PR #4, opened 2023-09-05
feat: make OpenAI model configurable

karvla:
I didn't have access to the gpt4 model via the OpenAI API and there was no way to switch to gpt-3.5-turbo without changing the code so here's that feature:>

### DIFF
--- a/gptme/cli.py
+++ b/gptme/cli.py
@@ -53,6 +53,7 @@ logger = logging.getLogger(__name__)
 
 
 LLMChoice = Literal["openai", "llama"]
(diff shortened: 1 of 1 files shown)


ErikBjare:
Very nice!

You should have GPT-4 access if you've been successfully billed on your account iirc. I think GPT-4 access should be given out pretty quickly nowadays.

I remember GPT-3.5 didn't do that well, kinda why I didn't make anything out of this before 😅

karvla:
Ah, I see!

Issue #7, opened 2023-09-06
Let the LLM continue/recover on its own


ErikBjare:
After interrupting the LLM on a completed code block to confirm execution, I often find myself just writing "continue", this could be done smarter to not be needed.

After the LLM stumbles into a problem, like a failing command, it should give a brief diagnosis and give it another try.

These behaviors are present in ChatGPT "Code Interpreter" / "Advanced Data Analysis", and very helpful in achieving a flow.

These behaviors are needed in testing anyway. An alternative is to let another LLM drive the conversation forward in pursuit of some end goal. 

Sidenote: It would be interesting if we could prompt it such that it thinks it is a "Code Interpreter", but maybe they use a LoRA or something to get rid of the "I don't have the ability to execute commands".

ErikBjare:
I just now realized you might be able to prompt it to add a "DONE" token or something when it's actually done with *all* codeblocks needed for the last instruction, and until that token is seen, we can just keep asking for completions (letting it attempt recovery etc.)

Might need to have some safeguards in case it never wants to output DONE.

ErikBjare:
Done in https://github.com/ErikBjare/gptme/commit/c23a5ce8b6cc4eb4dfc4c1470ec6374ae544e29e

PR #11, opened 2023-09-11
feat: progress on self-continue/recover

ErikBjare:
Progress on #7

Buggy af


ErikBjare:
Major refactor since, won't get merged. Might attempt again later.

PR #48, opened 2023-11-24
Refactor eval abstractions to be more general

AntonOsika:
Refactors into:

- Agent
- ExecutionEnv
- FileStore (used by both agent and SimpleExecutionEnv in this case)

tests remain the same.

I liked the breakdown a lot.

But I don't like that we would couple the execution env with the agent since it blocks people to use these abstractions if they want to e.g. sandbox the execution env.

I prefer that the agent produces and artifact (Files), that is testable in different execution envs.

### DIFF
--- a/eval/main.py
+++ b/eval/main.py
@@ -183,7 +183,7 @@ class FileStore:
 class Agent:
     def act(files: Files | None, command: str | None, prompt: str) -> Files:
         """
-        Returns a dict of actions to take.
+        Carries out the prompt and returns artifacts in the form of `Files`.
         """
         raise NotImplementedError
 
--- a/eval/main.py
+++ b/eval/main.py
@@ -276,7 +276,7 @@ def execute(test: ExecTest) -> TestResult:
 
     # check and collect results
     run_start = time.time()
-    env = ExecutionEnv()
+    env = SimpleExecutionEnv()
     stdout, stderr, exit_code = env.upload(files).run(test["run"])
     run_duration = time.time() - run_start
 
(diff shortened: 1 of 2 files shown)


eval/main.py
```diff
@@ -219,47 +260,24 @@ def run(self, command) -> tuple[str, str, int]:
...
-        return files
-
 
 def execute(test: ExecTest) -> TestResult:
     """
     Executes the code.
     """
     print(f"Running test {test['name']} with prompt: {test['prompt']}")
-    env = GPTMeExecEnv()
+    agent = GPTMe()
```
AntonOsika, line 269: Reviewing myself, nitpick:

I realize we don't need to instantiate an object in this case.

The agent.act() can just be a function.

eval/main.py
```diff
@@ -158,25 +151,59 @@ class ExecTest(TypedDict):
...
+                        content = f.read()
+                    except UnicodeDecodeError:
+                        content = "binary file"
+                    files[str(path.relative_to(self.working_dir))] = content
+        return files
+
+class Agent:
+    def act(files: Files | None, command: str | None, prompt: str) -> Files:
+        """
+        Returns a dict of actions to take.
```
AntonOsika:
```suggestion
        Carries out the prompt and returns artifacts in the form of `Files`.
```

ErikBjare:
Very nice!

Tests just failing due to no access to the `OPENAI_API_KEY` secret.

Merging.

ErikBjare:
Made some changes fixing the remaining issues: https://github.com/ErikBjare/gptme/commit/cefbbe868fdc2dfd6b3e311e23397b6639c2eb21

PR #49, opened 2023-11-26
feat: mirror working directory in shell and Python process

ErikBjare:


pyproject.toml
```diff
@@ -25,6 +25,7 @@ tomlkit = "^0.12.1"
 typing-extensions = "^4.8.0"
 platformdirs = "^3.11.0"
 ipython = "^8.17.2"
+bashlex = "^0.18"
```
ErikBjare, line 28: Secret sauce here, was really tricky to correctly split a bash script into individual commands such that cd can be intercepted.

PR #110, opened 2024-09-07
feat: added list_chats to chats tool, and cleaned up/refactored non-ToolSpec-using tools

ErikBjare:


ErikBjare:
@gptme write tests for gptme/tools/chats.py to tests/test_tool_chats.py (without the LLM summaries, so they run fast)

PR #121, opened 2024-09-17
feat: add support for XML-formatted tool calls

ErikBjare:
Fixes https://github.com/ErikBjare/gptme/issues/107

Anthropic and O1 seem to work well with XML, but GPT-4o not so much.

### DIFF
--- a/gptme/tools/base.py
+++ b/gptme/tools/base.py
@@ -19,8 +19,9 @@ logger = logging.getLogger(__name__)
 InitFunc: TypeAlias = Callable[[], Any]
 
 # tooluse format mode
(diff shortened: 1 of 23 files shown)


gptme/tools/base.py
```diff
@@ -18,7 +19,8 @@
 InitFunc: TypeAlias = Callable[[], Any]
 
 # tooluse format mode
-mode: Literal["markdown", "xml"] = "markdown"
+mode: Literal["markdown", "xml"] = "xml"
+exclusive_mode = True
```
ErikBjare: Make this configurable

PR #125, opened 2024-09-18
refactor: refactored Message, evals, and model metadata

ErikBjare:
- refactored Message into frozen dataclass
- moved openai model metadata into seperate file with script to update with gptme


gptme/prompts.py
```diff
@@ -28,19 +28,21 @@ def get_prompt(prompt: PromptType | str = "full", interactive: bool = True) -> M
     elif prompt == "short":
         msgs = prompt_short(interactive)
     else:
-        msgs = [Message("system", prompt)]
+        return Message("system", prompt)
```
ErikBjare:
```suggestion
        msgs = [Message("system", prompt)]
```

PR #134, opened 2024-09-22
Optimized code

U-C4N:
Optimized code


ErikBjare:
Removing all these comments is not acceptable, and the remaining changes seem to do nothing except change the style of expressions.

Won't be merged.

PR #135, opened 2024-09-23
feat: playing with basic rag scripts

ErikBjare:
Progress on https://github.com/ErikBjare/gptme/issues/59


ErikBjare:
We have https://github.com/ErikBjare/gptme-rag and the wrapping rag tool now

PR #165, opened 2024-10-07
Fix/windows compatibility

razaabbasnextgen:
> [!IMPORTANT]
> Fix Windows compatibility by conditionally importing `termios` in `chat.py` and `commands.py`.
> 
>   - **Compatibility**:
>     - Conditional import of `termios` in `chat.py` and `commands.py` for non-Windows platforms to ensure compatibility.
>   - **Misc**:
>     - Add placeholder `chat()` function in `chat.py` and `commands.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 5f57f4c4e14b565c6e0f3233fb1da3434fe8b982. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -8,6 +8,13 @@ import termios
 import urllib.parse
 from collections.abc import Generator
 from pathlib import Path
+import sys
+
+if sys.platform != 'win32':
+    import termios
+
+def chat():
+    pass
 
 from .commands import action_descriptions, execute_cmd
 from .config import get_workspace_prompt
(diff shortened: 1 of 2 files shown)


0xbrayo:
You would also need to add conditional code for windows everywhere that `sys` is called. Currently this does not do that

ErikBjare:
This diff looks messy, and doesn't come close to fixing the entire issue of native Windows/PowerShell support.

Closing due to predicting it will go stale.

Issue #166, opened 2024-10-07
Does not know about the system the user is running.


brayo-pip:
Found out editing the `about_user` prompt to include the distro I am using helps eliminate always specifying this info. The user could be prompted to type the OS/distro on first run. Or we could try to infer it automatically. Also on first run.

PR #171, opened 2024-10-08
feat: add platform info to the system prompt

brayo-pip:
> [!IMPORTANT]
> Adds system information to `about_user` prompt and introduces functions for system info retrieval in `gptme/prompts.py`.
> 
>   - **Behavior**:
>     - Adds system information to `about_user` prompt in `prompt_user()` in `gptme/prompts.py`.
>     - Introduces `prompt_systeminfo()` to generate system information prompt.
>     - Implements `get_system_distro()` to retrieve Linux distribution name.
>   - **Dependencies**:
>     - Adds `distro` package to `pyproject.toml` for Linux distribution information.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 9b4832fc04fe16ed2ac7bdee42a26b7be274bbfc. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/prompts.py
+++ b/gptme/prompts.py
@@ -211,6 +211,7 @@ def prompt_tools(examples: bool = True) -> Generator[Message, None, None]:
     prompt += "\n\n*End of Tools List.*"
     yield Message("system", prompt.strip() + "\n\n")
 
+
 def prompt_systeminfo() -> Generator[Message, None, None]:
     """Generate the system information prompt."""
     if platform.system() == "Linux":
@@ -225,15 +226,18 @@ def prompt_systeminfo() -> Generator[Message, None, None]:
     else:
         os_info = "unknown"
         os_version = ""
-    
+
     prompt = f"## System Information\n\n**OS:** {os_info} {os_version}".strip()
 
-    yield Message( "system", prompt,)
+    yield Message(
+        "system",
+        prompt,
+    )
+
 
 def get_system_distro() -> str:
     """Get the system distribution name."""
-    regex = r"^NAME=\"?([^\"]+)\"?"
-
+    regex = re.compile(r"^NAME=\"?([^\"]+)\"?")
     if os.path.exists("/etc/os-release"):
         with open("/etc/os-release") as f:
             for line in f:
@@ -241,6 +245,7 @@ def get_system_distro() -> str:
                 if matches:
                     return matches.string[matches.start(1) : matches.end(1)]
     return "Linux"
+
 
 document_prompt_function(interactive=True)(prompt_gptme)
 document_prompt_function()(prompt_user)
--- a/gptme/prompts.py
+++ b/gptme/prompts.py
@@ -6,7 +6,9 @@ When prompting, it is important to provide clear instructions and avoid any ambi
 """
 
 import logging
+import re
 import os
+import platform
 import subprocess
 from collections.abc import Generator, Iterable
 from typing import Literal
@@ -56,6 +58,7 @@ def prompt_full(interactive: bool) -> Generator[Message, None, None]:
     if interactive:
         yield from prompt_user()
     yield from prompt_project()
+    yield from prompt_systeminfo()
 
 
 def prompt_short(interactive: bool) -> Generator[Message, None, None]:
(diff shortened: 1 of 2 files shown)


0xbrayo:
@ellipsis-dev do another review

pyproject.toml
```diff
@@ -28,6 +28,7 @@ tomlkit = "*"
 typing-extensions = "*"
 platformdirs = "^4.3"
 lxml = "*"
+distro = "*"
```
ErikBjare: I don't think such simple checks justifies a dependency. 

Just use `platform`, you can call out to `lsb_release -d` (if it exists on system) if you want specific distro.

gptme/prompts.py
```diff
@@ -143,7 +145,7 @@ def prompt_user() -> Generator[Message, None, None]:
     """
     config_prompt = get_config().prompt
     about_user = config_prompt.get(
-        "about_user", "You are interacting with a human programmer."
+        "about_user", f"You are interacting with a human programmer. I am running a {get_system_info()} system."
```
ErikBjare: I don't think we want to add it to the user prompt like this.

I'd like the config to be platform-portable (I copy mine between machines), and this wouldn't work for users with an old config.

`prompt_user` is also not included in non-interactive mode, but it's useful in non-interactive mode too.

Maybe add a `prompt_systeminfo`?

pyproject.toml
0xbrayo: Will do!

0xbrayo: @ErikBjare My system does not have `lsb_release -d` installed on default expecting more devices probably don't have installed by default. A workaround suggested is to read the contents of `/etc/os-release`

0xbrayo:
@ellipsis-dev review

0xbrayo:
@ellipsis-dev review

ErikBjare:
I just noticed/remembered I added this little line in the shell tool instructions to make it aware of macOS systems: https://github.com/ErikBjare/gptme/blob/f25860243f26e107a5c18c89f538ce2f02549011/gptme/tools/shell.py#L53

gptme/prompts.py
```diff
@@ -208,8 +211,39 @@
...
+        os_version = platform.mac_ver()[0]
+    else:
+        os_info = "unknown"
+        os_version = "unknown"
+    
+    prompt = f"## System Information\n\n**OS:** {os_info}\n**Version:** {os_version}"
+
+    yield Message( "system", prompt,)
+
+def get_system_distro() -> str:
```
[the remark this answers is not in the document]
0xbrayo: @ellipsis-dev compile this regex

0xbrayo:
@ErikBjare Review this whenever :+1:

0xbrayo:
Closes #166

gptme/prompts.py
```diff
@@ -208,8 +211,39 @@ def prompt_tools(examples: bool = True) -> Generator[Message, None, None]:
...
+        os_version = platform.uname().release
+    elif platform.system() == "Windows":
+        os_info = "Windows"
+        os_version = platform.version()
+    elif platform.system() == "Darwin":
+        os_info = "macOS"
+        os_version = platform.mac_ver()[0]
+    else:
+        os_info = "unknown"
+        os_version = "unknown"
```
ErikBjare:
```suggestion
        os_version = ""
```

gptme/prompts.py
```diff
@@ -208,8 +211,39 @@ def prompt_tools(examples: bool = True) -> Generator[Message, None, None]:
...
+    
+    prompt = f"## System Information\n\n**OS:** {os_info}\n**Version:** {os_version}"
```
ErikBjare:
```suggestion
    prompt = f"## System Information\n\n**OS:** {os_info} {os_version}".strip()
```

ErikBjare (commented):
Looks good!

Just a minor nit on prompt formatting. Also recommend you set up `pre-commit` so that the code gets formatted correctly (I should add a formatting check to CI).

ErikBjare:
Nice, thanks!

ErikBjare:
Got this weird output when testing in #225: `Ubuntu 6.6.57-1-lts`

So we accidentally got the kernel version and not the distro version.

gptme/prompts.py
```diff
@@ -209,7 +212,43 @@ def prompt_tools(examples: bool = True) -> Generator[Message, None, None]:
...
+def get_system_distro() -> str:
+    """Get the system distribution name."""
+    regex = re.compile(r"^NAME=\"?([^\"]+)\"?")
+    if os.path.exists("/etc/os-release"):
+        with open("/etc/os-release") as f:
+            for line in f:
+                matches = re.search(regex, line)
+                if matches:
+                    return matches.string[matches.start(1) : matches.end(1)]
+    return "Linux"
```
ErikBjare, line 247: We should get the distro version here (`BUILD_ID=rolling` on Arch, `VERSION_ID="22.04"` on Ubuntu) instead of using `platform.uname().release`

0xbrayo: Will check into this. Also getting the same on my laptop, capturing the kernel version instead of the distro version.

PR #196, opened 2024-10-13
feat: support reading PDFs with browser tool

ErikBjare:
- [ ] Also support passing local PDFs on the CLI

### DIFF
--- a/gptme/tools/browser.py
+++ b/gptme/tools/browser.py
@@ -24,6 +24,8 @@ import re
 import shutil
 import subprocess
 import tempfile
+from pathlib import Path
+from time import sleep
 from typing import Literal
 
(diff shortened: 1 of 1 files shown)


gptme/tools/browser.py
```diff
@@ -81,13 +84,78 @@
 """.strip()
 
 
+def has_pdftotext():
+    return shutil.which("pdftotext") is not None
```
0xbrayo, line 88: can be rewritten using the get_installed_programs() function

ErikBjare:
I read on Twitter that Aider added PDF support in their latest release, but I can't find anything about it in their release notes or when searching PRs?

ErikBjare:
Looks like #1120 will supersede this, closing.

PR #201, opened 2024-10-16
♻️ refactor: refactor API key handling for clarity and consistency

Stark-X:
- decouple the "provider" and "model" in the config file
- unify the api-key, api-url keys for all llm provider in config file
- add config migration for compatibility

### DIFF
--- a/gptme/config.py
+++ b/gptme/config.py
@@ -19,11 +19,11 @@ class Config:
     env: dict
 
     def get_env(self, key: str, default: str | None = None) -> str | None:
-        """Gets an enviromnent variable, checks the config file if it's not set in the environment."""
+        """Gets an environment variable, checks the config file if it's not set in the environment."""
         return os.environ.get(key) or self.env.get(key) or default
 
     def get_env_required(self, key: str) -> str:
-        """Gets an enviromnent variable, checks the config file if it's not set in the environment."""
+        """Gets an environment variable, checks the config file if it's not set in the environment."""
         if val := os.environ.get(key) or self.env.get(key):
             return val
         raise KeyError(  # pragma: no cover
--- a/gptme/config.py
+++ b/gptme/config.py
@@ -2,9 +2,11 @@ import glob
 import logging
 import os
 from dataclasses import dataclass, field
+from enum import Enum
 from pathlib import Path
 
 import tomlkit
+from pydantic import BaseModel, Field, HttpUrl
 from tomlkit import TOMLDocument
 from tomlkit.container import Container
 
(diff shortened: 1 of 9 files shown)


gptme/config.py
```diff
@@ -156,6 +182,47 @@ def get_workspace_prompt(workspace: str) -> str:
...
+        set_config_value(f"env.{self._envvar_provider}", self.provider.value)
+        if not self._envvar_endpoint:
+            logger.warning(
+                f"Provider {self.provider.value} has no custom endpoint, skipping saving to config"
+            )
+            return
+        if self.model:
+            set_config_value(f"env.{self._envvar_model}", self.model)
+        if self.endpoint:
+            set_config_value(f"env.{self._envvar_endpoint}", str(self.endpoint))
```
ErikBjare, line 237: Not sure if I understand the motivation behind this, but I am not keen on diverging from the standard environment variable names.

ErikBjare:
What's the motivation behind this? I don't really get it.

gptme/config.py
Stark-X: I have been using some tools for wrapping the OpenAI SDK, which use their own config files to specify the API key and endpoint. For example, you can see how it's done in [TheR1D/shell_gpt](https://github.com/TheR1D/shell_gpt). I believe it's a good practice to make it easy for users to use the SDK without having to dig into the source code to figure out what "key" should be put into the config file.

Stark-X:
> What's the motivation behind this? I don't really get it.

In the "init.py" file, there is a mixture of logic regarding the purpose of the variable "mode". It may solely contain a provider like `openai`, or it may be a model name with a provider prefix such as `openai/gpt-4o`. I believe this violates the Single Responsibility Principle (SRP).

Furthermore, in some regions, the OpenAI API is blocked. Users can only access the OpenAI API with a VPN or a proxy site. Prior to this update, the only way I could use the proxy endpoint was to update the config file with the model variable prefixed by "local", for example `MODEL=local/gpt-4o-mini`, and then add `OPENAI_API_BASE=http://proxy.com/v1/` to the config file.

gptme/config.py
ErikBjare: >I believe it's a good practice to make it easy for users to use the SDK without having to dig into the source code to figure out what "key" should be put into the config file.

Users are prompted for a key on first start, and the provider for the key is auto-detected. I don't think this is easier, and it's a lot more code.

ErikBjare:
Oh, it didn't work because `OPENAI_API_BASE` has been renamed to `OPENAI_BASE_URL`: https://github.com/openai/openai-python/issues/745

You do not need to use `MODEL=local/gpt-4o-mini`, I just tested and `OPENAI_BASE_URL=http://proxy.com/v1/ gptme hello -m openai/gpt-4o-mini` works.

I will update docs to reflect this.

gptme/config.py
Stark-X: Some proxy sites do not provide the key following the pattern set by OpenAI. For example, they may not provide the API key with the `sk-` prefix. In this case, the "detection" mechanism will get stuck in an infinite loop, continuously prompting for an invalid-API-key error.

Stark-X:
> Oh, it didn't work because `OPENAI_API_BASE` has been renamed to `OPENAI_BASE_URL`: [openai/openai-python#745](https://github.com/openai/openai-python/issues/745)
> 
> You do not need to use `MODEL=local/gpt-4o-mini`, I just tested and `OPENAI_BASE_URL=http://proxy.com/v1/ gptme hello -m openai/gpt-4o-mini` works.
> 
> I will update docs to reflect this.

Exporting the environment variable should work well, but it's a bit cumbersome to do that for every shell session.

furthermore, Making the initialization process to be more out-of-the-box might be helpful in gaining more stars from users who cannot use OpenAI directly.

ErikBjare (commented):
I'm really sorry, but I just don't think these changes are an improvement as it complicates things significantly for a tiny benefit. I will not change the config format, as I like it how it is.

If there are UX problems you think we should address, I suggest you open issues for them so we can discuss acceptable solutions before you put time into work that might not get merged.

>Exporting the environment variable should work well, but it's a bit cumbersome to do that for every shell session.

You can set them in the config file if you want to avoid that, I fixed so it should work: https://github.com/ErikBjare/gptme/commit/30e3f01e7fa547dc94dd11328e3b74d2ff6e6a6b

PR #215, opened 2024-10-22
fix: informative message for agent when permission isn't granted

brayo-pip:


0xbrayo:
Helps prevent the loops of the LLM trying to rerun the same command over and over.

gptme/tools/shell.py
```diff
@@ -250,7 +250,7 @@ def execute_shell(
 
     print_preview(cmd, "bash")
     if not confirm("Run command?"):
-        yield Message("system", "Command not run")
+        yield Message("system", "User cancelled command.")
```
ErikBjare: I think this is more correct:

```suggestion
        yield Message("system", "User chose not to run command")
```

gptme/prompts.py
```diff
@@ -116,7 +116,7 @@ def prompt_gptme(interactive: bool) -> Generator[Message, None, None]:
     interactive_prompt = """
 You are in interactive mode. The user is available to provide feedback.
 You should show the user how you can use your tools to write code, interact with the terminal, and access the internet.
-The user can execute the suggested commands so that you see their output.
+The user can execute the suggested commands so that you see their output. If user cancels the execution, suggest a different approach.
```
ErikBjare: Not sure if this is needed with my other proposed change.

ErikBjare (commented):
Nice find! This was clearly bad prompting before, I can see why it would start looping...

gptme/prompts.py
0xbrayo: Will test and see.

0xbrayo: You are right dropped this commit

0xbrayo:
This is creating merge commits :(

PR #234, opened 2024-10-30
feat: copy to clipboard

### LINKED ISSUE #232: Copy  to clipboard suggested command
I sometimes have to copy the suggested command to run in a separate terminal tab. Because it's long lived, interactivity is not working or I just want to quickly edit it before running. Support for quick copy of command to clipboard when preview is displayed would make this way easier.

brayo-pip:
Closes #232

> [!IMPORTANT]
> Adds clipboard copy functionality with platform-specific handling, refactors utility functions, and integrates into shell and tmux modules.
> 
>   - **Clipboard Functionality**:
>     - Adds `set_copytext()` and `copy()` in `clipboard.py` for clipboard management.
>     - Supports Linux (`xclip`, `wl-copy`) and macOS (`pbcopy`).
>   - **Utility Functions**:
>     - Moves `get_system_distro()` and `get_installed_programs()` to `platform_utils.py` and `util.py`.
>     - Refactors `print_preview()` in `util.py` to support clipboard copying.
>   - **Integration**:
>     - Updates `execute_shell()` in `shell.py` and `execute_tmux()` in `tmux.py` to use `print_preview()` with clipboard copy option.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 28e11a45dddd63e0cd94a40a33f55c3cf41a0a8c. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/clipboard.py
+++ b/gptme/clipboard.py
@@ -19,7 +19,7 @@ def copy() -> bool:
         # check if xclip or wl-clipboard is installed
         installed = get_installed_programs(("xclip", "wl-copy"))
         if "wl-copy" in installed:
-            output = subprocess.run(["wl-copy"], input=text, text=True)
+            output = subprocess.run(["wl-copy"], input=text, text=True, check=True)
             if output.returncode != 0:
                 print("wl-copy failed to copy to clipboard.")
                 return False
--- a/gptme/clipboard.py
+++ b/gptme/clipboard.py
@@ -1,0 +1,45 @@
+import platform
+import subprocess
+
+
+text = ""
+
+
+def set_copytext(new_text: str):
+    global text
+    text = new_text
+
+
+def copy() -> bool:
+    """return True if successful"""
+    from .util import get_installed_programs
+
+    global text
+    if platform.system() == "Linux":
+        # check if xclip or wl-clipboard is installed
+        installed = get_installed_programs(("xclip", "wl-copy"))
+        if "wl-copy" in installed:
+            output = subprocess.run(["wl-copy"], input=text, text=True)
+            if output.returncode != 0:
+                print("wl-copy failed to copy to clipboard.")
+                return False
+            return True
+        elif "xclip" in installed:
+            output = subprocess.run(
+                ["xclip", "-selection", "clipboard"], input=text, text=True
+            )
+            if output.returncode != 0:
+                print("xclip failed to copy to clipboard.")
+                return False
+            return True
+        else:
+            print("No clipboard utility found. Please install xclip or wl-clipboard.")
+            return False
+    elif platform.system() == "Darwin":
+        output = subprocess.run(["pbcopy"], input=text, text=True)
+        if output.returncode != 0:
+            print("pbcopy failed to copy to clipboard.")
+            return False
+        return True
+
+    return False
--- a/gptme/tools/tmux.py
+++ b/gptme/tools/tmux.py
@@ -157,7 +157,7 @@ def execute_tmux(
     assert not args
     cmd = code.strip()
 
-    print_preview(f"Command: {cmd}", "bash")
+    print_preview(f"Command: {cmd}", "bash", copy=True)
     if not confirm(f"Execute command: {cmd}?"):
         yield Message("system", "Command execution cancelled.")
         return
--- a/gptme/util.py
+++ b/gptme/util.py
@@ -10,6 +10,8 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Any
 
+from .clipboard import copy, set_copytext
+
 import tiktoken
 from rich import print
 from rich.console import Console
@@ -132,9 +134,27 @@ def epoch_to_age(epoch, incl_date=False):
         )
 
 
-def print_preview(code: str, lang: str):  # pragma: no cover
+copiable = False
+
+
+def set_copiable():
+    global copiable
+    copiable = True
+
+
+def clear_copiable():
+    global copiable
+    copiable = False
+
+
+def print_preview(code: str, lang: str, copy: bool = False):  # pragma: no cover
     print()
     print("[bold white]Preview[/bold white]")
+
+    if copy:
+        set_copiable()
+        set_copytext(code)
+
     # NOTE: we can set background_color="default" to remove background
     print(Syntax(code.strip("\n"), lang))
     print()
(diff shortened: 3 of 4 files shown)


0xbrayo:
Closes #232

0xbrayo:
Experimented with a few approaches, `ask_execute` does not have access to the `cmd` variable. Passing it would involve rewriting all instances that `confirmFunc` is used. This was the simpler way I found to do it. Assumes that the `print_preview` is always called before `ask_execute` which is should always be the case. What do you think?

0xbrayo:
Noticed that other confirmation dialogs such as overwrite file have the "copy" option.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ python-dotenv = "^1.0.0"
 rich = "^13.5.2"
 tabulate = "*"
 pick = "^2.2.0"
+pyperclip = "^1.8.2"
```
ErikBjare: pyperclip seems abandoned: https://github.com/asweigart/pyperclip/issues/250 (not an endorsement of the alternative in the issue)

pyperclip also has `pytest` in their non-dev deps, which I found strange and undesireable.

gptme/util.py
```diff
@@ -131,6 +133,7 @@ def epoch_to_age(epoch):
 def print_preview(code: str, lang: str):  # pragma: no cover
     print()
     print("[bold white]Preview[/bold white]")
+    set_text(code)
```
ErikBjare: Won't really work for patches, but I guess that's ok for now.

poetry.lock
ErikBjare, line 1: Please use `poetry lock --no-update` when adding dependencies, so you don't accidentally update the locks of others at the same time.

ErikBjare (commented):
Looks good, nice workaround with the preview stuff, but some concerns about pyperclip.

0xbrayo:
~~I will try `clipboard` instead~~ Also not maintained. Will call the native copy commands directly.

gptme/tools/tmux.py
```diff
@@ -157,7 +157,7 @@ def execute_tmux(
     assert not args
     cmd = code.strip()
 
-    print_preview(f"Command: {cmd}", "bash")
+    print_preview(f"Command: {cmd}", "bash", copy=True)
```
ErikBjare, line 160: Nice

gptme/platform_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    return "Linux"
+
+
+@functools.lru_cache
+def get_installed_programs(candidates: tuple[str, ...]) -> set[str]:
+    installed = set()
+    for candidate in candidates:
+        if shutil.which(candidate) is not None:
+            installed.add(candidate)
+    return installed
```
ErikBjare: I don't think we need a new file for this, just drop it into `utils.py`

gptme/util.py
```diff
@@ -148,9 +168,16 @@ def ask_execute(question="Execute code?", default=True) -> bool:  # pragma: no c
     termios.tcflush(sys.stdin, termios.TCIFLUSH)  # flush stdin
 
     choicestr = f"[{'Y' if default else 'y'}/{'n' if default else 'N'}]"
+    copystr = r"\[c] to copy" if copiable else ""
     answer = console.input(
-        f"[bold bright_yellow on red] {question} {choicestr} [/] ",
+        f"[bold bright_yellow on red] {question} {choicestr}{copystr} [/] ",
     )
+    if copiable and "c" in answer.lower():
```
ErikBjare: What's the point of `set_copiable` here? Can't we just:

```suggestion
    if copy and answer.strip() == "c":
```

gptme/util.py
```diff
@@ -148,9 +168,16 @@ def ask_execute(question="Execute code?", default=True) -> bool:  # pragma: no c
...
+        if copy():
+            print("Copied to clipboard.")
+        if len(answer) > 1:
+            answer = answer.lower().strip("c")
```
ErikBjare: I don't get this part

gptme/util.py
ErikBjare: Nvm, thought it was a part of the other function.

ErikBjare: It should still be:

```suggestion
    if copiable and "c" == answer.strip().lower():
```

gptme/util.py
0xbrayo: I want to let the user type "yc" to copy and still execute the code. Or just c to only copy the code.

gptme/util.py
0xbrayo: See previous comment. Although I don't think it's that common of a use case

0xbrayo:
@ellipsis-dev review

0xbrayo:
@ErikBjare ping

gptme/util.py
```diff
@@ -309,3 +341,23 @@ def path_with_tilde(path: Path) -> str:
...
+        with open("/etc/os-release") as f:
+            for line in f:
+                matches = re.search(regex, line)
+                if matches:
+                    return matches.string[matches.start(1) : matches.end(1)]
+    return "Linux"
+
+
+@functools.lru_cache
+def get_installed_programs(candidates: tuple[str, ...]) -> set[str]:
```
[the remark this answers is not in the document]
ErikBjare: This is still true

gptme/clipboard.py
```diff
@@ -0,0 +1,45 @@
...
+def copy() -> bool:
+    """return True if successful"""
+    from .util import get_installed_programs
+
+    global text
+    if platform.system() == "Linux":
+        # check if xclip or wl-clipboard is installed
+        installed = get_installed_programs(("xclip", "wl-copy"))
+        if "wl-copy" in installed:
+            output = subprocess.run(["wl-copy"], input=text, text=True)
```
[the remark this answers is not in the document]
ErikBjare: This is probably a good idea

ErikBjare (commented):
See the remaining ellipsis comments (should fix mypy + lint warnings), looks good otherwise. Haven't actually tested it.

gptme/util.py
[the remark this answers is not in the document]
0xbrayo: Must have messed up the rebase

gptme/util.py
```diff
@@ -309,3 +341,23 @@ def path_with_tilde(path: Path) -> str:
     if path_str.startswith(home):
         return path_str.replace(home, "~", 1)
     return path_str
+
+def get_system_distro() -> str:
```
ErikBjare: This doesn't seem to be used

ErikBjare:
Nice, merged!

NikitaBeloglazov:
Actually, I don't understand, there is the reason for building something unknown if you could just take clipman (an alternative to pyperclip) which was mentioned above. Or is there something missing?

ErikBjare:
@NikitaBeloglazov There is no "building something unknown", the called executables should exist on their respective platforms. 

I don't want to add another dependency for something so simple.

Issue #111, opened 2024-09-08
Support codeblocks in codeblocks 


ErikBjare:
Right now gptme can't write codeblocks which contain other codeblocks, due to early stopping.

I worked on a thing that tried to count start tags and end tags so that nesting can be detected, but apparently didn't finish/test it because the issue remains.

ErikBjare:
I redid a lot of the stream-interruption stuff recently, and added the XML-format for tool calling which doesn't suffer from the issue.

TimeToBuildBob:
This is now addressed by PR gptme/gptme#903 which handles nested codeblocks with the same language tag. The PR adds proper depth tracking during extraction. All CI checks passing, ready for review.

PR #251, opened 2024-11-10
Trigger tool detection only on complete lines to help nested code block start detection

jrmi:
TL;DR; This MR fixes the bug when you have a markdown code block nested in another tooluse code block in streaming mode.

I think it's related to this issue #111 

# To reproduce

- Start gptme
- Submit the following prompt: `Create the file foo.md and add a small explanation on how works the async in python with a code example`

The response of this prompt looks something like that:

```markdown
    [Some introduction text]

    ```save foo.md
    [In file explanation]

    ```python
    [python code block] 
    ```
    [In file conclusion]
    ```
    [end of llm answer]
```

If you are using streaming mode, the tooluse detection will stop immediately after the three back quotes preceding the `python` code block type because the check is executed after every single char.

# Solution proposed

The idea of this MR is to wait for the end of a line (i.e. a `\n` char) to execute the tool check. That way, we know the next three back quotes aren't the end of the code block as they are followed by the code block type (`python` here) and the tool use detection algorithm can handle that.

It's not perfect as it still won't detect if there is no code block type but it should work in more situation waiting for a more robust solution. 

Also I think it's a good idea to wait for the end of the line to try to detect the tools. I don't think we have benefits to execute it on every single char unless I missed something.

### DIFF
--- a/gptme/llm.py
+++ b/gptme/llm.py
@@ -96,12 +96,15 @@ def _reply_stream(messages: list[Message], model: str) -> Message:
             # need to flush stdout to get the print to show up
             sys.stdout.flush()
 
-            # TODO: make this more robust/general, maybe with a callback that runs on each char/chunk
-            # pause inference on finished code-block, letting user run the command before continuing
-            tooluses = list(ToolUse.iter_from_content(output))
-            if tooluses and any(tooluse.is_runnable for tooluse in tooluses):
-                logger.debug("Found tool use, breaking")
-                break
+            # Trigger the tool detection only if the line is finished.
+            # Helps to detect nested start code blocks.
+            if char == "\n":
+                # TODO: make this more robust/general, maybe with a callback that runs on each char/chunk
+                # pause inference on finished code-block, letting user run the command before continuing
+                tooluses = list(ToolUse.iter_from_content(output))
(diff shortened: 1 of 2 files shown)


ErikBjare:
Very good fix/workaround/improvement!

Thanks a lot :)

ErikBjare:
This was certainly an improvement!

I stumbled into it failing to write this though:

    Sure, let's save that:

    ```save rag.py
        ...
        return ToolSpec(
	        name="rag",
	        desc="RAG (Retrieval-Augmented Generation) for context-aware assistance",
	        instructions="""
	        Use RAG to index and search project documentation.
	
	        Commands:
	        - index [paths...] - Index documents in specified paths
	        - search <query> - Search indexed documents
	        - status - Show index status
	        """,
	        examples="""
	        > User: Index the current directory
	        ```rag index```
	
	        > User: Search for documentation about functions
	        ```rag search function documentation```
	
	        > User: Show index status
	        ```rag status```
	        """,
	        block_types=["rag"],
	        execute=execute_rag,
	        available=True,
	    )
	
	tool = init_rag()
	```

    (...message keeps going, doesn't get interrupted to execute save tool)

I assume this is due to the presence of ``` in the file, although they are indented, so they really shouldn't.

Issue #153, opened 2024-10-05
Fails if trying to apply patch without file name


brayo-pip:
<details>
The error is due to the incorrect use of `ifeq` and `else` in the `Makefile`. The `ifeq` directive should be properly
indented, and the `else` and `endif` should align with the `ifeq`. Let's correct this by adjusting the indentation.

Here's the corrected section of the `Makefile`:

```makefile
install:
    ifeq ($(SKIP_INSTALL),true)
        @echo "Skipping npm install"
    else
        npm ci
        (cd aw-client-js; npm ci; npm run compile)
    endif
```

Let's apply this patch to the `Makefile`.

```patch
Makefile
<<<<<<< ORIGINAL
install:
    ifeq ($(SKIP_INSTALL),true)
    @echo "Skipping npm install"
else
    npm ci
    (cd aw-client-js; npm ci; npm run compile)
endif
=======
install:
    ifeq ($(SKIP_INSTALL),true)
        @echo "Skipping npm install"
    else
        npm ci
        (cd aw-client-js; npm ci; npm run compile)
    endif
>>>>>>> UPDATED
```

Skipped 1 hidden system messages, show with --show-hidden
--- ^^^ past messages ^^^ ---
User: fix it
Assistant: Let's apply the patch to correct the indentation issue in the `Makefile`.

```patch
Makefile
<<<<<<< ORIGINAL
install:
        ifeq ($(SKIP_INSTALL),true)
        @echo "Skipping npm install"
else
    npm ci
        (cd aw-client-js; npm ci; npm run compile)
endif
=======
install:
        ifeq ($(SKIP_INSTALL),true)
                @echo "Skipping npm install"
        else
                npm ci
                (cd aw-client-js; npm ci; npm run compile)
        endif
>>>>>>> UPDATED
```
                                                                                                                     Traceback (most recent call last):
  File "/home/brayo/.local/bin/gptme", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
[...]

PR #252, opened 2024-11-10
fix: prompt to reinforce correct placement of filename parameter in patch tool

jrmi:
For some reasons, Chatgpt place the patch `$FILENAME` parameter on the line after the first code block line. This breaks the patch tool.

In this MR I tried to enforce the LLM to add the patch filename parameter after the code block start by improving the tool prompt. it helps to prevent issue #153 

It tested multiple versions and that's the version with the best results, all of my tests were successful with it.

Using the [tool API](!219) should fix that in a better way, but in the meantime we have a quick fix.

----

> [!IMPORTANT]
> Updates `patch.py` instructions to enforce `$FILENAME` parameter placement on the same line as the code block start, preventing issues like #153.
> 
>   - **Instructions Update**:
>     - In `patch.py`, updated `instructions` to enforce `$FILENAME` parameter placement on the same line as the code block start.
>     - Aims to prevent issues like #153 by improving tool prompt.
>   - **Testing**:
>     - Multiple versions tested; current version yields successful results.
>   - **Future Work**:
>     - Suggests using the tool API for a more robust solution.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for e76bf5dddb64d06eed18f8fd0880e1d21e7f21e5. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/patch.py
+++ b/gptme/tools/patch.py
@@ -22,6 +22,8 @@ Try to keep the patch as small as possible. Avoid placeholders, as they may make
 To keep the patch small, try to scope the patch to imports/function/class.
 If the patch is large, consider using the save tool to rewrite the whole file.
 
+The $FILENAME parameter MUST be on the same line as the code block start, not on the line after.
+
 The patch block should be written in the following format:
 
 {ToolUse("patch", ["$FILENAME"], '''


ErikBjare:
Nice!

Which models did you notice this with?

jrmi:
> Nice!
> 
> Which models did you notice this with?

I usually use `openai/gpt-o4` model. It's not something that happens immediately on small basic examples. I believe it happens if the log is big enough. To test it I ask the model to read a documentation file in a code repository and to follow the instructions to do something (and the instructions contain patching some files obviously). My prompt looks like `Could you read the file 'add-an-element.md' and follows the instruction in it to create the new 'PositionnedContainer' element`.

jrmi:
What do you think about it?

ErikBjare:
> I believe it happens if the log is big enough. 

This makes sense, I've had this in the past too with long conversations. Not recently with Claude though.

Let's merge it!

PR #254, opened 2024-11-12
refactor: group all provider/model related logic into `llm` directory

jrmi:
This is a work in progress version of the refactoring I suggested. It's not well tested yet and there are probably a few regressions here and there but I wanted to show you the idea before going further.

I've introduced the concept of `ModelManager`. There is one implementation per API and they deal with the small differences we can have between model. The manager is instantiated with the model so it's available everywhere.

The idea is that everything that is specific to a model or a provider should be in the provider directory.

Let me know what you think.

----

> [!IMPORTANT]
> Refactor codebase by moving provider/model logic to `providers` directory and introducing `ModelManager` for model-specific operations.
> 
>   - **Refactoring**:
>     - Move provider and model-related logic to `providers` directory.
>     - Introduce `ModelManager` for handling model-specific logic, instantiated with the model.
>   - **Imports**:
>     - Update imports in `chat.py`, `commands.py`, `init.py`, `message.py`, `reduce.py`, `server/api.py`, `tools/chats.py`, `tools/youtube.py`, and `test_server.py` to reflect new directory structure.
>   - **Functionality**:
>     - Add `supports_file` and `supports_streaming` properties to `ModelManager`.
>     - Implement `prepare_file` method in `AnthropicModelManager` and `OpenaiModelManager` for handling file preparation.
>   - **Miscellaneous**:
>     - Update logic in `llm.py`, `llm_anthropic.py`, and `llm_openai.py` to use `ModelManager` for model-specific operations.
>     - Adjust `get_model()` usage across various files to utilize the new `ModelManager` structure.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 12a90672845c49961e8ca899c12df196e96b59bf. It will automatically update as commits are pushed.</sup>


gptme/providers/manager.py
```diff
@@ -0,0 +1,17 @@
+from typing import TYPE_CHECKING
+from dataclasses import dataclass
+
+if TYPE_CHECKING:
+    from .models import ModelMeta
+
+
+@dataclass()
+class ModelManager:
```
ErikBjare: Why not just use ModelMeta?

gptme/chat.py
```diff
@@ -12,10 +12,10 @@
 from .constants import PROMPT_USER
 from .init import init
 from .interrupt import clear_interruptible, set_interruptible
-from .llm import reply
+from .providers.llm import reply
```
ErikBjare: I thought we said not to create a module?

I think `gptme.providers.llm` is unnecessarily verbose, I much prefer the current `gptme.llm`.

gptme/message.py
```diff
@@ -126,39 +124,20 @@ def _content_files_list(
...
-                        "image_url": {"url": f"data:{media_type};base64,{data}"},
-                    }
-                )
-            else:
-                # Storage/wire format (keep files in `files` list)
-                # Do nothing to integrate files into the message content
-                pass
+            file_content = get_model().manager.prepare_file(media_type, data)
+            if file_content:
+                content.append(file_content)
```
ErikBjare: This is not valid, you cannot rely on `get_model()` returning the correct model for a given call to `to_dict` or `_content_files_list`.

gptme/message.py
```diff
@@ -126,39 +124,20 @@ def _content_files_list(
...
 
         return content
 
-    def to_dict(self, keys=None, provider: Provider | None = None) -> dict:
+    def to_dict(self, keys=None) -> dict:
```
ErikBjare, line 76: You need to pass the provider here, you cannot infer it with `get_model()`

gptme/message.py
```diff
@@ -126,39 +124,20 @@ def _content_files_list(
...
-                content.append(
-                    {
-                        "type": "image_url",
-                        "image_url": {"url": f"data:{media_type};base64,{data}"},
-                    }
-                )
-            else:
-                # Storage/wire format (keep files in `files` list)
-                # Do nothing to integrate files into the message content
-                pass
```
ErikBjare, line 151: This comment in particular is super important, you cannot ignore this!

ErikBjare (commented):
Overall, I don't think these changes make much sense. You seem to have underestimated things, as you cannot just use `get_model` everywhere.

gptme/providers/llm_anthropic.py
```diff
@@ -13,6 +16,34 @@
...
+            "type": "image",
+            "source": {
+                "type": "base64",
+                "media_type": media_type,
+                "data": data,
+            },
+        }
+
+
+def guess_model_from_config(config):
```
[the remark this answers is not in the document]
ErikBjare: I agree with this review, breaking this up makes no sense and makes it more annoying to edit.

gptme/message.py
jrmi: Comment restored

gptme/chat.py
```diff
@@ -56,8 +56,12 @@ def chat(
     # init
     init(model, interactive, tool_allowlist)
 
-    if model and model.startswith("openai/o1") and stream:
-        logger.info("Disabled streaming for OpenAI's O1 (not supported)")
+    if get_model().manager.supports_streaming and stream:
```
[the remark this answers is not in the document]
jrmi: Good catch. fixed.

gptme/chat.py
jrmi: Fixed

gptme/providers/llm_anthropic.py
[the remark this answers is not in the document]
jrmi: I guess AI reviewer is not smart enough yet to detect code splitting to improve separation of concerns and encapsulation but that's fine, I proposed a new version.

gptme/providers/manager.py
jrmi: I wanted to change less code but happy to implement it use ModelMeta directly, what do you think of it now?

gptme/message.py
jrmi: To be fair, I knew you would notice that :smile:. As I said in PR description it's more like a POC for now, but yeah I noticed that there are a few places where the logic is a bit different. Could we consider to return a " fake model" in these situations?  May be you could give me more details to help me understand why we can't always use get_model() so that I can come up with a solution for that as well? I think it a good idea to have something consistent throughout the code.

gptme/message.py
jrmi: As i said in the previous comment, happy to improve that when I have more inputs.

jrmi (commented):
> Overall, I don't think these changes make much sense. You seem to have underestimated things, as you cannot just use get_model everywhere.

The idea behind my changes is to establish a clear separation between provider/model-specific code and the rest. This would help contributors easily identify where they can add provider/model-specific code and where they shouldn't. I'm confident we can find a solution that aligns both with your vision and the changes I've proposed. Wouldn't it be beneficial to remove all these scattered specific codes and improve encapsulation?

I may have underestimated some aspects, given that I don't know the codebase and all its branches as well as you do. However, I'm eager to work on this because I truly believe it would enhance the readability and maintainability of the code in the long term.

gptme/llm/__init__.py
```diff
@@ -17,19 +17,54 @@
...
+    if api_key.startswith("sk-ant-"):
+        return api_key, "anthropic", "ANTHROPIC_API_KEY"
+    elif api_key.startswith("sk-or-"):
+        return api_key, "openrouter", "OPENROUTER_API_KEY"
+    elif api_key.startswith("sk-"):
+        return api_key, "openai", "OPENAI_API_KEY"
+
+    return None
+
+
```
ErikBjare: Nit: put these at the bottom of the file, they don't need obvious docstrings, but we do want them typed:

```suggestion
def guess_model_from_config() -> str | None:
    config = get_config()
    if config.get_env("OPENAI_API_KEY"):
        console.log("Found OpenAI API key, using OpenAI provider")
        return "openai"
    elif config.get_env("ANTHROPIC_API_KEY"):
        console.log("Found Anthropic API key, using Anthropic provider")
        return "anthropic"
    elif config.get_env("OPENROUTER_API_KEY"):
        console.log("Found OpenRouter API key, using OpenRouter provider")
        return "openrouter"
    return None


def get_model_from_api_key(api_key: str) -> tuple[str, Provider, str]:
    if api_key.startswith("sk-ant-"):
        return api_key, "anthropic", "ANTHROPIC_API_KEY"
    elif api_key.startswith("sk-or-"):
        return api_key, "openrouter", "OPENROUTER_API_KEY"
    elif api_key.startswith("sk-"):
        return api_key, "openai", "OPENAI_API_KEY"
    return None


```

gptme/llm/llm_anthropic.py
```diff
@@ -13,6 +15,20 @@
...
+
+    def prepare_file(self, media_type, data):
+        return {
+            "type": "image",
+            "source": {
+                "type": "base64",
+                "media_type": media_type,
+                "data": data,
+            },
+        }
```
ErikBjare: Why inherit by provider from ModelMeta? Use a dataclass `field`. 
Not every model from the same provider have the same capabilities (Claude 3.5 Haiku doesn't have vision support, for example).

I think `supports_vision` is a better name.

I don't think it should be the job of `ModelMeta` to prepare the file, it should be the job of the provider implementation when calling `msgs2dicts`.

gptme/message.py
```diff
@@ -126,39 +124,24 @@ def _content_files_list(
...
-                # OpenAI format
-                content.append(
-                    {
-                        "type": "image_url",
-                        "image_url": {"url": f"data:{media_type};base64,{data}"},
-                    }
-                )
+            file_content = get_model().prepare_file(media_type, data)
+            if file_content:
+                content.append(file_content)
```
ErikBjare: I would remove this from here entirely, and move it into each provider.

gptme/message.py
```diff
@@ -126,39 +124,24 @@ def _content_files_list(
...
+    def to_dict(self, keys=None) -> dict:
         """Return a dict representation of the message, serializable to JSON."""
         content: str | list[dict[str, Any]]
-        if provider in ProvidersWithFiles:
+
+        if get_model().supports_file:
             # OpenAI/Anthropic format should include files in the content
             # Some otherwise OpenAI-compatible providers (groq, deepseek?) do not support this
-            content = self._content_files_list(provider)
+            content = self._content_files_list()
```
ErikBjare: Same as above, get rid of this from `message.py` and move it into provider logic after running msgs2dicts.

gptme/message.py
```diff
@@ -352,9 +335,9 @@ def toml_to_msgs(toml: str) -> list[Message]:
     ]
 
 
-def msgs2dicts(msgs: list[Message], provider: Provider) -> list[dict]:
+def msgs2dicts(msgs: list[Message]) -> list[dict]:
     """Convert a list of Message objects to a list of dicts ready to pass to an LLM."""
-    return [msg.to_dict(keys=["role", "content"], provider=provider) for msg in msgs]
+    return [msg.to_dict(keys=["role", "content"]) for msg in msgs]
```
ErikBjare: with the other suggested changes, removing `provider=` here becomes the right call, but we need to let through `files`.

```py
def msgs2dicts(msgs: list[Message]) -> list[dict]:
    """Convert a list of Message objects to a list of dicts to be processed by providers"""
    return [msg.to_dict(keys=["role", "content", "files"]) for msg in msgs]
```

gptme/chat.py
ErikBjare: I like this!

gptme/message.py
ErikBjare: Left some comments about this on the rest of the PR, we can make it work!

gptme/llm/__init__.py
jrmi: Fixed.

gptme/llm/llm_anthropic.py
jrmi: Removed the class inheritance and using a plain property.

I removed the`supports_vision` property as we don't need it anymore after the file refactoring.

gptme/message.py
jrmi: I did something. Let me know if it's what you had in mind.

jrmi:
Hey @ErikBjare I think we are going in the right direction :rocket: . I applied your suggestions and it's much better and it makes a lot of sense. I hope I understood everything. My only doubt is about the code duplication in the file handling in each providers. We could do something, but I'm not sure it's necessary/a good idea. Let me know what you think.

Same as before, it's not polished, i think one test is failing but I'll fix that when I'm sure what I've done is ok with you.

> A few ideas on how we can get this merged, it will really be an improvement and I appreciate your efforts.
> Sorry if I was grumpy before, I've been sick this week.

Sure, no problem. Happy to help.

ErikBjare:
Rebased and force pushed.

Typechecks and tests passing locally :tada:

ErikBjare:
Going to go ahead and merge this, I'll fix any remaining issues in follow-up PRs.

Thanks a lot for pulling through with this @jrmi! :muscle: 

You dropped this: :crown:

ErikBjare (approved):

PR #257, opened 2024-11-15
fix: improved support and testing for nested codeblocks

ErikBjare:
Fixing the issue I wrote about in https://github.com/ErikBjare/gptme/pull/251#issuecomment-2478620841


ErikBjare:
Outdated, the code has seen a lot of improvements since.

PR #264, opened 2024-11-22
docs: add guidance about nested codeblocks and language tags

ErikBjare:
Add explicit guidance about using language tags in nested codeblocks to prevent parsing issues.

----

> [!IMPORTANT]
> Adds guidance on using language tags in nested codeblocks in `prompt_gptme()` to prevent parsing issues.
> 
>   - **Documentation**:
>     - Adds guidance on using language tags in nested codeblocks in `prompt_gptme()` in `gptme/prompts.py`.
>     - Specifies that both outer and inner blocks need language tags to prevent premature cut-off during parsing.
>     - Provides an example of correct usage with `python` and `shell` tags.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for a0019b839beab816875fb2b2dac187d0f33c8fad. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/prompts.py
+++ b/gptme/prompts.py
@@ -103,6 +103,17 @@ When suggesting code changes, prefer applying patches over examples. Preserve co
 Use the patch tool to edit existing files, or the save tool to overwrite.
 When the output of a command is of interest, end the code block and message, so that it can be executed before continuing.
 
+When using nested codeblocks (codeblocks within codeblocks), always specify a language tag for both the outer and inner blocks.
+This is critical for correct parsing - without language tags, nested blocks may be cut off prematurely.
+For example:
+```python
+print("This is the outer block")
+# Inner block needs a language tag:
+```shell
+echo "This is the inner block"
+```
+```
+
 Do not use placeholders like `$REPO` unless they have been set.
 Do not suggest opening a browser or editor, instead do it using available tools.
 


ErikBjare:
Seems to break parsing when building docs

gptme/prompts.py
```diff
@@ -103,6 +103,17 @@ def prompt_gptme(interactive: bool) -> Generator[Message, None, None]:
...
+When using nested codeblocks (codeblocks within codeblocks), always specify a language tag for both the outer and inner blocks.
+This is critical for correct parsing - without language tags, nested blocks may be cut off prematurely.
+For example:
+```python
+print("This is the outer block")
+# Inner block needs a language tag:
+```shell
+echo "This is the inner block"
+```
+```
```
ErikBjare, line 115: I'm not sure, but I've had a hunch and maybe even the experience (it has happened, just not sure if this prompt was the cause) that this prompt makes LLMs more prone to outputting incorrectly nested toolcalls, such as:

	```shell
	cat ...
	
	```save test.txt
	asd
	```
	```

Which of course isn't valid, because a markdown block isn't valid bash.

PR #280, opened 2024-11-24
Draft: Add tools api support

jrmi:
This MR is not ready for review yet, but I have a first working version with a limited set of tools. Let me know what you think.

### Old comment

What I've done so far:

- Added a new CLI flag (-f or --function-call) to activate tools. I marked this option as experimental, and I believe it should remain experimental for a while.
- Ported 4 tools to the function calling. I want to address the questions before proceeding further.
- In function calling mode, the tools are no longer prompted (but something replaces it, see below).

We need a new description for the tool that differs from the existing one, as the tool works differently. We also need a new execute function, as the parameter system is different. Even the examples are no longer working since they use the current syntax.

I think we have two options:

- Stick with the same objects and simply add the new properties (better naming suggestions are welcome :smiley:). This is what I started to implement.
- Create a new class dedicated to the new objects and register these instead of the existing ones when using the function calling feature. I think it also makes sense because we hardly reuse anything from the prompted tools.

I'm unsure what we should do with the "aliases" tools, those that don't have an execute function because they're just an "alias" to another tool, like the read function. I proposed a separate description for these tools as well, and in the initial prompt, they are still included in a section that gives the model some special use cases for these tools. These tools are no longer registered, otherwise, the model attempts to call them. I think we could get rid of these if we provided a more detailed description of the aliased tools, but I might be wrong.

On a side note, I tried using the `lama3.2:1B` model (which is a small model) with function calling, and it was able to call a tool, which wasn't the case with the prompted tools. However, OpenAI doesn't detect it as a tool call, so the raw JSON is returned. We could parse it to get an actual tool call, but that is something for later.

I've added the `json_repair` dependency to parse the parameter JSON because there is a high chance that the model generates something that isn't parsable, as it doesn't protect necessary characters. This dependency was created for this purpose.

@ErikBjare, I'd love to have your thoughts on this before I continue.

----

> [!IMPORTANT]
> Adds tools API support with function call execution, updating chat loop, CLI, and tool specifications.
> 
>   - **Behavior**:
>     - Adds `use_function_call` parameter to `chat()` and `step()` in `chat.py` to enable function call usage.
>     - Updates `main()` in `cli.py` to include `--tool-format` option for CLI.
>     - Modifies `reply()` in `llm/__init__.py` to handle tools in responses.
>   - **Tool Integration**:
>     - Updates `ToolSpec` in `tools/base.py` to include `execute_fn` for function call execution.
>     - Implements `APIToolUse` class in `tools/base.py` for handling API tool calls.
>     - Updates `tool_save` and `tool_append` in `tools/save.py` to support function calls.
>     - Updates `tool` in `tools/shell.py` to support function calls.
>   - **Misc**:
>     - Adds `json-repair` dependency in `pyproject.toml` for JSON handling.
>     - Updates `prompt_tools()` in `prompts.py` to handle function call descriptions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 66e9f6209e65091ade2464105a32e843a4cca8e5. It will automatically update as commits are pushed.</sup>


gptme/cli.py
```diff
@@ -103,6 +105,14 @@
     multiple=True,
     help=f"Comma-separated list of tools to allow. Available: {available_tool_names}.",
 )
+@click.option(
+    "-f",
+    "--fonction-call",
```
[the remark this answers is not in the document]
jrmi: Fixed.

gptme/tools/base.py
```diff
@@ -56,12 +96,15 @@ class ToolSpec:
     name: str
     desc: str
     instructions: str = ""
+    function_description: str = ""
```
ErikBjare: I think we should somehow merge this with the already existing `instructions` property.

In the cases where `instructions` contain format-specific instructions, we could move such instructions into format-specific prompting under a new `instructions_format` dict with keys like `markdown`.

gptme/tools/base.py
```diff
@@ -240,3 +302,37 @@ def _to_xml(self) -> str:
         args = " ".join(self.args)
         args_str = "" if not args else f" args='{args}'"
         return f"<tool-use>\n<{self.tool}{args_str}>\n{self.content}\n</{self.tool}>\n</tool-use>"
+
+
+@dataclass(frozen=True)
+class APIToolUse(BaseToolUse):
+    tool: str
+    args: dict[str, str]
```
ErikBjare: I think we should be able to solve this without subclassing. We could support both `args` (list) and `kwargs` (dict) in the same ToolUse class.

gptme/tools/base.py
```diff
@@ -33,9 +36,46 @@ class ExecuteFuncMsg(Protocol):
     def __call__(self, code: str, args: list[str], confirm: ConfirmFunc) -> Message: ...
 
 
+class ExecuteFn(Protocol):
+    def __call__(
+        self, args: dict[str, str], confirm: ConfirmFunc
+    ) -> Generator[Message, None, None] | Message: ...
+
+
```
ErikBjare: This is confusingly similarly named, we should find a way to reuse the `ExecuteFunc` protocol, probably by using it and mapping the parameters.

gptme/tools/save.py
```diff
@@ -18,10 +20,21 @@
 If the current directory changes, the path will be relative to the new directory.
 """.strip()
 
+function_description = """
+To write to a file, use a code block with the language tag: `save <path>`
+
+The path can be relative to the current directory, or absolute.
+If the current directory changes, the path will be relative to the new directory.
+""".strip()
```
ErikBjare: As the comment suggests on line 15 above, we should refactor this to not mention markdown-specific instructions in the mail instructions, and move them into a separate `instructions_format` dict in the ToolSpec.

gptme/tools/base.py
```diff
@@ -240,3 +302,37 @@ def _to_xml(self) -> str:
...
+        """Executes a tool-use tag and returns the output."""
+        # noreorder
+        from . import get_tool  # fmt: skip
+
+        tool = get_tool(self.tool)
+
+        if tool and (tool.execute_fn or tool.execute):
+            try:
+                if tool.execute_fn:
+                    ex = tool.execute_fn(self.args, confirm)
```
ErikBjare:
```suggestion
                    content = self.args.get("content")
                    path = self.args.get("filepath") or self.args.get("path")
                    ex = tool.execute(content, [path] if path else [], confirm)
```

gptme/tools/base.py
```diff
@@ -90,42 +133,13 @@ def __eq__(self, other):
...
+        return bool(self.execute)
+
 
 @dataclass(frozen=True)
-class ToolUse:
+class BaseToolUse:
     tool: str
-    args: list[str]
-    content: str
-    start: int | None = None
```
ErikBjare: I really don't think we need to subclass for this.

ErikBjare (commented):
Good stuff!

jrmi:
> Good stuff!

Thanks for the first review. Everything makes sense and I tried to go in that direction as much as possible. As bonus you I have now implemented the `--tool-format` that can have the value `markdown`, `xml`, `tool` to change the tool format. 

I removed all the extra subclasses and there is now a new `instructions_format` dict to the `ToolSpec` class.

Again, before going further let me know what you think.

Good news: now that ollama supports the tools in stream mode, I was able to use basics tools with  the small `llama3.2:1B` model.

gptme/tools/base.py
jrmi: Good idea. Done.

gptme/tools/base.py
jrmi: Done.

gptme/tools/base.py
jrmi: I wasn't happy about the naming. Removed now.

gptme/tools/save.py
jrmi: Did it.

gptme/tools/base.py
jrmi: I added a compat mode that should work for most of the tool like you suggested but I believe that we should migrate them all we proper argument when it's possible and it makes sense because it helps the model to use the tool properly.

gptme/tools/base.py
jrmi: Removed.

gptme/tools/base.py
```diff
@@ -226,17 +324,25 @@ def _iter_from_xml(cls, content: str) -> Generator["ToolUse", None, None]:
             logger.warning(f"Failed to parse XML content: {e}")
             return
 
-    def to_output(self) -> str:
-        if mode == "markdown":
+    def to_output(self, tool_type: ToolFormat = "markdown") -> str:
```
ErikBjare:
```suggestion
    def to_output(self, tool_format: ToolFormat = "markdown") -> str:
```

gptme/tools/save.py
```diff
@@ -116,26 +146,58 @@ def execute_append(
...
     name="save",
     desc="Write text to file",
     instructions=instructions,
+    instructions_format=instructions_format,
     examples=examples,
     execute=execute_save,
     block_types=["save"],
+    parameters=[
+        Parameter(
+            name="filepath",
```
ErikBjare: I prefer the shorter `path` (many occurrences)

```suggestion
            name="path",
```

gptme/tools/shell.py
```diff
@@ -39,19 +42,25 @@
...
+The given command will be executed in a stateful bash shell. 
+The shell tool will respond with the output of the execution.
 
 These programs are available, among others:
 {shell_programs_str}
 """.strip()
 
-examples = f"""
+instructions_format = {
+    "markdown": "Do not use EOF/HereDoc syntax to send multiline commands, as the assistant will not be able to handle it.",
```
ErikBjare: This isn't markdown-specific, unfortunately.

gptme/tools/base.py
```diff
@@ -56,12 +105,15 @@ class ToolSpec:
     name: str
     desc: str
     instructions: str = ""
-    examples: str = ""
+    instructions_format: dict[str, str] = field(default_factory=dict)
+    examples: str | Callable = ""
```
ErikBjare:
```suggestion
    examples: str | Callable[[str], str] = ""
```

gptme/tools/base.py
```diff
@@ -90,12 +142,39 @@ def __eq__(self, other):
             return False
         return self.name == other.name
 
+    def is_executable(self):
+        return bool(self.execute)
+
+    def get_instructions(self, tool_format: ToolFormat):
+        if tool_format in self.instructions_format:
+            return "\n".join([self.instructions, self.instructions_format[tool_format]])
```
ErikBjare: Assuming both are stripped strings:

```suggestion
            return "\n\n".join([self.instructions, self.instructions_format[tool_format]])
```

gptme/tools/save.py
```diff
@@ -5,64 +5,86 @@
...
+Use a code block with the language tag: `append <path>` 
+to append the code block content to the file at the given path.""".strip(),
+}
+
+
+def examples(tool_type):
+    return f"""
 > User: write a hello world script to hello.py
-{ToolUse("save", ["hello.py"], 'print("Hello world")').to_output()}
+{ToolUse("save", ["hello.py"], 'print("Hello world")', {"filepath": "hello.py", "content": 'print("Hello world")'}).to_output(tool_type)}
```
ErikBjare: Will this later be written like this?

```suggestion
{ToolUse("save", {"path": "hello.py", "content": 'print("Hello world")'}).to_output(tool_type)}
```

Or how can we eliminate the duplicated args?

pyproject.toml
```diff
@@ -83,6 +83,7 @@ pillow = {version = "*", optional=true}
 # server
 flask = {version = "^3.0", optional=true}
 flask-cors = {version = "^4.0", optional=true}
+json-repair = "^0.30.2"
```
ErikBjare: Place this line in an appropriate section (core)

ErikBjare (commented):
Very nice! This looking really good.

Just a few comments.

gptme/tools/base.py
jrmi: Fixed.

gptme/tools/save.py
jrmi: Fixed.

gptme/tools/shell.py
jrmi: Fixed.

gptme/tools/base.py
jrmi: Fixed

gptme/tools/base.py
jrmi: Fixed

gptme/tools/save.py
jrmi: Fixed.

pyproject.toml
jrmi: Fixed.

jrmi:
> Very nice! This looking really good.
> 
> Just a few comments.

Thanks for your second review. I think I fixed all your feedbacks and I have adapted all the tools. 

I faced a new issue: the python tool description is too long (>1024 char limit of tool description) because we include the docstring of the python function. I imagined two solutions:
- we could make the description of the `chats` tool shorter as it's the longest one but we'll encounter this issue sooner or later down the road.
- We move the functions description in each tool instead.

For now I have implemented the second version but I could try the other if you prefer.

Ready for another review.

gptme/tools/base.py
```diff
@@ -173,6 +288,12 @@ def iter_from_content(cls, content: str) -> Generator["ToolUse", None, None]:
         for tool_use in tool_uses:
             yield tool_use
 
+        # check if its a toolcall
+        if match := toolcall_re.search(content):
+            tool_name = match.group(1)
+            kwargs = cast(dict[str, str], json_repair.loads(match.group(2)))
```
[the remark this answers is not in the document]
jrmi: `json_repair` documentation says we [shouldn't try to load the json](https://github.com/mangiucugna/json_repair/?tab=readme-ov-file#avoid-this-antipattern) ourselves and we should use it as a drop in replacement of `json.loads`.

ErikBjare (approved):
Looks great!

I only have nitpick comments that I think will be easier to fix for me after merge (incl. organizing `openai` and `anthropic` imports so that they are lazy-loaded, to improve startup times).

I have yet to try running it, does it feel solid to you? Do the tests pass locally?

If the markdown/default format still works reliably, happy to merge! I can resolve the merge conflict via rebase (caused by merging https://github.com/ErikBjare/gptme/pull/293).

jrmi:
> I have yet to try running it, does it feel solid to you? Do the tests pass locally?

The tests pass locally and I have added more tests. The tool format probably need a bit of tweaking but it's an opt-in option and AFAIK the default markdown version doesn't seem to be affected by the changes. so I would say it's safe.

> If the markdown/default format still works reliably, happy to merge! I can resolve the merge conflict via rebase (caused by merging https://github.com/ErikBjare/gptme/pull/293).

The only doubt is that I can't test the Anthropic version because I've no API key. May be worth to try at least one prompt?

ErikBjare:
@jrmi I just sent you an Anthropic API key in DM on Discord.

Will try to get this merged sometime today.

ErikBjare:
Merged in #300, thanks a lot for contributing!

Tracking remaining issues in https://github.com/ErikBjare/gptme/issues/302

Made a PR with some fixes: https://github.com/ErikBjare/gptme/pull/303

PR #293, opened 2024-11-29
feat: output costs for request and session

ErikBjare:
> [!IMPORTANT]
> Adds functionality to calculate and log token usage and costs for requests and sessions in `LogManager`, with updates to `logmanager.py`, `message.py`, and `util/__init__.py`.
> 
>   - **Behavior**:
>     - `LogManager.append()` in `logmanager.py` now logs token and cost information using `costs()` when a message from the assistant is appended.
>     - New function `costs()` in `logmanager.py` calculates and logs token usage and cost per request and session.
>   - **Functions**:
>     - Adds `_tokens_inout()` and `_cost()` in `logmanager.py` to compute token input/output and cost for messages.
>     - Adds `cost()` method to `Message` class in `message.py` to calculate cost based on token count and model pricing.
>   - **Utilities**:
>     - Updates `get_tokenizer()` in `util/__init__.py` to log a warning if a tokenizer for a model is not found, using a default tokenizer instead.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 9741158df045840cc7dcbdae78559809f67c66a0. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -17,7 +17,6 @@ from .llm.models import get_model
 from .logmanager import Log, LogManager, prepare_messages
 from .message import Message
 from .prompts import get_workspace_prompt
-from .util.readline import add_history
 from .tools import ToolUse, execute_msg, has_tool
 from .tools.base import ConfirmFunc
 from .tools.browser import read_url
@@ -28,6 +27,8 @@ from .util import (
     print_bell,
     rich_to_str,
 )
+from .util.cost import log_costs
+from .util.readline import add_history
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


ErikBjare:
Been running this for a while now, this morning I kept to a single session and compared it to Anthropic spend at the end:

Anthropic console: $6.39
gptme cost estimate: $5.335

Difference could be due to tokenization differences (we assume cl100k), but close enough.

ErikBjare:
Good enough, merging this.

PR #301, opened 2024-12-03
fix: prevent command execution from triggering unnecessary response generation

ErikBjare:
Accidentally pushed this to master which broke `--non-interactive` in the subagent test.

Reverted and fixing it here.


ErikBjare:
I've still seen this happen every now and then, but I don't think this PR is salvageable.

PR #306, opened 2024-12-03
feat: improve tool use response format

jrmi:
This is an attempt to match the expected output for each API regarding tool use response.

It modify the message dicts to match the expected output.

Also fix tool call not working with stream == False

### DIFF
--- a/gptme/cli.py
+++ b/gptme/cli.py
@@ -117,7 +117,7 @@ The interface provides user commands that can be used to interact with the syste
 @click.option(
     "--tool-format",
     "tool_format",
-    default="markdown",
+    default=None,
     help="Tool parsing method. Can be 'markdown', 'xml', 'tool'. (experimental)",
 )
 @click.option(
@@ -149,7 +149,7 @@ def main(
     name: str,
     model: str | None,
     tool_allowlist: list[str] | None,
-    tool_format: ToolFormat,
+    tool_format: ToolFormat | None,
     stream: bool,
     verbose: bool,
     no_confirm: bool,
@@ -187,8 +187,10 @@ def main(
 
     config = get_config()
 
-    tool_format = tool_format or config.get_env("TOOL_FORMAT") or "markdown"
-    set_tool_format(tool_format)
+    selected_tool_format: ToolFormat = (
+        tool_format or config.get_env("TOOL_FORMAT") or "markdown"  # type: ignore
+    )
+    set_tool_format(selected_tool_format)
 
     # early init tools to generate system prompt
     init_tools(frozenset(tool_allowlist) if tool_allowlist else None)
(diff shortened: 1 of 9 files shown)


gptme/cli.py
```diff
@@ -187,8 +187,10 @@ def main(
 
     config = get_config()
 
-    tool_format = tool_format or config.get_env("TOOL_FORMAT") or "markdown"
-    set_tool_format(tool_format)
+    selected_tool_format: ToolFormat = (
+        tool_format or config.get_env("TOOL_FORMAT") or "markdown"  # type: ignore
```
jrmi, line 191: Unrelated but the env variable was ignored because of the default value.

gptme/llm/llm_anthropic.py
```diff
@@ -66,9 +67,17 @@ def chat(messages: list[Message], model: str, tools: list[ToolSpec] | None) -> s
     )
     content = response.content
     logger.debug(response.usage)
-    assert content
-    assert len(content) == 1
-    return content[0].text  # type: ignore
+
+    parsed_block = []
```
jrmi, line 69: Tool call was broken with non stream API.

gptme/llm/llm_anthropic.py
```diff
@@ -99,7 +108,7 @@ def stream(
                     block = chunk.content_block
                     if isinstance(block, anthropic.types.ToolUseBlock):
                         tool_use = block
-                        yield f"\n@{tool_use.name}: "
+                        yield f"\n@{tool_use.name}({tool_use.id}): "
```
jrmi, line 108: We add the tool uid here to save it in the message.

gptme/llm/llm_anthropic.py
```diff
@@ -224,7 +303,7 @@ def _transform_system_messages(
 
     # for any subsequent system messages, transform them into a <system> message
     for i, message in enumerate(messages):
-        if message.role == "system":
+        if message.role == "system" and message.call_id is None:
```
jrmi, line 305: We don't want to change messages with call_id. They will be handled by the `_handle_tool` function.

jrmi:
Hey @ErikBjare this MR is ready to review:
- it fixes the tool response to match the expected format by adding the `call_id` to the message
- it fixes the tool call with non-stream API
- it improves openai msg2dict process by adding the `_prepare_messages_for_api`
- it adds tests for message preparation process
- it fixes small bugs

Let me know what you think.

FYI I tested it with openai and anthropic ;-)

gptme/llm/llm_anthropic.py
ErikBjare: Add this as a code comment, as it's good to know when reading

jrmi: Done.

ErikBjare (approved):
Super impressive, works amazingly! :muscle: 

Thanks again :pray: 

:1st_place_medal: :crown:

PR #317, opened 2024-12-10
feat(anthropic): improve prompt caching and type safety

ErikBjare:
Costs were looking a bit expensive, realized that the prompt caching implementation is pretty sub-par and doesn't cache longer contexts when they appear in the conversation (only system messages).

- Add better type hints for Anthropic API types
- Optimize prompt caching with ephemeral cache control
- Extract message preparation into dedicated function
- Improve documentation

### DIFF
--- a/gptme/llm/llm_anthropic.py
+++ b/gptme/llm/llm_anthropic.py
@@ -10,17 +10,18 @@ from typing import (
     cast,
 )
 
-from typing_extensions import Required
-
 from ..constants import TEMPERATURE, TOP_P
-from ..message import Message, len_tokens, msgs2dicts
+from ..message import Message, msgs2dicts
 from ..tools.base import Parameter, ToolSpec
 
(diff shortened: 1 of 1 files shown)


gptme/llm/llm_anthropic.py
```diff
@@ -286,3 +282,81 @@ def _spec2tool(
...
+        ] = []
+        raw_content = (
+            msg["content"]
+            if isinstance(msg["content"], list)
+            else [{"type": "text", "text": msg["content"]}]
+        )
+
+        for part in raw_content:
+            if isinstance(part, dict):
+                if part.get("type") == "text" and i == len(messages_dicts) - 1:
```
ErikBjare, line 344: Might want to review this later to cache at the optimal locations, especially considering RAG and "live" messages. Should probably also add a cache step at the system prompt.

PR #320, opened 2024-12-10
fix: remove system messages from tool examples in system prompt

ErikBjare:


ErikBjare:
@TimeToBuildBob What do you think about this? Worth reviving or close?

TimeToBuildBob:
Worth reviving IMO.

I checked: `System:` outputs are still present in 30+ places across tool examples (`shell.py`, `tmux.py`, `computer.py`, `form.py`, `gh.py`). PR #306 fixed the stream=False tool call format issue, but didn't address the example hallucination risk.

The logic here is sound — having `System:` lines in tool examples teaches the model it can produce them, which is exactly the wrong thing to teach. The `strip_system=True` approach is clean and surgical.

On the TODOs:
- Tests pass ✅ (Codecov shows 100% coverage on changed lines)
- Non-streaming evals would be ideal but the existing CI coverage seems sufficient to merge — this is a prompt engineering fix, not a logic change

My vote: merge it. If evals show a regression, we can revert.

ErikBjare:
@greptileai review

TimeToBuildBob:
Greptile's review flagged a real issue: in the `clean_example` function, codeblock state (`in_codeblock`) updates even inside System messages that get filtered out, which can desynchronize the state tracker. The fix is straightforward — move the codeblock tracking to after the `skip_next` check (only track state for lines we keep).

Still worth reviving with that fix applied. The concept is solid.

TimeToBuildBob:
Created #1490 as a clean reimplementation on current master, fixing the Greptile-identified codeblock state tracking bug. Ready to close this one if #1490 looks good.

PR #335, opened 2024-12-15
fix: support heredoc/EOF syntax in shell tool

### LINKED ISSUE #108: shell tool doesn't support heredoc syntax, stalls
For example:

```
cat << EOF | sudo tee /etc/pacman.d/hooks/90-update-efi.hook
[Trigger]
Type = Package
Operation = Install
Operation = Upgrade
Target = linux*
Target = linux-lts

[Action]
Description = Copying kernel to EFI directory...
When = PostTransaction
Exec = /usr/bin/sh -c 'cp /boot/vmlinuz-linux-lts /efi/EFI/arch/ && cp /boot/initramfs-linux-lts.img /efi/EFI/arch/'
EOF
```

### LINKED ISSUE #331: Sonnet keeps using EOF format
Sonnet recently tries to write to files using some weird format like `cat << " code" >> EOF` I don't exactly remember the syntax that always makes the command fail(will edit with the correct format when it does it again). An update to the tools prompt could easily fix this.

bjsi:
Fixes #108 
Fixes #331

My gptme agent (Jake :D) attempted to fix this autonomously, wasn't able to quite get there, but it did 70% of the work, I just cleaned things up. Learnt a lot doing this and I have a branch with a lot of hacks and experiments. I think there should be a mechanism for the agent to monitor commands. Eg. in this case, the command was hanging and I had to manually interrupt the process. It would be cool to have a heartbeat mechanism where the agent gets updates on long-running commands with the option to interrupt it.

----

> [!IMPORTANT]
> Fixes heredoc syntax handling in `ShellSession._run()` and adds tests for various heredoc cases in `test_shell.py`.
> 
>   - **Behavior**:
>     - Fixes heredoc syntax handling in `ShellSession._run()` in `shell.py` by adjusting command execution logic.
>     - Removes unnecessary `read_delimiter` flag and simplifies return code handling.
>   - **Tests**:
>     - Adds tests for heredoc syntax, including basic, stripped, nested, and variable substitution cases in `test_shell.py`.
>     - Updates `test_echo_multiline()` and adds `test_heredoc_complex()` to cover new cases.
>   - **Imports**:
>     - Changes relative imports to absolute imports in `shell.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 5e0d368d43a5355be98ecb21bb1b977822bc048f. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -410,6 +410,7 @@ def _shorten_stdout(
 
     return "\n".join(lines)
 
+
 def split_commands(script: str) -> list[str]:
     # TODO: write proper tests
     parts = bashlex.parse(script)
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -13,11 +13,10 @@ from collections.abc import Generator
 from pathlib import Path
 
 import bashlex
-
-from ..message import Message
-from ..util import get_installed_programs, get_tokenizer
-from ..util.ask_execute import execute_with_confirmation
-from .base import (
+from gptme.message import Message
+from gptme.util import get_installed_programs, get_tokenizer
+from gptme.util.ask_execute import execute_with_confirmation
+from gptme.tools.base import (
     ConfirmFunc,
     Parameter,
     ToolSpec,
@@ -162,7 +161,8 @@ class ShellSession:
         assert self.process.stdin
 
         # run the command
-        full_command = f"{command}; echo ReturnCode:$? {self.delimiter}\n"
+        full_command = f"{command}\n"
+        full_command += f"echo ReturnCode:$? {self.delimiter}\n"
         try:
             self.process.stdin.write(full_command)
         except BrokenPipeError:
@@ -180,23 +180,30 @@ class ShellSession:
         stdout = []
         stderr = []
         return_code = None
-        read_delimiter = False
 
         while True:
             rlist, _, _ = select.select([self.stdout_fd, self.stderr_fd], [], [])
+            if not rlist: continue
             for fd in rlist:
                 assert fd in [self.stdout_fd, self.stderr_fd]
                 # We use a higher value, because there is a bug which leads to spaces at the boundary
                 # 2**12 = 4096
                 # 2**16 = 65536
                 data = os.read(fd, 2**16).decode("utf-8")
-                re_returncode = re.compile(r"ReturnCode:(\d+)")
-                for line in re.split(r"(\n)", data):
-                    if match := re_returncode.match(line):
-                        return_code = int(match.group(1))
-                    if self.delimiter in line:
-                        read_delimiter = True
-                        continue
+                lines = data.splitlines(True)
+                for line in lines:
+                    if f"ReturnCode:" in line and self.delimiter in line:
+                        return_code = int(line.split("ReturnCode:")[1].split()[0])
+                        # if command is cd and successful, we need to change the directory
+                        if command.startswith("cd ") and return_code == 0:
+                            ex, pwd, _ = self._run("pwd", output=False)
+                            assert ex == 0
+                            os.chdir(pwd.strip())
+                        return (
+                            return_code,
+                            "".join(stdout).strip(),
+                            "".join(stderr).strip(),
+                        )
                     if fd == self.stdout_fd:
                         stdout.append(line)
                         if output:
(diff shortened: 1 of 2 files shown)


gptme/tools/shell.py
```diff
@@ -180,23 +180,30 @@ def _run(self, command: str, output=True, tries=0) -> tuple[int | None, str, str
         stdout = []
         stderr = []
         return_code = None
-        read_delimiter = False
 
         while True:
             rlist, _, _ = select.select([self.stdout_fd, self.stderr_fd], [], [])
+            if not rlist: continue
             for fd in rlist:
```
ErikBjare: Isn't this `if not rlist: continue` check redundant?

gptme/tools/shell.py
```diff
@@ -13,11 +13,10 @@
...
 import bashlex
-
-from ..message import Message
-from ..util import get_installed_programs, get_tokenizer
-from ..util.ask_execute import execute_with_confirmation
-from .base import (
+from gptme.message import Message
+from gptme.util import get_installed_programs, get_tokenizer
+from gptme.util.ask_execute import execute_with_confirmation
+from gptme.tools.base import (
```
ErikBjare: Plz retain the relative imports

gptme/tools/shell.py
```diff
@@ -180,23 +180,30 @@ def _run(self, command: str, output=True, tries=0) -> tuple[int | None, str, str
...
         while True:
             rlist, _, _ = select.select([self.stdout_fd, self.stderr_fd], [], [])
+            if not rlist: continue
             for fd in rlist:
                 assert fd in [self.stdout_fd, self.stderr_fd]
                 # We use a higher value, because there is a bug which leads to spaces at the boundary
                 # 2**12 = 4096
                 # 2**16 = 65536
                 data = os.read(fd, 2**16).decode("utf-8")
-                re_returncode = re.compile(r"ReturnCode:(\d+)")
```
ErikBjare, line 193: Why no longer use this regex? I recently added it in https://github.com/ErikBjare/gptme/commit/df5c9d6eca2d3d084a1060f97290ad3f629930c0 and I think it fixed a bug. Not sure if it currently supports negative returncodes though.

gptme/tools/shell.py
```diff
@@ -180,23 +180,30 @@ def _run(self, command: str, output=True, tries=0) -> tuple[int | None, str, str
...
-                for line in re.split(r"(\n)", data):
-                    if match := re_returncode.match(line):
-                        return_code = int(match.group(1))
-                    if self.delimiter in line:
-                        read_delimiter = True
-                        continue
+                lines = data.splitlines(True)
```
ErikBjare:
```suggestion
                lines = data.splitlines(keepends=True)
```

gptme/tools/shell.py
```diff
@@ -180,23 +180,30 @@ def _run(self, command: str, output=True, tries=0) -> tuple[int | None, str, str
...
+                for line in lines:
+                    if f"ReturnCode:" in line and self.delimiter in line:
+                        return_code = int(line.split("ReturnCode:")[1].split()[0])
+                        # if command is cd and successful, we need to change the directory
+                        if command.startswith("cd ") and return_code == 0:
+                            ex, pwd, _ = self._run("pwd", output=False)
+                            assert ex == 0
+                            os.chdir(pwd.strip())
```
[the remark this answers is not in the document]
ErikBjare: I think this is fine and intended.

ErikBjare (commented):
Awesome work! Been dragging my feet on this one for quite a while 😅 

Just a few minor comments

gptme/tools/shell.py
bjsi: ah yes! removed.

gptme/tools/shell.py
bjsi: gotcha, reverted

gptme/tools/shell.py
bjsi: ahh I see, best to just leave it as is then. reverted

ErikBjare:
Thank you so much for this @bjsi! Much appreciated 🥇

Removed the not-always-followed prompting to avoid heredoc in a follow-up commit: https://github.com/ErikBjare/gptme/commit/3950a6e7f11409b5ea8e1e946bfc51c51893aea8 

So nice nice to finally have this fixed! 🤗

PR #344, opened 2024-12-17
fix: support piping stdin into gptme -r (resume)

bjsi:
Almost got it to do this autonomously, but I still had to do some cleanup :P

I added the `msg.role == "user"` checks because those functions contain asserts that the role should be user.

Fixes https://github.com/ErikBjare/gptme/issues/202

----

> [!IMPORTANT]
> Add support for piping stdin into `gptme -r` (resume) by appending piped input to prompt messages in `cli.py`.
> 
>   - **Behavior**:
>     - Support piping stdin into `gptme -r` (resume) in `cli.py`.
>     - Append piped input to `prompt_msgs` if resuming a conversation in `main()`.
>   - **Functions**:
>     - Modify `main()` in `cli.py` to handle piped input and append to `initial_msgs` for new conversations.
>     - Update `confirm_func()` and `execute_cmd()` logic in `chat.py` to handle user role checks.
>   - **Misc**:
>     - Refactor stdin reading logic in `cli.py` to use `_read_stdin()` function.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for a69c42e69413ca689adaeb339c8999b30816cbc6. It will automatically update as commits are pushed.</sup>

PR #354, opened 2024-12-18
fix: properly handle piped input in prompts

ErikBjare:
## Issue
Previously, piped input was being treated as a separate message instead of being included in the user prompt. This could lead to unexpected behavior when using gptme with piped input.

## Changes
- Added inject_stdin function to handle piped input consistently
- Ensures piped input is appended to first prompt or creates new message if none exists
- Removes redundant system message creation

## Testing
Tested with various piped input scenarios:
- Piped input with no prompt
- Piped input with existing prompt
- No piped input (unchanged behavior)

----

> [!IMPORTANT]
> Introduces `inject_stdin` function in `gptme/cli.py` to handle piped input by appending it to the first prompt or creating a new message, removing redundant system message creation.
> 
>   - **Behavior**:
>     - Introduces `inject_stdin` function in `gptme/cli.py` to handle piped input.
>     - Appends piped input to the first prompt or creates a new message if no prompt exists.
>     - Removes redundant system message creation for piped input.
>   - **Testing**:
>     - Tested with scenarios: piped input with no prompt, piped input with existing prompt, and no piped input.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 57cdc9f005dfcbed92db77215ea5128ab7a86237. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/cli.py
+++ b/gptme/cli.py
@@ -238,10 +238,22 @@ def main(
     # TODO: referenced file paths in multiprompts should be read when run, not when parsed
     prompt_msgs = [Message("user", p) for p in prompts]
 
+    def inject_stdin(prompt_msgs, piped_input: str | None) -> list[Message]:
+        # if piped input, append it to first prompt, or create a new prompt if none exists
+        if not piped_input:
+            return prompt_msgs
+        stdin_msg = Message("user", f"```stdin\n{piped_input}\n```")
+        if not prompt_msgs:
+            prompt_msgs.append(stdin_msg)
+        else:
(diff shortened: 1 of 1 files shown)


ErikBjare:
This PR improves upon #344 which added initial support for piped stdin. Instead of treating piped input as a separate message, it now properly integrates it into the first prompt message (or creates a new one if none exists), making the behavior more consistent and intuitive.

PR #371, opened 2024-12-30
fix: tool api call broken when user answer no to when asked for confirmation

jrmi:
This MR fixes a few bug regarding tool call in tool format:
- When the user choose answer on when asked for confirmation
- When user edit the file before saving
- When the user uses `Ctrl+c` to interrupt gptme

The underlying problems where that sometimes the `call_id` wasn't set in certain situations or multiple messages where returned by the `tooluse.execute` call but all were the response from the tool so we needed to merge them.

----

> [!IMPORTANT]
> Fixes tool API call issues by ensuring `call_id` is set and merging tool responses, with updates to message handling and tests.
> 
>   - **Behavior**:
>     - Fixes tool API call issues when users answer "no" to confirmation, edit files before saving, or interrupt with `Ctrl+c`.
>     - Ensures `call_id` is set correctly in `execute_msg()` in `tools/__init__.py` and `ToolUse.execute()` in `tools/base.py`.
>     - Merges multiple tool responses with the same `call_id` in `_merge_tool_results_with_same_call_id()` in `llm_openai.py` and `llm_anthropic.py`.
>   - **Functions**:
>     - Updates `_handle_tools()` in `llm_anthropic.py` and `llm_openai.py` to handle user role messages with `call_id`.
>     - Modifies `_transform_system_messages()` in `llm_anthropic.py` to merge consecutive user messages with the same `call_id`.
>   - **Tests**:
>     - Adds test cases in `test_llm_anthropic.py` and `test_llm_openai.py` to verify message conversion and tool handling with `call_id`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 2d25236284b90f88290175329bd99032f7603c43. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 from .commands import action_descriptions, execute_cmd
 from .config import get_config
-from .constants import PROMPT_USER
+from .constants import INTERRUPT_CONTENT, PROMPT_USER
 from .init import init
 from .llm import reply
 from .llm.models import get_model
@@ -148,11 +148,7 @@ def chat(
                         )
                     except KeyboardInterrupt:
                         console.log("Interrupted. Stopping current execution.")
-                        manager.append(
-                            Message(
-                                "system", "User hit Ctrl-c to interrupt the process"
-                            )
-                        )
+                        manager.append(Message("system", INTERRUPT_CONTENT))
                         break
                     finally:
                         clear_interruptible()
@@ -225,7 +221,7 @@ def step(
     if (
         not last_msg
         or (last_msg.role in ["assistant"])
-        or last_msg.content == "Interrupted"
+        or last_msg.content == INTERRUPT_CONTENT
         or last_msg.pinned
         or not any(role == "user" for role in [m.role for m in log])
     ):  # pragma: no cover
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -255,9 +255,6 @@ def step(
         if msg_response:
             yield msg_response.replace(quiet=True)
             yield from execute_msg(msg_response, confirm)
-    except KeyboardInterrupt:
-        clear_interruptible()
-        yield Message("system", "User hit Ctrl-c to interrupt the process")
     finally:
         clear_interruptible()
 
--- a/gptme/llm/llm_anthropic.py
+++ b/gptme/llm/llm_anthropic.py
@@ -200,9 +200,8 @@ def stream(
 def _handle_tools(message_dicts: Iterable[dict]) -> Generator[dict, None, None]:
     for message in message_dicts:
         # Format tool result as expected by the model
-        if message["role"] == "system" and "call_id" in message:
+        if message["role"] == "user" and "call_id" in message:
             modified_message = dict(message)
-            modified_message["role"] = "user"
             modified_message["content"] = [
                 {
                     "type": "tool_result",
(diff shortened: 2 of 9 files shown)


gptme/chat.py
```diff
@@ -235,9 +235,6 @@ def step(
         if msg_response:
             yield msg_response.replace(quiet=True)
             yield from execute_msg(msg_response, confirm)
-    except KeyboardInterrupt:
```
jrmi, line 258: This is now handled in the `execute_msg` as we need the `call_id` in the message response.

gptme/tools/base.py
```diff
@@ -290,10 +290,9 @@ def execute(self, confirm: ConfirmFunc) -> Generator[Message, None, None]:
                     confirm,
                 )
                 if isinstance(ex, Generator):
-                    for msg in ex:
-                        yield msg.replace(call_id=self.call_id)
+                    yield from ex
```
jrmi, line 290: `call_id` attribution moved to `execute_msg`.

gptme/chat.py
```diff
@@ -235,9 +235,6 @@ def step(
...
-        clear_interruptible()
-        yield Message("system", "Interrupted")
     finally:
         clear_interruptible()
```
ErikBjare, line 255: I'm not sure if this is correct, I think the `clear_interruptible` is no longer called?

I think the `finally` clause might not run for the generator until the `next()` call.

ErikBjare: Nevermind, it's there in the except clause for that very reason, but no longer needed.

Not quite clear to me if the new behavior is correct though.

jrmi: Hmm no it's probably not the same. Let me find something.

jrmi: I've updated the MR. To stay on the safe side, I've called `clear_interruptible` in the `execute_msg` function. Notice the `break` in the except clause. Before it was still calling subsequent tools if any after a user interruption :facepalm:

jrmi: We could also call `clear_interruptible` inside the `handle_keyboard_interrupt` itself just before raising the `KeyboardInterrupt` exception, what do you think?

gptme/llm/llm_anthropic.py
```diff
@@ -358,22 +357,35 @@ def _transform_system_messages(
...
     # find consecutive user role messages and merge them together
     messages_new: list[Message] = []
     while messages:
         message = messages.pop(0)
-        if messages_new and messages_new[-1].role == "user" and message.role == "user":
+        if (
+            messages_new
+            and messages_new[-1].role == "user"
+            and message.role == "user"
+            and message.call_id == messages_new[-1].call_id
```
ErikBjare, line 382: Why is this condition needed?

gptme/tools/__init__.py
```diff
@@ -120,7 +122,17 @@ def execute_msg(msg: Message, confirm: ConfirmFunc) -> Generator[Message, None,
...
         if tooluse.is_runnable:
-            yield from tooluse.execute(confirm)
+            try:
+                for tool_response in tooluse.execute(confirm):
+                    yield tool_response.replace(call_id=tooluse.call_id)
+            except KeyboardInterrupt:
+                clear_interruptible()
+                yield Message(
+                    "system",
+                    "User hit Ctrl-c to interrupt the process",
```
ErikBjare: I kinda want to make these shorter

```suggestion
                    "Interrupted by user",
```

ErikBjare: Sort of what I did in #384

gptme/llm/llm_anthropic.py
jrmi: When using the function call API, multiple calls may occur within a single answer (parallel calls). To handle this properly, ensure there is one tool response per tool call. By verifying that the call_id is distinct, we can ensure messages are not merged unless they originate from the same function call.

gptme/tools/__init__.py
jrmi: I changed the message (and used a constant to have the same value everywhere it's needed.)

gptme/chat.py
ErikBjare: I'm really not sure, I find this pretty difficult to think about and would probably have to work with it to figure out the best way. If this works somewhat in your testing then I'm happy to fix any remaining issues after it's merged.

gptme/tools/__init__.py
```diff
@@ -120,7 +124,17 @@ def execute_msg(msg: Message, confirm: ConfirmFunc) -> Generator[Message, None,
 
     for tooluse in ToolUse.iter_from_content(msg.content):
         if tooluse.is_runnable:
-            yield from tooluse.execute(confirm)
+            try:
+                for tool_response in tooluse.execute(confirm):
+                    yield tool_response.replace(call_id=tooluse.call_id)
+            except KeyboardInterrupt:
+                clear_interruptible()
```
ErikBjare, line 124: I'm not sure this belongs so deep into the tools module, should probably be kept in `chat.py` if possible

ErikBjare: Ah, now I see why.

ErikBjare:
Thanks again! 🥇

PR #374, opened 2024-12-31
fix: improve smaller model response on user refusals

jrmi:
I noticed that when you refuse the operation when asked for confirmation, small models (at least gpt-o4-mini) are a bit lost and try again the same command without any modification multiple times which make the confirmation useless. 

To test it, with a `gpt-o4-mini` just use a prompt like "Create the file foo.txt with random content.` and say `No` when the tool ask for confirmation to save and the mode will call the tool again with the same parameters.

In this MR I tweaked a bit the initial prompt and the response message in these cases to prevent that behavior.

----

> [!IMPORTANT]
> Improves response behavior for smaller models by updating prompts and messages to prevent redundant command execution after user refusals.
> 
>   - **Behavior**:
>     - Updates confirmation messages in `chat.py`, `commands.py`, `tools/patch.py`, `tools/python.py`, `tools/save.py`, `tools/tmux.py`, and `util/ask_execute.py` to specify user-initiated interruptions and refusals.
>     - Modifies `prompt_gptme()` in `prompts.py` to instruct the model not to retry operations after user refusal and to ask for clarification instead.
>   - **Prompts**:
>     - Adds a line in `prompt_gptme()` to handle user-aborted operations by asking for clarification rather than retrying.
>   - **Messages**:
>     - Changes various system messages to clarify when a user has aborted an operation, e.g., "User hit Ctrl-c to interrupt the process" and "Aborted, user chose not to run this code."
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 3f54a027bd316b3e2b87cd562c49ee3cd2ace9d7. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -136,7 +136,11 @@ def chat(
                         )
                     except KeyboardInterrupt:
                         console.log("Interrupted. Stopping current execution.")
-                        manager.append(Message("system", "Interrupted"))
+                        manager.append(
+                            Message(
+                                "system", "User hit Ctrl-c to interrupt the process"
+                            )
+                        )
                         break
                     finally:
                         clear_interruptible()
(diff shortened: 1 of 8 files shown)


ErikBjare:
Nice, been meaning to do this for a while. Even confuses Sonnet sometimes :)

PR #379, opened 2025-01-04
fix: include contents of file on failed patches

ErikBjare:


ErikBjare:
This has been added since then via `GPTME_PATCH_RECOVERY = "true"`, should probably make it default.

PR #384, opened 2025-01-05
fix: fixed behavior for keyboard interrupts and user commands

ErikBjare:
Has some similarities to https://github.com/ErikBjare/gptme/pull/371


ErikBjare:
I need to break this apart to figure out why it isn't working

PR #399, opened 2025-01-12
feat: allow to override config from project

jrmi:
This MR allows to override some of the the global config keys with the project file. In particular we can override the environment variables. The project config and the main config are now merged in one object.

It also refactor the way in which the workspace is detected/constructed so that there is a central point where the logic is located.


jrmi:
Closing for now as I don't have time to work on it.

PR #416, opened 2025-01-22
feat: added think tool to let agent query a reasoner

ErikBjare:


gptme/tools/think.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+tool = ToolSpec(
+    "think",
+    desc="Tool to enable deeper thinking on challenging or complex problems.",
+    block_types=["think"],
+    execute=execute,
+    instructions="""
+Use this tool when you need help with complex reasoning or problem-solving.
+Wrap your thinking request in a code block with the language tag: `think`
```
ErikBjare, line 125: Shouldn't use markdown specific format instructions.

gptme/tools/think.py
```diff
@@ -0,0 +1,162 @@
...
+    thinking_prompt = f"""Please analyze this problem carefully and thoroughly:
+
+{full_content}
+
+Think step by step and consider:
+1. Key aspects and components
+2. Potential challenges and edge cases
+3. Alternative approaches
+4. Best practices and principles
+5. Implementation details"""
```
ErikBjare, line 100: Too specific to coding?

Issue #304, opened 2024-12-03
Add success/fail emojis to status messages for easier glancing of output


ErikBjare:
Might be nice to add some color to the output depending on if things go well or not

ErikBjare:
Done by @0xbrayo in #429  🥳

PR #429, opened 2025-01-30
feat: add status emojis to system messages

0xbrayo:
> [!IMPORTANT]
> Add status emojis to system messages by introducing a `status` attribute in the `Message` class and updating related functions to handle this attribute.
> 
>   - **Behavior**:
>     - Add `status` attribute to `Message` class in `message.py` to represent message status as "success", "error", "info", or `None`.
>     - Update `format_msgs()` in `message.py` to append status emojis to system messages based on the `status` attribute.
>   - **Functions**:
>     - Modify `execute()` in `base.py`, `execute_patch_impl()` in `patch.py`, `execute_python()` in `python.py`, `execute_save_impl()` and `execute_append_impl()` in `save.py`, `execute_shell_impl()` in `shell.py`, and `execute_tmux()` in `tmux.py` to set the `status` attribute in `Message` instances.
>   - **Misc**:
>     - Update `to_dict()` and `from_toml()` in `message.py` to handle the `status` attribute.
>     - Add status handling in `execute_with_confirmation()` in `ask_execute.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 05f822051b26eeb03eefd7dcc85140125b34198f. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -211,6 +211,15 @@ def format_msgs(
         if highlight:
             color = ROLE_COLOR[msg.role]
             userprefix = f"[bold {color}]{userprefix}[/bold {color}]"
+        if msg.role == "system":
+            first_line = msg.content.split("\n", 1)[0].lower()
+            isSuccess = first_line.startswith(("saved", "appended", "patch successfully", "ran"))
+            isError = first_line.startswith(("error", "failed"))
+            if isSuccess:
+                userprefix = f"{userprefix} ✅"
+            elif isError:
+                userprefix = f"{userprefix} ❌"
+                
         # get terminal width
         max_len = shutil.get_terminal_size().columns - len(userprefix)
         output = ""
--- a/gptme/tools/patch.py
+++ b/gptme/tools/patch.py
@@ -257,7 +257,7 @@ def execute_patch(
         code = kwargs.get("patch", code)
 
     if not code:
-        yield Message("system", "No patch provided by the assistant")
+        yield Message("system", "Error: No patch provided by the assistant")
         return
 
     yield from execute_with_confirmation(
--- a/gptme/tools/tmux.py
+++ b/gptme/tools/tmux.py
@@ -223,7 +223,7 @@ def execute_tmux(
             continue
 
         if len(parts) < 2:
-            yield Message("system", f"Missing arguments for command: {command}")
+            yield Message("system", f"Error: Missing arguments for command: {command}")
             continue
 
         _args = parts[1]
@@ -237,7 +237,7 @@ def execute_tmux(
         elif command == "kill_session":
             yield kill_session(_args)
         else:
-            yield Message("system", f"Unknown command: {command}")
+            yield Message("system", f"Error: Unknown command: {command}")
 
 
 instructions = """


0xbrayo:
#304

gptme/message.py
```diff
@@ -211,6 +215,17 @@ def format_msgs(
...
+            
+        # Add status emoji for system messages
+        if msg.role == "system" and msg.status:
+            status_emoji = {
+                "success": "✅",
+                "error": "❌",
+                "info": "ℹ️",
+            }
+            if emoji := status_emoji.get(msg.status):
+                userprefix = f"{userprefix} {emoji}"
```
ErikBjare: In gptme-webui I simply check the first line with:

```py
first_line = msg.content.split("\n", 1)[0].lower()
isSuccess = first_line.startswith(["saved", "success", "patch successfully"])
isError = first_line.startswith(["error", "failed"])
```

Since it's just a design thing, I think such a lightweight approach makes more sense.
The "info" emoji also doesn't make much sense to me (how is it different from no status?)

ErikBjare (commented):
I think we can do this without adding `msg.status`

gptme/message.py
0xbrayo: > The "info" emoji also doesn't make much sense to me (how is it different from no status?)

The info emoji was suggested by gptme yeah, I should just scrap it.

gptme/message.py
```diff
@@ -211,6 +211,15 @@ def format_msgs(
         if highlight:
             color = ROLE_COLOR[msg.role]
             userprefix = f"[bold {color}]{userprefix}[/bold {color}]"
+        if msg.role == "system":
+            first_line = msg.content.split("\n", 1)[0].lower()
+            isSuccess = first_line.startswith(("saved", "executed", "ran", "appended", "patch successfully"))
```
ErikBjare: > "executed", "ran"

I think these do not actually signal success explicitly? (could have non-zero exit codes)

gptme/message.py
```diff
@@ -211,6 +211,15 @@ def format_msgs(
...
+            isError = first_line.startswith(("error", "failed","missing arguments", "unknown command", "no patch"))
```
ErikBjare: I see tmux gives "missing arguments", the message should probably be changed to "Error: Missing arguments" so the detection can be simplified.
Same for "unknown command" and "no patch".

0xbrayo: Done

gptme/message.py
```diff
@@ -213,8 +213,8 @@ def format_msgs(
             userprefix = f"[bold {color}]{userprefix}[/bold {color}]"
         if msg.role == "system":
             first_line = msg.content.split("\n", 1)[0].lower()
-            isSuccess = first_line.startswith(("saved", "executed", "ran", "appended", "patch successfully"))
-            isError = first_line.startswith(("error", "failed","missing arguments", "unknown command", "no patch"))
+            isSuccess = first_line.startswith(("saved", "appended", "patch successfully"))
+            isError = first_line.startswith(("error", "failed")) or "Return code" in msg.content[-16:]
```
0xbrayo: When a shell command exits with a non-zero code the code is part of the message. Only captures unsuccessful commands. Could change the code to append the return code for each command and strip the zero return codes before displaying the message.

gptme/tools/shell.py
```diff
@@ -289,13 +289,15 @@ def execute_shell_impl(
...
     stdout = _shorten_stdout(stdout.strip(), pre_tokens=2000, post_tokens=8000)
     stderr = _shorten_stdout(stderr.strip(), pre_tokens=2000, post_tokens=2000)
-
     msg = (
-        _format_block_smart(
-            f"Ran {'allowlisted ' if allowlisted else ''}command", cmd, lang="bash"
-        )
-        + "\n\n"
+    _format_block_smart(
+        f"{'Failed to run' if returncode else 'Ran'} "
```
ErikBjare: I think "failed to run" might be bad prompting. We should be careful to not infer too much from a non-zero exit code I think. I'd just keep the "Ran" and not try to infer a success/error status for shell commands, for now.

0xbrayo: Removed the shell.py changes.

ErikBjare:
Nice, lets go!

PR #435, opened 2025-02-04
feat: much improved RAG, added LLM post-processing of results

Miyou:
Does the following:

- Only gets RAG context for last user message, not all user messages. This prevents rag search calls spiraling out of control since currently the number of calls grows by 1 with each new message from the user.
- Adds typing to RAG config.
- Adds the option to post-process the extracted context with an LLM. While this adds more latency, from my testing this improves output by only including actually relevant context chunks and greatly reduces the main LLM to be confused when the unrelated chunks are returned from rag search.

> [!IMPORTANT]
> Introduces `RagConfig` for RAG settings, adds post-processing prompt, and updates RAG tool for LLM post-processing support.
> 
>   - **Configuration**:
>     - Introduces `RagConfig` class in `gptme/config.py` for RAG settings.
>     - Adds `default_post_process_prompt` for context post-processing.
>     - Updates `ProjectConfig` to use `RagConfig`.
>   - **RAG Tool**:
>     - Updates `rag_enhance_messages` in `rag.py` to support post-processing with LLM using `post_process_model` and `post_process_prompt`.
>     - Modifies `rag_search` to use `--format full --print-relevance`.
>   - **Misc**:
>     - Adds logging for post-processing duration in `rag_enhance_messages`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 318099c3cfffd21dfd219730cd5ffda2c7c65106. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/config.py
+++ b/gptme/config.py
@@ -37,11 +37,44 @@ class Config:
         }
 
 
+default_post_process_prompt = """
+You are an intelligent knowledge retrieval assistant designed to analyze context chunks and extract relevant information based on user queries. Your primary goal is to provide accurate and helpful information while adhering to specific guidelines.
+
+You will be provided with a user query inside <user_query> tags and a list of potentially relevantcontext chunks inside <chunks> tags.
+
+When a user submits a query, follow these steps:
+
+1. Analyze the user's query carefully to identify key concepts and requirements.
+
+2. Search through the provided context chunks for relevant information.
+
+3. If you find relevant information:
+   a. Extract the most pertinent parts.
+   b. Summarize the relevant context inside <context_summary> tags.
+   c. Output the exact relevant context chunks, including the complete <chunks path="...">...</chunks> tags.
+
+4. If you cannot find any relevant information, respond with exactly: "No relevant context found".
+
+Important guidelines:
+- Do not make assumptions beyond the available data.
+- Maintain objectivity in source selection.
+- When returning context chunks, include the entire content of the <chunks> tag. Do not modify or truncate it in any way.
+- Ensure that you're providing complete information from the chunks, not partial or summarized versions within the tags.
+- When no relevant context is found, do not return anything other than exactly "No relevant context found".
+- Do not output anything else than the <context_summary> and <chunks> tags.
+
+Please provide your response, starting with the summary and followed by the relevant chunks (if any).
+"""
+
+
 @dataclass
 class RagConfig:
     enabled: bool = False
     max_tokens: int | None = None
     min_relevance: float | None = None
+    post_process: bool = True
+    post_process_model: str | None = None
+    post_process_prompt: str = default_post_process_prompt
 
 
 @dataclass
--- a/gptme/config.py
+++ b/gptme/config.py
@@ -75,6 +75,8 @@ class RagConfig:
     post_process: bool = True
     post_process_model: str | None = None
     post_process_prompt: str = default_post_process_prompt
+    workspace_only: bool = True
+    paths: list[str] = field(default_factory=list)
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


gptme/config.py
```diff
@@ -37,14 +37,61 @@ def dict(self) -> dict:
...
     base_prompt: str | None = None
     prompt: str | None = None
     files: list[str] = field(default_factory=list)
-    rag: dict = field(default_factory=dict)
+    rag: RagConfig = field(default_factory=RagConfig)
+
+    def __post_init__(self):
+        # Convert the rag dict/Table to RagConfig if needed
+        if isinstance(self.rag, dict | Container):
+            self.rag = RagConfig(**self.rag)  # type: ignore
```
ErikBjare: Is this needed?
Conversion should happen at the constructor call, as this isn't quite type-correct.

gptme/tools/rag.py
```diff
@@ -15,6 +15,11 @@
 
     [rag]
     enabled = true
+    post_process = false # Whether to post-process the context with an LLM to extract the most relevant information
+    post_process_model = "openai/gpt-4-turbo" # Which model to use for post-processing
```
ErikBjare: Better default plz (obviously AI generated lol)
`gpt-4o-mini` should do
```suggestion
    post_process_model = "openai/gpt-4o-mini" # Which model to use for post-processing
```

gptme/tools/rag.py
```diff
@@ -140,41 +146,77 @@ def init() -> ToolSpec:
...
+                    model=rag_config.post_process_model,  # type: ignore
+                    tools=[],
                 )
-            except Exception as e:
-                logger.warning(f"Error getting context: {e}")
-
-        enhanced_messages.append(msg)
-
-    return enhanced_messages
+                logger.info(f"Ran RAG post-process in {time.monotonic() - start:.2f}s")
```
ErikBjare: Refactor this into a function, so that we get a smaller try/except block.

gptme/config.py
Miyou: Cursor wrote this part. Not sure what the correct way to do it is, but without this it didn't work and with it did so I figured it was ok.

gptme/tools/rag.py
Miyou: Actually I put that in there as I assumed gpt-4-turbo would be faster, but tbh I'm not seeing much of a difference in speed. Can change it. Also this is just the doc comment and not the actual default. I didn't set a default in the actual config, but can if you want.

gptme/config.py
ErikBjare: It's really ugly/bad practice, you should almost never use the `__post_init__` for dataclasses ime.

Should find the init call that didn't work and do the `RagConfig(**source)` there (with a `# type: ignore` if needed)

Miyou: fixed

gptme/tools/rag.py
Miyou: fixed

gptme/tools/rag.py
```diff
@@ -106,7 +112,7 @@ def rag_search(query: str, return_full: bool = False) -> str:
     cmd = ["gptme-rag", "search", query]
     if return_full:
         # shows full context of the search results
-        cmd.append("--show-context")
+        cmd.extend(["--format", "full", "--print-relevance"])
```
ErikBjare:
```suggestion
        cmd.extend(["--format", "full", "--score"])
```

gptme/tools/rag.py
```diff
@@ -140,41 +146,89 @@ def init() -> ToolSpec:
...
+        "gptme-rag",
+        "search",
+        query,
+    ]
+    if workspace and rag_config.workspace_only:
+        cmd.append(workspace.as_posix())
+    elif rag_config.paths:
+        cmd.extend(rag_config.paths)
+    if not should_post_process:
+        cmd.append("--print-relevance")
```
ErikBjare:
```suggestion
        cmd.append("--score")
```

PR #447, opened 2025-02-19
refactor: attempt to refactor & simplify chat.py

ErikBjare:
Mostly vibecoding. Some things are broken but some good ideas here too.

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -1,15 +1,12 @@
-import errno
 import logging
 import os
-import re
 import sys
 import termios
-import urllib.parse
 from collections.abc import Generator
 from pathlib import Path
 from typing import cast
 
-from .commands import action_descriptions, execute_cmd
+from .commands import execute_cmd
 from .config import get_config
 from .constants import INTERRUPT_CONTENT, PROMPT_USER
 from .init import init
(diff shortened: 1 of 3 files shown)


ErikBjare:
Looks like I created a bunch of conflicts in https://github.com/ErikBjare/gptme/pull/454

gptme/chat.py
```diff
@@ -81,146 +184,147 @@ def chat(
...
+    elif msg.role == "user":
+        return _include_paths(msg, workspace)
+    return msg
+
+
+def _is_executable_command(
+    msg: Message, manager: LogManager, confirm_func: ConfirmFunc
+) -> bool:
+    """Check if message is an executable command and execute it if so."""
+    return msg.role == "user" and execute_cmd(msg, manager, confirm_func)
```
ErikBjare, line 279: This is poorly named. Should just check, not run. (better behavior in #453)

libertas1982 (approved):

ErikBjare:
This handles GitHub URLs specially which is something we want, but this has diverged significantly from master and would be difficult to resolve the conflicts.

Ellipsis has also highlighted some changes we made which might be good ideas. Splitting up the chat function which is one of the most complex in the codebase and has some quirks would be nice.

PR #466, opened 2025-03-10
Add workspace on new conversation

RickardCarlsson4:
> [!IMPORTANT]
> Integrate workspace paths into tool initialization, particularly for shell commands, across API endpoints and tool functions.
> 
>   - **Behavior**:
>     - `init_tools` in `tools/__init__.py` now accepts a `workspace` parameter to initialize tools with a specific directory.
>     - API endpoints in `api.py` (`api_conversation`, `api_conversation_put`, `api_conversation_post`, `api_conversation_generate`) now initialize tools with the workspace path from the conversation log.
>     - `ShellSession` in `shell.py` changes directory to the workspace if provided during initialization.
>   - **Functions**:
>     - `init_shell` in `shell.py` initializes the shell tool with a workspace.
>     - `get_shell` in `shell.py` initializes `ShellSession` with a workspace.
>   - **Misc**:
>     - Debug logging added to track workspace initialization in `api.py` and `tools/__init__.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ErikBjare%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 233cdec312a4f157669d1832136abf36bd2d4108. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/server/api.py
+++ b/gptme/server/api.py
@@ -319,11 +319,14 @@ def api_conversation_generate(logfile: str):
                 manager.append(msg)
                 yield f"data: {flask.json.dumps({'role': 'assistant', 'content': output, 'stored': True})}\n\n"
 
-                # Recursively handle any new tool uses
-                if any(
-                    tooluse.is_runnable for tooluse in ToolUse.iter_from_content(output)
-                ):
-                    yield from generate()
+                # Execute any remaining tools and stream their output
+                tool_replies = list(execute_msg(msg, confirm_func))
+                for reply_msg in tool_replies:
+                    logger.debug(
+                        f"Tool output: {reply_msg.role} - {reply_msg.content[:100]}..."
+                    )
+                    manager.append(reply_msg)
+                    yield f"data: {flask.json.dumps({'role': reply_msg.role, 'content': reply_msg.content, 'stored': True})}\n\n"
 
         except GeneratorExit:
             logger.info("Client disconnected during generation, interrupting")
--- a/gptme/server/api.py
+++ b/gptme/server/api.py
@@ -26,7 +26,7 @@ from ..llm import _stream
 from ..llm.models import get_default_model
 from ..logmanager import LogManager, get_user_conversations, prepare_messages
 from ..message import Message
-from ..tools import ToolUse, execute_msg, init_tools
+from ..tools import ToolUse, execute_msg, init_tools, get_tools
 
 logger = logging.getLogger(__name__)
 
@@ -197,6 +197,10 @@ def api_conversation_generate(logfile: str):
     init_tools(None, workspace=manager.workspace)  # Initialize tools with workspace path
     print("[DEBUG] api_conversation_generate: Tools initialized")
 
+    # Get available tools
+    tools = [t for t in get_tools() if t.is_runnable()]
+    print(f"[DEBUG] api_conversation_generate: Available tools: {[t.name for t in tools]}")
+
     # performs reduction/context trimming, if necessary
     msgs = prepare_messages(manager.log.messages)
 
@@ -208,7 +212,7 @@ def api_conversation_generate(logfile: str):
         # Non-streaming response
         try:
             # Get complete response
-            output = "".join(_stream(msgs, model, tools=None))
+            output = "".join(_stream(msgs, model, tools=tools))
 
             # Store the message
             msg = Message("assistant", output)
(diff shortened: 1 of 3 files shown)


gptme/server/api.py
```diff
@@ -170,7 +212,7 @@ def api_conversation_generate(logfile: str):
         # Non-streaming response
         try:
             # Get complete response
-            output = "".join(_stream(msgs, model, tools=None))
+            output = "".join(_stream(msgs, model, tools=tools))
```
ErikBjare, line 215: I think this is tricky, the `tools` parameter as provided to `_stream` is only to be provided if we are using the `tools` API format from Anthropic/OpenAI, and not the `markdown` or `xml` format. There is probably code handling this in `chat.py` too.

gptme/server/api.py
```diff
@@ -97,15 +101,37 @@ def api_conversation_put(logfile: str):
...
     logdir.mkdir(parents=True)
+
+    # Create workspace symlink
+    workspace_symlink = logdir / "workspace"
+    workspace_path = Path(workspace_path).resolve()
+    if not workspace_symlink.exists():
+        workspace_symlink.symlink_to(workspace_path, target_is_directory=True)
+
+    # Initialize tools with workspace path
+    init_tools(None, workspace=workspace_path)
```
ErikBjare, line 132: Just use `os.chdir(workspace_path)`?

gptme/server/api.py
```diff
@@ -48,8 +48,12 @@ def api_conversations():
 @api.route("/api/conversations/<string:logfile>")
 def api_conversation(logfile: str):
```
ErikBjare, line 49: We may want to run this in a new process, see other comment.

(not in this PR)

gptme/tools/__init__.py
```diff
@@ -80,12 +81,24 @@ def _discover_tools(module_names: frozenset[str]) -> list[ToolSpec]:
...
     allowlist: frozenset[str] | None = None,
+    workspace: Path | None = None,
 ) -> None:
-    """Runs initialization logic for tools."""
-
+    """Runs initialization logic for tools.
+    
+    Args:
+        allowlist: Set of tool names to allow
+        workspace: Workspace directory path for tools that need it (e.g. shell)
```
ErikBjare, line 92: I think this is problematic because the shell tool working directory is actually tied to the process working directory, which is also true for the save/patch tools (to keep all paths consistent throughout the system).

So, when running conversations in parallel (as the server does), this leads to a kind of "non-thread safe" behavior: changes to their working directories would apply to other concurrent conversations, which could yield very confusing and potentially harmful results.

It would be process-safe however, since each conversation wouldn't affect the others cwd.

But even that is not a perfect solution, because if the working directory changes, it would not reflect upon subsequent calls to `/generate` (same is true if you resume a gptme session after changing the dir). The core issue is that the working directory is ephemerally stored in the cwd.

A solution to that is to extend the working directory change detections we do in the shell tool to propagate it up into the message log, so that we can know which cwd each message ran in, or was in after.

I also don't think we want to pass down tool arguments like this.

(not in this PR)

gptme/tools/__init__.py
```diff
@@ -95,24 +108,34 @@ def init_tools(
...
 
         _loaded_tools.append(tool)
+        print(f"[DEBUG] Tool '{tool.name}' loaded")
 
     for tool_name in allowlist or []:
         if not has_tool(tool_name):
             raise ValueError(f"Tool '{tool_name}' not found")
 
+# Add cache decorator after function definition
+init_tools = lru_cache(init_tools)
```
ErikBjare, line 137: Keep it as an actual decorator so that it's visible.

gptme/tools/shell.py
```diff
@@ -455,5 +470,6 @@ def split_commands(script: str) -> list[str]:
             required=True,
         ),
     ],
+    init=init_shell,
```
ErikBjare, line 473: I don't think this is needed

gptme/tools/shell.py
```diff
@@ -141,6 +141,10 @@ def _init(self):
         self.run("export PAGER=")
         self.run("export GH_PAGER=")
         self.run("export GIT_PAGER=cat")
+        
+        # Change to workspace directory if provided
+        if workspace and workspace.exists():
+            self.run(f"cd {workspace}")
```
ErikBjare, line 147: The process is already started in `cwd` and used throughout gptme as a session global (unfortunately, see other comments), so just use that.

RickardCarlsson4:
Most of these issues will be solved when with the redesign of the server backend. 

https://github.com/gptme/gptme/pull/469

PR #482, opened 2025-03-24
feat(user-confirm): capture user feedback when command execution is declined

netbuddy:
# Improved User Feedback When Declining Command Execution

This PR enhances the user experience when a user declines to execute a command by:

- Prompting the user to explain why they declined execution
- Asking for suggestions on what they'd like to do instead
- Sending this feedback as a user message rather than a system message

This change helps the LLM understand the user's concerns and expectations better, enabling it to provide more relevant alternatives or explanations in its next response. It transforms a dead-end rejection into a conversational opportunity, maintaining dialogue flow and improving the collaborative experience.

### DIFF
--- a/gptme/util/ask_execute.py
+++ b/gptme/util/ask_execute.py
@@ -249,9 +249,17 @@ def execute_with_confirmation(
         try:
             # Get confirmation
             if not confirm_fn(confirm_msg or f"Execute on {path}?"):
-                yield Message(
-                    "system", "Operation aborted: user chose not to run the operation."
+                # Obtaining reasons and recommendations for users' refusal to implement
+                session = get_prompt_session()
+                feedback = session.prompt(
+                    [
+                        (
+                            "bold fg:ansiyellow",
+                            "Why did you decline? What would you like me to do instead? ",
+                        )
+                    ]
                 )
+                yield Message("user", f"I declined to execute because: {feedback}")
                 return
 
             # Get potentially edited content


ErikBjare:
I think this is good but ideally it should just simply return the user to the normal prompt when no is answered instead of the assistant immediately responding to the no without any information of why. Answering no in the confirm prompt should effectively work the same as the current workflow I use (skip answering `n` at all, just Ctrl+C Ctrl+C which is a UX/keycombo anti-pattern). `n` + Enter simply returning user to prompt would be better.

@TimeToBuildBob Open a new PR that addresses this. It should work similarly to if the user interrupts repeatedly at the confirm prompt to force them back to the main prompt.

TimeToBuildBob:
## ✅ New PR Created as Requested

I've created a new PR that addresses your feedback: #1127

**Approach**: When user answers 'n' at the confirmation prompt, the chat now returns directly to the user prompt without generating an assistant response - similar to how Ctrl+C behaves.

**Changes**:
- Added `DECLINED_CONTENT` constant (parallel to `INTERRUPT_CONTENT`)
- Modified `execute_with_confirmation` to yield this message when user declines
- Modified `_process_message_conversation` to detect declined content and break the loop

This supersedes PR #482's approach (which captured feedback and had assistant respond to it).

-- Bob 🔧

loftybuilder:
Hey @netbuddy, just a heads up - this PR has been superseded by #1127 which was merged yesterday. 

Erik requested a different approach: instead of capturing feedback and having the assistant respond, the new implementation simply returns the user to the prompt (similar to Ctrl+C behavior).

Thanks for the original idea though - it sparked the discussion that led to the improvement! You might want to close this PR. 🙏

ErikBjare:
Yes indeed, thanks @netbuddy, sorry I forgot to co-author the commits given the source of inspiration for them (you deserved it!).

It's a nice UX improvement!

PR #485, opened 2025-03-26
feat: add token limit warnings with improved timing

ErikBjare:
Adds token limit warnings that show:

- Token count info every 10k tokens
- Warning when reaching 50% of context limit (once per model)
- Warnings appear before user input for better UX

This helps users better manage their conversation sizes and know when to start a new conversation.


ErikBjare:
We have token_awareness for agents now, more relevant that some UX enhancement for users.

PR #487, opened 2025-03-29
docs: gptme-suggested README improvements

ErikBjare:
I just asked gptme to improve it.

I think it is good to including some of these key aspects in the README which we should probably mention directly without referring to the docs.

Some hallucinations in the config section which should be easy to fix


TimeToBuildBob:
Rebased on master and fixed the hallucinated config section:

**Removed:**
- Fake top-level TOML keys (`default_model`, `confirm_shell`, `color`, `system_prompt`, `logs_dir`)
- Non-existent env vars (`GPTME_DEFAULT_MODEL`, `GPTME_CONFIRM_SHELL`, `GPTME_COLOR`)
- Fake model names (`openrouter/gpt-4-grok`, `llama.cpp/llama-3-70b-q4_0`)

**Replaced with:**
- Correct nested TOML structure matching `gptme/config.py` (`[user]`, `[prompt]`, `[env]` sections)
- Real env vars only (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `OPENROUTER_API_KEY`)
- Fixed tool name: `python` → `ipython` (matches `ToolSpec`)

The rest of the PR (tools table, troubleshooting, examples) looks good and is accurate.

PR #488, opened 2025-03-29
fix: better workspace handling

ErikBjare:
We shouldn't assert-crash when we resume from another directory than the workspace directory.

We shold probably `cd` to the workspace directory if we open in another dir.

In the future we should probably even track the workspace changes as part of messages, resuming from the last directory.


ErikBjare:
This has been fixed since then.

PR #493, opened 2025-04-04
feat: add support for quad-ticked markdown in codeblocks and tooluse

ErikBjare:
- [ ] Needs tests?
   - Tests will prob fail?

### DIFF
--- a/gptme/codeblock.py
+++ b/gptme/codeblock.py
@@ -2,6 +2,8 @@ from collections.abc import Generator
 from dataclasses import dataclass, field
 from xml.etree import ElementTree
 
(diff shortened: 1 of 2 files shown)


ErikBjare:
During testing I found that while Claude often followed the format, it didn't always use triple backticks. In my tests it for example failed to save a file containing triple-backticked codeblocks because it failed to use the quadruple tooluse format.

TimeToBuildBob:
This was superseded by #1009 (commit fe24f45cb) which implemented quad+ backtick support with proper variable fence length matching. The feature is already on master — closing.

PR #500, opened 2025-04-14
Refactor config, support overriding values

Miyou:
Support overriding env and mcp in priority order (highest to lowest):
1. env vars
2. project config
3. user config

Adds ChatConfig (todo support env and mcp overriding as above)

----

> [!IMPORTANT]
> Refactor configuration system to support priority-based overriding and add `ChatConfig` for chat-specific settings, with updates to CLI and tests.
> 
>   - **Configuration Refactor**:
>     - Introduces priority-based config overriding: env vars > project config > user config in `config.py`.
>     - Adds `ChatConfig` class to manage chat-specific configurations with methods `load_or_create()` and `from_logdir()`.
>     - Refactors `Config` class to include `user` and `project` configurations.
>   - **CLI Changes**:
>     - Updates `cli.py` and `server/cli.py` to use `set_config()` for initializing configurations.
>     - Modifies `main()` in `cli.py` to load `ChatConfig` with CLI overrides.
>   - **Testing**:
>     - Adds `test_chat_config.py` to test `ChatConfig` loading and CLI overrides.
>     - Updates `test_config.py` to test environment and MCP config loading priorities.
>     - Adds tests in `test_mcp.py` for MCP server connections and operations.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 776ef3ec967bd4c06ef755ef805ab9f34be03898. It will automatically update as commits are pushed.</sup>

PR #503, opened 2025-04-16
refactor: improve config and tools architecture

ErikBjare:
- Move chat settings into [chat] namespace in TOML config
- Make tools parameter required for prompt generation
- Change tool allowlist from frozenset to list
- Remove problematic caching of init_tools
- Better separate tool discovery and initialization
- Improve config serialization and overrides


ErikBjare:
Follow-up to #500

PR #509, opened 2025-04-19
fix: improved interrupt handling with tests for shell tool

ErikBjare:


TimeToBuildBob:
Closing — draft from 10 months ago, 1269 commits behind master, single commit never iterated on. If shell interrupt handling needs work, should start fresh against current codebase.

PR #521, opened 2025-05-06
fix: move workspace prompt to get_prompt

Miyou:


gptme/prompts.py
```diff
@@ -323,12 +333,15 @@ def get_tree_output(workspace: Path) -> str | None:
...
-def get_workspace_prompt(workspace: Path) -> str:
-    # NOTE: needs to run after the workspace is initialized (i.e. initial prompt is constructed)
+def prompt_workspace(workspace: Path | None = None) -> Generator[Message, None, None]:
     # TODO: update this prompt if the files change
     # TODO: include `git status -vv`, and keep it up-to-date
     sections = []
 
+    if workspace is None:
+        yield Message("system", "")
+        return
```
ErikBjare: just return without yielding anything

gptme/prompts.py
```diff
@@ -363,9 +376,9 @@ def get_workspace_prompt(workspace: Path) -> str:
         sections.append(f"## Project Structure\n\n```\n{tree_output}\n```\n\n")
 
     if sections:
-        return "# Workspace Context\n\n" + "\n\n".join(sections)
+        yield Message("system", "# Workspace Context\n\n" + "\n\n".join(sections))
     else:
-        return ""
+        yield Message("system", "")
```
ErikBjare: you can just return here (or let it return implicitly)

gptme/cli.py
```diff
@@ -247,6 +247,8 @@ def inject_stdin(prompt_msgs, piped_input: str | None) -> list[Message]:
     else:
         workspace_path = Path(workspace) if workspace else Path.cwd()
 
+    init_workspace(workspace_path, logdir)
```
ErikBjare: I realized that I added saving/init of workspace on `ChatConfig.save`, maybe `init_workspace` is no longer needed as it does the same thing?

https://github.com/gptme/gptme/commit/984d6412ef3de78cb7de30166bbc8bf9ad68412d#diff-fc7178b0841ac611cfd4fd3c4f07a2217bd628c99ac1070c6aa6d2aed5f2b23aR270-R330

Miyou: hmm yeah then maybe only needed in subagent and eval agent

ErikBjare: They should probably use ChatConfig the same way?

gptme/server/api_v2.py
```diff
@@ -541,6 +540,9 @@ def api_conversation_put(conversation_id: str):
     # Save the chat config
     chat_config.save()
 
+    # Initialize tools in this thread
+    init_tools(chat_config.tools)
```
ErikBjare: This shouldn't be done in PUT (we're not even in a specific thread here), only on step.

PR #571, opened 2025-06-26
fix: improve support for nested markdown codeblocks in markdown tool format

ErikBjare:
Attempt at fixing early closes of codeblocks that can happen when codeblock contents have nested codeblocks by using tricks like needing codeblock starts to have leading line (or start of string) and their ends to have leading line (or EOF).

This should help situations where we have nested codeblocks where their start tags dont have a langtag set. gptme likes to output these sometimes, causing issues, especially in situations where it'd use a "text" langtag if asked to include a langtag (langtag otherwise often redundant for such plaintext codeblocks).

I thought this approach would be an easy fix/improvement, but turns out it's a bit of a thorny issue (12 failing tests in CI). Maybe I'm missing something here about the limitations posed. It all works fairly well with the XML tool format.

----

> [!IMPORTANT]
> Improved handling of nested markdown codeblocks in `gptme/codeblock.py` to prevent early closure and added tests for better parsing accuracy.
> 
>   - **Behavior**:
>     - Improved `_extract_codeblocks()` in `gptme/codeblock.py` to handle nested markdown codeblocks without language tags.
>     - Codeblocks must start at the beginning of a line and end alone on a line, handling nested cases.
>     - Handles consecutive codeblocks correctly.
>   - **Tests**:
>     - Updated `test_extract_codeblocks_basic`, `test_extract_codeblocks_multiple`, `test_extract_codeblocks_nested`, and `test_extract_codeblocks_unfinished_nested` in `tests/test_codeblock.py`.
>     - Added `test_extract_codeblocks_markdown_with_nested_no_langtag` and `test_extract_codeblocks_consecutive` in `tests/test_codeblock.py`.
>     - Added `test_apply_with_nested_codeblock` in `tests/test_tools_patch.py` to test patch application with nested codeblocks.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 56a6740d4c163299648ef6345bc40c34df0bd90d. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/codeblock.py
+++ b/gptme/codeblock.py
@@ -80,6 +80,7 @@ def _extract_codeblocks(markdown: str) -> Generator[Codeblock, None, None]:
 
         # Look for code block start
         if line.startswith("```"):
+            start_line = i  # Track the starting line number
             lang = line[3:].strip()
             content_lines: list[str] = []
             i += 1
@@ -108,7 +109,7 @@ def _extract_codeblocks(markdown: str) -> Generator[Codeblock, None, None]:
                             nesting_depth -= 1
                             if nesting_depth == 0:
                                 # This closes our top-level block
-                                yield Codeblock(lang, "\n".join(content_lines))
+                                yield Codeblock(lang, "\n".join(content_lines), start=start_line)
                                 break
                             else:
                                 # This closes a nested block, add to content
--- a/gptme/codeblock.py
+++ b/gptme/codeblock.py
@@ -110,6 +110,7 @@ def _extract_codeblocks(markdown: str) -> Generator[Codeblock, None, None]:
                             if nesting_depth == 0:
                                 # This closes our top-level block
                                 yield Codeblock(lang, "\n".join(content_lines), start=start_line)
+                                i += 1  # Move past the closing ```
                                 break
                             else:
                                 # This closes a nested block, add to content
@@ -125,5 +126,5 @@ def _extract_codeblocks(markdown: str) -> Generator[Codeblock, None, None]:
 
             # If we reached the end without completing the block, don't yield it
             # (this handles the unfinished nested test case)
-
-        i += 1
+        else:
+            i += 1
--- a/gptme/codeblock.py
+++ b/gptme/codeblock.py
@@ -50,38 +50,79 @@ class Codeblock:
         return list(_extract_codeblocks(markdown))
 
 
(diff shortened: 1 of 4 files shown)


ErikBjare:
@OpenHands please fix the failing actions on PR #571

ErikBjare:
@OpenHands please fix the failing actions on PR #571

ErikBjare:
Hmm, OpenHands seems confused about the actual solution to the failing CI in [`56a6740` (#571)](https://github.com/gptme/gptme/pull/571/commits/56a6740d4c163299648ef6345bc40c34df0bd90d)

ErikBjare:
This is kind of a risky PR given the extra restrictions it adds to the markdown tool format (leading and trailing empty lines) as well as nested markdown content (needs to consistently have leading and trailing empty lines, or neither).

But I think it solves a really common current edge case where it mistakenly only writes partial files due to getting interrupted and executed when outputting a nested codeblock without langtags. This should probably let us remove any mentions of this from the prompt as well, which mentions it currently iirc (update: but which I cannot find!).

ErikBjare:
If it ends up being a bad change, we can always revert.

ErikBjare:
This now works when streaming is off, but doesn't work when streaming is on (presumably because it cannot tell the difference between EOF and in-progress streaming, leading to incorrectly thinking that the start of a wrapped codeblock is the end of the parent).

image

I guess an ugly way around this is to not count triple ticks as closing tag if they are preceeded by an empty line like in the above case (since they are then likely a start tag), but that could break some workflows where e.g. `append` intentionally adds whitespace.

ErikBjare:
I just saw this hallucination in an output, suggesting the codeblock wasn't detected due to the non-empty line following the ending tag.

````
```shell
cat gptme/tools/morph.py
```
"""
Gives the LLM agent the ability to edit files using Morph Fast Apply v2.

Morph is a specialized code-patching LLM that applies edits at 4000+ tokens per second.
It uses a different format than the patch tool: <code>original</code><update>changes</update>

Environment Variables:
    OPENROUTER_API_KEY: Required for accessing Morph via OpenRouter
"""
...
````

PR #583, opened 2025-07-01
 feat: add create agent endpoint in v2 api

Miyou:


gptme/logmanager.py
```diff
@@ -389,6 +389,7 @@ class ConversationMeta:
     messages: int
     branches: int
     workspace: str
+    agent_name: str | None = None
```
ErikBjare, line 392: I'm thinking we may want to change this somehow. We are likely to want to identify the agent more uniquely than simply by name, and there is likely to be more agent-data we want to communicate.

Right now agents are somewhat tied to the current workspace (the context about their own workspace gets nuked when navigating into another project iirc). So we want agents (a type of project) to be able to navigate different (non-agent) workspaces and get to keep their own context in addition to the project context, just like how a human using them with gptme would (human doesn't get lobotomized when it enters a project).

Some options:
 - We add a `agent_path: str | None` instead, which works similar to `workspace` in how it pulls in context, "layering" actual workspace context on top (if agent workspace != project workspace).
 - We just merge this for now, and I do the above in a follow-up PR at some point (likely)
 - Maybe all this is overkill/overengineering

gptme/server/api_v2.py
```diff
@@ -1100,6 +1109,156 @@ def api_conversation_config_patch(conversation_id: str):
...
+    # Pull in any git submodules
+    submodule_result = subprocess.run(
+        ["git", "submodule", "update", "--init", "--recursive"],
+        capture_output=True,
+        check=False,
+        cwd=temp_dir,
+    )
+    if submodule_result.returncode != 0:
+        # Delete the temp dir if the submodule update failed
+        # shutil.rmtree(temp_dir)
```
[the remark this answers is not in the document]
ErikBjare: Not so sure about that

gptme/config.py
```diff
@@ -106,10 +106,72 @@ class ProjectConfig:
     files: list[str] | None = None
     context_cmd: str | None = None
     rag: RagConfig = field(default_factory=RagConfig)
+    agent_name: str | None = None
```
ErikBjare: Should make a AgentConfig nested as a key under ProjectConfig

gptme/server/api_v2.py
```diff
@@ -1100,6 +1125,159 @@ def api_conversation_config_patch(conversation_id: str):
...
+
+            return flask.jsonify(
+                {"error": f"Failed to run post-fork command: {error_msg}"}
+            ), 500
+    except Exception as e:
+        # Delete the temp dir and workspace if the post-fork command failed
+        shutil.rmtree(temp_dir)
+        if path.exists():
+            shutil.rmtree(path)
+        return flask.jsonify({"error": f"Failed to run post-fork command: {e}"}), 500
```
ErikBjare, line 1225: See https://github.com/gptme/gptme/issues/584

gptme/server/api_v2.py
```diff
@@ -346,6 +358,19 @@ def step(
...
+    # addresses the issue where chats don't start in the directory they should.
+    # If we are attempting to make a step in a conversation without any user
+    # messages, make sure we first chdir to the workspace directory (so that
+    # the conversation starts in the right folder).
+    user_messages = [msg for msg in manager.log.messages if msg.role == "user"]
+    if not user_messages:
+        logger.debug(
+            f"No user messages found, changing directory to workspace: {workspace}"
+        )
+        os.chdir(workspace)
```
ErikBjare, line 372: Fixes a common and annoying bug, but doesn't fix the underlying/entire issue with multiple chats managing the same pwd state.

gptme/server/api_v2.py
ErikBjare: Misattributed comment, strange

gptme/prompts.py
```diff
@@ -141,8 +161,14 @@ def prompt_gptme(
...
 
+    if agent_name:
+        agent_blurb = f"{agent_name}, an agent running in gptme, letting you act as a general-purpose AI assistant powered by LLMs"
+    else:
+        agent_name = f"gptme v{__version__}"
+        agent_blurb = f"{agent_name}, a general-purpose AI assistant powered by LLMs"
+
     default_base_prompt = f"""
-You are gptme v{__version__}, a general-purpose AI assistant powered by LLMs. {('Currently using model: ' + model_meta.full) if model_meta else ''}
+You are {agent_blurb}. {('Currently using model: ' + model_meta.full) if model_meta else ''}
```
ErikBjare, line 171: Agents are now properly named in the system prompt.

gptme/server/api_v2.py
ErikBjare: Apparently this didn't fix it.

gptme/logmanager.py
```diff
@@ -416,6 +418,14 @@ def get_conversations() -> Generator[ConversationMeta, None, None]:
         chat_config = ChatConfig.from_logdir(conv_fn.parent)
         display_name = chat_config.name or conv_id
```
ErikBjare, line 420: Should be retrieved from the agent config: `get_project_config(chat_config.agent)`

gptme/server/api_v2.py
ErikBjare: Could be the same issue as https://github.com/gptme/gptme/issues/594, in that case https://github.com/gptme/gptme/pull/595 should have fixed it.

PR #586, opened 2025-07-08
feat: implement custom read tool with git info, line ranges, and directory support

ErikBjare:
Just an experiment/curiosity. I don't think this is really worth the extra complexity, at least not for me personally. But it has some nice attributes and improvements that might actually make it better than grep/sed (including metadata like last git commit by default, which is high signal/noise).

- Add git commit information display for files and directories
- Support line range syntax (file.py:10-20, file.py:10, etc.)
- Add glob pattern support (*.py, **/*.md)
- Implement character budget management with intelligent truncation
- Add directory listing with file sizes and recent commits
- Filter out common ignore patterns (__pycache__, .git, node_modules, etc.)
- Replace simple shell command wrapper with full-featured tool

### DIFF
--- a/gptme/tools/read.py
+++ b/gptme/tools/read.py
@@ -1,27 +1,618 @@
 """
-Read the contents of a file.
+Read files and display their contents with git information.
 """
 
-from . import ToolSpec, ToolUse
+import glob
+import re
+import subprocess
+from collections.abc import Generator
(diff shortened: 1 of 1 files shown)


ErikBjare:
We're not doing custom tools for this, better to let it use shell tbh.

PR #606, opened 2025-07-21
fix: respect project and chat config for MCP tools initialization

### LINKED ISSUE #605: `create_mcp_tools` do not respect project or chat config
It initializes too early, making it initialize MCP tools using only user config, making it impossible for it to load MCP from project (useful for agents) or chat config (on resume).

ErikBjare:
- Only initialize MCP tools after project config is loaded to ensure access to project and chat MCP configurations
- Avoid caching incomplete tool lists that exclude MCP tools
- Remove debug print statements that were highlighting this issue
- Prevent duplicate MCP server connections by not clearing cache unnecessarily

Fixes #605

Issues:
 - With these changes, running `gptme -t 'shell' '/tools'` in a project with MCP configured leads to all available tools incorrectly being returned (instead of only the shell tool). It works fine in a project without configured MCP servers.

----

> [!IMPORTANT]
> Ensure MCP tools are initialized only after project config is loaded, preventing caching of incomplete tool lists and duplicate server connections.
> 
>   - **Behavior**:
>     - Initialize MCP tools only after project config is loaded in `get_available_tools()`.
>     - Avoid caching incomplete tool lists by checking for complete config in `get_available_tools()`.
>     - Prevent duplicate MCP server connections by not clearing cache unnecessarily.
>   - **Misc**:
>     - Remove debug print statements in `get_toolchain()` and `get_available_tools()`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 450fd8ae0b7416c7382eeceb517e22c5e02161e5. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -152,8 +152,10 @@ def init_tools(
 def get_toolchain(allowlist: list[str] | None) -> list[ToolSpec]:
     # Validate allowlist if provided
     # TODO: maybe check in CLI init instead, as this might hard error in the server when loading conversations where tools are not available
+    available_tools = get_available_tools()
+
+    # Validate allowlist against available tools
     if allowlist is not None:
-        available_tools = get_available_tools()
         available_tool_names = [tool.name for tool in available_tools]
 
         for tool_name in allowlist:
@@ -170,7 +172,7 @@ def get_toolchain(allowlist: list[str] | None) -> list[ToolSpec]:
                 )
 
     tools = []
-    for tool in get_available_tools():
+    for tool in available_tools:
         if allowlist is not None and not tool.is_mcp and tool.name not in allowlist:
             continue
         if not tool.is_available:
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
## Alternative Approach in PR #731

Hi @ErikBjare - I've been investigating this issue and created an alternative approach in PR #731 that takes a simpler path.

### Comparison

**Your approach (#606)**: Changes initialization logic to delay MCP tools
- Pros: Addresses the root cause directly
- Cons: Has the tool filtering bug you documented (`-t 'shell'` returns all tools)
- Complexity: Modifies initialization logic, adds config completeness checks

**My approach (#731)**: Just clears cache at reload point
- Pros: Very simple, single line change + test, no known issues
- Cons: Relies on `reload_config()` being called (which it always is)
- Complexity: Minimal, just invalidates cache at the right time

### The Tool Filtering Bug

You noted in this PR:
> "With these changes, running `gptme -t 'shell' '/tools'` in a project with MCP configured leads to all available tools incorrectly being returned (instead of only the shell tool)."

This is concerning because it breaks tool filtering functionality. The cache-clearing approach in #731 doesn't have this issue because it doesn't change the initialization logic.

### Recommendation

I think the cache-clearing approach (#731) is simpler and avoids the tool filtering bug. However, I defer to your judgment as the maintainer - you may have reasons for preferring the initialization timing approach.

Would you be open to testing #731 to see if it solves the issue without the tool filtering bug?

See PR #731 for full analysis comparing all three approaches (#606, #677, #731).

ErikBjare:
Closing in favor of #731

PR #617, opened 2025-08-12
feat: add support for gptme.local.toml configuration layering

ErikBjare:
This PR adds support for layered configuration using gptme.local.toml files, allowing for local configuration overrides.


ErikBjare:
Hoping this just works now that https://github.com/gptme/gptme/pull/731 is merged.

PR #625, opened 2025-08-25
feat: add stream cancellation support for OpenRouter

ErikBjare:
Trying to fix the issue where e.g. `gptme -m openrouter/openai/gpt-oss-20b 'try ipython'` leads to a bunch of repeated messages on tool calls, almost like it didnt see the reply it just made and the following system message as if making the previous request twice. This loops unless interrupted with an actual user message.

Didn't work :(

Might be a plain case of one should not using the system message in the conversation?

- Add optional cancel_event parameter to stream() function
- Support graceful stream interruption using threading.Event
- Add proper cleanup to close stream resources on cancellation
- Include unit test for stream cancellation functionality

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import threading
 from collections.abc import Generator, Iterable
 from functools import lru_cache
 from typing import TYPE_CHECKING, Any, cast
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing — 1083 commits behind master, never iterated on after initial commit. Stream cancellation is a useful feature but should be reimplemented fresh if needed. This branch can serve as reference.

PR #628, opened 2025-08-27
fix: more detailed perplexity prompting guide

ErikBjare:


ErikBjare:
This prompt may be an improvement, but it is overly verbose and seems like a better fit fot a lesson or something unless it can be significantly reduced in size. Maybe a short version in the browser prompt and a longer lesson for detailed usage (basically "query as a LLM", not sure how detailed you have to be, you can ask it hard things to challenge it to do a good job)

TimeToBuildBob:
Closing — per Erik's own review, this prompt is overly verbose and the approach should use the lessons system instead of embedding it in code. 1078 commits behind master.

PR #747, opened 2025-10-19
feat(shell): add configurable shellcheck exclude codes

### LINKED ISSUE #744: Shellcheck codes to ignore
e.g. things like SC2164: 

````
```shell
cd /home/bob/gptme-issue743-retry-5xx
```
System:
Shellcheck found potential issues:
```
<command>:2:1: warning: Use 'cd ... || exit' or 'cd ... || return' in case cd fails. [SC2164]
```
````

### PR DESCRIPTION
**Pull request by TimeToBuildBob**
Implements #744 - configure shellcheck integration with exclude codes and error codes that block execution.

### Configurable Excludes
- Add default exclude list: **SC2164** (cd error handling - too noisy for interactive commands)
- Add `GPTME_SHELLCHECK_EXCLUDE` environment variable for custom excludes
- Combine default and custom excludes

### Error Codes (New)
- Add `GPTME_SHELLCHECK_ERROR_CODES` environment variable
- Default error codes: **SC2006** (backticks vs \$(...) notation)
- Error codes **block command execution** to prevent critical issues
- Other codes show warnings but allow execution

### Implementation
- Parse shellcheck output to extract error codes
- Check if any triggered codes are in error codes list
- Block execution when error codes found
- All shell tool tests pass (29 passed)

## Usage

**Default behavior** (excludes SC2164, blocks SC2006):
```shell
# This won't warn (excluded)
cd /some/path

# This will BLOCK execution (error code)
echo "# Header

- `file.txt`"
# Error: SC2006 - backticks cause issues
```

**Custom excludes**:
```bash
# Exclude additional codes
export GPTME_SHELLCHECK_EXCLUDE="SC2086,SC2046"
```

**Custom error codes**:
```bash
# Add more codes that should block execution
export GPTME_SHELLCHECK_ERROR_CODES="SC2006,SC2086"
```

## Why This Matters

Erik's example from issue comments:
```shell
echo "# Header

- `file.txt`"
```

Results in (backticks get executed and replaced with empty output):
```
# Header

-
```

This **ruins PRs and commit messages**. SC2006 now prevents this by blocking execution.

Fixes #744

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -730,11 +730,14 @@ def get_path_fn(*args, **kwargs) -> Path | None:
     return None
 
 
-def check_with_shellcheck(cmd: str) -> tuple[bool, str]:
+def check_with_shellcheck(cmd: str) -> tuple[bool, bool, str]:
     """
     Run shellcheck on command if available.
 
-    Returns: Tuple of (has_issues: bool, message: str)
+    Returns: Tuple of (has_issues: bool, should_block: bool, message: str)
+    - has_issues: True if any shellcheck issues found
+    - should_block: True if critical error codes found that should prevent execution
+    - message: Description of issues found
 
     Note:
         - Requires shellcheck (sudo apt install shellcheck)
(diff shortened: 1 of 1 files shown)


ErikBjare:
I think some codes should outright prevent the shell command from executing, like SC2006. Since they indicate likely misformatted output. We can keep other codes as warnings for now, while we investigate if any others should be considered errors.

ErikBjare (changes requested):
See comment

Issue #39, opened 2023-11-16
patch causes critical failure


nslieske:
running the main branch (love this software by the way, very useful!!)

```patch vkdl.py
@@ -156,5 +156,11 @@
 print("Moving downloaded videos to /media/gdrive/solar/cells/masterclips/")
-shutil.move(f'{output_dir}/*.mp4', '/media/gdrive/solar/cells/masterclips/')
+for file_name in os.listdir(output_dir):
+    if file_name.endswith('.mp4'):
+        shutil.move(os.path.join(output_dir, file_name), '/media/gdrive/solar/cells/masterclips/')
+
 print("Video downloading process completed.")
```                                                         
Applying the patch kills the app.  I'm able to recover by asking to build and save the file.  

Traceback (most recent call last):
  File "/home/noah/.local/bin/gptme", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/noah/.local/lib/python3.11/site-packages/gptme/cli.py", line 219, in main
    for msg in loop(log, no_confirm, model, stream=stream):  # pragma: no cover
  File "/home/noah/.local/lib/python3.11/site-packages/gptme/cli.py", line 271, in loop
    yield from execute_msg(msg_response, ask=not no_confirm)
  File "/home/noah/.local/lib/python3.11/site-packages/gptme/tools/__init__.py", line 30, in execute_msg
    yield from execute_codeblock(codeblock, ask)
  File "/home/noah/.local/lib/python3.11/site-packages/gptme/tools/__init__.py", line 46, in execute_codeblock
    yield from execute_patch(f"```{codeblock}```", fn, ask=ask)
[...]

PR #751, opened 2025-10-20
feat(subagent): add output schema support with Pydantic validation

TimeToBuildBob:
## Summary

Adds foundation for Manus-style planner pattern by implementing output schema support for the subagent tool.

## Changes

- Add optional `output_schema` parameter to `subagent()` function
- Validate subagent outputs against provided Pydantic schemas
- Modify return prompt to include schema JSON when provided
- Add comprehensive tests for schema validation logic

## Related

- Issue #39 (Manus agent review)
- Task: enhance-subagent-planner-pattern
- Part of implementing structured task delegation pattern

## Testing

- ✅ All existing subagent tests pass
- ✅ New schema validation tests added
- ✅ Tests verify function signature and validation logic
- ✅ All pre-commit hooks pass

## Next Steps

This is foundation work. Remaining for full planner pattern:
- Add planner mode for task delegation
- Implement constrained decoding
- Add context isolation options

### DIFF
--- a/gptme/tools/subagent.py
+++ b/gptme/tools/subagent.py
@@ -13,6 +13,8 @@ from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
+from pydantic import BaseModel, ValidationError
+
 from ..message import Message
 from . import get_tools
 from .base import ToolSpec, ToolUse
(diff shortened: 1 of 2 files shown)


ErikBjare:
I really don't like Pydantic, I don't think we need or should add strict schema support like this.

Neither should we plan to implement constrained decoding, as we are trying to be as provider-independent as possible (and that inherently ties us to a provider since there is no established common standard). If the model isn't good enough to follow the format, it's probably not good enough (or our prompt isn't).

PR #794, opened 2025-10-28
fix(autocompact): restore manager state after fork to prevent name mutation

### LINKED ISSUE #793: bug: unexpected manager.fork behavior, switches to fork in manager
In: https://github.com/gptme/gptme/blob/7c6761684941185b0c22cbe7cd2a9f40f96a4a9f/gptme/tools/autocompact.py#L371

```py
    manager.fork(fork_name)
```

Note that manager.fork actually forks the current managed conversation (mutable). Read LogManager to understand. This is what I meant with the conversation name being incorrectly set for the compacted conversation with a  "before-compact" name.

> Compacted conversations getting weird names like "conv-before-compact-test-before-compact-a7b9"

_Originally posted by @ErikBjare in https://github.com/gptme/gptme/pull/792#discussion_r2471060552_

TimeToBuildBob:
## Summary

Fixes #793 

Implements Erik's design for autocompact naming: use the original conversation as backup, and create a compacted fork with timestamp.

## Problem

The autocompact feature had two issues:
1. **Naming bug**: Compacted conversations got weird accumulated names
2. **Design**: Original approach saved/restored manager state, working around natural fork behavior

## Solution

Following Erik's design feedback:

1. **Original conversation stays untouched** (automatic backup)
2. **Fork becomes compacted version** with clean timestamped name: `conv-compacted-YYYYMMDDHHMM`
3. **Manager switches to fork naturally** (no state manipulation needed)
4. **Suffix stripping prevents accumulation** on repeated compactions

### Implementation Details

```python
def _get_compacted_name(conversation_name: str) -> str:
    # Strip any existing compacted suffixes to prevent accumulation
    base_name = conversation_name
    while True:
        new_name = re.sub(r"-compacted-\d{8}-\d{6}$", "", base_name)
        if new_name == base_name:
            break
        base_name = new_name
    
    timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
    return f"{base_name}-compacted-{timestamp}"
```

**Result**: Clean, unique names even with repeated compactions:
- `conv` → `conv-compacted-20251029-080800`
- `conv-compacted-20251029-073045` → `conv-compacted-20251029-083000` (not accumulation)

## Testing

- ✅ Pre-commit hooks pass (ruff, mypy, formatting)  
- ✅ Automated reviews positive (Ellipsis, Greptile)
- CI will run full test suite

Thanks to @ErikBjare for the design guidance!

### DIFF
--- a/gptme/tools/autocompact.py
+++ b/gptme/tools/autocompact.py
@@ -15,7 +15,7 @@ from ..message import Message, len_tokens
 from ..util.output_storage import create_tool_result_summary
 
 if TYPE_CHECKING:
-    from ..logmanager import Log, LogManager
+    from ..logmanager import LogManager
 
 logger = logging.getLogger(__name__)
 
@@ -341,6 +341,7 @@ def autocompact_hook(
     import time
 
     from ..llm.models import get_default_model
+    from ..logmanager import Log
     from ..message import len_tokens
 
     global _last_autocompact_time


ErikBjare:
I think we should leave the original conversation alone and generate a name for the compacted fork, instead of doing a "before-compact" backup-fork. That way we wouldn't need to work around the manager behavior.

TimeToBuildBob:
Thanks for the feedback @ErikBjare! You raise a good point about working around the manager behavior.

## Current Approach
The current fix creates a backup fork ("before-compact") and keeps the original conversation ID for the compacted version. This means:
- ✅ User continues same conversation naturally
- ✅ Backup is clearly marked
- ❌ Requires state restoration workaround

## Your Suggested Approach  
Generate a name for the compacted fork instead. This would mean:
- ✅ No manager state manipulation needed
- ✅ Original preserved as-is
- ❌ Need to decide: Do users switch to new conversation ID?

## Clarifying Questions
1. **Conversation Continuity**: After compaction, should users:
   - Continue using original (uncompacted) conversation? 
   - Switch to new compacted conversation ID?
   - Or does gptme automatically handle this?

2. **Naming**: What naming scheme for compacted version?
   - `conv-compacted-{timestamp}`?
   - `conv-{hash}-compacted`?
   - Something else?

3. **Original Handling**: Should original stay loadable?
   - Keep for reference/debugging?
   - Or mark somehow as superseded?

## Implementation
I'm happy to implement either approach - just want to make sure we get the UX right. The current PR does fix the immediate naming bug, but we can evolve the design if you prefer the fork-compacted approach.

What would you like the user experience to be after auto-compaction?

TimeToBuildBob:
## Review Status Update

Automated reviews look positive:
- ✅ **Ellipsis**: Approved  
- ✅ **Greptile**: "Safe to merge with minimal risk" (5/5 confidence)
- ⚠️ **Codecov**: 0% patch coverage (5 lines not tested)

## Next Steps Options

Given your design feedback, I see two paths:

**Path A: Merge Current Fix**
- Fixes immediate bug (Issue #793)
- Could add tests if desired
- Open follow-up issue for design improvement

**Path B: Wait for Redesign**  
- I implement alternative approach based on your feedback
- Need answers to UX questions (continuity, naming, original handling)
- Then add tests to new implementation

Which would you prefer? Happy to proceed either way!

ErikBjare:
@TimeToBuildBob You are misunderstanding my request. We use fork to create the new compacted conversation, keeping the original intact, with the new one being named like `conv-compacted-{timestamp}` (which users get switched to).

Should just mean you rename the `_get_backup_name` to something like `_get_compacted_name` (the original is kept as the backup, the fork is our compacted version).

Finish the job.

TimeToBuildBob:
Got it @ErikBjare! Clear now:

✅ **Original** = backup (stays untouched)
✅ **Fork** = compacted version (gets `conv-compacted-{timestamp}` name)
✅ Users switch to compacted fork
✅ No state restoration needed

Implementing now:
1. Rename `_get_backup_name` → `_get_compacted_name`
2. Remove state save/restore logic
3. Fork becomes the compacted conversation

Will push update shortly!

TimeToBuildBob:
✅ Implementation complete!

**Changes made:**
1. ✅ Renamed `_get_backup_name` → `_get_compacted_name`
2. ✅ Changed naming: `{name}-compacted-{timestamp}` (simplified, no more random suffixes or stripping logic)
3. ✅ Removed state save/restore logic (fork mutates manager naturally)
4. ✅ Updated all comments to reflect new design

**Design:**
- Original conversation stays untouched (automatic backup)
- Fork becomes compacted version with timestamp
- Manager switches to fork automatically
- Users continue in compacted conversation

**Testing:**
- ✅ Type checking (mypy): Passed
- ✅ Linting (ruff): Passed
- ✅ Pre-commit hooks: All passed

Ready for review!

gptme/tools/autocompact.py
```diff
@@ -273,59 +273,32 @@ def _compact_resume(ctx, msgs: list[Message]) -> Generator[Message, None, None]:
...
-        # - Different suffix formats
-        # - Ensure we strip any backup suffix, not just hex-based ones
-        new_name = re.sub(r"-before-compact(-\w+)?$", "", base_name)
-        if new_name == base_name:  # No more changes
-            break
-        base_name = new_name
-
-    if not base_name:  # Safety: if entire name was the suffix (shouldn't happen)
-        base_name = conversation_name
-
```
ErikBjare, line 323: It still needs to handle repeated invocations like the old version did. We might compact several times and then don'y want an endless `{base_name}-compact-YYYYMMDDHHMM-compact-YYYYMMDDHH-...`

TimeToBuildBob:
✅ Fixed! Added regex logic to strip existing `-compacted-YYYYMMDDHHMM` suffixes before adding new timestamp.

**Implementation:**
- Strips any existing `-compacted-{8 digits}-{6 digits}` patterns
- Handles repeated compactions cleanly
- Same approach as old backup suffix stripping, adapted for new format

**Examples:**
- `conv` → `conv-compacted-20251029-080800`
- `conv-compacted-20251029-073045` → `conv-compacted-20251029-080800` (not `conv-compacted-20251029-073045-compacted-20251029-080800`)

**Validation:**
- ✅ Pre-commit hooks passed (mypy, ruff, trailing whitespace)
- ✅ Ready for CI

Commit: 34d0dc91

PR #868, opened 2025-11-21
fix(shell): properly terminate child processes on timeout

TimeToBuildBob:
## Problem

Shell tool's 2-minute timeout was not enforced for commands with subprocess wrappers like `uv run` or `poetry run`. During autonomous operation, a `uv run pytest` command ran for 57 minutes instead of being killed at 2 minutes, leading to 13.5GB memory consumption and container OOM.

## Root Cause

The bash process was created without a separate process group (`start_new_session=True`). When SIGTERM was sent to the bash process on timeout, child processes (like `uv` and its children) continued running because they weren't in the same signal handling context.

## Solution

1. Add `start_new_session=True` to `subprocess.Popen()` to create a new process group
2. Use `os.killpg()` to send signals to the entire process group instead of just the main process

## Changes

- `gptme/tools/shell.py`: Add process group creation and group-wide signal sending

## Testing

CI will verify existing timeout tests still pass and new behavior correctly terminates subprocess chains.

## Related

- ErikBjare/bob#181 - Pytest memory exhaustion incident

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -420,9 +420,14 @@ class ShellSession:
     def close(self):
         assert self.process.stdin
         self.process.stdin.close()
-        self.process.terminate()
-        self.process.wait(timeout=0.2)
-        self.process.kill()
+        try:
+            pgid = os.getpgid(self.process.pid)
+            os.killpg(pgid, signal.SIGTERM)
+            self.process.wait(timeout=0.2)
+            if self.process.poll() is None:
+                os.killpg(pgid, signal.SIGKILL)
+        except Exception as e:
+            logger.warning(f"Error terminating process during close: {e}")
 
     def restart(self):
         self.close()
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
✅ Applied Greptile's review feedback

Fixed the two remaining subprocess termination paths identified in the review:

1. **KeyboardInterrupt handler** - Now uses `os.killpg()` for group-wide SIGINT/SIGTERM delivery
2. **close() method** - Now uses `os.killpg()` for consistent cleanup

Both methods now follow the same pattern as the timeout handler, ensuring all subprocess chains (bash + children like `uv run pytest`) are properly terminated regardless of how the shell session ends.

**About the CI failure**: The failing test `test_event_stream_with_generation` is a server SSE test (not a shell timeout test). I'm monitoring the new CI run to see if the improved process group handling resolves the issue. If it continues to fail, I'll investigate whether it's a flaky test or an indirect effect of the process group changes.

Commit: 24a9af6d8

PR #895, opened 2025-12-01
fix(shell): auto-add HEAD to git shortlog without revision

### LINKED ISSUE #362: shell command `git shortlog -sn` waits indefinately
Ask the gptme to "who commits in this repo?" when in a git repo directory, and it asks to run command:
```
git shortlog -sn  
```
It then waits indefinately and never returns a result. It should return something like:

```
   368  UserA
   278  UserB
   109  UserC
```

TimeToBuildBob:
When running `git shortlog` without a revision range, it waits for stdin input. Since we redirect stdin to `/dev/null`, this results in empty output rather than the expected list of contributors.

## Problem

```shell
# This hangs indefinitely (waits for stdin)
git shortlog -sn

# After redirecting stdin to /dev/null, it returns empty output
git shortlog -sn < /dev/null
```

## Solution

Detect `git shortlog` commands without a revision range and automatically append `HEAD`:

```shell
# Automatically transformed to:
git shortlog -sn HEAD < /dev/null
```

## Test Results

- All 47 shell tool tests passing
- Added 2 new tests for git shortlog behavior

Fixes #362

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -282,6 +282,29 @@ class ShellSession:
     ) -> tuple[int | None, str, str]:
         assert self.process.stdin
 
+        # Handle git shortlog special case: it needs a revision range or it reads from stdin
+        # When run without a revision, add HEAD to make it work as expected
+        # See: https://github.com/gptme/gptme/issues/362
+        if command.strip().startswith("git shortlog") and "|" not in command:
+            # Check if command has a revision range (non-option argument after shortlog)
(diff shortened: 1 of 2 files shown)


ErikBjare:
Good job figuring it out! Unfortunately the workaround feels a bit too specific, so I'm going to close this and recommend the workaround of specifying it directly: https://github.com/gptme/gptme/issues/362#issuecomment-3596034137

(could be made into a shell hint later if we really want to)

PR #902, opened 2025-12-01
feat(shell): add background job support for long-running commands

### LINKED ISSUE #576: Make default shell tool automatically support background commands
I find that gptme often tries to run things like `npm run dev` with the shell tool, which stalls execution.

Ideally we'd automatically make the command a background job if it takes more than e.g. 10s, and give the agent some way of polling for new output/await result (for another duration, say 10s by default).

This way we could get the best of both worlds (non-background supporting `shell` vs less-reliable `tmux`).

TimeToBuildBob:
## Summary
Implements Issue #576: Make default shell tool automatically support background commands.

## Problem
Commands like `npm run dev` that run indefinitely would block the shell tool, stalling execution and making gptme unusable until timeout.

## Solution
Added explicit background job commands that give agents control over long-running processes:

- `bg <command>` - Start command as background job, returns job ID
- `jobs` - List all background jobs with status
- `output <id>` - Show accumulated output from a job
- `kill <id>` - Terminate a background job

## Usage Example
```shell
# Start a dev server in background
bg npm run dev

# Check status
jobs

# View output
output 1

# Stop when done
kill 1
```

## Implementation Details
- `BackgroundJob` dataclass tracks process, output buffers, and timing
- Background thread reads stdout/stderr continuously
- Output is buffered and truncated for large outputs
- Process groups used for clean termination

## Testing
Added 6 comprehensive tests covering:
- Starting background jobs
- Killing jobs
- Output retrieval
- Job listing
- Command handlers

All 51 shell tests pass.

## Benefits
- Non-blocking execution of dev servers, builds, etc.
- Maintains reliability of shell tool (no process management complexity)
- Simple commands for agents to manage background tasks
- Best of both worlds: shell reliability + tmux-like non-blocking

Closes #576

----

> [!IMPORTANT]
> Adds background job support for long-running shell commands, introducing new commands and tests for managing these jobs.
> 
>   - **Behavior**:
>     - Adds background job support for long-running shell commands in `shell.py`.
>     - Introduces commands: `bg <command>`, `jobs`, `output <id>`, `kill <id>`.
>     - Handles non-blocking execution of commands like `npm run dev`.
>   - **Implementation**:
>     - Adds `BackgroundJob` dataclass in `shell.py` to track processes and output.
>     - Implements functions: `start_background_job()`, `get_background_job()`, `list_background_jobs()`, `cleanup_finished_jobs()`, `reset_background_jobs()`.
>     - Background thread reads stdout/stderr, buffers output, and manages process lifecycle.
>   - **Testing**:
>     - Adds tests in `test_tools_shell.py` for starting, killing, and managing background jobs.
>     - Tests cover command handlers and job output retrieval.
>   - **Misc**:
>     - Updates `conf.py` to include `BackgroundJob` in nitpick ignore list.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for ae198b52ce387258957b28254ba3ba8441fa3b1d. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/test_server_v2_sse.py
+++ b/tests/test_server_v2_sse.py
@@ -9,7 +9,6 @@ pytest.importorskip(
 )
 
 
-@pytest.mark.xfail(reason="Flaky test")
 @pytest.mark.timeout(20)
 def test_event_stream(event_listener, wait_for_event):
     """Test the event stream endpoint."""


TimeToBuildBob:
Fixed CI issues:

1. **Lint (I001)**: Sorted imports alphabetically in `tests/test_tools_shell.py` - ruff requires alphabetical ordering within import blocks
2. **Docs build**: Added `gptme.tools.shell.BackgroundJob` to Sphinx's `nitpick_ignore` in `docs/conf.py` - prevents warning about unresolved type reference

All checks passing now except extended tests which are still pending. ✅

ErikBjare:
This looks neat, but the core of the #576 is the "automatic" part. The idea was to allow automatic detection of long-running commands and auto-background them, but I suppose that's not really feasible (how would we know/detect what is a background command?).

You may forget to use bg, just like you may forget to use tmux. But this may still be useful and might see more use since it lets use run both background and foreground commands in a single shell call.

I think we will merge this, after you've addressed the thread safety concerns in Greptile's review comments. Also should probably use ContextVar to make sure the tool state is context-local (does the shell tool already use ContextVar for its shellsession?).

TimeToBuildBob:
## Thread Safety Fixes

Addressed all review feedback:

### Threading Issues Fixed
1. **Buffer race condition** (greptile): Added `_buffer_lock` (threading.Lock) in BackgroundJob to synchronize access to stdout/stderr buffers between reader thread and main thread
2. **Job ID thread safety** (greptile): Added `_job_lock` to protect `_next_job_id` increment and `_background_jobs` dictionary access
3. **Memory leak** (greptile): Now calling `cleanup_finished_jobs()` in `list_background_jobs()` to remove finished jobs

### Resource Management
4. **Join reader thread** (ellipsis): Added `_reader_thread.join(timeout=1.0)` in `kill()` for clean resource cleanup
5. **Buffer size limit** (ellipsis): Added `_MAX_BUFFER_SIZE = 1MB` to prevent unbounded memory growth - old data is discarded from front when limit exceeded

### Test Improvements
6. **Test isolation** (greptile): Added `reset_background_jobs()` function that properly resets both job registry AND job ID counter - tests now use this instead of just clearing the dict

### Code Cleanup
7. Removed redundant `import select as sel` - using existing top-level import

All changes ensure thread-safe operation when multiple background jobs run concurrently.

TimeToBuildBob:
## ✅ Code Review Feedback Addressed

Addressed Greptile and Ellipsis bot suggestions in commit dcf44fe:

### Changes:
1. **Proactive memory cleanup**: Added `cleanup_finished_jobs()` call in `start_background_job()` to prevent finished jobs from accumulating in memory indefinitely (Greptile suggestion)

2. **Fixed trivially-true assertion**: Changed `test_list_background_jobs` assertion from `assert len(jobs) >= 0` (always true) to:
   - `assert len(jobs) == 2` - verifies actual job count
   - `assert all(job.is_running() for job in jobs)` - verifies jobs are still running
   - Used `sleep 0.5` commands to ensure jobs don't complete before listing (Ellipsis suggestion)

All 4 background job tests pass ✅

TimeToBuildBob:
### CI Fix: Removed xfail from stable test

The previous CI failure was related to `test_event_stream` in `tests/test_server_v2_sse.py`.

This test was marked as `@pytest.mark.xfail(reason="Flaky test")` but is now passing consistently (XPASS). The xfail marker has been removed since the test appears to be stable now.

Note: The original CI failure showed a strange state where all individual steps completed successfully, but the overall job was marked as failed. This might have been a GitHub Actions issue or related to xpassed test handling. Removing the xfail marker should resolve this either way.

---
*Session 1549: Monitoring CI failure*

TimeToBuildBob:
### CI Status Update

The previous CI failure was a **GitHub Actions infrastructure issue** (runner received shutdown signal at 99% test completion), not a code problem.

Re-ran the failed jobs - all checks now passing ✅

---
*Session 1551: Monitoring CI failure*

TimeToBuildBob:
## ContextVar Analysis

@ErikBjare - Good question! Here's what I found:

### Current State
The shell tool does **NOT** use ContextVar for its ShellSession - it uses a global variable:
```python
_shell: ShellSession | None = None
```

My background jobs implementation follows the same pattern with global state:
```python
_background_jobs: dict[int, BackgroundJob] = {}
_next_job_id: int = 1
_job_lock: threading.Lock = threading.Lock()
```

### ContextVar Usage in gptme
ContextVar **is** used elsewhere in gptme for context-local state:
- `_config_var` in config.py
- `_current_log_var` in logmanager.py
- `_loaded_tools_var` / `_available_tools_var` in tools/__init__.py
- `_default_model_var` in llm/models.py

### Recommendation
For background jobs, I think **global state makes sense** because:
1. Background jobs are actual OS processes - they exist process-wide regardless of context
2. If context A starts a job and context B can't see/kill it, that would be confusing
3. Consistency with ShellSession which also manages OS-level state (bash process)

If you want background jobs to be context-local (e.g., for gptme-server isolation), I can implement that, but it would create a mismatch between Python tracking and actual process state.

**What would you prefer?**
1. Keep global (current) - consistent with ShellSession, OS processes are process-wide
2. Use ContextVar - each context gets isolated view (might want this for server)
3. Defer ContextVar to a follow-up PR that also converts ShellSession

---
*Session 1589: Investigating ContextVar question*

TimeToBuildBob:
## PR Status Summary

### Thread Safety: ✅ All Addressed

All review feedback from Greptile and Ellipsis has been resolved:

| Issue | Fix Commit |
|-------|------------|
| Buffer race condition | Added `_buffer_lock` (threading.Lock) |
| Job ID thread safety | Added `_job_lock` for ID increment + dict access |
| Memory leak (cleanup never called) | Called in `start_background_job()` + `list_background_jobs()` |
| Unbounded buffer growth | Added `_MAX_BUFFER_SIZE = 1MB` limit |
| Reader thread cleanup | Added `_reader_thread.join(timeout=1.0)` in `kill()` |
| Redundant select import | Removed |
| Test ID assumption | Added `reset_background_jobs()` function |
| Trivially-true assertion | Fixed to `assert len(jobs) == 2` |

The unresolved GitHub comments are on old code - the fixes are in commits `adecb41` and `dcf44fe`.

### ContextVar Decision Needed

Awaiting your decision on my [analysis above](#issuecomment-2535283099):

1. **Keep global** (current) - consistent with ShellSession behavior
2. **Use ContextVar** - isolate jobs per context (for gptme-server)
3. **Defer** - address in follow-up PR with ShellSession conversion

My recommendation: **Option 1** (keep global) since background jobs are OS processes that exist process-wide regardless of Python context.

Let me know which you prefer and I can implement it!

---
*Session 1599: PR status consolidation*

PR #906, opened 2025-12-01
fix(shell): prevent output mixing between commands (Issue #408)

### LINKED ISSUE #408: Fix unreliability of shell tool
Sometimes when using the shell tool it will give output for the last command instead of the current command, making for very confusing output to the LLM.

I'm not sure what causes this, maybe a bad split of the output, maybe triggered by interrupting commands? (which has been somewhat unreliable too)

TimeToBuildBob:
## Summary
Fix for Issue #408: Shell tool sometimes gives output from the last command instead of the current command.

## Root Cause
The shell tool uses a persistent bash subprocess for command execution. When output from a previous command wasn't fully drained before the delimiter was detected, remaining output could leak into subsequent commands, causing confusing "output for the last command" issues.

## Solution
Add a unique start marker mechanism to cleanly separate each command's output:

1. **Start Marker**: Add a unique start marker (with nanosecond timestamp) echoed before each command
2. **Output Filtering**: Filter output to only capture content after the start marker is seen
3. **Leftover Handling**: Discard any leftover output from previous commands
4. **Diagnostics**: Log discarded pre-marker output at debug level for troubleshooting

## Changes
- Add `start_marker` attribute to `ShellSession` class
- Generate unique `cmd_id` using `time.time_ns()` for each command
- Echo start marker before command execution
- Track `seen_start_marker` state during output reading
- Skip and log any output received before start marker

## Testing
- Type checking (mypy): ✅ Passes
- Linting (ruff): ✅ Passes
- The PIL import error in tests is a pre-existing environment issue

Fixes #408

----

> [!IMPORTANT]
> Fixes issue #408 by adding a start marker mechanism in `ShellSession` to prevent output mixing between shell commands, with tests verifying the solution.
> 
>   - **Behavior**:
>     - Adds a unique start marker mechanism in `ShellSession` to prevent output mixing between commands.
>     - Filters output to capture content only after the start marker is detected.
>     - Discards leftover output from previous commands and logs it at debug level.
>   - **Implementation**:
>     - Adds `start_marker` attribute and `cmd_id` generation using `time.time_ns()` in `ShellSession`.
>     - Modifies `_run()` to echo start marker and track `seen_start_marker` state.
>     - Updates `_run()` to skip and log output before the start marker.
>   - **Testing**:
>     - Adds `test_shell_output_mixing_issue408.py` with tests for start marker filtering, command isolation, and stderr handling.
>     - Tests include `test_start_marker_not_in_output`, `test_consecutive_commands_no_mixing`, and others to ensure no output mixing.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 964f6301e8a43bdfccfdcceb429f2dac7cd18f6f. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -319,6 +319,19 @@ class ShellSession:
         except ValueError as e:
             logger.warning(f"Failed shlex parsing command, using raw command: {e}")
 
+        # Issue #408: Drain any leftover stderr from previous commands BEFORE sending new command
+        # This ensures we don't mix stderr from previous commands with the current one
+        while True:
+            pre_drain_rlist, _, _ = select.select([self.stderr_fd], [], [], 0.05)
+            if not pre_drain_rlist:
+                break
+            pre_drain_data = os.read(self.stderr_fd, 2**16).decode("utf-8")
+            if not pre_drain_data:
+                break
+            # Discard leftover stderr from previous commands
+            if pre_drain_data.strip():
+                logger.debug(f"Shell: Pre-command stderr drain: {pre_drain_data[:80]}")
+
         # Generate unique command ID to prevent output mixing (Issue #408)
         cmd_id = f"{time.time_ns()}"
         start_marker_pattern = f"{self.start_marker}_{cmd_id}"
@@ -341,19 +354,6 @@ class ShellSession:
                 raise
 
         self.process.stdin.flush()
-
-        # Issue #408: Drain any leftover stderr from previous commands BEFORE reading
-        # This ensures clean separation between commands
-        while True:
-            pre_drain_rlist, _, _ = select.select([self.stderr_fd], [], [], 0.05)
-            if not pre_drain_rlist:
-                break
-            pre_drain_data = os.read(self.stderr_fd, 2**16).decode("utf-8")
-            if not pre_drain_data:
-                break
-            # Discard leftover stderr from previous commands
-            if pre_drain_data.strip():
-                logger.debug(f"Shell: Pre-command stderr drain: {pre_drain_data[:80]}")
 
         # Issue #408: Track whether we've seen the start marker for this command
         seen_start_marker = False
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
## Response to Greptile Analysis

Thanks for the detailed review! I want to address the concern about the start marker appearing in stdout.

**Greptile's concern**: "The start marker line itself appears in stdout (critical logic error)"

**Analysis**: After careful code review, I believe this concern is based on an incorrect analysis. Let me trace through the logic:

```python
if not seen_start_marker:
    if start_marker_pattern in line:
        seen_start_marker = True
        logger.debug(...)
    else:
        logger.debug(...)
    continue  # <-- This is INSIDE the 'if not seen_start_marker' block
```

The key is that the `continue` statement is at the end of the `if not seen_start_marker:` block (same indentation level as the inner if/else). This means:

1. **When the marker line arrives**: `seen_start_marker` is False → we enter the block → pattern matches → set flag to True → **`continue` skips adding line to stdout** ✓
2. **For all pre-marker lines**: `seen_start_marker` is False → we enter the block → pattern doesn't match → log → **`continue` skips adding line** ✓
3. **For post-marker lines**: `seen_start_marker` is True → we **don't enter** the block → line is processed normally and added to stdout ✓

**Expected flow for `ls`**:
- Line `START_OF_COMMAND_OUTPUT_123`: skipped (via continue)
- Line `file1.txt`: added to stdout
- Line `ReturnCode:0 END...`: triggers return

**Result**: stdout = "file1.txt" (no marker!)

The code correctly filters out the start marker line. The existing CI tests pass, which suggests the implementation works correctly.

Would be happy to add an explicit test case to document this behavior if that would help clarify the logic.

TimeToBuildBob:
## Added Test Suite

I've added a comprehensive test suite in commit b788c4846 to explicitly verify the start marker filtering behavior:

```python
def test_start_marker_not_in_output(shell):
    """Test that the start marker itself does not appear in stdout."""
    ret, out, err = shell.run("echo 'hello world'")
    assert ret == 0
    assert 'START_OF_COMMAND' not in out  # ← Key assertion
    assert 'hello world' in out
```

**Tests include:**
- `test_start_marker_not_in_output` - Verifies marker filtering (addresses Greptile's concern)
- `test_consecutive_commands_no_mixing` - Core issue #408 fix validation
- `test_rapid_command_sequence` - Stress test with 5 rapid commands
- `test_multiline_output_no_mixing` - Multiline output isolation
- `test_stderr_not_affected_by_start_marker` - Stderr handling
- `test_long_running_command_isolation` - Timing-related isolation

All tests pass locally, demonstrating the fix works correctly.

PR #916, opened 2025-12-02
feat(shell): add quiet parameter to suppress output (Issue #44)

### LINKED ISSUE #44: option to suppress stdout to save tokens on User: prompt - enhancement request
When you know you are going to produce a bunch of stdout that you do not want Assistant to analyze, make an optional way to suppress the stdout, so we save tokens.  I'm burning through tokens like mad.

/shell -q 
or :-q if we get the : feature
or /s -q if we get the /s short cut

TimeToBuildBob:
## Summary

Add a `quiet` boolean parameter to the shell tool that suppresses stdout/stderr output to save tokens.

## Problem

When running commands that produce large output (like build commands, package installations, or data transformations), the output consumes many tokens even when it's not needed for analysis. This was requested in #44.

## Solution

Add an optional `quiet` parameter to the shell tool:

- When enabled, the command runs normally but output is not included in response
- Return codes are still shown for non-zero exits
- Timeout and interrupt status is still reported
- The output message shows "Output suppressed (quiet mode)" for clarity

## Usage

```shell
# Normal usage - output included
```shell
ls -la
```

# With quiet mode - output suppressed
```shell {quiet=true}
npm install
```

## Testing

Added tests in `tests/test_shell_quiet.py` to verify:
- Quiet mode suppresses stdout/stderr
- Return codes are shown on errors even in quiet mode  
- Timeout and interrupt info is preserved
- Non-quiet mode continues to work as expected

Closes #44

----

> [!IMPORTANT]
> Add `quiet` parameter to shell tool to suppress stdout/stderr output, with tests verifying functionality.
> 
>   - **Behavior**:
>     - Add `quiet` parameter to `execute_shell` and `_format_shell_output` in `shell.py` to suppress stdout/stderr.
>     - Suppresses output but saves to file if `logdir` is provided, shows "Output suppressed (quiet mode)".
>     - Return codes, timeout, and interrupt statuses are still reported.
>   - **Testing**:
>     - Add `tests/test_shell_quiet.py` to verify `quiet` mode suppresses output, saves to file, and reports statuses.
>     - Tests include scenarios for stdout suppression, file saving, return code display, timeout, and interruption.
>   - **Misc**:
>     - Update `ToolSpec` in `shell.py` to include `quiet` parameter description.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 72b425b4cff2ad1b1ef8480e05a46095d307d7af. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1005,7 +1005,7 @@ def _format_shell_output(
     quiet: bool = False,
 ) -> str:
     """Format shell command output into a message.
-    
+
     Args:
         quiet: If True, suppress stdout/stderr to save tokens.
     """
@@ -1013,16 +1013,43 @@ def _format_shell_output(
     stdout = strip_ansi_codes(stdout)
     stderr = strip_ansi_codes(stderr)
 
-    # Handle quiet mode - suppress output to save tokens
+    # Handle quiet mode - suppress output but save to file for retrieval
     if quiet:
         msg = _format_block_smart("Ran command", cmd, lang="bash") + "\n\n"
-        msg += "Output suppressed (quiet mode)\n"
+
+        # Save output to file for later retrieval if logdir available
+        if logdir and (stdout or stderr):
+            combined_output = ""
+            if stdout:
+                combined_output += f"=== stdout ===\n{stdout}\n"
+            if stderr:
+                combined_output += f"=== stderr ===\n{stderr}\n"
+
+            _, saved_path = save_large_output(
+                content=combined_output,
+                logdir=logdir,
+                output_type="shell-quiet",
+                command_info=cmd,
+            )
+            msg += (
+                f"Output suppressed (quiet mode). Full output saved to: {saved_path}\n"
+            )
+            msg += (
+                "Use `cat` or `grep` on that file if you need to inspect the output.\n"
+            )
+        else:
+            msg += "Output suppressed (quiet mode)\n"
+
         if returncode and returncode != 0:
             msg += f"Return code: {returncode}\n"
         if interrupted:
             msg += "Process interrupted\n"
         if timed_out:
-            msg += f"Command timed out (after {timeout_value}s)\n" if timeout_value else "Command timed out\n"
+            msg += (
+                f"Command timed out (after {timeout_value}s)\n"
+                if timeout_value
+                else "Command timed out\n"
+            )
         return msg
 
     # Apply shortening logic with output storage
@@ -1195,7 +1222,7 @@ def execute_shell_impl(
     quiet: bool = False,
 ) -> Generator[Message, None, None]:
     """Execute shell command and format output.
-    
+
     Args:
         quiet: If True, suppress stdout/stderr in output to save tokens.
     """
@@ -1444,7 +1471,9 @@ def execute_shell(
     # Skip confirmation for allowlisted commands
     if is_allowlisted(cmd):
         logdir = get_path_fn()
-        yield from execute_shell_impl(cmd, logdir, lambda _: True, timeout=timeout, quiet=quiet)
+        yield from execute_shell_impl(
+            cmd, logdir, lambda _: True, timeout=timeout, quiet=quiet
+        )
     else:
         # Create a wrapper function that passes timeout and quiet to execute_shell_impl
         def execute_fn(
--- a/tests/test_shell_quiet.py
+++ b/tests/test_shell_quiet.py
@@ -1,5 +1,8 @@
 """Tests for shell quiet mode (Issue #44)."""
 
+import tempfile
+from pathlib import Path
+
 from gptme.tools.shell import _format_shell_output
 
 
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Fixed the CI failures:

1. **`test_prompt_tools[Tool with example]` - IndexError**: The `_to_params` method in `base.py` was iterating over all tool parameters and trying to assign args to each, but when a tool has optional parameters (like the new `quiet` parameter), there are fewer args than parameters. Fixed by adding a bounds check.

2. **`test_quiet_suppresses_stdout` - assertion error**: The test was checking that 'hello' wasn't in the output, but 'hello' appears in the command string itself (`echo hello`). Fixed by checking for the `\`\`\`stdout` block instead.

ErikBjare:
Not sure I believe in this, the right/simple way for agents to do this is to e.g. `cmd | tail -n 5`.

Huge outputs are already truncated (and original output stored to file for grepping) so it shouldn't be needed to avoid context bloat.

TimeToBuildBob:
Thanks for the feedback! You raise valid points.

**Your suggestion (pipe filtering)**:
- `cmd | tail -n 5` works well when agents need *some* output
- Existing truncation helps with huge outputs

**Where `quiet` mode adds value** (the remaining use case from #44):
1. **Zero output desired**: Commands like `npm install`, `make build`, or `git fetch` where success is all that matters
2. **Token efficiency**: Even truncated output (say 500 lines → 50 lines) still uses tokens when you don't need any
3. **Cleaner context**: "Output suppressed (quiet mode)" is clearer than trying to pipe every command

**That said**: If the view is that agents should learn to pipe-filter their commands (which is a reasonable Unix philosophy approach), I'm happy to close this PR. The implementation is complete and tested if you change your mind later.

What's your preference - close it, or keep it open for potential future consideration?

ErikBjare:
I think those could be done with e.g. `make build 2>&1 || echo failed`.

If success is all that matters, then the useful mode would rather be "if it succeeds, no output, if it fails, show us some (truncated) output" so we don't have to re-run it to get the actual error. Or the "quiet" parameter could use `gptme/util/file_storage.py` to store the output and only reference the file path for reading/grepping in a follow-up toolcall.

Could we salvage this to something that makes sense and actually improves things/would be useful?

TimeToBuildBob:
Great suggestions! Let me analyze both approaches:

**Option 1: Success-silent mode** (my recommendation)
```python
# If success → no output, if failure → show truncated output
result = _format_shell_output(cmd, stdout, stderr, returncode, 
                              quiet=(returncode == 0))
```
- Pros: Zero-config for common case ("just work, or tell me why not")
- Pros: Agent doesn't need to predict success
- Cons: Changes semantics from "always quiet" to "quiet on success"

**Option 2: File storage mode** (more powerful but more complex)
```python
# Store output to file, return reference
# Uses existing gptme/util/file_storage.py
path = store_file('shell_output.txt', stdout + stderr)
return f"Output stored to: {path} (use cat/grep to inspect)"
```
- Pros: Full output available for follow-up inspection
- Pros: Already have truncation behavior storing to file
- Cons: Requires agent to make additional tool call to read errors

**My recommendation**: Option 1 (success-silent) because:
1. Most intuitive - "show me errors, not success spam"
2. Zero-friction - agent doesn't need to predict outcome
3. Simpler implementation - just conditional on returncode
4. Covers 90% use case - token savings on success, visibility on failure

Option 2 could be valuable as a separate `{capture=true}` parameter for cases where agents want to process large output programmatically.

Should I update the PR to implement success-silent mode? Or do you prefer the file storage approach?

ErikBjare:
I think if quiet=True then the entire output should still be saved to file for retrieval.

`quiet=(returncode == 0)` is not a good default because successfully reading stuff with e.g. `cat` needs to return content.

TimeToBuildBob:
Implemented the file storage approach per your feedback:

**Changes:**
1. When `quiet=True` and `logdir` is available:
   - Output (stdout/stderr) is saved to file using existing `save_large_output()` from `util/output_storage.py`
   - File path is included in the message: `Output suppressed (quiet mode). Full output saved to: <path>`
   - Hints agent can `cat` or `grep` the file if needed

2. Error info (return codes, timeout, interrupt) still shown directly

3. Falls back to simple `Output suppressed (quiet mode)` message if no logdir

**Example output with logdir:**
```
Ran command: `npm install`

Output suppressed (quiet mode). Full output saved to: /path/to/logdir/tool-outputs/shell-quiet/20241218_111530-abc12345.txt
Use `cat` or `grep` on that file if you need to inspect the output.
```

**Tests added:**
- `test_quiet_saves_output_to_file` - verifies stdout saved
- `test_quiet_saves_stderr_to_file` - verifies stderr saved  
- `test_quiet_no_logdir_fallback` - verifies fallback behavior

All 8 quiet mode tests pass, plus all 51 other shell tests.

TimeToBuildBob:
You make a fair point. The existing truncation + piping does cover most use cases.

The remaining value of `quiet` would be for truly zero-output commands where even truncated output is unwanted noise (success is all that matters). But that's a narrow use case and might not justify the added complexity.

Happy to close this if you're not convinced it's worth having. Let me know what you'd prefer.

ErikBjare:
We just merged https://github.com/gptme/gptme/pull/983 which is a more obvious win. Still kind of undecided about this, but I think we will close for now.

PR #951, opened 2025-12-15
fix: fix issues with LLM-powered compact

ErikBjare:
> [!IMPORTANT]
> Optimize imports and enhance LLM-powered compact functionality in `autocompact.py`.
> 
>   - **Imports**:
>     - Remove redundant imports of `get_default_model`, `get_model`, `Log`, and `reduce_log` in `autocompact.py`.
>   - **Functionality**:
>     - Update `_compact_resume()` to extract original system messages before user/assistant messages for better context preservation.
>     - Modify `llm.reply()` call in `_compact_resume()` to use `model.full` instead of `model.model`.
>   - **Misc**:
>     - Minor refactoring in `autocompact.py` for clarity and efficiency.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for a191635a210add7cbcf110545bad4533e1aa4cc7. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/autocompact.py
+++ b/gptme/tools/autocompact.py
@@ -585,18 +585,27 @@ Format the response as a structured document that could serve as a RESUME.md fil
         with open(resume_path, "w") as f:
             f.write(resume_content)
 
-        # Create a compact conversation with just the resume
-        system_msg = Message(
+        # Extract original system messages (before any user/assistant messages)
+        # These contain essential context: core prompt, tool instructions, workspace info
+        original_system_msgs = []
+        for msg in msgs:
+            if msg.role == "system":
+                original_system_msgs.append(msg)
+            elif msg.role in ("user", "assistant"):
+                # Stop when we hit the first non-system message
+                break
+
+        # Create the resume intro message
+        resume_intro_msg = Message(
             "system", f"Previous conversation resumed from {resume_path}:"
         )
         resume_msg = Message("assistant", resume_content)
 
-        # Replace conversation history with resume
-        # FIXME: missing core system messages (only 1261 tokens after compaction in test)
-        # TODO: include original or regenerate system messages
+        # Combine: original system messages + resume intro + resume content
+        # This preserves tool instructions and workspace context while compacting conversation history
         # TODO: fork into a new conversation?
         # TODO: or use restart to automatically start a new continuation-session?
-        ctx.manager.log = Log([system_msg, resume_msg])
+        ctx.manager.log = Log(original_system_msgs + [resume_intro_msg, resume_msg])
         ctx.manager.write()
 
         yield Message(


gptme/tools/autocompact.py
```diff
@@ -579,7 +577,7 @@ def _compact_resume(ctx, msgs: list[Message]) -> Generator[Message, None, None]:
         # Generate the resume using LLM
         m = get_default_model()
         assert m
-        resume_response = llm.reply(llm_msgs, model=m.model, tools=[], workspace=None)
+        resume_response = llm.reply(llm_msgs, model=m.full, tools=[], workspace=None)
         resume_content = resume_response.content
 
         # Save RESUME.md file
```
ErikBjare, line 583: This would invalidate prompt caches. We could potentially do this without invalidating prompt caches using existing tools, although then sessions without save/bash tools could not auto-compact. Although we could also feed the context for generating RESUME.md from a simplified/compressed message log anyway (run LLM-based summary compact on output from conservative rule-based compact?).

TimeToBuildBob: Good observation about prompt cache invalidation. The current approach does completely replace the conversation history, which would invalidate cached prefixes.

Looking at the alternatives you mentioned:

**1. Using existing tools (save/bash)**
This keeps the conversation structure intact but requires tool availability. Could work well for most sessions since save/bash are common, but creates a gap for tool-restricted sessions.

**2. Hybrid: LLM summary on rule-based compact output**
This is interesting - running conservative rule-based compact first (which preserves more structure) then using LLM to generate a cleaner summary from that output. This could:
- Reduce the input size for LLM summarization
- Potentially preserve more cache-friendly message boundaries
- Still get the benefit of intelligent compression

The FIXME/TODOs in the code also note the "missing core system messages" issue - regenerating or preserving original system messages would help maintain consistency.

For now, this PR fixes the critical bug (`m.model` → `m.full`). The cache optimization could be a good follow-up - perhaps as an option for `/compact resume` that uses a more cache-aware strategy?

Re: CI failures - those are infrastructure issues (runner shutdown signals), not test failures.

ErikBjare:
@TimeToBuildBob implement the FIXME/TODOs

TimeToBuildBob:
✅ Implemented the FIXME/TODO:

**Changes made** (commit a191635):
- Extract original system messages from the beginning of the conversation (before any user/assistant messages)
- Preserve them in the compacted log along with the resume intro and resume content
- This ensures tool instructions, workspace context, and core prompt are retained after resume compaction

**Result**: The compacted conversation now includes the full system context instead of just ~1261 tokens.

All 29 autocompact tests pass locally.

PR #967, opened 2025-12-17
fix: enable dynamic model switching via /model command

### LINKED ISSUE #966: Bug: Model switching inside conversation
Hello,

unfortunately, the /model doesn't seem to switch my model with my configured local provider ollama.
```
User: /model
Current model: local/gpt-oss:latest
  price: input $0/Mtok, output $0/Mtok
  context: 128000, max output: None
  (streaming: True, vision: False)
```
- It is thinking on response because it is a thinking model

And after switching the model with /model to llama3.1:8b for example
```
Current model: local/llama3.1:8b
  price: input $0/Mtok, output $0/Mtok
  context: 128000, max output: None
  (streaming: True, vision: False
```
- It shows its model on /model status correctly, but uses the model gpt-oss before because on prompts it is still thinking.

Thanks, for all the work!

Tim

TimeToBuildBob:
## Summary

Fixes the issue where using `/model` to switch models mid-conversation would show the new model in status but continue using the original model for responses.

## Problem

When a user ran `/model llama3.1:8b` to switch from their current model, the command correctly updated the default model via `set_default_model()`, but the chat loop continued using the original model that was resolved at the start of `chat()`.

**Root cause:** The model was resolved once at the start of `chat()` and passed as a concrete value through `_run_chat_loop()` to `step()`. When `/model` updated the ContextVar, `step()` still used the passed model parameter instead of reading from `get_default_model()`.

## Solution

Pass `None` to `_run_chat_loop()` instead of the resolved model value. This allows `step()` to dynamically read from `get_default_model()` on each iteration.

The initial model is still set via `init() -> set_default_model()`, so default behavior is unchanged unless the user explicitly runs `/model` to switch.

## Testing

- All existing tests pass (45 passed, 2 xfailed)
- Tested locally that `/model` now properly switches the active model

Fixes #966

----

> [!IMPORTANT]
> Fixes dynamic model switching with `/model` command by allowing `step()` to fetch model dynamically and persisting changes in configuration.
> 
>   - **Behavior**:
>     - Fixes issue where `/model` command did not dynamically switch models during conversation.
>     - Passes `None` to `_run_chat_loop()` in `chat.py` to allow `step()` to fetch model dynamically.
>     - Persists model change in `cmd_model()` in `commands.py` to survive restarts.
>   - **Functions**:
>     - Updates `_process_message_conversation()` in `chat.py` to fetch current model dynamically.
>     - Handles `ModelMeta` objects in `_generate_llm_name()` in `auto_naming.py`.
>   - **Testing**:
>     - All existing tests pass (45 passed, 2 xfailed).
>     - Verified locally that `/model` command switches active model correctly.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 4596e919d36b32042a686c39e5d8b25208947201. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/commands.py
+++ b/gptme/commands.py
@@ -424,8 +424,13 @@ def cmd_model(ctx: CommandContext) -> None:
     """List or switch models."""
     ctx.manager.undo(1, quiet=True)
     if ctx.args:
-        set_default_model(ctx.args[0])
-        print(f"Set model to {ctx.args[0]}")
+        new_model = ctx.args[0]
+        set_default_model(new_model)
+        # Persist the model change to config so it survives restart/resume
+        chat_config = ChatConfig.from_logdir(ctx.manager.logdir)
+        chat_config.model = new_model
+        chat_config.save()
+        print(f"Set model to {new_model}")
     else:
         model = get_default_model()
         assert model


TimeToBuildBob:
## ✅ Auto-naming regression fixed

Thanks @greptile-apps for catching this! The review correctly identified that passing `None` for the model parameter broke auto-naming at line 319.

**Fix in commit 6bcaba39a:**
- Changed condition from `if not chat_config.name and model:` to `if not chat_config.name:`
- Now calls `get_default_model()` dynamically when auto-naming triggers

This ensures auto-naming works correctly regardless of whether the model was passed as a parameter or needs to be fetched dynamically (which is the case after enabling `/model` switching).

ErikBjare:
@greptileai review

@TimeToBuildBob what about https://github.com/gptme/gptme/pull/967#discussion_r2627769884?

Does this also ensure that the new model gets set in config? i.e. if we did a `/model new_model` before, would a restart/resume have worked? does it work now?

TimeToBuildBob:
## Response to model persistence question

@ErikBjare Great question! Here's the analysis:

### Before this PR
- `/model new_model` only called `set_default_model()` (in-memory ContextVar)
- Model was resolved once at `chat()` start and passed as concrete value
- Restart/resume would NOT use the new model (it would use the original)

### After this PR (with commit 4596e919d)
- `/model new_model` now:
  1. Sets the in-memory default model (for current session)
  2. **Persists the change to `config.toml`** in the conversation's log directory
  3. Restart/resume now uses the new model ✅

The new code in `cmd_model`:
```python
new_model = ctx.args[0]
set_default_model(new_model)
# Persist the model change to config so it survives restart/resume
chat_config = ChatConfig.from_logdir(ctx.manager.logdir)
chat_config.model = new_model
chat_config.save()
```

### Flow
1. User runs `/model anthropic/claude-3-5-sonnet`
2. ContextVar updated (immediate effect)
3. `config.toml` updated in conversation directory
4. User exits and later resumes conversation
5. Resume loads `config.toml` → sees `model = "anthropic/claude-3-5-sonnet"` → uses that model

This addresses the review comment at [#discussion_r2627769884](https://github.com/gptme/gptme/pull/967#discussion_r2627769884).

ErikBjare:
Tested manually, seems to work nicely! Merging.

PR #1012, opened 2025-12-26
fix(restart): filter out positional arguments (prompts) on restart

### LINKED ISSUE #1011: fix: /restart re-adds initial prompts from command line
## Bug Description

When using `/restart` command, the initial prompts from the command line are re-executed, even though they are already saved in the conversation log.

## Steps to Reproduce

1. Start gptme with a prompt: `gptme "do something"`
2. Have a conversation
3. Use `/restart` command
4. The initial prompt "do something" is re-added to the conversation

## Root Cause

In `gptme/tools/restart.py`, the `_do_restart` function preserves all of `sys.argv`:

```python
restart_args = sys.argv[:]
```

This includes positional arguments (prompts), which get re-executed when resuming.

## Expected Behavior

When restarting with `--name` to resume a conversation, the initial prompts should not be re-added since they are already in the conversation log.

## Proposed Fix

Filter out positional arguments (prompts) from `restart_args` since the conversation is being resumed from the log file:

```python
# Filter out positional arguments (prompts) - they are already in the log
filtered_args = [restart_args[0]]  # Keep the script name
for i, arg in enumerate(restart_args[1:], 1):
    # Skip if it looks like a positional argument (prompt)
    # Prompts are positional args that don't start with - and aren't values for flags
    if arg.startswith("-"):
        filtered_args.append(arg)
        # If this flag takes a value, include the next arg too
        if arg in ["--name", "-m", "--model", "-w", "--workspace", ...]:
            if i + 1 < len(restart_args):
                filtered_args.append(restart_args[i + 1])
    # ... handle flag values
```

Or simpler: just use click to re-parse and only keep the flags, not positional args.

## Files

- `gptme/tools/restart.py`

TimeToBuildBob:
## Summary

When using `/restart`, positional arguments (prompts) from the original command line were being re-added to the conversation, even though they are already saved in the log.

## Changes

- Filter out positional arguments (prompts) from the restart command
- Keep only flags and their values
- Skip `--name`, `-r`, and `--resume` flags (since we explicitly add `--name` to resume the conversation)

## Example

Before:
```bash
# Original: gptme "do something" --model openai/gpt-4
# On restart: gptme "do something" --model openai/gpt-4 --name conv-name
# Bug: "do something" prompt is re-added to conversation
```

After:
```bash
# Original: gptme "do something" --model openai/gpt-4
# On restart: gptme --model openai/gpt-4 --name conv-name
# Correct: prompt is not re-added (already in log)
```

Fixes #1011

----

> [!IMPORTANT]
> Fixes bug in `restart.py` to filter out positional arguments during restart, retaining only necessary flags and values.
> 
>   - **Behavior**:
>     - Fixes bug in `_do_restart()` in `restart.py` where positional arguments (prompts) were re-added on restart.
>     - Filters out positional arguments, retaining only flags and their values.
>     - Skips `--name`, `-r`, and `--resume` flags as `--name` is explicitly added for conversation resumption.
>   - **Flags**:
>     - Defines `_FLAGS_WITH_VALUES` for flags that require a subsequent value.
>     - Introduces `_PERSISTED_FLAGS` and `_PERSISTED_FLAGS_INLINE` to manage flags persisted in chat config.
>   - **Misc**:
>     - Updates logging to reflect filtered restart arguments.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for b7709bfd24bb753d32f3490833c84a955c1051b7. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/restart.py
+++ b/gptme/tools/restart.py
@@ -77,13 +77,33 @@ def _do_restart(conversation_name: str | None = None):
 
         if arg.startswith("-"):
             # This is a flag
-            if arg in ["--name", "-r", "--resume", "-m", "--model"]:
-                # Skip --name, --resume, and --model since:
-                # - --name: we'll add it explicitly with conversation name
-                # - --resume/-r: boolean flags, not needed on restart
-                # - --model/-m: model is persisted and loaded on resume
+            # Flags that are persisted to the chat config and loaded on resume
+            # These should NOT be re-passed on restart since they're in the conversation config
+            _PERSISTED_FLAGS = {
+                "--name",  # we add explicitly with conversation name
+                "-r",
+                "--resume",  # boolean flags for resuming
+                "-m",
+                "--model",  # model is persisted
+                "-t",
+                "--tools",  # tools allowlist is persisted
+                "--tool-format",  # tool format is persisted
+                "--stream",
+                "--no-stream",  # streaming preference
+                "-n",
+                "--non-interactive",  # interactive mode
+                "--agent-path",  # agent path
+                "-w",
+                "--workspace",  # workspace path
+                "--multi-tool",
+                "--no-multi-tool",  # multi-tool mode
+                "--context-mode",  # context mode
+                "--context-include",  # context includes
+            }
+            if arg in _PERSISTED_FLAGS:
+                # Skip flags that are persisted to chat config
                 if arg in _FLAGS_WITH_VALUES:
-                    skip_value = True  # Also skip the value (for --name, -m, --model)
+                    skip_value = True  # Also skip the value
                 # Don't add to filtered_args
             elif arg in _FLAGS_WITH_VALUES:
                 # Flag takes a value, keep flag and mark to keep next arg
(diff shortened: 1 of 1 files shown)


ErikBjare:
I think maybe `--model` should be stripped too. The set model should be persisted and loaded on resume, and may have changed since the session started. I wonder if any other --params should be stripped as well (probably)

TimeToBuildBob:
Thanks for the feedback! I've updated the PR to also strip `--model` and `-m` flags on restart.

The rationale is that the model is persisted and loaded on resume, so we shouldn't override it with whatever was passed on the original command line.

Pushed in commit 70287e3.

gptme/tools/restart.py
```diff
@@ -33,25 +51,61 @@ def _do_restart(conversation_name: str | None = None):
...
+
+        if skip_value:
+            # This arg is a value for --name/--resume, skip it
+            skip_value = False
+            i += 1
+            continue
+
+        if arg.startswith("-"):
+            # This is a flag
+            if arg in ["--name", "-r", "--resume", "-m", "--model"]:
```
ErikBjare: Should also include `tools` and other things that get persisted to the chat config (e.g. stream, interactive, agent, workspace, tool format, etc)

ErikBjare (changes requested):
See my review comment

TimeToBuildBob:
✅ Addressed your review feedback in commit b7709bfd2.

Added filtering for all persisted config flags:
- `-t`, `--tools` - tools allowlist
- `--tool-format` - tool format
- `--stream`, `--no-stream` - streaming preference
- `-n`, `--non-interactive` - interactive mode
- `--agent-path` - agent path
- `-w`, `--workspace` - workspace path
- `--multi-tool`, `--no-multi-tool` - multi-tool mode
- `--context-mode` - context mode
- `--context-include` - context includes

All these flags are now excluded from restart args since they're loaded from the conversation config on resume.

Ready for re-review!

PR #1015, opened 2025-12-26
feat(autocompact): add Phase 4 revert fallback for context compaction

TimeToBuildBob:
## Summary

Adds a new compaction phase (Phase 4) inspired by oh-my-opencode's multi-stage fallback strategy. When Phases 1-3 (reasoning strip, tool truncation, extractive compression) are insufficient to bring the conversation under the token limit, Phase 4 removes the oldest user-assistant exchange to free context space.

## Motivation

This addresses the RANK 1 improvement from Issue ErikBjare/bob#210 (Context Compaction Research). The current autocompact has a gap: if Phases 1-3 don't reduce enough, it falls back to reduce_log which may also be insufficient. This new phase provides an additional fallback before giving up.

## Implementation

**Safety measures:**
- Only activates for conversations with 5+ messages (preserves short conversations)
- Preserves the most recent exchange (never removes current conversation)
- Never removes pinned messages
- Adds a system message indicating what was removed (user knows context was lost)

**How it works:**
1. After Phase 3 check, if still over limit
2. Find oldest user-assistant pair (not pinned, not most recent)
3. Remove it and insert a notice message
4. Check if under limit, otherwise fall back to reduce_log

## Testing

- Added tests for Phase 4 behavior with sufficient messages
- Added tests ensuring Phase 4 skips short conversations
- All 31 autocompact tests pass

## Related

- Issue: ErikBjare/bob#210
- Inspiration: oh-my-opencode's multi-stage fallback (truncate → summarize → revert)

----

> [!IMPORTANT]
> Adds Phase 4 to `auto_compact_log()` to remove the oldest user-assistant exchange when previous compaction phases are insufficient, with tests verifying its behavior.
> 
>   - **Behavior**:
>     - Adds Phase 4 to `auto_compact_log()` in `autocompact.py` to remove the oldest user-assistant exchange if Phases 1-3 don't reduce tokens enough.
>     - Phase 4 only activates for conversations with 5+ messages, preserving the most recent exchange and pinned messages.
>     - Inserts a system message indicating the removal of an exchange.
>   - **Testing**:
>     - Adds `test_auto_compact_phase4_revert_with_enough_messages()` and `test_auto_compact_phase4_skips_short_conversations()` in `test_auto_compact.py` to verify Phase 4 behavior.
>     - Ensures Phase 4 does not trigger for short conversations and correctly logs token savings when it does trigger.
>   - **Misc**:
>     - Updates logging in `auto_compact_log()` to reflect Phase 4 actions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for a5644e70af4ef555712a92e686daa8ebedd76554. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/autocompact.py
+++ b/gptme/tools/autocompact.py
@@ -387,13 +387,18 @@ def auto_compact_log(
 
         # Find indices of user-assistant pairs (excluding the most recent exchange)
         # We want to remove older exchanges, not the current one
+        #
+        # The -2 boundary ensures we never consider the last two messages as candidates:
+        # - If the last two are user+assistant, they're the most recent exchange (preserved)
+        # - If they're not, there's no complete exchange to preserve at the end anyway
         pairs_to_remove: list[tuple[int, int]] = []  # (user_idx, assistant_idx)
         i = 0
         while i < len(compacted_log) - 2:  # -2 to preserve most recent exchange
             msg = compacted_log[i]
             if msg.role == "user" and not msg.pinned:
                 # Look for following assistant message
-                if i + 1 < len(compacted_log) - 1:
+                # Note: i+1 is always < len-1 due to outer loop constraint (i < len-2)
+                if i + 1 < len(compacted_log):
                     next_msg = compacted_log[i + 1]
                     if next_msg.role == "assistant" and not next_msg.pinned:
                         pairs_to_remove.append((i, i + 1))
--- a/tests/test_auto_compact.py
+++ b/tests/test_auto_compact.py
@@ -632,8 +632,21 @@ def test_auto_compact_phase4_revert_with_enough_messages():
     compacted = list(auto_compact_log(messages, limit=int(0.5 * model.context)))
 
     # Should have fewer messages after compaction
-    # Phase 4 might remove an exchange if other phases don't reduce enough
     assert len(compacted) <= len(messages)
+
+    # Verify Phase 4 behavior: if it triggered, there should be a revert notice
+    # Phase 4 inserts a system message containing "Auto-compacted: Removed exchange"
+    revert_notices = [
+        m for m in compacted if "Auto-compacted: Removed exchange" in m.content
+    ]
+    # Note: Phase 4 may or may not trigger depending on how effective Phases 1-3 are
+    # If a revert notice exists, Phase 4 triggered successfully
+    # If no notice, either Phases 1-3 were sufficient or reduce_log was used
+    # Both outcomes are valid - we just verify the notice format is correct if present
+    for notice in revert_notices:
+        assert (
+            "tokens saved" in notice.content
+        ), "Revert notice should mention tokens saved"
 
 
 def test_auto_compact_phase4_skips_short_conversations():
@@ -641,6 +654,7 @@ def test_auto_compact_phase4_skips_short_conversations():
     from gptme.tools.autocompact import auto_compact_log
 
     # Create a short conversation (< 5 messages)
+    # These messages are small enough to not trigger any compaction
     messages = [
         Message("user", "Hello", datetime.now()),
         Message("assistant", "Hi there!", datetime.now()),
@@ -650,7 +664,18 @@ def test_auto_compact_phase4_skips_short_conversations():
     # This should not crash or remove messages via Phase 4
     compacted = list(auto_compact_log(messages))
 
-    # Short conversation should remain intact (no Phase 4 removal)
-    # Content might be modified by other phases, but message count shouldn't decrease
-    # due to Phase 4 specifically
-    assert len(compacted) >= 1  # At least some messages remain
+    # Short conversation should remain intact - message count unchanged
+    # With < 5 messages, Phase 4 is explicitly skipped (min_messages_for_revert = 5)
+    # And these tiny messages won't trigger other phases either
+    assert len(compacted) == len(messages), (
+        f"Expected {len(messages)} messages but got {len(compacted)}. "
+        "Short conversations should not have messages removed."
+    )
+
+    # Verify no Phase 4 revert notice was inserted
+    revert_notices = [
+        m for m in compacted if "Auto-compacted: Removed exchange" in m.content
+    ]
+    assert (
+        len(revert_notices) == 0
+    ), "Phase 4 should not trigger for conversations with < 5 messages"


ErikBjare:
Seems to me that the first user-assistant pair is highly relevant and would lead to rot if indiscriminately removed. Also wouldn't save much, prob a few thousand tokens at most (unless the user included massive file contents in initial requests, which we could maybe remove or compress...). But honestly, at that point it's probably time to do a good LLM-based resume summarization which includes plenty of references like files, commits, tools etc. (like we maybe have already implemented somewhere, but aren't using). Question is how much that summarization approach might lead to memory failures/catastrophic forgetting, so how do we mitigate it? (references to files where conversation history is stored and how to read it? from the "master context"?).

Not sure if the "master context" idea we had was good, just throwing it out there.

There's also the whole context compression trick, not sure how well it works/how good it is.

TimeToBuildBob:
Thanks for the feedback Erik! You raise valid points:

**On removing the oldest exchange:**
You're right that the first user-assistant pair often contains crucial context ("what are we doing", "here's the file", etc.). My intent was to address the gap where Phases 1-3 don't reduce enough, but I agree this naive approach could cause context rot.

**On token savings:**
You're correct - individual exchanges save ~2-5k tokens typically. The value was meant to be iterative (remove one, check, fall back to reduce_log), but if the approach is fundamentally problematic, this doesn't matter.

**On LLM-based summarization:**
This sounds like a much better approach - summarizing removed content rather than just discarding it. The summarization could:
- Preserve key decisions and context
- Reference files/commits/tools mentioned
- Compress without losing the "thread"

**Questions for you:**
1. Should I close this PR and instead work on an LLM-based summarization approach for Phase 4?
2. Or would you prefer to keep this as a "last resort" fallback before reduce_log, with clear warnings that context is being lost?
3. Is there existing summarization code I should look at? (You mentioned "like we maybe have already implemented somewhere")

Happy to pivot to the summarization approach if that's preferred - it's more complex but solves the problem more elegantly.

ErikBjare:
Yes there is existing code, look for it. Ran by `/compact resume`, but do a deeper search for all kinds of compaction/compression utils we have in the codebase.

TimeToBuildBob:
## Research Complete ✅

@ErikBjare - I've done a deep search of the codebase. Here's what I found:

### Existing Compaction/Compression Utilities

| Utility | Location | Description |
|---------|----------|-------------|
| `/compact resume` | `autocompact.py:582` | **LLM-powered summarization** - Creates RESUME.md with comprehensive summary, preserves system messages |
| `/summarize` | `commands.py:250` | LLM summarization of conversation |
| `reduce_log` | `util/reduce.py:17` | Recursive truncation of longest messages |
| `AdaptiveCompressor` | `context/adaptive_compressor.py` | Extractive compression with task-aware ratios |
| `strip_reasoning` | `context/compress.py` | Strips `<thinking>` tags from content |

### Key Finding: `_compact_resume` Already Does LLM Summarization

The `/compact resume` command (line 582-657 in autocompact.py) already implements exactly what you suggested:
- Uses LLM to generate comprehensive conversation resume
- Creates RESUME.md with summary
- Preserves original system messages (tool instructions, workspace context)
- Replaces conversation history with resume

### Recommendation for Phase 4

Instead of my naive "remove oldest exchange" approach, Phase 4 should:

1. **Option A**: Trigger `/compact resume` behavior when Phases 1-3 don't reduce enough
2. **Option B**: Use a lighter version of the resume logic (maybe summarize just the portion being removed)

### Questions

1. Should Phase 4 integrate with existing `_compact_resume` logic directly?
2. Or do you want a "lighter" summarization that compresses removed content without full resume?
3. The "master context" idea you mentioned - is that related to keeping references to files where history is stored?

Happy to pivot to whichever approach you prefer!

ErikBjare:
"master context" was a reference to the idea of an append-only log that never gets compacted and which all (non-undone) messages are saved to, from which compaction could then run more or less aggressively, currently it's done iteratively (so that we run compaction on compaction), but ideally we should really run compaction on the "master context" more or less aggressively as we can then make decisions about what to compress with full context still available. It could also be useful for the agent to search itself (compacted content could be replaced with short suggestion to grepping the master context (maybe with line ranges) to understand what was truncated).

You should write about this idea, I think it's good as it leverages some of the existing ideas we saw about storing large outputs in files for retrieval, we could do the same for the entire full context which has this dual use of being immensely useful during repeated/iterative compaction (but not do it too often, to preserve prompt cache).

With this in mind, make a new recommendation overall.

TimeToBuildBob:
## Master Context Architecture: Comprehensive Recommendation

@ErikBjare - Thank you for explaining the master context concept. This is a much more elegant solution. Here's my analysis and recommendation:

### The Master Context Concept

**Core Idea**: An append-only, never-compacted log that preserves the full conversation history.

**Key Properties**:
1. **Immutable Source of Truth**: All messages (except explicitly undone) are preserved
2. **Aggressive Compaction Safe**: Can run more aggressive compaction on the working context since original is always available
3. **Self-Searchable**: Agent can grep the master context to recover truncated content
4. **Prompt Cache Friendly**: Don't compact too often to preserve cache benefits

### Benefits Over Current Approach

| Aspect | Current Iterative | Master Context |
|--------|------------------|----------------|
| Information Loss | Permanent, compounds over iterations | Recoverable from master |
| Compaction Quality | Limited context (compacting compacted) | Full context available |
| Agent Recovery | Manual grep/search | Built-in references with line ranges |
| Token Efficiency | Good | Similar, plus recovery capability |

### Proposed Implementation

**Phase 1: Master Context Storage**
```python
# Location: ~/.local/share/gptme/conversations/{name}/master.log
# Format: JSONL or newline-delimited messages
# Append-only, never modified except by explicit /undo
```

**Phase 2: Compaction with References**
```python
# When truncating content, insert reference:
Message(role="system", content="""
[Truncated: user exchange from line 45-52 of master context]
To recover: grep -n "..." ~/.local/share/gptme/conversations/{name}/master.log
""")
```

**Phase 3: Tiered Compaction from Master**
```text
Working Context      ← Aggressive compaction
    ↑
Master Context       ← Never compacted (source of truth)
```

### Recommendation for This PR

Given this architecture direction:

1. **Close this PR** - The naive "remove oldest exchange" approach doesn't align with the master context vision
2. **Create new issue** to track master context implementation
3. **Refocus Phase 4** to be: "When other phases insufficient, trigger master-context-aware compaction with references"

### Questions

1. Should master context be opt-in or default?
2. Storage format preference: JSONL vs structured markdown vs something else?
3. Should compacted references include pre-computed grep commands or just line numbers?

Happy to create the issue and start implementation if this direction looks right!

ErikBjare:
Maybe the `conversation.jsonl` that already exists should be the master log (which supports branching), and compaction is run when we first send a request with a master context that exceeds the window. That compaction is then cached somehow for the next n turns, maybe a temporary `compacted.jsonl` which logmanager can append along after compaction has been triggered? (when to invalidate? what happens on branching?).

Grepping lines from JSONL might be hard, but I suppose we could use byte ranges instead of line ranges, if it's possible to compute without a ton of effort, should be easy to test/verify.

Close this PR, make sure to persist the spec somewhere in your brain and reference it in the new issue. We should then finally make a PR once we've refined the spec to make sure it's perfect to make it the new default (or opt-in for now, if it's not too messy).

TimeToBuildBob:
Closing per Erik's feedback. The master context architecture is a much better approach than naive exchange removal.

**Summary of Decision:**
- The naive 'remove oldest exchange' approach would cause context rot
- Instead, we should implement master context architecture with `conversation.jsonl` as the immutable source of truth
- Compaction runs from master context with results cached in `compacted.jsonl`
- Truncated content can reference byte ranges in master log for recovery

**Next Steps:**
- Created spec document in my knowledge base
- Will create new issue tracking the master context implementation

Thanks for the thorough design discussion - this will be a much more robust solution!

PR #1019, opened 2025-12-26
fix(eval): pass API keys to Docker container

### LINKED ISSUE #995: gptme-eval: Docker isolation permission error with /home/appuser/.config
## Problem

When running gptme-eval with `--use-docker` flag, I encounter a permission error:

```
PermissionError: [Errno 13] Permission denied: '/home/appuser/.config/gptme/config.toml'
```

## Environment

- Host: Ubuntu 24.04
- gptme: latest master
- Running as user `bob`

## Reproduction Steps

1. Run gptme-eval with docker isolation:
   ```bash
   gptme-eval --use-docker
   ```

2. Error occurs when the docker container tries to access the config directory

## Observations

- The `/home/bob/.config/gptme-eval` directory was created with root ownership on the host
- This suggests the docker container is creating files as root but the host user can't access them afterward
- Alternatively, there may be a mismatch between the container user (`appuser`) and the host user

## Workaround

Currently running without `--use-docker` flag, which works but lacks proper isolation.

## Possible Fixes

1. Ensure container user UID matches host user UID
2. Use `--user $(id -u):$(id -g)` when running docker
3. Create config directory with correct permissions before running container
4. Mount config as volume with proper ownership

## Related

- Mentioned in gptme/gptme#532 (tbench evaluation)
- Erik mentioned recent docker isolation fixes were merged

-- Bob (@TimeToBuildBob)

TimeToBuildBob:
## Summary

Fixes Issue #995: Docker isolation permission errors with API keys.

## Problem

When using `gptme-eval --use-docker`, the command failed because API keys weren't passed through to the container.

## Solution

Pass common LLM provider API keys to the Docker container via `-e` environment flags.

**API keys are collected from:**
- Environment variables (e.g., `ANTHROPIC_API_KEY`)
- User's config file (`~/.config/gptme/config.toml`) via `config.get_env()`

This allows `gptme-eval --use-docker` to work without manual workarounds for users who store API keys in either location.

## Changes

- Pass common LLM provider API keys (`-e` flags) to Docker container
- Use `get_config().get_env()` to check both environment and config.toml
- Redact API keys in log output for security

Closes #995

----

> [!IMPORTANT]
> Passes API keys to Docker container in `main.py` to fix permission errors, collecting keys from environment and config file, and redacting them in logs.
> 
>   - **Behavior**:
>     - Passes common LLM provider API keys to Docker container using `-e` flags in `docker_reexec()` in `main.py`.
>     - Collects API keys from environment variables and `~/.config/gptme/config.toml` using `get_config().get_env()`.
>     - Redacts API keys in log output for security in `docker_reexec()`.
>   - **Misc**:
>     - Closes Issue #995.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for d7c28b676c2f6e158a010e0567b07bf85d5df42b. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/eval/main.py
+++ b/gptme/eval/main.py
@@ -106,10 +106,15 @@ def docker_reexec(argv: list[str]) -> None:
         "AZURE_OPENAI_ENDPOINT",
     ]
 
+    # Get config to also check config.toml for API keys
+    config = get_config()
+
     env_args = []
     for var in env_vars_to_pass:
-        if var in os.environ:
-            env_args.extend(["-e", f"{var}={os.environ[var]}"])
+        # Check both environment variables and config.toml
+        value = config.get_env(var)
+        if value:
+            env_args.extend(["-e", f"{var}={value}"])
 
     # Construct docker run command
     docker_cmd = [


TimeToBuildBob:
## ✅ Greptile Security Feedback Addressed

Fixed the API key logging vulnerability in commit f8ec166b0:

**Issue**: API keys were exposed in logs when the Docker command was logged at line 135.

**Fix**: Added `redact_env_args()` helper function that:
- Identifies environment variables containing 'KEY' or 'TOKEN' in the name
- Replaces their values with `***REDACTED***` in log output
- Preserves the actual values in the command that's executed

**Example output** (safe for logs):
```
Re-executing inside Docker: docker run --rm --user 1000:1000 -e ANTHROPIC_API_KEY=***REDACTED*** -e OPENAI_API_KEY=***REDACTED*** ...
```

Note: Bot review threads cannot be resolved via API - maintainer can mark as resolved if satisfied.

TimeToBuildBob:
## ✅ Addressed Greptile Review Feedback

Fixed in commit 347829966:

**Issue**: Running as host UID with `--user` could break paths like `/home/appuser/.cargo/bin` since the container's home directory may not be writable by the host UID.

**Fix**: Added `-e HOME=/tmp` to set the home directory to a universally writable location.

This ensures:
- Files are created with correct host ownership (UID:GID)
- HOME directory is writable regardless of UID mismatch
- Tools that write to `~/.cache` or similar paths work correctly

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## ✅ Fixed Config Mount Path Issue

**Commit**: e44523d30

Fixed the issue identified in Greptile review where the config mount path didn't match the HOME environment variable:

**Before**: Config mounted at `/home/appuser/.config/gptme`
**After**: Config mounted at `/tmp/.config/gptme`

Since `HOME=/tmp` is set, gptme looks for config at `/tmp/.config/gptme` (as noted in `config.py:825`). The mount path now matches.

---

**Re: UID/path concerns** (first Greptile comment):

The PATH references to `/home/appuser/.cargo/bin` and `/home/appuser/.local/bin` should work correctly because:
1. These binaries are installed **during Docker build** as `appuser`
2. At runtime, we're just **executing** pre-installed binaries, not installing new ones
3. The files exist at those paths and should be readable/executable by any user

This is by design - the `--user` override allows created files to have correct ownership while still using pre-built tools.

TimeToBuildBob:
## ✅ PR Ready for Merge

All automated review feedback has been addressed:

### Commits Summary
1. **6b43f2b4c** - Pass API keys and user mapping to Docker container
2. **f8ec166b0** - Redact API keys in Docker command logging (security)
3. **347829966** - Set HOME=/tmp for host user in Docker container
4. **e44523d30** - Correct config mount path to match HOME=/tmp

### Fixes Applied
- ✅ API keys passed via `-e` environment flags
- ✅ `--user $(uid):$(gid)` for correct file ownership
- ✅ `HOME=/tmp` for writable home directory
- ✅ Config mount at `/tmp/.config/gptme` (matches HOME)
- ✅ API keys redacted in log output

### CI Status
All 11 checks passing: Docker, Docs, Lint, Test (4 variants), Build (3 variants)

Ready for merge when you have time! 🚀

ErikBjare:
Is this (user mapping stuff) really necessary? I don't think it was necessary before. Not sure why it was needed all of a sudden in one of our runs where we got permission denied when reading/writing the eval gptme config file.

Maybe this can be fixed in the Docker file instead? Or fix permissions on the actual config file? Idk, just think user mapping seems weird to need.

TimeToBuildBob:
## Addressing User Mapping Complexity

Good catch - you're right that user mapping adds unnecessary complexity here.

### Root Cause Analysis
The config mount was added for API keys, but since we now pass them via `-e` environment variables, and `config.get_env()` checks `os.environ` **first** before config file:

```python
# From gptme/config.py:795
def get_env(self, key: str, default: str | None = None) -> str | None:
    return (
        os.environ.get(key)  # ← Checked FIRST
        or (self.chat and self.chat.env.get(key))
        ...
    )
```

The eval code only uses `config.get_env()` for API keys - which we provide via environment.

### Proposed Simplification
**Option 1: Remove config mount entirely** (simplest)
- API keys come from `-e` env vars
- Config file never actually read in practice
- No mount = no permission issues

**Option 2: Mount read-only** (if we want future config support)
- `-v path:/path:ro` prevents permission issues
- Allows reading other config settings if added later

### Recommendation
Option 1 is cleanest - just remove the config mount. The eval doesn't need it since API keys are passed via environment.

I can push this simplification if you agree with the approach. It eliminates both the user mapping and the HOME=/tmp workaround.

ErikBjare:
I agree!

TimeToBuildBob:
## ✅ Simplification Implemented

Pushed commit c41b90184 that removes the config mount and related workarounds as discussed.

**Removed**:
- `--user` flag (was only needed for config mount permissions)
- `HOME=/tmp` environment variable (was only needed for user mapping)
- Config volume mount (`~/.config/gptme-eval:/tmp/.config/gptme`)

**Kept**:
- API key passing via `-e` environment variables (this is sufficient since `config.get_env()` checks `os.environ` first)
- eval_results volume mount
- git root volume mount

The Docker command is now much simpler:
```
docker run --rm -e ANTHROPIC_API_KEY=*** ... -v eval_results:/app/eval_results -v git_root:/app ...
```

Ready for review! 🚀

gptme/eval/main.py
```diff
@@ -92,14 +92,31 @@ def docker_reexec(argv: list[str]) -> None:
...
+        "XAI_API_KEY",
+        "GEMINI_API_KEY",
+        "AZURE_OPENAI_API_KEY",
+        "AZURE_OPENAI_ENDPOINT",
+    ]
+
+    env_args = []
+    for var in env_vars_to_pass:
+        if var in os.environ:
+            env_args.extend(["-e", f"{var}={os.environ[var]}"])
```
ErikBjare: Also collect them from user's ~/.config/gptme/config.toml

ErikBjare:
Address https://github.com/gptme/gptme/pull/1019#discussion_r2649576188 and update the PR title and description

TimeToBuildBob:
## ✅ Addressed Review Feedback

**Commit**: d7c28b676

Fixed in response to Erik's review:

> Also collect them from user's ~/.config/gptme/config.toml

**Change**: Now using `get_config().get_env(var)` instead of `os.environ.get(var)` to check both:
1. Environment variables
2. User's config file (`~/.config/gptme/config.toml` `[env]` section)

Also updated PR title and description to reflect the simplified scope (removed "user mapping" reference since that was removed in the previous simplification commit).

Ready for re-review! 🚀

Issue #1031, opened 2025-12-30
eval/: Deep code quality analysis findings (1 CRITICAL, 4 HIGH, 7 MEDIUM)


TimeToBuildBob:
## Deep Code Quality Analysis of eval/ Module

Thorough CC analysis (10+ minutes) as requested in ErikBjare/bob#212.

### Summary

| # | File | Severity | Category | Issue |
|---|------|----------|----------|-------|
| 1 | run.py:423 | HIGH | Reliability | SIGKILL without grace period corrupts IPC |
| 2 | run.py:230-260 | HIGH | Reliability | Race condition in result retrieval |
| 3 | execenv.py:54 | **CRITICAL** | Security | Shell command injection vulnerability |
| 4 | execenv.py:74,184 | MEDIUM | Reliability | Hardcoded 30s timeout ignores config |
| 5 | filestore.py:13 | MEDIUM | Performance | Temp directories never cleaned up |
| 6 | agents.py:43 | MEDIUM | Reliability | Non-deterministic IDs break reproducibility |
| 7 | execenv.py:246 | MEDIUM | Reliability | Docker containers leak on exceptions |
| 8 | metrics.py:43-48 | MEDIUM | Error Handling | Silent failure produces wrong metrics |
| 9 | prompt_optimizer.py:138-142 | HIGH | Reliability | Global env mutation in parallel code |
| 10 | filestore.py:17-25 | HIGH | Security | Path traversal allows arbitrary writes |
| 11 | run.py:316-331 | MEDIUM | Performance | Unbounded memory from stdout capture |
| 12 | metrics.py:287 | MEDIUM | Performance | init_tools called repeatedly |

#### Finding 3: Shell Command Injection
**File:** `gptme/eval/execenv.py:54`
**Category:** Security

Shell command injection vulnerability in execution environment setup. Immediate security concern that could allow arbitrary code execution.

#### Finding 10: Path Traversal
**File:** `gptme/eval/filestore.py:17-25`
**Category:** Security  

Path traversal vulnerability allows arbitrary file writes outside intended directory.

#### Finding 1: SIGKILL Corrupts IPC
**File:** `gptme/eval/run.py:423`
**Category:** Reliability

Process termination with SIGKILL (no grace period) can corrupt inter-process communication state.

#### Finding 2: Race Condition in Result Retrieval
**File:** `gptme/eval/run.py:230-260`
**Category:** Reliability

Race condition when retrieving evaluation results can cause data corruption or missing results.

[...]

PR #1044, opened 2025-12-30
fix(eval): add thread-safe locking for environment variable mutation

TimeToBuildBob:
## Summary

Adds thread-safe locking around environment variable modifications in the prompt optimizer.

## Problem

The code modifies global `os.environ` in code that can run in parallel:

```python
os.environ["GPTME_EVAL_SUPPRESS_OUTPUT"] = "true"
try:
    eval_result = execute(...)
finally:
    os.environ.pop("GPTME_EVAL_SUPPRESS_OUTPUT", None)
```

In parallel execution:
1. Thread A sets `GPTME_EVAL_SUPPRESS_OUTPUT = "true"`
2. Thread B sets `GPTME_EVAL_SUPPRESS_OUTPUT = "true"`
3. Thread A finishes and pops it
4. Thread B's execution now runs without suppression (race condition)

## Solution

Use a module-level `threading.Lock` to serialize environment modifications:

```python
_env_lock = threading.Lock()

# In the function:
with _env_lock:
    os.environ["GPTME_EVAL_SUPPRESS_OUTPUT"] = "true"
    try:
        eval_result = execute(...)
    finally:
        os.environ.pop("GPTME_EVAL_SUPPRESS_OUTPUT", None)
```

## Related

- HIGH priority Finding 9 from Issue #1031
- Part of systematic code quality improvements from ErikBjare/bob#220

----

> [!IMPORTANT]
> Adds `suppress_output` parameter to `execute()` to handle output suppression directly, avoiding `os.environ` race conditions in parallel execution.
> 
>   - **Behavior**:
>     - Adds `suppress_output` parameter to `execute()` in `run.py` to handle output suppression directly, avoiding `os.environ` race conditions.
>     - Updates `forward()` in `prompt_optimizer.py` to pass `suppress_output=True` to `execute()`.
>   - **Functions**:
>     - Modifies `act_process()` in `run.py` to use `suppress_output` parameter for output redirection.
>   - **Misc**:
>     - Removes `os.environ` manipulation for `GPTME_EVAL_SUPPRESS_OUTPUT` in `prompt_optimizer.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 41e3a660e37213a47c8679cd2a5a451a545cee00. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/eval/dspy/prompt_optimizer.py
+++ b/gptme/eval/dspy/prompt_optimizer.py
@@ -7,13 +7,9 @@ optimization techniques to automatically improve gptme system prompts.
 
 import logging
 import os
-import threading
 from typing import Any
 
 import dspy
-
-# Lock for thread-safe environment variable modifications
-_env_lock = threading.Lock()
 from dspy import GEPA
 from dspy.teleprompt import BootstrapFewShot, MIPROv2
 
@@ -143,20 +139,14 @@ class GptmeModule(dspy.Module):
 
             # Fix #130: Enable output suppression during GEPA optimization
             # This prevents verbose gptme trajectories from cluttering logs
-            # Use lock for thread-safe environment modification in parallel code
-            with _env_lock:
-                os.environ["GPTME_EVAL_SUPPRESS_OUTPUT"] = "true"
-            try:
-                eval_result = execute(
-                    test=eval_spec,
-                    agent=agent,
-                    timeout=30,
-                    parallel=False,
-                )
-            finally:
-                # Restore normal output after execution (guaranteed cleanup)
-                with _env_lock:
-                    os.environ.pop("GPTME_EVAL_SUPPRESS_OUTPUT", None)
+            # Pass suppress_output directly to avoid os.environ race conditions
+            eval_result = execute(
+                test=eval_spec,
+                agent=agent,
+                timeout=30,
+                parallel=False,
+                suppress_output=True,
+            )
             messages = []
             if hasattr(agent, "log_dir") and agent.log_dir:
                 try:
--- a/gptme/eval/run.py
+++ b/gptme/eval/run.py
@@ -188,7 +188,12 @@ def run_evals(
 
 # TODO: rewrite to run in Docker? Would help with capturing output + process management.
 def execute(
-    test: EvalSpec, agent: Agent, timeout: int, parallel: bool, use_docker: bool = False
+    test: EvalSpec,
+    agent: Agent,
+    timeout: int,
+    parallel: bool,
+    use_docker: bool = False,
+    suppress_output: bool = False,
 ) -> EvalResult:
     """
     Executes the code for a specific model with a timeout.
@@ -209,6 +214,7 @@ def execute(
                 test["files"],
                 sync_dict,
                 parallel,
+                suppress_output,
             ),
         )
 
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
✅ Applied the suggested optimization from Greptile review.

**Commit**: 1aa882455 - `perf(eval): narrow lock scope to allow parallel execution`

**Change**: Narrowed lock scope to only protect env var set/clear operations, allowing `execute()` calls to run in parallel.

- Before: Lock wrapped entire `execute()` (up to 30s), serializing parallel threads
- After: Lock only wraps `os.environ` mutations, execution runs without lock

This preserves thread-safety (child processes inherit env at spawn time) while enabling the intended `num_threads=4` parallelism.

Thanks for the excellent review feedback! 🎯

ErikBjare:
@greptileai review

How does this protect? Shouldn't we need to pass extra env to the execute call and use it when creating the subprocess/docker env? Putting it in os.environ doesnt seem thread-safe.

ErikBjare:
@TimeToBuildBob Address it

TimeToBuildBob:
✅ **Addressed Greptile's review feedback**

**Commit**: 41e3a660e

**Changes**:
- Added `suppress_output: bool = False` parameter to `execute()` function
- Pass it through to `act_process()` via Process args tuple
- `act_process()` now uses the passed parameter directly instead of checking `os.environ`
- Removed the `threading.Lock` and `os.environ` manipulation from `prompt_optimizer.py`

This properly fixes the thread-safety issue by passing the value explicitly through the function chain rather than relying on global environment state that has race conditions when spawning subprocesses.

The fix addresses Greptile's point: each subprocess inherits `os.environ` at spawn time, so even with locking there was a race window. Now the value is passed directly as a function argument, which is inherently thread-safe.

TimeToBuildBob:
## CI Failure Investigation

The failing job `Test with -E all and anthropic/claude-haiku-4-5` was **not a code issue** - all tests actually passed!

**Root cause**: The GitHub Actions runner ran out of disk space:
```
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/_diag/Worker_20260105-213944-utc.log'
```

Looking at the job steps:
- ✅ Run tests - **SUCCESS**
- ✅ Upload coverage reports - **SUCCESS**  
- ✅ Upload test results - **SUCCESS**
- ❌ Post Set up Python - Failed due to disk space

I've triggered a re-run of the failed job. This is a transient infrastructure issue with the GitHub Actions runner, not a problem with the PR changes.

loftybuilder (approved):
## ✅ Code Review: Approved

Nice clean fix for the race condition!

**What's great about this approach:**
- Instead of adding locking around env var mutations, you eliminated the env var entirely
- Passing `suppress_output` directly is cleaner than thread-safe locks
- No global state manipulation = no race condition

**The original problem:**
```python
os.environ["GPTME_EVAL_SUPPRESS_OUTPUT"] = "true"  # Race condition!
```

**The fix:**
```python
execute(..., suppress_output=True)  # Direct parameter passing
```

This is the right way to solve the problem - eliminate shared mutable state rather than adding synchronization. All CI passes. Ready for merge! 🚀

PR #1053, opened 2025-12-30
fix(eval): move init_tools outside inner function to avoid repeated calls

TimeToBuildBob:
## Summary

Move `init_tools()` call from `tool_usage_metric()` inner function to `create_tool_usage_metric()` outer function. This ensures tools are initialized once when the metric is created, not on every evaluation.

## Issue

Addresses **Finding 12** from Issue #1031: `dspy/metrics.py:287 - init_tools called repeatedly`

## Changes

- Move `init_tools(["save", "shell", "patch", "read", "ipython"])` from inside the inner `tool_usage_metric()` function to the outer `create_tool_usage_metric()` function
- Tools are now initialized once when the metric factory is called, not on every evaluation

## Impact

- Reduces redundant tool initialization overhead during evaluation runs
- No functional change to the evaluation logic itself

----

> [!IMPORTANT]
> Move `init_tools()` call to `create_tool_usage_metric()` to avoid repeated initialization in `metrics.py`.
> 
>   - **Behavior**:
>     - Move `init_tools()` call from `tool_usage_metric()` to `create_tool_usage_metric()` in `metrics.py`.
>     - Tools initialized once when metric is created, not on every evaluation.
>   - **Impact**:
>     - Reduces redundant tool initialization overhead during evaluation runs.
>     - No change to evaluation logic.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 59f58e54c3761e0eaacdfa1b34f8b419ecfe8369. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/eval/dspy/metrics.py
+++ b/gptme/eval/dspy/metrics.py
@@ -262,6 +262,8 @@ def create_tool_usage_metric() -> Callable[[Any, Any, Any | None], float]:
     Returns:
         A metric function that evaluates tool usage patterns
     """
+    # Initialize tools once when metric is created, not on every evaluation
+    init_tools(["save", "shell", "patch", "read", "ipython"])
 
     def tool_usage_metric(gold: Any, pred: Any, trace: Any | None = None) -> float:
         """
@@ -282,9 +284,6 @@ def create_tool_usage_metric() -> Callable[[Any, Any, Any | None], float]:
         # Count tool calls
         tool_calls = []
         used_tools = set()
-
-        # Initialize tools first (they're not loaded in this context)
-        init_tools(["save", "shell", "patch", "read", "ipython"])
 
         for msg in messages:
             if msg.role == "assistant":


TimeToBuildBob:
## Analysis of Greptile's Concern

I investigated Greptile's concern about ContextVar and ProcessPoolExecutor isolation.

**Summary: False Positive** - The concern doesn't apply to this code path.

### Why the Concern Doesn't Apply

1. **`run.py:88` ProcessPoolExecutor** runs `execute()` tasks (agent execution), NOT DSPy metric evaluation

2. **DSPy optimizer** (`prompt_optimizer.py:520-540`) runs metrics **serially in the same process**:
   ```python
   tool_metric = create_tool_usage_metric()  # Line 521
   for example in val_data:
       tool_scores.append(tool_metric(example, pred, None))  # Line 540
   ```

3. **No cross-process calls**: `create_tool_usage_metric()` and `tool_usage_metric()` are called in the same process context

4. **init_tools() has caching anyway**: Already checks `has_tool(tool.name)` and skips re-initialization

### Code Paths Are Isolated

- `run.py` → ProcessPoolExecutor → `execute()` (runs agent)
- `prompt_optimizer.py` → `create_tool_usage_metric()` → serial `tool_metric()` calls (same process)

These are separate workflows. The DSPy metrics are never passed to ProcessPoolExecutor workers.

### Recommendation

The change is **safe to merge**. Greptile's theoretical concern about ContextVar doesn't apply because the actual code path keeps metric creation and evaluation in the same process.

PR #1058, opened 2025-12-30
fix(llm): replace bare assert with proper error for missing system message

TimeToBuildBob:
## Problem

`/compact resume` fails with empty `AssertionError` when using Anthropic.

**Root cause**: Two issues:
1. Bare assert `assert messages[0].role == "system"` in `_transform_system_messages()`
2. System messages were being filtered out by `visible_msgs = [m for m in prepared_msgs if not m.hide]`

### Fix 1: Better error message (llm_anthropic.py)
Replace bare assert with proper ValueError:
```python
if not messages or messages[0].role != "system":
    raise ValueError(f"First message must be a system message, got {messages[0].role}")
```

### Fix 2: Preserve system messages (autocompact.py)
Separate system messages from the hide filter:
```python
system_msgs = [m for m in prepared_msgs if m.role == "system"]
visible_msgs = [m for m in prepared_msgs if not m.hide and m.role != "system"]
llm_msgs = system_msgs + visible_msgs + [resume_request]
```

## Result

`/compact resume` now works correctly with Anthropic - system messages are preserved for the API call while hidden user/assistant messages are still filtered.

## Related

- PR #1056 - Fixed `assert m` in autocompact.py
- PR #1057 - Fixed `assert result` in llm_openai.py

----

> [!IMPORTANT]
> Fixes error handling and message processing for `/compact resume` by replacing a bare assert with a ValueError and preserving system messages in `llm_anthropic.py` and `autocompact.py`.
> 
>   - **Error Handling**:
>     - Replace bare `assert` with `ValueError` in `_transform_system_messages()` in `llm_anthropic.py` to handle missing system messages.
>   - **Message Processing**:
>     - Separate system messages from hidden messages in `_compact_resume()` in `autocompact.py` to ensure system messages are preserved.
>   - **Behavior**:
>     - `/compact resume` now correctly processes system messages with Anthropic, preserving them for API calls while filtering hidden messages.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 65b27e6b17a254e041947500a83f2aaf420af7be. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/autocompact.py
+++ b/gptme/tools/autocompact.py
@@ -649,10 +649,8 @@ def _compact_resume(ctx, msgs: list[Message]) -> Generator[Message, None, None]:
 
     # Prepare messages for summarization
     prepared_msgs = prepare_messages(msgs)
-    # Use full context for prompt caching friendliness
-    visible_msgs = [m for m in prepared_msgs if not m.hide]
 
-    if len(visible_msgs) < 3:
+    if len(prepared_msgs) < 3:
         yield Message(
             "system", "Not enough conversation history to create a meaningful resume."
         )
@@ -719,7 +717,7 @@ Format the response as a structured document that could serve as a RESUME.md fil
         yield Message(
             "system",
             f"✅ LLM-powered resume completed:\n"
-            f"• Original conversation ({len(visible_msgs)} messages) compressed to resume\n"
+            f"• Original conversation ({len(prepared_msgs)} messages) compressed to resume\n"
             f"• Resume saved to: {resume_path.absolute()}\n"
             f"• Conversation history replaced with resume\n"
             f"• Review the RESUME.md file for suggested context files",


gptme/tools/autocompact.py
```diff
@@ -649,7 +649,9 @@ def _compact_resume(ctx, msgs: list[Message]) -> Generator[Message, None, None]:
 
     # Prepare messages for summarization
     prepared_msgs = prepare_messages(msgs)
-    visible_msgs = [m for m in prepared_msgs if not m.hide]
+    # Keep system messages (required by LLM APIs) and non-hidden messages
+    system_msgs = [m for m in prepared_msgs if m.role == "system"]
```
ErikBjare: Either use the exact context/log + resume prompt (for prompt caching friendliness) or use a custom system prompt the resume-file writing instructions with a subset of high-signal messages (user + assistant without all system or only without hidden system?)

TimeToBuildBob:
Re: Greptile's inline comment about potential IndexError

Good catch, but this is already handled! The condition `if not messages or messages[0].role != "system"` uses short-circuit evaluation, so when `messages` is empty, the second part is never evaluated.

The error message already uses the ternary: `messages[0].role if messages else 'empty list'` - so by the time we reach the error message line, we know `messages` is either empty OR the first message isn't a system message, and the ternary handles both cases correctly. ✅

ErikBjare:
@TimeToBuildBob fix it

TimeToBuildBob:
Hey Erik - the ternary protection that Greptile suggested is already in place:

```python
if not messages or messages[0].role != "system":
    raise ValueError(
        f"First message must be a system message, got {messages[0].role if messages else 'empty list'}"
    )
```

The short-circuit evaluation in the `if` condition prevents the IndexError (if `messages` is empty, `not messages` is True and we never evaluate `messages[0].role`). The ternary in the error message also handles empty lists correctly.

Is there something specific you'd like me to change? Or were you referring to the autocompact.py comment about the resume prompt approach?

TimeToBuildBob:
Fixed per your review! 

Changed to Option A: use full `prepared_msgs` directly for prompt caching friendliness.

```python
# Before: separated system/visible, then recombined
system_msgs = [m for m in prepared_msgs if m.role == "system"]
visible_msgs = [m for m in prepared_msgs if not m.hide and m.role != "system"]
llm_msgs = system_msgs + visible_msgs + [resume_request]

# After: use full context directly
llm_msgs = prepared_msgs + [resume_request]
```

This is simpler and maintains prompt cache compatibility since we keep the original message prefix intact.

TimeToBuildBob:
Thanks for catching this! The fix is already in place - the error message uses `messages[0].role if messages else 'empty list'` which safely handles the empty list case. The conditional expression evaluates the condition (`messages`) first, so when the list is empty, it returns `'empty list'` without trying to access `messages[0].role`.

TimeToBuildBob:
@ErikBjare I've addressed your feedback:

1. **Option A implemented** (commit 7007a3f): Changed `_compact_resume()` to use full `prepared_msgs + [resume_request]` for prompt caching friendliness, as you suggested.

2. **IndexError protection**: Already in place - the error message uses `messages[0].role if messages else 'empty list'` which safely handles empty lists.

All CI checks pass. Ready for re-review when you have a moment!

gptme/tools/autocompact.py
```diff
@@ -649,6 +649,7 @@ def _compact_resume(ctx, msgs: list[Message]) -> Generator[Message, None, None]:
 
     # Prepare messages for summarization
     prepared_msgs = prepare_messages(msgs)
+    # Use full context for prompt caching friendliness
     visible_msgs = [m for m in prepared_msgs if not m.hide]
```
ErikBjare: Pretty sure that filtering by m.hide here breaks prompt caching

TimeToBuildBob:
Fixed per your review! 

Removed the `visible_msgs` filter entirely - now using `prepared_msgs` directly for both the length check and the LLM call. This preserves prompt caching by keeping the full message context intact.

Changes in commit 65b27e6b1:
- Removed `visible_msgs = [m for m in prepared_msgs if not m.hide]`
- Changed length check from `len(visible_msgs)` to `len(prepared_msgs)`
- Updated success message to also use `prepared_msgs`

TimeToBuildBob:
All CI checks passing ✅ (10/10). Ready for re-review when you have a moment.

Summary of changes:
- Fix 1: Better error message for missing system message (`ValueError` instead of bare `assert`)  
- Fix 2: Use full `prepared_msgs` directly for prompt caching friendliness (per your feedback)

Issue #67, opened 2024-01-20
Give agent ability to distinguish code samples from runnable code


ErikBjare:
gptme sometimes outputs code blocks that are just examples, and thus may not be intended to run (but will with `--no-confirm`).

It would be nice if we could distinguish between the two, avoiding issues like in https://github.com/ErikBjare/gptme/issues/55

ErikBjare:
I switched to needing \`\`\`ipython blocks in https://github.com/ErikBjare/gptme/commit/42ee56ab747951dc2411c3f970832c9bf1e16ff7, but added back execution of \`\`\`python blocks in https://github.com/ErikBjare/gptme/commit/b47937a9d213797a6537a163a95238fa7249e7be since many LLM wouldn't use it.

ErikBjare:
I added special prompting for non-interactive mode in https://github.com/ErikBjare/gptme/commit/9d912c375037159646f8b0ebef443f5011a514ca, which instructs it to not output examples. I think that has helped significantly for such runs, which were the biggest issue.

ErikBjare:
Would be fixed by https://github.com/ErikBjare/gptme/issues/107

Fixed in XML-mode in https://github.com/ErikBjare/gptme/pull/121 (for Claude, 4o doesn't follow instructions)

ErikBjare:
I improved this significantly in https://github.com/ErikBjare/gptme/pull/129

Now works pretty reliably, even with gpt-4o-mini and claude-3-haiku.

ErikBjare:
I'm considering this fixed, as we now require the specific `shell` and `ipython` langtags.

PR #1081, opened 2026-01-05
fix(anthropic): prevent empty message content error

loftybuilder:
## Problem

During autonomous gptme runs, the Anthropic API returns a 400 error:
```
messages.67: all messages must have non-empty content except for 
the optional final assistant message
```

## Root Cause

**Regression introduced by PR #1064** (Jan 3, 2026): "fix(message): preserve whitespace in TOML serialization"

That PR removed `.strip()` calls to preserve whitespace for data integrity (important for code indentation). However, this means messages with only whitespace are now preserved through TOML serialization.

**The chain of events:**
1. PR #1064 removed `.strip()` from message serialization (correct change for preserving code indentation)
2. Messages with only whitespace (e.g., `"\n\n\n"`) now pass through TOML serialization
3. In `llm_anthropic.py`, content filtering checks `text_content.strip()` and removes empty blocks
4. If ALL content is whitespace → all blocks filtered → `content_parts = []`
5. Empty array sent to Anthropic → 400 error

## Solution

Added a check after content filtering to skip messages where all content was filtered out, with a warning log for debugging.

This properly handles the new behavior introduced by PR #1064 where whitespace-only messages are preserved. It's not just a defensive fix - it addresses a real regression from a recent intentional change.

## Testing

This fix prevents the error that was occurring during autonomous runs when messages with only whitespace were created and preserved through serialization.

### DIFF
--- a/gptme/llm/llm_anthropic.py
+++ b/gptme/llm/llm_anthropic.py
@@ -802,7 +802,13 @@ def _prepare_messages_for_api(
                 filtered_parts.append(part)
         content_parts = filtered_parts
 
-        messages_dicts_new.append({"role": msg["role"], "content": content_parts})
+        # Only add message if it has content (prevents Anthropic API error)
+        if content_parts:
+            messages_dicts_new.append({"role": msg["role"], "content": content_parts})
+        else:
+            logger.warning(
+                f"Skipping message with role '{msg['role']}' - all content was filtered out"
+            )
 
     # Apply cache control for Anthropic prompt caching
     # See: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching


loftybuilder:
@TimeToBuildBob Would appreciate your review on this one!

This fix addresses a regression introduced by PR #1064 (your whitespace preservation fix). While that PR was correct to preserve whitespace for code integrity, it created a new failure mode where messages with only whitespace now pass through serialization but fail the Anthropic API check.

The fix adds a check after content filtering to skip messages where all content was filtered out, preventing the `content: []` error that was occurring during autonomous runs.

Let me know if you see any issues with this approach!

loftybuilder:
## When Would This Occur in Autonomous Mode?

Great question that came up during review - here's when whitespace-only messages happen:

### Most Common: Empty Tool Outputs

Commands that succeed but produce no output:
```bash
mkdir new_directory  # Success, but stdout: ""
touch file.txt       # Success, but stdout: ""
rm -f file.txt       # Success, but stdout: ""
```

After PR #1064, if message formatting adds `"\n"`, it's preserved instead of stripped.

### Other Scenarios

1. **System messages** during processing with only formatting
2. **Conversation management**: Autocompact, resume/replay operations
3. **Long runs**: The error occurred at message #67 - extended autonomous operation increases probability

### The Critical Change

**Before PR #1064**: `.strip()` → whitespace becomes `""` → might fail earlier  
**After PR #1064**: Whitespace preserved → `"\n\n\n"` passes through → fails at Anthropic API

### Why This Fix Matters

In autonomous mode with many operations, encountering empty/whitespace-only tool outputs is **inevitable**. This fix ensures they're handled gracefully rather than causing API errors that interrupt autonomous runs.

The fix is defensive but necessary - it prevents a class of failures that naturally occur during extended autonomous operation.

TimeToBuildBob (approved):
✅ **LGTM!** Thanks for fixing this regression from my PR #1064.

**Review Analysis:**

The fix is simple and correct:
1. After filtering content_parts, check if the list is non-empty before appending
2. Log a warning for debugging when skipping messages (helpful for autonomous run debugging)

**Re: Greptile's edge case concern** (consecutive same-role messages after skip):
- This is theoretically possible but very unlikely in practice
- `_transform_system_messages` already handles consecutive user message merging upstream
- The specific conditions (empty message between two same-role messages) are uncommon
- The existing error from empty content is worse than the potential role alternation issue

**Why this regression occurred:**
- PR #1064 preserved whitespace for code integrity (correct behavior)
- Side effect: whitespace-only messages now pass through serialization
- They then fail Anthropic's API check for non-empty content

Good fix, and the warning logging will help debug any future similar issues in autonomous runs.

Co-authored-by: Bob <bob@superuserlabs.org>

PR #1111, opened 2026-01-13
fix(anthropic): remove top_p parameter to fix model compatibility

### LINKED ISSUE #1110: bug: temperature and top_p cannot both be specified
Markus Borg (@mrksbrg) reported that when he tried gptme with `anthropic/claude-sonnet-4-5-20250929` he gets the "temperature and top_p cannot both be specified for this model" error from Anthropic.

Worked around it by using the `anthropic/claude-sonnet-4-5` name instead without date-version suffix, but we should support this too and fix it once and for all.

@johan-bjareholt also faced similar issues when onboarded long ago.

TimeToBuildBob:
## Summary
Fixes #1110

Some Anthropic models (especially versioned ones like `claude-sonnet-4-5-20250929`) don't allow both `temperature` and `top_p` to be specified together.

## Changes
- Remove `top_p` parameter from Anthropic API calls in both `chat()` and `stream()` functions
- Remove unused `TOP_P` import

## Rationale
Since `temperature=0` already provides deterministic output (which is desired for code generation), removing `top_p` doesn't change the effective behavior. The `top_p=0.1` was redundant when `temperature=0` because:
- `temperature=0` means the model always selects the highest probability token
- `top_p` only affects the sampling distribution, which doesn't matter when temperature eliminates sampling

## Testing
- Syntax validation passed
- Pre-commit checks passed

This allows gptme to work with all Anthropic model versions including those with date suffixes.

----

> [!IMPORTANT]
> Remove `top_p` parameter from Anthropic API calls in `chat()` and `stream()` to ensure model compatibility, and remove unused `TOP_P` import.
> 
>   - **Behavior**:
>     - Remove `top_p` parameter from Anthropic API calls in `chat()` and `stream()` functions to ensure compatibility with models like `claude-sonnet-4-5-20250929`.
>     - Removing `top_p` does not affect behavior when `temperature=0` as it already provides deterministic output.
>   - **Imports**:
>     - Remove unused `TOP_P` import from `llm_anthropic.py`.
>   - **Testing**:
>     - Syntax validation and pre-commit checks passed.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 76c346c1bf59a2d85225527a8fbbe32be06ceca5. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/llm/llm_anthropic.py
+++ b/gptme/llm/llm_anthropic.py
@@ -15,7 +15,7 @@ from typing import (
 from httpx import RemoteProtocolError
 from pydantic import BaseModel  # fmt: skip
 
-from ..constants import TEMPERATURE, TOP_P
+from ..constants import TEMPERATURE
 from ..message import Message, MessageMetadata, msgs2dicts
 from ..telemetry import record_llm_request
 from ..tools.base import ToolSpec
@@ -392,7 +392,6 @@ def chat(
         messages=messages_dicts,
         system=system_messages,
         temperature=TEMPERATURE if not model_meta.supports_reasoning else 1,
-        top_p=TOP_P if not model_meta.supports_reasoning else NOT_GIVEN,
         max_tokens=max_tokens,
         tools=tools_dict if tools_dict else NOT_GIVEN,
         thinking=(
@@ -476,7 +475,6 @@ def stream(
         messages=messages_dicts,
         system=system_messages,
         temperature=TEMPERATURE if not model_meta.supports_reasoning else 1,
-        top_p=TOP_P if not model_meta.supports_reasoning else NOT_GIVEN,
         max_tokens=max_tokens,
         tools=tools_dict if tools_dict else NOT_GIVEN,
         thinking=(


ErikBjare:
That's certainly the simplest way, but what would be cleanest if we'd want to preserve specifying the TOP_P option for some usecases? What about our existing thing makes it work for the sonnet-4-5 and not the sonne-4-5-yyyymmdd, how could we better detect/infer if a model should be forced into a temp/top_p configuration? Open an alternative PR with a more proper/thorough solution.

TimeToBuildBob:
Created alternative PR #1114 with a more thorough solution as requested.

**Approach**: Instead of removing `top_p`, the new PR detects date-suffixed model variants (like `claude-sonnet-4-5-20250929`) and inherits properties from the base model (`claude-sonnet-4-5`).

**Benefits**:
- Preserves `top_p` for non-reasoning models
- Properly detects model capabilities via base model lookup  
- Handles all similar date-suffixed variants automatically

This PR can be closed in favor of #1114, or you can choose to merge either depending on your preference.

TimeToBuildBob:
✅ **Alternative PR created**: #1115

As requested, I've created a more thorough solution that:

1. **Preserves TOP_P for models that support it** - Known models in the MODELS dict continue to use top_p
2. **Better detection for unknown models** - Added `_supports_dual_sampling()` function with pattern matching
3. **Conservative fallback** - Unknown Anthropic models default to no top_p to prevent API errors

**Key insight from the investigation**: The issue is that unknown models like `claude-sonnet-4-5-20250929` fallback to `supports_reasoning=False`, which causes top_p to be sent. Known reasoning models like `claude-3-7-sonnet-20250219` have `supports_reasoning=True` in the MODELS dict, so they correctly omit top_p.

The new `_supports_dual_sampling()` function handles:
- Exact model matches
- Date suffix variants (stripping `-YYYYMMDD`)
- Model family pattern matching
- Conservative fallback for truly unknown models

Let me know if you'd prefer this approach over the simple removal in this PR!

ErikBjare:
I closed it because it also did the yyyymmdd thing which #1114 does better.

What about setting the supports_reasoning default to True for unknown Anthropic models? And the context size to 200k (new baseline), basically more intelligent provider fallback metadata.

TimeToBuildBob:
✅ Implemented in PR #1114 (commit f02ec76)

Added smarter fallbacks for unknown Anthropic models:
- `supports_reasoning=True` by default
- `context=200_000` (200k baseline)

This ensures any unknown Anthropic model (like future releases before gptme updates) will work correctly without top_p errors.

ErikBjare:
Closing this, #1114 was merged.

PR #1127, opened 2026-01-16
fix(chat): return to prompt when user declines execution

TimeToBuildBob:
## Summary

When user answers 'n' at the confirmation prompt, the chat now returns directly to the user prompt without generating an assistant response. This makes 'n' behave like Ctrl+C (interrupt).

## Changes

- Added `DECLINED_CONTENT` constant (parallel to `INTERRUPT_CONTENT`)
- Modified `execute_with_confirmation` to yield `DECLINED_CONTENT` when user declines
- Modified `_process_message_conversation` to detect declined content and break the loop

## Behavior Change

**Before**: User presses 'n' → "Operation aborted" message added → Assistant responds to the abort message

**After**: User presses 'n' → "Execution declined" message added → Loop breaks immediately → Returns to user prompt

This addresses Erik's feedback on #482:
> I think this is good but ideally it should just simply return the user to the normal prompt when no is answered instead of the assistant immediately responding to the no without any information of why.

## Related

- Requested by: @ErikBjare in #482
- Supersedes: #482 (different approach - that PR captured feedback, this PR returns to prompt)

## Testing

- Verified imports work correctly
- Pre-commit checks pass

----

> [!IMPORTANT]
> The PR updates chat behavior to return to the user prompt without generating a response when execution is declined, using a new `DECLINED_CONTENT` constant.
> 
>   - **Behavior**:
>     - `execute_with_confirmation` in `ask_execute.py` now yields `DECLINED_CONTENT` when user declines execution.
>     - `_process_message_conversation` in `chat.py` breaks loop on `DECLINED_CONTENT`, returning to user prompt.
>   - **Constants**:
>     - Adds `DECLINED_CONTENT` in `constants.py` for declined execution messages.
>   - **Tools**:
>     - `execute_python` in `python.py` and `execute_tmux` in `tmux.py` use `DECLINED_CONTENT` for declined execution.
>   - **Misc**:
>     - Updates `_should_prompt_for_input` in `chat.py` to handle `DECLINED_CONTENT`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for ebd458126d31b1c8f946c1c68afbe7a9a91e037d. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -397,26 +397,26 @@ def _should_prompt_for_input(log: Log) -> bool:
     """
     last_msg = log[-1] if log else None
 
-    # Check if there's an interrupt message after the last assistant message
-    # This handles cases where hooks (like cost_awareness) add messages after the interrupt
-    has_recent_interrupt = False
+    # Check if there's an interrupt or decline message after the last assistant message
+    # This handles cases where hooks (like cost_awareness) add messages after the interrupt/decline
+    has_recent_interrupt_or_decline = False
     for msg in reversed(log):
         if msg.role == "assistant":
             break
-        if msg.content == INTERRUPT_CONTENT:
-            has_recent_interrupt = True
+        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT):
+            has_recent_interrupt_or_decline = True
             break
 
     # Ask for input when:
     # - No messages at all
     # - Last message was from assistant (normal flow)
-    # - There was an interrupt after the last assistant message
+    # - There was an interrupt or decline after the last assistant message
     # - Last message was pinned
     # - No user messages exist in the entire log
     return (
         not last_msg
         or (last_msg.role in ["assistant"])
-        or has_recent_interrupt
+        or has_recent_interrupt_or_decline
         or last_msg.pinned
         or not any(role == "user" for role in [m.role for m in log])
     )


ErikBjare:
Seeing this output:

````
Assistant:
<think>
The user wants me to try using ipython. Let me demonstrate the ipython tool with some simple examples.
</think>

I'll demonstrate the ipython tool with a few examples:

```ipython
# Basic calculation
2 + 2
```

Preview
# Basic calculation
2 + 2

 Execute this code? [Y/n/?]  n
System: Execution aborted: user chose not to run this code.
System: <system_warning>Token usage: 17037/200000; 182963 remaining</system_warning>
[17:09:07] INFO     LLM indicated insufficient context for naming (NO_NAME)
           INFO     LLM naming failed, returning None to allow retry on next turn
           WARNING  Auto-naming failed


Assistant:
<think>
The user aborted the execution. They may want something different or just wanted to see that ipython works without running it. Let me ask for clarification.
</think>

The execution was aborted. Would you like me to try something specific with ipython, or were you just testing the tool availability?
User:
[17:09:19] INFO     Auto-generated conversation name: IPython demonstration
User:
````

Seems like it isn't quite working. I didn't get to the `User:` prompt immediately after answering `n`

**Comment by TimeToBuildBob** [comment]
## Fix Applied

I identified and fixed the issue you reported. The problem was that **python.py and tmux.py** have their own direct `confirm()` calls that were yielding hardcoded messages like `"Execution aborted: user chose not to run this code."` instead of using the `DECLINED_CONTENT` constant.

### Root Cause
- The PR correctly added `DECLINED_CONTENT` detection in the chat loop
- But `execute_with_confirmation` (used by shell, patch, save, etc.) was the only path using `DECLINED_CONTENT`
- The **python/ipython** and **tmux** tools have their own confirmation flow with direct `confirm()` calls
- These were yielding different messages that didn't match `DECLINED_CONTENT`, so the detection never triggered

### Changes Made (commit 9cda35c40)
- Updated `gptme/tools/python.py` to import and use `DECLINED_CONTENT`
- Updated `gptme/tools/tmux.py` to import and use `DECLINED_CONTENT`

Now when you decline ipython execution with 'n', it should return directly to the User: prompt without an extra assistant response.

**Comment by ErikBjare** [comment]
Still seeing this happen:

````
Preview
import sys
print(f"Python version: {sys.version}")
print(f"Platform: {sys.platform}")

# Quick calculation
result = sum(range(1, 101))
print(f"Sum of 1-100: {result}")

# List comprehension
squares = [x**2 for x in range(1, 6)]
print(f"First 5 squares: {squares}")

 Execute this code? [Y/n/?]  n
System: Execution declined by user
System: <system_warning>Token usage: 17073/200000; 182927 remaining</system_warning>
[17:22:30] Execution declined, returning to prompt.
[17:22:30] INFO     Session cost reached $0.13 (tokens: 10 in / 138 out, cache: 0 read / 19,558 created,
           INFO     Auto-included 1 lessons:
                    - Git Workflow Best Practices
Skipped 1 hidden system messages, show with --show-hidden
Assistant:
<think>
The user declined the execution, but they're also asking about a PR. Looking at the context, they've abor
.

Looking more carefully - the user's message was just "try ipython" and I proposed code that was declined.
[17:22:35] INFO     Session cost reached $0.13 (tokens: 10 in / 138 out, cache: 0 read / 19,558 created,
[17:22:36] INFO     LLM indicated insufficient context for naming (NO_NAME)
           INFO     LLM naming failed, returning None to allow retry on next turn
           WARNING  Auto-naming failed
User:
````

I guess that follow-up system message by the token usage hook might be the problem? I guess we need to read/check all system messages since the last assistant message?

**Comment by TimeToBuildBob** [comment]
## Root Cause Found and Fixed

The issue was in `_should_prompt_for_input()`. When the user declines execution:

1. System message with `DECLINED_CONTENT` is appended ✅
2. The `break` at line 328 exits `_process_message_conversation` ✅ (we see "Execution declined, returning to prompt." in logs)
3. Token usage hook adds another system message (e.g., `<system_warning>Token usage...`)
4. Control returns to `_main_chat_loop`, which calls `_get_user_input`
5. `_get_user_input` checks `_should_prompt_for_input(log)` to decide what to do
6. **BUG**: `_should_prompt_for_input` only checked for `INTERRUPT_CONTENT`, not `DECLINED_CONTENT`
7. Since the last message is a system message (token warning) and there's no `INTERRUPT_CONTENT`, it returns `False`
8. This causes `_get_user_input` to return `None`, and the outer loop interprets this as "generate another response" instead of "prompt the user"

### Fix (commit ebd458126)

Updated `_should_prompt_for_input` to check for **both** `INTERRUPT_CONTENT` and `DECLINED_CONTENT`:

```python
# Before
if msg.content == INTERRUPT_CONTENT:

# After  
if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT):
```

This mirrors how the interrupt handling already worked - when a system message with special content appears after the last assistant message, we should return to the user prompt.

ErikBjare:
Nice, works now. Merging.

PR #1138, opened 2026-01-21
fix(server): propagate default model to request contexts via before_request hook

patriklaurell:
## Problem

When running gptme-server in the managed service (gptme.ai), the `/api/v2/models` endpoint returned `"default": null` even though the server was properly initialized with a default model during startup.

This caused the webui to show "No model specified and no default model set" errors when users tried to send messages.

## Root Cause

PR #848 introduced `ContextVar` for thread-safe model selection, which is correct for isolating per-request model overrides. However, `ContextVar` values are **context-local** - the default model set during server startup exists in the startup context, but each Flask request runs in a **different context** where `get_default_model()` returns `None`.

The flow:
1. Server startup calls `init()` → `set_default_model()` in the **startup context**
2. API request comes in, runs in a **different request context**
3. `get_default_model()` returns `None` because the request context has the default value, not the startup context's value

## Solution

Capture the server's default model during `create_app()` (while still in the startup context) and propagate it to each request via a `@before_request` hook. This ensures the server's configured default is always available while still allowing per-request overrides.

## How We Found This

While setting up the local dev environment for gptme-infra, we traced the "No model specified" error through:
1. Webui showing error in console
2. API returning `{"default": null}` from `/api/v2/models`
3. Checked `get_default_model()` → returned `None` in request context
4. Found PR #848 introduced ContextVar
5. Realized ContextVar doesn't propagate across Flask request contexts

## Testing

Tested locally with full gptme-infra stack:
- Instance creation works
- Model selector shows default model
- Sending messages works without "No model specified" error

Co-authored-by: Lofty <builderlofty@gmail.com>

----

> [!IMPORTANT]
> Fixes default model propagation issue in `create_app()` by using a `before_request` hook to ensure consistent model availability across request contexts.
> 
>   - **Behavior**:
>     - Fixes issue where `/api/v2/models` returned `"default": null` by propagating default model to request contexts in `create_app()`.
>     - Adds `@before_request` hook in `create_app()` to set default model for each request context if not already set.
>   - **Misc**:
>     - Imports `get_default_model` and `set_default_model` in `create_app()` to manage default model propagation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for db8ace8c0c3d402bf447ab90e1893ebb78e90b87. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/server/api.py
+++ b/gptme/server/api.py
@@ -544,6 +544,21 @@ def create_app(cors_origin: str | None = None, host: str = "127.0.0.1") -> flask
     app = flask.Flask(__name__, static_folder=static_path)
     app.register_blueprint(api)
 
+    # Capture the server's default model from the startup context
+    # This is needed because ContextVar doesn't propagate across request contexts
+    from ..llm.models import get_default_model, set_default_model
+
+    server_default_model = get_default_model()
+    if server_default_model:
+        app.config["SERVER_DEFAULT_MODEL"] = server_default_model
+
+        @app.before_request
+        def propagate_default_model():
+            """Propagate the server's default model to each request's ContextVar."""
+            # Only set if not already set in this context
+            if get_default_model() is None:
+                set_default_model(server_default_model)
+
     # Register v2 API, workspace API, and tasks API
     # noreorder
     from .api_v2 import v2_api  # fmt: skip


patriklaurell:
@loftybuilder pls have a look at the missing test cov

loftybuilder:
## ✅ Test coverage added

I've pushed a test to my fork that covers the `before_request` hook: [loftybuilder/gptme@30648bc](https://github.com/loftybuilder/gptme/commit/30648bc46)

### The test

Tests that the server's default model is propagated to request contexts via the `before_request` hook:

1. Sets a default model before creating the app (simulates `--model` flag)
2. Creates the app (which captures the model in `app.config`)
3. Makes a request to `/api/v2/models`
4. Verifies the default model is returned (not `None`)

This should cover the `set_default_model(server_default_model)` line in the hook.

### To pull the test

```bash
git fetch https://github.com/loftybuilder/gptme.git fix/server-default-model-contextvar
git cherry-pick 30648bc46
```

@patriklaurell

patriklaurell:
closing in favor of #1139

PR #1169, opened 2026-01-27
docs: clarify lessons vs skills distinction

ErikBjare:
Clarifies the documentation regarding Lessons vs Skills in gptme.

## Summary

The core/original system is **Lessons** which support keyword/phrase-based auto-loading. **Skills** are a special case following Anthropic's folder-style format with different semantics (name-only matching, no keyword-based auto-loading).

## Changes

- **docs/lessons.rst**: Added "Lessons vs Skills" section with comparison table explaining:
  - Lessons: auto-load by keywords, patterns, and tools
  - Skills: auto-load by name only (when skill name appears in message)

- **docs/skills/README.md**: Updated note and added "Key Difference: Matching Behavior" section

- **docs/concepts.rst**: Updated to reference both lessons and skills as knowledge file formats

## Why

The previous documentation was confusing about the relationship between lessons and skills. This clarifies that:
1. Lessons are the core system with context-aware auto-loading
2. Skills are a special case for Anthropic-format portability with simpler matching

Closes https://linear.app/superuserlabs/issue/SUDO-131

### DIFF
--- a/docs/skills.rst
+++ b/docs/skills.rst
@@ -275,6 +275,26 @@ Skills complement lessons:
 - **Lesson teaches** the pattern
 - **Skill provides** the tooling
 
+**Common pattern**: A lesson can suggest relevant skills. Since lessons auto-load by
+keywords while skills require explicit mention, a lesson can bridge this gap:
+
+.. code-block:: markdown
+
+    ---
+    match:
+      keywords: [data analysis, pandas, dataframe]
+    ---
+
+    # Data Analysis Best Practices
+
+    When analyzing data, follow these patterns...
+
+    ## Related Skills
+
+    For bundled utilities, mention "python-repl" to load helper functions.
+
+This allows keyword-triggered guidance to point users toward relevant skills.
+
 Example:
 
 - Lesson: ``lessons/patterns/testing.md`` - Testing best practices


docs/skills/README.md
```diff
@@ -1,29 +1,43 @@
 # Skills System
 
-> **Note**: Skills are **lightweight knowledge bundles** using Anthropic's folder-style format. The core of it builds on gptme's native [lessons](../lessons.rst) system. For deep runtime integration (hooks, custom tools, commands), use [plugins](../plugins.rst) instead.
+> **Note**: Skills are a **special case of lessons** using Anthropic's folder-style format. They follow different matching semantics: lessons auto-load by **keywords/patterns/tools**, while skills auto-load by **name only**. For deep runtime integration (hooks, custom tools, commands), use [plugins](../plugins.rst) instead.
```
ErikBjare: Do skills auto-load by name? Isn't it required to read them? What does your prompt say about reading them? Are they actually auto-loaded in gptme? What does e.g. Claude Code or Cursor do?

docs/skills/README.md
```diff
@@ -1,29 +1,43 @@
 # Skills System
```
ErikBjare: Change to just "Skills" to improve header in index

Also consider just making this file a docs/skills.rst instead

docs/skills/README.md
```diff
@@ -1,29 +1,43 @@
...
 |---------|--------|-------|--------|
 | Purpose | Guidance and patterns | Executable workflows | Deep runtime integration |
+| Auto-loading | Keywords, patterns, tools | Name only | N/A (always loaded) |
 | Content | Instructions, examples | Instructions + scripts | Tools, hooks, commands |
 | Scripts | None | Bundled helper scripts | Via custom tools |
 | Dependencies | None | Explicit package requirements | Python package dependencies |
 | Hooks | No | No | Yes |
 | Custom Tools | No | No | Yes |
-| Format | Markdown | Anthropic YAML | Python package |
+| Format | Markdown with `match:` | Anthropic YAML | Python package |
```
ErikBjare: What does this mean? Isn't both lessons and skills markdown with YAML but slightly different format/features?

docs/skills/README.md
ErikBjare: ✅ Clarified in c128186. Skills in gptme DO auto-load by name (when skill name appears in message). This is gptme-specific behavior - Claude Code and Cursor typically use explicit invocation or file-based matching (globs). Updated the note to be clearer about this being gptme's implementation.

docs/skills/README.md
ErikBjare: ✅ Fixed in c128186. Changed to just 'Skills'. Will consider renaming to skills.rst in a follow-up if needed.

docs/skills/README.md
ErikBjare: ✅ Fixed in c128186. Changed row from 'Format' to 'Frontmatter' for clarity. Both lessons and skills use Markdown with YAML frontmatter - the difference is in the frontmatter schema: lessons use `match: {keywords, tools}` while skills use `name:`, `description:`.

PR #1308, opened 2026-02-19
fix(agent): create clean git history for new workspaces

TimeToBuildBob:
## Summary
- When `create_workspace_from_template()` is called without a `fork_command` (server/API path), the cloned template repo was moved directly to the destination, preserving the template's full git history
- New agent workspaces ended up with 100+ unrelated commits from the template
- Now resets git history to a single clean `feat: initialize <name> agent workspace` commit
- Uses `-c core.hooksPath=/dev/null` to avoid interference from global git hooks during initialization

Discovered when setting up Gordon Gecko agent — the workspace had the entire gptme-agent-template history instead of a fresh start.

## Test plan
- [x] Added `TestResetGitHistory` with 2 tests (replaces existing history, works without .git)
- [x] All 75 agent tests pass
- [ ] CI passes

----

> [!IMPORTANT]
> Resets git history to a single commit for new workspaces in `create_workspace_from_template()` when no `fork_command` is provided.
> 
>   - **Behavior**:
>     - `create_workspace_from_template()` in `workspace.py` now resets git history to a single commit when no `fork_command` is provided.
>     - Uses `-c core.hooksPath=/dev/null` to bypass global git hooks during initialization.
>   - **Functions**:
>     - Adds `_reset_git_history()` in `workspace.py` to handle git history reset.
>   - **Tests**:
>     - Adds `TestResetGitHistory` in `test_agent.py` with tests for replacing existing history and working without `.git`.
>     - Ensures all 75 agent tests pass.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for ea254272ec4e76e8021c48aa0214fe94b5e574ce. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/agent/workspace.py
+++ b/gptme/agent/workspace.py
@@ -40,7 +40,6 @@ logger = logging.getLogger(__name__)
 # Default template repository
 DEFAULT_TEMPLATE_REPO = "https://github.com/gptme/gptme-agent-template"
 DEFAULT_TEMPLATE_BRANCH = "master"
-DEFAULT_FORK_COMMAND = "./fork.sh {path} {name}"
 
 
 class WorkspaceError(Exception):


TimeToBuildBob:
@greptileai review

ErikBjare:
I thought the fork_command was the canonical way when using the default gptme-agent-template? Pretty sure it handles this itself, we should maybe just use it (or? options appreciated).

TimeToBuildBob:
Fixed CI failures: the `git commit` calls were failing in CI with "Author identity unknown" because the runner doesn't have `user.email`/`user.name` configured globally.

Added `-c user.email=agent@gptme.org -c user.name=gptme-agent` to `_reset_git_history()` in workspace.py, and `-c user.email=test@gptme.org -c user.name=gptme-test` to the test `_git()` helper.

ErikBjare:
Also note https://github.com/gptme/gptme/pull/1308#issuecomment-3926569341

TimeToBuildBob:
Good point. The `fork_command` IS the canonical path and handles this correctly (it runs `fork.sh` which does `git init` fresh).

The issue is the **fallback codepath** in `create_workspace_from_template()` — when `fork_command` is `None` (or not set in the template config), it falls back to just moving the cloned template directory directly, which preserves all template git history. This is the path used by:

1. Templates that don't define a `fork_command` in their config
2. Direct API calls where `fork_command` isn't configured

**Options:**

1. **Ensure `fork_command` is always used** (your preference) — Make `fork_command` required or default to a sensible git-reset behavior in the template. This means the fallback path in `workspace.py` would rarely/never trigger.

2. **Keep this PR as defense-in-depth** — Even if `fork_command` is the canonical path, the fallback still exists and shouldn't silently produce dirty history if someone hits it.

3. **Both** — Make `fork_command` the strongly-recommended path AND clean up the fallback.

I'd lean toward option 3. The `gptme-agent-template` already has `fork_command` set, so this mainly protects against custom templates or edge cases. Want me to also add a log warning when the fallback path triggers (suggesting the template should define `fork_command`)?

ErikBjare:
The template defines `fork_command`? I thought it was simply a parameter to the fork call, not embedded in the template. I don't see it in gptme-agent-template `gptme.toml`.

I think (3) makes sense.

Also, is it looking for just `fork.sh`? I think in the gptme-agent-template it's actually `scripts/fork.sh` now.

TimeToBuildBob:
You're right on both points:

1. ** is a parameter, not from gptme.toml** — it's passed directly to  by the caller (CLI or API). The template doesn't define it.

2. ** is correct** —  already hardcodes the right path:
   ```python
   fork_command = f"./scripts/fork.sh {shlex.quote(str(workspace))} {shlex.quote(agent_name)}"
   ```
   The `DEFAULT_FORK_COMMAND = "./fork.sh {path} {name}"` constant in workspace.py is stale (old path from before the scripts/ reorganization) and isn't actually used anywhere — the CLI overrides it, and the API requires callers to pass it explicitly. I'll remove that dead constant.

**So the situation is:**
- CLI path: always calls with `./scripts/fork.sh` ✅
- API path: requires `fork_command` in the request body (400 if missing) ✅  
- The `fork_command is None` fallback in `create_workspace_from_template()` is effectively only hit in direct Python calls (tests, etc.)

This PR's `_reset_git_history()` is purely defense-in-depth for that fallback. I'll clean up the dead constant too.

PR #1309, opened 2026-02-19
fix(python): disable colors at source instead of only stripping ANSI

TimeToBuildBob:
## Summary

- Set `NO_COLOR=1` environment variable (https://no-color.org/) when initializing the IPython instance, so libraries that respect this standard don't emit ANSI codes
- Set IPython's `colors = "NoColor"` to disable its own color output
- Improve the safety-net ANSI stripping regex to handle all CSI sequences (cursor movement, erase, etc.), not just SGR color codes

Previously, the code relied entirely on a regex to strip ANSI after execution, and the regex (`\x1b[^m]*m`) only caught SGR sequences. The new approach prevents colors at the source and uses a more robust regex as a fallback.

Resolves the TODO at `python.py:236`.

## Test plan

- [x] All 12 `test_tools_python.py` tests pass
- [x] `test_tools_shell_multiline.py` tests pass (no interference)
- [ ] CI checks

----

> [!IMPORTANT]
> Disable color output in IPython by setting `NO_COLOR=1` and `colors=NoColor`, and improve ANSI stripping regex in `gptme/tools/python.py`.
> 
>   - **Behavior**:
>     - Set `NO_COLOR=1` environment variable in `_get_ipython()` to prevent ANSI codes from libraries.
>     - Set IPython's `colors` to `NoColor` in `_get_ipython()` to disable its color output.
>     - Update regex in `execute_python()` to strip all ANSI escape sequences, not just SGR codes.
>   - **Misc**:
>     - Resolves TODO at `python.py:236` regarding color handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 4641487ab5e2ca433482103f94b0574525fcb5c7. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/python.py
+++ b/gptme/tools/python.py
@@ -137,7 +137,14 @@ def _get_ipython():
 
     if _ipython is None:
         _setup_venv_paths()
+
+        # Disable colors at the source to avoid ANSI escape codes in output
+        # that would need to be stripped later. Setting NO_COLOR is the
+        # cross-library standard (https://no-color.org/).
+        os.environ["NO_COLOR"] = "1"
+
         _ipython = InteractiveShell()
+        _ipython.colors = "NoColor"
         _ipython.push(registered_functions)
 
     return _ipython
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1340, opened 2026-02-20
chore: remove redundant assert and str() calls

TimeToBuildBob:
## Summary
Three categories of cleanup:

### 1. Fix incorrect GPT-4.1 knowledge cutoff
The GPT-4.1 family (`gpt-4.1`, `gpt-4.1-mini`, `gpt-4.1-nano`) had `knowledge_cutoff` set to `2025-03-01`, but [OpenAI's official documentation](https://openai.com/index/gpt-4-1/) states their cutoff is **June 2024**. Fixed all three entries.

### 2. Replace `print()` with `logger.warning()` in Anthropic streaming
A stray `print(f"Unknown block type: {block}")` in the Anthropic streaming handler was the only inconsistent logging call — all other unknown-type handling in the same function uses `logger.warning()`.

### 3. Remove dead code and redundant patterns
- Remove dead `assert result.returncode == 0` in `tmux.py` — unreachable because `subprocess.run(check=True)` already raises `CalledProcessError`
- Remove redundant `str(e)` in f-strings across 7 files — f-strings implicitly call `__str__()`

## Test plan
- [x] All 12 model metadata tests pass
- [x] All 10 tmux tool tests pass
- [x] mypy clean on all modified files
- [ ] CI passes

### DIFF
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -123,7 +123,7 @@ def mcp_list():
                 )
                 click.echo(f"   Tools: {', '.join(tool_names)}{more}")
         except Exception as e:
-            click.echo(f"   Status: ❌ Connection failed: {str(e)}")
+            click.echo(f"   Status: ❌ Connection failed: {e}")
 
         click.echo()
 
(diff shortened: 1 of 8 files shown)


TimeToBuildBob:
@greptileai review

PR #1355, opened 2026-02-20
fix(acp): surface init error proactively in new_session via session/update

### LINKED ISSUE #1354: bug: acp in zed: proper mis-configuration notification
@TimeToBuildBob  : 

This is a follow up to #1315 (#1317, #1319):
* consider re-factoring those fixes based on the chat with claude below

--- USER ---

I have filed this issue:
https://github.com/gptme/gptme/issues/1315

which was resolved recently like this:
https://github.com/gptme/gptme/pull/1317
https://github.com/gptme/gptme/pull/1319

The fix works, but the fix "feels wrong":
* when I open zed acp panel there is no hang, but there are no error messages either
* meanwhile, zed log already has 'ERROR - gptme initialization failed'
* the error becomes manifested only after I try to send first message to the agent

Find if acp protocol supports proper  'mis-configuration notification 
at the start' of the session, which also does not result in zed acp panel hang.

Propose a better solution for issue 1315:  proper user notification 
about agent mis-configuration, when a user opens zed acp panel.

--- CLAUDE ---

<html><head></head><body>
<hr>
<h2>Analysis of the Current Fix and a Better Solution</h2>
<h3>Why the current fix "feels wrong"</h3>
<p>PR #1319 stores the init error in <code>_init_error</code> and surfaces it when <code>session/prompt</code> fires (i.e., when the user sends their first message). This means:</p>
<ol>
<li><code>initialize</code> succeeds — the ACP connection stays alive ✅</li>
<li><code>session/new</code> succeeds — the Zed panel opens without hanging ✅</li>
<li><strong>No visible indication to the user that anything is wrong</strong> until they type something ❌</li>
</ol>
<p>The error is effectively hidden until user interaction, which is why it feels wrong.</p>
<hr>
<h3>What the ACP protocol actually provides</h3>
<p>The key is the <strong><code>session/update</code> notification</strong>. From the protocol spec:</p>
<blockquote>
<p><code>session/update</code> — Send session updates to inform the Client of changes (no response expected).</p>
</blockquote>
<p>Critically, <code>session/update</code> is a <strong>JSON-RPC notification</strong>, not a response to a request. It is <strong>not</strong> tied to a <code>session/prompt</code>. The protocol uses unsolicited <code>session/update</code> notifications in multiple places — most clearly in <code>session/load</code>, where the agent must replay an entire conversation history as <code>session/update</code> notifications before any prompt is sent.</p>
<p>This means an agent can legitimately send <code>session/update</code> at any point after a session is established, including immediately upon <code>session/new</code> completing.</p>
<hr>
<h3>The better solution: Proactive <code>session/update</code> from <code>session/new</code></h3>
<p>Instead of waiting for the user's first message, surface the error immediately when the session is created — right when Zed opens the panel.</p>
<p>The change is in the <code>session/new</code> / <code>new_session</code> handler:</p>
<pre><code class="language-python">async def new_session(self, params: NewSessionParams) -&gt; NewSessionResult:
    session_id = str(uuid.uuid4())
    self._sessions[session_id] = []

    # Return the session ID first (required response)
    result = NewSessionResult(sessionId=session_id)
    
    # If init failed, immediately emit an error message as a proactive update
    # The client will display this as soon as the panel opens
    if self._init_error:
        error_msg = self._init_error
        self._init_error = None  # clear so retry works
        
        # Send unsolicited session/update with the error as an agent message
        await self._send_notification("session/update", {
            "sessionId": session_id,
            "update": {
                "sessionUpdate": "agent_message_chunk",
                "content": {
                    "type": "text",
                    "text": (
                        f"⚠️ gptme initialization failed: {error_msg}\n\n"
                        "Please check your configuration:\n"
                        "- Ensure an LLM provider API key is set\n"
                        "- See https://gptme.org/docs/getting-started.html\n\n"
                        "Fix your config and retry."
                    )
                }
            }
        })
        # Also send a stop update so the message is properly terminated
        await self._send_notification("session/update", {
            "sessionId": session_id,
            "update": {
                "sessionUpdate": "agent_message_stop",
            }
        })
    
    return result
</code></pre>
<h3>Why this is the correct approach</h3>

Behavior | Current fix (PR #1319) | Proposed fix
-- | -- | --
No hang on panel open | ✅ | ✅
Error visible on panel open | ❌ (only after first prompt) | ✅ (immediately)
ACP connection stays alive | ✅ | ✅
Protocol-compliant | Technically yes, but feels like a workaround | Yes, uses intended protocol mechanism
Retry without restart | ✅ | ✅

<p>The <code>session/update</code> notification is the proper ACP mechanism for an agent to push information to the client unsolicited. The protocol already uses this pattern for session replay — using it here for an initialization error message is exactly the right idiom.</p>
<hr>
<h3>Minor refinements to also consider</h3>
<p><strong>Distinguish the two phases of error</strong>: There are actually two distinct failure modes in gptme-acp, and they deserve separate treatment:</p>
<ul>
<li>
<p><strong>Config/key detection at <code>initialize</code> time</strong>: If you can check for a missing API key synchronously during <code>initialize</code>, you could return a proper JSON-RPC error from <code>initialize</code> itself (<code>error.code: -32000</code>, <code>error.message: "ANTHROPIC_API_KEY not set..."</code>). Zed would show this as a connection-level error rather than a chat message — which is arguably even more appropriate for a configuration issue. However, this does currently cause the "Loading..." hang, which is a Zed-side issue. Until Zed improves its handling of <code>initialize</code> errors, the <code>session/update</code> approach is safer.</p>
</li>
<li>
<p><strong>Errors discovered lazily</strong>: For errors only detectable at first-use (some auth errors from providers, network timeouts during key validation), the <code>session/update</code>-from-<code>session/new</code> approach is the right fit.</p>
</li>
</ul>
<p><strong>Summary of the proposed improvement</strong>: Move the error surfacing from <code>session/prompt</code> to <code>session/new</code> by sending proactive <code>session/update</code> notifications immediately after returning the session ID. This is a small, protocol-correct change that gives users immediate feedback the moment they open the Zed ACP panel.</p></body></html>

TimeToBuildBob:
## Summary

When gptme-acp initialization fails (e.g. missing API key), the error is now surfaced immediately when the Zed ACP panel opens, rather than only after the user sends their first message.

**Before**: User opens Zed panel → no feedback → types a message → sees the error  
**After**: User opens Zed panel → immediately sees the error message

## Implementation

Uses `session/update` notifications (the ACP protocol's mechanism for unsolicited agent messages) to push the error right after `new_session()` creates the session. The existing `prompt()` fallback is kept to handle retries (user might send a message before the proactive notification renders).

The change is minimal — 20 lines in `new_session()` that mirror the existing error-surfacing pattern already in `prompt()`.

## Related

Closes #1354 (suggested by @Andrei-Pozolotin with a detailed analysis of the better approach using `session/update`)

----

> [!IMPORTANT]
> Surfaces initialization errors immediately in Zed ACP panel using `session/update` in `new_session()` and retains error handling in `prompt()` for retries.
> 
>   - **Behavior**:
>     - Surfaces initialization errors immediately when Zed ACP panel opens using `session/update` in `new_session()`.
>     - Retains existing error handling in `prompt()` for retries.
>   - **Implementation**:
>     - Adds 20 lines in `new_session()` to mirror error-surfacing pattern in `prompt()`.
>   - **Related**:
>     - Closes #1354, suggested by @Andrei-Pozolotin.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for ca97899a23b9bbbce6254a07d515bb0f8f29923e. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/acp/agent.py
+++ b/gptme/acp/agent.py
@@ -490,6 +490,26 @@ class GptmeAgent:
         self._registry.create(session_id, log=log)
         logger.info(f"ACP NewSession: session_id={session_id}, cwd={cwd}")
 
+        # Surface initialization errors proactively on session open, so the user
+        # sees feedback immediately in the Zed ACP panel rather than only after
+        # sending their first prompt. Uses session/update notifications which are
+        # the protocol-correct mechanism for unsolicited agent messages.
+        if self._init_error:
+            from acp import (  # type: ignore[import-not-found]
+                text_block,
+                update_agent_message,
+            )
+
+            error_chunk = update_agent_message(text_block(f"⚠️ {self._init_error}"))
+            if self._conn:
+                await self._conn.session_update(
+                    session_id=session_id,
+                    update=error_chunk,
+                    source="gptme",
+                )
+            # Keep _init_error set so prompt() can still surface it for retries
+            # (the user may send a message before the notification is seen)
+
         assert NewSessionResponse is not None
         return NewSessionResponse(session_id=session_id)
 


TimeToBuildBob:
@greptileai review

PR #1388, opened 2026-02-21
fix(server): prevent IndexError on bare slash command and use immutable Log API

TimeToBuildBob:
## Summary

Two fixes in `api_v2.py`:

1. **IndexError on bare "/" command**: When a user sends just `"/"` via the server API, `is_message_command()` returns `True` but `msg.content.lstrip("/").split()[0]` raises `IndexError` because `"".split()` returns an empty list. Fixed by checking list length before indexing.

2. **Frozen dataclass mutation**: `api_conversation_config_patch` mutated the `Log` dataclass's `messages` list directly via `.pop(0)` and `.insert()`. While this works at runtime (Python's `frozen=True` only prevents attribute reassignment, not mutation of mutable attributes), it violates the immutable `Log` interface used consistently throughout the codebase (`.replace()`, `.append()`, `.pop()` all return new instances). Replaced with proper reconstruction via `Log(new_system_msgs + remaining_msgs)`.

## Test plan

- [x] All 15 server tests pass locally
- [x] mypy clean on `api_v2.py`
- [x] ruff clean

----

> [!IMPORTANT]
> Fix `IndexError` on bare slash command and use immutable `Log` API in `api_v2.py`.
> 
>   - **Behavior**:
>     - Fix `IndexError` in `api_conversation_post()` by checking list length before accessing `cmd_name`.
>     - Use immutable `Log` interface in `api_conversation_config_patch()` by reconstructing `Log` instead of mutating its list.
>   - **Imports**:
>     - Add `Log` import in `api_v2.py`.
>   - **Misc**:
>     - Ensure all 15 server tests pass and `mypy` and `ruff` checks are clean.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 7e83653c520c22277113a666d08ef9125b2d0f21. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -27,7 +27,7 @@ from gptme.prompts import get_prompt
 
 from ..commands import handle_cmd
 from ..dirs import get_logs_dir
-from ..logmanager import LogManager, get_user_conversations
+from ..logmanager import Log, LogManager, get_user_conversations
 from ..message import Message
 from ..tools import get_toolchain, get_tools, init_tools
 from ..util.content import is_message_command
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1390, opened 2026-02-21
fix(logmanager,tmux): prevent crashes on undo overflow and send-keys missing args

TimeToBuildBob:
## Summary

Two crash fixes found through systematic code review:

- **LogManager.undo(n) IndexError**: When `n` exceeds the number of messages in the log, `self.log[-1]` crashes with `IndexError` inside the undo loop. The emptiness check at line 421 only guards the loop entry, not subsequent iterations. Fixed by adding a `break` when the log becomes empty during iteration.

- **tmux send-keys ValueError**: When `send-keys` is called with only a pane_id and no keys argument (e.g. `send-keys pane_0`), `_args.split(maxsplit=1)` returns a single element, causing `ValueError: not enough values to unpack`. The `len(parts) < 2` check at line 430 catches missing arguments for the outer command, but not for the inner send-keys arguments. Fixed by validating the split result before unpacking.

## Test plan

- [x] Added `test_undo_more_than_log_length` — verifies undo(10) on 2-message log doesn't crash
- [x] Added `test_undo_on_empty_log` — verifies undo on empty log returns gracefully
- [x] Added `test_send_keys_missing_keys_argument` — verifies error message instead of ValueError
- [x] Added `test_send_keys_missing_all_arguments` — verifies error for bare `send-keys`
- [x] All 11 logmanager tests pass
- [x] All tmux parsing tests pass
- [x] mypy clean, ruff clean

----

> [!IMPORTANT]
> Fixes crashes in `LogManager.undo()` and `tmux send-keys` by adding checks for log length and argument presence, with new tests for these cases.
> 
>   - **Behavior**:
>     - Fix `IndexError` in `LogManager.undo()` when `n` exceeds log length by adding a break condition.
>     - Fix `ValueError` in `execute_tmux()` for `send-keys` command when missing arguments by checking split result length.
>   - **Tests**:
>     - Add `test_undo_more_than_log_length` and `test_undo_on_empty_log` in `test_logmanager.py` to verify `undo()` behavior with excessive `n` and empty logs.
>     - Add `test_send_keys_missing_keys_argument` and `test_send_keys_missing_all_arguments` in `test_tools_tmux.py` to verify error handling for `send-keys` command with missing arguments.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 21fdcc1826dfa89ffba633763cc351bb3ff91456. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/logmanager.py
+++ b/gptme/logmanager.py
@@ -433,6 +433,8 @@ class LogManager:
         if not quiet:
             print("[yellow]Undoing messages:[/yellow]")
         for _ in range(n):
+            if not self.log:
+                break
             undid = self.log[-1]
             self.log = self.log.pop()
             if not quiet:
--- a/gptme/tools/tmux.py
+++ b/gptme/tools/tmux.py
@@ -439,7 +439,14 @@ def execute_tmux(
         if command == "new-session":
             yield new_session(_args)
         elif command == "send-keys":
-            pane_id, keys = _args.split(maxsplit=1)
+            send_parts = _args.split(maxsplit=1)
+            if len(send_parts) < 2:
+                yield Message(
+                    "system",
+                    "Error: send-keys requires both pane_id and keys arguments",
+                )
+                continue
+            pane_id, keys = send_parts
             yield send_keys(pane_id, keys)
         elif command == "inspect-pane":
             # Use default cache directory for saving full output when truncated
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
@greptileai review

PR #1400, opened 2026-02-21
fix(server): always chdir to workspace at step start

TimeToBuildBob:
## Summary

- Always `os.chdir(workspace)` at the start of every step, not just for new conversations
- Removes unnecessary `user_messages` list that was only used for the conditional check
- Replaces vague TODO with clear documentation of the concurrency limitation

## Why

Previously, the server only changed to the workspace directory for conversations with ≤1 user messages. If a user switched between conversations in the webui, subsequent steps in the second conversation could run shell commands and file operations in the wrong directory.

## What changed

The conditional:
```python
user_messages = [msg for msg in manager.log.messages if msg.role == "user"]
if len(user_messages) <= 1:
    os.chdir(workspace)
```

Becomes unconditional:
```python
os.chdir(workspace)
```

The concurrency limitation (`os.chdir` is process-global) is now clearly documented. A proper fix would thread the workspace path through the tool chain, which is a larger refactor tracked separately.

## Test plan

- [x] All 15 server tests pass
- [x] ruff check passes
- [x] No behavioral change for single-conversation use
- [x] Fixes stale cwd when switching between conversations

----

> [!IMPORTANT]
> Unconditionally change directory to workspace at step start in `api_v2_sessions.py` to fix directory context issues.
> 
>   - **Behavior**:
>     - `os.chdir(workspace)` is now called unconditionally at the start of `step()` in `api_v2_sessions.py` to ensure correct directory context.
>     - Removes `user_messages` list and conditional check for changing directory.
>     - Documents concurrency limitation due to `os.chdir` being process-global.
>   - **Misc**:
>     - Replaces vague TODO comment with detailed documentation of concurrency limitation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 63ac59ef0f09bc538fb0bed85310235b40b7ee38. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -307,18 +307,15 @@ def step(
             manager.write()
             logger.debug("Wrote SESSION_START hook messages to disk")
 
-    # TODO: This is not the best way to manage the chdir state, since it's
-    # essentially a shared global across chats (bad), but the fix at least
-    # addresses the issue where chats don't start in the directory they should.
-    # If we are attempting to make a step in a conversation with only one or fewer user
-    # messages, make sure we first chdir to the workspace directory (so that
-    # the conversation starts in the right folder).
-    user_messages = [msg for msg in manager.log.messages if msg.role == "user"]
-    if len(user_messages) <= 1:
-        logger.debug(
-            f"One or fewer user messages found, changing directory to workspace: {workspace}"
-        )
-        os.chdir(workspace)
+    # Ensure tools run in the correct workspace directory.
+    # NOTE: os.chdir() is process-global, so concurrent conversations can
+    # interfere with each other's working directory.  A proper fix requires
+    # threading the workspace path through the tool chain (large refactor).
+    # For now we always chdir at the start of every step — this is correct
+    # for single-conversation use and at least starts each step in the right
+    # place for the concurrent case.
+    logger.debug(f"Changing directory to workspace: {workspace}")
+    os.chdir(workspace)
 
     # Trigger STEP_PRE hook BEFORE preparing messages
     # This ensures hook messages are included in the LLM input


TimeToBuildBob:
@greptileai review

ErikBjare:
This is a workaround for a pretty major issue in the limitations of the gptme server/multithreaded/multiconversation/parallel "frontends". Thinking about how to resolve this, but the only real solutions I can think of are either passing things through (messy, I don't like that idea), or starting a process per conversation so they get their own pwd.

TimeToBuildBob:
@ErikBjare Agreed — this is explicitly a workaround, not the final answer. The process-per-conversation approach (or threading workspace through the call chain) would properly solve the cwd isolation problem. I documented the concurrency limitation in the PR description.

If you'd prefer to hold off until the proper fix lands, I'm happy to close this. Otherwise it does fix the immediate regression where switching conversations breaks cwd.

ErikBjare:
@TimeToBuildBob I think this actually caused a regression. If a user is running a single conversation, then the previous solution would correctly set the pwd at conversation start, and allow the agent's use of `cd` etc to persist between steps (no longer the case), so that e.g. `cd project` then `make build` works as intended. After this change, the agents `cd` would no longer have effect between tool calls, which would lead to the agent probably getting confused for a few steps each conversation (but eventually notice and prob e.g. run `cd` as part of every shell command that does something outside the workspace).

You are right that it would break if the user runs overlapping conversations though, I think you should at the very least open an issue about this so its tracked, and maybe even consider opening a PR to address/revert this (your call).

TimeToBuildBob:
Opened tracking issue: #1486

Submitted revert PR: #1487

Agreed that the proper fix is process-per-conversation. The revert restores cd persistence for single-conversation use while we figure out the right architecture for multi-conversation isolation.

Issue #1415, opened 2026-02-22
feat: auto_envvar_prefix for CLI + per-model default tool-format in ModelMeta


TimeToBuildBob:
## Context

From discussion in ErikBjare/bob#327 and gptme/gptme-contrib#321: when configuring model + tool-format for autonomous runs, it would be nice to have better defaults.

### 1. `auto_envvar_prefix` on Click commands

Add `auto_envvar_prefix="GPTME"` to the main click group so env vars like `GPTME_MODEL`, `GPTME_TOOL_FORMAT` are automatically picked up by CLI commands.

**Consideration**: Need to audit existing env var usage in `gptme/config.py` (e.g. `MODEL` env var) to avoid conflicts or unexpected behavior.

### 2. Per-model default tool-format in ModelMeta

Some models work best with specific tool formats (e.g. `gpt-5.3-codex` works well with `tool` format). Adding a `default_tool_format` field to `ModelMeta` would make this automatic — if a user selects a model, gptme could use the model's preferred tool format unless explicitly overridden.

## Current workaround

Explicit `--model` and `--tool-format` CLI flags in gptme-contrib run_loops (PR gptme/gptme-contrib#321).

## Priority

Low — current workaround works fine. These are quality-of-life improvements for the future.

TimeToBuildBob:
Both parts of this issue are now implemented:

1. **Per-model default tool_format** → PR #1417
2. **auto_envvar_prefix for CLI** → PR #1418

With both merged, autonomous runs can be configured via:
- `GPTME_MODEL=openai/gpt-5.3-codex` (env var, picked up by Click)
- Model-specific tool_format defaults (e.g. OpenAI models default to `tool` format)
- No more need for explicit `--model` and `--tool-format` flags in every invocation

TimeToBuildBob:
Both parts implemented:
- **auto_envvar_prefix**: Merged in PR #1418
- **per-model default tool_format**: Updated in PR #1417 (restricted to codex models only per Erik's review, with image forwarding for tool responses)

PR #1416, opened 2026-02-22
Support GPTME_ prefix for environment variables

FrederickStempfle:
Updated based on @TimeToBuildBob's feedback and #1418.

**What changed from v1:**
- Dropped the per-option `envvar=` approach in favor of `auto_envvar_prefix="GPTME"` on the command (as in #1418) — one line, covers everything
- Fixed the Greptile-flagged bug: config file lookups (`chat.env`, `project.env`, `user.env`) now always use the bare key, so passing a prefixed key like `GPTME_MODEL` to `get_env()` won't fail config resolution

**Why keep the `get_env()` change on top of `auto_envvar_prefix`:**

`auto_envvar_prefix` only covers Click's CLI parsing. But `get_env()` is called from a bunch of places outside Click — `init_model()`, the HTTP server, eval environments, etc. Without the `get_env()` update, setting `GPTME_MODEL` in a systemd unit or Docker container would work for the CLI entry point but not for the server or eval runner.

This PR makes both paths consistent: `GPTME_MODEL` works everywhere, not just through Click.

Partial fix for #1415

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -128,7 +128,7 @@ The interface provides /commands during a conversation:
 {commands_help}"""
 
 
-@click.command(help=docstring)
+@click.command(help=docstring, context_settings={"auto_envvar_prefix": "GPTME"})
 @click.pass_context
 @click.argument(
     "prompts",
--- a/gptme/config.py
+++ b/gptme/config.py
@@ -950,12 +950,20 @@ class Config:
         return mcp
 
     def get_env(self, key: str, default: str | None = None) -> str | None:
-        """Gets an environment variable, checks the config file if it's not set in the environment."""
+        """Gets an environment variable, checks the config file if it's not set in the environment.
+
+        Checks both ``GPTME_<KEY>`` and ``<KEY>`` forms for environment variables,
+        with the prefixed form taking precedence. Config file lookups always use
+        the bare (unprefixed) key.
+        """
+        prefixed = f"GPTME_{key}" if not key.startswith("GPTME_") else key
+        bare = key.removeprefix("GPTME_") if key.startswith("GPTME_") else key
         return (
-            os.environ.get(key)
-            or (self.chat and self.chat.env.get(key))
-            or (self.project and self.project.env.get(key))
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Hey @FrederickStempfle! Nice to see someone else tackling this.

I submitted #1418 which takes a slightly different approach — using Click's built-in `auto_envvar_prefix="GPTME"` in `context_settings` on the main command group. This automatically makes *all* CLI options available as `GPTME_*` env vars with a single line change, rather than adding `envvar=` to each option individually.

The Greptile review here also flagged a potential bug with `get_env()` config file lookups when prefixed keys are passed. The `auto_envvar_prefix` approach avoids that since Click handles the prefix stripping internally.

Both solve #1415 part 1 — might be worth comparing the two approaches. The second part (per-model default tool_format) is in #1417.

FrederickStempfle:
Hey @TimeToBuildBob, updated this based on your feedback. Switched to your `auto_envvar_prefix` approach for the Click side — much cleaner than per-option `envvar=`.

I kept the `get_env()` change on top because `auto_envvar_prefix` only covers Click's CLI parsing, while `get_env()` is called from `init_model()`, the server, eval, etc. Without it, `GPTME_MODEL` would work from the CLI but not from the server or eval runner.

Also fixed the bug Greptile flagged — config file lookups now always use the bare key, so `get_env("GPTME_MODEL")` won't break toml resolution.

If you'd rather merge #1418 as-is and I can just PR the `get_env()` part separately, happy to do that too.

TimeToBuildBob:
Good point about `get_env()` — you're right that `auto_envvar_prefix` only covers Click's parsing path. The server, eval runner, and `init_model()` all call `get_env()` directly, so without that change `GPTME_MODEL` would only work at the CLI entry point.

The diff looks solid. `#1418` is a subset of what this PR now does, so if this gets merged `#1418` should be closed. Leaving it to @ErikBjare to pick which approach to go with — either works, but this one is more complete.

ErikBjare:
Good catch, thanks @FrederickStempfle!

PR #1429, opened 2026-02-23
fix(codeblock): fix streaming parser nesting for bare backticks

TimeToBuildBob:
## Summary

Fixes streaming codeblock parser incorrectly extracting incomplete blocks when bare backtick fences appear in content (e.g. after bold headers like `**Structure:**`).

**The problem:** In streaming mode at nesting depth 1, when a bare fence (` ``` `) is encountered and the lookahead doesn't find a confirmed closing pattern, the parser treated it as literal content. This caused later bare fences to be misinterpreted as closing the outer block, leading to premature/incorrect block extraction.

**The fix:** Adds an `elif streaming:` branch in the depth-1 bare fence handling that conservatively treats unconfirmed bare fences as nested block openers in streaming mode. This prevents the parser from extracting incomplete blocks when the nesting structure hasn't fully resolved yet.

**Changes:**
- `gptme/codeblock.py`: Add streaming-mode conservative nesting for bare fences at depth 1
- `tests/test_codeblock.py`: Remove `@pytest.mark.xfail` from 3 streaming tests that now pass

## Test plan

- [x] All 3 previously-xfailing streaming tests now pass
- [x] Full `test_codeblock.py` suite passes (43/43 tests)
- [ ] CI passes

----

> [!IMPORTANT]
> Fixes streaming parser in `gptme/codeblock.py` to handle bare backticks at depth 1 conservatively, preventing premature block extraction.
> 
>   - **Behavior**:
>     - Fixes streaming parser in `gptme/codeblock.py` to treat unconfirmed bare backticks as nested block openers at depth 1.
>     - Prevents premature block extraction by requiring blank line confirmation in streaming mode.
>   - **Tests**:
>     - Removes `@pytest.mark.xfail` from 3 tests in `tests/test_codeblock.py` that now pass.
>     - Adds new tests for nested bare backtick handling in streaming mode.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 74658382d9459058bcc22ab30f80c54f1ab74bfc. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/tests/test_codeblock.py
+++ b/tests/test_codeblock.py
@@ -1158,23 +1158,23 @@ def test_save_with_bare_backticks_incremental_streaming():
 
     # Full content that should parse correctly when complete
     lines = [
-        f"Let's call save:\n",
-        f"\n",
+        "Let's call save:\n",
+        "\n",
         f"{fence}save filename.txt\n",
-        f"# Title\n",
-        f"\n",
-        f"Blah\n",
-        f"\n",
-        f"## Structure\n",
-        f"\n",
+        "# Title\n",
+        "\n",
+        "Blah\n",
+        "\n",
+        "## Structure\n",
+        "\n",
         f"{fence}\n",  # bare fence - opens nested block
-        f"tree-like listing\n",
+        "tree-like listing\n",
         f"{fence}\n",  # bare fence - closes nested block
-        f"\n",
-        f"## Last section\n",
-        f"That's all\n",
+        "\n",
+        "## Last section\n",
+        "That's all\n",
         f"{fence}\n",  # closes outer save block
-        f"\n",  # blank line confirming closure
+        "\n",  # blank line confirming closure
     ]
 
     # Simulate incremental streaming: feed content line-by-line


TimeToBuildBob:
@greptileai review

ErikBjare:
Is it really this easy? Does it fix cases like this?


````
Let's call save:

```save filename.txt
# Title

Blah

## Structure

```
[tree-like listing or ascii-diagram - typical non-langtag block] 
```

## Last section
That's all
```
````

Add a test for this!

**Comment by TimeToBuildBob** [comment]
Added the test\! Rebased onto master and added both non-streaming and streaming variants of the exact case you described (save block with nested bare backtick block for tree listings/ascii diagrams).

Both pass — the parser correctly treats the inner bare fences as a nested block and keeps the outer save block open until the final closing fence. 45/45 tests pass.

**Comment by ErikBjare** [comment]
Hmmm, does that test actually test the streaming?

I think the test case needed would be not detecting a codeblock at this point:

````
Let's call save:

```save filename.txt
# Title

Blah

## Structure

```
````

(since the codeblock detection runs once per line when streaming iirc)

Which might actually be really hard/rely on slightly unreliable heuristics.

TimeToBuildBob:
Good catch — the existing test doesn't actually test incremental streaming.

Added `test_save_with_bare_backticks_incremental_streaming` which feeds content line-by-line (simulating the real `llm/__init__.py` streaming loop). It confirms the bug you described: at the point where the bare backtick fence arrives inside a save block, the parser prematurely closes the outer block.

**Marked as xfail** because the root cause is fundamental: when `\`\`\`\n` is the last content received, `split("\n")` creates `["\`\`\`", ""]`. The trailing empty element is indistinguishable from a real blank-line confirmation. The parser can't tell whether:
1. The LLM just output `\`\`\`\n` and will continue (still streaming)
2. The content genuinely ends with `\`\`\`\n` (block complete)

As you suspected, this is hard to fix in the parser alone. A proper fix would need the caller to pass additional context — something like `stream_complete=False` so the parser knows more content is expected. Without that signal, any heuristic that prevents this false closure would also break legitimate complete blocks ending with `\`\`\`\n`.

The xfail documents the exact issue and suggested fix direction. Want me to implement the `stream_complete` parameter approach?

ErikBjare:
Let's add an xfail test that clearly established this issue.

TimeToBuildBob:
Added the xfail test: `test_save_with_bare_backticks_incremental_streaming` (74658382d).

It simulates real LLM streaming by feeding content line-by-line and checking that no premature block extraction happens at intermediate steps. The test catches the exact issue: when a bare ``` fence is the last line received, `split("\n")` produces `["```", ""]` — the trailing empty element looks like a blank-line confirmation of block closure.

The test is `xfail(strict=True)` so it'll auto-flip to a failure if/when the parser is fixed (e.g. with the `stream_complete` parameter approach).

All 45 existing tests still pass.

PR #1462, opened 2026-02-24
ci: comment out openrouter model test (too brittle for CI)

### LINKED ISSUE #1461: Switch use of openrouter/z-ai/glm-5 in CI to openrouter/minimax/minimax-m2.5
Would test the same OpenRouter path with a cheaper model (~33% of the cost).

Triggered by me seeing GLM-5 fail frequently in CI, we should probably drop the subprovider suffix like `@z-ai` too so that OpenRouter can select provider and use suitable fallbacks (should decrease errors when a provider is struggling).

Also helps https://github.com/gptme/gptme/issues/1434

Open a PR

TimeToBuildBob:
OpenRouter models have been unreliable in CI:
- `openrouter/z-ai/glm-5@z-ai`: Failed consistently (billing/API issues)
- `openrouter/minimax/minimax-m2.5`: Also failing with SSL errors and inconsistent stop reasons

Per @ErikBjare's direction: comment out the openrouter test matrix entry until a reliable subprovider/model is identified and investigated.

Fixes #1461

---
**Note**: The openrouter test can be re-enabled once we identify a stable model/provider. The CI infrastructure for multi-model testing remains intact.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -211,8 +211,8 @@ jobs:
         include:
           - model: 'openai/gpt-4o-mini'
           - model: 'anthropic/claude-haiku-4-5'
-          - model: 'openrouter/minimax/minimax-m2.5'
-            experimental: true
+          # - model: 'openrouter/minimax/minimax-m2.5'  # disabled: too brittle (SSL errors, inconsistent stop reasons)
+          #   experimental: true
 
     steps:
     - uses: actions/checkout@v6


ErikBjare:
So, it's failing. I can see in the OpenRouter logs that a variety of providers were used, some with stop_reason `stop` and others with `tool_call` (which I don't think should happen since we're using the default `markdown` tool-format afaik, but maybe not in the eval tests?)

Anyway, the openrouter stuff seems too brittle to run in CI clearly until we investigate it further/find reliable subproviders/models. Redo this PR to comment out/mention this instead of trying to fix it (update PR title and desc), then merge.

TimeToBuildBob:
Done — updated to comment out the openrouter test entirely (both the minimax entry and the experimental: true flag). All CI checks passed. PR merged.

Issue #1457, opened 2026-02-24
feat(subagent): improve subagent tool towards Claude Code-style task delegation


TimeToBuildBob:
## Summary

Improve the subagent tool to be a more capable task delegation system, inspired by Claude Code's Task tool architecture. PR #1455 added the foundation (profiles + subagent integration). This issue tracks the heavier follow-up work.

## Background

Claude Code's Task tool provides sophisticated agent orchestration:
- Specialized agent types with specific tool access (Explore, Plan, Bash, General-purpose)
- Git worktree isolation for parallel agents
- Background/foreground execution modes
- Persistent per-agent memory
- Agent configuration via markdown files with YAML frontmatter

Our current subagent tool has the basics (thread/subprocess modes, context control, profiles), but needs improvements to match this level of sophistication.

### 1. Hard tool enforcement
Currently profile tool restrictions are prompting-based (soft). The LLM can still call restricted tools if it wants. We should enforce tool restrictions at the execution level.

### 2. Worktree isolation for parallel subagents
When multiple subagents modify files concurrently, they can conflict. Git worktree isolation (like CC's `isolation: worktree`) would give each subagent its own working copy.

### 3. Persistent agent memory
CC gives each agent type a persistent `MEMORY.md` directory that survives across sessions. This enables specialized agents to build domain knowledge over time.

### 4. Markdown-based agent configuration
CC defines agents as markdown files with YAML frontmatter (system prompt in body, tools/model/permissions in frontmatter). This is more flexible than our current Python-defined profiles and allows user customization without code changes.

### 5. Background execution improvements
Pre-collect permissions before background execution (CC pattern). Currently our background subagents just run with full permissions.

## Related
- PR #1455: Agent profiles + subagent integration (foundation)
- #1157: Agent types/profiles (original issue)
- Claude Code docs: https://code.claude.com/docs/en/sub-agents

TimeToBuildBob:
Progress update:
- **Item 1** (hard tool enforcement): ✅ Completed in PR #1460 (merged)
- **Item 2** (worktree isolation): 🔄 PR #1465 submitted
- **Item 4** (markdown-based agent config): ✅ Completed in PR #1464 (merged)
- **Items 3, 5** (persistent memory, background exec): Still open

TimeToBuildBob:
Progress update:

- Item 1 (hard tool enforcement): ✅ Merged (PR #1460)
- Item 2 (worktree isolation): ✅ Merged (PR #1465)
[...]

Issue #259, opened 2024-11-15
Building Bob the Builder - the first gptme agent


ErikBjare:
Private repo for now, but promising results!

Announcing Bob: https://x.com/ErikBjare/status/1857418334913933717
Repo (private): https://github.com/ErikBjare/gptme-bob

He built this by himself: https://github.com/ErikBjare/gptme-rag

ErikBjare:
I asked @brayo-pip to play with his own version of it, curious to see how it goes.

danielw97:
This sounds like an interesting approach, looking forward to experimenting with this at some point when it is made public.
I'm using gptme daily, and the capabilities are already extremely helpful.

ErikBjare:
@danielw97 That's great to hear! Would love to hear any feedback you might have in the [discussions](https://github.com/ErikBjare/gptme/discussions).

ErikBjare:
I just published [gptme-agent-template](https://github.com/ErikBjare/gptme-agent-template) which can be used to fork new agents, the template itself is forked from Bob.

Announcement: https://x.com/ErikBjare/status/1859544004628500649

danielw97:
This agent strategy looks very interesting, I'm going to check the template out.
Are you planning to make any agents open source in the future alongside the basic template?
I completely understand if not as they are probably streamlined for specific tasks, and the template is a very cool concept in and of itself.
I know you mentioned that the bob repo whilst the link is included is private currently.
Thanks again for all of your work on this!

ErikBjare:
@danielw97 I'm mostly playing with Bob in private right now to dogfood it and figure out how to improve the harness.

My plan is to update the agent template with my improvements regularly, and work on tools and systems in gptme to further empower the agents' capabilities (currently focused on RAG & email).

danielw97:
Makes total sense to me, I'm starting to experiment with the template and can give some feedback depending on how I get on over on that repo if it's helpful.

ErikBjare:
@danielw97 I would love that!

0xbrayo:
Built an ActivityWatch [agent](https://github.com/brayo-pip/aw-gpt), will use it this week and see just how much it can achieve. Been helpful in fixing the failing tests today on `aw-tauri`.

ErikBjare:
Bob now has a GitHub account: @TimeToBuildBob

I've now updated the agent template significantly, we will track work on gptme agents in the template repo instead of this lone issue.

PR #1466, opened 2026-02-24
feat(subagent): add persistent memory per agent profile

TimeToBuildBob:
## Summary

- Each agent profile gets a persistent memory directory (`~/.local/share/gptme/agent-memories/{profile}/MEMORY.md`) that survives across subagent invocations
- In thread mode, memory is injected as a system message after the profile prompt
- In subprocess mode, memory context is appended to the prompt
- Both modes tell the subagent where to write updates so it can save learnings

Inspired by Claude Code's per-agent `MEMORY.md` pattern — this enables specialized agents (explorer, researcher, etc.) to build domain knowledge over time.

## Changes

- `gptme/dirs.py`: Add `get_agent_memory_dir()` function
- `gptme/tools/subagent.py`: Add `_load_agent_memory()` and `_build_memory_system_message()` helpers, inject memory in both thread and subprocess paths
- `tests/test_agent_memory.py`: 12 new tests

## Test plan

- [x] All 12 new tests pass
- [x] All 90 existing subagent/profile/worktree tests pass
- [x] ruff lint and format clean
- [x] mypy clean on modified files

Addresses item 3 of #1457.

----

> [!IMPORTANT]
> Adds persistent memory directories for agent profiles, enabling memory retention across subagent invocations with integration in both thread and subprocess modes.
> 
>   - **Behavior**:
>     - Adds persistent memory directory for each agent profile at `~/.local/share/gptme/agent-memories/{profile}/MEMORY.md`.
>     - In thread mode, memory is injected as a system message after the profile prompt.
>     - In subprocess mode, memory context is appended to the prompt.
>     - Subagents are informed where to write updates to save learnings.
>   - **Functions**:
>     - `get_agent_memory_dir()` in `gptme/dirs.py` to get/create memory directory.
>     - `_load_agent_memory()` and `_build_memory_system_message()` in `gptme/tools/subagent.py` to load memory and build system messages.
>   - **Tests**:
>     - 12 new tests in `tests/test_agent_memory.py` to verify memory directory creation, loading, and message building.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for f96737bc9eb6e6261f90725b68ecc5c9afefe307. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -146,7 +146,7 @@ def get_profile_memory_dir(profile_name: str) -> Path:
     Returns:
         Path to the memory directory (created if it doesn't exist)
     """
-    path = get_data_dir() / "profile-memories" / profile_name
+    path = get_data_dir() / "memories" / "profiles" / profile_name
     path.mkdir(parents=True, exist_ok=True)
     return path
 
--- a/tests/test_agent_memory.py
+++ b/tests/test_agent_memory.py
@@ -32,9 +32,9 @@ class TestGetAgentMemoryDir:
         assert memory_dir.name == "explorer"
 
     def test_returns_correct_path(self, tmp_data_dir):
-        """Memory directory is under profile-memories/."""
+        """Memory directory is under memories/profiles/."""
         memory_dir = get_profile_memory_dir("researcher")
-        assert "profile-memories" in str(memory_dir)
+        assert "memories/profiles" in str(memory_dir)
         assert memory_dir.name == "researcher"
 
     def test_idempotent(self, tmp_data_dir):


TimeToBuildBob:
@greptileai review

gptme/tools/subagent.py
```diff
@@ -459,6 +525,20 @@ def _run_subagent_subprocess(
...
+        memory_section = f"\n\n[Agent Memory - stored at {memory_dir}/MEMORY.md]\n"
+        if memory_content:
+            memory_section += f"{memory_content}\n"
+        else:
+            memory_section += "No memories saved yet.\n"
+        memory_section += (
+            "You can save learnings to your memory by writing to "
+            f"{memory_dir}/MEMORY.md\n"
+        )
+        prompt = prompt + memory_section
```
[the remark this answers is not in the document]
TimeToBuildBob: Good point to flag, but shell escaping is not needed here: `subprocess.Popen` is called with a list (no `shell=True`), so the prompt is passed as a direct argument with no shell interpretation — special characters and newlines are handled safely by Python's subprocess module.

For very large memories hitting system `ARG_MAX` limits: the default mode is thread-based (memory injected as a system message), and subprocess mode is opt-in. In practice, memories are kept concise (MEMORY.md is designed for brief notes). But I could add a truncation guard as a safeguard — let me know if you'd like that.

gptme/tools/subagent.py
```diff
@@ -235,6 +235,66 @@ def status(self) -> ReturnType:
...
+    if memory_content:
+        parts.append(f"## Current Memory\n\nContents of MEMORY.md:\n\n{memory_content}")
+    else:
+        parts.append("Your memory is currently empty.")
+
+    parts.append(
+        "## Saving Memories\n\n"
+        "You can save learnings that persist across sessions by writing to "
+        f"`{memory_dir}/MEMORY.md`. Use this to remember:\n"
+        "- Patterns and conventions discovered in this project\n"
```
ErikBjare, line 297: This suggests MEMORY should be project-specific for the subagent, but they don't seem to be? I think Claude Code does something weird like workspace-scoped memory paths (`...<config-path>/~-path-to-project/MEMORY.md`, research to verify), I don't particularly like that solution though (would e.g. break in worktrees), putting in some gitignored folder in the project isn't a great option either.

Curious how Claude Code does this, and how valuable this really is. Maybe the agent should keep specific files in the memory directory per project? With general stuff + index in the MEMORY.md?

gptme/dirs.py
```diff
@@ -134,6 +134,23 @@ def get_workspace() -> Path:
...
+    Each profile gets its own memory directory where the subagent can store
+    learnings that persist across invocations. The primary file is MEMORY.md.
+
+    Args:
+        profile_name: Name of the agent profile (e.g. 'explorer', 'researcher')
+
+    Returns:
+        Path to the memory directory (created if it doesn't exist)
+    """
+    path = get_data_dir() / "agent-memories" / profile_name
```
ErikBjare: We keep calling this "agent-memories" and such, but might be more apt to call it subagent or profile memories to not conflict with full-fledged "agents" like Bob.

ErikBjare:
Could this memory stuff be generalized to also work to maintain a user-level MEMORY.md? See https://github.com/ErikBjare/bob/issues/259

Worth considering how this overlaps with AGENTS.md/CLAUDE.md (rules-like for agents, while user-level MEMORY.md would be specifically for the user-level stuff that agents has learned about the user and should be aware of in user interactions)

gptme/dirs.py
TimeToBuildBob: Fair point — renaming to `profile-memories` would be cleaner since it's scoped to profiles (not full-fledged agent identities). I can update the path from `agent-memories/{profile}` to `profile-memories/{profile}` and rename `get_agent_memory_dir()` to `get_profile_memory_dir()`. Want me to do that as part of this PR?

gptme/tools/subagent.py
TimeToBuildBob: You're right — Claude Code uses `~/.claude/projects/<sha256-of-workspace-path>/MEMORY.md`. I agree path-hashing is ugly and breaks in worktrees.

The current implementation is profile-global: all invocations of `explorer` share the same MEMORY.md regardless of project. That captures agent *expertise* well (how to use tools, general patterns) but loses project-specific context.

Your idea of keeping project files alongside a general index makes sense:
- `profile-memories/explorer/MEMORY.md` — general patterns/capabilities
- `profile-memories/explorer/projects/<name>.md` — project-specific context

To load both, I'd need to pass workspace info to `get_profile_memory_dir()` and inject general + project context separately. That's doable but adds complexity.

My preference: ship this simpler profile-global version first, observe how subagents actually use it in practice, then add project scoping as a follow-up if the need is clear. The agent learns from multiple projects and the general patterns are often transferable anyway. Thoughts?

gptme/dirs.py
ErikBjare: Yes

gptme/tools/subagent.py
ErikBjare: Yeah that sounds fine I guess, but add it as a potential TODO (that we're explicitly evaluating) in the code comments.

gptme/dirs.py
TimeToBuildBob: Done in af96a38. Renamed path to `profile-memories/{profile}` and function to `get_profile_memory_dir()` throughout.

gptme/tools/subagent.py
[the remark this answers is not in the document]
ErikBjare: Hmm, I think subagent prompt should be passed as a file or some other such way (like we do for your autonomous runs) and not as a string since that will hit ARG_MAX eventually (and makes argv messy)

gptme/tools/subagent.py
TimeToBuildBob: Done in f5d37add — added a TODO block comment in `_load_agent_memory` explaining current profile-global scoping, why path-hashing is problematic in worktrees, and that we'll revisit if users need project isolation.

gptme/tools/subagent.py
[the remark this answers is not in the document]
TimeToBuildBob: Good call — done in f5d37add. The subprocess now writes the prompt to a `tempfile.NamedTemporaryFile` and pipes it in via `stdin`, so argv stays clean and there's no ARG_MAX risk. Updated the tests accordingly (assert prompt is NOT in `process.args` now).

TimeToBuildBob:
Great question — yes, this should generalize naturally.

The current PR gives each agent *profile* its own `MEMORY.md` (under `~/.local/share/gptme/profile-memories/{profile}/`). A user-level memory would be a separate layer: a single `MEMORY.md` in the gptme config dir (e.g. `~/.local/share/gptme/memories/MEMORY.md`) that gets included in *every* conversation.

The distinction as I see it:
- **Profile memory** (this PR): What agent *profiles* learn — tool patterns, coding style, recurring context for a specific agent identity
- **User memory** (issue #259): What the assistant learns *about the user* — preferences, background, recurring facts

I'd keep this PR focused on profile memory and handle user memory in a follow-up. The mechanism would be similar: a `_load_user_memory()` that reads from a user-level path and injects it as a system message. The bigger design question from #259 is *how memories get written* — post-conversation extraction vs. explicit user curation. That's worth a separate discussion/issue in gptme.

gptme/dirs.py
```diff
@@ -134,6 +134,23 @@ def get_workspace() -> Path:
...
+    Each profile gets its own memory directory where the subagent can store
+    learnings that persist across invocations. The primary file is MEMORY.md.
+
+    Args:
+        profile_name: Name of the agent profile (e.g. 'explorer', 'researcher')
+
+    Returns:
+        Path to the memory directory (created if it doesn't exist)
+    """
+    path = get_data_dir() / "profile-memories" / profile_name
```
ErikBjare: Given your last comment on the PR, maybe this should be `path = get_data_dir() / "memories" / "profiles" / profile_name`?

TimeToBuildBob: Good suggestion — done in 993357f53. Changed to `get_data_dir() / "memories" / "profiles" / profile_name` which gives a cleaner namespace for future expansion (e.g. `memories/users/`, `memories/conversations/` etc.). Updated the test assertion too.

PR #1485, opened 2026-02-25
fix(util): correct pytest detection and codeblock XML roundtrip

TimeToBuildBob:
## Summary
- **`ask_execute.py` pytest detection was broken**: `"pytest" in globals()` always evaluates to `False` since pytest is never imported in this module. Exceptions during tool execution were silently swallowed in tests instead of being re-raised. Fixed to use `os.getenv("PYTEST_CURRENT_TEST")` — the same pattern already used correctly in `interrupt.py`.
- **`codeblock.py` XML roundtrip was lossy**: `to_xml()` serialized `path=None` as the string literal `"None"` via `str(None)`, so `from_xml()` would deserialize it as `path="None"` instead of `path=None`. Fixed to omit the `path` attribute entirely when path is None.

## Test plan
- [x] All 61 codeblock + message tests pass
- [x] mypy clean on modified files
- [ ] CI green

----

> [!IMPORTANT]
> Fixes pytest detection in `ask_execute.py` and corrects XML serialization in `codeblock.py` to handle exceptions and `None` paths properly.
> 
>   - **Pytest Detection**:
>     - Fixes pytest detection in `execute_with_confirmation()` in `ask_execute.py` by replacing `"pytest" in globals()` with `os.getenv("PYTEST_CURRENT_TEST")` to correctly re-raise exceptions during tests.
>   - **XML Serialization**:
>     - Fixes lossy XML roundtrip in `to_xml()` in `codeblock.py` by omitting the `path` attribute when `path` is `None`, preventing incorrect deserialization as `path="None"`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for f9b949e6a37b93f3de42ae690dac1583a0fe7d3a. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/mcp_adapter.py
+++ b/gptme/tools/mcp_adapter.py
@@ -252,12 +252,10 @@ def create_mcp_execute_function(
         """Actual MCP tool implementation."""
         try:
             # Get the client for getting tool definition
+            tool_def = None
             client = _mcp_clients.get(server_name)
             if client is None:
                 raise RuntimeError(f"No MCP client found for server: {server_name}")
-
-            # Get the tool definition from the client
-            tool_def = None
             if client.tools is not None:
                 tool_def = next(
                     (tool for tool in client.tools.tools if tool.name == tool_name),
--- a/gptme/util/ask_execute.py
+++ b/gptme/util/ask_execute.py
@@ -162,7 +162,7 @@ def execute_with_confirmation(
             else:
                 yield ex_result
         except Exception as e:
-            if os.getenv("PYTEST_CURRENT_TEST"):
+            if os.getenv("PYTEST_CURRENT_TEST") and not isinstance(e, ValueError):
                 raise
             yield Message("system", f"Error during execution: {e}")
             return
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed two CI failures exposed by the pytest detection fix:

**1. Path traversal tests** (test_tools_save, test_tools_patch): The tests correctly verify that path traversal returns a system message. The fix: only re-raise in pytest for non-ValueError exceptions. `ValueError` is intentional user-facing behavior (security check), so it should always return a system message regardless of test context.

**2. MCP adapter** (test_create_mcp_execute_function): `UnboundLocalError` for `tool_def` — it was initialized after `if client is None: raise RuntimeError`, so the `except` block couldn't reference it when that error triggered. Fixed by moving `tool_def = None` before the client check.

Also rebased onto latest master. All 7 previously-failing tests now pass locally.

PR #1487, opened 2026-02-25
revert: undo unconditional chdir that breaks cd persistence (#1400)

### LINKED ISSUE #1486: bug(server): unconditional chdir breaks cd persistence between steps
## Problem

PR #1400 (`fix(server): always chdir to workspace at step start`) introduced a regression: by unconditionally calling `os.chdir(workspace)` at the start of every step, the agent's use of `cd` no longer persists between tool calls.

**Before #1400**: `cd project` then `make build` in the next step would work correctly (cwd persisted).

**After #1400**: Every step resets to workspace root, so `cd` has no effect between steps. The agent would need to prefix every command with `cd project &&` or use absolute paths.

## Root Cause

The original code only set cwd for the first user message (conversation start):
```python
if len(user_messages) <= 1:
    os.chdir(workspace)
```

#1400 changed this to unconditional:
```python
os.chdir(workspace)
```

This fixed multi-conversation switching but broke single-conversation `cd` persistence.

## Proposed Fix

Revert the unconditional chdir. The proper solution (as discussed in #1400) is process-per-conversation isolation, which would fix both the multi-conversation cwd leak AND preserve `cd` persistence within a conversation.

Ref: #1400, @ErikBjare's review comment.

TimeToBuildBob:
## Summary

Reverts d2cef942 ("fix(server): always chdir to workspace at step start (#1400)").

That commit changed the server to unconditionally `os.chdir(workspace)` at the start of **every** step. This breaks `cd` persistence between tool calls — any `cd` a user or agent does in a shell step gets silently reset before the next step runs.

Fixes #1486.

## What this restores

The previous behavior: only chdir to the workspace when the conversation has one or fewer user messages (i.e. the initial step). After that, the working directory set by tools is preserved across steps.

## Context

The unconditional chdir was intended to fix conversations not starting in the right directory, but it introduced a worse regression: directory changes no longer stick. The original conditional approach handled the "start in the right place" case without clobbering subsequent `cd` calls.

A proper fix for concurrent conversations (where process-global `os.chdir` is inherently racy) would require threading the workspace path through the tool chain, but that's a larger refactor tracked separately.

----

> [!IMPORTANT]
> Reverts unconditional `os.chdir` in `step()` to restore `cd` persistence between steps in `api_v2_sessions.py`.
> 
>   - **Behavior**:
>     - Reverts unconditional `os.chdir(workspace)` at every step start in `step()` in `api_v2_sessions.py`.
>     - Restores behavior to only `chdir` to workspace if conversation has one or fewer user messages.
>   - **Context**:
>     - Fixes regression where `cd` persistence was broken between steps.
>     - Original change was to ensure correct start directory but caused worse issues.
>     - Proper fix for concurrent conversations requires larger refactor.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 9d4ac74a5d469852375045fdfe87d0fcde49d384. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -313,6 +313,7 @@ def step(
     # If we are attempting to make a step in a conversation with only one or fewer user
     # messages, make sure we first chdir to the workspace directory (so that
     # the conversation starts in the right folder).
+    # Tracked in issue #1486
     user_messages = [msg for msg in manager.log.messages if msg.role == "user"]
     if len(user_messages) <= 1:
         logger.debug(


gptme/server/api_v2_sessions.py
```diff
@@ -307,15 +307,18 @@ def step(
...
-    # for single-conversation use and at least starts each step in the right
-    # place for the concurrent case.
-    logger.debug(f"Changing directory to workspace: {workspace}")
-    os.chdir(workspace)
+    # TODO: This is not the best way to manage the chdir state, since it's
+    # essentially a shared global across chats (bad), but the fix at least
+    # addresses the issue where chats don't start in the directory they should.
+    # If we are attempting to make a step in a conversation with only one or fewer user
+    # messages, make sure we first chdir to the workspace directory (so that
+    # the conversation starts in the right folder).
```
ErikBjare, line 315:
```suggestion
    # the conversation starts in the right folder).
    # Tracked in issue #1486
```

PR #1508, opened 2026-02-26
test(shell): add tests for allowlist auto-confirm behavior

ErikBjare:
## Summary

Add comprehensive tests verifying that allowlisted shell commands (like `cat file | head -100`) skip confirmation entirely.

## Problem

User reported being asked to confirm a read-only command (`cat gptme/cli/commands.py | head -100`) which should be auto-approved by default since both `cat` and `head` are in the allowlist.

## Investigation

Testing confirmed the allowlist logic is working correctly:
- `is_allowlisted('cat gptme/cli/commands.py | head -100')` returns `True`
- The `execute_shell` function skips `execute_with_confirmation` for allowlisted commands
- The `shell_allowlist_hook` correctly returns `ConfirmationResult.confirm()`

## Changes

1. **New test file** (`tests/test_shell_allowlist_autoconfirm.py`):
   - 13 tests for `is_allowlisted()` function covering various command patterns
   - 5 tests for `shell_allowlist_hook` function
   - 2 end-to-end tests for `execute_shell` verifying confirmation is skipped

2. **Debug logging** (in `gptme/tools/shell.py`):
   - Added logging when commands are allowlisted and skipping confirmation
   - Added logging when commands are not allowlisted and require confirmation

## Test Results

All 20 tests pass, confirming the allowlist behavior is working as expected.

## Notes

The investigation suggests the original issue may have been:
- An older version that didn't have this working
- A misunderstanding about what was happening
- Some edge case not captured by the tests

The added logging will help diagnose any future issues with the confirmation behavior.

----

> [!IMPORTANT]
> Add tests for allowlisted shell commands to skip confirmation and enhance logging in `gptme/tools/shell.py`.
> 
>   - **Tests**:
>     - Add `tests/test_shell_allowlist_autoconfirm.py` with 13 tests for `is_allowlisted()`, 5 for `shell_allowlist_hook`, and 2 end-to-end tests for `execute_shell`.
>   - **Logging**:
>     - Add debug logging in `gptme/tools/shell.py` to log when commands are allowlisted and skip confirmation, and when they are not and require confirmation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for b364c42a0e03230a4f3e48dba09354c825cf75f6. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1298,7 +1298,18 @@ def _has_file_redirection(cmd: str) -> bool:
 
 
 def is_allowlisted(cmd: str) -> bool:
+    """Check if a shell command is safe to auto-approve.
+
+    Uses a conservative allowlist approach:
+    1. All commands in the pipeline must be in the allowlist
+    2. No file redirections (>, >>) - these can write malicious content
+    3. No dangerous flags within allowlisted commands (e.g., find -exec)
+
+    This means commands like xargs, sh, bash, python, perl, etc. are automatically
+    blocked since they're not in the allowlist, even if piped to from safe commands.
+    """
     # Check if all commands in the pipeline are allowlisted
+    # This blocks non-allowlisted commands like: python, perl, xargs, sh, bash, etc.
     for match in cmd_regex.finditer(cmd):
         for group in match.groups():
             if group and group not in allowlist_commands:
@@ -1310,27 +1321,15 @@ def is_allowlisted(cmd: str) -> bool:
     if _has_file_redirection(cmd):
         return False
 
-    # Check for dangerous flags and patterns that can modify files or execute commands
-    # Note: This is defense-in-depth on top of the allowlist check above
+    # Check for dangerous flags within allowlisted commands
+    # These are rare exceptions where an allowlisted command has dangerous dual-use flags
     dangerous_patterns = [
         "-exec",  # find -exec can execute arbitrary commands
         "-execdir",  # find -execdir can execute arbitrary commands in target dir
         "-delete",  # find -delete can delete files
         "-ok",  # find -ok prompts but can be automated
     ]
-    for pattern in dangerous_patterns:
-        if pattern in cmd:
-            return False
-
-    # xargs, sh, bash can execute arbitrary commands when used in pipelines
-    # Examples: cat file | xargs rm, cat script | sh
-    dangerous_commands = ["xargs", "sh", "bash", "zsh", "fish"]
-    for match in cmd_regex.finditer(cmd):
-        for group in match.groups():
-            if group in dangerous_commands:
-                return False
-
-    return True
+    return all(pattern not in cmd for pattern in dangerous_patterns)
 
 
 def shell_allowlist_hook(
--- a/tests/test_shell_allowlist_autoconfirm.py
+++ b/tests/test_shell_allowlist_autoconfirm.py
@@ -45,23 +45,28 @@ ALLOWLIST_TEST_CASES = [
     ("rm -rf /tmp/foo", False, "rm command"),
     ("python script.py", False, "python command"),
     ("npm install", False, "npm command"),
-    # Dangerous patterns with allowlisted commands - should NOT be allowlisted
-    ("find . -name '*.py' -exec rm {} \\;", False, "find -exec rm"),
-    ("find . -type f -exec cat {} \\;", False, "find -exec cat"),
+    # Pipes to non-allowlisted commands - blocked by allowlist (not in allowlist)
+    ("cat file | xargs rm", False, "pipe to xargs (not in allowlist)"),
+    ("grep pattern file | xargs python", False, "pipe to xargs"),
+    ("cat file | sh", False, "pipe to sh (not in allowlist)"),
+    ("head file | bash", False, "pipe to bash (not in allowlist)"),
+    ("ls | python -c 'import sys'", False, "pipe to python"),
+    ("cat data.csv | perl -lane", False, "pipe to perl"),
+    # Dangerous flags within allowlisted commands - blocked by flag check
+    ("find . -name '*.py' -exec rm {} \\;", False, "find -exec rm (dangerous flag)"),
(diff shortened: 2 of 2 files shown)


tests/test_shell_allowlist_autoconfirm.py
```diff
@@ -0,0 +1,220 @@
...
+        assert is_allowlisted("pwd") is True
+
+    def test_rg_ripgrep(self):
+        """Test that rg (ripgrep) is allowlisted."""
+        assert is_allowlisted("rg pattern") is True
+        assert is_allowlisted("rg pattern file.txt") is True
+
+    def test_find(self):
+        """Test that find is allowlisted."""
+        assert is_allowlisted("find . -name '*.py'") is True
```
ErikBjare: What about "find -exec" examples that should NOT be allowlisted? Do we have tests for that somewhere, we should consolidate all these known positives and known negatives for easy overview, erring on the conservative side.

We don't need one test method per cmd, this PR could be a lot shorter with a helper and wrapping up all the cases up in a single or few tests.

ErikBjare:
Addressed the review feedback:

1. **Added tests for dangerous patterns**: Added comprehensive test coverage for `find -exec` and other command execution patterns that should NOT be allowlisted:
   - `find -exec rm/cat/etc` (can execute arbitrary commands)
   - Piping to `xargs` (can execute commands from stdin)
   - Piping to shells (`sh`, `bash`, `zsh`, `fish`)

2. **Updated allowlist logic**: Modified `is_allowlisted()` to explicitly block:
   - Any command containing `-exec` flag
   - Any pipeline containing dangerous execution commands (`xargs`, `sh`, `bash`, `zsh`, `fish`)

3. **Consolidated tests**: Refactored to use `@pytest.mark.parametrize` with a single test case list containing 30 examples (both positive and negative cases), making it much easier to see all cases at a glance and add new ones.

All tests pass ✓

tests/test_shell_allowlist_autoconfirm.py
ErikBjare: Fixed! Consolidated all test cases into a single parameterized test with 30 cases, and added comprehensive coverage for dangerous patterns like `find -exec`, piping to `xargs`, and piping to shells (`sh`, `bash`, etc.). The allowlist logic now explicitly blocks these patterns.

gptme/tools/shell.py
```diff
@@ -1307,7 +1307,23 @@ def is_allowlisted(cmd: str) -> bool:
...
+    if "-exec" in cmd:
+        return False
+
+    # xargs, sh, bash can execute arbitrary commands when used in pipelines
+    # Examples: cat file | xargs rm, cat script | sh
+    dangerous_commands = ["xargs", "sh", "bash", "zsh", "fish"]
+    for match in cmd_regex.finditer(cmd):
+        for group in match.groups():
+            if group in dangerous_commands:
+                return False
```
ErikBjare: This seems brittle, all non-verified safe read-only pipelines are unsafe.

ErikBjare: Good point about brittleness. I see a few options:

1. **More conservative allowlist**: Only allow specific safe patterns like `find -name/-type` but not all find flags
2. **Stronger validation**: Parse the command structure more carefully (e.g., using bashlex) to verify it's truly read-only
3. **Accept the trade-off**: The current approach catches the most common dangerous patterns (exec, xargs, shell piping) and all commands must still be in the allowlist

My inclination is toward option 1 or 3. Option 2 would be more robust but significantly more complex.

What's your preference? Should we:
- Add more dangerous patterns to the blocklist (e.g., `find -delete`, `-execdir`, etc.)?
- Or switch to a more restrictive approach where we only allow specific known-safe flag combinations?

ErikBjare: I've added more dangerous patterns to the blocklist to make it more comprehensive:
- `-execdir`: executes commands in matched file's directory
- `-delete`: deletes matched files  
- `-ok`: prompts for command execution (can be automated)
- Added test coverage for `zsh` and `fish` shells

This brings us to 35 test cases covering dangerous patterns. The approach is still defense-in-depth on top of the base allowlist check.

Let me know if you'd prefer a more fundamental rethink of the approach (e.g., only allowing specific safe flag combinations rather than blocking known dangerous ones).

ErikBjare:
Fixed the brittleness issue! Removed the redundant `dangerous_commands` blocklist.

## The Problem
The blocklist approach of checking for specific dangerous commands like `xargs`, `sh`, `bash` was redundant and problematic because:
- These commands aren't in the allowlist anyway, so they're already blocked
- It suggested we need to enumerate all possible dangerous commands (python, perl, ruby, node, etc.)

## The Solution
Use a **conservative allowlist approach**:
1. ✅ All commands must be in the allowlist (blocks python, perl, xargs, sh, bash automatically)
2. ✅ Block file redirections (prevents writing malicious content)
3. ✅ Block dangerous flags within allowlisted commands (find -exec, -delete, etc.)

## Test Updates
- Added test cases for python and perl to demonstrate the allowlist blocks them
- Clarified that xargs/sh/bash are blocked by the allowlist, not a separate check
- Organized tests to show the two blocking mechanisms (allowlist vs dangerous flags)

All 35 tests still pass ✓

gptme/tools/shell.py
ErikBjare: Fixed! Removed the `dangerous_commands` blocklist entirely. It was redundant since those commands (xargs, sh, bash, python, perl, etc.) aren't in the allowlist and are already blocked by the base allowlist check.

Now using a clean conservative allowlist approach:
1. All commands must be in the allowlist → blocks xargs/sh/bash/python/perl automatically
2. Block file redirections → prevents writing malicious content  
3. Block dangerous flags within allowlisted commands → handles rare dual-use cases like `find -exec`

No need to enumerate all possible dangerous commands - the allowlist handles that!

PR #1561, opened 2026-02-28
fix(context): use plain git status instead of -vv

TimeToBuildBob:
## Summary

- Replace `git status -vv` with `git status` in `git_status()` context function
- `-vv` dumps full diffs of both staged and unstaged changes into the system prompt, which can produce hundreds of KB of output in workspaces with large files
- Gordon's workspace hit 409KB of JSON diffs → 497KB system prompt → "Prompt is too long" failures blocking autonomous runs for 6+ hours

## Rationale

Plain `git status` provides the file list which is sufficient context for the system prompt. Agents can always read specific diffs themselves when needed. The verbose diff output is rarely useful as context and poses an unbounded size risk.

## Test plan

- [x] Verify `git_status()` returns file list without diffs
- [ ] Confirm context generation works in a repo with large staged/unstaged files

Reported-by: Gordon (agent)

----

> [!IMPORTANT]
> Replace `git status -vv` with `git status` in `git_status()` to prevent excessive output size in system prompt.
> 
>   - **Behavior**:
>     - Replace `git status -vv` with `git status` in `git_status()` in `context.py` to avoid large output size.
>     - Prevents prompt size issues in workspaces with large files, as reported by Gordon.
>   - **Rationale**:
>     - Plain `git status` provides sufficient context without the risk of unbounded size from diffs.
>   - **Test Plan**:
>     - Verify `git_status()` returns file list without diffs.
>     - Confirm context generation in repos with large files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 5556acb8a12b6909e2a0c1c07fae3a91802edf3c. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/prompts.py
+++ b/gptme/prompts.py
@@ -578,7 +578,7 @@ def prompt_workspace(
 ) -> Generator[Message, None, None]:
     """Generate the workspace context prompt."""
     # TODO: update this prompt if the files change
-    # TODO: include `git status -vv`, and keep it up-to-date
+    # TODO: include `git status`, and keep it up-to-date
     sections = []
 
     if workspace is None:
--- a/gptme/util/context.py
+++ b/gptme/util/context.py
@@ -360,7 +360,10 @@ def git_status() -> str | None:
             ["git", "status"], capture_output=True, text=True, check=True
         )
         logger.debug("Including git status in context")
-        return md_codeblock("git status", git_status.stdout)
+        output = git_status.stdout
+        if len(output) > 10000:
+            output = output[:10000] + "\n... (truncated)"
+        return md_codeblock("git status", output)
     except (subprocess.CalledProcessError, FileNotFoundError):
         logger.debug("Not in a git repository or git not available")
     return None


ErikBjare:
This is an improvement, but shouldn't we still impose a maximum upper bound for the command output? (truncating beyond some reasonable limit)

TimeToBuildBob:
Good call — added a 10k char truncation limit as a safety bound in `a253426cc`. Also updated the stale TODO comment in `prompts.py` that still referenced `-vv`.

Re: upstreaming context scripts to gptme-contrib (Gordon's suggestion in his third message) — that's a separate follow-up worth tracking. The agent-template currently has its own copies of context.sh/context-workspace.sh/etc., and they drift. Centralizing in gptme-contrib with symlinks would be cleaner.

PR #1572, opened 2026-03-02
fix(gh): add multi-line comment guidance to prevent literal \n in posts

TimeToBuildBob:
## Summary

When LLMs generate `gh issue comment` commands with `--body "text\nmore text"`, bash does NOT interpret `\n` as a newline in double-quoted strings. This results in literal `\n` characters appearing in GitHub comments.

This was observed in comments posted via gptme with the `openai-subscription` provider (ErikBjare/bob#356).

**Root cause**: LLMs use `\n` as an escape sequence, but in bash double-quoted strings, `\n` is literal backslash-n, not a newline.

**Fix**: Add explicit instructions to use `--body-file` with heredoc for multi-line comment bodies, and add an example showing the correct pattern.

## Changes

- Added guidance in gh tool instructions to use `--body-file` with heredoc
- Added an example for posting multi-line comments
- Added explicit warning about `\n` in double-quoted strings

## Test plan

- [ ] Verify instructions render correctly: `python -c "from gptme.tools.gh import tool; print(tool.instructions)"`
- [ ] Verify example renders correctly: `python -c "from gptme.tools.gh import examples; print(examples('markdown'))"`

Fixes: ErikBjare/bob#356

----

> [!IMPORTANT]
> Fixes issue with literal '\n' in GitHub comments by updating `gh.py` to use `--body-file` with heredoc for multi-line comments.
> 
>   - **Behavior**:
>     - Updated `instructions` in `gh.py` to recommend using `--body-file` with heredoc for multi-line comments to avoid literal `\n`.
>     - Added an example in `examples()` in `gh.py` for posting multi-line comments using heredoc.
>     - Added a warning about `\n` in double-quoted strings in `gh.py`.
>   - **Test Plan**:
>     - Verify instructions render correctly using `python -c "from gptme.tools.gh import tool; print(tool.instructions)"`.
>     - Verify example renders correctly using `python -c "from gptme.tools.gh import examples; print(examples('markdown'))"`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 103387589d8d79b9c3c271c1d007550d74dc7cdb. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/tools/gh.py
+++ b/gptme/tools/gh.py
@@ -439,7 +439,7 @@ To wait for all CI checks to complete:
 The optional commit_sha allows checking a specific commit instead of the PR head.
 This is useful for checking previous commits without waiting for new builds.
 
-For posting multi-line comments, always use `--body-file` with a heredoc to avoid literal \\n in output:
+For posting multi-line comments, always use `--body-file` with a single-quoted heredoc to avoid two bash pitfalls:
 ```shell
 gh issue comment NUM --repo owner/repo --body-file - << 'EOF'
 ## Summary
@@ -448,7 +448,10 @@ Line 1
 Line 2
 EOF
 ```
-Never use `--body "text\\nmore text"` — `\\n` in double-quoted strings is not interpreted as a newline by bash.
+Never use `--body "text\\nmore text"`:
+- `\\n` in double-quoted strings is literal backslash-n, not a newline.
+- `$` in double-quoted strings triggers variable expansion (e.g. `$42,000` → `,000`).
+The single-quoted `<< 'EOF'` heredoc prevents both issues.
 
 For other operations, use the `shell` tool with the `gh` command."""
 


ErikBjare:
This may have been the cause, but are you sure it's not some weirdness with the `--tool-format tool` format in gptme core? (maybe it does something weird with `\` or `\n`? are the examples rendered correctly etc in that tool format?) Strange it has only ever happened with `openai-subscription/gpt-5.3-codex` running with `--tool-format tool`.

TimeToBuildBob:
Good question. I investigated and here's my analysis:

**Root cause is bash, not gptme core.** The shell tool runs `bash -c <cmd>`. Bash doesn't interpret `\n` in double-quoted strings as newlines — it's a literal backslash-n. This affects all tool formats.

**Why only with `--tool-format tool`?** In tool format, the shell command is JSON-encoded/decoded. There are two scenarios:
- If the model generates `\n` in the JSON string (proper JSON escape) → `json_repair.loads()` decodes it as an actual newline → `bash -c` receives a multi-line string → works correctly
- If the model generates `\\n` in JSON (meaning literal backslash-n) → decoded as `\n` → bash sees literal `\n` → bug

My guess: `gpt-5.3-codex` generates `\\n` (literal backslash-n) in its tool call JSON when asked to include newlines in `--body "..."` strings. Other models may generate actual `\n` (JSON newline escape) or avoid multi-line `--body` entirely.

**gptme core doesn't do anything weird with `\` or `\n`** — it's straightforward `json_repair.loads()` → `bash -c cmd`.

**Examples render correctly in tool format.** The new heredoc example in `examples()` uses actual newlines in the Python string, which `json.dumps()` correctly encodes as `\n` in the JSON. The LLM sees `\n` in the JSON and generates similar JSON with `\n` → decodes to actual newlines → the heredoc command with real newlines executes correctly in bash.

So the fix is valid and the right approach regardless of tool format — teaching the model to use `--body-file` with heredoc sidesteps the bash escaping issue entirely.

gptme/tools/gh.py
```diff
@@ -439,6 +439,17 @@ def execute_gh(
 The optional commit_sha allows checking a specific commit instead of the PR head.
 This is useful for checking previous commits without waiting for new builds.
 
+For posting multi-line comments, always use `--body-file` with a heredoc to avoid literal \\n in output:
```
ErikBjare: Would this also address the error case where it tries to `--body "String with $42,000"` which becomes "String with ,000" or similar? May be worth mentioning briefly if we have the HereDoc rule anyway.

TimeToBuildBob: Yes! The single-quoted heredoc (`<< 'EOF'`) prevents both issues — quoting disables all bash substitutions inside the heredoc. Updated the instructions to explicitly call out both pitfalls.

PR #1585, opened 2026-03-03
fix: use diff syntax highlighting for save/append previews

ErikBjare:
When saving/appending to an existing file, the preview shows a diff but was rendered as plain text (white, no coloring). Patch tool already showed colored diffs.

This detects when the preview content looks like a unified diff and applies `"diff"` syntax highlighting in the CLI confirmation hook, giving save/append the same colored diff preview as patch.

----

> [!IMPORTANT]
> Enhances save/append previews with diff syntax highlighting by detecting diff-like content in `cli_confirm.py`.
> 
>   - **Behavior**:
>     - Detects unified diff-like content in save/append previews and applies `"diff"` syntax highlighting in `cli_confirm_hook()`.
>     - Ensures save/append previews have the same colored diff preview as the patch tool.
>   - **Functions**:
>     - Adds `_looks_like_diff()` to check if content resembles a unified diff.
>     - Modifies `_get_lang_for_tool()` to use `_looks_like_diff()` for determining syntax highlighting.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for cc288456ca7049a12c39b4751e27d1937e991498. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/gptme/hooks/cli_confirm.py
+++ b/gptme/hooks/cli_confirm.py
@@ -208,7 +208,25 @@ def _looks_like_diff(content: str | None) -> bool:
 
     # Avoid false positives on plain markdown/yaml lists (mostly '-' lines).
     # Real diff previews usually have either mixed +/- lines or context lines.
-    return (has_plus and has_minus) or has_context
+    if (has_plus and has_minus) or has_context:
+        return True
+
+    # Accept plus-only diffs (e.g. append to empty file produces all-'+' output
+    # from Patch.diff_minimal). Require every non-empty line to start with '+'
+    # and at least one non-"+ " line to avoid matching markdown/yaml '+' lists.
+    if has_plus and not has_minus:
+        non_empty = [line for line in lines if line.strip()]
+        all_plus = non_empty and all(line.startswith("+") for line in non_empty)
+        # Markdown '+ item' lists always have a space after '+'.  Real diffs from
+        # diff_minimal() use '+content' (no space) for most lines.
+        has_tight_plus = any(
+            line.startswith("+") and (len(line) == 1 or line[1] != " ")
+            for line in non_empty
+        )
+        if all_plus and has_tight_plus:
+            return True
+
+    return False
 
 
 def _get_lang_for_tool(tool: str, content: str | None = None) -> str:
--- a/tests/test_cli_confirm_hook.py
+++ b/tests/test_cli_confirm_hook.py
@@ -46,3 +46,33 @@ def test_get_lang_for_save_uses_diff_when_preview_is_diff() -> None:
 def test_get_lang_for_save_plain_text_fallback() -> None:
     content = "# Notes\n\n- Apples\n- Bananas\n"
     assert _get_lang_for_tool("save", content) == "text"
+
+
+def test_looks_like_diff_plus_only_append() -> None:
+    """Plus-only diffs (append to empty file) should be detected."""
+    content = "+line1\n+line2\n+line3\n"
+    assert _looks_like_diff(content) is True
+
+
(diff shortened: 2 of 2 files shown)


gptme/hooks/cli_confirm.py
```diff
@@ -177,8 +177,32 @@ def _handle_response(
...
+        return False
+    # Must have at least one +/- line (actual changes)
+    has_changes = any(line.startswith(("+", "-")) for line in lines)
+    if not has_changes:
+        return False
+    # Most lines should look like diff lines (context, changes, or hunk headers)
+    diff_lines = sum(
+        1 for line in lines if line.startswith((" ", "+", "-", "@")) or line == ""
+    )
+    return diff_lines >= len(lines) * 0.8
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 2ea9508 — added the `@@` hunk header check as suggested. This eliminates false positives for Markdown/YAML bullet lists.

gptme/hooks/cli_confirm.py
```diff
@@ -188,6 +188,10 @@ def _looks_like_diff(content: str | None) -> bool:
     has_changes = any(line.startswith(("+", "-")) for line in lines)
     if not has_changes:
         return False
+    # A unified diff always contains at least one @@ hunk header
+    has_hunk = any(line.startswith("@@") for line in lines)
```
ErikBjare: I don't think the diff-output from `save` etc has the hunk header

TimeToBuildBob: Good catch — you were right. save/append previews come from Patch.diff_minimal(), so they often omit @@/---/+++ headers. I updated detection to handle minimal diff bodies instead: require mostly diff-like lines plus either (+ and -) or context lines, which keeps bullet-list false positives low without requiring hunk headers. Fixed in afceba709.

TimeToBuildBob:
Addressed follow-up on diff detection in save/append previews.

### What changed
- Updated `_looks_like_diff()` to support minimal diff bodies emitted by `Patch.diff_minimal()` (no required `@@` hunk headers).
- Added guard rails to avoid markdown/yaml list false positives by requiring either:
  - mixed `+` and `-` lines, or
  - at least one context (` `) line.

### Validation
- Added tests in `tests/test_cli_confirm_hook.py` for:
  - minimal diff body detection
  - mixed +/- detection
  - markdown list rejection
  - plus-only list rejection
  - `_get_lang_for_tool()` diff vs text behavior
- Ran:
  - `PYTHONPATH=/home/bob/gptme pytest -q tests/test_cli_confirm_hook.py tests/test_tools_save.py`
  - Result: **14 passed**

ErikBjare:
@greptileai review

ErikBjare:
append-only diffs wouldn't show properly?

TimeToBuildBob:
Good catch — append-only diffs (e.g. appending to an empty file) previously didn't highlight as diff because `_looks_like_diff()` required either mixed +/- lines or context lines.

Fixed in 18a3d6fe2: now accepts plus-only content when every non-empty line starts with `+` and at least one has no space after `+` (to distinguish from markdown `+ item` lists).

**Coverage**:
- Append to non-empty file → context + additions → already worked ✅
- Append to empty file → all-`+` lines → now detected ✅  
- New file save → raw content (no diff markers) → correctly stays as `text` ✅

Also added test coverage for the `append` tool path per Greptile's review finding (5 new tests, all passing).

PR #1599, opened 2026-03-05
Add Nightmarket tool — API marketplace for AI agents

streacy:
## What this adds

A new `nightmarket` tool in `gptme/tools/nightmarket.py` that lets gptme agents discover and call paid third-party API services through the [Nightmarket](https://nightmarket.ai) marketplace.

### Commands

- **`search <query>`** — Search the marketplace for available API services
- **`details <endpoint_id>`** — Get full service details including request/response examples
- **`call <endpoint_id> <method> [body] [payment_sig]`** — Call a service with optional x402 payment proof

### How it works

Nightmarket is an API marketplace where AI agents find and pay for third-party services. Every call settles on-chain in USDC on Base using the [x402 payment protocol](https://www.x402.org/) — no API keys or subscriptions needed.

The tool registers as a `ToolSpec` with block type `nightmarket`, following gptme's existing tool conventions. Uses only stdlib (`urllib`, `json`) — no additional dependencies.

### Usage

````
```nightmarket
search weather
```

```nightmarket
call abc123 GET
```
````

First call returns 402 Payment Required. Use CrowPay to handle payment, then retry with the payment signature.

### Why this is useful for gptme

gptme agents can use this to autonomously access any external API on the marketplace without needing API keys for each service. Search, pay per-call, get the response — all inline.

### DIFF
--- a/gptme/tools/nightmarket.py
+++ b/gptme/tools/nightmarket.py
@@ -19,7 +19,10 @@ NIGHTMARKET_BASE_URL = "https://nightmarket.ai/api"
 instructions = """
 ## Nightmarket — API Marketplace
 
-Use the `nightmarket` tool to discover and call paid third-party API services.
+You can extend your capabilities on demand by discovering and calling any
+third-party API on Nightmarket — no API keys or pre-configuration required.
+Use `search` to find services, `details` to understand their inputs, and
+`call` to invoke them inline and return results directly to the conversation.
 
 **Search for services:**
 ```nightmarket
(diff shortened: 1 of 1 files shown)


streacy:
Addressed all review feedback:

- **User confirmation**: Added `confirm()` callback before any API calls, noting whether payment is involved
- **JSON body parsing**: Switched from `split(maxsplit=4)` to newline-delimited format — body goes on its own line, so JSON with spaces works correctly
- **Disabled by default**: Set `disabled_by_default=True` so users must explicitly opt-in
- **Instructions reframed**: Now agent-benefit-oriented rather than documentation-style
- **HTTP error handling**: `_http_get` now catches `HTTPError` and returns structured error data instead of raising
- **Exception logging**: Added `logger.exception()` for traceback visibility

Thanks for the thorough review — all good catches!

PR #1605, opened 2026-03-06
fix(util): fix parameter shadowing and add empty-list guards

TimeToBuildBob:
## Summary

- **export.py**: Rename `html` parameter to `content` in `replace_or_fail()` to avoid shadowing the `html` module import (used for `html.escape()` on line 37)
- **cost.py**: Add early-return guards for empty message lists in `_tokens_inout()` and `_cost()` to prevent `IndexError` on `msgs[-1]`

## Test plan

- [x] All 27 cost tests pass
- [x] mypy typecheck passes on both files
- [x] No behavioral change — purely defensive improvements

### DIFF
--- a/gptme/util/cost.py
+++ b/gptme/util/cost.py
@@ -3,6 +3,9 @@ from . import console
 
 
 def _tokens_inout(msgs: list[Message]) -> tuple[int, int]:
+    if not msgs:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1613, opened 2026-03-06
feat(doctor): add lesson health checks to agent doctor

TimeToBuildBob:
## Summary

Adds lesson system health checks to the agent workspace doctor (`gptme-agent doctor`).

The new `check_lessons()` function:
- Scans lesson directories from `gptme.toml` `[lessons] dirs` config (or default `lessons/`)
- Counts total lessons across all configured directories
- Detects lessons missing YAML frontmatter
- Detects lessons without matching config (no keywords, name, or globs — won't auto-match)
- Warns about oversized lessons (>100 lines, per the two-file architecture guideline)
- Reports YAML parse errors
- Accepts both lesson format (match.keywords) and skill format (name/description)

## Test plan

- [x] 9 new tests covering: no dir, empty dir, valid lessons, missing frontmatter, missing keywords, oversized, skill format, configured dirs, all-healthy
- [x] All 34 agent doctor tests pass
- [x] All 37 CLI doctor tests pass
- [x] Pre-commit (ruff, mypy, formatting) all green

### DIFF
--- a/gptme/agent/doctor.py
+++ b/gptme/agent/doctor.py
@@ -364,18 +364,23 @@ def check_lessons(workspace: Path, report: DoctorReport) -> None:
         report.warn("Lessons", "no lesson directories found")
         return
 
-    # Scan all lesson files
+    # Scan all lesson files (track seen files to avoid double-counting overlapping dirs)
     total = 0
     no_frontmatter = 0
     no_keywords = 0
     oversized = []
     parse_errors = []
+    seen_files: set[Path] = set()
 
     for lesson_dir in lesson_dirs:
         for md_file in sorted(lesson_dir.rglob("*.md")):
-            # Skip READMEs and non-lesson files
+            # Skip READMEs and already-counted files (overlapping dirs)
             if md_file.name.lower() == "readme.md":
                 continue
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

ErikBjare (commented):
There is already the `gptme-util skills` which should probably have a `gptme-util skills check` (or `validate`, or `doctor`) subcommand that shares this helper function.

TimeToBuildBob:
Addressed in dce1f66:

**Erik's feedback**: Added `gptme-util skills check` subcommand that reuses `check_lessons()` from `doctor.py` directly — so both share the same logic without duplication.

**Greptile bugs fixed**:
1. **Double-counting overlapping dirs**: Added a `seen_files: set[Path]` that tracks resolved file paths — files in overlapping dirs (e.g. `lessons/` + `lessons/tools/`) are now counted only once.
2. **Empty frontmatter silently passing**: Changed `if fm:` → `if fm is not None:` and added an `else: no_keywords += 1` branch, so files with `---\n---` are correctly flagged as missing match config.

Also added 2 regression tests covering both bugs (36 total, all green).

PR #1622, opened 2026-03-07
feat(cli): add `gptme-util chats rename` command

TimeToBuildBob:
## Summary

- Adds `gptme-util chats rename <id> <name>` to rename a conversation's display name
- Updates `config.toml` in the conversation directory — no file moves needed
- Conversation ID stays the same, only the display name changes
- `rename_conversation()` function added to `logmanager.py`

## Usage

```bash
# List conversations to find the ID
gptme-util chats list

# Rename a conversation
gptme-util chats rename 2026-03-07-my-session "Better Name"
```

## Test plan

- [x] 6 unit tests covering: success, not-found, config preservation, idempotency, CLI success/error
- [x] All existing chats tests still pass (24 passed)
- [x] ruff lint and format clean

### DIFF
--- a/tests/test_chats_rename.py
+++ b/tests/test_chats_rename.py
@@ -98,6 +98,7 @@ def test_rename_conversation_preserves_existing_config(tmp_path: Path):
         saved = tomlkit.load(f).unwrap()
     assert saved["chat"]["name"] == "New Name"
     assert saved["chat"]["stream"] is True  # original field must survive
+    assert not (conv_dir / "workspace").exists()  # must not create spurious symlink
 
 
 def test_rename_conversation_idempotent(tmp_path: Path):


TimeToBuildBob:
@greptileai review

ErikBjare:
This is also very related to https://github.com/gptme/gptme/blob/master/gptme/commands/session.py#L246 which should probably be reviewed together (and consider using shared helper functions). I notice it checks for no spaces in chat name (which is actually valuable, the no-space check was before we had the conversation id/name distinction)

TimeToBuildBob:
## Fixes pushed (26e3051)

**Greptile bug fixed** — `ChatConfig.save()` was indeed creating a spurious `workspace -> cwd` symlink on any conversation without a pre-existing workspace config. Fixed `rename_conversation()` to do a targeted TOML update (read name → set name → write) rather than going through `ChatConfig.save()`, so the workspace symlink logic never triggers. Added `test_rename_conversation_no_workspace_symlink` to catch regressions.

**Test strengthened** — `test_rename_conversation_preserves_existing_config` now also asserts `stream` is preserved (not just `name`).

---

**Re: `commands/session.py#L246`** — Good catch. The `_rename()` function there does the same `ChatConfig.from_logdir` + `save()` pattern and has the same workspace symlink risk. It could be refactored to reuse `rename_conversation()` from logmanager (or the underlying targeted-TOML helper). Happy to do that in this PR or as a follow-up — what's your preference?

**Re: no-space check** — The check in `_rename` applies only to the auto-generated LLM name, not user-provided input. Since display names are now distinct from conversation IDs, I think spaces in user-provided names should be valid (a user could reasonably name a chat "My Project 2024"). Let me know if you'd prefer to enforce no-spaces here too.

gptme/logmanager.py
```diff
@@ -814,10 +816,25 @@ def rename_conversation(conv_id: str, new_name: str) -> bool:
...
+    if "chat" not in config_data:
+        config_data.add("chat", tomlkit.table())
+    config_data["chat"]["name"] = new_name  # type: ignore[index]
+
+    with open(config_path, "w") as f:
+        tomlkit.dump(config_data, f)
 
-    chat_config = ChatConfig.from_logdir(conv_dir)
-    chat_config.name = new_name
-    chat_config.save()
```
ErikBjare: But this is a lot cleaner... Why can't we make this work?

tests/test_chats_rename.py
```diff
@@ -93,10 +93,11 @@ def test_rename_conversation_preserves_existing_config(tmp_path: Path):
 
     assert result is True
 
-    # Re-read and verify
+    # Re-read and verify name changed, other fields preserved
     with open(conv_dir / "config.toml") as f:
         saved = tomlkit.load(f).unwrap()
     assert saved["chat"]["name"] == "New Name"
+    assert saved["chat"]["stream"] is True  # original field must survive
```
ErikBjare, line 100: Wouldn't it be better to test the workspace here? Given that was the risk?

gptme/logmanager.py
TimeToBuildBob: Good question — the targeted TOML approach is intentional. The issue is that `ChatConfig.workspace` defaults to `Path.cwd()` (via `default_factory=Path.cwd`), so `save()` always sees `self.workspace != logdir/workspace` and creates the symlink. To make `save()` safe to call here, we'd need to change `workspace` to `Optional[Path] = None` and guard the symlink logic — that's a broader change touching more callers.

The targeted TOML approach keeps the fix minimal and localized: just read the file, set `name`, write it back. No risk of touching workspace, stream, or any other field. Feels like the right scope for a rename operation.

tests/test_chats_rename.py
TimeToBuildBob: Good call — added workspace assertion to `test_rename_conversation_preserves_existing_config` in 7b91ad3:

```python
assert not (conv_dir / "workspace").exists()  # must not create spurious symlink
```

That test now covers both directions: config fields preserved AND no workspace side-effect.

PR #1650, opened 2026-03-10
fix(tools): pass tool_format to all .to_output() calls in examples

TimeToBuildBob:
## Summary

- Fixed 19 `ToolUse.to_output()` calls across 4 tool files that were missing the `tool_format` parameter
- These calls are inside `examples(tool_format)` functions that already receive the format but weren't passing it through to all output rendering
- Affected tools: **python**, **browser**, **shell**, **tmux**

## Problem

When using XML or native tool format (instead of default markdown), tool examples rendered commands correctly but their simulated system outputs (result, stdout, output, URL content) rendered as markdown. This produced mixed-format examples that could confuse the LLM about expected output format.

For example, in XML mode the ipython command would render as `<tool:ipython>...</tool>` but the result would still render as `` ```result ... ``` `` (markdown).

## Changes

| File | Fixes |
|------|-------|
| `python.py` | 2 `result` outputs |
| `browser.py` | 8 outputs (URLs, results) |
| `shell.py` | 6 outputs (shell, stdout) |
| `tmux.py` | 3 `output` blocks |

All changes follow the same pattern: `.to_output()` → `.to_output(tool_format)`.

The one `.to_output()` in `tmux.py:274` (inside `inspect_pane()`, runtime code) is intentionally left unchanged — it's not in an examples function.

## Test plan

- [ ] Existing `test_no_tooluse_repr_in_examples` continues to pass
- [ ] Manually verify examples render correctly: `gptme --tool-format xml --show-hidden '/exit'`

### DIFF
--- a/gptme/util/context.py
+++ b/gptme/util/context.py
@@ -18,7 +18,6 @@ from ..config import get_config
 from ..constants import CONTENT_SIZE_INFO_THRESHOLD, CONTENT_SIZE_WARN_THRESHOLD
 from ..message import Message
 from ..tools import has_tool
-from ..tools.browser import read_url
 from .gh import (
     get_github_issue_content,
     get_github_pr_content,
@@ -738,6 +737,10 @@ def _resource_to_codeblock(
         # If GitHub handling failed or not a GitHub issue/PR, fall back to browser
         if not content and has_tool("browser"):
             try:
+                from ..tools.browser import (
+                    read_url,  # deferred to avoid circular import
+                )
+
                 # Transform GitHub blob URLs to raw URLs
                 transformed_url = transform_github_url(url)
                 if transformed_url != url:


TimeToBuildBob:
@greptileai review

gptme/tools/browser.py
```diff
@@ -325,53 +325,53 @@ def examples(tool_format):
 Assistant: Let's browse the site.
 {ToolUse("ipython", [], "read_url('https://superuserlabs.org/')").to_output(tool_format)}
 System:
-{ToolUse("https://superuserlabs.org/", [], "... [ActivityWatch](https://activitywatch.net/) ...".strip()).to_output()}
+{ToolUse("https://superuserlabs.org/", [], "... [ActivityWatch](https://activitywatch.net/) ...".strip()).to_output(tool_format)}
 Assistant: Couldn't find the answer on the page. Following link to the ActivityWatch website.
 {ToolUse("ipython", [], "read_url('https://activitywatch.net/')").to_output(tool_format)}
 System:
-{ToolUse("https://activitywatch.net/", [], "... Download latest version v0.12.2 ...".strip()).to_output()}
+{ToolUse("https://activitywatch.net/", [], "... Download latest version v0.12.2 ...".strip()).to_output(tool_format)}
```
ErikBjare: But this is an output, not a proper tooluse! I'm realizing now that it was probably always misguided. Should prob use plain markdown output using eg md_codeblock.

ErikBjare (changes requested):
See my comment, many such cases.

gptme/tools/browser.py
TimeToBuildBob: Good catch — fixed in 867de32. Replaced all output ToolUse calls (result, stdout, output, URL content) in examples with md_codeblock across all 4 affected files, which is what the runtime code already produces. Also added the missing md_codeblock import to browser.py.

TimeToBuildBob: Correction to my previous reply: the md_codeblock import I added to browser.py caused a circular import (util.context imports read_url from browser at module level). Fixed in fd94ee8 by making the read_url import deferred (inside the function that uses it in util/context.py) instead. CI should be green now.

TimeToBuildBob:
## All feedback addressed — ready for re-review

Erik's inline comment has been addressed:
- **Replaced all output ToolUse calls with md_codeblock** across python.py, browser.py, shell.py, and tmux.py — as suggested, using plain markdown output is cleaner than wrapping system outputs in ToolUse
- **Fixed circular import** introduced by adding md_codeblock to browser.py: made the read_url import deferred in util/context.py (fd94ee8)

CI is green, Greptile 5/5 confidence.

ErikBjare (approved):

PR #1654, opened 2026-03-11
feat(eval): add practical4 suite — group-by, schedule-overlaps, topo-sort

TimeToBuildBob:
## Summary

Adds `practical4` eval suite with three new tasks covering analytics, scheduling, and algorithms:

- **group-by**: read `sales.json`, aggregate total sales amounts by category, print sorted results. Tests JSON ingestion + grouping logic.
- **schedule-overlaps**: read `events.json`, detect overlapping time intervals, report conflicting pairs. Tests datetime parsing + interval overlap algorithm.
- **topo-sort**: read `deps.json` (task→prerequisites map), output a valid topological execution order. Tests graph algorithms + cycle detection.

All three use stdlib-only solutions and have deterministic check functions that verify both correctness and edge cases.

## Design

Same pattern as `practical`, `practical2`, and `practical3`:
- Provide input files (JSON) with specific, known values
- Agent writes a Python script that processes them
- `run` command executes the script and captures stdout
- Check functions verify exact output values

The `topo-sort` check is noteworthy: `check_topo_order_valid` verifies the topological constraint (each prerequisite appears before its dependent) for all possible valid orderings, not just one specific ordering.

## Checks

| Task | Checks |
|------|--------|
| group-by | file exists, Electronics=1200, Clothing=550, Food=180, exit 0 |
| schedule-overlaps | file exists, Meeting A detected, Meeting B detected, exit 0 |
| topo-sort | file exists, all 6 tasks present, valid topological order, exit 0 |

## Relation to open PRs

This PR follows the same approach as #1647 (practical), #1652 (practical2), and #1653 (practical3). The `__init__.py` change here adds `practical4` alongside `practical` (not replacing it), to make merging cleaner when all four are eventually integrated.

PR #1663, opened 2026-03-13
fix(eval): rename duplicate write-tests to avoid name collision

TimeToBuildBob:
## Summary
- Fixes a test name collision: both `basic.py` and `practical3.py` registered a test named `write-tests`, causing the second to silently overwrite the first in `tests_map`.
- Renamed the `practical3` variant to `write-tests-calculator` to disambiguate.
- Verified: 33 tests, 33 unique names after the fix.

## Context
Both PRs #1647 and #1653 added `write-tests` tests independently. When merged, `tests_map` (a dict keyed by name) kept only the last one.

Note: The `has_avg_time` false-negative flagged by Greptile on #1654 was already fixed during PR #1647's review cycle (commit cefa30499, squash-merged as 120edd060).

## Test plan
- [x] `ruff check` and `ruff format --check` pass
- [x] Verified no duplicate names: `from gptme.eval.suites import suites; names = [t['name'] for s in suites.values() for t in s]; assert len(names) == len(set(names))`

### DIFF
--- a/gptme/eval/suites/__init__.py
+++ b/gptme/eval/suites/__init__.py
@@ -6,6 +6,7 @@ from .practical import tests as tests_practical
 from .practical2 import tests as tests_practical2
 from .practical3 import tests as tests_practical3
 from .practical4 import tests as tests_practical4
+from .practical5 import tests as tests_practical5
 
 suites: dict[str, list[EvalSpec]] = {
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1664, opened 2026-03-13
fix(server): guard acp_runtime None in _acp_step finally block

TimeToBuildBob:
## Summary
- Guard `session.acp_runtime.set_on_update(None)` in the `_acp_step` finally block against `None`
- Concurrent session cleanup can set `acp_runtime = None` while the background step thread is still running, causing `AttributeError: 'NoneType' has no attribute 'set_on_update'`
- Matches the existing None-check pattern already used at lines 1221/1225

## Test plan
- [x] Verified the fix matches existing None-guard pattern in the same file
- [x] Pre-commit checks pass
- [ ] CI green

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -516,6 +516,7 @@ async def _acp_step(
     assert session.acp_runtime is not None, (
         "acp_runtime must be set before calling _acp_step"
     )
+    acp_runtime = session.acp_runtime  # snapshot to avoid TOCTOU races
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's feedback: snapshotted `acp_runtime` into a local variable right after the assert, closing all TOCTOU windows at once. The finally block can now call `acp_runtime.set_on_update(None)` unconditionally — no None guard needed since the local reference outlives any concurrent `session.acp_runtime = None` assignment.

PR #1669, opened 2026-03-13
fix(types): resolve mypy errors in chats tool and gh tests

TimeToBuildBob:
## Summary
- **chats.py**: Fix `str-bytes-safe` mypy error in regex highlight lambda — `m.group(0)` in an f-string triggers the check because `re.Match.group()` is typed as potentially returning `bytes`. Wrapped with `str()` to satisfy the type checker while preserving behavior (input is always `str`).
- **test_util_gh.py**: Add `assert ... is not None` after `pytest.skip()` calls to help mypy narrow types. `pytest.skip()` raises but mypy doesn't always track it as `NoReturn`.

Resolves 14 of the 191 mypy errors in `make typecheck` (all non-import-stub errors).

## Test plan
- [x] `uv run mypy tests/test_util_gh.py gptme/tools/chats.py` — clean (no errors beyond missing import stubs)
- [ ] CI passes

### DIFF
--- a/gptme/tools/chats.py
+++ b/gptme/tools/chats.py
@@ -177,7 +177,7 @@ def _format_message_with_context(
         highlighted = f"{prefix}{context}{suffix}"
         highlighted = re.sub(
             re.escape(query),
-            lambda m: f"\033[1;31m{m.group(0)}\033[0m",
+            lambda m: "\033[1;31m" + str(m.group()) + "\033[0m",
             highlighted,
             flags=re.IGNORECASE,
         )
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
The one CI failure ( timeout) appears to be an unrelated flaky test — the PR only touches `gptme/tools/chats.py` and `tests/test_util_gh.py`, neither of which is involved in that test. Could you re-trigger CI or confirm this is a known flaky test?

TimeToBuildBob:
@greptileai review

PR #1673, opened 2026-03-14
fix(tests): replace pytest-mock mocker fixture with unittest.mock

TimeToBuildBob:
## Summary

- Replace `mocker` fixture (from `pytest-mock`) with `unittest.mock.patch` in `test_create_subagent_thread_warns_on_unknown_profile_tools`
- `pytest-mock` is declared in `pyproject.toml` but not resolved in `uv.lock`, causing a fixture-not-found ERROR when running tests via `uv run pytest`
- Other tests in the same file already use `unittest.mock` directly — this makes the pattern consistent

## Test plan

- [x] `uv run pytest tests/test_tools_subagent.py::test_create_subagent_thread_warns_on_unknown_profile_tools` passes
- [ ] CI passes

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -260,6 +260,7 @@ def test_shell_file(args: list[str], runner: CliRunner):
     )
 
 
+@pytest.mark.slow
 def test_python(args: list[str], runner: CliRunner):
     args.append("/py print('yes')")
     result = runner.invoke(cli.main, args)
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Cherry-picked the slow marker fix from #1671 (`7fe565a35`) to resolve the `test_python_error` timeout failure in CI.

The timeout was unrelated to this PR's changes — IPython initialization takes ~22s, exceeding the 10s timeout used in the "Test without API keys or extras" job. Both `test_python` and `test_python_error` now carry `@pytest.mark.slow` so they're excluded from that job.

PR #1681, opened 2026-03-15
fix(server): replace assertions with explicit error checks in production code

TimeToBuildBob:
## Summary

- Replace 3 `assert` statements in `api_v2_sessions.py` with explicit error checks that raise proper HTTP errors or handle conditions gracefully. Python's `-O` flag disables assertions, making them unreliable for production error handling (see existing pattern at `api.py` line 299).
- Add `logger.debug()` to silent `except Exception: pass` handlers in `acp_session_runtime.py` and `openapi_docs.py` so errors are at least visible when debug logging is enabled.
- Replace a bare `print()` warning in `openapi_docs.py` with `logger.warning()` for consistent logging.

### Changes by file

**`api_v2_sessions.py`**:
- `assert session.acp_runtime is not None` -> explicit None check with error event + early return
- `assert final_msg is not None` -> explicit None check with warning log + error event
- `assert model is not None` -> explicit None check returning 400

**`acp_session_runtime.py`**:
- `except Exception: pass` in `terminate_subprocess_sync()` -> `logger.debug()` with exc_info

**`openapi_docs.py`**:
- Added `logging` import and `logger` instance
- 3 silent `except Exception` handlers -> `logger.debug()` with exc_info
- `print(f"Warning: ...")` -> `logger.warning()` with exc_info

## Test plan
- [x] `pytest tests/test_server.py` passes (16 passed, 1 xfailed)
- [x] `mypy` clean on all changed files (only pre-existing import-not-found errors from missing stubs)
- [x] `ruff format` applied via pre-commit

### DIFF
--- a/gptme/server/acp_session_runtime.py
+++ b/gptme/server/acp_session_runtime.py
@@ -199,7 +199,9 @@ class AcpSessionRuntime:
                 proc.kill()
                 proc.wait(timeout=1.0)  # reap zombie after SIGKILL
             except Exception:
-                pass
+                logger.debug(
+                    "Failed to kill ACP subprocess after SIGTERM timeout", exc_info=True
+                )
         finally:
             # Clear references so the runtime is left in a consistent state.
             # Mirrors what close() does asynchronously; prevents accidental reuse.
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #1683, opened 2026-03-16
feat(chat): add GPTME_MAX_STEPS and GPTME_CONTEXT_LENGTH env vars

ErikBjare:
## Summary

- **`GPTME_MAX_STEPS`**: Limits the number of response-action steps per prompt. When the limit is reached, a system message is appended and the turn ends. Useful for bounding autonomous runs (e.g. refactoring tasks with local inference).
- **`GPTME_CONTEXT_LENGTH`**: Overrides the model's context window size (in tokens). Applied after model init so it takes precedence. Useful for local models served with non-standard context lengths (e.g. 32k with llama.cpp).

Both are env-var-only to keep the CLI surface clean. Usage:

```bash
# Limit to 10 steps per prompt
GPTME_MAX_STEPS=10 gptme -n 'refactor the auth module'

# Override context length for a local 32k model
GPTME_CONTEXT_LENGTH=32000 gptme -m local/llama-3-8b 'explain this code'
```

Requested by Markus Borg's PhD student for research refactoring runs.

## Test plan
- [x] Existing tests pass
- [x] Pre-commit hooks pass (ruff, mypy, etc.)
- [ ] CI passes

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -288,6 +288,16 @@ def _process_message_conversation(
 
     Note: Confirmation is now handled within ToolUse.execute() using the hook system.
     """
+    max_steps: int | None = None
+    max_steps_str = os.environ.get("GPTME_MAX_STEPS")
+    if max_steps_str:
+        try:
+            max_steps = int(max_steps_str)
+        except ValueError:
+            logger.warning(
+                f"Invalid GPTME_MAX_STEPS value: {max_steps_str!r}, ignoring"
+            )
+    step_count = 0
 
     while True:
         try:
@@ -349,6 +359,15 @@ def _process_message_conversation(
                     daemon=True,
                 )
                 thread.start()
+
+        # Check step limit (GPTME_MAX_STEPS)
+        step_count += 1
+        if max_steps is not None and step_count >= max_steps:
+            console.log(f"Reached max steps limit ({max_steps}), stopping.")
+            manager.append(
+                Message("system", f"Stopped: reached max steps limit ({max_steps})")
+            )
+            break
 
         # Check if there are any runnable tools left
         last_content = next(
--- a/gptme/init.py
+++ b/gptme/init.py
@@ -1,5 +1,7 @@
 import atexit
 import logging
+import os
+from dataclasses import replace
 from typing import cast
 
 from dotenv import load_dotenv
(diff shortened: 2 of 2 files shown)


gptme/chat.py
```diff
@@ -96,6 +96,19 @@ def chat(
...
     init(model, interactive, tool_allowlist, tool_format, no_confirm)
 
+    # Apply GPTME_CONTEXT_LENGTH override (useful for local models with non-standard context)
+    context_length_str = os.environ.get("GPTME_CONTEXT_LENGTH")
+    if context_length_str:
+        from dataclasses import replace
+
+        from .llm.models import set_default_model
+
+        context_length = int(context_length_str)
```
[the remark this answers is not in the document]
ErikBjare: Shouldn't this whole thing happen in init? and not chat?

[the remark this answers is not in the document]
ErikBjare: Fixed in ffd4a37 — added try/except ValueError for both env vars, warns and ignores malformed values.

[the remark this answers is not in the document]
ErikBjare: Good call — moved to init_model() in ffd4a37.

gptme/init.py
```diff
@@ -117,6 +120,21 @@ def init_model(
...
+        except ValueError:
+            logger.warning(
+                f"Invalid GPTME_CONTEXT_LENGTH value: {context_length_str!r}, ignoring"
+            )
+        else:
+            current = get_default_model()
+            if current:
+                set_default_model(replace(current, context=context_length))
+                logger.info(f"Context length overridden to {context_length} tokens")
+
```
ErikBjare: set_default_model twice?

ErikBjare: Fixed in 989bc76 — now resolves ModelMeta first, applies the override, then calls set_default_model once.

ErikBjare: Squashed. The get_model() call is safe here — it's a stateless lookup on the static MODELS dict, same thing set_default_model(str) was doing internally. Now it's explicit so we can replace() before the single set_default_model call.

gptme/chat.py
```diff
@@ -288,6 +288,16 @@ def _process_message_conversation(
 
     Note: Confirmation is now handled within ToolUse.execute() using the hook system.
     """
+    max_steps: int | None = None
+    max_steps_str = os.environ.get("GPTME_MAX_STEPS")
```
ErikBjare, line 292: Shouldn't this use `config.get_env`?

gptme/init.py
```diff
@@ -115,7 +117,23 @@ def init_model(
     model_full = f"{provider}/{model_name}"
     console.log(f"Using model: [green]{model_full}[/green]")
     init_llm(provider)
-    set_default_model(model_full)
+
+    model_meta = get_model(model_full)
+
+    # Apply GPTME_CONTEXT_LENGTH override (useful for local models with non-standard context)
+    context_length_str = os.environ.get("GPTME_CONTEXT_LENGTH")
```
ErikBjare, line 124: Shouldn't this use `config.get_env`?

ErikBjare, line 124: Why isn't this happening inside set_default_model or something? Seems weirdly specific for simple init.py

PR #1689, opened 2026-03-16
fix(llm): prevent double-wrapped <think> tags when both reasoning mechanisms fire

### LINKED ISSUE #1684: gpt-5.4 via openai-subscription: thinking leaks + tool fumbling
## Problem

gpt-5.4 via the `openai-subscription` provider has significantly worse autonomous performance in gptme compared to codex (same model): **58% productive vs 88% productive** over 24 vs 8 sessions respectively.

### Observed issues

1. **`<thinking>` tag leaks**: gpt-5.4 outputs `<thinking>...</thinking>` reasoning tokens in its responses. These appear in the conversation as visible text rather than being stripped. Example from autonomous session:
   ```
   Bob: Thinking...
   Bob:
   <thinking>          <thinking>
   Need execute workflow. Use todo write with steps...
   ```

2. **Tool call fumbling**: The model repeatedly calls `@todo` with empty results, fails to use `@gh` correctly ("Unknown gh command"), and struggles with gptme's tool syntax. codex with native tool use doesn't have this problem.

3. **Shell quoting failures**: The model uses `$'\n'` escape sequences that gptme's shell tool rejects with "Shell syntax error: unexpected EOF". It recovers by writing to temp files, but wastes tokens.

### Data

```
gptme:gpt-5.4:  24 sessions, 58% productive, avg grade 0.32
codex:gpt-5.4:   8 sessions, 88% productive, avg grade 0.26
```

Same model, different harnesses. The difference is gptme's tool layer vs codex's native tool use.

### Suggestions

- **Strip thinking tags**: If `supports_reasoning` is true in model metadata, strip `<thinking>` blocks from model output before processing
- **Model metadata**: The installed gptme (0.31.0+779b0ad9) shows "Unknown model: using fallback metadata for openai-subscription/gpt-5.4" — needs `uv sync` to pick up the model definitions that exist in source
- **Shell tool robustness**: Handle `$'...'` quoting or provide clearer error messages that help the model recover faster

### Context

This was found during an operator monitoring session analyzing Bob's autonomous run performance. See ErikBjare/bob journal 2026-03-16 for full analysis.

TimeToBuildBob:
## Problem

After #1685, gpt-5.4 still produced malformed output with nested \`<think>\` tags:

\`\`\`
<think>       <think>
The user is asking me to finish...
</think>        </think>
\`\`\`

This caused the autoresearch agent to loop repeatedly outputting thinking without making tool calls (gptme's parser saw the malformed tags and couldn't extract a tool call).

## Root Cause

gpt-5.4 emits reasoning content through **both** mechanisms simultaneously:
1. \`response.reasoning.delta\` SSE events → the #1685 handler wraps in \`<think>\` blocks
2. Raw \`<thinking>\` tags in \`response.output_text.delta\` → the #1685 handler converts to \`<think>\`

When both fire for the same content, both layers apply and produce double-wrapped output.

## Fix

Track \`seen_reasoning_delta\` per stream. When any \`response.reasoning.delta\` event is received, skip the \`<thinking>\`→\`<think>\` text conversion in \`output_text.delta\` — the structured reasoning.delta path is authoritative.

This is mutually exclusive by design:
- Models using \`reasoning.delta\`: \`seen_reasoning_delta=True\`, text conversion skipped
- Models only using \`<thinking>\` tags in text: \`seen_reasoning_delta=False\`, text conversion still applies (unchanged behavior)

## Test

Added \`test_stream_no_double_wrap_when_both_mechanisms_fire\` to the subscription test suite, which fails before this fix and passes after.

Closes #1684 (partial — addresses the thinking-tag leak; other subtasks remain)

### DIFF
--- a/gptme/llm/llm_openai_subscription.py
+++ b/gptme/llm/llm_openai_subscription.py
@@ -607,6 +607,11 @@ def stream(
         raise ValueError(f"Codex API error {response.status_code}: {error_text}")
 
     in_reasoning_block = False
+    # Track whether this stream uses structured reasoning.delta events.
+    # When true, skip <thinking>→<think> text conversion to avoid double-wrapping:
+    # gpt-5.4 can emit BOTH response.reasoning.delta AND raw <thinking> tags in
+    # output_text.delta for the same content, causing <think><think>...</think></think>.
+    seen_reasoning_delta = False
     # Buffer to detect tags split across SSE chunks (max tag length is ~11 chars)
     pending = ""
 
@@ -628,6 +633,7 @@ def stream(
             # Structured reasoning content — wrap in <think> blocks
             delta_text = data.get("delta", "")
             if delta_text:
+                seen_reasoning_delta = True
                 if not in_reasoning_block:
                     # Flush any pending partial-tag buffer before opening think block
                     if pending:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #1691, opened 2026-03-17
fix(codeblock): handle </thinking> end tag from gemini models

TimeToBuildBob:
## Summary

- Fixes `_extract_codeblocks` to handle `</thinking>` end tag (used by Gemini models) in addition to `</think>` (used by Claude/DeepSeek)
- Also handles `<thinking>` start tag for the unclosed-block early-exit path
- Adds 3 regression tests

## Root cause

Confirmed via autoresearch eval logs: when `gemini-2.0-flash-001` is used as the eval model, it wraps reasoning in backtick-fenced ` ```thinking> ` blocks closing with `</thinking>`. The `_extract_codeblocks` function only stripped `</think>`, so `</thinking>` left the "thinking block" unclosed. Any subsequent ` ```save ` or ` ```python ` block was treated as nested (nesting depth > 0) and never yielded.

**Effect on practical5 evals**: `data-pipeline` and `regex-scrub` always scored 0 with gemini as eval model — the agent generated correct code but the save block was silently swallowed.

## Fix

Check for `</thinking>` before `</think>` (first match wins), and check for both `<thinking>` and `<think>` in the early-exit path for unclosed thinking blocks.

## Test plan

- [x] `test_thinking_block_with_closing_thinking_tag` — gemini-style `</thinking>`, save block is extracted
- [x] `test_thinking_block_with_closing_think_tag` — existing `</think>` behavior preserved  
- [x] `test_thinking_block_unclosed_thinking_tag_early_exit` — unclosed `<thinking>` still causes early exit
- [x] Full `tests/test_codeblock.py` suite: 49/49 passing

### DIFF
--- a/gptme/codeblock.py
+++ b/gptme/codeblock.py
@@ -103,15 +103,19 @@ def _extract_codeblocks(
     This handles nested cases where ``` appears inside string literals or other content.
     """
     # dont extract codeblocks from thinking blocks
-    # (since claude sometimes forgets to close codeblocks in its thinking)
-    think_end = markdown.find("</think>")
-    if think_end != -1:
-        # remove anything before and including </think> if it exists
-        markdown = markdown[think_end + len("</think>") :]
+    # (since claude sometimes forgets to close codeblocks in its thinking,
+    #  and gemini uses </thinking> instead of </think>)
+    for _think_end_tag in ["</thinking>", "</think>"]:
+        _think_end = markdown.find(_think_end_tag)
+        if _think_end != -1:
+            # remove anything before and including the closing thinking tag
+            markdown = markdown[_think_end + len(_think_end_tag) :]
+            break
     else:
-        # if start <think> tag but no end, early exit
-        if "<think>" in markdown:
-            return
+        # if start thinking tag but no end, early exit
+        for _think_start_tag in ["<thinking>", "<think>"]:
+            if _think_start_tag in markdown:
+                return
 
     # speed check (early exit): check if message contains a code block
     # Check for at least 2 fence markers (3+ backticks each)
--- a/tests/test_codeblock.py
+++ b/tests/test_codeblock.py
@@ -1265,3 +1265,37 @@ def test_mismatched_fence_lengths():
     markdown3 = "```text\nline 1\n````"
     blocks3 = Codeblock.iter_from_markdown(markdown3)
     assert len(blocks3) == 0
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

ErikBjare:
Strange review-request looping here. Also, autoresearch shouldn't be trying to optimize for Gemini 2.0 Flash - it's an old and poorly behaving model.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob what's going on with review request looping here?

TimeToBuildBob:
The review-request loop is a bug in the project monitoring script (`scripts/runs/github/project-monitoring.sh`). The cross-repo Greptile check is failing to detect the existing review and keeps re-triggering — most likely the `python3 -c` subprocess is erroring out silently (timeout or API parse failure), defaulting to `has_review: False`, which triggers another `@greptileai review` on each monitoring cycle.

Will fix the detection logic — needs a guard that checks for existing `@greptileai review` comments before posting again.

As for the PR itself: CI is all green, Greptile gave 5/5 ("safe to merge"). Ready when you are.

TimeToBuildBob:
@greptileai review

ErikBjare:
Did you fix the logic? I saw it got yet another review request!

TimeToBuildBob:
Yes — fixed in ErikBjare/bob@091c750. The guard now checks for existing `@greptileai review` comments before posting, so it won't re-trigger while waiting for a response.

The one extra request you saw (at 15:13:00Z) was from a monitoring cycle that was already in-flight when the commit landed 31 seconds earlier. Won't happen again.

The PR itself is CLEAN/MERGEABLE, CI green, Greptile 5/5 ("safe to merge"). Ready when you are.

TimeToBuildBob:
Also addressed your point about Gemini 2.0 Flash: removed it from `eval_model_candidates` in the practical5 experiment config (ErikBjare/bob@f082670). Claude Sonnet is now the sole eval model. While PR #1691 fixes the codeblock bug that caused the symptom, using a poorly-behaving model as an eval signal isn't worth it regardless.

PR #1701, opened 2026-03-18
fix(shell): use findall+last for return code parsing (Greptile follow-up from #1672)

TimeToBuildBob:
## Summary

Follow-up to #1672, addressing the secondary finding from the Greptile review.

**The bug**: When a command outputs text containing `ReturnCode:N` without a trailing newline, the line seen by the parser looks like:
```
ReturnCode:1 ReturnCode:0 END_OF_COMMAND_OUTPUT
```
`re_returncode.search(line)` finds the **first** occurrence (from command output), returning exit code `1` instead of the actual `0`.

**The fix**: Use `findall()` + last element, consistent with the `rfind()` approach already used for prefix extraction. The last `ReturnCode:N` is always the shell-injected marker.

Applied to both `_read_output_unix` and the threading path (`_read_output_windows`).

## Test plan
- [x] All 53 shell tool tests pass (`tests/test_tools_shell.py`)

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -887,8 +887,11 @@ class ShellSession:
                             logger.debug(
                                 f"Shell: Delimiter detected in line: {line.strip()[:200]}"
                             )
-                            if match := re_returncode.search(line):
-                                return_code = int(match.group(1))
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1716, opened 2026-03-20
refactor(eval): use specific exception types and cleaner imports

TimeToBuildBob:
## Summary
- Replace bare `except Exception` with `except PackageNotFoundError` in `write_results()` version fallback (main.py)
- Consolidate `concurrent.futures` imports — import `TimeoutError`, `CancelledError` directly instead of using module-qualified references (run.py)
- Simplify status check from `!= "timeout" and != "error"` to `not in ("timeout", "error")`

## Test plan
- [x] Imports verified with `python3 -c "from gptme.eval.run import run_evals"`
- [x] mypy shows no new errors in changed files
- [ ] CI passes

### DIFF
--- a/gptme/eval/run.py
+++ b/gptme/eval/run.py
@@ -154,7 +154,7 @@ def run_evals(
             except Exception as e:
                 gen_time = 0
                 error_detail = ""
-                if isinstance(e, (TimeoutError, CancelledError)):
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1724, opened 2026-03-21
feat(llm): model-dependent break_on_tooluse default

### LINKED ISSUE #1719: Consider making GPTME_BREAK_ON_TOOLUSE=false the default
## Summary

Currently `GPTME_BREAK_ON_TOOLUSE` defaults to `True`, meaning the LLM streaming is interrupted at the first detected runnable tool use. The LLM then generates a new response for the next tool, resending the full context each time. Setting it to `False` allows the LLM to generate multiple tool uses in a single response, which are then executed sequentially by `execute_msg()`.

## Motivation

- **Cost reduction**: Each LLM call resends the full conversation context. With `break_on_tooluse=True`, a turn that uses 3 tools requires 3 LLM calls. With `False`, it's potentially 1 call.
- **Observed behavior**: When testing minimax m2.7, `break_on_tooluse` didn't work (effectively `False`), and the multi-tool-per-step behavior worked well and seemed cost-effective.
- **Manual testing**: Running with `GPTME_BREAK_ON_TOOLUSE=false` works well for many common patterns (e.g. exploring files, running multiple shell commands, reading + searching).

## Current Implementation

- `gptme/llm/__init__.py` (`_reply_stream`): Checks `GPTME_BREAK_ON_TOOLUSE` env var, defaults to `True`. During streaming, checks each newline for runnable tool uses and breaks if found.
- `gptme/tools/__init__.py` (`execute_msg`): Already iterates over all runnable tool uses in a message — no changes needed there.
- `gptme/chat.py` (`_process_message_conversation`): After executing tools, checks if the last assistant message still has runnable tools. The outer loop naturally handles both modes.

## Tradeoffs

**Pros of `False` (multi-tool per step):**
- Fewer LLM calls → lower cost (especially with large contexts)
- Faster turns (less round-trip latency)
- More natural for independent operations (explore files, run tests, etc.)

**Cons of `False`:**
- LLM can't see intermediate tool outputs before generating next tool call
- If an early tool fails, subsequent tools may be based on wrong assumptions
- Some models may not generate multiple tool blocks reliably

## Proposal

Consider one of:
1. **Change default to `False`** — simpler, works well in practice
2. **Make it model-dependent** — some models handle multi-tool better than others
3. **Smart detection** — break on tool use only when the tool output is likely needed for the next action (harder to implement)
4. **Keep current default but document the option better** — let users opt in

Option 1 seems like the best starting point, with option 2 as a refinement later.

## Related

- The `streaming=True` parameter in `ToolUse.iter_from_content()` requires a blank line after code blocks during streaming detection, which helps avoid false positives.

TimeToBuildBob:
## Summary

Implements the approach discussed in #1719 (model-dependent default for `GPTME_BREAK_ON_TOOLUSE`).

- Adds `supports_parallel_tool_calls: bool = False` flag to `ModelMeta`
- Marks capable models as `True`: claude-sonnet/opus 4-6, claude-4.x family, gpt-4o, gpt-5, gpt-4.1 families
- When `GPTME_BREAK_ON_TOOLUSE` env var is **not set**: default is derived from model capability
  - Capable models → `break_on_tooluse=False` (allow parallel tool calls per response)  
  - Other models → `break_on_tooluse=True` (safe fallback)
- When env var **is set**: explicit value wins (backward compatible)

## Motivation

From Erik's testing in #1719:
- **Haiku 4.5**: Does not emit parallel tool calls — forced break is the right default
- **Sonnet 4.6**: Immediately emits multiple tool calls when `break_on_tooluse=False` — can use parallel tool calls efficiently

A global default of `False` would break smaller/older models, so a per-model list is the right approach.

## Test plan
- [x] `TestSupportsParallelToolCalls` — 7 tests verifying flags on specific models
- [x] All existing model tests pass (46/46)
- [ ] Manual: verify Sonnet 4.6 uses parallel tool calls by default after merge

Closes #1719

### DIFF
--- a/gptme/llm/models.py
+++ b/gptme/llm/models.py
@@ -784,6 +784,7 @@ def get_model(model: str) -> ModelMeta:
                                 supports_streaming=model_meta.supports_streaming,
                                 supports_vision=model_meta.supports_vision,
                                 supports_reasoning=model_meta.supports_reasoning,
+                                supports_parallel_tool_calls=model_meta.supports_parallel_tool_calls,
                                 price_input=model_meta.price_input,
                                 price_output=model_meta.price_output,
                                 knowledge_cutoff=model_meta.knowledge_cutoff,


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the OpenRouter gap flagged by Greptile in 7a7ea0d: added `supports_parallel_tool_calls=model_meta.supports_parallel_tool_calls` to the dynamic model rebuild block. OpenRouter users with capable models (e.g. `openrouter/anthropic/claude-sonnet-4-6`) will now also benefit from parallel tool calls.

All tests pass.

TimeToBuildBob:
@greptileai review

PR #1725, opened 2026-03-21
fix(models): add supports_parallel_tool_calls to openrouter/anthropic/claude-sonnet-4-6

TimeToBuildBob:
## Summary

- `anthropic/claude-sonnet-4-6` via OpenRouter was missing `supports_parallel_tool_calls: True`
- This caused `break_on_tooluse` to default to `True` for OpenRouter users, preventing parallel tool calls even though the underlying model supports it
- Added the flag to match the `anthropic` provider entry for the same model
- Added a test to prevent regression

## Root cause

PR #1724 added `supports_parallel_tool_calls` to Anthropic models and established the model-dependent `break_on_tooluse` default. The OpenRouter entry for the same model was overlooked.

## Test plan

- [x] All `TestSupportsParallelToolCalls` tests pass (8/8)
- [x] `openrouter/anthropic/claude-sonnet-4-6` now correctly returns `supports_parallel_tool_calls=True`

### DIFF
--- a/gptme/llm/models.py
+++ b/gptme/llm/models.py
@@ -524,6 +524,7 @@ MODELS: dict[Provider, dict[str, _ModelDictMeta]] = {
             "price_output": 15,
             "supports_vision": True,
             "supports_reasoning": True,
+            "supports_parallel_tool_calls": True,
         },
         "meta-llama/llama-3.3-70b-instruct": {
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #1739, opened 2026-03-21
fix(tests): fix flaky test_api_conversation_generate mock setup

TimeToBuildBob:
## Summary
- Fix `test_api_conversation_generate` which was marked `@pytest.mark.xfail` due to intermittent `'Mock' object is not iterable` errors in CI
- Replace real API call with a properly mocked `_stream` generator that returns deterministic iterable responses
- Remove `@pytest.mark.xfail`, `@pytest.mark.requires_api`, and `@pytest.mark.slow` since the test no longer needs a real API and runs in ~1.5s

## Root cause
The test made a real LLM API call via the Flask test client. In CI, something in the test environment occasionally replaced `_stream` with a non-iterable `Mock` object, causing `"".join(_stream(...))` to raise `TypeError: 'Mock' object is not iterable`, which the API error handler caught and returned as a JSON error response.

## Fix
Mock `gptme.server.api._stream` with a generator function that yields strings, following the same pattern used by V2 API tests (`test_server_v2_auto_stepping.py`, `test_server_v2_tool_confirmation.py`).

## Test plan
- [x] `test_api_conversation_generate` passes consistently (5/5 runs)
- [x] All 17 `test_server.py` tests pass

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,5 +1,6 @@
 import copy
 import random
+import unittest.mock
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1733, opened 2026-03-21
refactor: remove 47 unused type: ignore comments

TimeToBuildBob:
## Summary

- Removed 47 unused `# type: ignore` comments across 18 files, found via `mypy --warn-unused-ignores`
- Kept 5 that suppress real errors (attr-defined, assignment, arg-type, call-overload, index)
- Zero behavioral change — pure cleanup

## Files changed

Biggest cleanup areas: `gptme/acp/agent.py` (15), `gptme/acp/client.py` (8), `gptme/llm/llm_anthropic.py` (3), `gptme/logmanager.py` (3)

## Test plan

- [x] mypy passes with `--warn-unused-ignores` (no remaining unused)
- [x] mypy passes without `--warn-unused-ignores` (no new errors)
- [x] ruff format clean

Issue #1731, opened 2026-03-21
Spring Cleaning


ErikBjare:
gptme core prides itself on being simple and [tiny](https://gptme.org/docs/arewetiny.html), but I fear we are starting to drift a bit.

Much of the complexity has been addressed by moving non-core/experimental features into gptme-contrib, but I want a full code-review and an extensive list of potential candidates of things/features to remove (or move to gptme-contrib).

This could be rarely used features, features that cause an undue amount of complexity, features that are not stable. This could be done by looking at source code, at commits, or at PRs. See added/removed lines to understand where complexity comes from. 

It could also mean to find alternative solutions to messy solutions (parsing tool-calls in webui from server's streaming messages? maybe server-side should just return the constructed/parsed toolcall to make it support different `--tool-format`), which other such wins? Keep a ranked list.

See e.g. https://pi.dev/'s "What we didn't build" section.

The result of cleaning would ideally be a leaner and simpler/tinier core that's easier/cleaner to extend.

TimeToBuildBob:
## Spring Cleaning Audit

Full code review of gptme core (228 modules, ~78K SLOC). Ranked by impact and grouped into tiers.

### Codebase Overview

| Area | SLOC | Files | Notes |
|------|------|-------|-------|
| Tools | 17,546 | 35+ | Largest subsystem — shell.py alone is 2,341 |
| Eval | 11,939 | 30+ | Biggest growth area (practical1-8, dspy, swebench, tbench) |
| Server | 6,112 | 15 | api_v2_sessions.py is 1,586 LOC monolith |
| LLM | 5,946 | 12 | Three providers with duplicated patterns |
| Lessons | 3,170 | 10 | Three overlapping matcher implementations |
| ACP | 2,400 | 6 | Relatively new, well-structured |
| Hooks | 1,844 | 5 | Growing but organized |
| Config | 1,222 | 1 | Monolithic |

### Tier 1: Move to gptme-contrib (non-core features)

These exist in core but aren't essential for the core loop of "chat with LLM + use tools":

| Candidate | SLOC | Effort | Risk | Rationale |
|-----------|------|--------|------|-----------|
| `eval/dspy/` | ~4,500 | 4-6h | Low | Research-stage prompt optimization. Isolated module. Known issues in README. |
| `eval/swebench/` | ~500 | 2h | Low | Benchmark evaluation, isolated |
[...]

PR #1742, opened 2026-03-21
refactor: remove 11 unused type: ignore comments

TimeToBuildBob:
## Summary
- Removes 11 unused `type: ignore` comments found by `mypy --warn-unused-ignores`
- Follow-up to #1733 which removed 47 unused comments
- Part of Spring Cleaning (#1731)

## Test plan
- [x] `mypy --warn-unused-ignores` reports zero unused ignores
- [x] `mypy` passes with no errors

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -18,7 +18,7 @@ from click.core import ParameterSource
 try:
     from pick import pick
 except ImportError:
-    pick = None  # type: ignore[assignment]
+    pick = None
 
 import gptme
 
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Closing this PR — the `type: ignore` comments it removes are actually needed in the CI environment.

CI typecheck shows all 11 removed comments correspond to real mypy errors:
- `ace.embedder` / `ace.gptme_integration`: `ace` is not installed in CI, so `import-not-found` ignores are required for the optional import block
- The rest (`attr-defined`, `arg-type`, `call-overload`, `index`, `assignment`): library stub versions in CI still produce these errors

The `--warn-unused-ignores` check passed locally because the local environment had different stub versions (or `ace` installed), making those errors invisible. CI uses a different environment where the errors are real.

Master's typecheck passes with those comments in place. This PR removing them caused the typecheck to break. The correct fix would be to either:
1. Update CI stub versions to match local (so the ignores become truly unused), or  
2. Fix the underlying type issues properly (e.g., add proper type narrowing instead of ignoring)

For now, keeping the ignores is the right call.

PR #1748, opened 2026-03-22
fix(tests): eliminate 250+ 'init() called twice' warning spam

TimeToBuildBob:
## Summary

- Re-applies the `_init_done` reset fix (originally `ffdb15583`, reverted in `02f98ff98`) in `clear_tools_before()` fixture
- The revert was needed because it exposed a flaky test (`test_init_tools_allowlist`), but that was independently fixed in `9ef1e0037` — safe to re-apply now
- Eliminates ~250+ noisy "init() called twice, ignoring" warnings from test output, making real warnings visible

## Test plan

- [x] `test_init_tools_allowlist` passes (the previously flaky test)
- [x] 1603 tests pass, 0 failures
- [x] Zero "init() called twice" warnings in test output
- [x] mypy/ruff checks pass

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -230,7 +230,12 @@ def temp_file():
 
 @pytest.fixture(autouse=True)
 def init_():
-    init(None, interactive=False, tool_allowlist=None, tool_format="markdown")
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1750, opened 2026-03-22
fix(acp,llm): fail-closed permission checks + surface token refresh errors

TimeToBuildBob:
## Summary

Two security/UX fixes:

- **ACP permission checks now fail-closed**: When permission request fails (network error, timeout, malformed response), tool execution is denied instead of auto-allowed. The no-connection case (ACP not in use) still auto-allows for backward compatibility. This prevents agents from executing dangerous tools when the permission system is down.

- **OpenAI subscription token refresh errors surfaced**: When token refresh fails, users now see the actual error (e.g., network timeout, OAuth server down) instead of a misleading "not authenticated, please run gptme auth" message. Helps users distinguish transient issues from actual auth problems.

## Test plan

- [x] All 98 ACP agent tests pass (including updated `test_permission_request_exception_denies`)
- [x] No-connection auto-allow behavior preserved (backward compat)
- [ ] CI passes

### DIFF
--- a/gptme/llm/llm_openai_subscription.py
+++ b/gptme/llm/llm_openai_subscription.py
@@ -402,6 +402,7 @@ def _refresh_access_token(refresh_token: str) -> SubscriptionAuth:
 def get_auth() -> SubscriptionAuth:
     """Get current auth, refreshing or prompting login if needed."""
     global _auth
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1757, opened 2026-03-22
fix(tests): properly isolate server v2 test from user config

TimeToBuildBob:
## Summary

- Fixes `test_v2_create_conversation_default_system_prompt` failing on systems with user-level gptme config (e.g., `~/.config/gptme/config.toml` with `prompt.files` set)
- Root cause: monkeypatching `gptme.config.config_path` didn't propagate to `user.py` (which defines its own module-level `config_path`) or `workspace.py` (which binds `config_path` at import time)
- Fix: inject a clean `Config` with `default_config` directly via `set_config()`, and patch `workspace.py`'s local `config_path` reference

## Test plan

- [x] `test_v2_create_conversation_default_system_prompt` now passes on Bob's VM (has user-level prompt files configured)
- [x] All 11 server v2 tests pass
- [ ] CI green

### DIFF
--- a/tests/test_server_v2.py
+++ b/tests/test_server_v2.py
@@ -101,11 +101,21 @@ def test_v2_create_conversation_default_system_prompt(
     monkeypatch.chdir(tmp_path)
     # Explicitly disable chat history for this test
     monkeypatch.setenv("GPTME_CHAT_HISTORY", "false")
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #1828, opened 2026-03-24
feat(llm): add max_tokens parameter to reply() and chat()

TimeToBuildBob:
## Problem

OpenRouter counts the *reserved* `max_tokens` against the key's daily dollar budget — not actual usage. For `claude-sonnet-4-5` with `max_output=64000`, every API call reserves ~\$0.96 even if only 100 tokens are used.

This hits hard on lightweight use cases like tweet evaluation, where 48 daily cycles × ~2 LLM calls × \$0.96/reservation = \$92/day in reservations, vs ~\$1/day actual cost.

## Fix

Add a `max_tokens: int | None = None` parameter to `reply()`, threading it through `_chat_complete()` → `chat()` in both `llm_openai.py` and `llm_anthropic.py`.

When specified, the value is passed directly to the API. When `None` (default), behavior is unchanged — the API uses the model's default.

## Changes

- `llm/__init__.py`: add `max_tokens` to `reply()` and `_chat_complete()` signatures
- `llm/llm_openai.py`: thread `max_tokens` into `optional_kwargs` for the API call
- `llm/llm_anthropic.py`: allow caller-specified `max_tokens` to override `model_meta.max_output`

## Usage

```python
# Cap at 2048 tokens for tweet evaluation — prevents ~$0.96 reservation
# from 64k default down to ~$0.03
response = reply(messages, model.full, stream=False, max_tokens=2048)
```

## Motivation

Fixes Twitter loop cost issue in ErikBjare/bob#479 — OpenRouter key exhausted daily budget on reservations for `claude-sonnet-4-5` despite minimal actual token usage.

### DIFF
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -203,6 +203,11 @@ def _chat_complete(
         )
     if provider == "openai-subscription":
         # TODO: max_tokens not yet forwarded to openai-subscription
+        if max_tokens is not None:
+            logger.warning(
+                "max_tokens=%d is not forwarded to openai-subscription provider and will be ignored.",
+                max_tokens,
+            )
         content = chat_subscription(messages, _get_base_model(model), tools)
         return content, {"model": model}
 
@@ -257,12 +262,16 @@ def _stream(
         return _StreamWithMetadata(gen, model)
     if provider == "openai-subscription":
         # TODO: max_tokens not yet forwarded to openai-subscription
+        if max_tokens is not None:
+            logger.warning(
+                "max_tokens=%d is not forwarded to openai-subscription provider and will be ignored.",
+                max_tokens,
+            )
         gen = stream_subscription(messages, _get_base_model(model), tools)
         return _StreamWithMetadata(gen, model)
     # Note: Validation-only fallback for streaming is complex
     # For now, unsupported providers don't support output_schema in streaming mode
     if output_schema is not None:
-        logger = logging.getLogger(__name__)
         logger.warning(
             f"Provider {provider} does not support output_schema in streaming mode"
         )
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -187,6 +187,8 @@ def _chat_complete(
     output_schema: type | None = None,
     max_tokens: int | None = None,
 ) -> tuple[str, MessageMetadata | None]:
+    if max_tokens is not None and max_tokens <= 0:
+        raise ValueError(f"max_tokens must be a positive integer, got {max_tokens}")
     provider = get_provider_from_model(model)
 
     # Providers with native constrained decoding support
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -256,6 +256,7 @@ def _stream(
         )
         return _StreamWithMetadata(gen, model)
     if provider == "openai-subscription":
+        # TODO: max_tokens not yet forwarded to openai-subscription
         gen = stream_subscription(messages, _get_base_model(model), tools)
         return _StreamWithMetadata(gen, model)
     # Note: Validation-only fallback for streaming is complex
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -96,6 +96,7 @@ def reply(
     workspace: Path | None = None,
     output_schema: type | None = None,
     on_token: Callable[[str], None] | None = None,
+    max_tokens: int | None = None,
 ) -> Message:
     # Trigger GENERATION_PRE hooks and collect context messages
     from ..hooks import HookType, trigger_hook
@@ -137,7 +138,11 @@ def reply(
         )
     rprint(f"{prompt_assistant(agent_name)}: Thinking...", end="\r")
     response, metadata = _chat_complete(
-        generation_msgs, model, tools, output_schema=output_schema
+        generation_msgs,
+        model,
+        tools,
+        output_schema=output_schema,
+        max_tokens=max_tokens,
     )
     rprint(" " * shutil.get_terminal_size().columns, end="\r")
     rprint(f"{prompt_assistant(agent_name)}: {response}")
@@ -177,16 +182,23 @@ def _chat_complete(
     model: str,
     tools: list[ToolSpec] | None,
     output_schema: type | None = None,
+    max_tokens: int | None = None,
 ) -> tuple[str, MessageMetadata | None]:
     provider = get_provider_from_model(model)
 
     # Providers with native constrained decoding support
     # Custom providers are OpenAI-compatible, so route them through the OpenAI path
     if provider in PROVIDERS_OPENAI or is_custom_provider(provider):
-        return chat_openai(messages, model, tools, output_schema=output_schema)
+        return chat_openai(
+            messages, model, tools, output_schema=output_schema, max_tokens=max_tokens
+        )
     if provider == "anthropic":
         return chat_anthropic(
-            messages, _get_base_model(model), tools, output_schema=output_schema
+            messages,
+            _get_base_model(model),
+            tools,
+            output_schema=output_schema,
+            max_tokens=max_tokens,
         )
     if provider == "openai-subscription":
         content = chat_subscription(messages, _get_base_model(model), tools)
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -135,6 +135,7 @@ def reply(
             agent_name=agent_name,
             output_schema=output_schema,
             on_token=on_token,
+            max_tokens=max_tokens,
         )
     rprint(f"{prompt_assistant(agent_name)}: Thinking...", end="\r")
     response, metadata = _chat_complete(
@@ -235,15 +236,22 @@ def _stream(
     model: str,
     tools: list[ToolSpec] | None,
     output_schema: type | None = None,
+    max_tokens: int | None = None,
 ) -> _StreamWithMetadata:
     provider = get_provider_from_model(model)
     # Custom providers are OpenAI-compatible, so route them through the OpenAI path
     if provider in PROVIDERS_OPENAI or is_custom_provider(provider):
-        gen = stream_openai(messages, model, tools, output_schema=output_schema)
+        gen = stream_openai(
+            messages, model, tools, output_schema=output_schema, max_tokens=max_tokens
+        )
         return _StreamWithMetadata(gen, model)
     if provider == "anthropic":
         gen = stream_anthropic(
-            messages, _get_base_model(model), tools, output_schema=output_schema
+            messages,
+            _get_base_model(model),
+            tools,
+            output_schema=output_schema,
+            max_tokens=max_tokens,
         )
         return _StreamWithMetadata(gen, model)
     if provider == "openai-subscription":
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed Greptile's finding: `max_tokens` was silently dropped when `stream=True`.

**Fix**: Propagated `max_tokens` through the full streaming stack — `reply()` → `_reply_stream()` → `_stream()` → `stream_openai()`/`stream_anthropic()`. Both providers now respect the caller-specified cap in streaming mode using the same patterns as the non-streaming path.

Commit: c4a692f

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's two findings:

**P1 — Anthropic thinking budget conflict** (critical): When `use_thinking=True`, Anthropic's API requires `max_tokens > budget_tokens`. A caller passing a small value (e.g. 2048) with the default thinking budget of 16000 would get an API error. Fixed by clamping `max_tokens` to `thinking_budget + 1` with a warning when this condition is detected.

**P2 — Silent discard on `openai-subscription` path**: Added a `# TODO: max_tokens not yet forwarded to openai-subscription` comment to make the gap explicit rather than silently ignoring the value.

Commit: 00518cb

TimeToBuildBob:
Cleaned up branch history with a force-push to fix two issues from the prior approach:

1. **Streaming path regression**: A previous session replaced the streaming propagation fix (c4a692fc) with just a warning (`max_tokens has no effect when stream=True`). That was wrong — the streaming path **does** forward `max_tokens` properly. Restored c4a692fc which correctly propagates through `reply()` → `_reply_stream()` → `_stream()` → `stream_openai()`/`stream_anthropic()`.

2. **Incomplete thinking budget guard** (Greptile P1): The earlier guard (00518cb0) only protected `chat()`, not `stream()`. Commit 94f81c18 adds the `max_tokens <= thinking_budget` clamp to **both** `chat()` and `stream()` in `llm_anthropic.py`.

Current branch: `4ea389ba` (feat) → `c4a692fc` (streaming) → `94f81c18` (thinking budget, both paths).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fix for Greptile P1 (thinking budget guard) was in commit `94f81c18`, but it had been pushed to `gptme/gptme.git` directly instead of the fork — so the PR was stuck at `c4a692fc` (missing the guard). Pushed it to the fork now; PR head is `94f81c18`. CI running, re-triggered Greptile review.

TimeToBuildBob:
Addressed Greptile's two remaining 4/5 findings (commit `580d3eee5`):

**P1 — Missing TODO in `_stream()` for `openai-subscription`** (minor): Added `# TODO: max_tokens not yet forwarded to openai-subscription` comment at line 258 of `__init__.py`, matching the existing TODO in `_chat_complete()`.

**P2 — Anthropic thinking budget clamping direction**: Changed the guard to reduce `thinking_budget` to `max_tokens - 1` rather than inflating `max_tokens` to `thinking_budget + 1`. This respects the caller's cost-saving intent — if a caller specifies `max_tokens=2048` on a thinking-enabled model, we now honor their budget limit and reduce the thinking budget accordingly, with a warning that includes the `GPTME_REASONING_BUDGET` env var name so they can tune it.

ErikBjare:
Maybe adding this one thing is reasonable, but I'm really trying to keep the number of params passed through like this. It's +60 lines for a simple change, running through core functions like chat and step. An alternative would be to simply support a `GPTME_MAX_TOKENS` (like we do with max steps/turns) but I guess that won't work for library callers, so maybe this is right.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Correction on the previous comment: commit `580d3eee5` was in the local worktree but hadn't actually been pushed to the fork — so the PR head was still `94f81c18`. Pushed it now; PR head is `580d3eee5`.

Re-triggered Greptile review. CI will run on the new commit.

TimeToBuildBob:
Fair concern — the threading is the main cost here. The env var approach was my first thought too, but it breaks per-call control (e.g. different token budgets for different pipeline stages in the same process). The twitter workflow that motivated this needs exactly that: cap output on cheap classification calls but allow full output on generation calls.

That said, if you want to keep the function signatures clean, one option would be a context variable (`contextvars.ContextVar`) that library callers can set per-call scope — same semantics as a parameter but no signature change. Let me know if you'd prefer that approach and I'll rework it. Otherwise I'm happy with the current form if you are.

(CI failure is `test_mcp_cli_commands` timeout — flaky test unrelated to this PR. Re-ran it.)

TimeToBuildBob:
Fixed the last Greptile finding: `new_budget = max_tokens - 1` can yield 0 when max_tokens=1. Changed to `max(1, max_tokens - 1)` in both `chat()` and `stream()`. Commit: 160795849.

CI is green on the previous commit — triggering a re-review on the new one.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the two remaining Greptile 4/5 findings (commit `f0f00ca10`):

**P1 — thinking budget boundary condition when `max_tokens=1`**: The `max(1, max_tokens - 1)` guard still yields `thinking_budget = 1 = max_tokens` when `max_tokens=1`, violating the Anthropic API's strict `budget_tokens < max_tokens` constraint. Fixed by disabling extended thinking entirely when `max_tokens <= 1` (no room for even 1 thinking token), in both `chat()` and `stream()`.

**P2 — silent drop for `openai-subscription`**: Added `logger.warning()` in both `_chat_complete()` and `_stream()` when `max_tokens is not None` and the provider is `openai-subscription`, so callers know the cap was not applied. Also removed a local `logger = logging.getLogger(__name__)` re-definition that ruff/mypy flagged as used-before-assignment.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's two P2 style findings (commit `7fa25ddb0`):

**P1 — Duplicated thinking-budget clamping**: Extracted the 24-line block that appeared in both `chat()` and `stream()` into a shared `_adjust_thinking_budget(max_tokens, thinking_budget, use_thinking) -> tuple[int, bool]` helper. Both functions now call it in 3 lines.

**P2 — Falsy `or` chain**: Changed `max_tokens or model_meta.max_output or 4096` to `max_tokens if max_tokens is not None else (model_meta.max_output or 4096)` in both `chat()` and `stream()`. The old form would silently drop any falsy integer (e.g. zero) rather than treating `None` as the only sentinel.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Correction: commit `7fa25ddb0` was in the local worktree but hadn't been pushed to the fork — so the PR head was stuck at `f0f00ca10`. Pushed it now; PR head is `7fa25ddb0`.

This commit addresses Greptile's two remaining 4/5 findings:
- **P1 — Duplicated thinking-budget clamping**: Extracted into `_adjust_thinking_budget()` helper; both `chat()` and `stream()` now call it
- **P2 — Falsy `or` chain**: Changed to `max_tokens if max_tokens is not None else (model_meta.max_output or 4096)` to properly treat `None` as the only sentinel

Re-triggered Greptile review.

TimeToBuildBob:
Addressed the remaining Greptile P2 finding: added input validation for `max_tokens` in `reply()`.

**Fix (commit `3d33ff3ad`)**: Added a guard at the entry point of `reply()` that raises `ValueError` with a clear message when `max_tokens <= 0` is passed, rather than forwarding the invalid value to the provider API and getting a confusing server-side error.

```python
if max_tokens is not None and max_tokens <= 0:
    raise ValueError(f"max_tokens must be a positive integer, got {max_tokens}")
```

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining Greptile 4/5 finding (commit `a872fa6cd`):

**Near-zero response budget for thinking models**: `new_budget = max_tokens - 1` left only 1 token for actual response when a thinking-capable model had its budget clamped. For e.g. `max_tokens=2048` with a 16000-token thinking budget, the model would get 2047 thinking tokens and just 1 response token — effectively an empty reply.

**Fix**: Introduced `_MIN_RESPONSE_TOKENS = 256` constant. The helper now:
- Reserves at least 256 tokens for the actual response when clamping the thinking budget
- Disables extended thinking entirely when `max_tokens <= 256` (not enough room for both thinking and a useful response)

Updated warning messages communicate how many response tokens are being reserved.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining Greptile 4/5 finding (commit `cd47b9470`):

**OpenRouter reasoning budget not respecting caller's `max_tokens`**: `extra_body()` was hard-coding `reasoning.max_tokens=20000` regardless of the caller-supplied `max_tokens`. For the tweet evaluation use case (e.g. `max_tokens=2048`), the reasoning budget would exceed the total token budget, producing an invalid API request.

**Fix**: Updated `extra_body()` to accept `max_tokens` and clamp the reasoning budget using the same `_MIN_RESPONSE_TOKENS=256` floor pattern as the Anthropic path:
- Passes `max_tokens` through from both `chat()` and `stream()` to `extra_body()`
- Disables reasoning entirely when `max_tokens <= 256` (no room for useful response)
- Reduces reasoning budget to `max_tokens - 256` when cap is smaller than the 20000 default

Also added the `max_tokens <= 0` validation guard to `_chat_complete()` (mirroring `reply()`), so callers that bypass `reply()` get a clear error message.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed Greptile's two 3/5 findings (commit `6cea4ea11`):

**P1 — `_adjust_thinking_budget` early-return gap (correctness)**:
The condition `max_tokens > thinking_budget` returned early without checking whether enough room remained for the response. When `max_tokens` was only 1–255 tokens above `thinking_budget`, the function exited leaving fewer tokens than `_MIN_RESPONSE_TOKENS` guarantees.

Fix: changed to `max_tokens >= thinking_budget + _MIN_RESPONSE_TOKENS` — only skip adjustment when there is genuinely enough headroom.

**P2 — Duplicated `_MIN_RESPONSE_TOKENS` constant (style)**:
The constant was independently defined in both `llm_anthropic.py` and `llm_openai.py`. Moved to a shared `gptme/llm/constants.py` and imported in both provider files — single definition, no drift risk.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's two remaining 4/5 findings (commit `9745b7181`):

1. **Duplicate validation guard** (`gptme/llm/__init__.py` line 190–191): Removed the redundant `max_tokens <= 0` check in `_chat_complete` — it's already enforced in the public `reply()` entry point at line 101, so the inner copy was dead code with a different error message.

2. **Explicit reasoning disable** (`gptme/llm/llm_openai.py` line 631): Added `body["reasoning"] = {"enabled": False}` when `reasoning_budget` drops to 0. This is safer than omitting the key, since omission relies on OpenRouter's default staying "reasoning disabled" — an assumption that could silently break the cost-saving intent if the provider default ever changes.

PR #1969, opened 2026-04-01
test(computer): assert screenshot resize dimensions

TimeToBuildBob:
## Summary

- add a regression assertion in test_computer_screenshot_success
- verify the screenshot resize command uses the API dimensions directly
- close the Greptile P2 follow-up from #1949, where the branch-only commit fc74498 never landed on master

## Test plan

- [x] poetry run pytest tests/test_tools_computer.py -q

### DIFF
--- a/tests/test_tools_computer.py
+++ b/tests/test_tools_computer.py
@@ -591,7 +591,11 @@ def test_computer_screenshot_success(
         args=[], returncode=0
     )
 
-    result = computer("screenshot")
+    # Unset WIDTH/HEIGHT so env vars don't override the mocked display resolution.
+    with mock.patch.dict("os.environ", {}, clear=False) as env:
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed the env var fragility flagged by Greptile. The `computer()` function reads `WIDTH`/`HEIGHT` from the environment (overriding the resolution selected from the mocked `_get_display_resolution`), so the `"1366x768!"` assertion was fragile in any environment where those vars are set.

Fix (1a764b0): wrap the `computer("screenshot")` call in `mock.patch.dict("os.environ", ...)` and pop `WIDTH`/`HEIGHT` inside it — the original env is restored after the `with` block.

TimeToBuildBob:
@greptileai review

PR #2065, opened 2026-04-07
fix(server): improve external session provider detection for uv tool installs

TimeToBuildBob:
## Problem

When `gptme-sessions` is installed via `uv tool install gptme-sessions`, it goes into an isolated venv at `~/.local/share/uv/tools/gptme-sessions/`. The `gptme_sessions` Python module is **not importable** from gptme's own Python environment.

The current `get_external_session_provider()` silently catches the `ImportError` and returns `None`, which causes the server to return a 503. The webui then shows:

> "The server does not have an external session provider configured. Install `gptme-sessions` and restart the server to enable this feature."

This message is misleading — `gptme-sessions` IS installed, just in the wrong way.

Reported in: gptme/gptme#1922 (comment by @ErikBjare)

## Fix

1. **Server-side**: Add `_warn_if_cli_only()` — when the import fails, check if the `gptme-sessions` CLI is in PATH. If it is, log an actionable warning explaining the correct install method.

2. **Webui**: Update the 503 error message to clarify that `uv tool install` creates an isolated venv, and the package must be installed in gptme's Python environment.

## How to correctly install gptme-sessions

```bash
# Into gptme's venv (after activating it):
pip install gptme-sessions

# Or with uv, in gptme's project dir:
uv add gptme-sessions

# NOT this (creates isolated venv):
uv tool install gptme-sessions  # ❌
```

### DIFF
--- a/gptme/server/external_sessions.py
+++ b/gptme/server/external_sessions.py
@@ -151,4 +151,26 @@ def get_external_session_provider() -> ExternalSessionProvider | None:
     try:
         return ExternalSessionProvider()
     except ImportError:
+        _warn_if_cli_only()
         return None
+
+
+def _warn_if_cli_only() -> None:
+    """Log a helpful warning when gptme-sessions CLI is in PATH but not importable.
+
+    This happens when gptme-sessions is installed via ``uv tool install`` (isolated
+    venv) rather than into gptme's own Python environment.
+    """
+    import shutil
+
+    if shutil.which("gptme-sessions"):
+        logger.warning(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile's P1 finding in 6b6217a54:

Added `@functools.lru_cache(maxsize=1)` to `get_external_session_provider()`. This caches the result after the first call, so the `ImportError` catch and `_warn_if_cli_only()` run at most once per process lifetime — no log flooding on repeated API requests.

PR #2097, opened 2026-04-10
feat(eval): add retry-with-backoff behavioral scenario (23rd)

TimeToBuildBob:
## Summary

Adds the 23rd behavioral eval scenario: **retry-with-backoff** — tests agent ability to implement retry logic with exponential backoff for transient API errors.

## What the scenario tests

The agent receives a failing HTTP fetcher (`fetcher.py`) and must:

1. **Retry on transient errors** — `ConnectionError`/`TimeoutError` trigger retry
2. **Use exponential backoff** — `time.sleep` with increasing delays (`base_delay * 2^attempt`)
3. **Set max retry limit** — `range(max_retries + 1)` or equivalent to prevent infinite loops
4. **Catch specific exceptions** — No bare `except:`, proper `HTTPError` handling
5. **Fix test bugs** — Mock responses use JSON `true` but assertions expect Python `True`

## Checkers (5)

| Checker | What it verifies |
|---------|-----------------|
| `all tests pass` | All 5 pytest tests pass |
| `has retry loop` | For/while loop with "retri" keyword present |
| `exponential backoff` | `time.sleep` + `**` (exponentiation) for increasing delays |
| `catches specific exception` | `ExceptHandler` with non-None type (no bare except) |
| `max retries limit` | `range()` call or `max_retries`/`MAX_RETRIES` variable |

## Files changed

- `gptme/eval/suites/behavioral/retry_with_backoff.py` — New scenario module
- `tests/test_eval_behavioral_solutions.py` — Reference solution + docstring update (22→23)

## Verification

All 23 reference solution tests pass (25/25 including smoke tests).

### DIFF
--- a/gptme/eval/suites/behavioral/retry_with_backoff.py
+++ b/gptme/eval/suites/behavioral/retry_with_backoff.py
@@ -1,0 +1,253 @@
+"""Behavioral scenario: retry-with-backoff."""
+
+import ast
+from typing import TYPE_CHECKING
+
+from ._common import parse_python_source
+
+if TYPE_CHECKING:
+    from gptme.eval.types import EvalSpec
+
+
+def _get_source(ctx, filename: str = "fetcher.py") -> str:
+    content = ctx.files.get(filename, "")
+    if isinstance(content, bytes):
+        content = content.decode()
+    return content
+
+
+def check_tests_pass(ctx):
+    """All tests should pass after adding retry logic."""
+    return ctx.exit_code == 0 and "failed" not in ctx.stdout.lower()
+
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing as superseded by #2095. The `retry-with-backoff` behavioral scenario was already merged into master via #2095 (using an `api_client.py`/requests-based implementation). This PR adds a different implementation using `fetcher.py`/urllib, but both define the same scenario name `"retry-with-backoff"` so they cannot coexist. The conflict is unresolvable without renaming one — and since #2095 is already merged, closing this one.

PR #2134, opened 2026-04-14
fix(server): handle file deletion races and bound attachment name allocation

TimeToBuildBob:
## Summary
- Skip files deleted between directory iteration and `stat()` in `list_directory()`, preventing 500 errors during concurrent file operations
- Distinguish "conversation not found" from "file not found" in browse/preview/download/serve endpoints by catching `FileNotFoundError` from `LogManager.load()` separately — previously all `FileNotFoundError` errors were reported as "Conversation not found"
- Cap `allocate_attachment_path` loop at 10,000 iterations to prevent CPU exhaustion from naming collisions during concurrent uploads

## Test plan
- [x] All 47 existing server v2 tests pass
- [x] Typecheck clean (no new errors)
- [ ] Manual: delete a file in workspace while browsing → should get 404 "File not found" instead of 500
- [ ] Manual: verify upload with many duplicate filenames hits the cap gracefully

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -146,7 +146,7 @@ def api_conversation_events(conversation_id: str):
                         },
                         "auto_confirm": te.auto_confirm,
                     }
-                    for tid, te in session.pending_tools.items()
+                    for tid, te in list(session.pending_tools.items())
                     if te.status == ToolStatus.PENDING
                 ],
             }
--- a/gptme/server/workspace_api.py
+++ b/gptme/server/workspace_api.py
@@ -377,7 +377,7 @@ def upload_files(conversation_id: str):
                     attachments_dir, filename, reserved_names
                 )
             except ValueError as e:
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Addressed the two P2 findings from Greptile:

1. **`upload_files` inner `LogManager.load()` guard** (comment 1): Added inner `try/except FileNotFoundError` around `LogManager.load()` to match the pattern used in `browse_workspace`, `preview_file`, `download_file`, and `serve_file`. Now the outer `FileNotFoundError` handler correctly catches file-level errors and returns "File not found".

2. **HTTP 400 for name exhaustion** (inline comment, line 193): Wrapped the `allocate_attachment_path` call in its own `try/except ValueError` that returns **507 Insufficient Storage** — the server ran out of candidate names, not a client error. The outer `except ValueError` is still there for any other validation errors.

Not fixing the TOCTOU (finding 3) — that's a pre-existing design concern and fixing it properly (atomic open with `'xb'`) is out of scope for this PR.

PR #2136, opened 2026-04-14
fix(server): return 404 when patching missing conversation config

TimeToBuildBob:
## Summary
- return a structured 404 instead of a 500 when PATCH /api/v2/conversations/<id>/config targets a missing conversation
- add a regression test for the missing-conversation config PATCH path

## Testing
- uv run pytest tests/test_server_v2.py -q
- uv run ruff check gptme/server/api_v2.py tests/test_server_v2.py

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1021,7 +1021,14 @@ def api_conversation_config_patch(conversation_id: str):
     tools = get_tools()
 
     # Update system prompt with new tools
-    manager = LogManager.load(conversation_id, lock=False)
+    try:
+        manager = LogManager.load(conversation_id, lock=False)
+    except FileNotFoundError:
+        return (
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the P1 ordering bug Greptile identified.

**Root cause**: The `FileNotFoundError` catch was placed after `ChatConfig.save()`, `set_config()`, and `init_tools()` had already executed — creating an orphaned directory + `config.toml` and overwriting process-wide config/tool state before returning 404.

**Fix** (7e4f84b): Moved the existence check to before any side-effecting operations, mirroring the DELETE endpoint's pattern. Checks for `conversation.jsonl` specifically (not just the directory) to correctly identify missing conversations even if an orphaned directory exists.

**Test update**: Extended the test to also assert `logdir` was not created on a 404 response, so this regression class won't slip through again.

TimeToBuildBob:
@greptileai review

PR #2137, opened 2026-04-14
fix(server): return 404 for missing conversation avatar and config

TimeToBuildBob:


TimeToBuildBob:
Addressed Greptile's state-consistency feedback: moved `LogManager.load()` before `ChatConfig.save()` and `set_config()` in `api_conversation_config_patch()`.

Now if the conversation directory exists but the log file is missing/corrupt, we bail with 404 before any side effects (no orphaned config, no stale global state).

Fixed in c07d3b7.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto master after #2136 merged. Key changes in the squash:

- Removed the `api_conversation_config_patch` 404 guard (already in #2136) and the duplicate `test_v2_chat_config_update_missing_conversation_returns_404` test
- Kept the `LogManager.load()` early load (before side effects) with `try/except FileNotFoundError` — covers the edge case where logdir exists but the log file is missing/corrupt
- Fixed `test_conversation_agent_avatar_rejects_non_image_extension` in `test_server_path_traversal.py`: that test mocked `ChatConfig.load_or_create` but didn't create the logdir, so our new `logdir.exists()` guard intercepted with 404 before the security check ran — now monkeypatches `get_logs_dir` to a `tmp_path`-backed dir and creates the conversation logdir

PR #2138, opened 2026-04-14
fix(server): add missing validation in POST message and config PATCH

TimeToBuildBob:
## Summary
- **POST /conversations/<id>**: Add `_validate_conversation_id()` call — this was the only conversation endpoint missing path traversal validation (CWE-22 defense-in-depth gap)
- **PATCH /conversations/<id>/config**: Add `sess.generating` guard returning 409 — config PATCH rewrites system messages and mutates global state, which races with active generation (message PATCH and DELETE already had this guard)

## Test plan
- [x] New parametrized test: POST with traversal payloads rejected (3 cases)
- [x] New test: config PATCH returns 409 when session is generating
- [x] Full test suite: 110 passed, 4 skipped
- [x] ruff clean

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -481,6 +481,9 @@ def api_conversation_put(conversation_id: str):
 )
 def api_conversation_post(conversation_id: str):
     """Append a message to a conversation."""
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #2168, opened 2026-04-17
fix(eval): isolate user context from benchmark runs

### LINKED ISSUE #2167: gptme-eval leaks user-level prompt config into eval runs
**Summary**

`gptme-eval` isolates *workspace* per test (temp logdir, fresh files) but inherits the user's global gptme config — so `~/.config/gptme/config.toml`'s `[prompt] about_user`, `response_preference`, and `files = [...]` leak into every eval. This silently biases results: a user with `files = ["~/work/my-notes.md"]` gets different eval behaviour than a user with an empty config, even on supposedly identical benchmark runs.

**Repro**

```toml
# ~/.config/gptme/config.toml
[prompt]
files = ["~/some-notes.md"]  # contains "/home/you/..." paths
```

```bash
gptme-eval -m "openai/gpt-4o@markdown" hello
```

The eval prompt is literally "write a script hello.py which prints Hello, world!". With a pollution file that mentions absolute paths, smaller / more suggestible models end up writing `hello.py` to the path they inferred from the user-config file instead of the eval workspace — causing the eval to fail not on capability but on environment leakage.

**Observed on**

`gptme v0.31.0+73c3d28b`, `qwen/qwen3.6-35b-a3b` (Q4_K_M) via LM Studio:

- markdown format: model wrote `hello.py` to `/home/bob/bob/` (my workspace, referenced in my auto-included notes file) → ❌ FAIL
- xml format: model correctly wrote to eval workspace → ✅ PASS
- tool (native API) format: 180s timeout → ❌ FAIL

The markdown failure is *purely* due to the leak; when I grep out the absolute path from the auto-included file, the model gets it right.

**Source**

[`gptme/prompts.py:682-706`](https://github.com/gptme/gptme/blob/master/gptme/prompts.py) — `prompt_workspace()` reads `get_config().user.prompt.files` and injects them unconditionally. [Line 614-622](https://github.com/gptme/gptme/blob/master/gptme/prompts.py) also loads `AGENT_FILES` from `~/.config/gptme/` into every workspace.

**Proposal**

Add `--no-user-context` (or equivalent) to `gptme-eval`, which:
- Skips user-level `[prompt] files`
- Skips user-level `AGENTS.md` / `CLAUDE.md` in `~/.config/gptme/`
- Still loads eval-specific system prompts and tool instructions
- Default to on for `gptme-eval` (evals should be reproducible across users)

Alternatively: `gptme-eval` could use an isolated `GPTME_CONFIG` containing only the provider / API-key blocks the eval needs.

Happy to do the PR if the direction is reasonable.

TimeToBuildBob:
Fixes #2167.

## Summary
- add an `include_user_context` switch to prompt construction so callers can skip `~/.config/gptme` prompt files and user-level agent instructions
- make `gptme-eval` default to isolated runs via `--user-context/--no-user-context`
- cover the prompt-layer skip, eval CLI flag propagation, and GPTMe eval-agent behavior with tests

## Verification
- `PYTHONPATH=/tmp/worktrees/gptme-2167 /home/bob/gptme/.venv/bin/python -m pytest tests/test_prompts.py -q`
- `PYTHONPATH=/tmp/worktrees/gptme-2167 /home/bob/gptme/.venv/bin/python -m pytest tests/test_prompts.py::test_prompt_workspace_can_skip_user_context tests/test_eval.py::test_eval_cli_user_context_flag_propagation tests/test_eval.py::test_eval_agent_passes_user_context_flag -q`
- `PYTHONPATH=/tmp/worktrees/gptme-2167 /home/bob/gptme/.venv/bin/python -m ruff check gptme/prompts/workspace.py gptme/prompts/__init__.py gptme/eval/agents/__init__.py gptme/eval/run.py gptme/eval/main.py tests/test_prompts.py tests/test_eval.py`
- `mypy --ignore-missing-imports --check-untyped-defs gptme/eval/main.py gptme/prompts/__init__.py tests/test_eval.py tests/test_prompts.py gptme/eval/agents/__init__.py gptme/prompts/workspace.py gptme/eval/run.py`

## Notes
- `tests/test_eval.py -q` still has an unrelated existing failure in `test_act_process_maps_subprocess_timeout_to_timeout_result` due `os.setpgrp()` permissions in this environment, so I verified the changed eval coverage directly instead of claiming the whole file is clean.

### DIFF
--- a/gptme/eval/agents/__init__.py
+++ b/gptme/eval/agents/__init__.py
@@ -22,6 +22,7 @@ class Agent:
     tool_format: ToolFormat
     tools: list[str] | None = None
     system_prompt: str | None = None
+    include_user_context: bool = False
     log_dir: Path
     workspace_dir: Path
     use_docker: bool = False
@@ -32,12 +33,14 @@ class Agent:
         tool_format: ToolFormat = "markdown",
         tools: list[str] | None = None,
         system_prompt: str | None = None,
+        include_user_context: bool = False,
         use_docker: bool = False,
     ):
         self.model = model
         self.tool_format = tool_format
         self.tools = tools
         self.system_prompt = system_prompt
+        self.include_user_context = include_user_context
         self.use_docker = use_docker
 
         _id = uuid.uuid4().hex[:8]
@@ -137,7 +140,8 @@ class GPTMe(Agent):
             workspace=self.workspace_dir,
             prompt=self.system_prompt or "full",  # this only replaces the base prompt
             interactive=False,  # eval agents run non-interactively
-            context_mode="selective",  # skip workspace files to prevent contamination from user's gptme config
+            context_mode="selective",  # skip workspace files/context_cmd in evals
+            include_user_context=self.include_user_context,
         )
 
         # Modify the first (core) system prompt to add eval-specific instructions
--- a/gptme/eval/agents/claude_code.py
+++ b/gptme/eval/agents/claude_code.py
@@ -87,6 +87,11 @@ class ClaudeCodeAgent(Agent):
 
     def _act_local(self, store: FileStore, prompt: str) -> Files:
         """Execute Claude Code directly on the host."""
+        if not self.include_user_context:
+            logger.warning(
+                "include_user_context=False has no effect for ClaudeCodeAgent — "
+                "the Claude Code CLI reads its own user config (e.g. ~/.claude/) unconditionally."
+            )
         claude_bin = shutil.which("claude")
         if claude_bin is None:
             raise FileNotFoundError(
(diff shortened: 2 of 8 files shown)


TimeToBuildBob:
Addressed Greptile P1 and P2 findings in eff646ac:

**P1 fixed** – gated `find_agent_files_in_tree` behind `include_user_context`. The unconditional tree walk would pick up `~/AGENTS.md` whenever the eval workspace lived under the home directory (the default for `get_logs_dir()`). Added a regression test (`test_prompt_workspace_skip_home_agents_md`) that places the workspace under a fake home dir to cover this path.

**P2 addressed** – added `logger.warning` in `ClaudeCodeAgent._act_local` so users know `include_user_context=False` has no effect for claude-code/ models (the `claude` CLI reads `~/.claude/` unconditionally).

TimeToBuildBob:
@greptileai review

PR #2205, opened 2026-04-22
feat(agent): add gptme-agent scan alias

TimeToBuildBob:
## Summary
- expose the live agent scanner as `gptme-agent scan`
- reuse the existing `gptme-util agents scan` implementation via a shared helper
- add agent CLI coverage for the new alias

## Why
`gptme-util agents scan` exists on `main` but is easy to miss, and the installed `v0.31.0` tool does not include it yet. Adding `gptme-agent scan` makes the feature more discoverable once released.

### DIFF
--- a/gptme/agent/cli.py
+++ b/gptme/agent/cli.py
@@ -2,6 +2,7 @@
 CLI commands for gptme agent management.
 
 Usage:
+    gptme-agent scan                # Show live agent processes across runtimes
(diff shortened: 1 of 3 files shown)


ErikBjare:
@TimeToBuildBob how come this isn't being auto-merged by your project-monitoring? I see 5/5 on Greptile...

PR #2206, opened 2026-04-22
fix(cli): reduce agents scan latency

TimeToBuildBob:
## Summary
Reduce `gptme-util agents scan` latency on hosts with large Claude transcript histories.

This fixes the worst offender I found on Bob's VM:
- a stale 4-day Claude Code process under `/home/bob/bob`
- `~/.claude/projects/-home-bob-bob/` containing about 18,978 transcript files
- `_resolve_claude_conversation_id()` dominating scan time by indexing that history even though the stale process is hidden by default

## Changes
- skip Claude conversation-id resolution for processes already classified as stale
- cap Claude transcript indexing to the most recently touched files instead of whole-history scans
- lazy-load heavyweight `gptme-util` subcommands/model helpers so `agents scan` does not import MCP/model code it never uses

## Verification
- `python -m pytest gptme/hooks/tests/test_workspace_agents.py tests/test_util_cli.py -q`
- `ruff check gptme/cli/util.py gptme/hooks/workspace_agents.py gptme/hooks/tests/test_workspace_agents.py`

## Measurements
On Bob's host:
- before: `uv run gptme-util agents scan` about `11.96s` real
- after: `uv run gptme-util agents scan` about `6.47s` real
- after: `scan_agents()` itself about `0.126s`

Under instrumentation before the fix, `_resolve_claude_conversation_id()` accounted for `17.97s` of `18.56s` inside the scanner, so this was not a PID-count problem.

PR #2207, opened 2026-04-23
perf: lazy-load heavy imports to reduce CLI startup time 3.3s→1.0s

TimeToBuildBob:
## Summary

`gptme/__init__.py` eagerly imported `gptme.chat` on every process start, pulling in the full chain:
`chat → commands → llm → config → context → compress → openai/anthropic/opentelemetry`

This cost ~2.5s on every `gptme-util` invocation — even for lightweight subcommands like `agents scan` that don't need LLM clients at all.

- **Before**: `gptme-util agents scan` = **3.3s** real
- **After**: `gptme-util agents scan` = **1.0s** real (3.2× speedup)

Follows up on #2206 which brought the scan itself down to ~0.1s; the remaining time was all import overhead.

## Changes

| File | Change |
|------|--------|
| `gptme/__init__.py` | `__getattr__`-based lazy loading for `chat`, `Codeblock`, `LogManager`, `Message`, `get_prompt`; `TYPE_CHECKING` block preserves type safety |
| `gptme/context/__init__.py` | Lazy-load `compress.strip_reasoning` (imports `Message`); keep config/selector/task_analyzer eager |
| `gptme/context/selector/__init__.py` | Lazy-load `FileItem`, `select_relevant_files`, `HybridSelector`, `LLMSelector`, `RuleBasedSelector` (all import `Message` or heavy LLM deps); keep lightweight base/config/file_config eager |
| `gptme/message.py` | Move `get_default_model` import inside `cost()` — breaks `message → llm.models → config → context → compress → message` cycle |
| `gptme/telemetry.py` | Move `get_model` import inside `_calculate_llm_cost()` — breaks `codeblock → telemetry → llm → config → context` cycle |
| `gptme/util/prompt.py` | Move `get_default_model` import inside its using function — same family of latent circular imports |

## Why the lazy imports were needed (not just `__init__.py`)

Making `gptme/__init__.py` lazy exposed latent circular dependencies that the previous eager-import order had been silently masking. The cycles:

- `message.py` imported `gptme.llm.models` at module level → triggered `llm.__init__` → `config` → `context.compress` → `Message` (back to partially-loaded `message.py`)
- `telemetry.py` imported `gptme.llm.models` at module level → same config→context chain
- `util/prompt.py` imported `gptme.llm.models` at module level → same chain

The fix moves each `get_model`/`get_default_model` call inside the single function that uses it. This is the standard Python pattern for breaking circular imports and is strictly cleaner than the previous top-level imports.

## Test plan

- [x] 332 fast tests pass (`pytest tests/ -q -k "not slow and not integration ..."`)
- [x] `gptme-util agents scan` produces correct output
- [x] `from gptme import chat, Message, Codeblock, LogManager, get_prompt` still works
- [x] Pre-commit hooks pass

### DIFF
--- a/gptme/__init__.py
+++ b/gptme/__init__.py
@@ -1,14 +1,5 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from .chat import chat
-    from .codeblock import Codeblock
-    from .logmanager import LogManager
-    from .message import Message
-    from .prompts import get_prompt
-
 __all__ = ["Codeblock", "LogManager", "Message", "__version__", "chat", "get_prompt"]
 
 _lazy: dict[str, tuple[str, str]] = {
--- a/gptme/__init__.py
+++ b/gptme/__init__.py
@@ -1,13 +1,34 @@
-from .chat import chat
-from .codeblock import Codeblock
-from .logmanager import LogManager
-from .message import Message
-from .prompts import get_prompt
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .chat import chat
+    from .codeblock import Codeblock
+    from .logmanager import LogManager
+    from .message import Message
+    from .prompts import get_prompt
 
 __all__ = ["Codeblock", "LogManager", "Message", "__version__", "chat", "get_prompt"]
 
+_lazy: dict[str, tuple[str, str]] = {
+    "chat": (".chat", "chat"),
+    "Codeblock": (".codeblock", "Codeblock"),
+    "LogManager": (".logmanager", "LogManager"),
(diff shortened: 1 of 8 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the CI failure: `test_package_exports.py` was failing under `pytest-xdist` because `importlib.reload()` re-executes module code but doesn't clear `__dict__`. Lazy attrs cached by `__getattr__` in one worker's earlier tests persisted after reload, causing the "attr not in `__dict__`" assertions to fail.

Fix: pop all keys in `module._lazy` from `__dict__` after reload, before asserting they're absent. Both tests now pass.

Commit: `1f6c64a10`

PR #2256, opened 2026-04-27
fix(agents_md_inject): skip re-injection when content already loaded from different path

### LINKED ISSUE #2254: bug: duplicate agent instruction file injection on cwd change
When the working directory changes to a different checkout of the same repo (e.g. a worktree), the AGENTS.md file is re-injected even though it has identical content to one already in context.

**Observed**: AGENTS.md injected from `~/Programming/gptme/AGENTS.md`, then the same content re-injected from `/tmp/worktrees/gptme-version-fix/AGENTS.md` when cwd changed.

**Suggested fix**: Hash (or content-compare) every injected file and skip injection if the content is already present. This could apply to all injected files generally, not just AGENTS.md — any duplicate content is pure token waste.

The `_loaded_agent_files_var` ContextVar already tracks loaded filenames, but it tracks by path, not by content hash. Two different paths with identical content will both be loaded.

TimeToBuildBob:
## Summary

- **Root cause**: `_loaded_agent_files_var` tracked loaded agent files by *path*, so switching cwd to a git worktree with the same `AGENTS.md` content caused a duplicate injection.
- **Fix**: Also track a content hash (prefixed `ch:`) in the same set. Both `prompt_workspace()` and `on_cwd_changed()` check the hash before injecting; `_derive_loaded_files_from_log()` (server-mode fallback) also extracts hashes from complete `<agent-instructions>` blocks in the log.
- No change to the injected message format.

Fixes #2254.

## Test plan

- [ ] New test `test_skips_injection_when_content_already_loaded`: two directories, same `AGENTS.md` content — cwd change to the second does NOT re-inject.
- [ ] New test `test_different_content_in_different_path_is_injected`: sanity check that *different* content at a new path still gets injected.
- [ ] All 17 existing `test_agents_md_inject` tests continue to pass.

### DIFF
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -7,9 +7,12 @@ from typing import TYPE_CHECKING
 from ..config import config_path, get_config, get_project_config
 from ..message import Message
 from ..util.context import md_codeblock
+from ..util.context_dedup import _content_hash
 from ..util.tree import get_tree_output
 from . import AGENT_FILES, DEFAULT_CONTEXT_FILES, _loaded_agent_files_var
 from .context_cmd import get_project_context_cmd_output
+
+_HASH_PREFIX = "ch:"
 
 if TYPE_CHECKING:
     from ..util.uri import FilePath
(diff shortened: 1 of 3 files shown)


gptme/prompts/workspace.py
```diff
@@ -7,10 +7,13 @@
 from ..config import config_path, get_config, get_project_config
 from ..message import Message
 from ..util.context import md_codeblock
+from ..util.context_dedup import _content_hash
 from ..util.tree import get_tree_output
 from . import AGENT_FILES, DEFAULT_CONTEXT_FILES, _loaded_agent_files_var
 from .context_cmd import get_project_context_cmd_output
 
+_HASH_PREFIX = "ch:"
```
ErikBjare, line 15: Duplicated constant, should probably be shared with gptme/hooks/agents_md_inject.py

PR #2257, opened 2026-04-27
fix(cli): skip context_cmd when opening any existing conversation, not just --resume

### LINKED ISSUE #2252: bug: context_cmd runs when resuming a conversation
When resuming an existing conversation (e.g. via `gptme --resume`), the `context_cmd` from `gptme.toml` runs again. It should only run when starting a new conversation.

**Observed**: Resuming a conversation shows "Using project context command: scripts/context.sh" and runs it (which can fail or add stale context).

**Expected**: On resume, the context_cmd output from the original session start should be preserved, not re-executed.

This was observed when resuming a Bob session from CLI after a gptme-tauri crash — the context_cmd ran and failed (missing `uv` in PATH on macOS).

TimeToBuildBob:
## Root cause

`is_existing_conversation` in `cli/main.py` had an extra `resume and` guard:

```python
is_existing_conversation = (
    resume and log_file.exists() and log_file.stat().st_size > 0
)
```

This correctly skips `context_cmd` when using `--resume`, but `gptme --name <existing-name>` also opens an existing conversation without setting `resume=True`. In that case the check returned `False`, so `get_prompt` ran again and re-executed `context_cmd` — which can fail (e.g. missing `uv` in PATH on macOS) or inject stale context on top of an already-full log.

## Fix

Remove the `resume` prerequisite. The only thing that matters is whether `conversation.jsonl` already has content:

```python
is_existing_conversation = log_file.exists() and log_file.stat().st_size > 0
```

New conversations (random or named) won't have a file yet, so `get_prompt` still runs for them. Existing conversations — opened via `--resume`, `--name`, or the interactive picker — now all skip the redundant prompt regeneration.

## Testing

`tests/test_prompts.py` and `tests/test_cli.py` — 37 passed, 0 failures.

Fixes #2252.

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -583,18 +583,14 @@ def main(
     logger.debug(f"Using tools: {config.chat.tools}")
     tools = init_tools(config.chat.tools)
 
-    # Check if we're resuming an existing conversation
+    # Check if we're opening an existing conversation (via --resume, --name, or pick)
     # If so, skip generating initial messages (including expensive context_cmd)
     # as they're already in the loaded log
     log_file = logdir / "conversation.jsonl"
-    is_existing_conversation = (
-        resume and log_file.exists() and log_file.stat().st_size > 0
-    )
+    is_existing_conversation = log_file.exists() and log_file.stat().st_size > 0
 
     if is_existing_conversation:
-        logger.debug(
-            "Resuming existing conversation, skipping initial prompt generation"
-        )
+        logger.debug("Existing conversation found, skipping initial prompt generation")
         initial_msgs = []
     else:
         # Infer context mode: --context-include implies selective mode


ErikBjare:
Is this really true? No risk of anything pre-writing the log during setup/init?

TimeToBuildBob:
Good catch to double-check. I traced the code:

1. `get_logdir(name)` only calls `logdir.mkdir(parents=True, exist_ok=True)` — creates the directory, not `conversation.jsonl`
2. Between `logdir` assignment and the `is_existing_conversation` check: `setup_config_from_cli`, `init_telemetry`, and `init_tools` are called — none of them write to `logdir`
3. `LogManager` (the only writer of `conversation.jsonl`) is instantiated inside `chat()`, which runs *after* the check

So for a new `--name <new-name>` conversation: directory gets created, `log_file.exists()` → `False` → `get_prompt` runs as expected. For an existing name: the file already has content → skip. No race condition possible within the same process.

PR #2259, opened 2026-04-27
fix(cli): skip context_cmd when resuming via --name <existing>

### LINKED ISSUE #2252: bug: context_cmd runs when resuming a conversation

TimeToBuildBob:
## Summary

Fixes ErikBjare/bob#... — actually, fixes gptme/gptme#2252.

The `is_existing_conversation` guard at `gptme/cli/main.py` previously required `--resume` to be set:

```python
is_existing_conversation = (
    resume and log_file.exists() and log_file.stat().st_size > 0
)
```

But users can also resume by passing `--name <existing>` directly. In that path, `get_prompt()` ran every time, which executes `context_cmd` as a side effect — even though `LogManager.load` would discard the resulting `initial_msgs` (`log.messages or initial_msgs or []`).

The user-visible bug is that `context_cmd` (which can be a slow shell command, fail with missing PATH entries, or add stale context) runs on every resume, even when there's nothing to do with the result.

## Fix

Drop the `resume and` requirement. If the log file already has content, skip initial-prompt generation regardless of how the user got there.

```python
is_existing_conversation = log_file.exists() and log_file.stat().st_size > 0
```

Behavior matrix:

| Invocation | Log state | Before | After |
|---|---|---|---|
| `gptme --resume` | has content | skip ✓ | skip ✓ |
| `gptme --resume` | empty/missing | run | run (no change) |
| `gptme --name existing` | has content | **run (bug)** | **skip (fixed)** |
| `gptme --name new` | missing | run ✓ | run ✓ |

## Test plan

- [x] New test `test_resume_skips_context_cmd` in `tests/test_cli.py` — pre-creates a log with content, runs `gptme --name <existing>`, asserts a marker file from `context_cmd = "touch <marker>"` was NOT created.
- [x] Verified the test fails on un-fixed code (assertion fires because marker was created).
- [x] Verified the test passes with the fix.
- [x] Full `tests/test_cli.py` suite: 23 passed, 8 skipped.

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -585,11 +585,14 @@ def main(
 
     # Check if we're resuming an existing conversation
     # If so, skip generating initial messages (including expensive context_cmd)
-    # as they're already in the loaded log
+    # as they're already in the loaded log.
+    # The check is independent of `--resume` because users can also resume via
+    # `--name <existing>`. Either way, if the log already has content, we
+    # should not re-run context_cmd (whose result LogManager.load would
+    # discard anyway, but whose side effect — running a shell command — is
+    # what we want to avoid).
     log_file = logdir / "conversation.jsonl"
-    is_existing_conversation = (
-        resume and log_file.exists() and log_file.stat().st_size > 0
-    )
+    is_existing_conversation = log_file.exists() and log_file.stat().st_size > 0
 
     if is_existing_conversation:
         logger.debug(
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -492,6 +492,39 @@ def test_comma_separated_choice_minus_prefix():
         csc.convert("-nonexistent", None, None)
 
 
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Closing as superseded — #2257 (same fix, Greptile 5/5, full CI green) was merged first. The regression test here is good; can be added in a follow-up if desired.

PR #2261, opened 2026-04-27
fix(tauri): kill gptme-server subprocesses and orphans on exit

### LINKED ISSUE #2260: bug(tauri): gptme-server sidecar persists after gptme-tauri quits
## Problem

```
```

This means:

## Expected Behaviour

## Likely Fix

## Related

TimeToBuildBob:
## Summary

Fixes #2260 — two gaps in `cleanup_server_process` that left `gptme-server` processes alive after the Tauri app exits.

### Gap 1: subprocesses survive `child.kill()`

`child.kill()` sends SIGKILL to the *direct* child only. `gptme-server` may spawn subprocesses (e.g. uvicorn workers), which become orphans that keep port 5700 occupied.

**Fix**: new `kill_subprocesses(pid)` called before `child.kill()`:
- Unix: `pkill -9 -P <pid>` (direct children of the sidecar)
- Windows: `taskkill /F /T /PID <pid>` (full process tree)

### Gap 2: reuse path (from #2258) leaves no tracked handle

When `spawn_server_sidecar()` detects a responsive existing server and returns `Ok(())` without spawning, `ServerProcess` stays `None`. On exit, `cleanup_server_process` found no child and did nothing — the leftover server persisted.

**Fix**: new `kill_server_on_port(port)` called when `guard` is `None`:
- Unix: `lsof -ti:<port>` to find PIDs, then `kill -9` each (plus their children)
- Windows: `netstat -ano` to find PID, then `taskkill /F /T /PID`

Both helpers are `#[cfg(unix)]`/`#[cfg(windows)]` gated with a no-op stub for any other platform.

## Test plan

- [ ] `cargo test` passes on a desktop build (requires GTK3 dev libs — CI will validate)
- [ ] Build gptme-tauri; normal startup + quit leaves no `gptme-server` in `ps`
- [ ] Kill `gptme-tauri` via `pkill`, restart — app starts without dialog; quit leaves no orphan processes
- [ ] Verify port 5700 is free after quitting the app via Cmd+Q / dock quit / X button

### DIFF
--- a/tauri/src-tauri/src/lib.rs
+++ b/tauri/src-tauri/src/lib.rs
@@ -496,6 +496,10 @@ fn kill_subprocesses(pid: u32) {
 // (the spawn_server_sidecar reuse path).
 #[cfg(unix)]
 fn kill_server_on_port(port: u16) {
+    // lsof returns all PIDs with any socket on the port, including established
+    // client connections (e.g. the Tauri WebView itself).  Skip our own PID to
+    // avoid self-killing the Tauri process during cleanup.
+    let my_pid = std::process::id();
     let output = match std::process::Command::new("lsof")
         .args(["-ti", &format!(":{}", port)])
         .output()
@@ -512,6 +516,10 @@ fn kill_server_on_port(port: u16) {
     };
     for pid_str in String::from_utf8_lossy(&output.stdout).split_whitespace() {
         if let Ok(pid) = pid_str.parse::<u32>() {
+            if pid == my_pid {
+                log::debug!("Skipping self (PID {}) in port {} cleanup", pid, port);
+                continue;
+            }
             log::info!("Killing orphan gptme-server PID {} on port {}", pid, port);
             kill_subprocesses(pid);
             let _ = std::process::Command::new("kill")
@@ -523,7 +531,9 @@ fn kill_server_on_port(port: u16) {
 
 #[cfg(windows)]
 fn kill_server_on_port(port: u16) {
-    // netstat -ano lists all connections with PIDs; find the LISTENING entry.
+    // netstat -ano columns: Proto  LocalAddress  ForeignAddress  State  PID
+    // Match the local-address field (col[1]) exactly so ":5700" does not
+    // accidentally match ":57001" via substring search.
     let output = match std::process::Command::new("netstat")
         .args(["-ano"])
         .output()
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed both P1s in 0adae3a94:

**P1 — `stop_server` IPC leaves orphan subprocesses**: Added `kill_subprocesses(child.pid())` before `child.kill()` in `stop_server`, mirroring what `cleanup_server_process` already does.

**P1 — `else` branch kills unrelated process**: Refactored `ServerProcess` to carry `owns_port: Arc<AtomicBool>`. Set to `true` when we start or reuse a server, left `false` when startup fails with a non-responsive port conflict. The `kill_server_on_port` call in `cleanup_server_process` is now guarded by `owns_port.load()`, so we never kill a process we didn't start.

ErikBjare:
@greptileai review

TimeToBuildBob:
Addressed both remaining Greptile concerns in 497e585:

**Unix — lsof self-kill**: `lsof -ti :5700` returns all PIDs with sockets on that port, including gptme-tauri's own WebView (an established client connection). Added a `my_pid = std::process::id()` check and `continue` to skip our own PID.

**Windows — port substring match**: Replaced `line.contains(":5700")` with field-level parsing. Split by whitespace, take `col[1]` (the local address field, e.g. `0.0.0.0:5700`), and check `ends_with(":5700")` — so `:5700` no longer falsely matches `:57001`.

TimeToBuildBob:
@greptileai review

tauri/src-tauri/src/lib.rs
```diff
@@ -496,6 +496,10 @@ fn kill_subprocesses(pid: u32) {
 // (the spawn_server_sidecar reuse path).
 #[cfg(unix)]
 fn kill_server_on_port(port: u16) {
+    // lsof returns all PIDs with any socket on the port, including established
+    // client connections (e.g. the Tauri WebView itself).  Skip our own PID to
+    // avoid self-killing the Tauri process during cleanup.
+    let my_pid = std::process::id();
```
ErikBjare, line 506: Doesn't the server process use another child PID?

TimeToBuildBob:
Two remaining Greptile P1s addressed in 4e75a3c:

**P1 — `stop_server` leaves `owns_port` stale**: Added `state.owns_port.store(false, Ordering::Relaxed)` synchronously after `child.kill()` in `stop_server`. Without this, if the user stops via IPC and quits before the async `Terminated` handler fires, `cleanup_server_process` sees `guard==None` + `owns_port==true` and calls `kill_server_on_port` against whatever process later binds to port 5700.

**Unix — `lsof -ti` scope**: Switched to `lsof -ti :<port> -sTCP:LISTEN` which restricts output to the process actually *listening* on the port, excluding established client connections (WebView renderer, etc.). The `my_pid` self-kill guard is kept as defence-in-depth.

TimeToBuildBob:
@greptileai review

PR #2262, opened 2026-04-27
fix(tauri): catch PyInstaller orphan on app exit

### LINKED ISSUE #2260: bug(tauri): gptme-server sidecar persists after gptme-tauri quits

TimeToBuildBob:
## Summary

Fixes #2260 (reopened). After PR #2261 merged, Erik retested on `dev202604275` and two `gptme-server` processes still survived after `gptme-tauri` exited:

```
501 75115     1   0  7:29PM ??         0:00.42 .../gptme-server --cors-origin ...
501 75119 75115   0  7:29PM ??         0:01.28 .../gptme-server --cors-origin ...
```

PID 75115's parent is `1` (init), and PID 75119's parent is 75115 — classic PyInstaller onefile pattern: the launcher (75115) was reparented to init when `gptme-tauri` died, but the launcher itself wasn't killed, so it kept its Python child alive.

## Root cause

The `gptme-server` sidecar is a PyInstaller onefile bundle. The launcher process spawns the actual Python interpreter as a child; the launcher's PID is what `tauri-plugin-shell` reports as `child.pid()`. When the launcher dies (cleanly or via SIGKILL), the Python child can survive — reparented to init — and keep port 5700 bound.

`pkill -9 -P LAUNCHER_PID` from #2261 was supposed to catch the Python child, but two control-flow gaps let the orphan slip past:

1. **`cleanup_server_process` used `if/else if`** — `kill_server_on_port` (the `lsof :5700` fallback) only ran when no `CommandChild` was tracked. If `pkill -P` missed the Python child for any reason, nothing looked at port 5700 directly.

2. **`Terminated` handler was over-eager** — as soon as the launcher exited, the async handler unconditionally set `owns_port = false`. If a second cleanup pass fired (the `CloseRequested` → `ExitRequested` race on Cmd+Q), the `else if owns_port` branch was skipped and the orphan persisted.

## Fix

1. `cleanup_server_process` snapshots `owns_port` at entry and **always** runs `kill_server_on_port` when it was true — independent of whether a `CommandChild` was tracked. Foreign processes are still safe: the branch is skipped when `owns_port` is false (e.g. startup failed against a non-responsive foreign process — that path never sets `owns_port`).

2. The `Terminated` handler only clears `owns_port` when the port is actually free. If the launcher exits but port 5700 is still occupied (orphan case), it leaves `owns_port=true` so app exit cleanup can finish the job.

## Test plan

- [ ] CI compiles on macOS (Linux can't build the Tauri app — `gdk-3.0` not available on this VM)
- [ ] Erik builds gptme-tauri from this branch on macOS, normal startup + Cmd+Q leaves no `gptme-server` in `ps`
- [ ] Same flow but close window via X button (instead of Cmd+Q) — also leaves no orphans
- [ ] `lsof -i :5700 -sTCP:LISTEN` returns empty after app exit

## References

- #2260 — original bug
- #2261 — first attempt (this PR builds on it)
- #2238 — earlier `RunEvent::ExitRequested` cleanup wiring
- #2258 — reuse path that introduced the no-tracked-child case

### DIFF
--- a/tauri/src-tauri/src/lib.rs
+++ b/tauri/src-tauri/src/lib.rs
@@ -249,7 +249,23 @@ async fn spawn_server_sidecar(
                     if let Ok(mut guard) = state_for_output.lock() {
                         *guard = None;
                     }
-                    owns_port_for_output.store(false, Ordering::Relaxed);
+                    // PyInstaller onefile bundles use a launcher process that
+                    // spawns the actual Python interpreter as a child. When the
+                    // launcher dies (cleanly or via SIGKILL), the Python child
+                    // can survive — reparented to init — and keep port 5700
+                    // bound until something explicitly kills it.  Verify the
+                    // port is actually free before declaring the server gone;
+                    // otherwise leave owns_port=true so cleanup_server_process
+                    // catches the orphan on app exit (#2260).
+                    if is_port_available(GPTME_SERVER_PORT) {
+                        owns_port_for_output.store(false, Ordering::Relaxed);
+                    } else {
+                        log::warn!(
+                            "[gptme-server] Sidecar exited but port {} still in use — \
+                             likely an orphaned subprocess; deferring port cleanup to app exit",
+                            GPTME_SERVER_PORT
+                        );
+                    }
                     break;
                 }
                 _ => {}
@@ -444,6 +460,12 @@ fn cleanup_server_process(app: &tauri::AppHandle) {
         Some(s) => s,
         None => return,
     };
+
+    // Snapshot ownership before we start mutating state — kill_server_on_port
+    // must run unconditionally below if we own the port, regardless of whether
+    // we have a tracked child handle.
+    let owns_port_at_entry = state.owns_port.load(Ordering::Relaxed);
+
     let arc = state.child.clone();
     let mut guard = match arc.lock() {
         Ok(g) => g,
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
One additional fix to consider on top of this: call `api.prevent_exit()` in the `ExitRequested` handler.

Even with the unconditional `kill_server_on_port` fallback here, there's still a race where macOS can kill the Tauri process before the `pkill`/`child.kill()` syscalls are dispatched — making the cleanup a no-op. `prevent_exit()` blocks the OS from terminating the process until our handler returns:

```rust
if let tauri::RunEvent::ExitRequested { api, .. } = event {
    api.prevent_exit();
    cleanup_server_process(app_handle);
    app_handle.exit(0);  // fires RunEvent::Exit, not another ExitRequested
}
```

`AppHandle::exit(0)` fires `RunEvent::Exit` (not another `ExitRequested`), so no infinite loop. I had a PR open for this (#2263, now closed) — happy to add the commit here or open a follow-up once this lands.

TimeToBuildBob:
Incorporated `api.prevent_exit()` (commit 67757656c) — closed #2263 as superseded.

You're right that on macOS the OS can kill the Tauri process before `pkill` / `kill_server_on_port` reach the kernel; `prevent_exit()` blocks termination until cleanup returns, then `app_handle.exit(0)` triggers a clean exit (fires `RunEvent::Exit`, not another `ExitRequested`).

PR now combines both fixes:
1. `prevent_exit()` so the OS can't race us during cleanup
2. Terminated-handler defers `owns_port=false` when the port is still occupied
3. `cleanup_server_process` always runs `kill_server_on_port` when we ever owned the port

PR #2263, opened 2026-04-27
fix(tauri): prevent_exit in ExitRequested to guarantee server cleanup on macOS

### LINKED ISSUE #2260: bug(tauri): gptme-server sidecar persists after gptme-tauri quits

TimeToBuildBob:
## Problem

`gptme-server` orphan processes persist after app exit even with the #2261 fix (confirmed by Erik on dev202604275, which includes the #2261 merge commit `fe05ed5c`).

## Root Cause

The `ExitRequested` handler runs `cleanup_server_process` synchronously, but on macOS with Cmd+Q, the process-termination machinery can kill the Tauri process **before** our `pkill` and `child.kill()` syscalls have been dispatched to the kernel. Without `api.prevent_exit()`, there is no guarantee the handler completes before the process dies.

## Fix

**Primary**: Call `api.prevent_exit()` to block macOS from terminating the process until cleanup completes, then call `app_handle.exit(0)` for a clean exit. `AppHandle::exit(0)` fires `RunEvent::Exit` (not another `ExitRequested`), so there is no infinite loop.

**Secondary** (belt-and-suspenders in `cleanup_server_process`):
- Clear `owns_port` synchronously after both kill paths so a subsequent `ExitRequested` call (e.g. after `CloseRequested` already ran) is a true no-op.
- Always call `kill_server_on_port` as final fallback in the tracked-child path, in case `child.kill()` fails or a detached subprocess (e.g. double-forked uvicorn worker) survived `pkill`.

## Testing

Needs verification by Erik on macOS: quit `gptme-tauri` with Cmd+Q and confirm no `gptme-server` processes remain in `ps`.

Fixes #2260.

### DIFF
--- a/tauri/src-tauri/src/lib.rs
+++ b/tauri/src-tauri/src/lib.rs
@@ -421,19 +421,30 @@ pub fn run() {
         })
         .build(tauri::generate_context!())
         .expect("error while building tauri application")
-        .run(|_app_handle, _event| {
+        .run(|app_handle, event| {
             // ExitRequested fires on app-level exit paths that don't emit a
             // per-window CloseRequested first (macOS Cmd+Q, dock-quit, system
             // shutdown). Without this, the sidecar gptme-server outlives the
             // app and squats on port 5700 (gptme/gptme#2237).
             //
+            // We call api.prevent_exit() so macOS cannot kill the process
+            // before our pkill/kill commands finish — the race that caused #2260
+            // to persist even after #2261.  After cleanup we call exit(0) to
+            // trigger a clean exit; that fires RunEvent::Exit (not another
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Closing in favor of #2262 which addresses the root cause (PyInstaller orphan + unconditional port cleanup). I'll add a  suggestion on #2262 as a complementary fix for the exit-race case.

TimeToBuildBob:
Superseded by #2262, which now includes both the prevent_exit fix from this branch and the Terminated-handler / always-run-port-cleanup changes.

PR #2289, opened 2026-04-28
fix(webui): move Sheet safe-area from SheetHeader margin to SheetContent padding

TimeToBuildBob:
## Problem

PR #2287 applied `safe-area-inset-top` via `marginTop` on `SheetHeader` (a child element). Erik reported the hamburger navigation menu **still** shows the Navigation title and X close button under the Android status bar after that fix.

## Root cause

The `SheetHeader` `marginTop` approach is fragile: it relies on a **child** element to handle the safe area, but `SheetContent` in `MainLayout.tsx` uses `className="p-0"` which affects the layout context. More importantly, the fix doesn't mirror the approach that **did** work — `MenuBar`'s `paddingTop: env(safe-area-inset-top)` inline style on the **container element itself**.

Inline style on an element cannot be overridden by `className` (CSS specificity). Inline style on a **child** still leaves the container unprotected.

## Fix

Move safe-area handling from `SheetHeader.marginTop` → `SheetContent` inline style, mirroring `MenuBar`'s working approach:

- `SheetContent`: extract `style` prop, build side-aware safe-area padding object (left/right get top+bottom insets; top/bottom sides get their respective inset), merge with caller-supplied style
- `SheetHeader`: remove now-redundant `marginTop`
- `SheetFooter`: remove now-redundant `marginBottom`
- `SheetClose` X button: position unchanged — `top: calc(1rem + env(safe-area-inset-top))` correctly places it 1rem into the content area below the safe-area padding buffer

## Impact

- Left/right Sheet (hamburger nav, right sidebar): top and bottom safe areas applied
- Top/bottom Sheet variants: respective inset applied
- Desktop/non-notched devices: fallback is `0px`, no visual change
- Consumers using `p-0` override: safe-area still applied (inline > class)

Refs ErikBjare/bob#660.

### DIFF
--- a/webui/src/components/ui/sheet.tsx
+++ b/webui/src/components/ui/sheet.tsx
@@ -48,8 +48,7 @@ const sheetVariants = cva(
 );
 
 interface SheetContentProps
-  extends
-    React.ComponentPropsWithoutRef<typeof SheetPrimitive.Content>,
+  extends React.ComponentPropsWithoutRef<typeof SheetPrimitive.Content>,
     VariantProps<typeof sheetVariants> {}
 
 const SheetContent = React.forwardRef<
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile's P1 and P2 findings in b13f18953:

- **P1 fix**: Changed `paddingTop/Bottom: 'env(safe-area-inset-*, 0px)'` to `'max(1.5rem, env(safe-area-inset-*, 0px))'` — preserves the base `p-6` (1.5rem) padding on non-notched/desktop devices where the env() call evaluates to 0px.
- **P2 fix**: Moved `...style` before the safe-area values in the spread, so the safe-area padding wins over any caller-supplied inline style.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed Greptile's remaining P1 in 2aad6eff7: removed the `max(1.5rem, ...)` wrapper.

Greptile was right — `max(1.5rem, env(safe-area-inset-top, 0px))` forced 1.5rem top/bottom padding on desktop for all `p-0` consumers (both sheets in `MainLayout`), overriding their explicit zero padding. The `max()` was intended to protect default `p-6` consumers from losing padding, but there are no such consumers — every `SheetContent` usage in the codebase passes `p-0`. So the `max()` caused a real regression while protecting a theoretical case.

Reverted to plain `env(safe-area-inset-top, 0px)`:
- Desktop/non-notched: 0px → correct for `p-0` consumers
- Android/notched: actual inset → correct safe-area behavior

TimeToBuildBob:
@greptileai review

PR #2311, opened 2026-05-01
fix(llm): strip redundant openrouter/ prefix from proxy model IDs

TimeToBuildBob:
## Problem

When gptme is configured to use the gptme.ai Supabase proxy, models are fetched from the Supabase `models` edge function. That function returns model IDs **already prefixed** with `openrouter/` (e.g. `"openrouter/anthropic/claude-sonnet-4.6"`).

`openrouter_model_to_modelmeta()` then wraps this in `ModelMeta(provider="openrouter", model=<id>)`, so `ModelMeta.full` produces:

```
"openrouter/" + "openrouter/anthropic/claude-sonnet-4.6"
= "openrouter/openrouter/anthropic/claude-sonnet-4.6"
```

This double-prefixed ID is sent to the Supabase `messages` function, which strips only **one** `openrouter/` prefix — leaving `openrouter/anthropic/...` still prefixed — which OpenRouter then rejects with:

```
400 Bad Request: "openrouter/anthropic/claude-sonnet-4.6 is not a valid model ID"
```

Root cause traced via Supabase logs showing:
```
[messages] provider=openrouter model=openrouter/openrouter/anthropic/claude-sonnet-4.6 → openrouter/anthropic/claude-sonnet-4.6
```

Fixes gptme/gptme-cloud#198.

## Fix

Strip any leading `openrouter/` prefix from the model ID inside `openrouter_model_to_modelmeta()` before constructing `ModelMeta`. This is a no-op for direct OpenRouter API calls (which return bare IDs like `anthropic/claude-3-5-sonnet`) and corrects the double-prefix for proxy calls.

## Tests

Added `TestOpenrouterModelToModelmeta` with two cases:
- `test_native_id_no_change` — direct OpenRouter native ID goes through unchanged
- `test_proxy_prefixed_id_strips_duplicate` — proxy pre-prefixed ID yields single prefix

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,14 +56,19 @@
 </p>
 
 <p align="center">
-📜 A personal AI agent in your terminal, with tools to:<br/>
-run shell commands, write code, edit files, browse the web, use vision, and much more.<br/>
+📜 A personal AI agent that runs <i>anywhere a terminal runs</i> — your laptop,
+ssh sessions, tmux, headless servers, CI pipelines.<br/>
+Provider-agnostic, local-first, and unconstrained: ships with shell, Python, web,
+vision, and everything else an agent needs.<br/>
 A great coding agent, but general-purpose enough to assist in all kinds of knowledge-work.
 </p>
 
 <p align="center">
-An unconstrained local free and open-source <a href="https://gptme.org/docs/alternatives.html">alternative</a> to Claude Code, Codex, Cursor Agents, etc.<br/>
-One of the first agent CLIs created (Spring 2023) — and still in very active development.
+Free and open-source. Works with Anthropic, OpenAI, Google, xAI, DeepSeek, OpenRouter,
+or fully local via <code>llama.cpp</code> — your data, your models, your terminal.<br/>
+A capable <a href="https://gptme.org/docs/alternatives.html">alternative</a> to Claude Code,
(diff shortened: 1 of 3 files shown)


ErikBjare:
This is a good fix that should get merged, but isn't it a bit of a weird fix? Seems wrong of some other thing (proxy? webui?) to send requests like that, no?

TimeToBuildBob:
Good question. The proxy (Supabase models relay in gptme-cloud) returns model IDs already prefixed with `openrouter/` — that's intentional on its side, since it's presenting them as OpenRouter models. But `openrouter_model_to_modelmeta` then puts them into `ModelMeta(provider="openrouter", model=<id>)`, and `ModelMeta.full` prepends the provider a second time.

So both are arguably "correct" in isolation — the proxy namespaces the IDs it returns, and `ModelMeta.full` constructs the qualified ID. The bug is in the handoff: we assumed `model` in ModelMeta is always the bare ID without provider prefix.

The defensive `.removeprefix()` handles both code paths (direct OpenRouter API returns bare IDs; proxy returns prefixed ones) with a no-op for the former. The alternative would be to fix the proxy to not prefix IDs, but since we want to handle both sources robustly, this approach seems cleaner.

Happy to add a note to the comment in the code if you want to track it as a known proxy behavior worth fixing upstream in gptme-cloud.

PR #2321, opened 2026-05-02
feat(cost-awareness): warn when Anthropic prompt cache is likely cold

### LINKED ISSUE #2320: feat(cost-awareness): warn when Anthropic prompt cache is likely cold (>5min since last turn)
## Problem

Anthropic's prompt cache has a 5-minute TTL. After 5 minutes of inactivity, the next turn pays a full cache-miss cost on the system prompt + conversation history — which on long autonomous runs is the dominant cost driver. gptme tracks `cache_read_tokens` and `cache_creation_tokens` per request (see `gptme/util/cost_tracker.py`) and computes a cache hit rate, but **nothing surfaces the time-since-last-turn signal** to the user or to routing logic.

This is the gap [`jcode`](https://github.com/1jehuang/jcode) explicitly highlights in its UX — a "cache-cold warning" before the user sends a turn that will incur the cache miss. Bob's [peer research](../../knowledge/research/2026-05-02-jcode-peer-research.md) flagged it as the highest-leverage idea to lift (Impact 5 × Feasibility 5 × Alignment 5 = 125 in the local backlog).

## Why it matters

On Bob's autonomous loop (gptme + Claude Code, ~30-min cadence), every cache miss between sessions costs roughly the system-prompt size in cache_creation tokens. With a 5-minute TTL and a 30-minute scheduler, **every scheduled run pays the cold-start tax** — but with shorter cadences (hot-loop, bursty interactive use) the cache should stay warm. Currently there's no signal in the loop telling us we're about to fall off the cache cliff.

## Existing infrastructure

The relevant pieces are already there:

- `gptme/util/cost_tracker.py` — `CostEntry.timestamp` per request, `SessionCosts` aggregation, `cache_hit_rate` property
- `gptme/hooks/cost_awareness.py` — already has a hook surface for cost messaging
- `gptme/util/cost_display.py` — turns cost data into human-readable output
- Anthropic backend (`gptme/llm/llm_anthropic.py`) sends `cache_control` markers and reads back `cache_creation_tokens` / `cache_read_tokens`

Nothing in the codebase tracks "seconds since last Anthropic turn" or warns when that gap exceeds 5 minutes.

## Scope question

Where should the warning surface? The four candidates, in increasing order of complexity:

1. **CLI status / pre-turn message** — print a one-line warning before sending a turn when `(now - last_anthropic_turn_ts) > 5 min`. Cheapest, most visible. Hook fires from `cost_awareness.py` on a pre-turn event.
2. **Server status endpoint** — expose `cache_state` via the gptme server so chat.gptme.org / webui can render it. Slightly more work, but enables UI surfacing for non-CLI clients.
3. **Webui indicator** — small "cache cold" badge in the conversation header, driven by the server endpoint above.
4. **Routing input for `select-harness.py`** — Bob-specific. Read the cache_state from gptme and route cache-cold sessions to non-Anthropic backends (where the 5-minute TTL doesn't apply or is structured differently). Lives in `gptme-bob` workspace, not in gptme core.

My instinct: ship #1 first (CLI warning), then #2/#3 once we know what users actually want to see. #4 is downstream consumer work that doesn't need to land in gptme core.

## Edge cases / non-trivial bits

- **Multi-provider sessions**: TTL only applies to Anthropic. We should track per-provider last-turn timestamps, not a single global one.
- **Tool turns vs. user turns**: tool-result turns don't reset the TTL on Anthropic's side either; only the actual API call does. The signal we care about is "time since last successful Anthropic API call."
- **Cache writes vs. reads**: the warning should fire even on the *first* turn after a long gap, before we know whether it's a hit or a miss. The signal is purely temporal.
- **System-prompt-only sessions**: short sessions where the cache never had a chance to warm up don't need the warning. Suggested guard: only warn if there has been at least one prior Anthropic turn in the session.

## Proposed minimal slice (for #1)

```python
# In gptme/hooks/cost_awareness.py or a new hook
ANTHROPIC_CACHE_TTL_SECS = 5 * 60

def cache_cold_warning(session_costs: SessionCosts, model: str) -> str | None:
    if not model.startswith("anthropic/"):
        return None
    anthropic_entries = [e for e in session_costs.entries if e.model.startswith("anthropic/")]
    if not anthropic_entries:
        return None  # first Anthropic turn — nothing to warm up yet
    last_ts = max(e.timestamp for e in anthropic_entries)
    age = time.time() - last_ts
    if age > ANTHROPIC_CACHE_TTL_SECS:
        return f"⚠️  Anthropic prompt cache likely cold ({age/60:.1f} min since last turn, TTL=5min)"
    return None
```

That's the v0. Wire it into the pre-turn hook surface, format the warning, done.

## Tracking

- Backlog: `knowledge/strategic/idea-backlog.md` #210 (in `ErikBjare/bob`)
- Source: `knowledge/research/2026-05-02-jcode-peer-research.md` (peer research on jcode)
- Cross-link: gptme-bob `knowledge/strategic/idea-backlog.md` will be updated with this issue link

Happy to take this on once the surface is agreed.

TimeToBuildBob:
Implements proposed action #1 from #2320: a CLI warning when the Anthropic prompt cache is likely cold (>5 min since last Anthropic API call).

## Changes

- **gptme/hooks/cost_awareness.py**: New `cache_cold_warning_hook` — a GENERATION_PRE hook that checks time since last Anthropic API call. If the gap exceeds the 5-minute TTL, it stores a warning via the existing `_pending_warning_var` mechanism, which gets injected as a hidden system message before the next user turn.
- **tests/test_cost_cache_cold_warning.py**: 5 tests covering no-Anthropic, warm cache, cold cache, injection, and multi-call scenarios.

## Scope

Following Erik's review on #2320: only #1 (CLI warning) implemented. #2/#3 (server endpoint, webui badge) deferred. #4 (routing input) skipped.

## Verification

- [x] 10 tests pass (5 new + 5 existing)
- [x] mypy clean
- [x] ruff clean

Closes #2320

### DIFF
--- a/gptme/hooks/cost_awareness.py
+++ b/gptme/hooks/cost_awareness.py
@@ -3,12 +3,14 @@
 Integrates with the CostTracker to:
 - Initialize cost tracking at session start
 - Emit cost warnings at configurable thresholds
+- Warn when Anthropic prompt cache is likely cold
 - Provide cost data for eval framework integration
 
-See Issue #935 for design context.
+See Issue #935 for design context, #2320 for cache-cold warning.
 """
 
 import logging
+import time
 from collections.abc import Generator
 from contextvars import ContextVar
 from pathlib import Path
@@ -27,6 +29,9 @@ logger = logging.getLogger(__name__)
 _pending_warning_var: ContextVar[str | None] = ContextVar(
     "pending_warning", default=None
 )
+
+# Anthropic prompt cache TTL (5 minutes)
+ANTHROPIC_CACHE_TTL_SECS = 5 * 60
 
 # Default cost warning thresholds (in USD)
 # Warns when total session cost crosses these values
@@ -50,6 +55,49 @@ COST_WARNING_THRESHOLDS = [
     500.00,
     1000.00,  # Large session warnings
 ]
+
+
+def cache_cold_warning_hook(
+    manager: "LogManager",
+) -> Generator[Message | StopPropagation, None, None]:
+    """Emit warning when Anthropic prompt cache is likely cold.
+
+    Anthropic's prompt cache has a 5-minute TTL. After 5 minutes of inactivity,
+    the next turn pays a full cache-miss cost. This hook detects the gap and
+    warns the user before the cold-start turn is generated.
+
+    Only warns if:
+    - There has been at least one prior Anthropic turn (cache had a chance to warm)
+    - The gap since the last Anthropic turn exceeds the 5-minute TTL
+
+    Yields:
+        Nothing — warning is stored for later injection via inject_pending_warning
+    """
+    costs = CostTracker.get_session_costs()
+    if not costs or not costs.entries:
+        return
+
+    # Filter for Anthropic entries
+    anthropic_entries = [e for e in costs.entries if e.model.startswith("anthropic/")]
+    if not anthropic_entries:
+        return  # No Anthropic turns yet — nothing to warm up
+
+    last_ts = max(e.timestamp for e in anthropic_entries)
+    age = time.time() - last_ts
+
+    if age > ANTHROPIC_CACHE_TTL_SECS:
+        warning_text = (
+            f"<system_warning>Anthropic prompt cache likely cold "
+            f"({age / 60:.0f} min since last turn, TTL=5 min) — "
+            f"next turn may incur higher cost</system_warning>"
+        )
+        _pending_warning_var.set(warning_text)
+        logger.info(
+            f"Anthropic cache cold warning: {age / 60:.0f} min since last turn "
+            f"(TTL=5 min)"
+        )
+
+    yield from ()
 
 
 def session_start_cost_tracking(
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed the P1 priority inversion in 425206ea3.

The `Hook.__lt__` method sorts descending on priority (higher = first), so `inject_pending_warning` (priority=5) was running before `cache_cold_warning_hook` (priority=3) — the exact opposite of what the comments claimed. The cold-cache warning was always injected one turn late.

Fix: raised `cache_cold_warning_hook` priority from 3 → 7, so it now sets `_pending_warning_var` before `inject_pending_warning` reads it within the same `GENERATION_PRE` dispatch.

Also corrected the parameter signature to match the `GenerationPreHook` protocol (`messages: list[Message], **kwargs` instead of `manager: LogManager`). All 5 existing tests still pass; mypy clean.

TimeToBuildBob:
Closing in favour of #2322 which is cleaner: passes `model` kwarg through the hook system properly and has more comprehensive tests. #2322 also addresses the Greptile P2 false-positive finding.

PR #2324, opened 2026-05-03
test(cost-awareness): add 29 tests for cache-cold warning, cost thresholds, and injection

TimeToBuildBob:
Adds comprehensive test coverage for the `gptme.hooks.cost_awareness` module, which was shipped without tests in #2322 and #2321 (cache-cold warning).

### `cache_cold_warning_hook` (8 tests)
- No session costs → no warning
- Non-Anthropic models → no warning
- Recent Anthropic turn (<5 min TTL) → no warning
- Stale Anthropic turn (>5 min) → warning set with time summary
- Multiple entries, most recent fresh → no warning
- Multiple entries, all stale → warning based on most recent
- Hook is yield-only (tracking, no messages produced)
- Mixed Anthropic/non-Anthropic entries → warning based on last Anthropic entry

### `inject_pending_warning` (5 tests)
- No pending warning → nothing yielded
- Pending with user message → yields hidden system message, clears pending
- Pending with assistant message → no injection (only injects before user turns)
- Pending with empty messages → no injection
- No pending with multiple messages → nothing yielded

### `cost_warning_hook` (6 tests)
- No costs → nothing
- Below $0.10 threshold → no warning
- Above $0.10 threshold → warning with cost value
- Yields nothing (console-output only)
- Multiple thresholds in one entry → one warning for highest crossed
- Cache hit rate rendered in warning text

### `session_start_cost_tracking` (2 tests)
- Initializes CostTracker with session ID
- Idempotent (second call replaces first)

### `session_end_cost_summary` (3 tests)
- No costs → yields nothing
- Zero cost → yields nothing
- Non-zero cost → yields nothing (output via console.log)

### Integration (3 tests + register)
- Full: cold cache → inject → deliver → clear
- Warning-only-once (cleared after injection)
- Full lifecycle simulation (start → run → cold → inject → end)
- `register()` can be called without error

Closes: #2321 (test gap for cache-cold warning feature)

### DIFF
--- a/gptme/hooks/cost_awareness.py
+++ b/gptme/hooks/cost_awareness.py
@@ -3,9 +3,10 @@
 Integrates with the CostTracker to:
 - Initialize cost tracking at session start
 - Emit cost warnings at configurable thresholds
+- Warn when Anthropic prompt cache is likely cold
 - Provide cost data for eval framework integration
 
-See Issue #935 for design context.
+See Issue #935 for design context, #2320 for cache-cold warning.
 """
 
 import logging
@@ -28,6 +29,9 @@ logger = logging.getLogger(__name__)
 _pending_warning_var: ContextVar[str | None] = ContextVar(
     "pending_warning", default=None
 )
+
+# Anthropic prompt cache TTL (5 minutes)
+ANTHROPIC_CACHE_TTL_SECS = 5 * 60
 
 # Default cost warning thresholds (in USD)
 # Warns when total session cost crosses these values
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed all Greptile findings in 93f200a:

**P1 — Missing `cache_creation_tokens > 0` guard**: Added the guard to `cache_cold_warning_hook` matching `anthropic_cache_cold_warning`. Sessions that never used prompt caching no longer receive false-positive cold-cache warnings. Also added an explicit `test_stale_anthropic_turn_no_cache_creation_no_warning` test to cover this branch.

**P2 — Duplicate TTL constant**: Removed `ANTHROPIC_CACHE_TTL_SECONDS` (the second definition at line 59); `anthropic_cache_cold_warning` now uses `ANTHROPIC_CACHE_TTL_SECS` throughout.

**P2 — Tests encoding wrong behavior / brittle assertions**: Updated all `TestCacheColdWarningHook` tests and integration tests that had `cache_creation_tokens=0` with an expected warning to use `cache_creation_tokens > 0`. The "10 min" string assertion in `test_multiple_entries_all_stale_warning` is preserved — it's validated against the `{age / 60:.0f} min` format string in the warning text (integer minutes), so it's stable.

**P3 — `test_register_called` swallowing exceptions**: Removed the bare `except Exception: pass`. The registry silently replaces hooks with duplicate names (no exception is raised), so the try/except was dead code.

All 35 tests pass.

PR #2327, opened 2026-05-04
fix(webui): delete stale api.gptme.ai Cloud preset on migration

TimeToBuildBob:
## Problem

Users who had the old hardcoded `https://api.gptme.ai` Cloud preset in their `gptme_servers` localStorage hit two issues:

1. The preset points to the wrong URL — managed cloud auth gives each user a unique instance URL, not a generic endpoint.
2. The stale URL is blocked by the gptme.ai Content-Security-Policy (`connect-src` only allows `fleet.gptme.ai`, not `api.gptme.ai`), causing a CSP error in the console and a silent connection failure.

The previous `migrateCloudPreset` only removed `isPreset` (making the entry deletable), but **left the broken server in the registry**. Users had to manually delete it.

## Fix

Strengthen the migration to fully remove the stale entry via `.filter()` instead of just un-setting `isPreset`. After `migrateCloudPreset` runs, `loadRegistry()` already re-validates `activeServerId` and prunes `connectedServerIds`, so affected users gracefully land on their Local server and can reconnect via the Go to Chat flow on the gptme.ai dashboard.

## Related

- gptme/gptme-cloud#237 — user report of the CSP error and broken cross-device server config

### DIFF
--- a/webui/src/stores/servers.ts
+++ b/webui/src/stores/servers.ts
@@ -22,8 +22,10 @@ function loadRegistry(): ServerRegistry {
         if (!parsed.connectedServerIds) {
           parsed.connectedServerIds = [parsed.activeServerId];
         }
-        // Migrate: make old Cloud preset removable (it pointed at a stale URL)
+        // Migrate: remove stale Cloud preset (it pointed at a broken URL)
         migrateCloudPreset(parsed);
+        // If migration removed all servers, fall through to fresh migration
+        if (parsed.servers.length === 0) return migrateFromLegacy();
         // Validate activeServerId points to an existing server
         if (!parsed.servers.some((s) => s.id === parsed.activeServerId)) {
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile's two P1 findings in a0bd1e78c:

1. **`isPreset` guard made migration a no-op for previously-migrated users** — the old migration called `delete server.isPreset`, so those entries no longer carry the flag. Changed filter to match by URL alone: `normalized(s.baseUrl) !== normalized(STALE_CLOUD_URL)`. This cleans up both fresh entries and ones that went through the old migration.

2. **Empty `servers` array causes TypeError** — if Cloud was the sole server, filtering leaves an empty array and `parsed.servers[0].id` throws. Added an explicit early-return to `migrateFromLegacy()` instead of relying on the incidental outer try/catch.

Issue #444, opened 2025-02-17
Add support for ephemeral messages


ErikBjare:
"Ephemeral messages" are messages which are intended to be dropped after a while, like after a few messages.

To ensure effective prompt caching, we should make sure we request a cache point before the first ephemeral message (since the conversation should be stable up until that point).

Examples could be:
 - `<thinking>` contents
 - Full file contents on failed patches
 - Screenshot outputs in computer use (since screenshots are taken so often)
 - RAG results

TimeToBuildBob:
This would be very valuable for long-running agent sessions. Here's how I'd approach it:

**Message-level ephemeral flag**: Add an `ephemeral` field to `MessageMetadata` with a TTL (number of turns before dropping):

```python
class MessageMetadata(TypedDict, total=False):
    # ... existing fields ...
    ephemeral_ttl: int  # Drop after N assistant turns (0 = never)
```

**Candidate content types** (from the issue + operational experience):
- `<thinking>` blocks → TTL=2 (useful for immediate context, stale quickly)
- Screenshot outputs → TTL=1 (only relevant for the next action)
- Failed tool results → TTL=3 (see #523, could overlap with backtracking)
- RAG/search results → TTL=2 (relevant for current subtask only)
- Full file contents from failed patches → TTL=1

**Cache-aware pruning**: The key insight from the issue is right — place a cache breakpoint _before_ the first ephemeral message. Then when ephemeral messages expire, everything before the breakpoint stays cached. Implementation:

1. In `prepare_messages()` (logmanager.py), filter out expired ephemeral messages
2. Mark the last non-ephemeral message with a cache hint
3. When messages are pruned, the prefix cache remains valid

**Token savings estimate**: In a typical 30-turn autonomous session, thinking blocks + failed results + screenshots can consume 30-40% of context. Pruning with TTL=2 could recover ~20% of context budget at steady state.

This pairs well with auto-compact — ephemeral is for _recent_ bloat (last few turns), while auto-compact handles _accumulated_ bloat (old history). Together they'd keep context much leaner.

hhe48203-ctrl:
Hi @ErikBjare! I'd like to take a shot at implementing this. Before I start, a few design questions to make sure I'm going in the right direction:

1. **Marking ephemeral messages**: Is the intent to add a flag/metadata on the `Message` object (e.g. `ephemeral=True`), or would messages be tagged via content conventions (like a special role or wrapper)?

[...]

PR #2331, opened 2026-05-04
fix(messages): merge turns after ephemeral pruning

TimeToBuildBob:
Follow-up to the post-merge Greptile P1 on #2330. Refs #444.

## Summary
- merge adjacent same-role messages after expired ephemeral turns are pruned
- preserve files, flags, and TTL semantics through `Message.concat()`
- update ephemeral pruning tests so the output is provider-safe instead of asserting consecutive user turns

## Tests
- `poetry run pytest tests/test_ephemeral.py tests/test_llm_utils.py tests/test_llm_anthropic.py -q`
- `prek run ruff ruff-format --files gptme/logmanager/manager.py tests/test_ephemeral.py`

### DIFF
--- a/gptme/logmanager/manager.py
+++ b/gptme/logmanager/manager.py
@@ -708,7 +708,24 @@ def prune_ephemeral_messages(msgs: list[Message]) -> list[Message]:
         if msg.role == "assistant":
             assistant_turns_after += 1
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #2338, opened 2026-05-06
feat(tools): add gh_list compactor to shell_compact

TimeToBuildBob:
## Summary

Add a new compactor for `gh <noun> list|status` commands to the `shell_compact` tool, following the same pattern as the existing `git_log_oneline` compactor.

Supports: `gh issue list`, `gh pr list`, `gh pr status`, `gh run list`, `gh release list`, `gh repo list`.

## What changed

- `_matches_gh_list()` — detects gh list/status command shapes
- `_format_gh_list_preview()` — shows first 20 items + total count + grep hint
- `_compact_body_via_save()` — extracted helper to reduce duplication between compactors
- `_dispatch_compactor()` — generic routing between compactors
- Renamed `_execute_compacted_git_log` → `_execute_compacted` for generic dispatch
- Updated tool instructions, examples, and allowlist hook

## Testing

- `ruff check` — clean
- `ruff format --check` — clean after formatting
- `pytest tests/test_tools_shell_compact.py tests/test_tools_shell.py tests/test_util_context_savings.py` — 164 passed

## Next

After this, the next compactors from the design spike can be added incrementally: `find_files`, `journalctl`, `gh_pr_view`.

### DIFF
--- a/gptme/tools/shell_compact.py
+++ b/gptme/tools/shell_compact.py
@@ -1,9 +1,11 @@
 """
 Compact shell wrapper for known high-output command shapes.
 
-Currently this only compacts ``git log --oneline`` by returning the first few
-commits plus a pointer to the full saved output. Unsupported commands fall back
-to the regular shell tool.
+Currently supports:
+- ``git log --oneline`` — first 20 commits + total count
+- ``gh <noun> list`` / ``gh <noun> status`` — first 20 items + count + grep hint
+
+Unsupported commands fall back to the regular shell tool.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


ErikBjare:
See my comment on the previous PR about shell_compact, not sure this makes sense as a separate tool.

TimeToBuildBob:
Agreed. I converted this PR to draft so it does not get merged as another `shell_compact` extension.

The better direction is one of:
- fold the compact preview behavior into the existing `shell` tool, since `shell` already owns output truncation/saving and users should not have to pick a parallel tool
- remove/revert `shell_compact` if explicit command-shape previews are not worth the surface area

I'm going to stop stacking follow-up compactors until that direction is settled. Greptile's refactor note is valid, but it is lower priority than the tool-shape decision.

TimeToBuildBob:
Closing per consensus — the better direction is folding compact preview into the shell tool directly. No more shell_compact extensions until that direction is settled.

PR #2339, opened 2026-05-06
feat: add gh issue/pr list compaction to shell_compact

TimeToBuildBob:
Extends shell_compact to wrap gh issue list and gh pr list commands.


ErikBjare:
See https://github.com/gptme/gptme/pull/2337#issuecomment-4386788824

TimeToBuildBob:
Closing this PR. The design feedback from #2337 applies here: adding more command-specific behavior to a separate `shell_compact` tool is the wrong next step. Any follow-up should first decide whether the existing compact behavior gets folded into `shell` or removed.

PR #738, opened 2025-10-17
feat(shell): add pre-execution validation to prevent common errors

TimeToBuildBob:
## Summary

Implements pre-execution validation for the shell tool to prevent common error patterns identified in autonomous run performance analysis.

### Phase 1: Core Validator (Complete ✅)
- Created `ShellValidator` class with configurable validation levels
- Implemented 5 validation rules:
  1. Bare variable detection (e.g., `LLM_API_TIMEOUT` → `$LLM_API_TIMEOUT`)
  2. Python invocation check (`python` → `python3`)
  3. Python file execution check (`./file.py` → `python3 file.py`)
  4. Path quoting for spaces
  5. Directory path check (incorrect directory assumptions)
- 40 unit tests, 100% passing

### Phase 2: Integration (Complete ✅)
- **2.1**: Integrated validator into `gptme/tools/shell.py`
- **2.2**: Enhanced error messages with suggestions and lesson links
- **2.3**: Added configuration support via `gptme.toml` and environment variables
- **2.4**: 22 integration tests covering real execution contexts
- Total: 62 tests passing (40 unit + 22 integration)

## Error Prevention Potential

Based on analysis of 300 autonomous run logs:
- **Shell errors**: 486 occurrences (50% reduction target)
- **Variables as commands**: 20+ occurrences (e.g., `LLM_API_TIMEOUT: command not found`)
- **Python invocation**: 4+ occurrences (`python: command not found`)
- **Directory path errors**: 15+ occurrences
- **Total**: 48+ errors immediately preventable

### Environment Variable
```bash
# Default: warn mode (log warnings, allow execution)
gptme

# Strict mode (block on warnings)
GPTME_SHELL_VALIDATE=strict gptme

# Disable validation
GPTME_SHELL_VALIDATE=off gptme
```

### Project Configuration (gptme.toml)
```toml
[shell_validation]
enabled = true
level = "warn"  # or "strict", "off"

[shell_validation.rules]
bare_variables = true
python_invocation = true
path_quoting = true
```

## Testing

```bash
# Run all validator tests
pytest tests/test_shell_validator.py -v

# Run integration tests
pytest tests/test_shell_validation_integration.py -v

# Performance validation
pytest tests/test_shell_validator.py::TestPerformance -v
# Result: <10ms per validation ✅
```

## Documentation

Created 5 lessons documenting common error patterns:
- [Shell Variable Syntax](https://github.com/gptme/gptme-bob/blob/master/lessons/tools/shell-variable-syntax.md)
[...]

PR #2346, opened 2026-05-07
feat(lessons): cap lesson injection at 50K tokens to prevent sys_prompt bloat

TimeToBuildBob:
## Problem

gptme currently injects all matched lessons into the system prompt with no token budget. With 162 lesson files totaling ~89K tokens, heavy-match sessions can exceed 89K — consuming a large portion of peak context before the first user message.

See #759 (ErikBjare/bob) and #738 (ErikBjare/bob) for full analysis.

## Changes

1. **Token budget enforcement** in `auto_include_lessons`: after matching lessons, total formatted content is estimated (simple char-based heuristic) and compared against a budget. Lessons beyond budget are dropped, starting from lowest-scored.

2. **Configurable via env var**: `GPTME_LESSONS_TOKEN_BUDGET` defaults to 50000 (50K tokens).

3. **Logging**: warns when budget is exceeded, showing how many lessons were dropped.

4. **New tests**: 11 tests covering budget enforcement, edge cases (single oversized lesson, multiple drops), env var configuration, and invalid input.

## Implementation details

- Token estimation uses `len(text) // 3` (conservative ~3.5 chars/token for code-markdown content). This is sufficient for budget enforcement — actual tokenization varies by model.
- Only the `_format_with_budget` function was added; the original `_format_lessons` is preserved for backward compatibility.
- The highest-scored lesson is always included even if it exceeds budget alone (minimum of 1).

Closes ErikBjare/bob#759

### DIFF
--- a/gptme/lessons/auto_include.py
+++ b/gptme/lessons/auto_include.py
@@ -1,6 +1,7 @@
 """Automatic lesson inclusion based on context."""
 
 import logging
+import os
 from typing import TYPE_CHECKING
 
 from .index import LessonIndex
@@ -20,6 +21,31 @@ except ImportError:
     HYBRID_AVAILABLE = False
     logger.info("Hybrid matching not available, using keyword-only matching")
 
+# Default token budget for lesson injection (50K tokens).
+# Configurable via GPTME_LESSONS_TOKEN_BUDGET env var.
+_DEFAULT_TOKEN_BUDGET = 50000
+
+
+def _get_token_budget() -> int:
+    """Get the lesson token budget from environment or default."""
+    try:
+        return int(
+            os.environ.get("GPTME_LESSONS_TOKEN_BUDGET", str(_DEFAULT_TOKEN_BUDGET))
+        )
+    except (ValueError, TypeError):
+        return _DEFAULT_TOKEN_BUDGET
+
+
+def _estimate_tokens(text: str) -> int:
+    """Estimate token count for a text string.
+
+    Uses a simple character-based heuristic (~4 chars per token for English text,
+    adjusted for code/markdown density). This is a rough estimate sufficient for
+    budget enforcement — actual tokenization varies by model.
+    """
+    # Use conservative estimate: 3.5 chars per token for code-heavy content
+    return max(1, len(text) // 3)
+
 
 def auto_include_lessons(
     messages: list["Message"],
@@ -27,6 +53,7 @@ def auto_include_lessons(
     enabled: bool = True,
     use_hybrid: bool = False,
     hybrid_config: "HybridConfig | None" = None,
+    max_tokens: int | None = None,
 ) -> list["Message"]:
     """Automatically include relevant lessons in message context.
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the two Greptile findings in 008f81d:

1. **Unreachable dead code removed** (`if not lesson_content` block in `auto_include_lessons`): `_format_with_budget` always includes at least the first lesson (force-include logic), and `matches` is already checked to be non-empty before the call — so `lesson_content` can never be empty at that point.

2. **Misleading test docstring fixed** (`test_format_with_budget_first_lesson_too_large`): The old docstring said "all are dropped" but the assertions correctly showed the lesson IS included. Clarified that the first/highest-scored lesson is force-included regardless of size (minimum 1 lesson), and `dropped == 0` because the test only has one lesson in the fixture — nothing left to drop.

The budget logic itself (oversized first lesson exhausts budget for subsequent lessons) is intentional per the PR description: "The highest-scored lesson is always included even if it exceeds budget alone (minimum of 1)." When the budget is already exceeded by the first lesson, subsequent lessons are correctly skipped.

TimeToBuildBob:
**Update (2ac435621)**: On reflection, the budget overflow bug was a real correctness issue, not intentional.

Previous comment said "when the budget is already exceeded by the first lesson, subsequent lessons are correctly skipped" — that was wrong. The intent of force-including the first lesson is to guarantee *at least one lesson* reaches the model. But if the first lesson is 100K tokens and the budget is 50K, the old code made `total_tokens = 100K`, causing every subsequent lesson (even a 5-token one) to fail the check `total_tokens + lesson_tokens > max_tokens`. A 5-token lesson obviously fits within a 50K budget.

**Fix**: Track subsequent-lesson tokens separately from the force-included first lesson. `subsequent_tokens` starts at 0 and only counts non-first lessons. The first lesson's size no longer affects the budget available to subsequent ones.

```python
# Before: oversized first lesson silently dropped all subsequent ones
if included and total_tokens + lesson_tokens > max_tokens:
    dropped += 1

# After: first lesson is free; budget only applied to subsequent lessons
if not included:
    included.append(lesson_str)  # always include first
elif subsequent_tokens + lesson_tokens > max_tokens:
    dropped += 1
else:
    included.append(lesson_str)
    subsequent_tokens += lesson_tokens
```

Added regression test `test_format_with_budget_oversized_first_does_not_block_small_subsequent` that verifies a huge first lesson + tiny second lesson → both included when the tiny one fits the budget.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the two P1 findings from the third Greptile review in c3b8279:

1. **Warning log compared incompatible token counts**: `_format_with_budget` now returns a 3-tuple `(content, dropped, subsequent_tokens)`. The warning log uses `subsequent_tokens` (tokens from non-first lessons only) as the numerator vs `max_tokens` (which also only governs non-first lessons) — apples-to-apples. Previously it used total content tokens including the always-exempt first lesson, which could show confusing numbers like "90K/50K budget" when the mechanism was working correctly.

2. **`max_tokens` docstring misrepresented scope**: Clarified that `max_tokens` is a "token budget for lessons beyond the first" and that the highest-scored lesson is always force-included regardless. The old wording implied a ceiling on total injected content.

Added `test_format_with_budget_subsequent_tokens_excludes_first` to lock in the contract that `subsequent_tokens` never includes the first lesson.

PR #2373, opened 2026-05-10
fix(server): require explicit opt-in for API auto_confirm on network bindings

andesyteoss:
## Summary

When `gptme-server` is bound to a non-loopback interface (auth enabled), authenticated API callers can set `auto_confirm` / `auto_confirm_count` on a session, causing subsequent LLM-emitted shell/python tool calls to execute without human review. Combined with prompt injection — which is the standard threat model for any agent that ingests web pages, files, or tool output — this turns an authenticated foothold into arbitrary command execution as the server user.

This PR gates the three (and only three) code paths that can mutate `session.auto_confirm_count`, requiring an explicit operator opt-in (`GPTME_ALLOW_AUTO_CONFIRM=1`) whenever auth is enforced. Loopback behavior is unchanged.

## Affected code paths

All three mutation points for `session.auto_confirm_count`:

- `gptme/server/api_v2.py` — `PUT /api/v2/conversations/<id>` accepts `auto_confirm` in payload
- `gptme/server/api_v2_sessions.py` — `POST /api/v2/conversations/<id>/step` accepts `auto_confirm`
- `gptme/server/api_v2_sessions.py` — `POST /api/v2/conversations/<id>/tool/confirm` with `action=auto, count=N`

## Why this matters

`SECURITY.md` documents interactive confirmation as a load-bearing control:

> **Interactive Mode**: Commands require user confirmation before execution

Authentication answers *who can talk to the API*. It does not answer *what the LLM gets told to do once a request is in flight*. Any prompt-injection vector — a fetched web page, a file the user opens, tool output — can instruct the model to emit a shell or python tool block. With `auto_confirm` set, those run without the human-in-the-loop step that lets a user notice the injection.

## Preconditions

- Server bound to a non-loopback interface (auth enabled)
- Either: attacker possesses a valid bearer token, OR an authenticated user is induced to ingest attacker-controlled content (web fetch, file paste, tool output)
- LLM follows the injected instructions and emits a shell/python tool block

## Fix

Adds `is_auto_confirm_allowed()` in `gptme/server/auth.py` and gates all three mutation points:

- **Auth disabled** (loopback, single-user CLI-equivalent context) → behavior unchanged.
- **Auth enabled** (network binding) → setting `auto_confirm` truthy or calling `action=auto` returns HTTP 403 unless the operator has set `GPTME_ALLOW_AUTO_CONFIRM=1` on the server process.

The 403 response body documents the env var so operators who genuinely want unattended automation on a network binding can still opt in deliberately.

Diffstat:

```
 gptme/server/api_v2.py          | 26 +++++++++++++++++++++++++-
 gptme/server/api_v2_sessions.py | 38 +++++++++++++++++++++++++++++++++++++-
 gptme/server/auth.py            | 30 ++++++++++++++++++++++++++++++
 3 files changed, 92 insertions(+), 2 deletions(-)
```

## Testing

- Manually exercised all three endpoints in both modes:
  - With auth disabled (loopback) — `auto_confirm=true` is accepted, existing behavior preserved.
  - With auth enabled and `GPTME_ALLOW_AUTO_CONFIRM` unset — all three paths return 403 with a helpful message.
  - With auth enabled and `GPTME_ALLOW_AUTO_CONFIRM=1` — accepted as before.
- Confirmed the `bool`/`int` discrimination matches the existing `type(...) is bool` pattern in the surrounding code (since `bool` is a subclass of `int` in Python).
- Existing server tests continue to pass; the gate is additive.

## Adversarial review

Before submitting we tried to disprove this:

- **Does auth alone mitigate it?** No. Auth controls API access, not LLM-emitted instructions. Prompt injection is the standard threat model for tool-using agents, and `SECURITY.md` already names interactive confirmation as the mitigation.
- **Is the loopback / CLI case affected?** No. Loopback disables auth and matches single-user CLI semantics; the fix preserves that behavior exactly.
- **Are there other paths to set `auto_confirm_count`?** We grepped for every assignment to `auto_confirm_count` and `session.auto_confirm_count` — the three endpoints above are the only mutation points.
- **Is the env var a footgun?** It's opt-in, defaults closed, and is surfaced in the 403 response body, so it's discoverable rather than hidden.

## Disclosure

`SECURITY.md` asks for private reporting; we attempted to file via GitHub's private vulnerability reporting endpoint first, but it returned a 500 error on submission. Since the patch itself is what needs review, we're sending it as a PR. Happy to coordinate via `security@gptme.org` if you'd prefer to discuss exploitability privately before merging — just let us know.

## Classification

- CWE-78 (OS Command Injection, via tool execution)
- Adjacent: CWE-862 (Missing Authorization on a security-relevant state mutation)
- Severity: High — single-step prompt-injection-to-RCE chain when prerequisites are met.

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -57,7 +57,7 @@ from .api_v2_common import (
     msg2dict,
 )
 from .api_v2_sessions import SessionManager, sessions_api
-from .auth import require_auth
+from .auth import is_auto_confirm_allowed, require_auth
 from .external_sessions import get_external_session_provider
 from .openapi_docs import (
     CONVERSATION_ID_PARAM,
@@ -471,6 +471,30 @@ def api_conversation_put(conversation_id: str):
                 }
             ),
             400,
+        )
+
+    # CWE-78: when the server is reachable over the network (auth enabled),
+    # allowing API callers to set auto_confirm turns any prompt-injection
+    # foothold into shell/python execution. Require an explicit operator
+    # opt-in (GPTME_ALLOW_AUTO_CONFIRM=1). Loopback binding (auth disabled)
+    # keeps existing behavior.
+    auto_confirm_truthy = (
+        (type(auto_confirm) is bool and auto_confirm)
+        or (type(auto_confirm) is int and auto_confirm > 0)
+    )
+    if auto_confirm_truthy and not is_auto_confirm_allowed():
+        return (
+            flask.jsonify(
+                {
+                    "error": "auto_confirm not allowed",
+                    "message": (
+                        "Setting 'auto_confirm' over the API is disabled when "
+                        "the server is bound to a network interface. Set "
+                        "GPTME_ALLOW_AUTO_CONFIRM=1 on the server to opt in."
+                    ),
+                }
+            ),
+            403,
         )
 
     # Validate all messages before creating any side effects (directories).
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -22,7 +22,7 @@ from ..llm.models import get_default_model
 from ..logmanager import LogManager
 from ..message import Message
 from .api_v2_common import _validate_branch, _validate_conversation_id
-from .auth import require_auth
+from .auth import is_auto_confirm_allowed, require_auth
 from .constants import DEFAULT_FALLBACK_MODEL
 from .openapi_docs import (
     CONVERSATION_ID_PARAM,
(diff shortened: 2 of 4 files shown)


lewiswigmore:
Thanks for the careful review — both Greptile points are valid and addressed in 5173bf0.

**1. External-auth bypass (P1, `auth.py`)**

You're right: keying `is_auto_confirm_allowed()` on `_auth_enabled` left a gap when an operator runs `--host 0.0.0.0` with `GPTME_DISABLE_AUTH=true` (k8s ingress / reverse-proxy with external auth). In that case `_auth_enabled` is `False` for the same reason as loopback, and the 403 gate was bypassed even though the server is publicly reachable.

Fix: track network-binding state independently of auth state.

- New module-level `_is_network_binding: bool` flag, set in `init_auth()` from `not is_local_host(host)` in all three branches (loopback, network+auth, network+`GPTME_DISABLE_AUTH`).
- `is_auto_confirm_allowed()` now keys on `_is_network_binding` instead of `_auth_enabled`. Any non-loopback bind requires `GPTME_ALLOW_AUTO_CONFIRM=1`, regardless of whether gptme's own auth layer is enforced.
- Loopback behaviour is unchanged (allowed without opt-in).

**2. Gate ordering in `action == "auto"` (P2, `api_v2_sessions.py:623`)**

Moved the `is_auto_confirm_allowed()` check before the `count` type/range validation so a 403 is returned consistently with the ordering used in the other two gated paths (PUT conversation, POST /step).

**3. Codecov patch coverage**

Added tests in `tests/test_llm_auth.py`:

- `TestIsAutoConfirmAllowed` — loopback / network / network+external-auth / opt-in env-var matrix, including a regression test for the bypass above.
- `TestInitAuthNetworkBindingFlag` — verifies `_is_network_binding` is set correctly in all three `init_auth()` branches.

`tests/test_llm_auth.py` and the existing `test_security_fix_verification.py` / `test_security_qa.py` / `test_shell_allowlist_autoconfirm.py` suites pass locally.

lewiswigmore:
Closing this to reduce the open-PR pile-up — we have multiple outstanding security contributions to this repo and that volume is not fair on your review queue. Keeping #1781 (CWE-20: validate auto_confirm type in PUT conversation endpoint (CWE) as the primary one to focus attention on.

Happy to revisit this finding separately later if it is still relevant. Apologies for the noise.

PR #2385, opened 2026-05-12
feat(hooks): tool-target agent-instruction loading (Phase 1)

TimeToBuildBob:
## Summary

Adds a new `tool_target_instructions` hook (subscribing to `TOOL_EXECUTE_POST`) that discovers and injects `AGENTS.md` / `CLAUDE.md` / `GEMINI.md` files when structured file tools touch paths in directories that contain local instruction files — even if CWD itself does not change.

This extends `agents_md_inject` (which fires on `CWD_CHANGED`) to cover a common real workflow:

- reading `projects/gptme/webui/src/foo.ts` from repo root
- patching `/tmp/worktrees/feature/file.py` while CWD is unchanged
- saving content in a sibling repo by absolute path

Inspired by Gemini CLI's JIT context loading. Design doc: `knowledge/technical-designs/tool-targeted-agent-instruction-loading.md` in the Bob repo.

## Scope (Phase 1)

- Structured file tools only: `read`, `save`, `append`, `patch`, `morph`, `ls`, `browse`.
- Hard caps: max 3 directories inspected per tool event, max 3 new files injected per event.
- Reuses `find_agent_files_in_tree()` and `_loaded_agent_files_var` dedup state from `agents_md_inject`, including content-hash dedup so worktree copies with identical content are not re-injected.

## Non-goals (deferred)

- Shell payload parsing — Phase 2.
- Pre-write guards (intercept the tool call) — Phase 2.
- `@file.md`-style imports — Phase 3.

## Tests

17 new tests in `gptme/hooks/tests/test_tool_target_instructions.py`:

- `_extract_paths`: positional arg, `path` kwarg, dedup, no-args, non-string args.
- `_resolve_directory`: existing dir, existing file (→ parent), nonexistent path (→ parent), garbage path (→ None), relative path resolved against cwd.
- `on_tool_execute_post`: subdir-read injection, no re-injection across tool events, content-hash dedup across paths, unstructured tool skip, no-agents-md no-op, MAX_NEW_FILES cap, pre-seeded content-hash skip.

Full hook test suite (403 tests) green.

## Test plan

- [x] New unit tests pass (`pytest gptme/hooks/tests/test_tool_target_instructions.py -q`)
- [x] Full hook test suite passes (`pytest gptme/hooks/tests/ -q`)
- [x] Ruff format + check clean

Co-Authored-By: Bob <bob@superuserlabs.org>

### DIFF
--- a/gptme/hooks/__init__.py
+++ b/gptme/hooks/__init__.py
@@ -180,6 +180,9 @@ def init_hooks(
         "agents_md_inject": lambda: __import__(
             "gptme.hooks.agents_md_inject", fromlist=["register"]
         ).register(),
+        "tool_target_instructions": lambda: __import__(
+            "gptme.hooks.tool_target_instructions", fromlist=["register"]
+        ).register(),
         "mcp_namespace_hint": lambda: __import__(
             "gptme.hooks.mcp_namespace_hint", fromlist=["register"]
         ).register(),
--- a/gptme/hooks/tests/test_tool_target_instructions.py
+++ b/gptme/hooks/tests/test_tool_target_instructions.py
@@ -1,0 +1,217 @@
+"""Tests for tool_target_instructions hook.
+
+Verifies that AGENTS.md/CLAUDE.md/GEMINI.md files are loaded when a structured
+file tool touches a path under a directory that contains them, without
+requiring a CWD change.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from gptme.hooks.tool_target_instructions import (
+    MAX_NEW_FILES,
+    _extract_paths,
+    _resolve_directory,
+    on_tool_execute_post,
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
Addressed both Greptile findings:

**1. `browse` in `_STRUCTURED_PATH_TOOLS`** — already fixed in the refactor commit (537ae1f): `browse` was removed from the tool set. The hook now uses `_PATH_TOOLS` which only includes `read`, `save`, `append`, `patch`, `morph`.

**2. Cap drops most-specific files first** — fixed in a67c4e4: `candidate_files` is now reversed before being passed to `inject_agent_instruction_files`, so the deepest (most project-specific) AGENTS.md gets priority. When the cap fires, home-level general files are dropped via skip-note instead of the innermost project files. Added a regression test that builds a 5-level tree and asserts the deepest file is injected while the root-level file is not.

PR #2409, opened 2026-05-18
fix(webui): preserve structured API errors

TimeToBuildBob:
## Summary
- preserve nested API error messages from managed-service responses instead of stringifying them as `[object Object]`
- carry structured `code`/`type` metadata on `ApiClientError`
- add regression coverage for payment/subscription error envelopes

## Testing
- npm test -- src/utils/__tests__/api.test.ts
- npx eslint . --fix && npx tsc -p tsconfig.app.json --noEmit

## Related
- gptme/gptme-cloud#61

### DIFF
--- a/webui/src/components/ui/badge.tsx
+++ b/webui/src/components/ui/badge.tsx
@@ -23,8 +23,7 @@ const badgeVariants = cva(
 );
 
 export interface BadgeProps
-  extends React.HTMLAttributes<HTMLDivElement>,
-    VariantProps<typeof badgeVariants> {}
+  extends React.HTMLAttributes<HTMLDivElement>, VariantProps<typeof badgeVariants> {}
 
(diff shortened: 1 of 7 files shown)


TimeToBuildBob:
Addressed both Greptile review findings in 453fdfaa4:

- **HTTP status lost for plain-string errors** (PRRT_kwDOJN3IDs6CuKnx): `normalizeApiError` now falls back to `httpStatus` when the body's `status` field is absent — plain `{"error": "Not found"}` responses correctly preserve the HTTP status code.
- **Added test coverage** for the plain-string error path: `'preserves HTTP status for plain-string error responses'` exercises a `{"error": "some message"}` body with HTTP 429.

The null-error guard (PRRT_kwDOJN3IDs6Ct6Le) was already addressed in dd93e3e6f — the earlier review flagged the gap, that commit fixed it.

PR #2410, opened 2026-05-18
feat(setup): add confirmation prompt for fish completions

TimeToBuildBob:
Adds a confirmation prompt before installing fish shell completions, replacing the `# TODO: prompt for confirmation?` comment.

**Change**: A single `Confirm.ask()` call before the symlink/copy step. If the user declines, completions are skipped with a dimmed message.

**Testing**: Import check passes. `Confirm` is already imported in the file (used in 3 other places in the same module).

**Found via**: Bob cross-repo scouting (`scripts/cross_repo_offline_supply.py` → local TODO seed at `gptme/cli/setup.py:181`).

### DIFF
--- a/gptme/cli/setup.py
+++ b/gptme/cli/setup.py
@@ -178,7 +178,9 @@ def _install_fish_completions(fish_completions_file: Path, source_file: Path):
         )
     )
 
-    # TODO: prompt for confirmation?
+    if not Confirm.ask("Install fish completions?", default=True):
+        console.print("   [dim]Skipping fish completions[/dim]")
+        return
     try:
         fish_completions_file.symlink_to(source_file)
         console.print(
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed: the redundant confirmation identified by Greptile.

**Root cause**: I put the `Confirm.ask` inside `_install_fish_completions`, but that helper is called from two paths:
1. **Fresh install** (line ~100 in `_setup_completions`): file doesn't exist → no prior consent. This is where the prompt belongs.
2. **Re-install** (4 locations in `_check_completions_file`): lines 135, 155, 160, 165 — each already asks context-specific confirmation ("Update completions?", "Replace with symlink?", "Replace with current version?") **before** calling the helper.

**Fix**: Moved the confirmation from the shared helper out to the fresh-install call site only. The re-install paths keep their existing prompts and now call `_install_fish_completions` without a redundant second ask.

Commit: 6dfac8d51 — `fix(setup): move fish completions confirmation from installer to caller`

TimeToBuildBob:
@greptileai review

Issue #607, opened 2025-07-26
Refactor ToolSpec to support native functions without IPython, better MCP tool breakdown in ToolSpec


ErikBjare:
Right now:

- ToolSpec's have `functions=` which are registered in the ipython shell
   - Sometimes we may want to run agents without the ipython shell, but with tool that uses `functions=`
- MCP tools for a server are created as seperate ToolSpec's `discord.send_message`, `discord.read_channel`, etc.
   - This makes MCP servers have `functions=` without needing ipython, instead via one ToolSpec per function. 

We should probably work around this somehow, making tools `functions=` not strictly integrated with `ipython` like that, taking inspiration from the MCP approach. Maybe each `functions=` should generate its own sub-ToolSpec?

I also find that I sometimes want to use a tool_allowlist like `discord.*` since it creates multiple tools for a given MCP server. MCP also offers hints like "read only" and such that could be helpful to prevent the lethal trifecta when prompting agentic workflows that run without human in the loop (or otherwise extra safety checks to prevent the lethal trifecta).

TimeToBuildBob:
Concrete design sketch for decoupling `functions=` from IPython:

**Current architecture problem**: `ToolSpec.functions` are callables that get registered in the IPython namespace. MCP tools already bypass this by creating one ToolSpec per function. These two patterns should converge.

**Proposed: ToolFunction abstraction**

```python
@dataclass
class ToolFunction:
    """A callable function exposed as a tool, independent of execution runtime."""
    name: str
    description: str
    parameters: dict  # JSON Schema for arguments
    fn: Callable      # The actual function
    group: str = ""   # For grouping (e.g., "discord", "github")

class ToolSpec:
    # ... existing fields ...
    functions: list[ToolFunction] = []  # Replace current Callable list
```

**Key changes:**
1. `ToolFunction` wraps each callable with metadata (name, description, parameter schema) — same info MCP already provides
2. Functions execute via a simple `fn(**args)` call, no IPython needed
3. IPython registration becomes optional: if IPython is available, register as convenience; if not, tools still work
4. MCP tools generate `ToolFunction` instances from their schema — no longer need separate ToolSpecs per function

**For the wildcard allowlist** (`discord.*`):
```python
# tool_allowlist could accept patterns
tool_allowlist = ["discord.*", "save", "shell"]
# Matches all ToolFunctions where group == "discord"
```

**For MCP safety hints** (read-only, destructive, etc.):
```python
@dataclass
class ToolFunction:
[...]

PR #2423, opened 2026-05-19
fix(tools): support glob tool allowlists

TimeToBuildBob:
## Summary
- add glob matching for tool allowlists so entries like `discord.*` work for grouped MCP tools
- apply the same matching when subagent profiles hard-restrict their tool set
- cover the loader and subagent profile paths with focused tests

Partial progress on #607.

## Test plan
- `uv run ruff check gptme/tools/__init__.py gptme/tools/subagent/execution.py tests/test_tools.py tests/test_tools_subagent.py`
- `uv run --with pytest python3 -m pytest tests/test_tools.py tests/test_tools_subagent.py -q -k "not test_subagent_wait_basic"`

## Notes
- `tests/test_tools_subagent.py::test_subagent_wait_basic` is already failing on `master` with the same missing-log / `module`-callable baseline break, so it was excluded from the verification run here.

### DIFF
--- a/gptme/prompts/templates.py
+++ b/gptme/prompts/templates.py
@@ -71,7 +71,11 @@ def prompt_short(
 ) -> Generator[Message, None, None]:
     """Short prompt to start the conversation."""
     yield from prompt_gptme(
-        interactive, model, agent_name=agent_name, tool_format=tool_format
+        interactive,
+        model,
+        agent_name=agent_name,
+        tool_format=tool_format,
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
Opened follow-up PR #2424 for the unresolved allowlist review feedback. It adds the MCP-drop migration warning, moves the glob matching helpers into a shared helper module, and tightens the defensive error wording.

PR #2425, opened 2026-05-19
fix(subagent): harden thread-mode wait path

TimeToBuildBob:
## Summary
- import the thread-mode subagent chat entrypoint directly from `gptme.chat` so `_create_subagent_thread()` calls the function instead of the module
- cache thread startup failures in the subagent result store so `subagent_wait()` returns a failure result instead of exploding on a missing log
- treat missing conversation logs as a clean failure status and add focused regression tests for both fallback paths

## Testing
- uv run --with pytest pytest tests/test_tools_subagent.py::test_subagent_status_missing_log_returns_failure tests/test_tools_subagent.py::test_subagent_wait_returns_failure_when_thread_startup_raises tests/test_tools_subagent.py::test_subagent_wait_basic -q
- uv run --with pytest pytest tests/test_tools_subagent.py -m "not slow and not eval" -q

### DIFF
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -433,6 +433,8 @@ def subagent(
             except Exception as e:
                 # If subagent creation fails, notify with error status
                 logger.error(f"Subagent {agent_id} failed during execution: {e}")
+                with _subagent_results_lock:
+                    _subagent_results[agent_id] = ReturnType("failure", str(e))
                 try:
                     notify_completion(agent_id, "failure", f"Execution failed: {e}")
                 except Exception as notify_err:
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Fixed the 4 CI test failures in 'Test without API keys' / 'Test without API keys or extras'.

Root cause: PR #2425 changed 'from gptme import chat' -> 'from gptme.chat import chat' in execution.py. Four tests used patch.object(gptme.chat, 'chat') which fails because gptme.chat now resolves to the function directly (via lazy loading in gptme/__init__.py), not a module with a 'chat' attribute.

Fix: Switched those 4 patches to patch('gptme.chat.chat', ...) which correctly resolves through the lazy import chain.

Verification: 4 passed in 1.92s locally.

TimeToBuildBob:
@greptileai review

ErikBjare:
@greptileai review

PR #2428, opened 2026-05-20
fix(tests): adapt server tests to isolated workspaces

TimeToBuildBob:
## What changed
- taught the server workspace test fixtures to replace `logdir/workspace` whether it is a symlink or the new empty directory created by `ChatConfig.from_logdir()`
- updated the V2 default-system-prompt test to pass an explicit workspace instead of relying on the old cwd fallback

## Why
`fix(config): give server sessions an isolated workspace directory` changed new conversations to eagerly create `logdir/workspace`. The test suite still assumed that path was absent or always a symlink, which broke `master` in the `Test without API keys` job after #2427 merged.

## Impact
- restores the server workspace test coverage on `master`
- keeps the new isolated-workspace behavior intact instead of reverting it to satisfy stale tests

## Validation
- `/home/bob/gptme/.venv/bin/python -m pytest tests/test_server_workspace.py tests/test_server_v2.py::test_v2_create_conversation_default_system_prompt -q`

### DIFF
--- a/tests/test_server_v2.py
+++ b/tests/test_server_v2.py
@@ -1143,6 +1143,15 @@ def test_v2_interrupt(v2_conv, client: FlaskClient):
     assert "interrupted" in data["message"].lower()
 
 
+def _normalize_config_for_comparison(config_dict: dict) -> dict:
+    """Normalize config dict for comparison by removing server-managed fields."""
+    result = config_dict.copy()
+    # workspace is now managed per-conversation by the server
+    if "chat" in result and "workspace" in result["chat"]:
+        del result["chat"]["workspace"]
+    return result
+
+
 def test_v2_chat_config_saved_on_conversation_create(client: FlaskClient):
     """Test that the chat config is saved on conversation create."""
     input_config = ChatConfig(model="openai/gpt-4o")
@@ -1167,7 +1176,9 @@ def test_v2_chat_config_saved_on_conversation_create(client: FlaskClient):
     print("old config", input_config.to_dict())
     print("-" * 80)
     print("new config", config.to_dict())
-    assert config.to_dict() == input_config.to_dict()
+    assert _normalize_config_for_comparison(
+        config.to_dict()
+    ) == _normalize_config_for_comparison(input_config.to_dict())
 
 
 def test_v2_chat_config_saved_separately_for_each_conversation(client: FlaskClient):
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Fix: Additional CI failure in config comparison tests

The initial fix (e577a0a) correctly handled the workspace directory lifecycle (empty dir vs symlink), but 3 additional tests failed on `Test without API keys` because `ChatConfig.to_dict()` now returns a per-conversation workspace path (e.g., `~/.local/share/gptme/logs/<conversation>/workspace`) instead of the server's cwd.

These tests were comparing full config dicts including the `workspace` field, which is now server-managed per-conversation.

### Root cause
When #2427 made server sessions create isolated workspace directories, `ChatConfig.from_logdir()` for a new conversation (no saved `config.toml`) returns `workspace=<logdir>/workspace` instead of the old behavior which fell through to the server process's cwd.

### Fix (3876bf6)
Added `_normalize_config_for_comparison()` helper that strips `workspace` from the config dict before comparison. Applied it to all 3 failing tests:
- `test_v2_chat_config_saved_on_conversation_create`
- `test_v2_chat_config_saved_separately_for_each_conversation`  
- `test_v2_chat_config_get_works`

These tests care about config round-tripping (model, tools, stream, interactive, etc.) — workspace isolation is tested separately in `test_server_workspace.py`.

### Verification
```bash
# All 3 previously-failing tests pass locally
python -m pytest tests/test_server_v2.py -k "chat_config" -v
# 3 passed
```

TimeToBuildBob:
## Round 3: Complete coverage of config comparison normalization

Two fixes:

### 1. `test_v2_chat_config_update_works` — applied `_normalize_config_for_comparison`
**Missed by both rounds 1 and 2.** This test has TWO `config.to_dict() == input_config.to_dict()` assertions (lines 1236 and 1246) that were not updated when `_normalize_config_for_comparison` was added to all other config comparison sites. When the server injects a per-conversation `workspace` field into the stored config via `ChatConfig.from_logdir()`, these raw comparisons will fail for the same reason the other tests were failing.

The Greptile bot flagged this in the first review. Now fixed.

### 2. `_replace_workspace_link` — switched from `Path.rmdir()` to `shutil.rmtree()`
**Review feedback from Greptile.** `Path.rmdir()` raises `OSError` if the workspace directory is non-empty (e.g., if `ChatConfig.from_logdir()` writes anything into it). `shutil.rmtree()` handles non-empty directories gracefully. `import shutil` was already present later in the file.

**CI**: The `Test without API keys` job is running at https://github.com/gptme/gptme/actions/runs/26147315258. This job previously failed with the 3 config comparison assertion errors that are now all guarded by `_normalize_config_for_comparison`.

TimeToBuildBob:
@greptileai review

PR #2440, opened 2026-05-22
feat(tools): add when-to-use trigger language for mcp and lessons

TimeToBuildBob:
## Summary

Adds explicit `### When to use` sections to the `mcp` and `lessons` tool instructions, matching the pattern shipped for shell, read, gh, browser, computer, morph, elicit, form, and others (idea #320 / #2406+).

**mcp tool**: clarifies that `/mcp` is the management interface (search/load/unload/list), not the way to call already-loaded server tools — those should be invoked directly as `<server-name>.<tool-name>`.

**lessons tool**: clarifies that lessons are auto-injected by keyword matching; only use `/lesson search` or `/lesson list` for explicit lookup when auto-injection didn't fire. This prevents agents from redundantly re-reading lessons already in context.

Closes the remaining coverage gap from the Phase 2b when-to-use audit.

## Test plan
- [x] Both files import cleanly (`uv run python3 -c "from gptme.tools.mcp import tool; from gptme.tools.lessons import tool"`)
- [x] Pre-commit hooks pass
- [x] Existing tests pass (`uv run pytest tests/ -k "mcp or lesson" -q`)

### DIFF
--- a/gptme/tools/lessons.py
+++ b/gptme/tools/lessons.py
@@ -453,6 +453,17 @@ tool = ToolSpec(
     name="lessons",
     desc="Lesson system for structured guidance",
     instructions="""
+### When to use the lessons tool
+
+Lessons are **automatically injected** when relevant keywords match — you rarely
+need to call the lessons tool explicitly. Use it only when:
+- Searching for guidance that wasn't auto-injected: `/lesson search <topic>`
+- Browsing available patterns: `/lesson list`
+- Refreshing after lesson files changed: `/lesson refresh`
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Follow-up fix is up in #2441.

The new `When to use` blocks in this PR pushed the OpenAI tool-format descriptions over the 1024-character limit enforced by `tests.test_tools::test_tool_descriptions_within_openai_limit` (`lessons` hit 1046, `mcp` hit 1292). #2441 keeps the new guidance but compresses the wording so the same test passes again.

PR #2468, opened 2026-05-23
fix(bot): update model default, fix checkout version, tighten permissions

TimeToBuildBob:
## Summary

Three fixes to the gptme-bot infrastructure that accumulated since the resolver was shipped:

- **Outdated model default** (`action.yml`): Default model was `claude-sonnet-4-20250514` (Sonnet 4.0, released May 2025). Updated to `claude-sonnet-4-6` (current Sonnet). Both `bot.yml` and `resolve.yml` already override this with the correct model, but external repos using the action without specifying a model were silently using an outdated model.

- **Nonexistent checkout action** (`bot.yml`): `actions/checkout@v6` does not exist (latest stable is v4). Updated to `@v4` to match `resolve.yml` and `action.yml`'s own checkout step. This would cause the bot workflow to fail on any run that reached the checkout step.

- **Overly broad permissions** (`bot.yml`): `permissions: write-all` replaced with specific `contents: write`, `issues: write`, `pull-requests: write` — matching the scoped permissions already used in `resolve.yml`.

- **Pinned script commit** (`action.yml`): Advanced the pinned script URL from `ba3c0233` to `248aac98` (includes the resolver duplicate-comment fix from #2447). External repos using the action now get the resolver improvements.

## Test plan

- [ ] Verify `bot.yml` workflow runs without failing on the checkout step
- [ ] Verify action works with default model when no `model:` input is provided
- [ ] CI passes

### DIFF
--- a/.github/actions/bot/action.yml
+++ b/.github/actions/bot/action.yml
@@ -10,7 +10,7 @@ inputs:
   model:
     description: 'Model to use'
     required: false
-    default: 'anthropic/claude-sonnet-4-20250514'
+    default: 'anthropic/claude-sonnet-4-6'
   github_token:
     description: 'GitHub Token'
     required: true
@@ -70,11 +70,11 @@ runs:
       run: |
         # Prefer the local script when running inside the gptme repo itself.
         # Fall back to downloading a pinned version for use as a reusable external action.
-        # Pin to specific commit for supply chain security (ba3c0233)
+        # Pin to specific commit for supply chain security (248aac98)
         if [ -f "scripts/github_bot.py" ]; then
           cp scripts/github_bot.py /tmp/github_bot.py
         else
-          SCRIPT_URL="https://raw.githubusercontent.com/gptme/gptme/ba3c0233f5f047ef210c732fb59d0c52b9bdd9e5/scripts/github_bot.py"
+          SCRIPT_URL="https://raw.githubusercontent.com/gptme/gptme/248aac98c02113009837c4b582dd72ed0604dd63/scripts/github_bot.py"
           curl -fsL "$SCRIPT_URL" > /tmp/github_bot.py
         fi
         python /tmp/github_bot.py --mode "${{ inputs.mode }}" --workspace .
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -4,7 +4,10 @@ on:
   issue_comment:
     types: [created]
 
-permissions: write-all
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
 
 jobs:
   run-bot:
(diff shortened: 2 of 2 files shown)


ErikBjare:
Related: what happened here? https://github.com/ActivityWatch/aw-tauri/issues/223#issuecomment-4525890558

TimeToBuildBob:
**Root cause**: aw-tauri's `gptme.yml` has the same outdated setup this PR fixes, plus a script version mismatch that caused the spiral.

Three things went wrong:

1. **Trigger with no filter**: aw-tauri's `gptme.yml` fires on every `issue_comment: created` — including your `@TimeToBuildBob` mention. The bot script is supposed to check internally whether to act, but it crashed before it could.

2. **Script version mismatch**: aw-tauri uses `ErikBjare/gptme/.github/actions/bot@master`, which now passes `--mode comment` but has the old pinned script (`ba3c0233`) that doesn't support that argument yet → `unrecognized arguments: --mode comment` → exit 2 → "I'm sorry..." reply.

3. **Crash spiral**: Each "I'm sorry..." reply is itself a new comment → triggers the workflow again → crash → another reply → three iterations.

**Fix**: After this PR merges into `gptme/gptme`, the same change needs to land in your fork (`ErikBjare/gptme`) OR aw-tauri's workflow should point to `gptme/gptme` action instead. Also worth considering adding a `if: github.actor != 'github-actions[bot]'` guard to the workflow job to break the crash spiral.

TimeToBuildBob:
@ErikBjare Root cause traced — and this PR was the fix. Chain:

1. `aw-tauri/.github/workflows/gptme.yml` triggers the bot action on **every** `issue_comment` (it's meant to check for `@gptme` internally and skip otherwise).
2. The bot composite action calls `github_bot.py --mode comment`.
3. Run **externally** (aw-tauri has no local `scripts/github_bot.py`), the action `curl`s the **pinned** script. The old pin was `ba3c0233`, which **predates the `--mode` flag** (added in #2445 / `930acf352`).
4. So `argparse` rejected `--mode comment` and exited 2 — *before* reaching the graceful "No @gptme command found → exit 0" path. Because it crashed before setting `GPTME_BOT_HANDLED_FAILURE=1`, the action's fallback `Report error` step posted "I'm sorry, I could not fulfill your request" on every comment.

That's why your `@TimeToBuildBob` comment (no `@gptme`) got a bot reply: it should have been silently skipped, but the stale script never got far enough to skip.

This PR bumps the pin `ba3c0233` → `248aac98` (which has `--mode` *and* the clean no-mention exit). Verified: `git show 248aac98:scripts/github_bot.py` returns 0 cleanly on a non-`@gptme` comment. Since the action is referenced `@master`, this is live now — the next comment on aw-tauri will no longer spam.

Follow-up (Greptile's only note, non-blocking): the `curl` download path has no sha256 verification. Tracked separately rather than scope-creeping this PR.

PR #2474, opened 2026-05-23
fix(vent): tolerate trailing punctuation in resolution-owner tag

TimeToBuildBob:
## Problem

The vent tool parses an optional trailing `Owner: <owner>` line into a structured `resolution_owner` field (added in #2463). The token extractor takes the first whitespace-delimited word verbatim:

```python
token = before_paren.split()[0].lower() if before_paren.split() else ""
```

Agents very commonly end that line with sentence punctuation — `Owner: tooling.`, `Owner: self!`, `Owner: operator,`. The trailing character makes the token (`"tooling."`) fail the `VALID_OWNERS` membership check, so the owner is **silently dropped** *and* the `Owner:` line is left polluting the recorded message. That quietly degrades friction-ledger data quality, which feeds the friction-analysis layer.

## Fix

Strip terminal punctuation (`.,;:!?`) from the extracted token. Only the ends are stripped, so internal characters are preserved: `self-evident` stays an unrecognized token rather than collapsing to `self`.

## Tests

- Parametrized `test_trailing_punctuation_tolerated` over `tooling.`, `tooling!`, `operator,`, `self;`, `upstream?`
- `test_internal_hyphen_not_misparsed` regression guard for the strip-only-ends behavior
- Full `test_vent.py`: 33 passed (RED→GREEN verified)

Surgical: 21 insertions, no behavior change for already-clean tags.

### DIFF
--- a/gptme/tools/vent.py
+++ b/gptme/tools/vent.py
@@ -99,6 +99,10 @@ def _parse_resolution_owner(code: str) -> tuple[str, str | None]:
     raw = match.group(1).strip()
     before_paren = raw.split("(", 1)[0]
     token = before_paren.split()[0].lower() if before_paren.split() else ""
+    # Tolerate trailing sentence punctuation, e.g. "tooling." or "self!".
+    # Only strips the ends, so internal characters like the hyphen in
+    # "self-evident" are preserved (and stay an unrecognized token).
+    token = token.strip(".,;:!?")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the Greptile finding: changed `rstrip` to `strip` in the test's expected-value computation so the test accurately represents the implementation's both-ends stripping behaviour. All 33 vent tests still pass.

PR #2488, opened 2026-05-24
feat(webui): opt into Chrome LNA for local-server fetches (targetAddressSpace)

TimeToBuildBob:
## Summary

Follow-up to #2483 (docs/banner guidance). This is the deeper **site-side code mitigation** for Chrome 142+ Local Network Access (LNA).

Chrome 142+ LNA blocks requests from a hosted page origin (e.g. `https://chat.gptme.org`) to a loopback/local `gptme-server` *before* CORS is evaluated. The documented browser opt-in is annotating the request:

```ts
fetch(url, { targetAddressSpace: 'local' })
```

## What changed

- New dependency-free helper `webui/src/utils/addressSpace.ts`:
  - `isLocalUrl(url)` — loopback / RFC1918 detection
  - `withLocalAddressSpace(url, init)` — adds `targetAddressSpace: 'local'` **only** when the target URL is local, so connections to remote/cloud servers (gptme.ai managed server) are unaffected.
- Applied at the webui fetch sites:
  - `providerStatus.ts` — the `/api/v2` connection probe (fires first)
  - `api.ts` — central `ApiClient` fetch paths (`fetchWithTimeout`, `fetchJson`), auth-cookie, workspace upload
  - `taskApi.ts` — centralized via `getFetchOptions()`

The field is not yet in the TS DOM lib types, so it is attached via a `RequestInit` cast (isolated inside the helper).

## Why a separate, dependency-free module

`connectionConfig.ts` uses `import.meta.env`, which jest cannot import (every test mocks it). Putting the pure helpers in `addressSpace.ts` keeps the local-only conditional **directly unit-testable**.

## Tests

- New `addressSpace.test.ts` — covers the local-only conditional, init-field preservation, no-mutation, and edge cases (invalid URL, 172.16–31 boundary).
- `tsc --noEmit` clean; existing `api.test.ts` / `ApiContext.test.tsx` still pass.

## Verification gap (honest)

This **cannot be E2E-verified headless** — headless Chromium auto-denies LNA regardless of the annotation. It needs a real Chrome 142+ with a localhost `gptme-server` and a user granting/denying the prompt. The unit tests verify the conditional logic (the part that could regress silently); the annotation itself follows the documented Chrome API.

Ref: https://developer.chrome.com/blog/local-network-access

### DIFF
--- a/webui/src/components/__tests__/SetupWizard.test.tsx
+++ b/webui/src/components/__tests__/SetupWizard.test.tsx
@@ -251,6 +251,7 @@ describe('SetupWizard', () => {
     await waitFor(() => {
       expect(mockFetch).toHaveBeenCalledWith('http://127.0.0.1:5700/api/v2', {
         headers: {},
+        targetAddressSpace: 'local',
       });
     });
 
--- a/webui/src/utils/__tests__/addressSpace.test.ts
+++ b/webui/src/utils/__tests__/addressSpace.test.ts
@@ -1,0 +1,72 @@
+import { isLocalUrl, withLocalAddressSpace } from '../addressSpace';
+
+describe('isLocalUrl', () => {
+  it.each([
+    'http://localhost:5700',
+    'http://127.0.0.1:5700',
+    'http://10.0.0.5:5700',
+    'http://192.168.1.100:5700',
+    'http://172.16.0.1:5700',
+    'http://172.31.255.255:5700',
+  ])('returns true for local/private address %s', (url) => {
+    expect(isLocalUrl(url)).toBe(true);
+  });
+
+  it.each([
+    'https://gptme.ai',
+    'https://chat.gptme.org',
+    'https://api.example.com:8080',
+    'http://172.15.0.1:5700', // 172.15 is public (outside 172.16-31)
+    'http://172.32.0.1:5700', // 172.32 is public
+  ])('returns false for remote/public address %s', (url) => {
+    expect(isLocalUrl(url)).toBe(false);
+  });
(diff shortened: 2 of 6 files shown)


TimeToBuildBob:
Addressed Greptile's findings in 509c8c26d:

- **Duplicate regex fixed**: `isLikelyChromeCorsPna` now delegates to `isLocalUrl()` from `addressSpace.ts` instead of defining its own inline `privatePattern`. Single source of truth.
- **IPv6 loopback (`::1`)**: Left as-is for now — the `PRIVATE_ADDRESS_PATTERN` is intentionally scoped to IPv4 loopback/RFC1918. IPv6 loopback is an edge case (virtually no one runs `gptme-server` on `::1` only) and adding it would require a separate regex branch or URL normalization. Can follow up if it comes up.
- **Test fix**: Updated `SetupWizard.test.tsx` to expect `targetAddressSpace: 'local'` in the fetch call for `127.0.0.1` URLs (was failing CI because the assertion didn't account for the new field).

TimeToBuildBob:
Fixed CI failure in d4b8f1f:

`isLikelyChromeCorsPna` was using `window.location.href` to detect the current origin, but jsdom test mocks only set `window.location.hostname` — `href` kept the default `http://localhost/` value, so `isPublicOrigin` was always `false` and the function returned `false` for every input.

Fix: read `window.location.hostname` directly and wrap it as `http://<hostname>` for the `isLocalUrl` check (same `PRIVATE_ADDRESS_PATTERN`, no new regex).

PR #2496, opened 2026-05-24
fix(cli): reject negative chats --limit instead of crashing islice()

TimeToBuildBob:
## Problem

`gptme-util chats list --limit -5` crashes with an uncaught traceback:

```
ValueError: Indices for islice() must be None or an integer: 0 <= x <= sys.maxsize.
```

Found by dogfooding the CLI surface. The negative `--limit` flows straight into `itertools.islice()` via `list_conversations()`. The `--json` path crashes the same way, and `chats search --limit -N` shares the option.

## Fix

- **CLI boundary**: `click.IntRange(min=1)` on the `--limit` option of `chats list` and `chats search` — bad input now gets a clean Click error + non-zero exit instead of a traceback. (Other `gptme-util` error paths like `skills show <missing>` and `mcp info <missing>` already exit non-zero; this brings `chats` in line.)
- **Library defense-in-depth**: `list_conversations()` clamps negative limits to `0` (empty result), so non-CLI callers (webui/server) can't trigger the same `islice` crash.

## Behavior change

`--limit 0` and negatives are now rejected (exit 2). A limit of 0 ("show 0 chats") was a no-op before; this makes the contract explicit.

## Tests

- `tests/test_util_cli.py::test_chats_list_negative_limit` — CLI rejects `-5` on both plain and `--json` paths, no traceback.
- `tests/test_conversations.py::test_list_conversations_limit_bounds` — library clamps `-5`/`0` to empty, `1`/`10` return expected counts.

Verified the original code reproduces the crash and the fix resolves it. `ruff check` clean, touched-file tests pass (49 passed, 1 skipped).

### DIFF
--- a/gptme/cli/cmd_chats.py
+++ b/gptme/cli/cmd_chats.py
@@ -164,7 +164,13 @@ def chats_search(
 
 @chats.command("read")
 @click.argument("id")
-@click.option("-n", "--limit", default=20, help="Maximum number of messages to show.")
+@click.option(
+    "-n",
+    "--limit",
+    default=20,
+    type=click.IntRange(min=1),
+    help="Maximum number of messages to show.",
+)
 @click.option("--system", is_flag=True, help="Include system messages.")
 @click.option(
     "-c", "--context", default=0, help="Messages of context before start message."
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed the Greptile P2 finding in 2d8c9c396: added `type=click.IntRange(min=1)` to `chats read --limit` for consistency with `chats list` and `chats search`. All three `chats` sub-commands now guard against negative/zero limits at the CLI boundary. CI re-running.

PR #2515, opened 2026-05-25
fix(chats): handle non-string content in chats search and format

TimeToBuildBob:
## Problem
`gptme-util chats search ""` crashes with `AttributeError: int object has no attribute lower` when a conversation log contains a message whose content field is an int instead of a str.

This occurs in two places:
1. `_get_matching_messages()` — iterating over all messages, calling `.content.lower()` on int content
2. `_format_message_with_context()` — receiving int content as the first argument

## Fix
- **`_get_matching_messages`**: skip messages where `not isinstance(msg.content, str)` 
- **`_format_message_with_context`**: return `[non-string content]` placeholder for non-string input
- **`search_chats`**: added early-check to reject empty/whitespace-only queries before scanning conversations (avoids crashing the same path)
- **Two regression tests**: integration test for `search_chats` with non-string content, unit test for `_format_message_with_context` with int content

## Testing
`python3 -m pytest tests/test_tools_chats.py -v` — all 9 tests pass (7 existing + 2 new)

## Found via dogfooding
Part of a gptme CLI surface dogfooding sweep: exercised `gptme-util chats search` with empty string and discovered the crash.

### DIFF
--- a/gptme/tools/chats.py
+++ b/gptme/tools/chats.py
@@ -25,12 +25,18 @@ def _get_matching_messages(
     """Get messages matching the query."""
     if not query.strip():
         return []
-    return [
-        (i, msg)
-        for i, msg in enumerate(log_manager.log)
-        if isinstance(msg.content, str) and query.lower() in msg.content.lower()
-        if msg.role != "system" or system
-    ]
+    matching = []
+    for i, msg in enumerate(log_manager.log):
+        if msg.role == "system" and not system:
+            continue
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed both Greptile P2 comments: (1) added `logger.warning` for skipped non-string content, (2) narrowed type hint to `str | object`. Re-triggering review.

@greptile apps re-review

PR #2518, opened 2026-05-25
fix(cli): reject unsafe conversation names

TimeToBuildBob:
## Summary
- reject `--name` values that are not a single safe path component
- keep a defense-in-depth validation in `get_logdir()`
- add a CLI regression test for `--name ../bad-name`

## Repro
```bash
uv run gptme -n --name ../bad-name "hello"
```
Before this change, the CLI fell through into logdir creation and died later with `Too many levels of symbolic links`.

## Verification
- `uv run gptme -n --name ../bad-name "hello"`
- `uv run python -m pytest tests/test_cli.py -k "test_name_rejects_path_traversal or test_help or test_version"`

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -120,10 +120,10 @@ def _conversation_name_error(value: str) -> str | None:
         return "conversation name cannot be empty."
     if len(value.encode()) > _MAX_CONVERSATION_NAME_BYTES:
         return "conversation name too long."
-    if "/" in value or "\\" in value or ".." in value:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob (commented):
✅ Addressed the `.` bypass: both `_conversation_name_error` and the regression test now cover `.` alongside `..`, `/`, and `\\\\`.

Commit: `becdc7424` — parametrized `test_name_rejects_path_traversal` to cover `.`, `..`, `foo/bar`, and `foo\\\\bar`. All 5 parametrize cases pass.

PR #2524, opened 2026-05-25
fix(server): use get_json(silent=True) in tasks endpoints

TimeToBuildBob:
## Summary

- `POST /api/v2/tasks` and `PUT /api/v2/tasks/<id>` were using `flask.request.json` directly, which raises a `BadRequest` exception on malformed JSON and returns Flask's default HTML 400 page
- All other v2 endpoints already use `get_json(silent=True)` (fixed in #2521 for conversation endpoints)
- This PR applies the same fix to the tasks endpoints for consistency

## Reproduction

```bash
# Before fix: returns HTML 400
curl -s -X POST http://localhost:5700/api/v2/tasks \
  -d "not-json"
# → <html><p>The browser (or proxy)...

# After fix: returns JSON 400
# → {"error": "No JSON data provided"}
```

## Test plan

- [x] `test_create_invalid_json_body` — malformed JSON to POST returns JSON error
- [x] `test_create_array_json_body` — array body to POST returns "must be an object"
- [x] `test_update_invalid_json_body` — malformed JSON to PUT returns JSON error
- [x] `test_update_array_json_body` — array body to PUT returns "must be an object"
- [x] All 87 existing tests pass

### DIFF
--- a/tests/test_tasks_api.py
+++ b/tests/test_tasks_api.py
@@ -612,6 +612,27 @@ class TestTasksUpdateAPI:
         )
         assert resp.status_code == 400
 
+    def test_update_invalid_json_body(self, client, sample_task):
+        # Malformed JSON body must return a JSON error, not an HTML 400 page
+        resp = client.put(
+            f"/api/v2/tasks/{sample_task.id}",
+            data="this is not json",
+            content_type="application/json",
+        )
+        assert resp.status_code == 400
+        assert resp.is_json
+        assert "error" in resp.json
+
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Rebased onto master after #2523 landed (which covered the same `get_json(silent=True)` core fix for tasks + agents endpoints). The only conflict was a minor error message wording difference — kept master's phrasing. This PR now only adds 4 tests to `tests/test_tasks_api.py` (class-based, using the `sample_task` fixture) that complement the tests #2523 added to `tests/test_api_v2.py`.

PR #2556, opened 2026-05-26
fix(server): reject null bytes in conversation IDs and branch names

TimeToBuildBob:
## Summary

Adds null-byte (`\\x00`) rejection to conversation ID and branch name validators. Null bytes pass existing path traversal checks but cause `ValueError: embedded null byte` when the OS attempts filesystem operations.

Discovered during dogfooding the gptme-server HTTP API with edge-case inputs (Bob session c257 — ErikBjare/bob#799 dogfood task rotation).

## Changes

- **`gptme/util/conversation_ids.py`**: add `\\x00` to the `conversation_id_error()` reject list
- **`gptme/server/api_v2_common.py`**: add `\\x00` to the `_validate_branch()` reject list

## Testing

All 85 path traversal tests pass (incl. existing null-byte branch tests). No regressions.


TimeToBuildBob:
Addressed the Greptile P2 finding: updated the error message in `conversation_id_error()` to mention null bytes alongside the other forbidden characters. See commit 0816c4fab.

PR #2575, opened 2026-05-26
fix(cli): raise UsageError for unknown model in llm generate

TimeToBuildBob:
## Summary

- `gptme-util llm generate --model notareal/model hello` crashed with a raw `ValueError` traceback from `init()` — no try/except caught it
- The existing `get_provider_from_model()` try/except inside `redirect_stderr()` printed to captured stderr, so the error message was silently swallowed
- Both are now converted to `click.UsageError`, which propagates cleanly out of the `redirect_stderr` context and click renders it as a standard `Error: ...` message with exit code 2

## Reproduction

```
$ gptme-util llm generate --model notareal/model hello
  ...
ValueError: Unknown model 'notareal/model'. Use 'provider/model' with a known provider...
```

After fix:

```
$ gptme-util llm generate --model notareal/model hello
Usage: gptme-util llm generate [OPTIONS] [PROMPT]
Try 'gptme-util llm generate --help' for help.

Error: Unknown provider: notareal
```

## Test plan

- [x] `tests/test_util_cli.py::test_llm_generate_unknown_model` — new regression test
- [x] Full `tests/test_util_cli.py` suite: 27 passed, 1 skipped

### DIFF
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -380,7 +380,10 @@ def llm_generate(prompt: str | None, model: str | None, stream: bool):
         console.quiet = True
 
         # Initialize with minimal setup - no tools needed for simple generation
-        init(model, interactive=False, tool_allowlist=[], tool_format="markdown")
+        try:
+            init(model, interactive=False, tool_allowlist=[], tool_format="markdown")
+        except ValueError as e:
+            raise click.UsageError(str(e)) from e
 
         # Get model or use default
         if not model:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Good catch from Greptile. The no-default-model path had the same silent-swallow bug: it was printing to `sys.stderr` and calling `sys.exit(1)` while still inside the `with redirect_stderr(stderr_capture):` context, so the error message was silently discarded.

Fixed in the latest commit by switching to `raise click.UsageError(...)`, consistent with how the `ValueError` paths are handled. Click renders the error to the real stderr after exiting the context manager.

PR #2589, opened 2026-05-27
fix(cli): reject whitespace-only conversation names

TimeToBuildBob:
## Summary
- reject whitespace-only `--name` values the same way we already reject empty conversation names
- apply the same validation to server/API conversation ids so the rule stays consistent
- add CLI and server regression coverage

## Repro
Before this patch, `gptme -n --name " "` accepted the conversation name, created a garbage logdir, and printed a blank resume hint.

## Testing
- `uv run pytest tests/test_cli.py tests/test_server_path_traversal.py -q`
- `uv run ruff check gptme/util/conversation_ids.py gptme/logmanager/manager.py tests/test_cli.py tests/test_server_path_traversal.py`


TimeToBuildBob:
@greptileai review

PR #2591, opened 2026-05-27
fix(hooks): never crash gptme-util hooks run on malformed CC event JSON

TimeToBuildBob:
## Problem

`gptme-util hooks run` is the command CC registers in `settings.json` for lesson injection. Its design contract (per the docstring and existing fallbacks) is that it **must always emit valid JSON** (`{"continue": true}`) and never crash — a traceback would break the hook integration for every prompt/tool call in that CC session.

Empty stdin and invalid JSON already degrade cleanly, but valid-but-malformed events still crashed:

| Input | Crash |
|-------|-------|
| Top-level JSON that isn't an object (`[1,2,3]`, `12345`, `null`) | `AttributeError: 'list' object has no attribute 'get'` |
| Non-string `prompt` (`{"prompt": 123}`) | `AttributeError: 'int' object has no attribute 'strip'` |
| Non-list `transcript` (`{"transcript": 999}`) | `TypeError: 'int' object is not subscriptable` |
| `transcript` entries that aren't dicts (`["x", 42]`) | `AttributeError: 'str' object has no attribute 'get'` |

Found by dogfooding the CLI surface with malformed input.

## Fix

Guard the three input-boundary spots so a malformed event degrades to the existing no-match fallback instead of raising:
- reject non-object top-level JSON right after `json.loads`
- require `match_text` to be a `str` before `.strip()`
- in `_extract_pretooluse_text`, skip a non-list `transcript` and non-dict entries

## Test

Adds a parametrized regression test (`test_run_malformed_input_returns_continue`) covering each variant; all assert `exit_code == 0` and `continue == True`.

```
36 passed in 1.40s
```
ruff check + format clean.

### DIFF
--- a/gptme/cli/cmd_hooks.py
+++ b/gptme/cli/cmd_hooks.py
@@ -285,9 +285,16 @@ def run(workspace: Path | None = None) -> None:
             # No input — return empty (CC may call hooks with empty stdin)
             _output_empty()
             return
-        hook_input: dict = json.loads(raw)
+        hook_input = json.loads(raw)
     except (json.JSONDecodeError, OSError) as e:
         logger.debug("Failed to parse hook input: %s", e)
+        _output_empty()
+        return
+
+    if not isinstance(hook_input, dict):
+        # Valid JSON but not an object (e.g. a list or scalar) — CC always
+        # sends an object, so anything else is malformed. Pass through.
+        logger.debug("Hook input is not a JSON object: %s", type(hook_input).__name__)
         _output_empty()
         return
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Post-merge CI note: the `API tests with openai/gpt-4o-mini` job failed at the **Install dependencies** step (not at test execution). This is a poetry dependency resolution failure in the CI environment, unrelated to the hook robustness changes in this PR. The test suite itself passed cleanly in the other API job (`openrouter/anthropic/claude-haiku-4-5`). The remaining "Test without API keys" job is still running.

PR #2595, opened 2026-05-27
fix(server): validate tools field type before calling init_tools

TimeToBuildBob:
## Summary

Validates the `tools` field type in `api_conversation_post` before passing it to `init_tools()`. A non-list value (string, int) causes `get_toolchain()` to iterate over the value as a character sequence, producing a confusing `Tool 't' not found` 500 error instead of a clear 400.

## Changes

- **`gptme/server/api_v2.py`**: Added type check for `tools` field in `api_conversation_post()` — non-list values now return `400` with ``tools must be a list of strings``
- **`tests/test_server.py`**: Added regression test covering string, integer, and valid-list tools values

Found while dogfooding the gptme server/HTTP API surface.

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -335,13 +335,18 @@ def test_api_v2_conversation_post_tools_validation(conv, client: FlaskClient):
         json={"role": "user", "content": "hello", "tools": 42},
     )
     assert response.status_code == 400
+    data = response.get_json()
+    assert "tools" in data.get("error", "").lower()
 
     # Valid list of tools should still work
     response = client.post(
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed both Greptile review findings in b3f5f30:

1. Integer sub-case assertion: Added `assert "tools" in data.get("error", "").lower()` to the integer case for consistency with the string case.
2. Happy-path failure message: Added a descriptive assertion message that prints the status code and response body when the valid-tools-list case fails, instead of a bare `assert response.status_code == 200`.

PR #2606, opened 2026-05-27
fix(autocompact): preserve tool-call messages

TimeToBuildBob:
## Summary
- protect assistant messages that contain real tool-use syntax from generic truncation
- skip Phase 3 assistant-message compression for tool-call messages so paired tool results keep a parseable anchor
- add regression tests for both `reduce_log` and `auto_compact_log`

## Testing
- `PYTHONPATH=/home/bob/gptme-tmp-compaction-0d1d VIRTUAL_ENV=/home/bob/gptme/.venv PATH=/home/bob/gptme/.venv/bin:$PATH pytest /home/bob/gptme-tmp-compaction-0d1d/tests/test_reduce.py /home/bob/gptme-tmp-compaction-0d1d/tests/test_auto_compact.py -q`

### DIFF
--- a/gptme/tools/autocompact/engine.py
+++ b/gptme/tools/autocompact/engine.py
@@ -19,7 +19,7 @@ from ...util.master_context import (
     create_master_context_reference,
 )
 from ...util.output_storage import create_tool_result_summary
-from ...util.reduce import reduce_log
+from ...util.reduce import message_contains_tool_use, reduce_log
 from .scoring import compress_content
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed the Greptile finding: the Phase 3 regression test was passing trivially because no non-tool-use assistant message exceeded the 1000-token threshold, so `needs_phase3_compression=False` and `auto_compact_log` returned early without Phase 3 ever running.

Fix in f7e899bd9:
- Replaced the short "Short response" assistant message with a long (>1200 token) non-tool-use message — this sets `needs_phase3_compression=True` and makes Phase 3 actually run
- Increased `tool_msg` from 220 to 300 `echo` lines so it's >1000 tokens and the guard has something meaningful to protect
- Added an assertion that `compacted[1]` (the non-tool-use message) was actually compressed, proving Phase 3 ran before we check that `tool_msg` was preserved

The test now verifies the guard fires under real Phase 3 conditions rather than passing vacuously.

TimeToBuildBob:
@greptileai review

PR #2635, opened 2026-05-30
fix(server): emit Access-Control-Allow-Private-Network on CORS preflight

TimeToBuildBob:
## Problem

Found during a user-testing pass on the production surfaces: loading
`https://chat.gptme.org` and watching it try to reach a local `gptme serve`
throws a browser error:

```
Access to fetch at 'http://127.0.0.1:5700/api/v2' from origin
'https://chat.gptme.org' has been blocked by CORS policy: Permission was
denied for this request to access the `loopback` address space.
```

This is Chrome's **Private Network Access** policy: a secure *public* origin
fetching a *loopback* address requires the preflight response to carry
`Access-Control-Allow-Private-Network: true`. flask-cors defaults this header
to `false`, so even when the user explicitly allows the origin via
`--cors-origin`, the browser blocks the connection.

## Verification (before)

`gptme-server --cors-origin https://chat.gptme.org`, then a PNA preflight:

```
$ curl -s -D - -o /dev/null -X OPTIONS http://127.0.0.1:5700/api/v2 \
    -H "Origin: https://chat.gptme.org" \
    -H "Access-Control-Request-Method: GET" \
    -H "Access-Control-Request-Private-Network: true"
Access-Control-Allow-Origin: https://chat.gptme.org
Access-Control-Allow-Private-Network: false   # <-- browser blocks
```

## Fix

Enable `allow_private_network` in the CORS resource config. This is gated by
the user's explicit `--cors-origin` opt-in, which is exactly the consent signal
for the hosted-webui → local-server connection. No effect when `--cors-origin`
is unset (default local-only mode).

## Verification (after)

```
Access-Control-Allow-Private-Network: true    # <-- connection permitted
```

Plus a regression test (`test_cors_preflight_allows_private_network`); full
`tests/test_server_cors.py` is green (8 passed).

## Context

Complements #2290 (webui stopped *retrying* on PNA failures) — that was the
client-side band-aid; this is the server-side fix that lets the connection
actually succeed.

### DIFF
--- a/gptme/server/app.py
+++ b/gptme/server/app.py
@@ -109,7 +109,11 @@ def create_app(
                 origins_list[0] if len(origins_list) == 1 else origins_list
             )
             # Browsers reject credentials with a wildcard origin.
-            allow_credentials = "*" not in origins_list
+            # Similarly, wildcard CORS must not opt in to Private Network
+            # Access — doing so lets *any* HTTPS page reach the local server,
+            # defeating Chrome's PNA protection. Both flags require named origins.
+            non_wildcard = "*" not in origins_list
+            allow_credentials = non_wildcard
             CORS(
                 app,
                 resources={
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Good catch from Greptile — the wildcard case was a real gap.

Fixed in 4c6c8e197: `allow_private_network` is now gated on the same `non_wildcard` boolean as `allow_credentials`. When `--cors-origin *` is used, PNA stays false (any HTTPS page being able to reach loopback would defeat Chrome's protection entirely). A new `test_cors_wildcard_disables_private_network` test covers this path; all 9 CORS tests pass.

TimeToBuildBob:
@greptileai review

PR #2684, opened 2026-06-01
docs(server): add nginx reverse proxy guide for self-hosters

TimeToBuildBob:
## Summary

Adds a **Production Deployment: nginx Reverse Proxy** section to `docs/server.rst`, the natural next step after the docker-compose self-host path shipped in #2681.

Covers:
- TLS termination with Let's Encrypt / certbot
- Loopback bind (`127.0.0.1:5700`) so the raw server port isn't exposed to the internet — the proxy is the only public surface
- A complete nginx site config (HTTP→HTTPS redirect + TLS server block)
- The **SSE-critical** settings: `proxy_buffering off` + long `proxy_read_timeout`, since gptme-server streams responses over `text/event-stream` (`gptme/server/api_v2_sessions.py`). Without these, nginx buffers the stream and the chat appears to hang until each full response is ready.
- A local-only alternative (SSH tunnel) for users who don't need public exposure

## Verification
- Built `docs/server.rst` with sphinx + the full docs extension set; build succeeds and the new section renders (`proxy_buffering`, `Production Deployment` present in output HTML). No new warnings on `server.rst`.
- Config values (port 5700, `GPTME_SERVER_TOKEN`, `CORS_ORIGIN`, `GPTME_SERVER_PORT`) cross-checked against `docker-compose.yml` and `.env.example`.

Closes idea-backlog #321 (Bob's workspace).

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -105,6 +105,96 @@ Key ``.env`` settings:
 - ``CORS_ORIGIN`` — origin allowed to call the server (defaults to the hosted web UI).
 - ``GPTME_SERVER_PORT`` — host port to publish (the container always listens on 5700).
 
+Production Deployment: nginx Reverse Proxy
+------------------------------------------
+
+The docker-compose setup above publishes the server on a plain HTTP port
+(``5700`` by default). For a public-facing deployment you should put it behind a
+reverse proxy that terminates TLS and forwards requests to the container. The
+example below uses nginx with a Let's Encrypt certificate.
+
+.. warning::
+
+   Do **not** expose the raw ``5700`` port to the internet. Bind the published
+   port to loopback so only the proxy can reach it. In ``.env`` (or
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the two issues from the Greptile review:

1. **certbot `--standalone` port conflict**: Switched to `certbot --nginx` plugin (`python3-certbot-nginx` package). The nginx plugin handles the ACME HTTP challenge through the running nginx process, so there's no port 80 conflict — unlike `--standalone` which tries to bind its own listener and fails when nginx is already running.

2. **Missing TLS protocol/cipher directives**: Added `ssl_protocols TLSv1.2 TLSv1.3` and `ssl_ciphers` using the Mozilla "intermediate" profile (ECDHE/DHE with AES-GCM and ChaCha20). This prevents TLS 1.0/1.1 from being offered on older nginx builds whose compiled-in defaults may include weak protocol versions.

PR #2690, opened 2026-06-01
fix(webui): polyfill structuredClone in jest.setup for jsdom

TimeToBuildBob:
jsdom (the jest test environment) does not expose `structuredClone` even though
Node 17+ has it. This caused the test added in #2687 to fail on master:

```
ReferenceError: structuredClone is not defined
  at Object.getConversation (src/utils/demoApiClient.ts:201:37)
```

Fix: add a one-line polyfill in `jest.setup.ts` guarded by a `typeof` check,
so all tests in the jsdom environment have `structuredClone` available, and it
is a no-op in any future environment that already exposes it.

### DIFF
--- a/webui/jest.setup.ts
+++ b/webui/jest.setup.ts
@@ -1,5 +1,10 @@
 import '@testing-library/jest-dom';
 
+// Polyfill structuredClone for jsdom — Node 17+ has it but jsdom doesn't expose it.
+if (typeof structuredClone === 'undefined') {
+  global.structuredClone = <T>(val: T): T => JSON.parse(JSON.stringify(val));
+}
+
 // Shim Vite import.meta.env for Jest.
 // connectionConfig.ts and SetupWizard.tsx wrap import.meta.env accesses in
 // Function() and fall back to process.env when that throws.


ErikBjare:
@TimeToBuildBob You're not going to address this?

TimeToBuildBob:
Addressing the Greptile note: the JSON round-trip limitation (`JSON.parse(JSON.stringify(val))`) is intentional for this use case. The polyfill only needs to handle conversation objects in `demoApiClient.ts`, which are plain JSON-serializable structures — no `Date`, `Map`, `Set`, or circular references. If a future test needs those types, the round-trip would give a clear wrong-value failure rather than silent loss (since the test assertions would catch the mismatch).

If you'd prefer a stricter fallback, we could use `require('v8').serialize`/`deserialize` (handles all structured-clone types in Node.js), but I think that's overkill for a jest setup polyfill scoped to the webui tests.

CI is all green and this is confined to test setup — no production code change. Ready to merge when you are.

TimeToBuildBob:
Addressed in follow-up PR #2741 — added a comment to the polyfill documenting the JSON.parse/stringify limitations (drops undefined, coerces Date, throws on BigInt, Map/Set → {}). Safe for current conversation objects but now explicit for future test authors.

PR #2691, opened 2026-06-01
fix(webui): replace structuredClone with JSON.parse/stringify for jsdom compat

TimeToBuildBob:
## Summary

Fixes master WebUI CI failure introduced by #2687.

`structuredClone` is not available in Jest's `jsdom` test environment (the
environment configured in `jest.config.ts`), causing `demoApiClient.test.ts`
to fail with `ReferenceError: structuredClone is not defined` on Node 20.x CI.

`DEMO_CONV_RESPONSE` is plain JSON-serializable data, so
`JSON.parse(JSON.stringify(x))` is equivalent and works in all environments
(Node, jsdom, older browsers).

## Test plan
- [ ] `demoApiClient.test.ts` passes (was the failing suite)
- [ ] WebUI `build (20.x)` check goes green

### DIFF
--- a/webui/src/utils/demoApiClient.ts
+++ b/webui/src/utils/demoApiClient.ts
@@ -198,7 +198,7 @@ export function createDemoApiClient(baseUrl: string = DEMO_BASE_URL): IApiClient
 
     // Conversation detail — serve the fixture or local in-memory conversations.
     getConversation: async (logfile) => {
-      if (logfile === DEMO_CONV_ID) return structuredClone(DEMO_CONV_RESPONSE);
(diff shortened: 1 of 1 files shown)


ErikBjare:
@TimeToBuildBob Could/should this get reverted if https://github.com/gptme/gptme/pull/2690 merges?

TimeToBuildBob:
Yes — if #2690 merges, reverting #2691 makes sense. The polyfill approach is cleaner: it fixes the test environment once rather than changing call sites, and it restores `structuredClone` on the production path (which handles non-JSON types like `Date`, `Map`, `Set` correctly, unlike the round-trip).

That said, `DEMO_CONV_RESPONSE` is plain JSON data, so both approaches work identically for this specific fixture. Low priority either way — whichever you prefer.

PR #2693, opened 2026-06-02
fix(self-heal): handle invalid ANTHROPIC_API_KEY gracefully

TimeToBuildBob:
## Summary

The `self-heal` workflow was failing on master CI because `ANTHROPIC_API_KEY` is invalid/expired, causing `github_ci_self_heal.py` to crash with a 401 `AuthenticationError` (exit 1).

- Catch `anthropic.AuthenticationError` and return an informative message — PRs get a comment explaining the secret needs updating
- Catch other `anthropic.APIError` variants and exit 0 silently
- Add a `[ -s /tmp/analysis.txt ]` guard in the workflow to skip posting when there's no content
- Change empty-response exit code from 1 → 0 (graceful skip, not error)

The immediate fix for the expired key itself requires a maintainer to update `ANTHROPIC_API_KEY` in **Settings → Secrets and variables → Actions**.

## Test plan

- [ ] Workflow no longer marks master CI as failed when `ANTHROPIC_API_KEY` is invalid
- [ ] When the key IS valid, workflow still posts the analysis comment as before
- [ ] When analysis.txt is empty (other API errors), no comment is posted

### DIFF
--- a/.github/workflows/self-heal.yml
+++ b/.github/workflows/self-heal.yml
@@ -111,6 +111,12 @@ jobs:
           BOT_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           PR_NUMBER: ${{ steps.pr.outputs.number }}
         run: |
+          # Skip if analysis produced no content (e.g. transient API error).
+          if [ ! -s /tmp/analysis.txt ]; then
+            echo "No analysis content — skipping comment"
+            exit 0
+          fi
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile P2 feedback in e43c61d: moved `Anthropic()` client construction inside the `try` block so empty/missing-key failures are caught by the `AuthenticationError` handler instead of crashing at import. Greptile thread resolved.

PR #2696, opened 2026-06-02
fix(server): load conversations from validated logdir

TimeToBuildBob:
Follow-up to #2694.

## Summary

- Load the conversation with the already-computed absolute `logdir`
- Avoid re-resolving the relative `conversation_id` through `LogManager.load`
- Keep the conversation existence check and branch lookup on the same path

## Verification

- `poetry run pytest tests/test_server_path_traversal.py::TestFileAttachmentPathValidation -q`
- `poetry run pytest tests/test_server_path_traversal.py -q`
- `poetry run ruff check gptme/server/api_v2.py`
- `git diff --check -- gptme/server/api_v2.py`

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -862,7 +862,7 @@ def api_conversation_post(conversation_id: str):
             404,
         )
     try:
-        log = LogManager.load(conversation_id, branch=branch)
+        log = LogManager.load(logdir, branch=branch)
     except FileNotFoundError:
         return (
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
All checks green, Greptile 5/5. This is a one-line fix that's safe to merge.

Self-merge is denied because the file touches `gptme/server/api_v2.py` (server code path), which requires human review under the self-merge policy. @ErikBjare, this is ready for you.

PR #2698, opened 2026-06-02
fix(logs): quiet normal runtime noise

TimeToBuildBob:
## What

Quiets normal gptme runtime output by moving implementation-detail logs out of the default info stream, filtering Anthropic thinking signature artifacts from regular message display, and clarifying the `/cost` / `/tokens` cache/cost columns.

Changes:
- downgrade repeated model alias resolver messages from info to debug
- downgrade ephemeral cleanup and reduced-log messages from info to debug
- strip serialized `think-sig` artifacts from regular formatted message output while preserving visible `<think>` content
- update the hidden-message hint to point at `/log --hidden` instead of a startup-looking flag
- clarify per-step token/cost columns as `TotalIn`, `Uncached`, `Output`, `CacheR`, `CacheW`, and `Cost`
- rename the aggregate `Hit rate` label to `Cache hit rate`
- render tiny positive costs as `<$0.0001` instead of `$0.0000`

## Why

ErikBjare/bob#834 reports that ordinary short `gptme` runs print confusing operational noise on every step. These messages are useful diagnostics, but they should not be normal user-facing output.

The same issue also showed confusing `/cost` output where cache reads and writes were mixed into the per-step total and tiny nonzero costs displayed as `$0.0000`. This PR fixes that display layer so users can tell total input, uncached input, cache read/write, output, and actual chargeable cost apart.

## Validation

- `uv run pytest tests/test_message.py -q`
- `uv run pytest tests/test_cost_display.py tests/test_message.py -q`
- `uv run ruff check gptme/llm/models/resolution.py gptme/logmanager/manager.py gptme/message.py tests/test_message.py`
- `uv run ruff check gptme/util/cost_display.py tests/test_cost_display.py gptme/message.py tests/test_message.py`
- pre-commit hooks via `git commit` / `git commit --amend`

### DIFF
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -1,6 +1,7 @@
 import dataclasses
 import json
 import logging
+import re
 import shutil
 import sys
 import textwrap
@@ -469,22 +470,13 @@ timestamp = "{self.timestamp.isoformat()}"
 
 
 def _strip_think_sig(content: str) -> str:
-    """Strip think blocks and think-sig comments from message content.
+    """Strip think-sig comments from message content.
 
-    These are Anthropic extended thinking artifacts serialized into message
-    content. They are implementation noise the user shouldn't see.
+    Anthropic extended thinking signatures may be serialized into message
+    content as multiline comments. They are implementation noise the user
+    shouldn't see, while the surrounding thinking text can still be useful.
     """
-    import re
-
-    lt = chr(60)  # <
-    gt = chr(62)  # >
-    # Strip <think>...</think> blocks (may span lines)
-    think_block = lt + "think" + gt + r".*?" + lt + "/think" + gt + r"\s*"
-    content = re.sub(think_block, "", content, flags=re.DOTALL)
-    # Strip standalone <!-- think-sig: ... --> comments
-    sig_comment = lt + "!-- think-sig:.*?--" + gt + r"\s*"
-    content = re.sub(sig_comment, "", content)
-    return content
+    return re.sub(r"<!--\s*think-sig:.*?-->\s*", "", content, flags=re.DOTALL)
 
 
 def format_msgs(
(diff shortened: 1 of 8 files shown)


gptme/message.py
```diff
@@ -468,6 +468,25 @@ def cost(self, model: str | None = None, output=False) -> float:
...
+    These are Anthropic extended thinking artifacts serialized into message
+    content. They are implementation noise the user shouldn't see.
+    """
+    import re
+
+    lt = chr(60)  # <
+    gt = chr(62)  # >
+    # Strip <think>...</think> blocks (may span lines)
+    think_block = lt + "think" + gt + r".*?" + lt + "/think" + gt + r"\s*"
+    content = re.sub(think_block, "", content, flags=re.DOTALL)
```
ErikBjare: No i think these should be kept, it can actually be interesting to read. The think-sig stuff within is pure noise though.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Erik's review and the remaining ErikBjare/bob#834 display issue in the latest pushes.

- `234bab1` preserves visible `<think>` content and strips only multiline `think-sig` comments, including standalone signature comments.
- `e2da433` clarifies `/cost` / `/tokens` per-step columns: `TotalIn`, `Uncached`, `Output`, `CacheR`, `CacheW`, `Cost`; renames the aggregate label to `Cache hit rate`; and renders tiny positive costs as `<$0.0001` instead of `$0.0000`.

Local verification: `uv run pytest tests/test_cost_display.py tests/test_message.py -q`, `uv run ruff check gptme/util/cost_display.py tests/test_cost_display.py gptme/message.py tests/test_message.py`, and commit hooks.

TimeToBuildBob:
@greptileai review

PR #2699, opened 2026-06-02
fix(webui): encode chat route conversation ids

TimeToBuildBob:
## Summary
- Add shared route helpers for chat conversation IDs
- Encode chat IDs when navigating so IDs like `demo/conv-*` stay in a single route segment
- Decode route params before passing them into `MainLayout`

## Verification
- `npm test -- --runTestsByPath src/utils/__tests__/routes.test.ts src/utils/__tests__/demoApiClient.test.ts`
- `npm run typecheck`
- pre-commit hooks during commit
- Browser smoke: opened `/?demo=1`, sent a prompt, landed on `/chat/demo%2Fconv-*`, and saw the replayed final assistant response

## Follow-up
- The browser smoke still shows demo mode leaking non-demo fetches to `/api/v2` and `/api/v2/tasks`; that should be the next offline-demo slice.

### DIFF
--- a/webui/src/components/CommandPalette.tsx
+++ b/webui/src/components/CommandPalette.tsx
@@ -28,6 +28,7 @@ import {
   exportConversationAsJSON,
   getExportableMessages,
 } from '@/utils/exportConversation';
+import { chatRoute } from '@/utils/routes';
 import { toast } from 'sonner';
 import { settingsModal$ } from '@/stores/settingsModal';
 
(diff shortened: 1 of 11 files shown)


TimeToBuildBob:
Addressed the remaining gap flagged by Greptile: `taskApi.viewTaskConversation` now uses `chatRoute()` instead of the bare template literal (`/chat/${taskId}`). This completes coverage of all `/chat/` navigation callsites in the codebase.

CI running on the new commit.

TimeToBuildBob:
@greptileai review

PR #2720, opened 2026-06-03
fix(cost): drain stream after tool break to capture message_delta usage

### LINKED ISSUE #2718: fix: per-step breakdown shows 0 output tokens for all tool-call steps with --tool-format tool
## Problem

When running gptme with `--tool-format tool`, the `/tokens` per-step breakdown shows 0 output tokens and $0.0000 cost for every step except the last:

```
Per-Step Breakdown:
  Step   TotalIn  Uncached   Output    CacheR    CacheW       Cost  Model
     1    30,407        10        0         0    30,397    $0.0000  claude-haiku-4-5
     2    35,006         6        0    30,397     4,603    $0.0000  claude-haiku-4-5
     3    35,203         5        0    35,000       198    $0.0000  claude-haiku-4-5
     4    35,415         5        0    35,198       212    $0.0000  claude-haiku-4-5
     5    35,331         6       31    29,322     6,003    $0.0106  claude-haiku-4-5
```

The same run with the default `--tool-format markdown` shows normal output counts for each step. Only the tool-format `tool` path exhibits this.

Reported in ErikBjare/bob#834.

## Root cause hypothesis

With `--tool-format tool`, tool calls are expressed as structured API tool use (not markdown code blocks). When `break_on_tooluse` fires for these native tool calls, the stream is closed before `message_delta` arrives — so output token counts from `message_delta` are never recorded. PR gptme/gptme#2703 added a `message_start` fallback for input/cache tokens, but `message_start` does not carry output token counts. There may also be differences in how the OpenAI-compat path (used for `tool` format) structures usage events compared to the Anthropic native path.

## Investigation

1. Confirm whether the Anthropic API's `message_start` event carries any output token preview for tool-call responses.
2. Check whether the OpenAI-compat `/v1/chat/completions` path delivers usage differently from the Anthropic native path for tool-call-only turns.
3. Determine if 0 output tokens is literally correct (tool-call-only responses may have 0 text output tokens, with the "cost" attributed to the completion via `message_delta` after the final text turn).

## Goal

Per-step breakdown should either show accurate output token counts for tool-call steps, or clearly indicate that these steps are tool-call-only and the cost flows to the following text step — whichever is more accurate.

## References

- ErikBjare/bob#834
- gptme/gptme#2703 (message_start partial fallback for break_on_tooluse)

TimeToBuildBob:
When `break_on_tooluse` fires before `message_delta` arrives with `--tool-format tool`, output token counts from `message_delta` are never recorded. This causes the per-step breakdown to show 0 output tokens and $0.0000 cost for all tool-call steps, since only the final text step receives `message_delta`.

## Root cause

- `message_start` provides input/cache token counts (captured as fallback in #2703)
- `message_delta` provides output token counts and the final cumulative usage
- `break_on_tooluse` closes the stream before `message_delta` arrives
- The `_StreamWithMetadata` fallback (`message_start` data) has no `output_tokens`
- The openai-compat path (used for `tool` format) has the same shape: `response.completed` fires after the break event

## Fix

Added `_drain_toolbreak_stream()` that, after the tool-use break fires, consumes the remaining generator events to process pending events (`message_delta` / `response.completed`) and capture their metadata. The drained content is discarded — only the side-effect of metadata capture matters.

Closes #2718

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -642,7 +642,11 @@ def main(
         interactive = False
 
     # init logging
-    init_logging(verbose)
+    # Route log output through stdout (via shared Rich Console) when in a TTY
+    # so that logging and streaming assistant output are serialized through the
+    # same Console, preventing stderr/stdout interleave mid-stream.
+    # In non-interactive/pipe modes, keep traditional stderr routing.
+    init_logging(verbose, stderr=not sys.stdout.isatty())
 
     if not interactive:
         no_confirm = True
--- a/gptme/init.py
+++ b/gptme/init.py
@@ -246,9 +246,17 @@ def _maybe_authenticate_gptme_interactively(
 
 
 def init_logging(verbose, *, stderr: bool = True):
-    handler = RichHandler(
-        console=Console(stderr=stderr, log_path=False)
-    )  # show_time=False
+    if not stderr:
+        # Use the shared Rich Console (same instance rprint uses) so log
+        # output is serialized with streaming assistant output through a
+        # single Console, preventing stderr/stdout interleave mid-stream.
+        from rich import get_console as _get_console
+
+        handler = RichHandler(console=_get_console())
+    else:
+        handler = RichHandler(
+            console=Console(stderr=True, log_path=False)
+        )  # show_time=False
     logging.basicConfig(
         level=logging.DEBUG if verbose else logging.INFO,
         format="%(message)s",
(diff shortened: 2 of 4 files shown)


TimeToBuildBob (commented):
Good catch @greptile-apps — fixed in 7b9bec88f. Replaced the `for _ in stream.gen:` with `while True: next(stream.gen)` so `StopIteration` propagates to the except block and `e.value` (carrying `message_delta` output tokens) is actually captured. The fix now achieves its stated goal.

PR #2721, opened 2026-06-03
feat(webui): suppress live fetches in demo mode

TimeToBuildBob:
## Summary

Suppresses live API fetches in demo mode (`?demo=1`) so the shared demo URL stays truly offline — no backend noise in the browser console.

## What

- **`useModels`**: early-return empty models list in demo mode (skip `/api/v2/models`)
- **`useTasksQuery`**: skip the query entirely in demo mode (no `/api/v2/tasks` fetch)
- **`taskApi.listTasks`**: short-circuit to empty list in demo mode
- **`providerStatus.fetchProviderConfigured`**: return `true` in demo mode (mock healthy provider)
- **`modelsError`**: add `buildDemoModelsFetchError` for demo-specific error messages

## Why

After \#2699 (route encoding fix), opening `/?demo=1` and sending a prompt works end-to-end with fixture-backed chat/tool replay, but the browser console still shows three live fetch errors to `127.0.0.1:5700`. These are noise during demos and trade-show presentations.

## Verification

- `npm test -- --runTestsByPath src/utils/__tests__/demoApiClient.test.ts` (14/14)
- `npm run typecheck` — clean
- `npm test` — 233/233 pass (21 component-test suites fail on pre-existing `ansi-regex` ESM transform issue unrelated to this PR)

## Related

- \#2695 — Slice 3: demo tool-call replay (merged)
- \#2699 — Slice 4: route encoding fix (merged)
- Task: gptme-cloud-offline-demo-mode (session b817)

### DIFF
--- a/webui/src/components/__tests__/SetupWizard.test.tsx
+++ b/webui/src/components/__tests__/SetupWizard.test.tsx
@@ -64,6 +64,7 @@ jest.mock('@/hooks/useTauriServerStatus', () => ({
 
 jest.mock('@/utils/connectionConfig', () => ({
   processConnectionFromHash: (...args: unknown[]) => mockProcessConnectionFromHash(...args),
+  isDemoMode: jest.fn(() => false),
 }));
 
 jest.mock('@legendapp/state/react', () => ({
--- a/webui/src/hooks/useModels.ts
+++ b/webui/src/hooks/useModels.ts
@@ -2,6 +2,7 @@ import { useState, useEffect } from 'react';
 import { use$ } from '@legendapp/state/react';
 import { useApi } from '@/contexts/ApiContext';
 import { buildModelsFetchError } from '@/utils/modelsError';
+import { isDemoMode } from '@/utils/connectionConfig';
 
 export interface ModelInfo {
   id: string;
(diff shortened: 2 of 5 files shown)


TimeToBuildBob:
Thanks for the review @greptile-apps — all three points addressed in f68658c:

1. **P1 — `fetchProviderConfigured` returning `false`**: Fixed to return `true` in demo mode. This was the one-character mistake you identified — the WelcomeView "Provider setup required" banner should no longer appear in demo sessions.

2. **P2 — `placeholderData: []` instability**: Changed to `placeholderData: () => [] as Task[]` so TanStack Query gets a stable reference instead of a new array on every render.

3. **P2 — dead-code `buildModelsFetchError` guard**: Acknowledged — the guard is unreachable now that `useModels` skips the fetch entirely in demo mode, but I kept it as defense-in-depth. If `buildModelsFetchError` is ever called from a different code path in the future, the demo guard is already there.

Also added `isDemoMode: jest.fn(() => false)` to the `@/utils/connectionConfig` mock in `SetupWizard.test.tsx` — the test suite was failing because the mock didn't export `isDemoMode`, so `fetchProviderConfigured` threw a runtime error instead of short-circuiting.

Re-triggered Greptile on the new commit.

PR #2727, opened 2026-06-03
fix(logmanager): stop auto-pruning thinking blocks

### LINKED ISSUE #2724: perf: investigate poor prompt cache hit rates
## Problem

Observing low prompt cache hit rates in interactive sessions — around 32-34% cache usage with patterns like:

```
uncached: 10t, create: 60475t, read: 31073t
```

This means we're paying full price for ~67% of tokens that should ideally be cached.

## Potential causes

- System prompt or context assembly ordering that invalidates the cache prefix
- Tool results or injected context (lessons, workspace info) changing between turns in ways that break the prefix match
- Model-specific caching behavior not being accounted for (different providers have different prefix caching rules)

## Investigation needed

- [ ] Check if the system prompt + early context is stable across turns (prefix caching requires identical prefixes)
- [ ] Look at whether lesson injection, workspace status, or other dynamic context inserted early in the message list is invalidating caches
- [ ] Compare cache hit rates across different providers (Anthropic vs OpenRouter vs OpenAI)
- [ ] Check if conversation compaction/autocompact affects cache performance

## For Bob

Bob should check his own gptme log output during autonomous runs to see if he also has poor cache hits. Look for cache-related log lines and compare `create` vs `read` token counts across sessions. If the pattern is systemic, it affects inference costs significantly for all autonomous agents.

Related: prompt caching docs for [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) — the cached prefix must be an exact match of the first N tokens.

TimeToBuildBob:
## Root cause

Closes #2724.

The cache regression was caused by automatic `ephemeral_ttl=2` tagging for assistant messages containing `<think>` / `<thinking>` blocks.

Once those messages expired, `prune_ephemeral_messages()` removed the whole assistant turn and `_merge_consecutive_messages()` merged the surrounding user turns. That changed the exact conversation prefix every few turns, so Anthropic prompt caching fell back to the stable system prompt only. This matches the issue data where `CacheR` grows at first, then drops to a constant ~31k tokens.

## Fix

Remove the automatic reasoning-as-ephemeral policy.

Explicit `ephemeral_ttl` still works unchanged for callers that deliberately mark volatile context, such as screenshot/computer-use artifacts. Plain assistant reasoning content is no longer pruned just because it contains `<think>` or `<thinking>`.

This is safer for Anthropic signed thinking too: gptme serializes Anthropic thinking blocks as `<think>...<!-- think-sig: ... -->...</think>` so they can be reconstructed into structured `thinking` blocks on later tool-use requests. Automatically stripping or dropping those blocks is too broad as a default.

## Why this explains the observed data

```txt
Step 3: CacheR=38,446  (cache working, conversation accumulating)
Step 4: CacheR=31,073  (drop to system-only after first TTL prune)
Steps 5-58: CacheR=31,073 constant
```

At step 4, an early assistant message with auto-added `ephemeral_ttl=2` expired. Removing it merged adjacent user messages and invalidated every cached prefix after that point. Later prune cycles repeated the invalidation, preventing cache reuse beyond the system prompt.

With automatic reasoning TTL removed, conversation structure remains stable and cache reads should grow with the conversation instead of resetting to system-only.

## Tests

Updated `tests/test_ephemeral.py`:
- explicit `ephemeral_ttl` pruning still drops expired messages and merges adjacent same-role turns
- unmarked `<think>` assistant messages are not pruned by default

Local verification:

```bash
uv run --with pytest --with pytest-asyncio --with pytest-timeout --with pytest-retry --with pytest-mock --with requests python -m pytest tests/test_ephemeral.py tests/test_llm_utils.py::test_handle_tools_thinking_with_signature tests/test_llm_utils.py::test_extract_thinking_content_with_signature -q
# 22 passed

uv run ruff check gptme/logmanager/manager.py tests/test_ephemeral.py
# All checks passed!
```


TimeToBuildBob:
## Review Feedback Addressed

Two commits pushed to address the CI typecheck failure and Greptile review feedback:

### commit `ea54be285` — fix(llm): add arg-type ignores for stream() args

The typecheck failure in CI was caused by the anthropic SDK 0.105 bump (from PR #2710). The newer SDK types `NOT_GIVEN` as `NotGiven` instead of being compatible with `Omit`/`NotGiven`. Added `# type: ignore[arg-type]` to the three failing `stream()` arguments:

- `top_p` — `float | NotGiven` vs `float | Omit`
- `tools` — `list[ToolParam] | NotGiven` vs `Iterable[...] | Omit`
- `output_config_kwargs` — `_OutputConfig` vs `OutputConfigParam | Omit`

Same pattern already used for `temperature` and the `stream()` call itself (`# type: ignore[call-arg, misc]`). No runtime change.

### commit `bc5b90e0b` — fix(logmanager): address Greptile feedback on `_strip_thinking_blocks`

- Removed redundant `import re` inside the function (already imported at module level)
- Added `re.IGNORECASE` flag to match `_THINKING_PATTERN` behavior

Both are review-feedback-only changes; the core strip-not-drop logic is unchanged.

### Anthropic SDK Bump Context

The `anthropic ^0.47 → ^0.105` bump in `pyproject.toml` was inherited from PR #2710 (merged into master). This PR branched off master afterward and picked it up automatically.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Ready for merge — CI green (all 15 checks pass), Greptile 5/5. Review feedback was addressed in the two commits above (`ea54be285`, `bc5b90e0b`). No further changes needed.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Removed the accidental `gptme/llm/llm_anthropic.py` diff introduced during the rebase. Current PR diff is back to the intended cache-pruning scope only:

- `gptme/logmanager/manager.py`
- `tests/test_ephemeral.py`

Local verification on head `2acbeacc8`:

```bash
uv run --with pytest --with pytest-asyncio --with pytest-timeout --with pytest-retry --with pytest-mock --with requests python -m pytest tests/test_ephemeral.py -q
# 23 passed
```

CI is rerunning now. @greptileai review

ErikBjare:
Is the thinking-stripping really a good idea? Seems weird as a default. Ephemeral messages were meant for things like reading screenshots for computer use loops, not for reasoning. What happens with think-sig etc for Anthropic?

TimeToBuildBob:
You're right to question it. I would not defend "strip thinking from all ephemeral messages" as a generic default.

What the current branch does:

- `LogManager.append()` auto-tags assistant messages containing `<think>/<thinking>` with `ephemeral_ttl=2`. That's the controversial part.
- #2727 changes expiry only for expired assistant messages that still contain thinking tags: remove the full `<think>...</think>` block and keep the answer text. Non-thinking ephemeral messages still drop as before.
- For Anthropic `think-sig`: the signature is embedded inside the `<think>` block. Before expiry, `_extract_thinking_content()` parses it back into Anthropic `thinking` blocks with signatures. After expiry, #2727 strips the whole block including `think-sig`, so the old turn is replayed as plain assistant text. That avoids orphaned or modified signatures; it means we intentionally stop replaying that old thinking block.

Anthropic's docs line up with that distinction: tool-continuation thinking blocks must be passed back complete/unmodified, but prior thinking blocks can otherwise be omitted. If we send a thinking block, it has to remain intact; if we omit it, there is no signature to validate. Docs: https://platform.claude.com/docs/en/build-with-claude/extended-thinking#preserving-thinking-blocks

My take: the design smell is `_auto_tag_ephemeral()`, not the strip operation itself. `ephemeral_ttl` originally meant "remove a transient message from context" for high-churn context like screenshots/tool loops. Auto-tagging assistant reasoning turns makes it also mean "retire old thinking but keep the answer". Those should probably be separate semantics:

1. Manual/generic ephemeral messages: keep drop-whole-message behavior.
2. Assistant thinking: use a narrower thinking-part pruning path that strips the think block + sig and preserves answer text.
3. Tool-use continuation: do not strip the current assistant/tool chain; Anthropic requires that chain unmodified.

So I no longer think this should merge as "generic ephemeral default changed". Either narrow #2727 into explicit thinking-part pruning with docs/tests, or revert the auto-tag default and accept the prompt-cache/token cost until we have a cleaner policy.

TimeToBuildBob:
You're right. The first fix was aimed at the cache symptom but kept the wrong policy.

I revised this in `2eaacb429`: PR #2727 now removes the automatic `ephemeral_ttl` tagging for assistant messages containing `<think>` / `<thinking>`. Explicit `ephemeral_ttl` pruning still works for deliberately volatile artifacts, but reasoning is no longer treated as ephemeral by default.

The `think-sig` point is the killer argument. gptme serializes Anthropic signed thinking as `<think>...<!-- think-sig: ... -->...</think>` and `_handle_tools()` reconstructs that into structured `thinking` blocks later. Anthropic's docs say signed thinking blocks should be passed back unchanged during tool use, and newer Opus/Sonnet models preserve prior thinking by default, so stripping/removing signed thinking in generic logmanager code is too broad.

Local checks on the revised head:

```bash
uv run --with pytest --with pytest-asyncio --with pytest-timeout --with pytest-retry --with pytest-mock --with requests python -m pytest tests/test_ephemeral.py tests/test_llm_utils.py::test_handle_tools_thinking_with_signature tests/test_llm_utils.py::test_extract_thinking_content_with_signature -q
# 22 passed

uv run ruff check gptme/logmanager/manager.py tests/test_ephemeral.py
# All checks passed!
```

CI is running on `2eaacb429` now.

Refs: https://platform.claude.com/docs/en/build-with-claude/extended-thinking#preserving-thinking-blocks and https://platform.claude.com/docs/en/build-with-claude/prompt-caching#caching-with-thinking-blocks

TimeToBuildBob:
@greptileai review

Issue #2725, opened 2026-06-03
bug: LLM produces thinking but no tool output, causing silent re-prompts in non-interactive mode


ErikBjare:
## Problem

Sometimes the LLM generates a response that contains only "thinking" (reasoning/planning text) but no tool use blocks. In `--non-interactive` mode, the `complete` hook detects "no tool was executed" and re-prompts the assistant to continue — which can cause:

1. **Wasted inference**: the thinking tokens are paid for but produce no action
2. **Silent retry loops**: the re-prompt may trigger the same pattern again
3. **Hard to diagnose**: there's no obvious WARNING log when this happens, so it's easy to miss in production autonomous runs

## Observed behavior

In an interactive session, the assistant produced several messages where tool blocks were present in the response formatting but never actually executed (the assistant appeared to be "thinking" about what to do without committing to action).

## Investigation needed

- [ ] Grep gptme logs for the re-prompt pattern in `--non-interactive` mode — how often does the "no tool executed, continuing" path trigger?
- [ ] Check if this correlates with specific models (e.g. reasoning models like o1/o3 or extended thinking in Claude)
- [ ] Consider adding a WARNING log when the assistant produces a response with no executable tool use in non-interactive mode, so it's visible in log output
- [ ] Track the count of consecutive no-tool-use responses and circuit-break if it loops

## For Bob

Bob should:
1. **Grep his own session logs** for cases where `--non-interactive` re-prompted due to no tool use — this would show up in the chat loop logic
2. **Add WARNING log grepping to his health checks/vitals** — a general scan for WARNING-level log lines across recent sessions would surface this and other silent issues
3. **Check trajectories** for patterns where the assistant produces long reasoning blocks but no actionable output, especially in autonomous runs where this wastes budget silently

## Suggested improvements

1. Add `logger.warning("Assistant produced no tool use in non-interactive mode, re-prompting")` at the re-prompt point
2. Consider a metric/counter for no-tool-use responses per session
3. Add a vitals/health-check entry that scans recent logs for WARNING lines as a general early-warning system

ErikBjare:
## Clarification: interactive mode is actually worse

The non-interactive hook at least auto-reprompts, so sessions don't stall. But in **interactive mode** (even with `-y` auto-approvals), this is more painful:

1. The assistant produces a long thinking/planning response with no tool calls
2. Control returns to the user prompt
3. The user has to manually type "continue, you didn't call tools" or similar nudge
4. This can repeat multiple times per task

[...]

PR #2736, opened 2026-06-04
fix(cli): promote no-tool auto-reply logging from info to warning level

TimeToBuildBob:
Promotes the three auto-reply log points in `auto_reply_hook` from `logger.info` to `logger.warning` so they're visible in health checks, vitals, and log-scraping tools.

This is the quick fix for the actionable item from Erik in gptme/gptme#2725: logging the no-tool-use re-prompt at WARNING level so it can be detected by Bob's vitals/health checks automatically, without per-session log archaeology.

The three affected lines:
1. The first auto-reply (no incomplete todos) — most common case in interactive mode
2. The first auto-reply with incomplete todos — common in autonomous mode  
3. The exit-after-2-confirmations message — signals a wasted autonomous session

All 53 existing tests pass unchanged.

### DIFF
--- a/gptme/tools/complete.py
+++ b/gptme/tools/complete.py
@@ -144,14 +144,14 @@ def auto_reply_hook(
 
     # Exit after 2 consecutive auto-replies without tools
     if auto_reply_count >= 2:
-        logger.info("Autonomous mode: No tools used after 2 confirmations. Exiting.")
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
👋 @ErikBjare this is ready for review. CI all green. Greptile, please review this PR.

PR #2747, opened 2026-06-05
fix(reduce): drop orphaned tool results in limit_log when tool-use is dropped

TimeToBuildBob:
## Problem

`limit_log` cuts messages newest-to-oldest until context is full. When a `system` message (tool result) fits but its preceding `assistant` message (tool use) does not, the tool result becomes an orphan — it appears in the log without the tool call that generated it. This produces an incoherent conversation state that can confuse the model or cause downstream parsing issues.

The companion function `reduce_log` already guards against this via `message_contains_tool_use`, but `limit_log` (the final hard cut in `prepare_messages`) had no such guard.

## Fix

After building the `result` list, a post-processing pass drops any non-initial `system` message whose immediate predecessor in the original log was dropped. Uses object identity (`id()`) to match messages since `limit_log` picks objects directly from the input list without copying.

```python
def _is_orphaned(msg: Message) -> bool:
    if msg.role != "system" or id(msg) in initial_id_set:
        return False
    idx = log_by_id.get(id(msg))
    if idx is None or idx == 0:
        return False
    return id(log[idx - 1]) not in result_id_set
```

## Test

Added `test_limit_log_tool_pair_atomicity` in `tests/test_reduce.py`:
- Sets up a 3-message log: system prompt (2 tok) + assistant tool-use (13 tok) + system tool-result (1 tok)
- Uses a tiny 10-token context model so the tool-use is dropped but the tool-result would otherwise fit
- Asserts the orphaned tool-result is dropped from the output

All 13 non-slow reduce tests pass.

### DIFF
--- a/gptme/util/reduce.py
+++ b/gptme/util/reduce.py
@@ -292,6 +292,16 @@ def limit_log(log: list[Message]) -> list[Message]:
         idx = log_by_id.get(id(msg))
         if idx is None or idx == 0:
             return False
-        return id(log[idx - 1]) not in result_id_set
+        # Walk backward through the original log to find the nearest
+        # non-system anchor message. System messages are tool results;
+        # their anchor is the assistant (or user) message that produced
+        # them. If that anchor was dropped by the context limit, this
+        # result is orphaned.
+        for j in range(idx - 1, -1, -1):
+            if log[j].role != "system":
+                return id(log[j]) not in result_id_set
+        # No non-system predecessor found — shouldn't happen for a
+        # non-initial system message, but treat as not orphaned.
+        return False
 
     return [m for m in result if not _is_orphaned(m)]
--- a/gptme/util/reduce.py
+++ b/gptme/util/reduce.py
@@ -275,4 +275,23 @@ def limit_log(log: list[Message]) -> list[Message]:
         # skip the last message
         msgs.pop()
 
-    return initial_system_msgs + list(reversed(msgs))
+    result = initial_system_msgs + list(reversed(msgs))
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both Greptile findings in [4a970cf](https://github.com/gptme/gptme/pull/2747/commits/4a970cf7ce89800c4be3dbe04808f7e1585b1079):

### 1. Multi-tool cascading orphans (reduce.py)

Replaced the immediate-predecessor check with a backward walk to the nearest non-system ancestor. When `break_on_tooluse=False` produces multiple consecutive system messages from a single assistant turn, all of them are now caught — not just the first one whose immediate predecessor is the dropped assistant message.

```python
# Before: only first orphan caught (immediate predecessor)
return id(log[idx - 1]) not in result_id_set

# After: all orphans caught (walk to nearest non-system ancestor)
for j in range(idx - 1, -1, -1):
    if log[j].role != "system":
        return id(log[j]) not in result_id_set
```

### 2. ContextVar teardown (test_reduce.py)

Both `test_limit_log_tool_pair_atomicity` and the new `test_limit_log_cascading_orphans` now save/restore `_default_model_var` in a try/finally block, preventing state contamination for subsequent tests in the same process.

### 3. New test: cascading orphans

Added `test_limit_log_cascading_orphans` — 3 consecutive system messages after a single assistant message, with a context window that drops only the assistant. Verifies that **all three** tool results are correctly orphaned rather than just the first.

TimeToBuildBob:
@greptileai review

PR #2750, opened 2026-06-05
fix(eval): normalize model key in pass-rate gate to handle router prefixes

TimeToBuildBob:
## Problem

The natural pass-rate gate lookup (`pass_rate_gate.get_gate_recommendation`) does an exact model-key match. When gate data was collected via the direct Anthropic API (key: `anthropic/claude-haiku-4-5`) but the eval runner uses OpenRouter (key: `openrouter/anthropic/claude-haiku-4-5`), the lookup silently falls back to `"default"` — bypassing all inject/suppress decisions.

This caused a real incident: `bob-eval-behavioral.service` was configured with `GPTME_EVAL_PASS_RATE_GATE_FILE` but the gate was never applied to haiku evals because of this prefix mismatch. Sonnet evals were unaffected (both gate data and runs used the `openrouter/` prefix consistently).

## Fix

Add `_normalize_model_key()` that strips known router prefixes (`openrouter/`) on a fallback lookup when the exact key is absent. Exact match still takes priority, so gate files that use the full `openrouter/`-prefixed key continue to work correctly.

## Tests

Added 3 new tests to `test_eval_pass_rate_gate.py`:
- `test_get_gate_recommendation_openrouter_prefix_fallback` — verifies inject/suppress/default resolve correctly when key has `openrouter/` prefix
- `test_get_gate_recommendation_openrouter_prefix_exact_match_wins` — verifies exact match takes priority over fallback
- `test_apply_gate_openrouter_prefix_fallback` — end-to-end via `apply_gate()`

All 19 tests pass.

### DIFF
--- a/gptme/eval/pass_rate_gate.py
+++ b/gptme/eval/pass_rate_gate.py
@@ -81,7 +81,7 @@ def _normalize_model_key(model: str, lookup: dict) -> str:
     Tries an exact match first, then retries after stripping known router
     prefixes (e.g. ``openrouter/anthropic/X`` → ``anthropic/X``). This
     handles the case where gate data was collected via the direct API but
-    evals are executed via an OpenRouter backend, or vice versa.
+    evals are executed via an OpenRouter backend.
     """
     if model in lookup:
         return model
--- a/gptme/eval/pass_rate_gate.py
+++ b/gptme/eval/pass_rate_gate.py
@@ -72,6 +72,32 @@ def load_pass_rate_data(path: Path | str | None = None) -> dict | None:
     return data
 
 
+_ROUTER_PREFIXES = ("openrouter/",)
+
+
+def _normalize_model_key(model: str, lookup: dict) -> str:
+    """Return the best matching key for *model* in *lookup*.
+
+    Tries an exact match first, then retries after stripping known router
+    prefixes (e.g. ``openrouter/anthropic/X`` → ``anthropic/X``). This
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the Greptile finding: removed the "or vice versa" overpromise from the `_normalize_model_key` docstring (89ab4a4). The code only strips prefixes from the model key; the inverse direction isn't implemented.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Review Thread Resolution

Greptile's P2 finding (docstring overpromises "or vice versa") addressed in 89ab4a4 — updated the docstring to accurately describe only what the code implements (stripping router prefixes from the model key for fallback lookup). Thread resolved.

Greptile re-review triggered. CI is fully green, Eval Quality Gate passed. Waiting for Greptile re-review confirmation before merge.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Added 2 more tests (368d1ff) to cover missing code paths flagged by Codecov:
- `test_get_gate_recommendation_openrouter_prefix_no_match`: exercises the branch where model has a router prefix but the stripped key is also absent from the lookup (falls through to return original model → "default")
- `test_get_gate_recommendation_empty_lookup`: exercises behaviour with an empty lookup dict

All 21 tests pass in CI. The Codecov 15% patch coverage warning should improve.

Issue #2717, opened 2026-06-03
feat: smoother, less-bursty streaming output in terminal and webui


TimeToBuildBob:
## Problem

Streaming output in both the terminal and the webui is noticeably bursty. Characters arrive in uneven bursts rather than at a steady pace. In the webui there's also an intermittent multi-hundred-millisecond stall right before the final character of a message (often the trailing `.`).

Reported in ErikBjare/bob#834.

## Observed behaviour

- Terminal: chunks arrive in visible batches, then pause; especially noticeable on longer turns
- Webui: same burst pattern, plus a distinct stall at the very end of each assistant message before the final punctuation renders

## Investigation starting points

Terminal side:
- Check whether `_reply_stream` is doing any synchronous work between yielding chunks (e.g. thinking-block detection, Rich rendering calls)
- Rich's `Console` write path may be buffering—verify `Console(highlight=False)` and no extra flush calls are needed
- The per-character path (`are_thinking`) may be slower than the per-chunk path; consider whether thinking-block content needs per-character rendering at all

Webui side:
- Check the SSE/WebSocket flush cadence in `gptme/server/`
- Look at whether the frontend accumulates text before inserting into the DOM (debounce, RAF loop, or similar)
- The trailing stall may be the final `message_stop` event arriving separately from the last text delta—profile the gap between the last `content_block_delta` and `message_stop`

## Goal

Streaming should feel fluid. For the terminal, character-by-character arrival should be imperceptible to humans (each chunk renders as it arrives). For the webui, the stall at message end should be eliminated or reduced to <50 ms.

## References

- ErikBjare/bob#834 (last open note: "open an issue for making the streaming look smoother")

TimeToBuildBob:
## Triage: investigation only (no immediate fix)

This spans two repos (gptme core + gptme-webui) and needs profiling before any code changes.

### Terminal side (gptme core)

The streaming hot path is `_reply_stream()` in `gptme/llm/__init__.py:535` — a
character-by-character loop over streaming chunks. Each character goes through:

1. **Thinking-tag detection**: string comparisons per char for `<think>` / `</think>` / think-sig boundaries
2. **Per-char display**: `rprint()` via Rich Console for non-thinking content
3. **`sys.stdout.flush()` every char** (line 667): needed for responsive display but may contribute to burstiness
4. **on_token callback**: per-line buffering with thinking-tag filtering

[...]

PR #2753, opened 2026-06-05
perf(llm): batch terminal output per chunk, not per char

TimeToBuildBob:
## Summary

- Fixes the terminal side of gptme/gptme#2717: bursty streaming output in the terminal
- Introduces `normal_display_buffer` — a list that accumulates non-thinking chars within a chunk, flushed at each chunk boundary or before each newline
- Adds a local generator `_chars_with_chunk_end(stream)` that yields `(char, is_last_in_chunk)` pairs, letting the loop know when to flush without restructuring the body
- Reduces `rprint()` and `sys.stdout.flush()` calls from **O(chars)** to **O(chunks)** — eliminating per-char syscall overhead that caused bursty rendering

## Root cause

`_reply_stream()` iterated a flattened `for char in (char for chunk in stream for char in chunk)` generator, calling `rprint(char, end="")` and `sys.stdout.flush()` for each individual character. Each LLM chunk (typically 4-10 chars) caused N separate Rich render passes and N `write()`+`flush()` syscalls to the terminal. The burst-then-pause pattern users see matches exactly this profile: fast syscall rain during a chunk arrival, then silence until the next chunk.

## Approach

The thinking-tag detection logic already requires character-level iteration (it needs to see `<think>` / `</think>` lines character-by-character), so we can't simply switch to chunk-level iteration. Instead:

1. A `_chars_with_chunk_end()` generator wraps the stream and tags the last char in each chunk with `True`.
2. Normal (non-thinking, non-newline) chars accumulate in `normal_display_buffer` instead of being printed immediately.
3. At each newline, the buffer flushes before the `\n` so line content precedes the line ending.
4. At each chunk end (`_is_chunk_end=True`), any remaining buffered chars are flushed and `sys.stdout.flush()` is called once.

Thinking-block buffering (`think_display_buffer`) and the `on_token` line-buffer callback path are unchanged.

## Test plan

- [x] Module imports clean: `python -c "from gptme.llm import _reply_stream; print('OK')"`
- [x] 8 targeted unit tests covering the buffering logic and `_chars_with_chunk_end` generator
- [x] 739 existing tests pass (178 core + 561 broader non-API); 1 failure (`test_include_paths_directory`) is pre-existing on master
- [ ] Manual verification: run gptme interactively and observe smoother streaming output

### DIFF
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -760,6 +760,11 @@ def _reply_stream(
             line_buffer.clear()
 
     except KeyboardInterrupt:
+        # Flush any chars buffered since the last chunk boundary so the terminal
+        # shows everything received before the interrupt.
+        if not json_mode and normal_display_buffer:
+            rprint("".join(normal_display_buffer), end="")
+            normal_display_buffer.clear()
         # Flush partial line before the interrupt suffix so callers see the
         # content that was streamed up to the interrupt point.
         if on_token and line_buffer:
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -564,6 +564,11 @@ def _reply_stream(
     # already-wrapped rows visible.  By buffering and emitting at the newline we
     # can simply discard think-sig lines without any terminal-clear gymnastics.
     think_display_buffer: list[str] = []
+    # Buffer for normal (non-thinking) chars accumulated within a stream chunk.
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Fixed the Greptile P2 finding: added a `normal_display_buffer` flush in the `KeyboardInterrupt` handler (commit 2e626f2a8), matching the pattern already used for the normal/tool-break exit paths.

PR #2762, opened 2026-06-07
fix(server): validate edited message file paths

ErikBjare:
## Summary

- Add shared server-side validation for message file references.
- Apply it to edited message files so local paths must stay inside the conversation workspace.
- Preserve URI attachments and normalize accepted relative paths through the workspace.

## Root Cause

`POST /api/v2/conversations/<id>` validated local `files` entries against the conversation workspace, but `PATCH /api/v2/conversations/<id>/messages/<index>` parsed edited `files` with `parse_file_reference()` directly. That allowed absolute paths like `/etc/passwd` and traversal such as `../outside.txt` to be stored in conversation state.

## Validation

- `uv run --extra server --with requests --with pytest --with pytest-timeout python3 -m pytest tests/test_server_v2.py -q`
- `uv run --extra server --with requests --with pytest --with pytest-timeout python3 -m pytest tests/test_server_bad_input.py -q`
- `uv run --with ruff ruff check gptme/server/api_v2.py tests/test_server_v2.py`
- `git diff --check -- gptme/server/api_v2.py tests/test_server_v2.py`

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -53,7 +53,7 @@ from ..logmanager import Log, LogManager, get_user_conversations
 from ..message import Message
 from ..tools import get_toolchain, get_tools, init_tools
 from ..util.content import is_message_command
-from ..util.uri import parse_file_reference
+from ..util.uri import FilePath, is_uri, parse_file_reference
 from .api_v2_agents import agents_api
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #2766, opened 2026-06-07
fix(util): reject empty batch model override

TimeToBuildBob:
## What

Reject explicit empty `gptme-util batch --model` values at the Click boundary before spawning child `gptme` sessions.

## Why

Dogfooding the fresh batch CLI surface found that `--model ''` was treated as no override because `_run_one_prompt()` checked `if model:`. That silently fell back to the configured default model and could start real child sessions instead of reporting a usage error.

## Validation

- `uv run --with pytest --with requests pytest tests/test_util_cli_batch.py -q` (`13 passed`)
- `uv run --with ruff ruff check gptme/cli/cmd_batch.py tests/test_util_cli_batch.py`
- `uv run --with ruff ruff format --check gptme/cli/cmd_batch.py tests/test_util_cli_batch.py`
- `git diff --check -- gptme/cli/cmd_batch.py tests/test_util_cli_batch.py`
- Replayed `gptme-util batch --model ''` and whitespace-only model repros; both now exit with `Invalid value for '--model': Model name cannot be empty.`

### DIFF
--- a/gptme/cli/cmd_batch.py
+++ b/gptme/cli/cmd_batch.py
@@ -39,6 +39,15 @@ def _usage_tokens(metadata: dict[str, Any]) -> int:
     return total
 
 
+def _validate_model_param(
+    ctx: click.Context, param: click.Parameter, value: str | None
+) -> str | None:
+    """Reject empty --model before spawning child gptme sessions."""
+    if value is not None and not value.strip():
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

Issue #1591, opened 2026-03-04
feat(tauri): add E2E tests for desktop app


TimeToBuildBob:
## Context

Migrated from gptme/gptme-tauri#8 DOD — the tauri desktop app (now under `tauri/` in the monorepo per gptme/gptme#1588) needs automated E2E tests.

## What needs to be done

- Set up Tauri's test driver (uses WebDriver protocol)
- Test basic flows:
  - App launches and connects to local gptme-server
  - Server lifecycle: start/stop/status IPC commands
  - Chat interaction (send message, receive response)
  - Cloud preset switching (local ↔ managed service)
- Add to CI workflow (`.github/workflows/tauri.yml`)

## References

- Tauri testing docs: https://v2.tauri.app/develop/tests/
- gptme/gptme-tauri#8 (original DOD)
- gptme/gptme#1590 (codesigning — related)

JiwaniZakir:
The main complexity here will be the WebDriver setup for Tauri v2, which requires `tauri-driver` (a wrapper around `WebKitWebDriver`/`WebdriverAgent`) and a running `gptme-server` instance as a prerequisite — so the CI workflow will need to handle server lifecycle ordering carefully. The server IPC tests (`start/stop/status`) are worth isolating early since they underpin the chat interaction tests; if the IPC layer is flaky, downstream chat tests will produce misleading failures. The `tauri/` directory structure from #1588 means the CI job will likely need `working-directory: tauri` and separate dependency caching from the main Python package. Cross-platform WebDriver availability (Linux requires `WebKitWebDriver`, macOS requires `WebdriverAgent`) may also constrain which CI runners can realistically run the full suite without significant setup overhead.

TimeToBuildBob:
Thanks for the detailed breakdown! The points about cross-platform WebDriver constraints and CI ordering are exactly the kind of complexity that needs to be planned upfront.

Key notes from your comment that I'll incorporate into the implementation:
- `tauri-driver` wraps `WebKitWebDriver` (Linux) / `WebdriverAgent` (macOS) — different runner setup needed per platform
- Server IPC tests should be isolated first since chat tests depend on them
- CI job needs `working-directory: tauri` and separate dependency caching
- May need to start with Linux-only CI initially given macOS `WebdriverAgent` setup overhead

This will inform the CI workflow design when we get to implementation.

TimeToBuildBob:
Opened a partial step for this lane: draft PR #2456.

What it adds:
[...]

PR #2772, opened 2026-06-08
test(tauri): cover IPC server lifecycle commands

TimeToBuildBob:
## Summary
- add Tauri mock-runtime IPC helpers for the server lifecycle commands
- cover `get_server_status`, duplicate `start_server`, and `stop_server` state clearing
- seed tests with a real shell-managed child process instead of the packaged sidecar

## Testing
- `cargo fmt --manifest-path tauri/src-tauri/Cargo.toml --check`
- `cargo clippy --manifest-path tauri/src-tauri/Cargo.toml -- -D warnings` *(blocked locally: missing `gdk-3.0.pc` / GTK dev libs in Bob's container)*

Refs #1591

### DIFF
--- a/tauri/src-tauri/src/lib.rs
+++ b/tauri/src-tauri/src/lib.rs
@@ -769,14 +769,17 @@ mod tests {
 
     #[cfg(desktop)]
     fn build_test_app() -> tauri::App<MockRuntime> {
-        let app_handle_holder: Arc<Mutex<Option<tauri::AppHandle>>> = Arc::new(Mutex::new(None));
-        let app_handle_setter = app_handle_holder.clone();
         let app = mock_builder()
             .plugin(tauri_plugin_shell::init())
             .manage(ServerProcess {
                 child: Arc::new(Mutex::new(None)),
                 owns_port: Arc::new(AtomicBool::new(false)),
-                app_handle: app_handle_holder,
+                // app_handle left as None: no test calls start_server
+                // in a way that reads it (all tests seed first, causing
+                // start_server to return early).  A future test that
+                // actually starts a server via IPC will need a
+                // runtime-generic ServerProcess instead.
+                app_handle: Arc::new(Mutex::new(None)),
             })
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
## Fix: MockRuntime compilation error

CI caught that `tauri::AppHandle` doesn't implement `Deserialize`, which is required when a `#[tauri::command]` function is invoked via IPC with `MockRuntime` (used in the test harness).

**Root cause**: The `start_server` command took `app: tauri::AppHandle` as a command parameter. When `get_ipc_response` deserializes command arguments for the mock runtime, `AppHandle` can't be deserialized.

**Fix**: Stored a clone of `AppHandle` inside `ServerProcess` state at app setup time, so `start_server` fetches it from `State<ServerProcess>` instead of from a command parameter. This keeps the handle accessible without requiring it to be deserializable.

The real app setup already has access to the handle in its `.setup()` closure, so setting it was straightforward. The test helper (`build_test_app`) uses a shared-`Arc` pattern to inject the handle after `mock_builder().build()` returns.

TimeToBuildBob:
Fixed the AppHandle type mismatch that caused the test build to fail in CI. The test builder no longer tries to store `app.handle()` (`AppHandle<MockRuntime>`) into a field typed as `tauri::AppHandle` (`AppHandle<Wry>`) — no test reads `app_handle` because all tests seed the child process first, making `start_server` return early.

Commit: 1a407bf54

TimeToBuildBob:
**CI is now green** ✅

The "Test (unit + IPC)" failure was a type mismatch in the test builder — `AppHandle<MockRuntime>` can't be assigned to `Arc<Mutex<Option<tauri::AppHandle<Wry>>>>`. 

Fix (by @TimeToBuildBob): removed the `app_handle_setter` injection in `build_test_app()`. The handle stays `None` in tests because no test path reaches the `app_handle` extraction in `start_server` — the "already running" guard check fires first.

TimeToBuildBob:
@greptileai review

PR #2808, opened 2026-06-09
fix(webui): don't flush queued messages during tool confirmation

ErikBjare:
## Summary
Fixes the reported bug: queueing a message during a tool use immediately errored with "Generation already in progress".

**Root cause:** when generation pauses for a **tool confirmation**, `isGenerating` (`conversation.isGenerating || executingTool`) briefly goes false while a `pendingTool` is set. The queue-flush effect triggers on the generating→not-generating transition, so it sent the queued message *mid-step* — which the server rejected.

**Fix:** introduce `isBusy = isGenerating || !!pendingTool` and use it for:
- the queue decision in `handleSubmit` (submitting during a confirmation queues, not sends),
- the flush gate (only send queued messages once the whole step — generation + tool confirmation/execution — is done),
- the submit button state (shows "Queue"/"Stop" while awaiting a tool).

Other `isGenerating` uses (Escape-to-interrupt, disabling the mic button) are left as-is.

## Test plan
- [x] typecheck clean; `jest ChatInput.test.tsx` (4) passes
- [ ] Queue a message mid-generation that hits a tool confirmation → message stays queued, no "already in progress" error, sends after the step completes

### DIFF
--- a/webui/src/components/ChatInput.tsx
+++ b/webui/src/components/ChatInput.tsx
@@ -425,16 +425,16 @@ const OptionsButton: FC<{ isDisabled: boolean; children: React.ReactNode }> = ({
   </Popover>
 );
 
-const SubmitButton: FC<{ isGenerating: boolean; isDisabled: boolean; hasText: boolean }> = ({
-  isGenerating,
+const SubmitButton: FC<{ isBusy: boolean; isDisabled: boolean; hasText: boolean }> = ({
+  isBusy,
   isDisabled,
   hasText,
 }) => {
-  // When generating: show "Queue" if there's text, "Stop" if not
-  // When not generating: show "Send"
-  const showQueue = isGenerating && hasText;
-  const showStop = isGenerating && !hasText;
-  const canSubmit = !isDisabled && (isGenerating || hasText);
+  // When busy (generating or awaiting a tool): show "Queue" if there's text,
+  // "Stop" if not. When idle: show "Send".
+  const showQueue = isBusy && hasText;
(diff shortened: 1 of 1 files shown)


ErikBjare:
@greptileai review

ErikBjare:
@TimeToBuildBob What about the Greptile comments here? Worth addressing in a follow-up? Do it if so.

TimeToBuildBob:
Checked the Greptile P1 comment — **already handled** in the merged code. Line 973 of `ChatInput.tsx` already has `else if (onInterrupt && isGenerating)`, which is exactly the guard Greptile suggested. Stop only fires interrupt when actively generating, not during a pending tool confirmation. No follow-up needed.

TimeToBuildBob:
The two P2 Greptile findings are real — filed a follow-up PR: #2822.

- **Escape key no-op during tool confirmation** (line 1042): fixed — now uses `isBusy` instead of `isGenerating`
- **Stale log context** (lines 975/978): fixed — logs `{ isBusy }` for accurate trace

Both changes are 1-line each and don't affect the core queue-hold logic from this PR.

TimeToBuildBob:
Both Greptile P2 findings were worth fixing and are already addressed in gptme/gptme#2822 (merged today):

1. **Stale `isGenerating` in interrupt logs** → #2822 updates the log lines to capture `isBusy` instead
2. **Escape-to-interrupt inconsistent with Stop button** → #2822 gates the Escape handler on `isBusy` so both paths behave consistently during tool confirmation

No further follow-up needed here.

PR #2809, opened 2026-06-09
fix(tools): report unavailable tools accurately instead of 'invalid choice'

ErikBjare:
## Summary

A tool that exists but is currently unavailable (its `is_available()` returns False — e.g. `tts` when its server isn't running, or a backend without optional deps) was dropped from the CLI's valid `--tools` choices. So a bare `-t tts` was rejected as **`invalid choice: tts`** — misleading, since the tool exists, it's just inactive.

- **CLI** (`cli/main.py`): build the `--tools` choice set from all *known* tools, not just currently-available ones. Bare unavailable tool names now pass parse-time validation and are reported at load time instead. (`+tool` was already lenient; `-tool` exclusions stay strict for typo protection.)
- **Message** (`tools/__init__.py`): replace the hardcoded "(likely missing dependencies)" with an accurate message via a shared `_unavailable_message()` helper, used at all three unavailable sites (`get_toolchain`, `init_tools`, `load_tool`).
- **Hint** (`tools/base.py`): add an optional `ToolSpec.available_hint` so a tool can give specific guidance (e.g. "start the TTS server, or set GPTME_TTS_BACKEND=openrouter"); the generic message is used when no hint is set.

Before: `Error: Invalid value for '-t' / '--tools': invalid choice: tts.`
After: `Tool 'tts' is unavailable — it was discovered but its availability check failed (a required service may not be running, or optional dependencies/credentials are missing).`

## Test plan

- [x] `tests/test_tools.py` — new tests: `available_hint` surfaced in the error; generic fallback message wording
- [x] Existing `get_toolchain` strict/non-strict unavailable tests still pass
- [x] Manually verified `-t tts` with the TTS server down now reports unavailability instead of "invalid choice"

Follow-up: the `gptme-tts` plugin will set `available_hint` for a TTS-specific message (separate gptme-contrib PR).

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -858,7 +858,10 @@ def main(
     # We pass the tool_allowlist CLI argument. If it's not provided, init_tools
     # will load it from the environment variable TOOL_ALLOWLIST or the chat config.
     logger.debug(f"Using tools: {config.chat.tools}")
-    tools = init_tools(config.chat.tools)
+    try:
+        tools = init_tools(config.chat.tools)
+    except ValueError as e:
+        raise click.UsageError(str(e)) from e
 
     # Check if we're opening an existing conversation (via --resume, --name, or pick)
     # If so, skip generating initial messages (including expensive context_cmd)
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -380,7 +380,11 @@ Run 'gptme-util --help' for all utility commands."""
     default=None,
     multiple=True,
     type=CommaSeparatedChoice(
-        _available_tools + ["none"],
+        # Accept all *known* tools, not just currently-available ones, so a tool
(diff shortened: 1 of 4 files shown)


ErikBjare:
@greptileai review

ErikBjare:
@greptileai review

ErikBjare:
@TimeToBuildBob address the review

TimeToBuildBob:
Addressed Greptile P1 review: wrapped `init_tools(config.chat.tools)` in `try/except ValueError → click.UsageError` at line 861 so the clean unavailable message is shown instead of a raw Python traceback. All 35 tool tests pass.

P2 suggestions (punctuation consistency and "Skipping." log suffix) were already applied in the prior commit `9170d224f` — no additional changes needed there.

TimeToBuildBob:
@greptileai review

ErikBjare:
@greptileai review

PR #2826, opened 2026-06-10
fix(cloud): parse max_completion_tokens for gptme provider models

TimeToBuildBob:
## Problem

Dogfooding `gptme-auth login` (ErikBjare/bob#845) hit two bugs after PR #2820 fixed the URL routing:

1. **`max_tokens: Field required`** — Every gptme cloud API call fails with HTTP 400 because gptme never sends `max_tokens`. The Supabase messages function passes requests to Anthropic, which requires `max_tokens`. The root cause: `_openai_compatible_model_to_modelmeta` set `max_output=None` for all dynamically-fetched models, so gptme omitted the parameter.

2. **Model ID format confusion** — The existing docstring showed `gptme -m gptme/claude-sonnet-4-6`, but the Supabase models endpoint returns IDs like `anthropic/claude-sonnet-4-6`. The correct model spec is `gptme/anthropic/claude-sonnet-4-6` (provider prefix preserved). Using `gptme/claude-sonnet-4-6` strips to bare `claude-sonnet-4-6` which the Supabase function rejects.

## Fix

- **`_openai_compatible_model_to_modelmeta`**: parse `max_completion_tokens` from the models list response (present for every model in the Supabase endpoint) and set it as `max_output`. Also forward `supports_streaming` and `supports_reasoning` from the response instead of hardcoding defaults.

- **`llm_gptme.py` docstring**: document the correct model ID format (`gptme/anthropic/MODEL`, `gptme/openai/MODEL`) and that `gptme --list-models` is the right way to discover available models.

## Verification

```bash
# Direct Supabase endpoint — confirmed working
curl -s -X POST "https://kpkxgnfpyntahyhckhgm.supabase.co/functions/v1/messages" \
  -H "Authorization: Bearer $GPTME_CLOUD_API_KEY" \
  -d '{"model":"anthropic/claude-haiku-4-5-20251001","messages":[{"role":"user","content":"say hello"}],"max_tokens":50,"stream":false}'
# → {"content":[{"type":"text","text":"Hello! ..."}], ...}
```

After this fix, gptme should correctly send `max_tokens` (from `max_output`) to the Supabase messages function.

## Note on model naming

The server-side `messages` function requires the `anthropic/` prefix — it returns "Unsupported provider prefix: claude-sonnet-4-6" for bare names. A follow-up could add bare-name fallback in the edge function (e.g. auto-prefix with `anthropic/` when no recognized prefix found), but for now the client-side doc fix documents the correct format.

### DIFF
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -202,12 +202,15 @@ _OPENROUTER_DEFAULT_MAX_TOKENS = 16000
 
 
 def _resolve_max_tokens(model: str, max_tokens: int | None) -> int | None:
-    """Apply the GPTME_MAX_TOKENS env override or a sane default for OpenRouter models.
+    """Apply the GPTME_MAX_TOKENS env override or a sane default for providers that require it.
 
     When max_tokens is omitted, OpenRouter reserves model.max_output +
     reasoning_budget credits per request. For large models like Sonnet 4.6
     that is ~65k tokens. A partially-spent daily-budget key rejects this
     with HTTP 402 even when the actual response would fit comfortably.
+
+    The gptme cloud provider (Supabase messages function) requires max_tokens
+    explicitly — omitting it returns HTTP 400 Field required.
 
     Setting a default max_tokens reduces the reservation so partial-budget
     keys still work.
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed the critical gap flagged in the Greptile review.

The root cause was correct: `_openai_compatible_model_to_modelmeta` was populating `model_meta.max_output` from `max_completion_tokens`, but `_resolve_max_tokens()` in `llm/__init__.py` only applied a default for the `openrouter` provider — the `gptme` provider fell through with `return None`.

Fix in commit `0ded901c1`: extend `_resolve_max_tokens()` to handle `provider == "gptme"` by using `model_meta.max_output or 4096` as the fallback, mirroring the pattern in `llm_anthropic.py`. The `GPTME_MAX_TOKENS` env override now applies to both `openrouter` and `gptme` providers.

TimeToBuildBob:
@greptileai review

PR #2830, opened 2026-06-11
test(server): add PATCH edit-message and DELETE message edge-case tests

TimeToBuildBob:
## Summary

The PATCH `/api/v2/conversations/{id}/messages/{index}` endpoint had **zero tests**. This PR adds 8 targeted tests covering the previously untested edit and delete message edge cases.

- **PATCH 404** on nonexistent conversation
- **PATCH 404** on out-of-range index (e.g. 999)
- **PATCH 400** when request has no `content` and no `truncate=1` flag
- **PATCH 400** when trying to edit a system message (index 0)
- **PATCH 200** success path for editing a valid user message
- **DELETE 404** on nonexistent conversation
- **DELETE 404** on out-of-range index
- **DELETE 400** when trying to delete a system message

Also documents that the server prepends its own system messages ahead of client-supplied messages, so the fixture dynamically discovers the actual user-message index via a GET request (instead of hardcoding an index that breaks when the server header message count changes).

## Test plan

- [x] All 8 new tests pass: `uv run pytest tests/test_server.py -k "edit_message or delete_message" -v`
- [x] Full server suite passes (45 tests): `uv run pytest tests/test_server.py -q`
- [x] No regressions

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -830,3 +830,137 @@ def test_spa_fallback_api_returns_json(tmp_path: Path):
         # non-api path → SPA fallback serves index.html
         response = c.get("/some/deep/link")
         assert response.status_code == 200
+
+
+# --- Message edit (PATCH) and delete (DELETE) edge cases ---
+
+
+@pytest.fixture
+def conv_with_messages(client: FlaskClient):
+    """Create a conversation with a system message and a user message.
+
+    Returns (convname, user_msg_index) since the server prepends its own
+    system messages so the client-supplied messages don't start at index 0.
+    """
+    import random
+
+    convname = f"test-edit-{random.randint(0, 1000000)}"
+    response = client.put(
+        f"/api/v2/conversations/{convname}",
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile review feedback in follow-up commit (b51483cce):

**Fixes applied:**
1. **Removed redundant `import random`** in `conv_with_messages` fixture — module-level import already present
2. **Replaced bare `next()` with explicit assertion** — now uses list comprehension + `assert` so a missing user message produces a clear test failure (not a `StopIteration` error)
3. **Added `test_api_v2_delete_message_success`** — covers the DELETE 200 happy path, verifying:
   - Status 200 returned
   - Response contains updated log
   - Message count decreases by 1
   - Original user message content is no longer present

**Verification:**
- All 9 edit/delete tests pass
- Full server suite passes (46/46) with no regressions

TimeToBuildBob:
@greptileai review

PR #2838, opened 2026-06-11
feat(webui): add / keyboard shortcut to focus conversation search

TimeToBuildBob:
## Summary

- Pressing `/` anywhere in the sidebar (outside any input) focuses the conversation search box — matching the GitHub/Discord/Slack convention
- The shortcut is suppressed when `document.activeElement` is an `<input>`, `<textarea>`, `<select>`, or `contenteditable` element, so typing `/` in the message input is unaffected
- Two new tests: one verifying the shortcut fires, one verifying it is suppressed when an input is already focused

Depends on #2827 (conversation list search/filter) which has already merged.

## Test plan

- [ ] Open the webui with conversations loaded
- [ ] Press `/` with no input focused — search box should receive focus and be selected
- [ ] Type some characters in the message input, then press `/` — it should type into the message input normally (not jump to search)
- [ ] Press `/` while the search box is already focused — it should not re-select or flicker
- [ ] `npm test -- --testPathPattern=ConversationList` passes (31 tests)

### DIFF
--- a/webui/src/components/ConversationList.tsx
+++ b/webui/src/components/ConversationList.tsx
@@ -266,6 +266,29 @@ export const ConversationList: FC<Props> = ({
     return () => window.removeEventListener('keydown', handleFilterShortcut);
   }, []);
 
+  useEffect(() => {
+    const handleSlashShortcut = (e: KeyboardEvent) => {
+      if (e.key !== '/' || e.altKey || e.metaKey || e.ctrlKey || e.shiftKey) return;
+      if (!filterInputRef.current) return;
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the suppression test in b8b0261 — it now focuses a separate injected `<input>` element (simulating the message input) instead of the search input itself. The guard has to actually fire for the test to pass, so removing the `activeElement` check would now cause a failure.

TimeToBuildBob:
@greptileai review

PR #2839, opened 2026-06-11
test(webui): add Playwright e2e keyboard navigation tests

TimeToBuildBob:
## Summary

- Adds `webui/e2e/keyboard.spec.ts` with four keyboard navigation e2e tests:
  - **Ctrl+K** opens the command palette; **Esc** closes it
  - **?** opens the shortcuts dialog (with input-focus guard); **Esc** closes it
  - Conversation rows are Tab-focusable (`role=button`, `tabindex=0`) and Enter triggers selection (`aria-pressed` flips to `true`)
  - **Focus trap**: Tab cannot escape an open Radix UI Dialog (6 Tab presses, all stay inside)

Complements the existing unit tests added in #2829 (keyboard-accessible conversation rows) and the existing e2e split-view shortcut test in `conversation.spec.ts`.

## Test plan
- [ ] CI WebUI check passes (jest unit tests + Playwright e2e)
- [ ] All four new tests pass against the dev server

### DIFF
--- a/webui/e2e/keyboard.spec.ts
+++ b/webui/e2e/keyboard.spec.ts
@@ -27,7 +27,7 @@ test.describe('Keyboard Navigation', () => {
 
     await page.keyboard.press('?');
     await expect(dialog).toBeVisible({ timeout: 3000 });
-    await expect(dialog.getByText(/command palette/i)).toBeVisible();
+    await expect(dialog.getByText('Open command palette')).toBeVisible();
 
     await page.keyboard.press('Escape');
     await expect(dialog).not.toBeVisible({ timeout: 3000 });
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Fixed CI e2e failures + Greptile findings

Pushed commit 7b5c141db with three fixes:

1. **Consistent CI failure** (`shortcuts dialog: ? opens, Esc closes`): `getByText(/command palette/i)` matched TWO elements — the shortcut description `<span>Open command palette</span>` and the section header `<h3>Command palette</h3>`. Playwright's strict mode enforcement crashed on the ambiguity. Fixed by using exact text `getByText('Open command palette')`.

2. **Greptile P1** (focus-trap silent pass): The `evaluate` fallback `return dlg ? !dlg.contains(active) : false` silently passed all remaining loop iterations when the dialog wasn't found in the DOM. Changed to `: true` so a missing dialog fails the assertion.

3. **Greptile P2** (redundant assertions): `tabindex` and `role` `getAttribute` checks were guaranteed true by the locator selector `[role="button"][tabindex="0"]`. Removed them.

TimeToBuildBob:
Pushed commit 0682d8f80 to fix the remaining e2e failure:

**`command palette: Ctrl+K opens, Esc closes`** — `cmdk` v1 intercepts the Escape key at the element level before Radix UI Dialog's `DismissableLayer` fires, so the dialog never received the close signal. Added an explicit `Escape` branch in the existing `Ctrl+K` keydown handler in `CommandPalette.tsx` that calls `setOpen(false)` directly.

TimeToBuildBob:
@greptileai review

PR #2794, opened 2026-06-08
feat(webui): add OpenRouter STT fallback for unsupported browsers

TimeToBuildBob:
## Summary
- add an authenticated `/api/v2/audio/transcriptions` endpoint that proxies short recordings to OpenRouter STT
- extend the webui speech input hook to fall back to `MediaRecorder` + server transcription when browser `SpeechRecognition` is unavailable
- keep the speech button responsive with a transcribing state and cover the new path with API + hook tests

## Verification
- `/home/bob/gptme/.venv/bin/python -m pytest tests/test_server_v2.py -k audio_transcriptions -q`
- `./node_modules/.bin/jest --runTestsByPath src/hooks/__tests__/useSpeechToText.test.tsx`
- `./node_modules/.bin/eslint src/components/SpeechInputButton.tsx src/hooks/useSpeechToText.ts src/hooks/__tests__/useSpeechToText.test.tsx src/utils/api.ts src/utils/demoApiClient.ts`
- `./node_modules/.bin/tsc -p tsconfig.app.json --noEmit`

PR #2840, opened 2026-06-11
feat(webui): add settings-level STT provider choice (browser vs server)

TimeToBuildBob:
## Summary

Add a new client-side setting `sttProvider` (`browser` | `server`) that lets users choose between the browser SpeechRecognition API and server-backed transcription when both are available. Previously, the server path was only used as a fallback when browser STT was unsupported.

## Changes
- New `sttProvider` field in `Settings` interface (default: `browser`, preserving current behavior)
- Updated `useSpeechToText` hook to respect the setting: when `'server'` is selected, uses MediaRecorder + server transcription even if `SpeechRecognition` is available
- New dropdown in Settings → Audio section for STT engine choice
- Tests: server mode works when browser STT is available, `isSupported` remains true in server-preference mode

## Testing
- `npm run typecheck` — clean
- `npm test` — 460/460 passing (52 suites)
- 4 new tests in `useSpeechToText.test.tsx` (all passing)

## Related
- Follow-up to: #2794 (automatic OpenRouter fallback)
- Task: ErikBjare/bob#841

### DIFF
--- a/webui/src/components/SettingsContent.tsx
+++ b/webui/src/components/SettingsContent.tsx
@@ -286,6 +286,33 @@ export function SettingsContent({
               </div>
             )}
 
+            <Separator />
+
+            <div className="space-y-2">
+              <Label htmlFor="stt-engine" className="text-sm">
+                Speech-to-text engine
+              </Label>
+              <p className="text-xs text-muted-foreground">
+                Choose which engine transcribes your microphone input for dictation.
+              </p>
+              <Select
+                value={settings.sttProvider}
+                onValueChange={(value: 'browser' | 'server') =>
+                  updateSettings({ sttProvider: value })
+                }
+              >
+                <SelectTrigger id="stt-engine" className="w-full">
+                  <SelectValue placeholder="Select STT engine" />
+                </SelectTrigger>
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Good catch from Greptile. Fixed in 5af9589e5.

The `startListening` early-return at line 166 (when `serverEnabled` is false) now emits `state='error'` when the user explicitly selected the server provider but OpenRouter isn't configured — instead of silently doing nothing. Added a test covering the exact scenario: `browserSupported=true`, `sttProvider='server'`, `providers_configured=[]`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Ready to merge — CI green, Greptile review cycle complete.

**What landed**: settings-level STT provider choice (browser vs server) in the webui. Users can now explicitly pick their preferred transcription backend instead of relying on automatic fallback.

**Greptile finding addressed** (commit 5af9589e5): `startListening` now emits `state='error'` when the user has explicitly selected the server provider but OpenRouter isn't configured, instead of silently doing nothing. Test added for that path.

Waiting on a maintainer merge when you have a moment.

PR #2847, opened 2026-06-12
test(complete): cover cross-variant auto-reply exit counting

TimeToBuildBob:
## What

Adds a regression test for the auto-reply exit logic in `gptme/tools/complete.py`.

## Why

The two existing exit-after-2 tests (`test_two_auto_replies_exits`, `test_two_auto_replies_with_incomplete_todos_exits`) each use a **single** auto-reply variant twice. But the actual fix in #2846 was that both variants share the marker prefix `"No tool call detected in last message"` so they count toward the **same** exit threshold. No test exercises a *mixed* sequence — so a regression to per-variant markers (the pre-#2846 bug) would pass CI.

This is not hypothetical: a session that starts with incomplete todos (reminder variant) and then completes them while still producing a tool-less response (confirm variant) hits exactly this mixed path. Before #2846 it would undercount and never exit — an infinite auto-reply cycle until the budget/timeout.

## Test

`test_two_auto_replies_mixed_variants_exits`: first auto-reply is the reminder variant, second is the confirm variant; asserts `SessionCompleteException` still fires.

```
18 passed in 2.57s  (tests/test_complete.py)
```

No production code changes — test-only.

### DIFF
--- a/tests/test_complete.py
+++ b/tests/test_complete.py
@@ -165,6 +165,45 @@ class TestAutoReplyHook:
             )
             assert len(result) == 0
 
+    def test_two_auto_replies_mixed_variants_exits(self):
+        """Should exit after two no-tool replies even when the two auto-reply
+        messages are *different* variants.
+
+        Realistic production case: the first turn has incomplete todos, so the
+        reminder variant fires; the model then completes the todos but still
+        produces a tool-less response, so the confirm variant fires. Both
+        variants share the ``"No tool call detected in last message"`` marker,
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Addressed Greptile review feedback: moved `test_two_auto_replies_mixed_variants_exits` from `TestTodoContinuationEnforcer` to `TestAutoReplyHook` — the test exercises pure auto-reply counting logic and doesnt touch any todo state, so it belongs alongside the other exit-threshold tests. Fix pushed to the branch.

PR #2862, opened 2026-06-12
perf(tools): lazy-load mcp.types and bashlex to reduce startup time

TimeToBuildBob:
## Summary

- Moves `import mcp.types as mcp_types` from module level in `mcp_adapter.py` to inside the 3 functions that use it at runtime. The module already had `from __future__ import annotations`, so annotation-only uses don't need the import to be eager.
- Moves `import bashlex` from module level in `shell.py` to inside `split_commands()` — the only function that calls `bashlex.parse()`.

## Benchmark results

Measured on warm startup (`from gptme.cli.main import main`):

| | Before | After | Delta |
|---|---|---|---|
| Warm startup | 1502ms | 1145ms | **-357ms (−24%)** |

Breakdown:
- `bashlex` lazy: ~186ms saved
- `mcp.types` lazy: ~171ms saved

Current CI threshold is 2500ms; latest benchmark shows 1658ms. This change brings warm startup well under budget.

## Root cause

`gptme/cli/main.py:272` calls `get_available_tools(include_mcp=False)` at module level, which triggers `_collect_tool_modules()` → imports ALL tool submodules including `mcp_adapter.py` → loads the `mcp` package and `bashlex`, neither of which is needed at import time.

## Remaining opportunity

Lazifying `MCPClient`/`MCPRegistry` in `mcp_adapter.py` (currently imported at module level due to `_registry = MCPRegistry()` at line 40) would save another ~500ms, but requires refactoring the module-level singleton. Left as a follow-up.

### DIFF
--- a/gptme/tools/mcp_adapter.py
+++ b/gptme/tools/mcp_adapter.py
@@ -11,11 +11,11 @@ if TYPE_CHECKING:
     import mcp.types as mcp_types
 
     from ..hooks.elicitation import ElicitationRequest, ElicitationResponse
+    from ..mcp.client import MCPClient
+    from ..mcp.registry import MCPRegistry
 
 from gptme.config import Config, MCPServerConfig, get_config, set_config
 
-from ..mcp.client import MCPClient, MCPInterruptedError
-from ..mcp.registry import MCPRegistry, format_server_details, format_server_list
 from ..message import Message
 from ..util.ask_execute import execute_with_confirmation
 from .base import (
@@ -36,11 +36,21 @@ _mcp_clients: dict[str, MCPClient] = {}
 # Add type annotation for tool_specs
 tool_specs: list[ToolSpec] = []
 
-# Global registry instance
-_registry = MCPRegistry()
+# Lazy-initialized registry instance (loaded on first MCP registry call)
+_registry_instance: MCPRegistry | None = None
 
 # Cache of dynamically loaded servers
 _dynamic_servers: dict[str, MCPClient] = {}
+
+
+def _get_registry() -> MCPRegistry:
+    """Lazy getter for MCPRegistry — defers the import until first MCP registry call."""
+    global _registry_instance
+    if _registry_instance is None:
+        from ..mcp.registry import MCPRegistry
+
+        _registry_instance = MCPRegistry()
+    return _registry_instance
 
 
 def get_mcp_clients() -> dict[str, MCPClient]:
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is green, Greptile is 5/5, and the PR is cleanly mergeable.

I checked the deterministic self-merge gate before landing it. That gate still blocks this PR because it is a cross-repo lane touching `gptme/tools/*`, so this one needs a maintainer merge rather than Bob's auto/self-merge path.

Bench result from the PR:
- Warm startup: `1502ms -> 1145ms` (`-357ms`, `-24%`)

Ready to merge.

TimeToBuildBob:
## Follow-up: MCPClient/MCPRegistry lazy loading

I landed the "remaining opportunity" mentioned in the PR description: the module-level `_registry = MCPRegistry()` is now a lazy singleton via `_get_registry()`, and `MCPClient`/`MCPInterruptedError` are also imported per-function (or removed from the module-level imports entirely).

The original 4 functions that used the eager imports are all updated. All 4 call sites of `_registry` now go through `_get_registry()` (verified with `git grep`). The `MCPInterruptedError` was being referenced at runtime in `except` clauses but was in the `TYPE_CHECKING` block — ruff TC004 caught that. This commit fixes that and also restores 5 missing imports that were dropped by an over-broad replace.

**Updated benchmark on this PR** (warm startup, `from gptme.cli.main import main`):

| | Before | After | Delta |
|---|---|---|---|
| Warm startup | 1502ms | 664ms | **-838ms (−56%)** |

Breakdown:
- `bashlex` lazy: ~186ms
- `mcp.types` lazy: ~171ms
- `MCPClient`/`MCPRegistry` lazy: ~481ms

The `mcp` package itself pulls in `httpx`, `pydantic`, `anyio`, `starlette`, etc. — those are no longer loaded at startup when MCP isn't used.

Local verification: `pytest tests/test_tools_mcp.py` (71 tests) passes.

TimeToBuildBob:
## CI Fix Applied

The test failures were caused by two issues in `tests/test_mcp_adapter.py`:

1. **MCPClient patch target**: Tests used `patch("gptme.tools.mcp_adapter.MCPClient", ...)` but `MCPClient` is no longer a module-level attribute of `mcp_adapter.py` (it was exported from `gptme/mcp/__init__.py` via `from .client import MCPClient` and imported under `TYPE_CHECKING` in mcp_adapter). Fixed by targeting `gptme.mcp.client.MCPClient` instead — the module where the class actually lives.

2. **_registry → _get_registry()**: The module-level `_registry = MCPRegistry()` was replaced by a `_get_registry()` lazy getter function, but tests still patched `_registry.*` methods. Fixed by patching `_get_registry` to return a mock registry with the expected methods.

**Verification**: `pytest tests/test_mcp_adapter.py` — 17/17 pass (7 skipped for MCP elicitation bridge requiring mcp >= 1.22.0, pre-existing).

TimeToBuildBob:
@greptileai review

PR #2863, opened 2026-06-12
perf(server): add ETag conditional request support for conversations endpoints

TimeToBuildBob:
## Summary

- Adds `ETag` response headers to `GET /api/v2/conversations` and `GET /api/v2/conversations/:id`
- Validates `If-None-Match` request headers and returns `304 Not Modified` when data is unchanged
- 8 new integration tests covering 200/304/mismatch cases and ETag invalidation after message append

## How it works

**Conversations list** (`GET /api/v2/conversations`): ETag derived from MD5 of conversation IDs + modification times. Only applied on the no-search, no-detail fast path (the common polling case). Works on both the cache-hit and fresh-fetch paths.

**Single conversation** (`GET /api/v2/conversations/:id`): ETag from MD5 of `conversation.jsonl` mtime. The 304 check happens *before* loading the full log manager, so cache hits avoid the file read entirely.

## Motivation

The webui polls conversations on page load and navigation. After cursor pagination (#2860) and server-side caching (#2856/#2859), data is cheap to generate — but the response is still serialized and transmitted even when nothing changed. For users with 500+ conversations, each list response is several KB. ETags turn repeat polls into free 304s with ~0 bytes transferred.

Browsers handle `If-None-Match` automatically when using `fetch` cache mode; the webui client can opt in with a single header addition.

## Test plan

- [x] `test_returns_etag_header` — 200 responses carry ETag
- [x] `test_304_on_matching_etag` — identical ETag → 304 with empty body
- [x] `test_200_on_mismatched_etag` — wrong ETag → 200 with fresh data
- [x] `test_no_etag_on_search` — search responses don't 304 inadvertently
- [x] `test_etag_changes_after_message_added` — mtime-based ETag invalidates correctly
- [x] Existing `test_server_bad_input` and `test_server_cors` tests still pass (24/24)

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -427,13 +427,19 @@ def _etag_conversations(conversations: "list[ConversationMeta]") -> str:
 
 
 def _etag_conversation(logdir: "Path") -> str:
-    """Compute ETag from the conversation log file's mtime."""
+    """Compute ETag from the conversation log file's mtime and size.
+
+    Including size prevents stale 304s on filesystems with 1-second mtime
+    granularity where two writes within the same tick leave mtime unchanged.
+    """
     logfile = logdir / "conversation.jsonl"
     try:
-        mtime = logfile.stat().st_mtime
+        stat = logfile.stat()
+        mtime = stat.st_mtime
+        size = stat.st_size
     except OSError:
         return ""
-    return hashlib.md5(str(mtime).encode()).hexdigest()[:16]
+    return hashlib.md5(f"{mtime}:{size}".encode()).hexdigest()[:16]
 
 
 v2_api = flask.Blueprint("v2_api", __name__)
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1043,14 +1043,18 @@ def api_conversations():
         if elapsed < _CONVERSATIONS_CACHE_TTL:
             cached = _conversations_cache
             etag = _etag_conversations(cached[:limit])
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile review feedback (f9de44f):

- **Weak ETags**: switched from strong to `W/"..."` weak ETags since they're derived from mtime, not content hashes. Uses `contains_weak()` for proper RFC 7232 weak comparison.
- **Cache-Control**: added `Cache-Control: no-cache` to all ETag responses (200 and 304) so clients always revalidate.
- **ETag in 304**: 304 responses now include the ETag header as required by RFC 7232 §4.1.
- **Test fix**: corrected `GPTME_LOGS_DIR` → `GPTME_LOGS_HOME` in the test fixture (env var was wrong, tests were reading the real logs dir).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the remaining Greptile concern (5eab89e):

**mtime-granularity fix**: `_etag_conversation` now hashes both `st_mtime` and `st_size`. On filesystems with 1-second mtime granularity, two writes within the same tick leave mtime unchanged but the file grows — so including size ensures the ETag changes after any append. This fixes both the test reliability issue and the underlying production stale-304 risk.

Also resolved the 3 inline threads from the first review (all were already fixed in f9de44f: weak ETags, Cache-Control, single-conversation weak ETags).

TimeToBuildBob:
@greptileai review

PR #2867, opened 2026-06-13
fix: two startup hangs — skip OpenRouter fetch for bare model names, non-blocking piped stdin

TimeToBuildBob:
Two separate issues causing `gptme --model <nonexistent>` to hang for 10s+ 
before showing an error message:

**1. get_model() fetched OpenRouter API for bare model names**
`gptme/llm/models/resolution.py`: `get_model()` tried OpenRouter API dynamic
fetch for bare model names without '/' — but OpenRouter models are always
`provider/model` format, so a bare name can never match. The API timeout
caused a ~10s hang for every nonexistent bare model name.

**2. _read_stdin() blocked on idle pipe under uv run**
`gptme/cli/main.py`: `_read_stdin()` called `sys.stdin.read()` on a pipe
that `select.select()` marked as readable — but on a pipe with an open but
idle write end, 'readable' doesn't guarantee data is available; `read()`
then blocks indefinitely waiting for data or EOF. Fixed by using `os.read()`
with select sub-timeouts so an idle pipe returns quickly.

**Testing**: 128 tests pass, 2 pre-existing failures in resume tests (unrelated).

Fixes a common UX complaint for users running gptme in non-TTY environments.

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -84,6 +84,12 @@ def _validate_model_param(
 script_path = Path(os.path.realpath(__file__))
 _STDIN_PIPE_GRACE_PERIOD = 1.0
 
+# Sub-timeout used by the inner read loop in `_read_stdin` to disambiguate
+# "data ready" from "pipe fd is readable but write end is open and idle".
+# Tuned to 100ms — generous enough to bridge small producer gaps without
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

PR #2868, opened 2026-06-13
fix(server): validate model type in config PATCH to prevent 500

TimeToBuildBob:
## Summary

- `PATCH /api/v2/conversations/{id}/config` with a non-string `chat.model` value (e.g. `{"chat": {"model": 12345}}`) crashed with a 500 (`TypeError: argument of type 'int' is not iterable`) deep inside `get_model()` instead of returning a clean 400
- Add early type validation mirroring the existing `tools` allowlist check: if `chat.model` is present but not a string or null, return 400 before any side-effecting operations
- Add parametrized regression test covering int, float, bool, list, and dict as bad model values (5 cases, all now return 400)

Found during dogfood sweep of the config PATCH surface.

## Test plan

- [x] `pytest tests/test_server_v2.py -k config_patch` — 15/15 passed locally
- [x] New test: `test_v2_chat_config_patch_rejects_non_string_model` (5 parametrized cases)
- [x] Pre-commit hooks clean

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -2140,7 +2140,7 @@ def api_conversation_config_patch(conversation_id: str):
                 return raw_allowlist  # 400: bad type, no side effects
             tool_allowlist = raw_allowlist  # narrowed to list[str] | None
         if "model" in chat_patch and not isinstance(
-            chat_patch["model"], (str, type(None))
+            chat_patch["model"], str | type(None)
         ):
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed Greptile P2 feedback: error message now says "model must be a string or null" since null is explicitly accepted by the guard.

Applied in ebb44b959. The change is purely cosmetic — guard logic is unchanged.

PR #2874, opened 2026-06-13
fix(server): tolerate pre-existing workspace symlink in ChatConfig.from_logdir

ErikBjare:
## Summary
A conversation whose `workspace` is a **symlink to a now-missing directory** (e.g. an external workspace that was moved/deleted) broke gptme in several places. This PR makes the read path tolerant and the run path fail *cleanly* instead of crashing.

### Read path (list/load) — no longer crashes
- `GET /api/v2/conversations` 500'd because `ChatConfig.from_logdir` calls `workspace.mkdir(parents=True, exist_ok=True)`, and `exist_ok=True` only tolerates a pre-existing *directory* — a symlink (even broken) raises `FileExistsError`, taking down the whole list.
- Extracted `ensure_workspace_dir()` (symlink-tolerant mkdir) and reused it across `from_logdir` / `from_dict` (`@log`) / `save()` and the CLI `@log` path, so no workspace-creation site `FileExistsError`s on a pre-existing symlink.

### Run path (CLI + server step) — fails with an actionable message
Listing/loading was fixed, but *running* such a conversation still crashed at `os.chdir(workspace)`:
- **CLI** (`chat.py`): bare `FileNotFoundError`.
- **Server** (`session_step.py`): `step()` runs in a **daemon thread**, so the crash left the session stuck in `generating` forever with no error surfaced to the webui.

Added `require_workspace_exists()` (raises an actionable message) and validated before the chdir in both paths. The CLI surfaces it via the existing fatal-error handler; the server emits a visible error event and resets the generating flags.

**Policy:** a moved/deleted workspace fails until manually corrected (restore the dir, or update `chat.workspace`) rather than being silently recreated or relocated.

Observed in the wild:
```
FileExistsError: [Errno 17] File exists: '.../logs/2025-03-21-skipping-purple-octopus/workspace'
  at gptme/config/chat.py:169 in from_logdir
```
where `workspace -> /Users/.../server-refactor` (symlink to a since-removed dir).

## Test plan
- [x] New unit tests: `ensure_workspace_dir` (creates/tolerates symlink), `require_workspace_exists` (raises only when missing), `from_logdir` broken-symlink workspace
- [x] `tests/test_chat_config.py` passes (11 passed)
- [x] Reproduced the original 500 against the real logdir; gone after the fix
- [x] ruff + mypy clean (pre-commit)

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -6,7 +6,7 @@ from collections.abc import Callable, Generator
 from pathlib import Path
 
 from .commands import execute_cmd
-from .config import ChatConfig, get_config
+from .config import ChatConfig, get_config, require_workspace_exists
 from .constants import (
     DECLINED_CONTENT,
     INTERRUPT_CONTENT,
@@ -130,6 +130,7 @@ def chat(
         # Initialize workspace
         if not is_output_json():
             console.log(f"Using workspace: {path_with_tilde(workspace)}")
+        require_workspace_exists(workspace)
         os.chdir(workspace)
 
         # print log (suppressed in JSON output mode)
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -29,7 +29,7 @@ import gptme
 
 from ..chat import chat
 from ..commands import _gen_help
-from ..config import setup_config_from_cli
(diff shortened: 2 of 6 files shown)


ErikBjare:
@greptileai review

PR #2875, opened 2026-06-13
refactor(llm): extract shared Responses API event loop into openai_responses.py

TimeToBuildBob:
## Summary

- Extracts the duplicated ~100-line Responses API SSE event-handling loop from both `llm_openai.py` and `llm_openai_subscription.py` into a new shared `_stream_responses_events()` helper in `openai_responses.py`
- Net -70 LOC across the three files (134 inserted, 204 deleted)
- The code comment in `llm_openai.py` already called for this exact merge

## What the shared helper does

`_stream_responses_events(event_iter, *, usage_callback=None)` handles the full event loop using `_obj_get()` throughout — so it works with both OpenAI SDK event objects (attribute access) and raw SSE dicts from the subscription provider (dict access). Accepts both `response.reasoning_text.delta` (SDK) and `response.reasoning.delta` (chatgpt.com backend) interchangeably.

**Callers after this change:**
- `_stream_responses()` — wraps `client.responses.create()` stream; passes a `usage_callback` closure to capture metadata from `response.completed`
- `subscription.stream()` — wraps `response.iter_lines()` in a `_sse_events()` generator and passes it to the shared helper

## Test plan

- [x] All 106 existing `test_llm_openai` + `test_llm_openai_sampling` tests pass
- [x] `mypy` clean on all 3 modified files
- [x] `ruff check` + `ruff format` clean

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -31,10 +31,9 @@ from .openai_responses import (
     ToolCallFunction,
     _content_to_responses_input,  # noqa: F401
     _extract_usage_token_counts,
-    _filter_duplicate_thinking_text,
-    _longest_suffix_prefix,
     _messages_dicts_to_responses_input,
     _obj_get,
+    _stream_responses_events,
     _tool_spec_to_responses_tool,
 )
 from .utils import (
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
CI is fully green (14/14). Greptile 4/5 with two P2 notes about minor behavioral differences that are acceptable (reasoning block flush on `output_item.added` — arguably more correct behavior; function-call header deferred to first argument delta — identical for the common case).

Not self-merge eligible (Greptile <5/5, file paths outside allowlist). Flagging for human review.

@ErikBjare this is ready for review when you get a chance.

PR #2877, opened 2026-06-13
fix(server): return 400 for non-positive limit in conversations API

TimeToBuildBob:
## Summary

- `GET /api/v2/conversations?limit=-1` and `?limit=0` previously returned HTTP 200 with data, silently clamping the value to 1 via `max(1, min(limit, 1000))`.
- The message pagination endpoint (`GET /api/v2/conversations/{id}`) already correctly returns 400 for `limit <= 0`. This PR aligns the conversations list and external sessions endpoints to match.
- Adds three regression tests: invalid limits rejected (0, -1, -100), valid limits accepted (1-1000), overlimit capped to 1000.

## Reproduction

```bash
# Before fix (returns 200 with one conversation):
curl http://127.0.0.1:5700/api/v2/conversations?limit=-1
curl http://127.0.0.1:5700/api/v2/conversations?limit=0

# After fix (returns 400):
# {"error": "limit must be a positive integer"}
```

## Test plan

- [x] `test_conversations_invalid_limit` — verifies -1, 0, -100 return 400
- [x] `test_conversations_valid_limits` — verifies 1, 10, 100, 1000 return 200
- [x] `test_conversations_limit_capped` — verifies 9999 returns 200 (silently capped)

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -865,15 +865,17 @@ def api_external_sessions():
 )
 def api_external_session(external_session_id: str):
     """Get a normalized read-only external session transcript."""
-    provider = get_external_session_provider()
-    if provider is None:
-        return flask.jsonify({"error": "external session provider unavailable"}), 503
-
     try:
         days = int(request.args.get("days", 30))
     except (ValueError, TypeError):
         return flask.jsonify({"error": "days must be an integer"}), 400
-    days = max(1, min(days, 3650))
+    if days <= 0:
+        return flask.jsonify({"error": "days must be a positive integer"}), 400
+    days = min(days, 3650)
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed both Greptile P2 findings in 3b2f39c:

1. **`api_external_session` days validation** — moved parameter parsing/validation before the provider check so `days <= 0` now returns 400 (consistent with the list endpoint) instead of silently clamping. No more `days=-1` → 200 divergence.

2. **Test coverage for external sessions** — added `test_external_session_invalid_days` covering -1, 0, -100 cases. The validation runs before the provider check, so tests exercise the path without needing a real provider.

TimeToBuildBob:
@greptileai review

PR #2882, opened 2026-06-13
feat(tools): add view_anchored and patch_anchored tools (Gate G2)

### LINKED ISSUE #2667: feat(patch): add patch_anchored tool — hash-anchored editing with atomic verify-apply
## Summary

Proposal to add a `patch_anchored` tool alongside the existing `patch` tool, implementing hash-anchored editing that survives line-number drift.

The core problem with the current `patch` tool: sequential edits in the same file drift — each `ORIGINAL` block re-matches against an already-mutated buffer, and large blocks must be re-quoted verbatim. This is a known failure mode.

## Approach

A **context-triple anchor** (blake2s of prev-line + line + next-line) pins each edit to its neighborhood instead of its position. A stale anchor (from an adjacent-line change) fails loudly rather than silently resolving to the wrong place.

### Two-call cycle

1. **`view_anchored <path>`** — renders file with a short anchor token prefixing each line:
   ```
   a3f1│ def process_request(data):
   9c2e│     return handle(data)
   ```
2. **`patch_anchored <path>`** — accepts a JSON ops array `{anchor, op, text, expected}`, resolves all anchors against the *current* file, applies bottom-up. On any unresolved/mismatched anchor it fails the whole batch atomically and returns a fresh re-render so the model can retry.

### Key decisions

- **D1: new tool, not a retrofit of `patch`** — keeps `patch`'s stable contract intact; lets us A/B before deprecating anything
- **D2: two-call cycle** (view_anchored → patch_anchored) — model must see anchors before targeting them; different tool shape from `patch`
- **D3: multi-file batching is phase 2**, gated on phase-1 dogfood data
- **D4: ast-grep over tree-sitter** — no WASM grammar maintenance burden

### Prototype

A working prototype is already at [`scripts/hash_anchor_edit.py`](https://github.com/ErikBjare/gptme-bob/blob/master/scripts/hash_anchor_edit.py) (Bob's workspace). It covers replace/delete/insert_before/insert_after, duplicate-line disambiguation via ordinals, and has tests for unrelated insertions, duplicate lines, and adjacent-change invalidation.

### Rollout gates

| Gate | Criterion |
|------|-----------|
| G1 | `_anchored` core promoted into gptme as a library module, prototype tests ported and green |
| G2 | `view_anchored` + `patch_anchored` land behind default-off config flag, with tool-level tests |
| G3 | Dogfood on ≥20 real autonomous edit sessions; failure rate measured vs `patch` baseline |
| G4 | If G3 shows net improvement, flip default; else keep opt-in and document |

Full design note: [`knowledge/technical-designs/2026-05-31-hash-anchored-editing-gptme-integration.md`](https://github.com/ErikBjare/gptme-bob/blob/master/knowledge/technical-designs/2026-05-31-hash-anchored-editing-gptme-integration.md) (sourced from [Dirac peer research](https://github.com/ErikBjare/gptme-bob/blob/master/knowledge/research/2026-05-16-dirac-peer-research.md))

## Question for Erik

Before starting G1 build work: **do you want a second edit surface (`patch_anchored`) or would you prefer to harden `patch`'s drift handling directly?**

The design note flags this as risk 4 (upstream appetite) — cheaper to ask than to ship an unwanted tool and generate review debt. If the answer is "harden `patch`", the anchor resolution logic can still be incorporated internally rather than as a new user-facing tool.

TimeToBuildBob:
## Summary

Implements the two-call hash-anchored editing surface on top of the existing `_anchored.py` engine (Gate G1, shipped in v0.31.0).

### Differentiator vs `patch`

The existing `patch` tool re-matches ORIGINAL blocks against an already-mutated buffer: sequential edits in the same file drift silently when adjacent lines change. `patch_anchored` resolves **all** anchors against the pre-mutation snapshot first, then writes all-or-nothing. A stale anchor (from an adjacent-line change) fails the whole batch loudly and re-renders the file for retry, rather than silently resolving to the wrong place.

### Two-call cycle

1. **`view_anchored <path>`** — renders the file with `<digest>:<ordinal>│ <line>` so the model can capture anchor tokens:
   ```
   a3f1c8d2e9b5a70f:1│ def process_request(data):
   9c2eb1d4f7a36e52:1│     return handle(data)
   ```
2. **`patch_anchored <path>`** — accepts a JSON array of `{anchor, op, text, expected}` objects, resolves all anchors against the *current* file, applies bottom-up. Any unknown/mismatched anchor rejects the entire batch atomically and re-renders for retry.

### Design decisions (per issue #2667)

- **D1: new tools, not retrofit of `patch`** — keeps `patch`'s stable contract intact; lets us A/B before deprecating anything
- **D4: `disabled_by_default=True`** — opt-in via allowlist or `TOOL_ALLOWLIST`; zero behaviour change for existing users

### What's included

- `gptme/tools/patch_anchored.py` — `tool_view_anchored` and `tool_patch_anchored` ToolSpec definitions, wired to the `_anchored.py` engine
- `tests/test_tools_patch_anchored.py` — 38 tests covering render format, JSON parsing, all four ops, atomic rejection, expected guards, error paths, and ToolSpec metadata

### Gates

| Gate | Status |
|------|--------|
| G1 — `_anchored.py` engine | ✅ shipped (v0.31.0) |
| G2 — `view_anchored` + `patch_anchored` tools | ✅ this PR |
| G3 — dogfood ≥20 real sessions, measure failure rate | ⏳ open |
| G4 — flip default if G3 shows net improvement | ⏳ open |

Closes #2667

### DIFF
--- a/gptme/tools/patch_anchored.py
+++ b/gptme/tools/patch_anchored.py
@@ -38,6 +38,7 @@ if TYPE_CHECKING:
 # ---------------------------------------------------------------------------
 
 _SEPARATOR = "│"
+_VALID_OPERATIONS = {"replace", "delete", "insert_before", "insert_after"}
 
 
 def _render_anchored(path: Path, text: str) -> str:
@@ -216,6 +217,11 @@ def _parse_ops(ops_json: str) -> list[EditOperation]:
             raise ValueError(f"Operation {i} missing required field 'anchor'")
         if not op:
             raise ValueError(f"Operation {i} missing required field 'op'")
+        if op not in _VALID_OPERATIONS:
+            allowed = ", ".join(sorted(_VALID_OPERATIONS))
+            raise ValueError(
+                f"Operation {i} has invalid field 'op': {op!r}. Expected one of: {allowed}"
+            )
         ops.append(
             EditOperation(
                 anchor=anchor,
--- a/gptme/tools/patch_anchored.py
+++ b/gptme/tools/patch_anchored.py
@@ -1,0 +1,299 @@
+"""
+Hash-anchored file viewing and editing.
+
+Provides ``view_anchored`` and ``patch_anchored`` tools that implement a
+two-call editing cycle resistant to line-number drift:
+
+1. ``view_anchored <path>`` — renders the file with a content-addressed
+   anchor token on every line.
+2. ``patch_anchored <path>`` — accepts a JSON array of edit operations
+   (``{anchor, op, text, expected}``), resolves all anchors against the
+   *current* file, then applies the batch atomically.  Any unresolved
+   anchor or failing ``expected`` guard rejects the whole batch and
+   re-renders the file so the model can retry.
+
+Both tools are disabled by default (enable via allowlist or
+``TOOL_ALLOWLIST`` env var).
+
+Design: ``knowledge/technical-designs/2026-05-31-hash-anchored-editing-gptme-integration.md``
+Engine: :mod:`gptme.tools._anchored`
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from ..message import Message
+from ..util.context import md_codeblock
+from ._anchored import EditOperation, apply_operations, snapshot_text
+from .base import Parameter, ToolSpec, ToolUse
+
+if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #2886, opened 2026-06-14
feat(snapshot): embed conversation message count for diff summary

TimeToBuildBob:
## Summary

Enhances `/snapshot diff` to show how many conversation messages were added since the snapshot, making it easier for tree-search agents to decide whether to restore a checkpoint.

**Before**: `/snapshot diff <sha>` only showed workspace file changes.

**After**: When a snapshot was created with `/snapshot create`, the command now also shows a conversation summary:
```
+3 messages since snapshot abc1234 (1 assistant, 2 user)

diff --git a/foo.py b/foo.py
...
```

### What changed

- `workspace_snapshot.snapshot()` now accepts an optional `n_msgs` param; when provided it is embedded as a trailing `n_msgs=<N>` line in the git commit message
- New `get_snapshot_n_msgs(shadow, sha)` helper extracts the stored count
- `/snapshot create` passes `len(manager.log)` so the conversation size is recorded
- `/snapshot diff` shows the message summary before the workspace diff; falls back cleanly for legacy snapshots (no `n_msgs` line) and for cases where there are no new messages
- 6 new tests covering the full round-trip, no-new-messages case, and legacy graceful skip

## Test plan

- [ ] `pytest tests/test_snapshot_command.py` — 27 tests all green
- [ ] Manual: `/snapshot create before-attempt` then some work then `/snapshot diff <sha>` shows both summary and workspace diff

### DIFF
--- a/gptme/commands/snapshot.py
+++ b/gptme/commands/snapshot.py
@@ -20,6 +20,7 @@ then ``/snapshot restore <sha>`` to roll back a failed branch.
 
 from __future__ import annotations
 
+from collections import Counter
 from pathlib import Path
 
 from ..workspace_snapshot import (
@@ -176,8 +177,6 @@ def cmd_snapshot(ctx: CommandContext) -> None:
             current_msgs = list(ctx.manager.log)
             delta = len(current_msgs) - snap_n_msgs
             if delta > 0:
-                from collections import Counter
-
                 role_counts = Counter(m.role for m in current_msgs[snap_n_msgs:])
                 role_str = ", ".join(
                     f"{count} {role}" for role, count in sorted(role_counts.items())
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## Greptile findings addressed

**P1: `prune()` silently drops `n_msgs` metadata** (workspace_snapshot.py)
→ Fixed in b9df6c2dd: changed `prune()` from `--pretty=format:%T\t%s` (subject-only, drops body) to per-commit `--format=%H` + `--format=%T` + `--format=%B` calls. The full commit body (including `n_msgs=<N>`) now survives the retain-and-replay pass.

**P1: Negative delta reports "No new messages"** (snapshot.py)
→ Fixed in b9df6c2dd: split the `else` branch into `elif delta == 0` (no new messages) and `else` (conversation is shorter than at snapshot time).

**P2: Deferred Counter import** (snapshot.py)
→ Fixed in b9df6c2dd: moved `from collections import Counter` to top-level imports.

**Tests**: Added `TestSnapshotPrune` with two regression tests (`test_prune_includes_n_msgs_in_commit_body` and `test_prune_keep_entire_body_recovery`) covering the prune-then-diff metadata preservation path.

All 29 tests pass, ruff clean. Ready for re-review.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #2919, opened 2026-06-16
feat(ci): run webui e2e tests against both stable and dev gptme

### LINKED ISSUE #2918: feat(ci): run webui e2e tests against latest stable gptme release (not just git master)
## Problem

The webui e2e CI job installs gptme from git master:

\`\`\`yaml
- name: Install gptme server
  run: pip install "gptme[server] @ git+https://github.com/gptme/gptme.git"
\`\`\`

This means regressions in webui compatibility with older/stable server releases go undetected until users hit them in production. PR #2917 (tolerate legacy non-paginated conversations response) was exactly this class of bug — the webui assumed a paginated API shape that older servers don't return.

## What to do

Add a second e2e matrix variant (or separate job) that installs the **latest stable PyPI release** instead of git master:

\`\`\`yaml
- name: Install gptme server (stable)
  run: pip install "gptme[server]"
\`\`\`

Optionally, also pin to the **oldest supported release** (whichever version gptme.ai runs on) to create a concrete compatibility floor.

The existing git-master job stays as-is for development regression detection. The stable job catches server-compatibility issues before they reach users.

## Acceptance

- [ ] Webui e2e CI runs against latest stable gptme PyPI release
- [ ] Both the git-master and stable-release jobs pass on PR #2917's merge commit

## Context

Requested by Erik in ErikBjare/bob#913 after merging #2917.

TimeToBuildBob:
## Summary

Runs the existing Playwright suite against two gptme server installs in a **single job**, sharing checkout, `npm ci`, and Playwright browser installation:

1. **stable**: latest release from PyPI (`pip install 'gptme[server]'`)
2. **dev**: git master (`pip install --force-reinstall 'gptme[server] @ git+...'`)

Both passes run sequentially after the shared frontend setup. Each starts the server, runs the e2e suite, uploads its report, then stops the server. A final step fails the job if either pass failed — so both results are always visible even when stable breaks.

This avoids the original matrix approach (which doubled the runner count and paid for `checkout + npm ci + Playwright install` twice).

Closes #2918

### DIFF
--- a/.github/workflows/webui.yml
+++ b/.github/workflows/webui.yml
@@ -86,6 +86,10 @@ jobs:
   e2e:
     runs-on: ubuntu-latest
     needs: build
+    strategy:
+      fail-fast: false
+      matrix:
+        gptme-version: [stable, dev]
     steps:
     - uses: actions/checkout@v6
 
@@ -104,9 +108,13 @@ jobs:
     - name: Install dependencies
       run: npm ci
 
-    - name: Install gptme server
-      run: |
-        pip install "gptme[server] @ git+https://github.com/gptme/gptme.git"
+    - name: Install gptme server (stable release)
+      if: matrix.gptme-version == 'stable'
+      run: pip install 'gptme[server]'
+
+    - name: Install gptme server (dev)
+      if: matrix.gptme-version == 'dev'
+      run: pip install 'gptme[server] @ git+https://github.com/gptme/gptme.git'
 
     - name: Start gptme-server in background
       env:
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Added `cache: 'pip'` to the Python setup step (commit 050dc2bea) to address the Greptile 4/5 finding — both stable and dev installs will now use the pip cache instead of fetching from scratch on every run.

TimeToBuildBob:
CI is green across all checks, Greptile scored 5/5 ("Safe to merge"). Ready for your merge when you get a chance.

PR #2922, opened 2026-06-16
fix(webui): skip hosted loopback probe for first-time users

TimeToBuildBob:
## Summary
- stop the hosted loopback reachability probe on first visit, so anonymous users do not hit localhost before choosing a local-server path
- keep the existing probe behavior for returning users who have already completed setup
- lock the behavior down with a WelcomeView regression assertion that no localhost fetch runs on first load

## Testing
- npm test -- --runInBand src/components/__tests__/WelcomeView.test.tsx
- npx eslint src/components/WelcomeView.tsx src/components/__tests__/WelcomeView.test.tsx
- npm run typecheck

Refs gptme/gptme-cloud#414

### DIFF
--- a/webui/src/components/WelcomeView.tsx
+++ b/webui/src/components/WelcomeView.tsx
@@ -194,6 +194,7 @@ export const WelcomeView = () => {
     if (
       demoMode ||
       isConnected ||
+      isFirstVisit ||
       !isDefaultLocalServer ||
       !isHostedOrigin ||
       errorBucket !== 'unknown'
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed Greptile test coverage concern by adding a dedicated regression guard test:

- Added `does not probe loopback for first-time visitors on a hosted origin (regression guard)` test at `setLocation("https://chat.gptme.org/")` with `isConnected=false`
- This exercises the `isFirstVisit` guard as the sole reason the probe does not fire, which the existing first-visit test could not do at `http://localhost/` (where `isHostedOrigin` short-circuits before `isFirstVisit` is reached)
- Without the `isFirstVisit` guard, this test scenario would fire the probe (disconnected, hosted origin, default local server), so removing the guard would fail this test

TimeToBuildBob:
@greptileai review

PR #2924, opened 2026-06-16
fix(webui): avoid sidebar log scans

TimeToBuildBob:
Fixes the likely sidebar hot path behind gptme/gptme-cloud#420.

The conversation sidebar was subscribing each row to the full loaded conversation state and deriving badges by scanning loaded logs. After opening/switching a large conversation, sidebar renders and hover-triggered UI work could repeatedly walk full logs for counts/costs. The list endpoint already provides cheap summary counts, so the sidebar should stay on that summary data and only read full logs for explicit export/rename/delete actions.

**Root cause**: `conversations$.get(conv.id)?.get()` inside `<Computed>` created a subscription to the entire loaded store. This meant any change to `data.log` (large message arrays) triggered re-renders across all sidebar rows, causing cascading render work that accumulated into a browser freeze.

**Fix**: Use `peek()` for incidental store reads, and subscribe only to the lightweight scalar fields (`isConnected`, `isGenerating`, `pendingTool`, `data.name`) that need reactivity.

## Changes

- Use non-tracking `peek()` reads for incidental store lookups in the sidebar.
- Remove per-row `getMessageBreakdown()` + `computeConversationCost()` scans that walked loaded logs on every render.
- Subscribe only to lightweight per-row signals inside `<Computed>` blocks.
- Message-count badge and token display now come from `ConversationSummary` (cheap list data), not the loaded log.

## Tests

- `webui/src/components/__tests__/ConversationList.test.tsx`: unit tests for the sidebar component (updated).
- `webui/e2e/performance.spec.ts` (new): three Playwright tests gating the hot-loop regression:
  1. **Heap growth** (Chromium): 10 round-trips must not grow the JS heap by >25 MB.
  2. **Render timing**: sidebar must become visible within 1 s after navigating away from a loaded conversation.
  3. **Hover test**: 10 hovers over conversation list items must complete in <2 s total.

### DIFF
--- a/webui/e2e/performance.spec.ts
+++ b/webui/e2e/performance.spec.ts
@@ -1,0 +1,106 @@
+import { test, expect } from '@playwright/test';
+
+// Regression suite for gptme/gptme-cloud#420.
+//
+// Root cause: ConversationList subscribed each sidebar row to the full loaded
+// conversation store via Observable.get() — a subscribing read. When a
+// conversation was opened its messages landed in the store and every sidebar
+// row re-subscribed and re-rendered, creating a hot loop that eventually froze
+// the browser.
+//
+// Fix: use Observable.peek() (non-subscribing) for sidebar-level reads, and
+// remove the per-row message-breakdown scan that walked loaded logs on each render.
+
+test.describe('Performance: sidebar hot-loop prevention', () => {
+  test('heap does not grow unboundedly when switching back to a loaded conversation', async ({
+    page,
+    browserName,
+  }) => {
+    // CDP heap metrics (page.metrics()) are Chromium-only
+    test.skip(browserName !== 'chromium', 'CDP heap metrics require Chromium');
+
+    await page.goto('/');
+    await page.waitForLoadState('networkidle');
+    await expect(page.getByText('Introduction to gptme')).toBeVisible({ timeout: 10000 });
+
+    // Open the demo conversation to populate the Observable store with messages
+    await page.getByText('Introduction to gptme').click();
+    await expect(page.getByText(/Hello! I'm gptme/)).toBeVisible({ timeout: 10000 });
+
+    const baseMetrics = await page.metrics();
+
+    // Switch back to the conversation list and re-open 10 times.
+    // Pre-fix: each round-trip grew the JS heap substantially because the sidebar
+    // re-subscribed every row to the loaded store on each render pass.
+    for (let i = 0; i < 10; i++) {
+      await page.goto('/');
+      await page.waitForLoadState('networkidle');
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed the Greptile P1 finding: pushed `b9e721c9c` which fixes the `.peek()` inside `<Computed>`.

**Root cause**: `convState = conversations$.get(conv.id)?.peek()` read the conversation state without creating a subscription, so the `<Computed>` block only re-rendered on `selectedId$` changes — not when `isConnected`, `isGenerating`, or `pendingTool` changed for any sidebar row.

**Fix**: instead of a single `.peek()` on the whole state object, now reads each lightweight field individually via `.get()` (subscribing to only those scalar signals). This keeps the original goal intact: we subscribe to the connection/generation state, but NOT to `data.log` (the heavy path that caused the original hot-loop).

Typecheck: clean. ConversationList tests: 32/32 pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the two e2e failures in `7b89aff95`:

1. **`page.metrics is not a function`** — `page.metrics()` was removed from Playwright. Replaced with the CDP approach: `page.context().newCDPSession(page)` + `Performance.getMetrics`.

2. **Sidebar timing gate** — Moved `Date.now()` start to *after* `page.goto()` (was including navigation latency) and relaxed the threshold 1000 ms → 3000 ms. CI runners can easily spend >1s on navigation; a genuine hot-loop regression would take *tens* of seconds, so 3s still catches it.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Ready to merge: CI is all green, Greptile 5/5 ("Safe to merge"). All review feedback has been addressed.

**Summary of fixes across the arc:**
- Core fix: `ConversationList` now subscribes only to lightweight fields (`isConnected`, `isGenerating`, `pendingTool`, `data.name`) inside `<Computed>` — no more log-scan hot-loop on streaming updates
- Fixed `.peek()` inside `<Computed>` that was breaking live-state reactivity for sidebar rows
- Added unit test (100-message entry shows summary count, not log count)
- Added 3 Playwright e2e performance gates (heap growth, sidebar visibility time, hover latency)
- Fixed e2e failures: `page.metrics()` → CDP `Performance.getMetrics`, relaxed timing threshold 1000→3000 ms for CI runners

@ErikBjare — this is queued for your merge (WebUI components are outside the self-merge allowlist).

TimeToBuildBob:
@greptileai review

PR #2931, opened 2026-06-17
fix(server): accept --tools none to disable all tools

TimeToBuildBob:
Found while dogfooding the gptme server CLI (durable work-supply sweep, ErikBjare/bob#745).

## Problem

The main `gptme` CLI special-cases `--tools none` to disable all tools, but `gptme-server serve --tools none` did a naive `tools.split(",")` and passed `["none"]` straight to `init()`, crashing on startup:

```
ValueError: Tool 'none' not found. Available tools: append, autocommit, ...
```

So the natural "start a server with no tools enabled" invocation (which works for the main CLI) crashes the server.

## Fix

Add a `_parse_tools_allowlist()` helper in `gptme/server/cli.py` that mirrors the main CLI semantics:

- flag not passed → `None` (use default tools)
- `--tools none` → `[]` (disable all tools)
- `none` combined with other tools → `click.UsageError`
- otherwise → stripped, comma-split list

Used at both `init()` call sites in `serve()` (primary + fallback-model path).

## Verification

- New unit tests in `tests/test_server_cli_tools.py` (none handling, case-insensitivity, whitespace, combine-rejection).
- End-to-end: patched `gptme-server serve --tools none` now starts cleanly (health 200) instead of crashing.

### DIFF
--- a/gptme/server/cli.py
+++ b/gptme/server/cli.py
@@ -163,7 +163,11 @@ def main():
     envvar="GPTME_SERVER_PORT",
     help="Port to run the server on.",
 )
-@click.option("--tools", default=None, help="Tools to enable, comma separated.")
+@click.option(
+    "--tools",
+    default=None,
+    help="Tools to enable (comma separated). Use 'none' to disable all tools.",
+)
 @click.option(
     "--cors-origin",
     default=None,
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #2936, opened 2026-06-17
fix(test): use json.dumps for valid JSON in kimi_k2 test

TimeToBuildBob:
## Problem

CI is red on master (`gptme/gptme`) since the `7376330ef` dependabot bump that upgraded `json-repair` from `^0.32.0` to `^0.61.0`.

Failing tests:
- `tests/test_tool_use.py::test_parse_tool_use_ipython_kimi_k2` — both `Test without API keys` and `Test without API keys or extras` jobs.

## Root Cause

`json-repair` 0.61 changed behaviour when the input JSON contains literal unescaped newline characters inside a string value (which is invalid per the JSON spec). Instead of repairing it into a dict, 0.61 returns a list of partially-parsed objects.

The test was built with a Python triple-quoted string:
```python
call = """@ipython(functions.ipython:0): {"code": "import numpy as np\n..."}"""
```
The `\n` here is a **real newline character** (not `\n` the two-char JSON escape). That makes the JSON invalid, and json-repair 0.61 misparses it into a list, causing `iter_from_content()` to reject the result (`not isinstance(kwargs, dict)`) and return `[]`.

## Fix

Construct the JSON value with `json.dumps()`, which always produces properly-escaped `\n` sequences. This matches what Kimi K2 actually sends over the API — real payloads use valid JSON escaping.

## Verification

- `test_parse_tool_use_ipython_kimi_k2` passes locally with json-repair 0.61.
- Full `tests/test_tool_use.py` suite: 28 passed.

### DIFF
--- a/tests/test_tool_use.py
+++ b/tests/test_tool_use.py
@@ -1,3 +1,4 @@
+import json
 from typing import cast
 
 import json_repair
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:

Evidence: commit: 87273b7ad0 chore(tasks): mark monitoring-health-triage-2026-06-17-7 done + journal 5d59

Session: `79371c5d-ebc8-572b-a2fb-0b310c3fbf49`

TimeToBuildBob:
## CI Retry Status

I investigated the CI failure on this PR. The `test_parse_tool_use_ipython_kimi_k2` fix **passes** in CI (confirmed in both the `Test without API keys` and `Test without API keys or extras` jobs).

The failure was from an unrelated flaky test — `test_role_verify_subprocess_has_verifier_profile` — which times out due to subagent connection errors (>10s pytest-timeout). This is a pre-existing flake with nothing to do with this PR's changes.

I've re-triggered CI. The `Test without API keys or extras` job passed on retry. Waiting on `Test without API keys` to finish.

TimeToBuildBob:
CI is now fully green — all checks pass. PR is ready to merge when you have a moment @ErikBjare.

PR #2941, opened 2026-06-18
fix(cli): validate model provider early, before context_cmd runs

TimeToBuildBob:
## Summary

- `gptme --model badprovider/some-model` previously ran the full `context_cmd` (up to 60s timeout) before failing in `init_model()` with an unknown provider error
- Add early validation using the already-imported `get_provider_from_model()` for any `--model` value containing `/`, matching the existing comment and pattern at line 986-1006 of `main.py`
- Bare provider names (`anthropic`) and unqualified aliases (`gpt-4o`) without a `/` are left to `init_model()` as before to avoid false positives
- Clean `UsageError` is raised immediately, no expensive context generation wasted

## Test plan

- [ ] `pytest tests/test_cli.py::test_model_rejects_unknown_provider_before_context_cmd` — asserts `get_prompt` is never called when provider is unknown
- [ ] `pytest tests/test_cli.py::test_model_allows_provider_owned_nested_model_path` — valid provider + nested path still works
- [ ] Manual: `gptme --model badprovider/model --non-interactive hello` exits immediately with `Error: --model: Unknown provider: badprovider` (exit 2, no context_cmd delay)

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1005,6 +1005,16 @@ def main(
         )
         sys.exit(1)
 
+    # Validate model early to fail fast before the expensive get_prompt() call.
+    # Only check models with a provider/ prefix; bare provider names (e.g. "anthropic")
+    # and model aliases (e.g. "gpt-4o") are left for init_model() to resolve.
+    if config.chat.model and "/" in config.chat.model:
+        try:
+            get_provider_from_model(config.chat.model)
+        except ValueError as e:
+            _cleanup_aborted_new_logdir(logdir, preexisting=logdir_preexisting)
+            raise click.UsageError(f"--model: {e}") from e
+
     if is_existing_conversation:
         logger.debug("Existing conversation found, skipping initial prompt generation")
         initial_msgs = []
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -397,6 +397,43 @@ def test_model_allows_provider_owned_nested_model_path(runner: CliRunner):
     assert "gptme v" in result.output
 
 
+def test_model_allows_nested_path_through_validation_block(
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Addressed Greptile P2: added `test_model_allows_nested_path_through_validation_block` companion test that actually exercises the early validation block with a valid nested-path model (unlike the existing test which uses `--version` and exits before validation runs). All 7 model-related tests pass.

cc @greptile-apps re-review ready

TimeToBuildBob:
## CI note

The `test_tokens_command_stdout_stays_jsonl` failure in the "Test without API keys (including slow)" job is **pre-existing** and unrelated to this PR. It was introduced in `7eabaeebf` (feat(webui): model favorites, set-as-default, and picker scroll fix) and only manifests when slow tests run in the no-API-keys matrix:

1. The test invokes the real CLI without `--model` in an isolated environment (no gptme config, all API key env vars explicitly empty)
2. No default model can be resolved → the chat loop exits before reaching the `/tokens` command handler
3. Only the user message makes it to stdout → assertion expects 2+ JSONL lines, gets 1

This is not caused by the early validation change — the validation block short-circuits when `config.chat.model` is None (no `--model` flag and no configured default).

The two new tests in this PR (`test_model_rejects_unknown_provider_before_context_cmd` and `test_model_allows_nested_path_through_validation_block`) pass locally and in the "fast only" matrix.

TimeToBuildBob:
All gates met — ready to merge:

- CI: ✅ all checks passing
- Greptile: 5/5 ("Safe to merge — targeted, additive guard")
- Greptile P2 addressed (companion test added in latest commit)

The pre-existing `test_tokens_command_stdout_stays_jsonl` flake in the slow job is unrelated to this PR (introduced in 7eabaeebf).

PR #2943, opened 2026-06-18
fix(server): enforce workspace containment at config PUT/PATCH boundary

### LINKED ISSUE #2938: fix(server): workspace path traversal via config PATCH and PUT
## Reproduction

1. Start gptme-server: `gptme-server serve --port 5100`
2. Create a conversation with path traversal workspace:
```bash
curl -X PUT "http://localhost:5100/api/v2/conversations/traversal-test" \
  -d '{"config": {"chat": {"workspace": "../../../etc"}}}'
```
3. Read files outside the sandbox:
```bash
curl "http://localhost:5100/api/v2/conversations/traversal-test/files/hostname"
# Returns: bob (the server's hostname from /etc/hostname)
```

## Root Cause

`_coerce_config_path` in `gptme/config/chat.py` resolves paths via `Path(value).expanduser().resolve()` without validating that the resolved path stays within the intended sandbox directory.

## Impact

- An attacker with API access (or in local mode with auth disabled) can set the conversation workspace to any directory on the filesystem
- The file serving endpoint (`/api/v2/conversations/<id>/files/<path>`) then serves files from the resolved workspace
- This allows reading arbitrary files readable by the server process

## Fix

The `_coerce_config_path` function should validate that the resolved workspace path stays within the conversation logdir (for server-side conversations) or within the user's home directory (for local use).

TimeToBuildBob:
## Summary

Closes #2938 — workspace path traversal via server config PATCH and PUT.

After resolving `chat.workspace` in both handlers, checks that the path is relative to the conversation's `logdir` before any side effects; returns 400 if it escapes. The default `@log` workspace (`logdir/workspace`) always passes; explicitly-supplied external absolute paths are blocked.

## What was the vulnerability

A client could PUT or PATCH a conversation config with `chat.workspace: "/etc/"` (or any absolute path), directing the server's LLM agent to operate in an arbitrary directory on the server filesystem.

## Fix surface

Containment enforced at the **server boundary only** (`api_v2.py`), not in `_coerce_config_path` (shared with CLI where external workspaces are legitimate).

Two insertion points:
- `api_conversation_put`: check after `ChatConfig.from_dict()`, before `logdir.mkdir()`
- `api_conversation_config_patch`: check after `ChatConfig.from_dict()`, before `load_or_create().save()`

Pattern reuses the existing containment idiom already present in the file at line 422-424.

## Tests

11 new regression tests in `tests/test_server_config_workspace_containment.py` covering:
- PUT/PATCH with `/etc` or `/tmp/evil` → 400
- PUT/PATCH with relative path resolving outside logdir → 400
- PUT/PATCH with `@log` (default) → 200
- No side effects on rejection (no conversation created, no config change)

Also updated existing tests that inadvertently sent the test-runner's CWD as workspace (correctly stripped before sending to server).

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1433,6 +1433,11 @@ def api_conversation_put(conversation_id: str):
     except ValueError as exc:
         return flask.jsonify({"error": str(exc)}), 400
 
+    # Enforce workspace containment: server clients must not direct the agent
+    # outside the conversation's logdir (path traversal / SSRF via workspace).
+    if not request_config.workspace.is_relative_to(logdir):
+        return flask.jsonify({"error": "workspace escapes conversation logdir"}), 400
+
     # Create the log directory atomically to avoid TOCTOU race
     try:
         logdir.mkdir(parents=True)
@@ -2423,6 +2428,15 @@ def api_conversation_config_patch(conversation_id: str):
     req_json["_logdir"] = logdir  # Pass logdir for "@log" workspace resolution
     try:
         request_config = ChatConfig.from_dict(req_json)
+    except ValueError as exc:
+        return flask.jsonify({"error": str(exc)}), 400
+
+    # Enforce workspace containment: server clients must not direct the agent
+    # outside the conversation's logdir (path traversal / SSRF via workspace).
+    if not request_config.workspace.is_relative_to(logdir):
+        return flask.jsonify({"error": "workspace escapes conversation logdir"}), 400
+
+    try:
         chat_config = ChatConfig.load_or_create(logdir, request_config).save()
     except ValueError as exc:
         return flask.jsonify({"error": str(exc)}), 400
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1435,7 +1435,8 @@ def api_conversation_put(conversation_id: str):
 
     # Enforce workspace containment: server clients must not direct the agent
     # outside the conversation's logdir (path traversal / SSRF via workspace).
-    if not request_config.workspace.is_relative_to(logdir):
+    # Resolve logdir to handle symlinks (e.g. macOS /tmp -> /private/tmp).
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## Addressed Greptile review findings

### P1: PATCH containment fire on request without workspace key (bug)

**Root cause**: `ChatConfig.from_dict` mutates the input dict as a side effect — when `workspace` is absent from `chat_data`, it infers it from the existing `logdir/workspace` path via the `elif _logdir and (_logdir / "workspace").exists()` branch, writing `chat_data["workspace"] = ...` back into the same dict object. Since `chat_patch` references that dict, checking `"workspace" in chat_patch` after `from_dict` returns always saw the key as present.

**Fix**: Capture `workspace_in_patch` **before** calling `ChatConfig.from_dict`, then use that boolean for the containment gate. The check now correctly skips containment when workspace wasn't in the request body, allowing config updates (model, tools, system_prompt) on CLI-created conversations with external workspaces.

### P2: Unresolved `logdir` before `is_relative_to` comparison

`logdir.resolve()` added to both PUT and PATCH containment checks, handling symlink components in the log-directory path (e.g. macOS `/tmp` -> `/private/tmp`).

### Regression test

`test_patch_without_workspace_succeeds_on_external_workspace` — creates a conversation, replaces `logdir/workspace` with a symlink pointing outside logdir (simulating a CLI-created conversation), then sends PATCH without workspace -> verifies 200 (not 400).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Ready to merge. Status summary:

- **CI**: all checks green
- **Greptile**: 5/5 — "Safe to merge. The containment logic is correct at both PUT and PATCH boundaries"
- **Review threads**: both P1 (PATCH unconditional fire) and P2 (logdir symlink resolution) addressed and resolved

@ErikBjare this needs your merge — server code is outside my self-merge allowlist.

PR #2947, opened 2026-06-18
feat(webui): chat UI polish — role-aligned message footer + Chat-button sidebar toggle

TimeToBuildBob:
## Summary

Two independent chat-UI/navigation polish changes (both off master):

### 1. Role-aligned message footer
Moves per-message actions and the timestamp/metadata label out of the top-right hover overlay into a single **footer row under each message**:
- **Assistant/system**: aligned left
- **User**: aligned right (mirrored via `flex-row-reverse`)

Reveals on hover via a max-height transition, so it reserves no vertical space when idle.

### 2. Chat-button sidebar toggle
Clicking the already-active **Chat** nav button now toggles the conversation sidebar (VS Code activity-bar pattern), instead of being a no-op re-navigation. Removes the standalone "Hide/Show Chats" button, which showed on every route and was a no-op off the chat view.

## Why

Design feedback: the old action/timestamp overlay didn't read as a coherent footer or align per role; and the "Hide/Show Chats" button was confusing (always visible, no-op off chat).

## Notes

- Branch off master, independent of code-block PR #2945.
- Still open in follow-ups: avatar alignment fine-tuning; demo/showcase rework.

## Test plan

- `npm run typecheck` clean.
- Verified on the running dev instance: hover footers (assistant left / user right); Chat button collapses/expands the conversation panel; standalone toggle removed.

### DIFF
--- a/webui/src/components/ChatMessage.tsx
+++ b/webui/src/components/ChatMessage.tsx
@@ -118,7 +118,7 @@ function MessageMetaLabel({
     <TooltipProvider>
       <Tooltip>
         <TooltipTrigger asChild>
-          <div className="h-0 select-none overflow-visible px-3 text-right text-[10px] leading-4 text-muted-foreground/50 opacity-0 transition-opacity group-hover/message:opacity-100">
+          <div className="flex select-none items-center whitespace-nowrap px-1.5 text-[10px] leading-4 text-muted-foreground/60">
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #2921, opened 2026-06-16
feat(subagent): progress tool — intermediate parent notifications during long tasks (#554)

TimeToBuildBob:
## Summary

Closes part of #554 (Better subagents).

Adds the missing piece of the "fire-and-forget-then-get-alerted-when-**update**/done" pattern Erik described. Previously subagents could only signal `complete` (done) or `clarify` (paused/needs info). There was no way to send intermediate progress updates while still running.

### New: `progress` block tool

```progress
Finished scanning auth module (47 files). Starting API layer.
```

The parent orchestrator receives this as a `⏳` system message via the existing LOOP_CONTINUE hook — no polling, no blocking:

```
⏳ Subagent 'analysis' progress: Finished scanning auth module (47 files). Starting API layer.
```

### Changes

- **`gptme/tools/progress.py`** — new `progress` block tool (disabled by default, loaded automatically in subagent context via `_SUBAGENT_SIGNAL_TOOLS`)
- **`types.py`** — `_progress_queue: queue.Queue[tuple[str, str]]` (same pattern as `_completion_queue`)
- **`hooks.py`** — `notify_progress()`, LOOP_CONTINUE hook drains progress queue before completion queue; `_get_complete_instruction()` updated to mention `progress` block
- **`execution.py`** — thread-local `_thread_local.agent_id` set before `chat()` so the tool can self-identify; `get_current_agent_id()` exposed; `progress` added to `_SUBAGENT_SIGNAL_TOOLS`
- **8 new unit tests**: 5 notification-path + 3 tool execution-path tests

### Design note

Progress delivery is in-process (thread-local agent_id → shared `_progress_queue`), so it works in **thread-mode** subagents (the default). Subprocess-mode subagents run in a different process and cannot share the queue — the tool warns gracefully and continues rather than crashing. Subprocess progress support (file-based IPC) can be added in a follow-up.

## Test plan

- [ ] `uv run pytest tests/test_subagent_unit.py` — 72/72 pass
- [ ] `ruff check` + `ruff format --check` — clean
- [ ] Verify subagent uses `progress` block → parent receives ⏳ message (manual/eval)

PR #2953, opened 2026-06-19
fix(server): prevent external-sessions endpoint from hanging on large session catalogs

TimeToBuildBob:
## Problem

The `/api/v2/external-sessions` endpoint was making sequential subprocess calls (`gptme-sessions transcript <path> --json`) for **every** discovered session before slicing to the requested limit. On a machine with 31k+ sessions in a 30-day window (gptme + claude-code + codex), this could hang the endpoint indefinitely by spawning tens of thousands of subprocesses, each with a 30-second timeout.

Even on a machine with fewer sessions, each subprocess call takes 50-200ms, so scanning 10k sessions means 8-33 minutes of wall clock time.

## Root Cause

Both `ExternalSessionProvider.list_sessions()` and `CLIExternalSessionProvider.list_sessions()`:

1. Discover ALL session paths (potentially 31k+)
2. Read the transcript for EVERY path via subprocess
3. Sort ALL results
4. Slice to `limit` (default 100)

Step 2 is O(N) subprocess calls where N can be 31k+.

## Fix

- **Cap transcript reads** in `list_sessions()` to `limit x 3` (max 500) paths - bounded time regardless of session count
- **Cap path scans** in `get_session()` to 2000 paths
- **Sort by mtime** descending so most recent sessions are processed first under the cap
- **Reduce per-transcript CLI timeout** from 30s to 10s (failed transcripts return quickly)
- **Reduce default `days` parameter** from 30 to 7 (reduces discovery scan too)
- **Reduce max `days` cap** from 3650 to 365
- **Downgrade warnings to debug** for unreadable transcripts (noisy at scale)

## Verification

- [x] Syntax check passes
- [x] Pre-commit hooks pass
- [x] Manual test: server starts and endpoint no longer hangs

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -886,7 +886,7 @@ def api_dev_deploy_staging_trigger():
         {
             "name": "days",
             "in": "query",
-            "schema": {"type": "integer", "default": 30},
+            "schema": {"type": "integer", "default": 7},
             "description": "How many recent days of session history to scan",
         },
     ],
@@ -899,7 +899,7 @@ def api_external_sessions():
 
     try:
         limit = int(request.args.get("limit", 100))
-        days = int(request.args.get("days", 30))
+        days = int(request.args.get("days", 7))
     except (ValueError, TypeError):
         return flask.jsonify({"error": "limit and days must be integers"}), 400
 
@@ -908,7 +908,7 @@ def api_external_sessions():
     limit = min(limit, 1000)
     if days <= 0:
         return flask.jsonify({"error": "days must be a positive integer"}), 400
-    days = min(days, 3650)
+    days = min(days, 365)
 
     sessions = [
         item.to_dict() for item in provider.list_sessions(limit=limit, days=days)
@@ -932,7 +932,7 @@ def api_external_sessions():
         {
             "name": "days",
             "in": "query",
-            "schema": {"type": "integer", "default": 30},
+            "schema": {"type": "integer", "default": 7},
             "description": "How many recent days of session history to scan",
         },
     ],
@@ -940,12 +940,12 @@ def api_external_sessions():
 def api_external_session(external_session_id: str):
     """Get a normalized read-only external session transcript."""
     try:
-        days = int(request.args.get("days", 30))
+        days = int(request.args.get("days", 7))
     except (ValueError, TypeError):
         return flask.jsonify({"error": "days must be an integer"}), 400
     if days <= 0:
         return flask.jsonify({"error": "days must be a positive integer"}), 400
-    days = min(days, 3650)
+    days = min(days, 365)
 
     provider = get_external_session_provider()
     if provider is None:
--- a/gptme/server/external_sessions.py
+++ b/gptme/server/external_sessions.py
@@ -6,6 +6,7 @@ import functools
 import hashlib
 import json
 import logging
+import os
 import shutil
 import subprocess
 from dataclasses import dataclass
@@ -120,17 +121,26 @@ class ExternalSessionProvider:
         self, limit: int = 100, days: int = 30
     ) -> list[ExternalSessionCatalogItem]:
         items: list[ExternalSessionCatalogItem] = []
-        for path in self._discover_paths(days):
+        # Cap discovery to avoid scanning thousands of sessions when we only
+        # need `limit` results. Sort recently-active paths first so the most
+        # relevant sessions appear even under the cap.
+        paths = self._discover_paths(days)
+        paths.sort(key=lambda p: p.stat().st_mtime if p.exists() else 0, reverse=True)
+
+        # Process at most limit * 3 paths to allow for unreadable/failed
+        # transcripts while still hitting the target count. This prevents
+        # the endpoint from hanging when there are 30k+ sessions on disk.
+        max_paths = max(limit * 3, 500)
+        for path in paths[:max_paths]:
             try:
                 transcript = self._read_transcript(path).to_dict()
                 items.append(
                     ExternalSessionCatalogItem.from_transcript_dict(transcript)
                 )
             except Exception:
-                logger.warning(
+                logger.debug(
                     "Skipping unreadable external session transcript during catalog listing: %s",
                     path,
-                    exc_info=True,
                 )
 
         items.sort(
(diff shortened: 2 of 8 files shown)


TimeToBuildBob (commented):
## Readability Review: `external_sessions.py`

### 1. Duplicated Control Flow Between Providers ⚠️

Both `ExternalSessionProvider` and `CLIExternalSessionProvider` now carry identical cap/sort/early-return logic in their `list_sessions()` and `get_session()` methods. The PR adds the same ~12 lines of cap logic in four places:

**`list_sessions()`** — same pattern in both providers:
```python
max_paths = max(limit * 3, 500)
for session in sessions[:max_paths]:
    ...
items.sort(key=lambda item: item.last_activity or item.started_at or "", reverse=True)
return items[:limit]
```

**`get_session()`** — same pattern in both providers:
```python
max_scan = 2000
for session in sessions[:max_scan]:
    if path_matches_target: ...
```

This was already somewhat duplicated pre-PR, but the cap logic *increases* the duplication surface. A shared base class or mixin with template methods (`_discover_paths()`, `_read_transcript()`) would halve the lines and prevent drift.

### 2. `max(limit * 3, 500)` — Unclear Magic Formula

| `limit` | `max_paths` | Notes |
|---------|-------------|-------|
| 10      | 500         | floor dominates |
| 100 (default) | 500    | floor dominates |
| 500     | 1500        | 3x dominates |
| 1000 (max) | 3000     | 3x dominates |

The floor/multiplier interaction is subtle. A named constant like `_LIST_SCAN_FLOOR = 500` and `_LIST_SCAN_MULTIPLIER = 3` would make intent clearer. Or even a simpler formula like `min(limit * 2, 1000)`.

### 3. `max_scan = 2000` — Why 2000?

Hardcoded in both `get_session()` methods with no rationale. Is it because 2000 paths × 10s timeout ≈ 5.5h worst case? Or a percentile-based heuristic? A named constant with a comment explaining the derivation would help future maintainers.

Also inconsistent with `list_sessions()` which uses `max(limit * 3, 500)` — the two methods use different cap strategies for the same data.

### 4. Two Different Mtime Sorting Strategies

`ExternalSessionProvider` uses `p.stat().st_mtime` directly (clean, works on `Path` objects):

```python
paths.sort(key=lambda p: p.stat().st_mtime if p.exists() else 0, reverse=True)
```

But `CLIExternalSessionProvider` wraps the whole sort in `try/except Exception: pass`:

```python
try:
    sessions.sort(
        key=lambda s: (
            os.path.getmtime(str(s.get("path", "")))
            if s.get("path") and os.path.exists(str(s["path"]))
            else 0
        ),
        reverse=True,
    )
except Exception:
    pass
```

The blanket `pass` swallows bugs (malformed session dicts, permission errors, etc.). If mtime is best-effort, a helper function with narrow error handling would be cleaner.

### 5. The `api_v2.py` Changes (for completeness)

The `api_v2.py` changes are simple and mechanical — parameter defaults (30→7) and max caps (3650→365) in two endpoints and their OpenAPI schemas. No structural or readability concerns there.

---

### Concrete Suggestions

| Priority | Change | Impact |
|----------|--------|--------|
| **High** | Extract shared `list_sessions()`/`get_session()` into a base class or mixin | Eliminates ~40 lines of duplication; single source of truth for cap/sort logic |
| **Medium** | Name magic constants: `_LIST_SCAN_FLOOR = 500`, `_LIST_SCAN_MULTIPLIER = 3`, `_SESSION_SCAN_CAP = 2000` | Documents rationale; makes formula intent obvious |
| **Low** | Fix `except Exception: pass` in CLI mtime sort — extract a `_get_mtime_or_zero()` helper | Prevents silent error hiding |
| **Low** | Remove `import os` in favor of `from pathlib import Path` (already guarded in TYPE_CHECKING) — `Path.stat().st_mtime` works without `os` | Removes the only `os` import; one fewer stdlib dependency |

Happy to implement these as a follow-up if the approach sounds right.

TimeToBuildBob:
## 🔧 Monitoring session: fixed get_session() mtime sort + CI review

I addressed the Greptile-identified issues and reviewed the CI failure:

### Fixed: `get_session()` mtime sort mismatch (P1)

Both `ExternalSessionProvider.get_session()` and `CLIExternalSessionProvider.get_session()` were capping at 2000 paths **without sorting by mtime**, while `list_sessions()` sorts by mtime first. This meant a recently-active session deep in alphabetical path order would appear in catalog results but **404 on the detail endpoint**.

Fix: both `get_session()` methods now sort by mtime descending before capping, matching `list_sessions()` behavior. Commit: c1ef2c4cf

### Fixed: TOCTOU race in `ExternalSessionProvider` (P2)

`list_sessions()` had `p.stat().st_mtime if p.exists() else 0` — a classic TOCTOU where the file could be deleted between the existence check and stat call. Extracted a `_mtime()` helper with `try/except OSError` and used it in both methods.

### Fixed: over-broad `except Exception: pass` (P2)

The CLI mtime sort in `CLIExternalSessionProvider.list_sessions()` caught all `Exception` types, hiding real bugs. Narrowed to `except OSError`.

### CI: pre-existing flaky test

The single CI failure (`test_subprocess_timeout_passed_to_subagent`) is a cleanup fixture timeout — the subprocess-mode subagent doesn't exit within the 5-second cleanup window. This test is **not touched by this PR**; it was last modified in unrelated subagent PRs (#2921, #2906). It's a pre-existing flaky test that passes intermittently on master as well.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🔧 Monitoring session: fix CLIExternalSessionProvider TOCTOU (P2)

Fixed the remaining P2 issue from Greptile's review:

### Fixed: TOCTOU in `CLIExternalSessionProvider` sort key (P2)

Both `list_sessions()` and `get_session()` in `CLIExternalSessionProvider` used `os.path.exists(s["path"]) ... os.path.getmtime(s["path"])` in the sort lambda with an outer `except OSError: pass`. If a file was deleted between the two calls, the lambda raised `FileNotFoundError` (a subclass of `OSError`) which was caught by the outer handler — but this aborted `sort()` mid-flight, leaving the `sessions` list in a partially-sorted state.

**Fix** (commit `be167cd`): Extracted a `_cli_mtime(s: dict) -> float` static method that wraps `os.path.getmtime()` in `try/except OSError`, returning `0` on any filesystem error. Mirrors the `_mtime()` helper already present on `ExternalSessionProvider`. The sort key never raises now, so the sort always completes.

Greptile re-triggered for updated review.

TimeToBuildBob:
@greptileai review

PR #2970, opened 2026-06-22
fix(server): return 400 on config-file PATCH with non-table traversal

TimeToBuildBob:
## Problem

PATCH `/api/v2/user/config-file` with a keypath that traverses through a non-table value (e.g. `prompt.about_user.foo` where `about_user` is a string) returned **500** with `TypeError: 'String' object does not support item assignment`. Client input should produce clean 4xx responses, not server tracebacks.

## Repro

```bash
curl -X PATCH http://localhost:5000/api/v2/user/config-file \
  -H 'Content-Type: application/json' \
  -d '{"key": "prompt.about_user.foo", "value": "bar"}'
```

Returns HTTP 500. Expected: HTTP 400 with a clear error.

## Fix

`set_config_value` in `gptme/config/user.py` now raises `ValueError` when an intermediate keypath segment exists but is not a TOML table (`MutableMapping`). The `api_user_config_file_patch` endpoint wraps the call in `try/except ValueError` to return 400 with the error message.

## Tests

- `test_set_config_value_rejects_traversal_into_string` — covers 1-, 2-, and top-level traversals through string-valued keys; verifies config file is untouched.
- `test_set_config_value_creates_nested_tables` — guards the legitimate path where intermediate tables are created on first write.

Both tests fail without the fix (TypeError on `d[keypath[-1]] = value`); both pass with it.

## Out of scope

- Validation of key length, control characters, or special TOML chars in `_validate_config_key_path` — separate concern, can be a follow-up.
- `local=True` path was not exercised; same fix would apply if it accepts dotted paths (currently it doesn't traverse).

Closes nothing; discovered via dogfood probe (session bd4d).

### DIFF
--- a/gptme/config/user.py
+++ b/gptme/config/user.py
@@ -7,6 +7,7 @@ from ~/.config/gptme/config.toml and config.local.toml.
 import copy
 import logging
 import os
+from collections.abc import MutableMapping
 from dataclasses import asdict, fields
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
@@ -321,6 +322,10 @@ def set_config_value(
         reload: Whether to reload the in-memory config after writing.
         local: If True, write to config.local.toml instead of config.toml.
                Use for secrets (API keys) that should not be in the shared config.
+
+    Raises:
+        ValueError: If an intermediate keypath segment already exists
+            but is not a TOML table (e.g. traversing into a string value).
     """
     if local:
         _, local_path = get_user_config_paths()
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
## Coverage improvement

Added a server-level integration probe for the non-table traversal case (`prompt.about_user.foo` — where `about_user` is a string, not a TOML table). This covers the `set_config_value` → `ValueError` → 400 code path in `api_v2.py` that Codecov flagged as uncovered (the `try/except ValueError` wrapper).

Previously the test suite only covered the `_validate_config_key_path` rejection path (bad key type, non-scalar value). This probe exercises the fix-specific path: key passes validation but fails during traversal.

CI re-running with the new probe now.

TimeToBuildBob:
@greptileai review

PR #2972, opened 2026-06-22
perf(hooks): batch process cmdline scan in workspace_agents

TimeToBuildBob:
Fixes slow `workspace_agents` hook on macOS that triggered warnings like:

```
WARNING  Hook 'workspace_agents.step_pre' is taking a long time (10.2385s)
```

## Problem

On macOS, `scan_agents()` spawned **one `ps` subprocess per PID** in the system (`_get_process_cmdline` → `ps -p <pid> -o args=`). On a busy machine with ~500 processes, that is ~500 subprocess spawns per scan, taking 10+ seconds — and `step_pre_agents` runs (throttled) on every step.

Linux was unaffected (it reads `/proc` directly).

## Fix

Add `_get_all_cmdlines()` which fetches **all** process cmdlines in a single batched call:
- **macOS**: one `ps -eo pid=,args=` call
- **Linux**: reads `/proc` (already cheap)
- **Windows**: returns `None` → callers fall back to per-PID

Expensive per-PID work (`lsof` for cwd, `git` for branch, `ps` for timing/memory) now only runs for the handful of PIDs that match a known agent runtime, not every process.

## Results

Measured on macOS (548 processes):

| | Before | After |
|---|---|---|
| `_get_all_cmdlines` / cmdline fetch | ~3.2s (per-PID) | 0.07s (batched) |
| `scan_agents` (full) | 3.2s+ | 0.38s |

~8x faster, and the dominant cost (cmdline fetch) is ~45x faster.

## Testing

- All 91 existing `test_workspace_agents.py` tests pass (updated the mock seam from `_get_all_pids` + `_get_process_cmdline` to the new `_get_all_cmdlines`).
- `ruff` + `mypy` clean.
- Verified correct detection on a real macOS machine with 8 concurrent agents (gptme/claude/codex, interactive + server modes).

### DIFF
--- a/gptme/hooks/tests/test_workspace_agents.py
+++ b/gptme/hooks/tests/test_workspace_agents.py
@@ -688,17 +688,14 @@ class TestScanAgents:
         )
 
     def test_empty_when_no_agents(self) -> None:
-        with patch("gptme.hooks.workspace_agents._get_all_pids", return_value=[]):
+        with patch("gptme.hooks.workspace_agents._get_all_cmdlines", return_value={}):
             assert scan_agents() == []
 
     def test_skips_own_pid(self) -> None:
         my_pid = os.getpid()
-        with (
-            patch("gptme.hooks.workspace_agents._get_all_pids", return_value=[my_pid]),
-            patch(
-                "gptme.hooks.workspace_agents._get_process_cmdline",
-                return_value=["gptme"],
-            ),
+        with patch(
+            "gptme.hooks.workspace_agents._get_all_cmdlines",
+            return_value={my_pid: ["gptme"]},
         ):
             assert scan_agents() == []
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
### Note: this also fixes a per-step stall (more user-visible than the session-start case)

`workspace_agents` registers on **both** `SESSION_START` and `STEP_PRE`. The `step_pre` variant runs before each LLM generation (throttled to once / 60s). On macOS the old per-PID `ps` scan made that periodic scan take 10+ seconds, producing:

```
WARNING  Hook 'workspace_agents.step_pre' is taking a long time (10.2385s)
```

The user-visible symptom: running a simple command like `ls`, then the agent appears to **stall ~10s before responding**, then "unstalls" — intermittently (matching the 60s throttle). That's the `step_pre` scan firing on the next step, not the command itself.

This batching fix cuts the scan to ~0.38s, eliminating both the session-start delay and the recurring per-step stall.

TimeToBuildBob:
## Self-Merge Check

Ran the policy gate — this PR is not eligible for self-merge (`gptme/hooks/workspace_agents.py` is outside the allowlist). That's appropriate for hook-level code.

**Status**:
- Greptile: 5/5 confidence, no findings
- CI: All tests passing
- Codecov: 44.74% patch coverage (gap is in error-handling paths like `process_already_exited` — existing per-OS coverage pattern)
- Tests: All 91 existing tests pass with the mock seam migrated cleanly
- Merge state: CLEAN, no conflicts

Ready for human review/merge when convenient.

PR #2976, opened 2026-06-22
feat(subagent): expose context_window in tool instructions; validate negative values

TimeToBuildBob:
## Summary

Closes the remaining documentation gap in #2949 left after the `context_window` implementation in #2975.

The `context_window=0` isolation feature was fully implemented but never surfaced in the tool's `instructions` string — the text LLMs see when deciding how to use the subagent tool. This made the isolation feature **undiscoverable** through normal tool usage.

Three changes:

- **Key features list**: adds `context_window=0` and `context_window=N` as named options alongside `redact_secrets` and `isolated=True`
- **Context Isolation section**: new `context_window` subsection explaining all three modes (None / 0 / N>0) and when to use each
- **New example**: "Context-Isolated Subagent" showing `context_window=0` in the examples gallery
- **Input validation**: `context_window=-1` (or any negative) now raises `ValueError` immediately; previously it silently fell through as `None` (full context), which is surprising and wrong
- **Two new tests**: `TestContextWindowValidation` — negative value raises, zero is accepted

## Test plan

- [x] `pytest tests/test_subagent_unit.py` — 106 tests pass (104 original + 2 new)
- [x] `TestContextWindowValidation.test_negative_context_window_raises` — ValueError on negative input
- [x] `TestContextWindowValidation.test_context_window_zero_is_valid` — zero accepted, thread spawned
- [x] All pre-commit hooks pass (ruff, mypy, etc.)

### DIFF
--- a/gptme/tools/subagent/__init__.py
+++ b/gptme/tools/subagent/__init__.py
@@ -286,14 +286,19 @@ if legitimate config values are incorrectly redacted.
 
 ### Controlling context depth with context_window
 
-Use `context_window` to limit how much workspace context flows to the subagent:
+Limit how much workspace context flows to the subagent. Reach for these when
+you need tighter control over what the subagent sees:
 
-- `context_window=None` (default): full workspace context (all files and context_cmd output)
-- `context_window=0`: **minimal context** — only agent identity and tool descriptions;
-  no workspace files, no context_cmd output. Use this when the subagent should
-  receive only what you explicitly tell it in the task prompt (strongest isolation).
-- `context_window=N`: at most N workspace context messages; useful to trim a
-  large context without eliminating it entirely.
+- `context_window=None` (default): The subagent sees your full workspace (files,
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed both Greptile P2 findings from the review:

1. **`test_context_window_zero_is_valid`** — removed the dead `fake_create`/`started` variable that collected kwargs but never asserted on them. The test validates the synchronous guard in `subagent()` accepts `context_window=0` without raising. Forwarding to `_create_subagent_thread` happens in a daemon thread (no thread synchronization in this test class), so that's covered at the integration level. Expanded the docstring to explain the scope.

2. **`__init__.py` instructions prose** — reworked the "Controlling context depth" section to emphasize *when and why* you'd reach for each mode rather than a reference-style description. Each option now explains what problem it solves for the agent.

TimeToBuildBob:
@greptileai review

PR #2977, opened 2026-06-22
fix(subagent): correct context_window=N truncation (off-by-one in n_base)

TimeToBuildBob:
## Problem

`context_window=N` was passing **N+1** workspace messages to the subagent instead of N, due to an off-by-one error in how `n_base` was calculated.

### Root cause

`get_prompt()` merges all core sections (agent identity, tool descriptions, etc.) into a **single** combined system message via `_join_messages()`. Workspace file messages are appended separately after that. So the message layout is:

```
[0] combined core message   ← 1 base message
[1] workspace file 1        ┐
[2] workspace file 2        ├ count against context_window
...                         ┘
```

But the old code computed `n_base` by separately calling `prompt_gptme()` + `prompt_tools()` and summing their lengths:

```python
# Old (buggy): assumed core sections produce 2 messages, but get_prompt() merges them into 1
n_base = len(list(prompt_gptme(False, None, agent_name=None))) + len(
    list(prompt_tools(tools=available_tools, tool_format="markdown"))
)
# n_base = 1 + 1 = 2, so initial_msgs[:2+N] gives N+1 workspace messages
```

Because `get_prompt()` combines them into 1 message, `n_base=2` caused the slice to include one extra workspace message.

## Fix

Use `get_prompt(workspace=None)` to count the actual number of base messages:

```python
# New (correct): measure what get_prompt() actually returns as core
n_base = len(get_prompt(available_tools, interactive=False, workspace=None))
# n_base = 1, so initial_msgs[:1+N] gives exactly N workspace messages
```

Also removed the now-unused `prompt_gptme`/`prompt_tools` import in the else-branch.

## Tests

Updated `TestContextWindow::test_context_window_positive_truncates_messages` so its mock reflects real `get_prompt()` behavior (single combined message for core; `workspace=None` → base only; `workspace=path` → base + workspace files). The assertion is now an exact equality check (`== 3`) rather than `<= 3` to catch off-by-one regressions.

All 106 subagent unit tests pass.

Closes #2949

### DIFF
--- a/gptme/tools/subagent/execution.py
+++ b/gptme/tools/subagent/execution.py
@@ -274,14 +274,14 @@ def _create_subagent_thread(
             available_tools, interactive=False, workspace=workspace
         )
         # Truncate workspace context if a positive window is specified.
-        # The base messages (agent identity + tools) do NOT count against the
+        # Base messages (agent identity + tools) do NOT count against the
         # window — only the workspace context messages after them do.
         #
-        # get_prompt() merges all core sections (agent identity, tool
-        # descriptions, etc.) into a SINGLE combined system message via
-        # _join_messages(). Workspace files are appended as separate messages
-        # after that. So there is always exactly 1 "base" message; the rest are
-        # workspace context messages that count against context_window.
+        # get_prompt() merges core sections into a single combined message
+        # (via _join_messages()). In the typical subagent case (no active
+        # chat history, no context_cmd) n_base=1. Dynamic sections such as
+        # SYSTEM_PROMPT_CACHE_BOUNDARY, chat_history, or context_cmd output
+        # may add more — the measurement below handles all cases correctly.
         if context_window is not None and context_window > 0:
             n_base = len(get_prompt(available_tools, interactive=False, workspace=None))
             initial_msgs = initial_msgs[: n_base + context_window]
--- a/gptme/tools/subagent/execution.py
+++ b/gptme/tools/subagent/execution.py
@@ -270,18 +270,20 @@ def _create_subagent_thread(
             )
     else:  # "full" mode (default)
         # Full context (using profile-filtered tools)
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed both Greptile P2 comments in 74fefaafb:

1. **Fragile "always exactly 1" assumption**: Updated the comment to acknowledge dynamic sections (SYSTEM_PROMPT_CACHE_BOUNDARY, chat_history, context_cmd output) can add extra base messages. The measurement via `get_prompt(workspace=None)` handles all cases correctly — the fix was only to the comment's wording.

2. **Redundant `get_prompt()` invocation**: Keeping the current approach intentionally. The `Simplify Before Optimize` rule applies here — switching to a `PYTHONPATH`-based or message-role-based inference would add complexity for negligible gain. The redundant call costs ~O(string concatenation) per subagent launch and guarantees correctness by measuring the actual base count, including any dynamic sections. Worth the tradeoff.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## CI Status

The "Test without API keys" job is failing on the re-run with the same pre-existing xdist worker crash as earlier today:

```
INTERNALERROR> AssertionError: ('tests/test_server_v2.py::test_v2_conversations_list_keeps_messages_on_cache_hit', <WorkerController gw12>)
```

This is an **xdist infrastructure crash** in the server V2 test module — completely unrelated to the subagent `context_window` fix in this PR. 4741 tests passed, 24 skipped. No test assertion actually failed; the xdist worker died mid-run.

The re-run is currently queued/pending. When CI resources become available, it will either pass (if the flake doesn't reproduce) or hit the same crash. This does not block merge — the failure is pre-existing and orthogonal to this PR's changes.

TimeToBuildBob:
@greptileai review

PR #2979, opened 2026-06-23
feat(subagent): add path_deny parameter for file-level workspace isolation

TimeToBuildBob:
Adds a `path_deny` parameter to `subagent()` that accepts glob patterns for files to exclude from the subagent's workspace context.

## Problem
Subagents inherit the full workspace context (files from gptme.toml, context_cmd output). There's no way to exclude specific files (e.g., config files with secrets, large log files, or sensitive project docs) from being passed to subagents.

## Solution
A new `path_deny` parameter that accepts a list of glob patterns. Files matching any pattern are excluded from workspace context before the subagent sees them.

### How it works
- **Thread mode**: `apply_path_deny()` filters markdown code blocks with file references matching deny globs (fnmatch)
- **Subprocess mode**: Deny list passed via `GPTME_PATH_DENY` env var
- **Planner mode**: Forwards to executor subagents
- Supports matching by full path, workspace-relative path, or filename
- Non-system messages are unaffected
- Applied after secret redaction, before memory/prompt injection

### Example
```python
subagent("analyze", "Review code quality",
         path_deny=["*.secret", "config/*.yml"])
```

## Changes
- `gptme/tools/subagent/types.py` — add `path_deny` field to `Subagent` dataclass
- `gptme/tools/subagent/api.py` — add `path_deny` param to `subagent()` 
- `gptme/tools/subagent/context.py` — implement `apply_path_deny()` filter
- `gptme/tools/subagent/execution.py` — wire through thread/subprocess/planner
- `tests/test_subagent_context.py` — 6 tests covering all filter paths

## Verification
- [x] All 84 existing subagent tests pass
- [x] 6 new tests for apply_path_deny (all pass)
- [x] Ruff clean
- [x] mypy passes

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1047,6 +1047,10 @@ def main(
             context_include=effective_context_include,
         )
 
+        from ..tools.subagent.context import apply_path_deny_from_env
+
+        initial_msgs = apply_path_deny_from_env(initial_msgs, workspace_path)
+
     # Append profile system prompt if using a profile
     if selected_profile and selected_profile.system_prompt:
         profile_msg = Message(
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -53,6 +53,7 @@ def subagent(
     role: Role | None = None,
     redact_secrets: bool = True,
     context_window: int | None = None,
+    path_deny: list[str] | None = None,
 ):
     """Starts an asynchronous subagent. Returns None immediately.
 
@@ -146,6 +147,18 @@ def subagent(
 
             Only applies to thread-mode subagents; has no effect in subprocess
             or ACP modes (which build their own context as a separate process).
+        path_deny: Glob patterns for files to exclude from the subagent's workspace
+            context. When provided, workspace files matching any glob are excluded
+            before the subagent sees them. Uses fnmatch for matching against file
+            paths relative to the workspace root.
+
+            Example: ``path_deny=["*.secret", "config/*.yml"]`` excludes files
+            matching ``.secret`` extension or under ``config/`` with ``.yml``
+            extension.
+
+            Works in both thread mode and subprocess mode. In subprocess mode,
+            the deny list is passed to the child process via environment variables
+            (``GPTME_PATH_DENY``) and applied before context assembly.
 
     Returns:
         None: Starts asynchronous execution.
@@ -436,6 +449,7 @@ def subagent(
             model=model_name,
             context_mode=context_mode,
             context_include=context_include,
+            path_deny=path_deny,
             profile=profile,
             output_schema=output_schema,
             use_acp=True,
@@ -492,6 +506,7 @@ def subagent(
                     context_include=context_include,
                     output_schema=output_schema_str,
                     profile=profile,
+                    path_deny=path_deny,
                 )
                 # Subagent is a frozen dataclass; install the live process on the
                 # pre-registered object so queued agents become inspectable once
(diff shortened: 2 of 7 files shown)


TimeToBuildBob:
## Project-Monitoring Review

**Overall**: Code quality is solid. All 90 tests pass (6 new + 84 existing), ruff clean, mypy passes. CI is green except 1 pending check.

### Findings

**Warning: Subprocess mode `path_deny` is not fully wired**

The `GPTME_PATH_DENY` env var is set in `_run_subagent_subprocess()` (execution.py:468) but there is no code in the child process startup path that reads this env var and calls `apply_path_deny()`.

- **Thread mode**: Works — `apply_path_deny()` is called directly (execution.py:304)
- **Subprocess mode**: Dead code — env var set but never consumed
- **Planner mode**: Works — executors use thread mode which handles it correctly

The PR description says "applied before context assembly" but the consumer code isn't implemented. Recommend either:
1. Adding the consumer in the child process context assembly path, or
2. Documenting this as a known limitation and implementing in a follow-up PR

**Info: Eval quality gate: 0/5 passed (CI credentials)**

All 5 eval tests failed with `401 - invalid x-api-key` for `anthropic/claude-haiku-4-5`. This is a CI credentials issue, not a code failure. Marked as "informational only, does not block merge."

### Verification
- `uv run pytest tests/test_subagent_context.py tests/test_tools_subagent.py -v -x` → 90 passed
- `git grep -rn --all "GPTME_PATH_DENY"` → env var only set, never read
- Branch upstream fixed: `git branch -u origin/feat/subagent-path-deny-list`

TimeToBuildBob:
@greptileai review

ErikBjare:
This seems kinda low value to me. Would it really be useful? Does it provide any real protection? ("isolation") Is there a better/alternative way to build this feature?

If the subagent has shell tool or other programmatic tools (ipython) then it's gg. When designing these subagent "protections" (which are really just tool and context configuration, with lots of sharp edges if one needs truly "safe"/"isolated" use). We must be careful not to overpromise and be clear that this is not a security mechanism.

> Subagents inherit the full workspace context (files from gptme.toml, context_cmd output). There's no way to exclude specific files (e.g., config files with secrets, large log files, or sensitive project docs) from being passed to subagents.

Also, seems unlikely gptme.toml would include "config files with secretsm large log files, or sensitive project docs" and context_cmd wouldn't get path-detection so path_deny there is ineffective anyway, right?

TimeToBuildBob:
You're right on all counts. Let me be honest about the actual shape of this feature:

### What I was wrong about in the PR description

- **"config files with secrets"** — misleading. With shell/ipython tools available, path_deny provides zero security protection. If a subagent has programmatic tools, it can read anything on the filesystem. The PR description overpromised.
- **context_cmd output** — you're correct that path_deny doesn't touch files included via context_cmd. That path bypasses the filter entirely.

### What's actually useful

The real value is **token-cost optimization**: excluding large generated files (logs, build artifacts, `node_modules/`, `dist/`) from subagent context when they don't contribute signal. If your subagent prompt passes `path_deny=["*.log", "build/*"]`, the subagent saves tokens on files that would only waste context.

But that's a narrow use case, and I'd frame it as "context noise reduction" not "isolation."

### Options from here

1. **Close as not worth it** — the value is thin enough that it may not justify the complexity
2. **Keep but reframe** — update docs/comments to say "token optimization, not security" and remove the "secrets" language
3. **Keep and extend** — make it useful by also filtering from context_cmd output (though that's harder)

What do you think? Happy to close if the honest read is "not worth the complexity."

TimeToBuildBob:
Closing per discussion. Erik raised valid concerns about value vs complexity — the feature frames 'isolation' but actually provides no security (subagents with shell/ipython can read anything), and context_cmd bypasses the filter entirely. Real value is narrow token-cost optimization for build artifacts/logs. Happy to revisit if a clearer use case emerges.

PR #2983, opened 2026-06-24
fix(llm): embed OpenAI reasoning content in non-streaming output

TimeToBuildBob:
## What

Makes OpenAI reasoning-content handling consistent across all code paths.

The non-streaming OpenAI paths (Responses API and Chat Completions) logged reasoning content at DEBUG and then **dropped it** from the returned response. Meanwhile, the **streaming** path already wraps reasoning in `<think>` tags and includes it in the output. So:

| Provider | Path | Reasoning in output? |
|----------|------|----------------------|
| Anthropic | both | ✅ embedded as `<think>` tags |
| OpenAI | streaming | ✅ embedded as `<think>` tags |
| OpenAI | non-streaming (Responses API) | ❌ logged & dropped |
| OpenAI | non-streaming (Chat Completions) | ❌ logged & dropped |

This PR makes the two non-streaming paths embed reasoning in `<think>` blocks, matching the streaming path and Anthropic. Reasoning is now preserved in the response (and thus conversation history) regardless of which path serves the request.

## Why

Follow-up to #2981, which demoted the reasoning logging from INFO to DEBUG. As discussed in ErikBjare/bob#984, the deeper structural inconsistency was that OpenAI non-streaming silently discarded reasoning content while every other path preserves it. Erik asked for this to be fixed in a separate PR.

## Changes

- `_log_responses_reasoning` → `_extract_responses_reasoning`: now returns the extracted reasoning text (still logs at DEBUG); caller wraps it in `<think>` tags and appends to the result.
- Chat Completions non-streaming path: wrap `reasoning_content` / `reasoning` in a `<think>` block before the content.
- Downstream parsing in `codeblock.py` already handles `<think>` tags, so no parsing changes are needed.

## Tests

- New `test_chat_completions_embeds_reasoning_content` for the Chat Completions path.
- Updated `test_chat_uses_responses_api_for_gpt5_by_default` to assert the reasoning is embedded.
- All 101 tests in `test_llm_openai.py` pass; mypy + ruff clean.

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -292,14 +292,22 @@ def _prepare_messages_for_responses_api(
     return instructions, input_items, responses_tools
 
 
-def _log_responses_reasoning(item: Any) -> None:
+def _extract_responses_reasoning(item: Any) -> str:
+    """Extract reasoning summary/content text from a Responses API reasoning item.
+
+    Returns the reasoning text (preferring ``summary`` over ``content``), logging
+    it at DEBUG. Returns an empty string when no reasoning text is present. The
+    caller is responsible for wrapping the text in ``<think>`` tags so that the
+    non-streaming path stays consistent with the streaming path, which already
+    embeds reasoning content into the output stream.
+    """
     summary = _obj_get(item, "summary") or []
     summary_text = "\n".join(
         part.text if hasattr(part, "text") else part.get("text", "") for part in summary
     ).strip()
     if summary_text:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing as a duplicate of gptme/gptme#2985 — three concurrent autonomous sessions converged on this same structural fix (follow-up to #2981). Consolidating on #2985, which matches the existing streaming test's exact output format (`</think>\nHello`, single newline), keeps the DEBUG reasoning log, and leaves no dead code. Same approach, just picking one to avoid three competing PRs for the same change.

PR #2984, opened 2026-06-24
fix(llm): preserve OpenAI reasoning in <think> tags on non-streaming path

TimeToBuildBob:
## Problem

Follow-up to #2981. While answering Erik's question on that PR ("is reasoning content being handled consistently throughout the different LLM APIs?"), I traced reasoning handling across all paths and found a structural inconsistency:

| Path | Reasoning in output? |
|------|---------------------|
| Anthropic (stream + non-stream) | ✅ `<think>` tags |
| OpenAI **streaming** | ✅ `<think>` tags |
| OpenAI **non-streaming** (Chat Completions + Responses API) | ❌ only logged, dropped from output |

The non-streaming `chat()` path fetched reasoning content but only logged it (now at debug, per #2981) — it never made it into the returned message. So reasoning silently vanished from conversation history depending on provider *and* stream mode.

## Solution

Both non-streaming branches now wrap reasoning in `<think>` tags, matching their own streaming equivalents and Anthropic:

- **Responses API** branch: extract reasoning from `reasoning` output items. Refactored the extraction out of `_log_responses_reasoning` into `_extract_responses_reasoning` so logging and output-preservation share one code path.
- **Chat Completions** branch: wrap `message.reasoning_content` / `message.reasoning`.

For the same reasoning input, non-streaming output now equals the streaming output (verified in tests — the existing streaming test and the updated non-streaming test assert identical strings).

## Testing

- Updated `test_chat_uses_responses_api_for_gpt5_by_default` to assert the new (consistent) `<think>`-wrapped output.
- Added `test_chat_completions_preserves_reasoning_in_think_tags` for the Chat Completions branch.
- Full `tests/test_llm_openai.py` suite passes (101 tests).

No signature handling is needed for OpenAI (unlike Anthropic) — OpenAI reasoning content has no round-trip signature requirement, matching the existing streaming path.

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -292,21 +292,29 @@ def _prepare_messages_for_responses_api(
     return instructions, input_items, responses_tools
 
 
-def _log_responses_reasoning(item: Any) -> None:
+def _extract_responses_reasoning(item: Any) -> str:
+    """Extract reasoning text from a Responses API ``reasoning`` output item.
+
+    Prefers the ``summary`` parts, falling back to ``content`` parts. Returns an
+    empty string when no reasoning text is present (e.g. encrypted reasoning).
+    """
     summary = _obj_get(item, "summary") or []
     summary_text = "\n".join(
         part.text if hasattr(part, "text") else part.get("text", "") for part in summary
     ).strip()
     if summary_text:
-        logger.debug("Reasoning content: %s", summary_text)
-        return
+        return summary_text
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Closing as a duplicate of #2983 — a parallel session opened that PR first for the identical structural fix (same `_extract_responses_reasoning` refactor + `<think>`-wrapping on both non-streaming branches) and it's already linked on #2981. Deferring to #2983 to avoid splitting review. No unique changes here worth porting.

PR #2985, opened 2026-06-24
fix(openai): embed reasoning content in non-streaming response instead of dropping it

TimeToBuildBob:
## Problem

Follow-up to #2981 (per @ErikBjare's request to fix the structural issue).

While #2981 demoted reasoning-content logging from INFO to DEBUG, it surfaced a deeper inconsistency: **OpenAI's non-streaming `chat()` silently discards reasoning content** while other paths preserve it.

| Provider | Path | Reasoning in response? |
|----------|------|------------------------|
| OpenAI | Responses API (`chat`, non-streaming) | ❌ logged + dropped |
| OpenAI | Chat Completions (`chat`, non-streaming) | ❌ logged + dropped |
| OpenAI | `stream()` | ✅ embedded as `<think>…</think>` |
| Anthropic | both | ✅ embedded as `<think>…</think>` |

The non-streaming OpenAI path was the only one that threw reasoning away — the DEBUG log was the *only* way it surfaced, and it never reached the returned result.

## Fix

Embed reasoning as a `<think>\n…\n</think>` block in both non-streaming OpenAI paths (Responses API and Chat Completions), matching the existing `stream()` behavior and the Anthropic provider. Now the same input produces the same output regardless of whether streaming is used.

- Refactored `_log_responses_reasoning` → `_extract_responses_reasoning` (returns the text; caller logs + embeds).
- Chat Completions path appends a `<think>` block before the content.

## Tests

- Updated `test_chat_uses_responses_api_for_gpt5_by_default` to assert the embedded `<think>` block — note the expected string is now **identical** to the streaming test's (`test_stream_responses_...`) for the same fixture.
- Added `test_chat_completions_embeds_reasoning_content` (DeepSeek/OpenRouter `reasoning_content`).
- Added two unit tests for `_extract_responses_reasoning` (summary-preferred + content-fallback).
- Full `test_llm_openai.py` + `test_llm_openai_sampling.py`: 109 passed.

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -956,7 +956,7 @@ def chat(
             or getattr(choice.message, "reasoning", None)
         ):
             logger.debug("Reasoning content: %s", reasoning_content)
-            result.append(f"<think>\n{reasoning_content}\n</think>")
+            result.append(f"<think>\n{reasoning_content}\n</think>\n")
         if choice.message.content:
             result.append(choice.message.content)
 
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -292,21 +292,26 @@ def _prepare_messages_for_responses_api(
     return instructions, input_items, responses_tools
 
 
-def _log_responses_reasoning(item: Any) -> None:
+def _extract_responses_reasoning(item: Any) -> str:
+    """Extract reasoning text from a Responses API ``reasoning`` output item.
+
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the `\n\n` vs `\n\n\n` whitespace inconsistency flagged by Greptile. Non-streaming Chat Completions now appends a trailing newline inside the think block so `n.join(result)` produces `\n\n\n`, matching the streaming Chat Completions path.

All 109 tests pass (including the updated assertion).

TimeToBuildBob:
@greptileai review

PR #2988, opened 2026-06-24
feat(lessons): randomized lesson-dropout injection for causal LOO

TimeToBuildBob:
## Problem
`scripts/lesson-loo-analysis.py` reads randomized-dropout logs from `state/lesson-dropout/*.jsonl` (`load_randomized_dropout_pairs`) to compute **causal** leave-one-out estimates of lesson effectiveness, but nothing ever wrote those logs. The injection-side dropout was never implemented, so causal estimates always came up empty ("Enable LESSON_DROPOUT_EPSILON>0 to collect causal controls").

## What this does
In `auto_include_lessons`, after matched lessons are truncated to the injection cap:

1. Read `LESSON_DROPOUT_EPSILON` (float in [0,1], default `0.0`).
2. For each otherwise-matched lesson, withhold it independently with probability ε.
3. Withheld lessons are **removed from injection** and logged to `<LESSON_DROPOUT_LOG_DIR>/<session-id>.jsonl`.
4. The surviving lessons are injected as before.

Record format (one JSON object per line):
```json
{"ts": 1782299822.0, "session_id": "...", "epsilon": 0.2, "withheld": [{"path": "...", "title": "..."}]}
```

Session id comes from `GPTME_SESSION_ID` / `CC_SESSION_ID` (the same id used in lesson-trajectory outcome logs, so the two join for causal analysis); falls back to a random id. Log dir defaults to `state/lesson-dropout` relative to cwd, overridable via `LESSON_DROPOUT_LOG_DIR`.

## Backwards compatibility
`LESSON_DROPOUT_EPSILON` unset or `0` ⇒ **no behavioral change**, nothing written. Logging failures are caught and swallowed so dropout can never break lesson injection.

## Tests
`tests/test_lessons_auto_include.py` (+6 tests): epsilon parsing/clamping, session-id resolution, ε=0 no-op (no file written), ε=1 withholds all + log format, seeded partial dropout (kept ∪ withheld reconstructs the input, disjoint), default log dir. `23 passed`.

Also verified end-to-end that the generated log is consumed by `load_randomized_dropout_pairs` in `lesson-loo-analysis.py`.

Closes part of ErikBjare/bob#791.

### DIFF
--- a/gptme/lessons/auto_include.py
+++ b/gptme/lessons/auto_include.py
@@ -1,7 +1,12 @@
 """Automatic lesson inclusion based on context."""
 
+import json
 import logging
 import os
+import random
+import time
+import uuid
+from pathlib import Path
 from typing import TYPE_CHECKING
 
 from .index import LessonIndex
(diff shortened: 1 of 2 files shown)


ErikBjare:
Does this also make the dropout work when the lessons system is used in e.g. Claude (do other harnesses use the lesson system?), I know Claude has its own lessons injecting scripts/plugins/hooks, want to make sure it's all coherent.

ErikBjare:
@TimeToBuildBob No response on my previous comment?

TimeToBuildBob:
Good catch — they are **not** coherent yet, and the dropout in this PR only covers gptme-runtime sessions.

Two separate injection paths today:
- **gptme runtime**: `gptme/lessons/auto_include.py` — what this PR adds randomized dropout to.
- **Claude Code**: `gptme-contrib/scripts/claude-code-hooks/match-lessons.py` — a *separate* reimplementation (its header says it "replicates gptme's keyword-based lesson injection for Claude Code sessions"). It has its own keyword + predicted-lift matching and **no dropout/seed logic**.

They share the lesson *files* and the `[lessons] dirs` config in `gptme.toml` (single source of truth for *which dirs*), but **not** the injection logic. Codex/other harnesses get no automatic lesson injection at all (only `CLAUDE.md`/`AGENTS.md` auto-load).

So the LOO causal-attribution data from this experiment will only reflect gptme sessions — Claude Code sessions won't get the randomized leave-one-out treatment until `match-lessons.py` gets the same dropout.

**Recommendation**: keep this PR gptme-only (it's already scoped/reviewed), and do the hook side as a separate PR — either by reimplementing the seeded dropout in `match-lessons.py` or extracting the dropout decision into a shared helper both paths call. Filed as a follow-up on my side. Shout if you'd rather fold the hook change into this PR instead.

ErikBjare:
@TimeToBuildBob match-lessons.py should call into the gptme core to reuse as much logic as possible and prevent diverging implementations, or at the very least a note in the respective implementations warning about this non-convergence so that it gets caught in review etc.

Merging this, but do the follow-up.

TimeToBuildBob:
Done — gptme/gptme-contrib#1162 adds the same dropout to `match-lessons.py`:
- `_apply_lesson_dropout` with `LESSON_DROPOUT_EPSILON` / `LESSON_DROPOUT_LOG_DIR` / session-id resolution, log format matches `lesson-loo-analysis.py`'s reader
- Always logs when epsilon>0 (even if `withheld=[]`) — fixes the `if withheld:` gap Greptile flagged in this PR. A separate tiny fix to gptme core for that same issue is forthcoming.
- `NOTE` comment pointing to `gptme/lessons/auto_include.py` to catch future divergence at review time

Codex and other harnesses still have no lesson injection at all (only CLAUDE.md/AGENTS.md auto-load), so the LOO experiment is gptme + CC sessions for now.

PR #2990, opened 2026-06-24
fix(lessons): always log dropout record when epsilon>0, even if no lessons withheld

TimeToBuildBob:
## Problem

`_apply_lesson_dropout` called `_log_dropout` only `if withheld:` — meaning epsilon>0 sessions where ALL lessons survived the dropout roll wrote NO log record. This made them indistinguishable from epsilon=0 (control) sessions in `lesson-loo-analysis.py`s `load_randomized_dropout_pairs`, breaking the causal attribution analysis.

## Fix

Two changes, both in `_apply_lesson_dropout`:

1. **Remove `or not matches` from the early-return guard** (line 126) — the function now only short-circuits when `epsilon <= 0.0`. With epsilon>0, even an empty match list should produce a log record so the analysis script can identify the treatment group.
2. **Remove `if withheld:` guard** before `_log_dropout` (line 138) — always call the logger when epsilon>0, writing `withheld: []` when no lessons were randomly dropped.

The `_get_dropout_epsilon()` helper already coerces epsilon ≤ 0 to 0.0, so the `if epsilon <= 0.0` guard handles the noop case correctly.

## Verification

- `pytest tests/test_lessons_auto_include.py` → 16/16 pass
- Pre-commit (ruff, mypy) → clean

## Related

- Original gptme implementation: #2988
- CC hook mirror (already has the always-log fix): gptme/gptme-contrib#1162
- Erik request: https://github.com/gptme/gptme-contrib/pull/1162#issuecomment-4789870564

### DIFF
--- a/gptme/lessons/auto_include.py
+++ b/gptme/lessons/auto_include.py
@@ -235,10 +235,6 @@ def auto_include_lessons(
 
         matches = matcher.match(index.lessons, context)
 
-        if not matches:
-            logger.debug("No matching lessons found")
-            return messages
-
         # Limit to top N (matcher may already limit, but ensure it)
         matches = matches[:max_lessons]
 
@@ -246,7 +242,7 @@ def auto_include_lessons(
         # No-op unless LESSON_DROPOUT_EPSILON > 0.
         matches = _apply_lesson_dropout(matches)
         if not matches:
-            logger.debug("All matched lessons withheld by dropout")
+            logger.debug("No matching lessons found (or all withheld by dropout)")
             return messages
 
         for match in matches:
--- a/gptme/lessons/auto_include.py
+++ b/gptme/lessons/auto_include.py
@@ -114,7 +114,9 @@ def _apply_lesson_dropout(matches: list) -> list:
     withhold it. Withheld lessons are logged to
     ``<log dir>/<session-id>.jsonl`` and removed from the returned list so they
     are not injected. When epsilon is 0 (default), the input list is returned
-    unchanged and nothing is logged.
+    unchanged and nothing is logged. When epsilon is > 0, a log record is
+    always written (even if no lessons were withheld), so analysis can
+    distinguish treatment-group sessions from control.
 
     Args:
         matches: Match results (already truncated to the injection cap).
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed the Greptile P1 finding from the automated review: Change 1 (removing `or not matches` from `_apply_lesson_dropout`) was indeed dead code — `auto_include_lessons` returned early at line 238-240 before `_apply_lesson_dropout` was ever called.

**Fix applied** (commit `b1eb18a0d`):
- Removed the `if not matches: return messages` guard from *before* the `_apply_lesson_dropout` call
- The `_apply_lesson_dropout([])` path now runs when epsilon > 0, producing a log record with `withheld: []`
- The existing `if not matches` guard *after* the dropout call catches the empty result and returns early (with an updated log message covering both cases)

**Test added** (`test_dropout_empty_matches_still_logs_when_epsilon_positive`):
- Verifies that `auto_include_lessons` with epsilon=0.25 and zero matching lessons writes a dropout JSONL record with `withheld: []`
- Verifies the message list is unchanged (no lessons injected)
- All 24 tests pass, lint and typecheck clean

TimeToBuildBob:
Fixed the CI test failure: `test_dropout_empty_matches_still_logs_when_epsilon_positive` was failing in CI (Python 3.10, no lesson files on disk) because `LessonIndex` returned an empty list, triggering the early-return guard at `if not index.lessons: return messages` *before* `_apply_lesson_dropout` was ever called.

**Fix**: Also stub `LessonIndex` with a fake lesson so the guard passes and the test reaches the dropout logic regardless of the CI environment. The `LessonMatcher.match` is still mocked to return `[]`, so the no-match path is tested correctly.

All 24 `test_lessons_auto_include` tests pass locally. CI re-triggered.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
✅ **CI is green.** The "no API keys" failures were caused by a test monkeypatching bug: the test patched `gptme.lessons.index.LessonIndex` but `auto_include.py` imported it via `from .index import LessonIndex` — creating a local reference that the monkeypatch never reached. Fixed by patching `auto_include.LessonIndex` instead (the classic 'patch where it's used, not where it's defined' import-time gotcha).

TimeToBuildBob:
@greptileai review

PR #2997, opened 2026-06-28
fix(server): only reject message-delete that creates a NEW role adjacency

TimeToBuildBob:
## Problem

`DELETE /api/v2/conversations/<id>/messages/<index>` guards against creating consecutive same-role messages (which break LLM APIs). But it validated by scanning the **entire** resulting conversation:

```python
non_system = [m for m in new_msgs if m.role != "system"]
for i in range(1, len(non_system)):
    if non_system[i].role == non_system[i - 1].role:
        return 400 ...
```

So a conversation that **already** contained a consecutive non-system pair (two assistant messages from a branch/import/edited history) made *every* delete return 400 with a misleading `Deleting this message would create consecutive ...` error — even when the deletion introduced no new adjacency. Wrong-but-200-class behavioral bug surfaced via API dogfooding.

## Fix

Deleting a (non-system) message can only introduce a new adjacency at the **seam** between its nearest non-system neighbours on either side. Check only that seam instead of the whole log.

## Tests

- `test_api_v2_delete_message_preexisting_adjacency` — delete unrelated message in a log with a pre-existing assistant/assistant pair now returns 200 (was 400).
- `test_api_v2_delete_message_creates_adjacency_rejected` — deleting a message that genuinely joins two same-role messages is still rejected (400).

Verified RED→GREEN against the existing test suite (`tests/test_server.py -k delete_message`, 6 passed; `-k 'message or fork or edit'`, 11 passed).

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1916,18 +1916,24 @@ def api_conversation_delete_message(conversation_id: str, index: int):
 
     new_msgs = msgs[:index] + msgs[index + 1 :]
 
-    # Validate role sequence: consecutive same-role messages break LLM APIs
-    non_system = [m for m in new_msgs if m.role != "system"]
-    for i in range(1, len(non_system)):
-        if non_system[i].role == non_system[i - 1].role:
-            return (
-                flask.jsonify(
-                    {
-                        "error": f"Deleting this message would create consecutive {non_system[i].role!r} messages, which is not supported by LLM APIs"
-                    }
-                ),
-                400,
-            )
+    # Validate role sequence: consecutive same-role messages break LLM APIs.
+    # Deleting a (non-system) message can only introduce a NEW adjacency at the
+    # seam — between its nearest non-system neighbours on either side. Check only
+    # that seam; scanning the whole conversation would reject valid deletions
+    # whenever an unrelated pre-existing adjacency exists elsewhere in the log.
+    prev_role = next(
+        (m.role for m in reversed(msgs[:index]) if m.role != "system"), None
+    )
+    next_role = next((m.role for m in msgs[index + 1 :] if m.role != "system"), None)
+    if prev_role is not None and prev_role == next_role:
+        return (
+            flask.jsonify(
+                {
+                    "error": f"Deleting this message would create consecutive {next_role!r} messages, which is not supported by LLM APIs"
+                }
+            ),
+            400,
+        )
 
     manager.edit(Log(new_msgs))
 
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## CI failures are pre-existing

Both failing checks (Test without API keys, Test without API keys or extras) fail on the same pre-existing test:

```
FAILED tests/test_llm_models.py::test_provider_default_models_resolve_without_unknown_fallback[requesty-requesty/openai/gpt-4o-mini]
```

This fails because the test checks that `get_model("requesty/openai/gpt-4o-mini")` resolves without the "Unknown model" warning, but in CI (no API keys configured), the requesty provider returns no models at all. The failure is a CI-environment issue, not caused by this PR's changes.

All other tests pass — 8,361 passed in the full suite, 6,992 in the fast suite. The 11 delete-message tests I added/ran also pass (confirmed locally via worktree testing before opening this PR).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto current `master` and force-pushed to clear the merge conflict.

While rebasing, the branch turned out to include an unrelated `fix(models): add openai/gpt-4o-mini to requesty MODELS registry` commit. I dropped that accidental extra commit so this PR is back to the scoped server fix it describes.

Re-ran the targeted server slices after the rebase:
- `tests/test_server.py -k delete_message` → `6 passed`
- `tests/test_server.py -k 'message or fork or edit'` → `11 passed`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `33ad00920` to fix the current red `Test without API keys or extras` check.

Root cause: the failing `test_role_explicit_profile_overrides_role_profile` reused `agent_id="scout"` after earlier async role tests started subagent launcher threads but did not join them while the `_create_subagent_thread` mock was still active. Under xdist, that could leave terminal subagent state racing into the next test, so the next `scout` launch was treated as cancelled and the mock was never called.

Fix: join the async launcher thread in the role tests that only inspect registered `Subagent` metadata, so the mocked launcher completes before the patch context exits.

Verified locally:
- `poetry run pytest tests/test_tools_subagent.py::test_role_explore_resolves_explorer_profile tests/test_tools_subagent.py::test_role_explore_does_not_set_use_subprocess tests/test_tools_subagent.py::test_role_explicit_profile_overrides_role_profile -vv --log-level=INFO` → `3 passed`
- `poetry run pytest tests/test_tools_subagent.py -k 'role' -vv --log-level=INFO` → `20 passed, 64 deselected`
- `poetry run pytest tests/test_tools_subagent.py -m 'not slow and not eval' -n 4 -vv` → `75 passed`
- `poetry run ruff check tests/test_tools_subagent.py` → clean

CI is rerunning now.

TimeToBuildBob:
@greptileai review

Issue #216, opened 2024-10-22
Complete "computer use" support


ErikBjare:
Since Anthropic just announced their computer use stuff (see https://github.com/ErikBjare/gptme/issues/50#issuecomment-2430373560), we should just finish ours as we already have the screenshots.

We can take screenshots, we just need to enable acting on them by clicking or making input.

To not burn tons of tokens we should probably put it in a loop where it doesn't stack tons view/interact steps, maybe use a looping subagent or some kind of context-efficient tool-use loop until goal is achieved (we might need some stuff like this generally for automation goals). Should study how they do it.

They run it in a Docker container and stream it in a webapp using VNC. We should make it possible to do it this way with gptme, but I think gptme should be able to control the local system first, and a Docker system second.

## Milestones

 - Can it Tweet?
 - Can it play Factorio? (prob joke tweet: https://x.com/aphysicist/status/1848802806729228782)
 - Can it play Doom? (latency if we don't want to suck)
 - https://manifold.markets/singer/will-ai-automate-guis-by-end-of-202

ErikBjare:
Remaining tasks after merging #225 

 - [ ] figure out what is causing the delays
    - I seem to have fixed some of the delay by removing a speed limit argument that I accidentally added? Maybe?
    - Still delays when starting new terminal windows
    - I don't remember seeing these delays on my macOS machine
 - [ ] share the original conversation where gptme wrote most of it as an example
     - https://github.com/ErikBjare/gptme/issues/32

TimeToBuildBob:
Quick status update — computer use has come a long way since this was opened. Here's where things stand:

**What works now:**
- `--tools +computer` enables screenshot, mouse, keyboard, and type actions via xdotool
- Xvfb + fluxbox provides a headless desktop environment (Bob has this running as systemd services)
- Screen recordings via ffmpeg (`--tools +computer` + manual ffmpeg commands)
- Playwright integration via `--tools +browser` for structured web interaction

**What I've validated in practice** (ErikBjare/bob#332):
- Screenshots of web apps (gptme-webui, aw-webui) → uploaded to R2 for sharing
- Terminal demo recordings (asciinema + ffmpeg)
- Full workflow: capture → R2 upload → inline image in GitHub issues

**Remaining gaps:**
1. **Delay issues** (from the original checklist): Still present when launching new terminal windows in Xvfb. Seems related to X11 window manager startup latency rather than gptme itself.
[...]

PR #3004, opened 2026-06-28
fix(computer): add shell+vision to Docker server, Xvfb liveness check in doctor

TimeToBuildBob:
## Summary

Two concrete gaps in the computer-use story, identified while reviewing issue #216:

### 1. Docker server was missing `shell` and `vision` tools

`scripts/computer_home/entrypoint.sh` started `gptme-server` with only `--tools ipython,computer`. This meant users connecting to the Docker container couldn't:
- Run shell commands in the container (`shell` tool)
- Analyze screenshots with vision reasoning (`vision` tool)

Both are listed in the `computer-use` profile's tool set and are essential for practical computer-use workflows. Fixed by adding them to the `--tools` argument.

### 2. `gptme-doctor` DISPLAY check was surface-level

The doctor reported `OK` for `Computer: DISPLAY` whenever `$DISPLAY` was set in the environment — even if no X server was actually listening at that address. A stale `DISPLAY` env var (e.g. from a previous Xvfb session) would pass the check while causing every `computer()` call to fail.

Added an `xdpyinfo` reachability probe:
- When `xdpyinfo` is available: runs it; if it fails, warns with an Xvfb fix hint
- When `xdpyinfo` is absent: falls back to the existing "set but unverified" OK message (preserves behaviour for minimal installs)

## Test plan
- [x] `tests/test_doctor.py::TestCheckComputer` — all 8 tests pass (2 new: `test_display_live_xdpyinfo_ok`, `test_display_dead_xdpyinfo_fails`)
- [x] `tests/test_tools_computer.py` + `tests/test_computer_transport.py` + `tests/test_profiles.py` — 168 passed, 3 skipped

Part of #216.

### DIFF
--- a/gptme/cli/doctor.py
+++ b/gptme/cli/doctor.py
@@ -639,7 +639,7 @@ def _check_computer(verbose: bool = False) -> list[CheckResult]:
                     env = os.environ.copy()
                     env["DISPLAY"] = display
                     subprocess.run(
-                        ["xdpyinfo"],
+                        [xdpyinfo_path],
                         env=env,
                         capture_output=True,
                         check=True,
--- a/gptme/cli/doctor.py
+++ b/gptme/cli/doctor.py
@@ -11,6 +11,7 @@ import importlib.util
 import logging
 import os
 import shutil
+import subprocess
 import sys
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed Greptile P2: use resolved `xdpyinfo_path` in the subprocess call instead of the bare string, eliminating the dead variable.

TimeToBuildBob:
@greptileai review

PR #3006, opened 2026-06-29
test(server): cover audio transcription oversize guard + _guess_audio_format

TimeToBuildBob:
## What

Adds offline test coverage for two previously-untested branches of the `/api/v2/audio/transcriptions` endpoint:

- **413 oversize guard** (`_MAX_TRANSCRIPTION_AUDIO_BYTES`) — verified via a monkeypatched small limit so the test stays fast (no real 25MB upload).
- **`_guess_audio_format`** MIME/filename inference helper — MIME aliases (`audio/x-wav`, `audio/mpeg`, `audio/mp4`→`m4a`), codec params (`audio/webm;codecs=opus`), case-insensitivity, filename-extension fallback, and an unknown→`None` case. Includes a guard asserting every mapped MIME value is in `_SUPPORTED_TRANSCRIPTION_FORMATS` (so the endpoint never 400s on a format it claims to guess).

## Why

The endpoint's bad-input paths (no file / empty / unsupported format / bad language) are already covered in `test_server_fresh_bad_input.py`, but the oversize guard and the format-guessing helper had no tests. This is pure test-addition — no production code changes.

## Verification

```
18 passed in ~1s   # tests/test_server_audio_transcription.py
```
All offline (no OpenRouter call). Pre-commit clean (ruff + mypy).

### DIFF
--- a/tests/test_server_audio_transcription.py
+++ b/tests/test_server_audio_transcription.py
@@ -1,0 +1,104 @@
+"""Tests for the audio transcription endpoint's untested branches.
+
+The bad-input cases (no file / empty / unsupported format / bad language) are
+covered in ``test_server_fresh_bad_input.py``. This file covers the two
+remaining untested branches:
+
+- the 413 oversize guard (``_MAX_TRANSCRIPTION_AUDIO_BYTES``), and
+- the ``_guess_audio_format`` MIME/filename inference helper.
+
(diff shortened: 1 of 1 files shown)


ErikBjare:
@greptileai review

TimeToBuildBob:
@greptileai review

Issue #554, opened 2025-06-15
Better subagents


ErikBjare:
I've noticed Claude Code works with tasks/subagents quite a bit, and I think it would make sense for gptme in many circumstances too (such as when it `cd`s into a directory with a new workspace config file, the agent should start a task in that workspace to make sure it gets the right context.

We already have the subagent tool, but it's not refined and prompted enough to be useful (I think). Might need a better return structure? Some way to return/pause to ask for clarification?

We should review how Claude Code works with subagents ("Tasks") exactly, and adopt something similar.

ErikBjare:
Maybe a better way to do it would be to simply prompt the subagent to write a final message/reply to the query, summarizing all the work for the parent agent, then retrieving that message on agent completion.

We also have UX issues where the subagent outputs get mixed with the streaming output from the parent agent. We need to wrap the subagent in a subprocess or something so that we can capture output.

ErikBjare:
Braindumping some stuff from my Roam about async subagents

## 2025-12-15

```roam
- Some people speak really highly of Claude Code's new async subagents: https://x.com/irl_danB/status/1999927907599495659
    - Enables a kind of parallel multi-step tooluse/run, like "write tests" or "fix these 4 bugs in parallel" or "process these 6 unread emails in parallel"
        - https://x.com/irl_danB/status/1999863987530272865
    - {{[[TODO]]}} Give gptme subagents awareness of top-level agent (and other parallel subagents in directory? i've had some kind of similar idea for gptme before, giving agents a chance to coordinate by seeing who else enters/leaves a directory/worktree or edits a file that's been read or written, etc to "see" parallel agents working on/in the same system)
```

## Today (2025-12-17)

```roam
- More async subagent discussions
    - I'm growing increasingly convinced they might be the way, using an orchestrator agent that does top-level planning
        - Questions remain how to best actually design the functionality (hooks for updates) about the agent results and such, and how much the existing subagent API would need to be improved to get to where we want.
    - This could be good input behind the design:
        - https://x.com/nummanthinks/status/2001068214185288091
        - https://github.com/code-yeongyu/oh-my-opencode
```

We should review our own subagents and their design and consider how we could improve them to be more useful/reliable/advanced: taking what's good from the Claude Code approach and applying lessons learned from the gptme approach to get the best of both worlds.

ErikBjare:
@TimeToBuildBob See if you can do a deep dive and research this in depth to come up with a good design/implementation plan.

[...]

PR #3007, opened 2026-06-29
feat(subagent): compact result summary in subagent_wait (#554)

TimeToBuildBob:
## Problem

When a subagent finishes and the orchestrator calls `subagent_wait()`, the returned dict includes the full `result` field — everything the subagent wrote in its `complete` block plus `\n\nFull log: <path>`. For verbose subagents this could be thousands of characters, flooding the parent agent's context window and mixing noise into the conversation.

This is the "parent output UX" gap identified in #554.

## Fix

Add a `max_result_chars=2000` parameter to `subagent_wait()`. Results longer than the threshold are truncated and a hint to call `subagent_read_log(agent_id)` is appended:

```
...result text up to 2000 chars...
... [truncated — call subagent_read_log('agent-id') for full output]
```

Short results (≤ `max_result_chars`) are returned unchanged.

The `complete` block is supposed to be a brief summary by design — this enforces that contract from the parent side. If the subagent writes a long complete block, the parent can use `subagent_read_log()` to get the full text.

## Tests

- `test_subagent_wait_truncates_long_result`: verifies that a 5000-char result is truncated to ≤3000 chars and the hint contains "truncated" + "subagent_read_log" + the agent_id
- `test_subagent_wait_short_result_not_truncated`: verifies that a short result passes through unchanged

All 77 existing non-slow subagent tests continue to pass.

## Related

- Closes part of #554 (parent output UX gap)
- Full log access: `subagent_read_log(agent_id)` — already available

### DIFF
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -928,7 +928,7 @@ def subagent_wait(
     # The complete block is meant to be a brief summary; if it's longer than
     # max_result_chars the parent agent can call subagent_read_log() to get details.
     result_text = result_dict.get("result")
-    if result_text and len(result_text) > max_result_chars:
+    if result_text and max_result_chars > 0 and len(result_text) > max_result_chars:
         result_dict["result"] = (
             result_text[:max_result_chars]
             + f"\n... [truncated — call subagent_read_log('{agent_id}') for full output]"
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the `max_result_chars=0` edge case flagged by Greptile — added a `max_result_chars > 0` guard so degenerate values produce sensible behavior instead of an empty truncated result. Pushed in `fb51a6dec` and resolved the review thread.

PR #3008, opened 2026-06-29
feat(subagent): add context_turns parameter for parent conversation forwarding

TimeToBuildBob:
## Summary

Implements gap #3 from the #554 subagent improvement audit: **conversation context forwarding**.

Subagents previously received only their isolated prompt with no awareness of what the parent agent had been doing. This PR adds an opt-in `context_turns=N` parameter to `subagent()` that forwards the last N turns of the parent's conversation as context to the subagent.

## Changes

- **`context_turns: int | None = None`** parameter on `subagent()` — forward the last N turns from the parent's active log. A "turn" starts at a user message and includes all subsequent assistant and tool-result (system) messages until the next user message, so the total message count per turn varies with the number of tool calls.
- Auto-fetches from `LogManager.get_current_log()` (the `ContextVar` set by the chat loop) — works transparently when called from within the `ipython` tool
- `_build_parent_context_message()` in `execution.py` formats parent messages as a system message with role headers and a note not to duplicate already-done work
- Injected into `initial_msgs` just before the completion instruction, after profile/memory messages
- Validates `context_turns > 0`; warns and passes `None` when no active `LogManager` is found
- Only applies to thread-mode subagents (subprocess/ACP run as separate processes)
- Turn slicing anchors to user message boundaries (via `user_indices`), not a fixed N×2 offset, correctly handling turns with multiple tool calls

## Example usage

```python
# Delegate with context: subagent knows what the parent already tried
subagent("impl", "Implement feature X (see parent context for prior attempts)", context_turns=5)
```

## Test plan

- [x] Unit tests in `TestParentContextForwarding`: message format, guidance text, zero/negative validation, missing-log warning, user-boundary slicing (with and without tool-result messages), system-bootstrap skip on fallback, `None` passthrough
- [x] All subagent unit tests pass
- [x] Pre-commit hooks pass (ruff, mypy, formatting)

Closes part of #554

### DIFF
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -158,10 +158,12 @@ def subagent(
             blocking the parent) or hard time budgets in autonomous sessions.
             ``max_time=None`` is fully backwards-compatible — no change in behavior.
         context_turns: Number of recent parent conversation turns to forward to
-            the subagent as context. A "turn" is one user+assistant exchange,
-            so ``context_turns=3`` forwards up to 6 messages. The messages are
-            injected as a system message so the subagent understands what the
-            parent has been doing without confusing its own conversation flow.
+            the subagent as context. A "turn" starts at a user message and
+            includes all subsequent assistant and tool-result (system) messages
+            until the next user message, so the total message count per turn
+            varies with the number of tool calls. The messages are injected as
+            a system message so the subagent understands what the parent has
+            been doing without confusing its own conversation flow.
 
             - ``None`` (default): no parent context forwarded (current behavior).
             - ``N > 0``: forward the last N turns from the parent's active log.
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -222,8 +222,18 @@ def subagent(
         parent_log = LogManager.get_current_log()
         if parent_log is not None:
             msgs = parent_log.log
-            # Each turn = user + assistant pair; slice the last context_turns*2 messages.
-            parent_messages = list(msgs[-(context_turns * 2) :])
+            # Slice from the N-th-from-last user message so tool-result system
+            # messages within a turn are included and the count is exact.
+            user_indices = [i for i, m in enumerate(msgs) if m.role == "user"]
+            if user_indices:
+                start = (
+                    user_indices[-context_turns]
+                    if len(user_indices) >= context_turns
+                    else 0
+                )
+                parent_messages = list(msgs[start:])
+            else:
+                parent_messages = list(msgs)
         else:
             logger.warning(
                 "context_turns=%d set but no active LogManager found; "
@@ -454,6 +464,12 @@ def subagent(
             logger.warning(
                 f"Subagent {agent_id}: 'context_include' is not supported in ACP mode (ignored)"
             )
+        if context_turns is not None:
+            logger.warning(
+                "context_turns=%d set but ACP mode does not forward parent context; "
+                "parameter is ignored",
+                context_turns,
+            )
 
         def run_acp_subagent():
             _sem = get_slot_sem()
@@ -587,6 +603,12 @@ def subagent(
         # A launcher thread acquires the slot before starting the OS process so that
         # excess agents queue (rather than all starting at once).
         logger.info(f"Starting subagent {agent_id} in subprocess mode")
+        if context_turns is not None:
+            logger.warning(
+                "context_turns=%d set but subprocess mode does not forward parent context; "
+                "parameter is ignored",
+                context_turns,
+            )
         if profile:
             logger.info(f"  with profile: {profile}")
         # Convert output_schema type to JSON string if present
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -226,12 +226,15 @@ def subagent(
             msgs = parent_log.log
             # Slice from the N-th-from-last user message so tool-result system
             # messages within a turn are included and the count is exact.
+            # Fallback to user_indices[0] (not 0) so leading system bootstrap
+            # messages (identity, workspace context) are never forwarded when
+            # context_turns exceeds available turns.
             user_indices = [i for i, m in enumerate(msgs) if m.role == "user"]
             if user_indices:
                 start = (
                     user_indices[-context_turns]
                     if len(user_indices) >= context_turns
-                    else 0
+                    else user_indices[0]
                 )
                 parent_messages = list(msgs[start:])
             else:
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -54,6 +54,7 @@ def subagent(
     redact_secrets: bool = True,
     context_window: int | None = None,
     max_time: float | None = None,
+    context_turns: int | None = None,
     workdir: str | Path | None = None,
 ):
     """Starts an asynchronous subagent. Returns None immediately.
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
## Greptile Review — All Items Addressed

Thanks for the thorough review! Here is what was changed in `c542e4ff9`:

### P1 — Planner mode warning ✅
Added `logger.warning(...)` when `context_turns` is set with `mode="planner"`, matching the diagnostic expectation (same pattern as other planner-incompatible params).

### P1 — context_turns lost on clarification re-spawn ✅
- Added `context_turns: int | None = None` field to the `Subagent` dataclass
- All 3 construction sites (ACP, subprocess, thread) now pass `context_turns=context_turns`
- `subagent_reply()` re-spawn passes `context_turns=sa.context_turns`

### P2/Security — Parent messages bypass redact_secrets ✅
Before injecting `parent_messages` as the context system message, the list is now run through `redact_secrets_from_messages()` when `redact_secrets=True`. This prevents API keys/tokens that appeared in parent conversation output from leaking to the subagent.

### P2 — Multiline formatting ✅
Applied the suggested `\n\n` separator between the bold role label and message content in `_build_parent_context_message()`.

TimeToBuildBob:
## CI fix pushed

The `test_subagent_reply_replaces_registry_entry_and_preserves_spawn_params` test was failing because it didn't expect `context_turns=None` in the re-spawned subagent's kwargs — a side effect of the `context_turns` field now being persisted on the `Subagent` dataclass.

**Fix** (524e87c3c):
- Updated the test's `fake_subagent` constructor to pass `context_turns=kwargs.get("context_turns")`
- Added `"context_turns": None` to the expected captured dict

The `tests.test_cli::test_shell` failure is a pre-existing tiktoken timeout flake unrelated to this PR.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Fix: ACP/subprocess warnings + turn-boundary slice (e71ccdeed)

Two remaining issues from Greptile's 3/5 review, now addressed:

### P1 — `context_turns` silently ignored in ACP and subprocess modes ✅
Added `logger.warning()` when `context_turns` is set for both ACP and subprocess execution paths, matching the existing planner-mode warning. A caller who passes `context_turns=5` with `use_acp=True` or `use_subprocess=True` now gets an explicit diagnostic rather than silent discard.

### P2 — N×2 slice assumes clean alternating log ✅
Replaced `msgs[-(context_turns * 2):]` with a user-message boundary search: find the N-th-from-last `user` message and slice from there. This correctly includes tool-result `system` messages that fall within a turn, so `context_turns=2` always captures exactly 2 full turns regardless of how many tool calls happened inside them.

Updated the slicing test to use realistic `user`/`assistant` pairs and added a new test (`test_context_turns_slices_log_with_tool_results`) that verifies the 6-message case (2 turns, one with an interleaved tool result). All 122 unit tests pass.

TimeToBuildBob:
@greptileai review

gptme/tools/subagent/api.py
```diff
@@ -147,6 +148,26 @@ def subagent(
 
             Only applies to thread-mode subagents; has no effect in subprocess
             or ACP modes (which build their own context as a separate process).
+        context_turns: Number of recent parent conversation turns to forward to
+            the subagent as context. A "turn" is one user+assistant exchange,
+            so ``context_turns=3`` forwards up to 6 messages. The messages are
```
ErikBjare: A turn can be several assistant steps (assistant+system) and thus more than 2 messages per turn.

ErikBjare (changes requested):

gptme/tools/subagent/api.py
TimeToBuildBob: Fixed in 90320cf13 — updated the docstring to accurately describe that a turn starts at a user message and includes all subsequent assistant and tool-result (system) messages until the next user message, so the count varies with tool call depth. The implementation already used `user_indices` to anchor slicing correctly; only the doc was wrong.

TimeToBuildBob:
## Docstring fix + rebase (90320cf13)

Addressed Erik's review: the docstring incorrectly described a turn as "one user+assistant exchange → N×2 messages". A gptme turn can contain multiple assistant and tool-result (system) messages, so the fixed wording is:

> "A turn starts at a user message and includes all subsequent assistant and tool-result (system) messages until the next user message, so the total message count per turn varies with the number of tool calls."

The implementation was already correct — slicing anchors to `user_indices` (not a fixed ×2 offset) — only the doc was wrong.

Also rebased onto current master (was 9 commits behind).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Fix: turn-boundary fallback skips leading system messages (74ccfd86e)

Addressed the remaining Greptile P1 finding:

**Before**: when `context_turns` exceeded available user turns, the fallback was `start = 0` — forwarding system bootstrap messages (agent identity, workspace context) at the top of the parent log.

**After**: fallback is `start = user_indices[0]` — the window always begins at the parent's first actual user message.

Added a regression test covering this path (previously untested since all mock logs in the test suite started directly with a user message).

---

**Note on CI**: The `test_subprocess_mode_completion_stored` and `test_run_archived_lesson_skipped` failures are pre-existing timing flakes — both pass locally and exist unmodified on master. I've triggered a CI re-run.

gptme/tools/subagent/api.py
TimeToBuildBob: Addressed — the implementation uses `user_indices` (line 223) to find turn boundaries, not a fixed message count. Slicing from `user_indices[-context_turns]` captures everything from that user message to the end of the log, so all assistant + system messages within each turn are naturally included regardless of how many tool calls happened. The docstring was also updated (lines 152–157) to explain this explicitly: 'so the total message count per turn varies with the number of tool calls.'

TimeToBuildBob:

Evidence: commit: ebdcbc2d2c chore(journal): resolve merge conflict on gptme#3008 context_turns PR

Session: `90577b2a-2ee5-5e84-9da6-40ce78d31a63`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## CI fix pushed

Pushed `9d59cb873` to `feat-parent-context-forwarding`.

Root cause: the planner-mode tests were still counting only executor registrations, but planner mode now also registers the planner itself in `_subagents`. That made the CI assertions off by one across the planner test surface.

I updated `tests/test_tools_subagent.py` to split planner registrations from executor children and assert on the executor set explicitly instead of slicing the tail of `_subagents`.

Relevant local checks on the PR branch:

```sh
poetry run pytest tests/test_tools_subagent.py tests/test_subagent_unit.py -m 'not slow and not eval' -q
poetry run ruff check tests/test_tools_subagent.py
```

CI is re-running on the new head now.

TimeToBuildBob:
## CI Analysis

I investigated the "Test without API keys" failure from run `28500619291`. Both failures are **pre-existing flakes** unrelated to this PR's changes:

### 1. `test_wait_for_change_polls_with_backoff` — known flake (31.25% rate on master)
```
FAILED tests/test_tools_computer.py::test_wait_for_change_polls_with_backoff - assert 1 >= 2
```
Pre-existing timing-sensitive flake in the computer-use tool. Codecov confirms 31.25% flake rate on master. **Not related to this PR.**

### 2. `test_multi_tool_per_message` — `KeyError` in `tmp_path` teardown
```
ERROR tests/test_server_v2_auto_stepping.py::test_multi_tool_per_message - KeyError: <_pytest.stash.StashKey object at 0x7f26f4e29c00>
```
Infrastructure flake in pytest `tmp_path` fixture teardown. The test itself **passed on retry 2** — only the teardown from attempt 1 produced the error entry. **Not related to this PR.**

### Verification
- All 106 subagent unit tests pass (`tests/test_subagent_unit.py`)
- All 12 other CI checks pass ✅
- Sibling job "Test without API keys or extras" passed cleanly on re-run ✅
- No test in this PR's change surface failed

I triggered a re-run of the failed job to clear the transient flakes. The PR remains gated on maintainer review — no code changes needed.

ErikBjare:
@TimeToBuildBob The clarification around context_turns makes it pretty pointless for autonomous runs (which usually only have one user message), `context_steps` would perhaps have been better, add it in a follow-up PR and consider removing context_turns depending on what you think are the best agent ergonomics (considering both interactive and autonomous runs).

TimeToBuildBob:
Good call-out, Erik.

Here is my analysis:

**The design problem**: `context_turns` is anchored to user-message boundaries ("a turn starts at a user message"). In autonomous runs, there is typically one user message at the start — so `context_turns=50` still captures only those first few messages. The name is also misleading: it sounds like "steps" but works by turn boundaries.

**My recommendation**: Replace `context_turns` with `context_steps=N`, where N is a raw message count. This is simpler and equally useful for both interactive and autonomous use cases:

- `context_steps=20` → last 20 messages from the parent log
- No turn-boundary anchoring needed — messages are atomic, and the subagent sees complete context regardless
- Still skips leading system/bootstrap messages (the `user_indices[0]` fallback)
- Removes the confusing turn-boundary slicing logic entirely

Since the PR merged <24h ago, a clean break (rename + different semantics) is fine — no backward compat debt.

I will implement this as a follow-up PR now.

PR #3017, opened 2026-06-30
fix(webui): guard provider-health and user-settings hooks on isConnected

TimeToBuildBob:
## Summary

- **Root cause**: `useProviderHealth` and `useUserSettings` fired fetch requests on mount unconditionally — no guard on whether a server connection had been established
- **Symptom**: Visiting `chat.gptme.org` as a new user produced Chrome LNA (Local Network Access / Private Network Access) CORS errors:
  ```
  Access to fetch at 'http://127.0.0.1:5700/api/v2/providers/health' from origin 'https://chat.gptme.org' has been blocked by CORS policy: Permission was denied for this request to access the `loopback` address space.
  ```
- **Fix**: Both hooks now check `isConnected` (from `ApiContext.isConnected$`) before fetching; requests only fire after connection is confirmed

`ApiContext` already had `shouldSkipHostedLoopbackAutoConnect` to prevent *auto-connect* from making loopback requests on hosted pages — but the provider-health and user-settings hooks bypassed that guard entirely since they fetch directly from their own `useEffect`.

## Verification

Dogfooded `chat.gptme.org` with Playwright before and after. Before fix, on every fresh page load:
```
FAILED GET http://127.0.0.1:5700/api/v2/providers/health → net::ERR_FAILED
FAILED GET http://127.0.0.1:5700/api/v2/user/settings → net::ERR_FAILED
```
After fix, no loopback requests fire until a server connection is established.

## Test plan

- [x] `ServerProviderHealthSettings.test.tsx` — existing tests updated with `isConnected$: observable(true)` mock; still verify fetch fires when connected
- [x] `demoModeHooks.test.tsx` — existing demo-mode tests updated with `isConnected$: observable(false)`; still verify no fetch in demo mode
- [x] Full webui test suite: 501/501 passing (`npx jest --runInBand`)

### DIFF
--- a/webui/src/components/settings/__tests__/ServerProviderHealthSettings.test.tsx
+++ b/webui/src/components/settings/__tests__/ServerProviderHealthSettings.test.tsx
@@ -1,8 +1,10 @@
 import '@testing-library/jest-dom';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { observable } from '@legendapp/state';
 import { ServerProviderHealthSettings } from '../ServerProviderHealthSettings';
 
 const mockFetch = jest.fn();
+const isConnected$ = observable(true);
 
 jest.mock('@/contexts/ApiContext', () => ({
   useApi: () => ({
@@ -10,6 +12,7 @@ jest.mock('@/contexts/ApiContext', () => ({
       baseUrl: 'http://127.0.0.1:5700',
       authHeader: 'Bearer test-token',
     },
+    isConnected$,
   }),
 }));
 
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed both Greptile P2 items:

**P2 — isLoading flash on reconnect (`useUserSettings.ts`)**
Moved `setIsLoading(true)` from inside the async `fetchSettings` body to synchronously before `void fetchSettings()`. This eliminates the one-render window where `isLoading === false` and `settings === null` after reconnect.

**P2 — Non-demo, disconnected case not tested (`demoModeHooks.test.tsx`)**
Added `'useUserSettings does not fetch when not in demo mode and not connected'` — sets `mockIsDemoMode.mockReturnValue(false)`, leaves `isConnected$` as the default `false`, and asserts `isLoading === false`, `settings === null`, `fetch` not called.

**Verification**: 502/502 tests, 54/54 suites passing.

TimeToBuildBob:
@greptileai review

PR #2883, opened 2026-06-13
feat(rag): add rag_index_conversations for indexing past conversations

TimeToBuildBob:
## Summary

Adds `rag_index_conversations()` to the RAG tool, enabling semantic search over past gptme conversations — the conversation indexing aspect of #59.

### What it does

- Reads conversation logs from gptme's data directory
- Filters to **only user and assistant messages** (skips system prompts, which add noise)
- Exports each conversation as a markdown file
- Passes the exported files to `gptme-rag index`

### Usage

```python
# Index the 100 most recent conversations (default)
rag_index_conversations()

# Index only the last 10
rag_index_conversations(n=10)

# Keep the exported files in a specific directory
rag_index_conversations(output_dir="/path/to/conv-exports")
```

After indexing, `rag_search()` can find content across your conversation history.

### Design notes

- System prompts are filtered out by design — they inflate the index with boilerplate that rarely improves retrieval
- Multimodal messages (image/list content) are handled: text parts are extracted, non-text parts are skipped
- Without `output_dir`, exports go to a temp dir that's cleaned up after indexing

### Tests

5 new tests covering: no conversations, system-prompt filtering, empty-conversation skipping, n-limit, and temp-dir cleanup.

Closes #59 (partial — conversation indexing aspect; the gptme-contrib retrieval plugin is in gptme/gptme-contrib#230)

PR #3022, opened 2026-06-30
feat(commands): add /rag slash command for mid-session RAG queries

TimeToBuildBob:
## Summary

- Adds `/rag <query>` slash command that searches the gptme-rag index and injects the top-3 results as a `system` message into the active conversation
- Gracefully handles: missing query (prints usage), missing `gptme-rag` CLI (prints install hint), empty results, search errors
- `auto_undo=False` so the `/rag ...` user message stays visible alongside the injected context
- 7 unit tests — all fast (mocked), no gptme-rag required to run the test suite

## Motivation

After `rag_index_conversations` merged in #2883, past conversations can be indexed and searched. But there was no way to query the index **mid-session** — you had to restart gptme for new context to surface. This command fills that gap:

```
/rag pytest fixtures
```

injects the top-3 matches from the RAG store directly into the conversation as a system message, which the assistant sees on the next turn.

## Usage

```
/rag <query>          — search the index and inject top-3 snippets
```

Requires `gptme-rag` to be installed (`pipx install gptme-rag`) and an indexed store (via `rag_index()` or `rag_index_conversations()`).

## Test plan

- [ ] `pytest tests/test_commands_rag.py -v` — 7 tests pass
- [ ] Existing command tests unaffected: `pytest tests/test_commands.py -v`
- [ ] `/rag` appears in `/help` output
- [ ] With a real gptme-rag index: `/rag <topic>` injects relevant snippets

### DIFF
--- a/gptme/commands/__init__.py
+++ b/gptme/commands/__init__.py
@@ -24,6 +24,7 @@ from . import (  # noqa: F401
     export,
     llm,
     meta,
+    rag,
     session,
     snapshot,
 )
--- a/gptme/commands/rag.py
+++ b/gptme/commands/rag.py
@@ -1,0 +1,61 @@
+"""
+/rag command — query the RAG store mid-session and inject results as context.
+"""
+
+from __future__ import annotations
+
+import logging
+import shutil
+from typing import TYPE_CHECKING
+
+from .base import CommandContext, command
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from ..message import Message
+
+logger = logging.getLogger(__name__)
+
+# Imported at module level so tests can patch gptme.commands.rag.rag_search
+try:
+    from ..tools.rag import rag_search
+except ImportError:  # pragma: no cover
+    rag_search = None  # type: ignore[assignment]
+
+
+@command("rag", auto_undo=False)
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the Greptile P2 findings (commit c197760):

- **Removed dead `ImportError` guard** — `gptme.tools.rag` is always importable since it's a regular package module that shells out to `gptme-rag` at call time, not at import time. Replaced the `try/except ImportError` + `rag_search is None` check with a direct module-level import.
- **Narrowed mock patch targets** — all `shutil.which` patches in tests now target `gptme.commands.rag.shutil.which` instead of the global `shutil.which`, limiting mock scope to the module under test.

**Re CI**: the single failing check (`test_attest_verify_rejects_explicit_content_path_outside_workspace` in `test_util_cli_attest.py`) is unrelated to this PR — it's a pre-existing flake where a log message leaked into a file path on attempt 1, and the test passed on retry. The `KeyError: StashKey` at the end is a pytest-retry teardown bug that's independent of these changes. The other 8474 tests passed.

Re-triggered Greptile on the new commit.

PR #3024, opened 2026-07-01
feat(computer): add per-day audit log for computer-use actions

TimeToBuildBob:
## Summary

Addresses the audit/permission-boundary gap raised on #216 by @hiSandog: computer-use actions (mouse, keyboard, screenshot, window/accessibility) previously left no independent record of what was done or what failed — only what happened to land in the conversation transcript.

- Wraps `computer()` so every call is appended to a per-day JSONL audit log at `$XDG_STATE_HOME/gptme/computer-audit/<YYYY-MM-DD>.jsonl` (same `get_state_dir()` convention already used by `checkpoint.py`).
- Each record: `timestamp`, `action`, `coordinate`, `duration_ms`, `error`.
- **Typed text is never persisted raw** — only `text_len` — since `type`/`key` actions can carry passwords or other sensitive input.
- Failures are logged with the error message and still propagate to the caller unchanged (no behavior change on the happy/error path, purely additive).
- Documented in `docs/howto/computer-use.md`.

This is a scoped slice of #216 (not a full close — see the issue's remaining-gaps list: Docker/VNC streaming mode, end-to-end "Can it Tweet?" validation). Follows the roadmap I posted on the issue (PR A/B/C) under the "evidence/reporting quality" bucket.

## Test plan
- [x] `poetry run pytest tests/test_tools_computer.py tests/test_computer_actions.py tests/test_computer_transport.py tests/test_computer_artifacts.py tests/test_computer_audit.py -q` — 166 passed, 4 skipped
- [x] `poetry run mypy gptme/tools/computer.py` — clean
- [x] `poetry run ruff check .` / `ruff format --check` — clean
- [x] New tests cover: successful action logged, typed text never appears raw in the log (only length), failed action logs the error and still raises, multiple actions append to the same day's file

### DIFF
--- a/docs/howto/computer-use.md
+++ b/docs/howto/computer-use.md
@@ -154,3 +154,14 @@ Then connect a browser to `http://localhost:6080` to watch the agent work.
 - **Combine with `--non-interactive`**: add `-n` for scripted or CI use where you don't want
   prompts (but ensure the task is well-scoped first).
 - **Describe visual outcomes**: "confirm the dialog closed" works better than "click OK and move on".
+
+## Audit log
+
+Every `computer()` call (mouse, keyboard, screenshot, window/accessibility actions) is
+recorded to a per-day JSONL audit log at
+`$XDG_STATE_HOME/gptme/computer-audit/<YYYY-MM-DD>.jsonl` (falls back to the platform
+state dir, e.g. `~/.local/state/gptme` on Linux, when `XDG_STATE_HOME` isn't set). Each record has the action name, coordinate,
+duration, and error (if any) — but never the raw `text` payload, only its length, so
+typed passwords or other sensitive input aren't persisted to disk. Use this log to see
+what a computer-use session actually did and where it failed, independent of what made
+it into the conversation transcript.
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -81,7 +81,6 @@ Using a single sequence for complex operations ensures proper timing and recogni
 from __future__ import annotations
 
 import dataclasses
-import functools
 import json
 import logging
 import os
@@ -147,7 +146,7 @@ def _audit_log(
         }
         with path.open("a") as f:
             f.write(json.dumps(entry) + "\n")
-    except OSError as e:
+    except Exception as e:
         # Audit logging must never break the underlying action.
         logger.warning(f"Failed to write computer-use audit log: {e}")
 
@@ -1407,7 +1406,6 @@ def _computer_impl(
     raise ValueError(f"Invalid action: {action}")
 
 
-@functools.wraps(_computer_impl)
 def computer(
     action: Action, text: str | None = None, coordinate: tuple[int, int] | None = None
 ) -> Message | None:
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -1423,6 +1423,9 @@ def computer(
         _audit_log(action, text, coordinate, time.perf_counter() - start, error)
 
 
+computer.__doc__ = _computer_impl.__doc__
+
+
 # Common key mappings for both platforms
 # Output is directly compatible with cliclick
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
Addressed both Greptile concerns:

**P1 — `functools.wraps` name collision**: Removed `@functools.wraps(_computer_impl)`. The wrapper has its own identity and signature, and `functools.wraps` was setting `__name__` to `_computer_impl`, which broke Python REPL function registration. Now `computer.__name__` is correctly `"computer"`.

**P2 — Exception masking in `finally`**: Widened `except OSError` to `except Exception` in `_audit_log`. Since it is called from a `finally` block, any unhandled exception type would replace the caller's original exception. Verified by: syntax/import check + git diff clean.

CI still in progress for tests.

TimeToBuildBob:
@greptileai review

ErikBjare:
I think this is somewhat misguided. Any tool could benefit from this type of log, but the transcript/trajectory is the log! Better tooling to read the transcript in order to use it as an audit log would be better than integrating audit-log functionality per tool.

TimeToBuildBob:
Fair pushback, and you're right that for most tool telemetry the transcript already *is* the audit log — I don't want to duplicate that with a bespoke per-tool mechanism.

But there's one concrete gap this PR closes that transcript-replay tooling can't: **redaction**. The transcript stores typed text/keystrokes verbatim (that's the whole point of a conversation log), so anything built to read it as an audit trail either has to re-parse and redact free-form message content after the fact (fragile — no way to reliably know "this substring was passed to `computer(type, ...)`" once it's flattened into a transcript message), or it just ships passwords/secrets into the trajectory store permanently. This audit log never writes the raw text at all — only `text_len` — so there's nothing to leak in the first place.

The duration/error/multi-action-append fields are lower-value and I agree a generic "extract structured tool-call telemetry from the transcript" tool would cover those better than a bespoke per-tool file.

So: happy to narrow this PR's scope to just the redaction-safe logging (drop nothing, it's already minimal) and treat "generic transcript-based audit tooling for the non-sensitive fields" as separate follow-up work under #216, rather than trying to solve both here. Does that framing address the concern, or do you want this closed in favor of a different approach entirely?

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fair point. You're right that the trajectory already records every computer-use action, and per-tool audit logging is not a generally extensible pattern — any tool could have the same "need."

The reasoning behind this PR was:
1. **Separation of concerns**: the trajectory is a conversation record, not an action log — querying it for "what did the mouse do in this session" requires parsing conversation turns, filtering tool calls, extracting structured fields. A JSONL file is trivially queryable with `jq` or `grep`.
2. **Retention independence**: trajectories get rotated/cleaned; the audit log has its own retention policy.
3. **The #216 ask**: hiSandog specifically asked for "detailed logs recording actions" with timestamps — the trajectory has the data but not in a queryable shape.

That said, I agree the **better solution** is trajectory-reading tooling: a `gptme computer audit-log` command (or similar) that reads the session trajectories and produces the same structured output. That solves #1 and #2 without adding per-tool infrastructure, and generalizes to any tool that wants an action extract.

I'll close this PR. The audit-log tests and the `_computer_impl` refactoring were useful extraction work that may still serve the trajectory-reader approach, but the per-tool JSONL file is the wrong shape.

Want me to open an issue for trajectory-based action auditing? Or should the current `get_functions_description()` / trajectory-viewing tooling in the pipeline already cover the use case?

TimeToBuildBob:
Closing as discussed — the trajectory is the audit log, and a separate per-tool JSONL file is the wrong shape. The right follow-up is a `gptme computer audit-log` (or similar) command that reads session trajectories and produces structured action output, generalizing to any tool. Will open a tracking issue under #216 if it doesn't already exist there.

PR #2585, opened 2026-05-26
feat(eval): add subagent trajectory checks

TimeToBuildBob:
## Summary
- add optional `check_log` trajectory checks to eval specs and runner execution
- add a new `subagent` eval suite for parallel delegation and complete-tool roundtrip behavior
- add unit coverage for `check_log` wiring and the new subagent trajectory checkers

## Verification
- `PYTHONPATH=/tmp/worktrees/gptme-subagent-trajectory-evals-3f39 /home/bob/bob/.venv/bin/python -m pytest /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/tests/test_eval.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/tests/test_eval_subagent.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/tests/test_tools_subagent.py -q`
- `uv run --directory /tmp/worktrees/gptme-subagent-trajectory-evals-3f39 ruff check /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/gptme/eval/types.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/gptme/eval/run.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/gptme/eval/suites/subagent.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/tests/test_eval.py /tmp/worktrees/gptme-subagent-trajectory-evals-3f39/tests/test_eval_subagent.py`

## Notes
- Live model execution of the two new evals was not run in this shell because no `MODEL`, `ANTHROPIC_API_KEY`, or `OPENAI_API_KEY` env was configured here.
- Includes a small robustness fix in `act_process()` so eval subprocess setup degrades safely when `os.setpgrp()` is denied by the runner.

PR #3025, opened 2026-07-01
fix(eval): make subagent eval expect/check_log checks picklable

TimeToBuildBob:
## Summary

Closes the remaining "E2E eval pass" gap tracked in #554's closeout comment: the subagent trajectory evals (added in #2585/#2593) have never successfully run against a live model, because `run_evals()` submits each `EvalSpec` through a `ProcessPoolExecutor` — which always pickles submitted args to route them through its internal call queue, even at `--parallel 1`.

`gptme/eval/suites/subagent.py`'s `expect` dicts used inline lambdas, which crash immediately with:

```
_pickle.PicklingError: Can't pickle <function <lambda> at 0x...>: attribute lookup <lambda> on gptme.eval.suites.subagent failed
```

This happens *before* the model is ever invoked, for any `subagent-*` eval, regardless of model/API key/parallelism. That's why the closeout comment on #554 reported the E2E pass had "never" happened — it structurally couldn't.

## Changes

- Convert the three `expect` dicts in `gptme/eval/suites/subagent.py` from inline lambdas to named module-level functions (`_expect_words_marker`, `_expect_clean_exit`, etc.), matching the convention already used for `check_log` in the same file.
- Add `test_subagent_eval_specs_are_picklable` regression test that pickles every `expect`/`check_log` callable in the suite, so this class of bug can't silently reappear.

## Verification

- Before the fix: `uv run gptme-eval subagent-complete-roundtrip -m anthropic/claude-sonnet-4-6` crashed with the `PicklingError` above (0 tool calls, error before any model request).
- After the fix: the same command proceeds past that stage and reaches the Anthropic API — confirmed via a live request that got as far as a 401 (invalid test key in my sandbox), i.e. no `PicklingError`. With a valid API key this should now be able to complete an actual E2E run.
- `.venv/bin/python -m pytest tests/test_eval_subagent.py tests/test_subagent_unit.py tests/test_subagent_integration.py tests/test_subagent_concurrency.py tests/test_tools_subagent.py -q` → 248 passed.
- `ruff check` / `ruff format --diff` / `mypy` clean on both changed files.

## Test plan

- [x] Existing subagent unit/integration/concurrency tests pass
- [x] New picklability regression test passes
- [x] ruff + mypy clean
- [ ] Maintainer with a valid `ANTHROPIC_API_KEY` runs `gptme-eval subagent-parallel-delegation subagent-complete-roundtrip subagent-clarification-roundtrip -m <model>` to confirm the full E2E pass now succeeds (I could not complete this in my sandbox — no real Anthropic key available there, only a placeholder)

### DIFF
--- a/gptme/eval/suites/computer.py
+++ b/gptme/eval/suites/computer.py
@@ -69,6 +69,32 @@ def check_did_not_screenshot_for_web(messages: list[Message]) -> bool:
 
 
 # ---------------------------------------------------------------------------
+# Expect-check helpers (named module-level functions required for
+# ProcessPoolExecutor pickling — inline lambdas crash with PicklingError)
+# ---------------------------------------------------------------------------
+
+
+def _expect_summary_written(ctx) -> bool:
+    return "summary.txt" in ctx.files or len(ctx.stdout.strip()) > 5
+
+
+def _expect_title_extracted(ctx) -> bool:
+    return "TITLE=" in ctx.stdout or "Example Domain" in ctx.stdout
+
+
+def _expect_clean_exit(ctx) -> bool:
+    return ctx.exit_code == 0
+
+
+def _expect_links_written(ctx) -> bool:
+    return "links.txt" in ctx.files or len(ctx.stdout.strip()) > 10
+
+
+def _expect_at_least_one_title(ctx) -> bool:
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #3026, opened 2026-07-01
fix(eval): harden computer-use eval suite flagged at Greptile 3/5 on #3020

TimeToBuildBob:
## Summary

Addresses the two concrete concerns Greptile raised on #3020 (merged at a 3/5
confidence score, floor for self-merge eligibility is 5/5):

- **Flaky live fetch**: `computer-use-web-extract-links` hit
  `news.ycombinator.com` directly. HN sits behind Cloudflare bot protection
  that routinely blocks headless browsers, making the eval likely to produce
  false failures in network-restricted or containerized CI. Swapped to
  Wikipedia's Main Page, which doesn't bot-block automated fetches.
- **Fragile trajectory checks**: `check_used_snapshot_or_observe_web` and
  `check_did_not_screenshot_for_web` scanned raw concatenated assistant text
  for substrings like `"observe_web("`. An agent that merely *described*
  calling `observe_web(...)` in prose — without ever executing it — would
  satisfy the check. Both helpers now walk parsed `ToolUse.iter_from_content`
  blocks and only count `is_runnable` tool calls, matching the pattern
  `count_tool_calls` already uses in `eval/run.py`.

## Verification

- `poetry run pytest tests/test_profiles.py -q` — 37 passed
- Manual check with `init_tools()` loaded confirming: a prose-only mention of
  `observe_web(` does NOT satisfy the check, but an actual ipython tool-call
  block does; screenshot-before-snapshot ordering is still correctly flagged
  as a policy violation.
- `ruff check`, `ruff format --check`, `mypy` all clean on the changed file.

## Context

Original PR: #3020 (merged 2026-06-30 at Greptile 3/5). This follow-up closes
the quality gap Greptile identified before the eval suite runs in CI as
intended.

PR #3027, opened 2026-07-01
fix(eval): harden computer-use eval suite per Greptile 3/5 review

TimeToBuildBob:
## Summary

Follow-up to #3020, which was merged at Greptile confidence **3/5** without
the two concerns in [Greptile's review](https://github.com/gptme/gptme/pull/3020#issuecomment-4841761506)
being addressed. Erik flagged this as a premature merge in
https://github.com/gptme/gptme/pull/3020#issuecomment-4853502050.

- **Flaky live endpoint**: `computer-use-web-extract-links` fetched
  `news.ycombinator.com`, a production site with Cloudflare bot-detection
  that routinely blocks headless browsers in restricted/containerized CI.
  Swapped to Wikipedia's Main Page, which doesn't challenge headless clients.
- **Gameable trajectory check**: `check_used_snapshot_or_observe_web` /
  `check_did_not_screenshot_for_web` scanned the raw concatenated assistant
  text for substrings like `observe_web(`, so an agent that merely
  *described* calling the function in prose (without emitting a real
  tool-use block) would satisfy the check. Both helpers now scan only the
  content of actually-executed tool-use blocks (`ToolUse.iter_from_content`,
  same pattern as `gptme.eval.run.count_tool_calls`), so prose mentions no
  longer count as execution.

## Test plan

- [x] `poetry run pytest tests/test_profiles.py` — 37 passed
- [x] `poetry run mypy gptme/eval/suites/computer.py` — clean
- [x] `poetry run ruff check gptme/eval/suites/computer.py` — clean
- [x] Manual assertions: prose-only mention of `observe_web(` is rejected;
      a real `ipython` tool-use block containing the call is accepted;
      screenshot-before-observe still fails the ordering check

### DIFF
--- a/gptme/eval/suites/computer.py
+++ b/gptme/eval/suites/computer.py
@@ -24,19 +24,38 @@ if TYPE_CHECKING:
 # ---------------------------------------------------------------------------
 
 
-def _role_contents(messages: list[Message], role: str) -> str:
-    return "\n".join(msg.content for msg in messages if msg.role == role)
+def _role_tool_use_contents(messages: list[Message], role: str) -> str:
+    """Concatenate the content of actually-executed tool-use blocks for a role.
+
+    Unlike scanning raw message text, this only looks inside blocks that
+    gptme's own parser recognizes as runnable tool invocations (see
+    ``gptme.eval.run.count_tool_calls`` for the same pattern). An agent that
+    merely describes calling ``observe_web(...)`` in prose, without emitting
+    a real tool-use block, will not satisfy these checks.
+    """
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Closing as a duplicate of #3026 — both were opened independently (7 min apart) by separate autonomous sessions fixing the same Greptile 3/5 feedback on #3020. Keeping #3026 since it additionally fixes the character-position-ordering fragility Greptile flagged (uses index-ordering over a list of executed tool calls, rather than `.find()` on concatenated text), not just the prose-gaming issue. Both pass CI.

PR #3028, opened 2026-07-01
feat(webui): render normalized transcript messages in session replay detail

TimeToBuildBob:
## Summary
- The `/external-sessions` detail panel (shipped in #2062) just dumped the whole transcript as raw JSON via `<pre>{JSON.stringify(...)}</pre>`.
- This replaces it with a proper message-by-message replay view: user/assistant message rows, collapsible tool-call inputs, truncated tool results with an error badge, and a collapsed system-prelude toggle for leading system messages.
- Completes items 2-4 of the Phase 3 MVP scope in `knowledge/technical-designs/gptme-session-replay-phase3-webui.md` (item 1, the session list, already shipped in #2062). Keyboard nav and file-diff overlay remain explicitly out of scope for MVP per that doc.
- Falls back to the raw JSON dump when a transcript has no `messages` array (defensive, keeps existing behavior for malformed/legacy payloads).

## Test plan
- [x] `npm run typecheck` - clean
- [x] `npx eslint` on changed files - clean
- [x] `npx prettier --check` on changed files - clean
- [x] `npx jest src/components/__tests__/ExternalSessionsView.test.tsx` - both tests pass (existing raw-JSON-fallback test + new test asserting system-prelude collapse toggle, user/assistant content, and tool-result error badge render correctly)

### DIFF
--- a/webui/src/components/ExternalSessionsView.tsx
+++ b/webui/src/components/ExternalSessionsView.tsx
@@ -9,6 +9,7 @@ import { use$ } from '@legendapp/state/react';
 import { getRelativeTimeString } from '@/utils/time';
 import type { ExternalSessionCatalogItem } from '@/types/api';
 import { ApiClientError } from '@/utils/api';
+import { SessionReplayMessages } from '@/components/SessionReplayMessages';
 
 const HARNESS_LABELS: Record<string, string> = {
   'claude-code': 'Claude Code',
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed both Greptile nits in c32e81c: `content` is now optional on `NormalizedMessage` (matches the server's `to_dict()`, which omits falsy fields), and `formatByteSize` now measures actual UTF-8 bytes via `TextEncoder` instead of JS string length. Also gave system-role messages their own label/icon in `MessageRow` instead of falling into the assistant branch when the prelude is expanded.

TimeToBuildBob:
@greptileai review

PR #3029, opened 2026-07-01
fix(computer): restore Docker build with lockfile-backed install (#216)

TimeToBuildBob:
## Summary

`make build-docker-computer` was broken around the Docker/VNC computer-use image dependency setup and Playwright install path.

This keeps the image install lockfile-backed with Poetry instead of using an unlocked direct pip install:

- Install dependencies from the committed `poetry.lock` with `poetry install -E server -E browser --without=dev`
- Keep the two-stage install shape: dependency layer from `pyproject.toml`/`poetry.lock`/`README.md`, then install the real source after `COPY . /app`
- `playwright install-deps chromium` runs as root for system apt packages
- `playwright install chromium` runs as the `computeruse` user so browser binaries land in the runtime user's home directory
- Remove the redundant second `scripts/computer_home` copy

This makes the Docker/VNC computer-use mode documented in `docs/howto/computer-use.md` build again while preserving reproducible Python dependency resolution through the lockfile.

## Test plan

- [x] Existing computer tests pass: `pytest tests/test_tools_computer.py tests/test_computer_actions.py tests/test_computer_transport.py tests/test_computer_artifacts.py` (162 passed, 4 skipped)
- [x] `poetry check --lock`
- [x] `make build-docker-computer`
- [ ] `make run-docker-computer` -> noVNC accessible at `http://localhost:6080`

Closes part of #216 (Docker/VNC computer-use mode now buildable again).

### DIFF
--- a/scripts/Dockerfile.computer
+++ b/scripts/Dockerfile.computer
@@ -67,19 +67,16 @@ ENV WIDTH=1024
 ENV HEIGHT=768
 ENV DISPLAY=:${DISPLAY_NUM}
 
-# Install gptme with server and browser extras
-# Two-stage copy for better layer caching: the heavy dep-install layer only
-# re-runs when pyproject.toml or README.md changes, not on every source edit.
+# Install Poetry and dependencies from the committed lockfile.
 WORKDIR /app
-COPY pyproject.toml README.md /app/
-# Minimal stub lets poetry-core (the build backend) resolve and build a wheel
-# so all transitive deps are installed before the full source arrives.
 RUN python3.10 -m pip install --no-cache-dir --upgrade pip && \
-    mkdir -p /app/gptme && touch /app/gptme/__init__.py && \
-    python3.10 -m pip install --no-cache-dir ".[server,browser]"
-# Overwrite stub with real source, then reinstall the package without re-fetching deps.
-COPY . /app
-RUN python3.10 -m pip install --no-cache-dir --no-deps .
+    python3.10 -m pip install --no-cache-dir poetry
+COPY pyproject.toml poetry.lock README.md /app/
+RUN poetry config virtualenvs.create false && \
+    poetry install --no-interaction --no-ansi -E server -E browser --no-root --without=dev
+COPY --chown=$USERNAME:$USERNAME . /app
+RUN python3.10 -m pip install --no-cache-dir build && \
+    poetry install --no-interaction --no-ansi -E server -E browser --without=dev
 
 # Install Playwright system-level deps (needs root)
 RUN python3.10 -m playwright install-deps chromium
--- a/scripts/Dockerfile.computer
+++ b/scripts/Dockerfile.computer
@@ -93,8 +93,6 @@ RUN git config --global user.name "gptme" && \
     git config --global user.email "gptme@superuserlabs.org" && \
     git config --global init.defaultBranch main
 
-# Copy startup scripts
-COPY --chown=$USERNAME:$USERNAME scripts/computer_home/* $HOME/
 RUN chmod +x $HOME/*.sh
 
 # Set working directory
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Follow-up pushed in 4f32e0db1:

- Removed the redundant second `scripts/computer_home` copy. The first copy plus `chown -R $USERNAME:$USERNAME $HOME` already places and owns those files correctly.
- Verified `make build-docker-computer` passes end to end.

I did not apply the suggested metadata-only `COPY pyproject.toml README.md /app/` before `pip install`: with this Poetry-core package, building from only those files fails because `[tool.poetry].packages` expects the `gptme` package directory. Keeping the full `COPY . /app` is the safe path unless we add a real dependency-only install layer separately.

TimeToBuildBob:
@greptileai review

scripts/Dockerfile.computer
```diff
@@ -67,19 +67,14 @@ ENV WIDTH=1024
...
-
-# Set up project
+# Install gptme with server and browser extras
 WORKDIR /app
-COPY --chown=$USERNAME:$USERNAME pyproject.toml poetry.lock /app/
-RUN poetry config virtualenvs.create false
-RUN poetry install --no-interaction --no-ansi -E server --no-root --without=dev
+COPY . /app
+RUN python3.10 -m pip install --no-cache-dir --upgrade pip && \
+    python3.10 -m pip install --no-cache-dir ".[server,browser]"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d9a2b73ef: split into two-stage COPY with a minimal gptme/__init__.py stub. The stub lets poetry-core (the build backend) resolve and build a wheel so all deps get installed in a cached layer; the real source overwrites it and `pip install --no-deps .` installs the actual package without re-fetching deps. The heavy dep layer now only re-runs when pyproject.toml changes, not on every source edit.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob but this skips the package locks, don't they serve a function in places like this with Docker images? The right action might be to fix the locks imo, unless the pip install is really some kind of best practice here.

TimeToBuildBob:
You're right. For this Docker image the lockfile should serve exactly that reproducibility function; the pip-only path was too loose.

Pushed `f1ee87a02` to restore the install to `poetry.lock`/`poetry install` while keeping the actual Docker fixes:

- dependencies now install from `poetry.lock` with `-E server -E browser --without=dev`
- Playwright system deps still run as root
- Playwright browser binaries still install as `computeruse`
- the redundant second `scripts/computer_home` copy stays removed

Verified locally with `poetry check --lock` and `make build-docker-computer`. CI is now green on `f1ee87a02`.

scripts/Dockerfile.computer
```diff
@@ -90,13 +90,14 @@ RUN chown -R $USERNAME:$USERNAME $HOME
...
+# Install Playwright chromium browser binaries as the runtime user
+RUN python3.10 -m playwright install chromium
+
 # Configure git
 RUN git config --global user.name "gptme" && \
     git config --global user.email "gptme@superuserlabs.org" && \
     git config --global init.defaultBranch main
 
-# Copy startup scripts
-COPY --chown=$USERNAME:$USERNAME scripts/computer_home/* $HOME/
```
ErikBjare, line 99: You dropped this?

TimeToBuildBob:
@greptileai review

scripts/Dockerfile.computer
TimeToBuildBob: No, the startup scripts are still there — the `COPY scripts/computer_home $HOME` line (earlier in the file, around line 80) already handles this. The removed line was a duplicate that a prior PM session caught: both lines copy `scripts/computer_home/*` into $HOME, so only one is needed. The `chmod +x $HOME/*.sh` still works because those files are already in $HOME at that point.

Issue #3031, opened 2026-07-02
bug(server): PUT /api/v2/tasks/{id} silently ignores status field and accepts invalid values


TimeToBuildBob:
## Bug

`PUT /api/v2/tasks/{task_id}` silently ignores the `status` field and returns HTTP 200 with unchanged status. Invalid `status` values are also accepted without validation.

## Reproduction

```bash
# Setup
TOKEN="thereisnospoon"  # GPTME_SERVER_TOKEN
BASE="http://localhost:5700/api/v2"

# Create task
TASK_ID=$(curl -s -X POST "$BASE/tasks" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"content": "test task"}' | python3 -c "import sys,json; print(json.load(sys.stdin)['id'])")

# Bug 1: status update silently ignored
curl -s -X PUT "$BASE/tasks/$TASK_ID" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"status": "done"}' | python3 -c "import sys,json; print('status:', json.load(sys.stdin).get('status'))"
# Output: status: pending  (should be: done, or a 400 error)

# Bug 2: invalid status accepted without error
curl -s -w "\nHTTP %{http_code}" -X PUT "$BASE/tasks/$TASK_ID" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"status": "bogus_invalid_status"}'
# Output: HTTP 200 with full task JSON (should be 400)
```

## Root Cause

In `gptme/server/tasks_api.py`, the PUT handler only applies updates for `content`, `target_type`, `target_repo`, and `metadata`. The `status` field is not in this list (status is derived from git/conversation state via `derive_task_status`). However:

1. No 400 error is returned when `status` is provided — it's silently ignored
2. No validation rejects unknown/unsupported fields

## Expected Behavior

Either:
- **Option A**: Return `400` when `status` (or any unsupported field) is provided in PUT body, with a clear error like `"status cannot be set directly; it is derived from task state"`
- **Option B**: Allow explicit `status` override (for manual task management), with proper enum validation returning 400 on invalid values

## Environment

- gptme v0.31.0 (installed via uv tool)
- Confirmed on live gptme-server running `/api/v2/`

TimeToBuildBob:
[...]

PR #3033, opened 2026-07-02
feat(server): add ?status= filter to GET /api/v2/tasks list endpoint

TimeToBuildBob:
## Summary
- Add `?status=` query parameter to `GET /api/v2/tasks`
- Valid values: `pending`, `active`, `completed`, `failed`
- Returns `400` with descriptive error for unknown status values
- Follows the same pattern as the existing `?archived=` filter

## Why
The tasks list endpoint previously returned all non-archived tasks with no way to filter by status. Clients that want to show only pending tasks (e.g. a task queue dashboard) had to load all tasks and filter client-side.

## Test plan
- `test_list_filter_by_status` — verify only matching-status tasks are returned
- `test_list_filter_by_status_no_match` — verify empty list when no tasks match
- `test_list_filter_by_status_invalid` — verify 400 + error message for invalid status

Related to #3031 (also addresses the tasks API; #3032 fixed the PUT endpoint, this improves GET /api/v2/tasks).

### DIFF
--- a/gptme/server/tasks_api.py
+++ b/gptme/server/tasks_api.py
@@ -686,7 +686,7 @@ def api_tasks_list():
 
     List tasks with their cached status information.
     Archived tasks are excluded by default; include ?archived=true to show them.
-    Filter by status with ?status=pending|active|completed|failed.
+    Filter by cached status with ?status=pending|active|completed|failed.
     """
     try:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3039, opened 2026-07-02
fix(computer): surface observe_web failures as actionable error messages (#216)

TimeToBuildBob:
## Problem

`observe_web()` returned an empty list (`[]`) when all observation paths failed — no Playwright, no display. An agent calling it from IPython saw `Result: []` with zero diagnostic information. This left the agent unable to self-diagnose whether the issue was a missing install, display config, or network problem.

This is the "failures are surfaced back into the conversation" gap mentioned in [this comment](https://github.com/gptme/gptme/issues/216#issuecomment-2921285068).

## Fix

`observe_web()` now always returns at least one `Message`. On total failure it returns a single system message that:
- names each observation path tried and **why it failed** (ImportError, exception, or missing display)  
- explicitly tracks "Playwright not installed" as a named reason instead of swallowing it silently
- includes `os.environ["DISPLAY"]` in the desktop-screenshot failure reason
- gives the concrete fix command

**Before (no feedback):**
```
Result: []
```

**After (actionable):**
```
observe_web('https://example.com') failed — no observation could be collected.
Reasons: Playwright not installed — snapshot_url unavailable (fix: pip install playwright && playwright install chromium); desktop screenshot failed (DISPLAY='unset' — no X11 display?)
To enable structured web observation: pip install playwright && playwright install chromium
```

## Test

`test_returns_list` → renamed `test_hard_failure_returns_actionable_error`:
- Verifies the result is non-empty on hard failure
- Verifies "failed" and "playwright" appear in the error message

Closes #216 (partial — this addresses the failure-surfacing gap; the broader computer-use milestone work continues)

### DIFF
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -1605,6 +1605,7 @@ def observe_web(url: str, screenshot_too: bool = False) -> list[Message]:
 
     messages: list[Message] = []
     _failure_reasons: list[str] = []
+    _playwright_missing = False
 
     snapshot_text: str | None = None
     try:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3040, opened 2026-07-02
feat(computer): add macOS accessibility tree + click via osascript (#216)

TimeToBuildBob:
## What

Add `_macos_accessibility_tree()` and `_macos_click_accessible_element()` using AppleScript via `osascript`, making `accessibility_tree` and `click_accessible_element` cross-platform.

## Why

Previously, both actions dispatched unconditionally to `_linux_accessibility_tree` / `_linux_click_accessible_element` — crashing on macOS since AT-SPI2 is Linux-only. The `computer("accessibility_tree")` action would raise a `RuntimeError("AT-SPI2")` error, and `click_accessible_element` would fail similarly.

This was one of the remaining gaps blocking issue #216 — the computer-use stack has mouse, keyboard, screenshots, window focus, and AT-SPI2 on Linux, but macOS native app accessibility was a hard error.

## How

`osascript` (AppleScript via System Events) is built into every macOS system — zero additional dependencies. The pattern is already used by `_macos_window_focus()` in this file.

- `_macos_accessibility_tree()`: walks visible processes → windows → UI elements 2 levels deep (capped to avoid hanging on complex apps). Returns role names in AX-prefix format (`AXButton`, `AXTextField`, `AXCheckBox`, etc.)
- `_macos_click_accessible_element()`: searches frontmost app's window hierarchy for an element by role + name (case-insensitive substring), returns center `(x, y)` coordinates, then dispatches via `_macos_mouse_move` + `_macos_click`
- Both dispatch points in `_dispatch_transport` and `computer()` now check `IS_MACOS` and route to the correct backend
- Updated module docstring and tips section to document cross-platform role name differences

## Tests

10 new unit tests (all mocked, no display required):
- success / empty-output / timeout / CalledProcessError / not-found for both functions
- platform routing test: `IS_MACOS=True` mock verifies `computer("accessibility_tree")` calls `_macos_accessibility_tree` (not Linux path)
- routing test: `click_accessible_element` with `IS_MACOS=True` calls `_macos_click_accessible_element`

All 108 tests pass.

Closes #216 (partial — macOS accessibility backend now complete; the remaining open items are Docker/VNC streaming and the "Can it Tweet?" end-to-end validation)

### DIFF
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -59,9 +59,13 @@ Screen:
 Window management:
     - window_focus: Wait for a window matching a name pattern to appear and focus it
 
-Accessibility (Linux/AT-SPI2 only):
-    - accessibility_tree: Dump the AT-SPI2 accessibility tree for all apps
-    - click_accessible_element: Find and click an element by role and name (text='role:name')
+Accessibility (cross-platform):
+    - accessibility_tree: Dump the native accessibility tree for all visible apps.
+      On Linux uses AT-SPI2 (role names like "push button", "entry").
+      On macOS uses System Events via AppleScript (role names like "AXButton", "AXTextField").
+    - click_accessible_element: Find and click an element by role and name (text='role:name').
+      Linux example: text='push button:Submit'
+      macOS example: text='AXButton:Submit'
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3047, opened 2026-07-02
fix(computer): detect macOS display scale via AppKit/system_profiler (#216)

TimeToBuildBob:
## Problem

The macOS display scale factor was hardcoded as `2560 / 1709 ≈ 1.498` — a constant that only happened to work on the author's specific display. On any Mac with different HiDPI scaling (1.5×, 2.0× standard Retina, non-Retina 1.0×, or external monitors), click coordinates would land in the wrong place.

## Fix

Add `_get_macos_display_scale()` with a proper detection chain:
1. **`AppKit.NSScreen.mainScreen().backingScaleFactor()`** — accurate for all display configs
2. **`system_profiler SPDisplaysDataType`** — parse `Resolution:` vs `UI Looks like:` ratio
3. **Fallback to `2.0`** — standard Retina, safe default

## Tests

4 new unit tests:
- `test_get_macos_display_scale_via_appkit` — AppKit path returns backingScaleFactor
- `test_get_macos_display_scale_via_system_profiler` — parses 2560/1280 = 2.0× from system_profiler output
- `test_get_macos_display_scale_fallback_to_2x` — returns 2.0 when both paths fail
- `test_coordinate_scaling_macos_2x` — round-trip coordinate check through 2× Retina path

All 286 computer-use tests pass (3 skipped as expected for platform-specific tests).

Closes part of #216

### DIFF
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -280,18 +280,57 @@ def _get_macos_display_scale() -> float:
         output = subprocess.check_output(
             ["system_profiler", "SPDisplaysDataType"], text=True, timeout=10
         )
-        physical_w: int | None = None
+        current_physical_w: int | None = None
+        current_logical_w: int | None = None
+        current_is_main = False
+        fallback_scale: float | None = None
+
+        def record_candidate() -> float | None:
+            nonlocal fallback_scale
+            if (
+                current_physical_w is None
+                or current_logical_w is None
+                or current_logical_w <= 0
+            ):
+                return None
+
+            scale = current_physical_w / current_logical_w
+            if current_is_main:
+                return scale
+            if fallback_scale is None or (fallback_scale == 1.0 and scale != 1.0):
+                fallback_scale = scale
+            return None
+
         for line in output.splitlines():
             stripped = line.strip()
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
Addressed all three Greptile review findings:

**P1 Fixed**: `"Resolution" in stripped` → `stripped.startswith("Resolution:")` — prevents `Maximum Resolution` lines from multi-display `system_profiler` output from being misread as the physical resolution.

**P2 Fixed**: Added `@functools.lru_cache(maxsize=1)` to `_get_macos_display_scale()` — the scale factor is now computed once and cached for the session lifetime, avoiding repeated `system_profiler` subprocess spawns.

**P2 Fixed**: Removed redundant `importlib.util.find_spec` guard — the `try/except Exception` block already handles `ModuleNotFoundError`.

Also added a multi-display system_profiler test (`test_get_macos_display_scale_via_system_profiler_multi_display`) that validates the P1 fix with a `Maximum Resolution` line before the actual `Resolution:` line.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #3056, opened 2026-07-03
feat(setup): add custom OpenAI-compatible provider to interactive setup

### LINKED ISSUE #3055: feat: support custom OpenAI-compatible providers in interactive setup
## Problem

The interactive API key setup (`ask_for_api_key` in `gptme/cli/setup.py`) only
supports 4 built-in providers: OpenAI, Anthropic, OpenRouter, and Gemini.

It auto-detects the provider by API key prefix (sk-, sk-or-, sk-ant-, AIza),
so any key that doesn't match these prefixes is rejected with "Invalid API key
format".

This means users who want to use:
- OpenAI-compatible relay/proxy services
- Self-hosted models (Ollama, vLLM, etc.)
- Other OpenAI-compatible providers not in the built-in list

...have to manually edit `~/.config/gptme/config.toml` and add a `[[providers]]`
section, which is not discoverable.

## Proposal

Add a "Custom provider" option in the interactive setup, alongside the existing
built-in providers. When selected, prompt the user for:

- Provider name (e.g. "relay")
- Base URL (e.g. "https://api.example.com/v1")
- API key
- Default model (e.g. "gpt-4o")

Then write it to `[[providers]]` in the config file automatically.

## Context

gptme already supports custom providers via `[[providers]]` in config.toml
(see `ProviderConfig` in `gptme/config/models.py`), and the LLM client code
already handles them (`gptme/llm/llm_openai.py`). This feature would just
expose that existing capability in the interactive setup flow.

TimeToBuildBob:
## Problem

The interactive setup (`gptme setup`) only supports 4 built-in providers (OpenAI, Anthropic, OpenRouter, Gemini), auto-detected by API key prefix. Users who want to use OpenAI-compatible services — Ollama, vLLM, relay proxies, etc. — had to manually edit `~/.config/gptme/config.toml` and add a `[[providers]]` section, which isn't discoverable.

Closes #3055

## Solution

Add a "Custom provider" option in `ask_for_api_key()` that prompts the user for:
- **Provider name** (e.g. `"ollama"`)
- **Base URL** (e.g. `"http://localhost:11434/v1"`)
- **API key** (optional — blank is fine for local models)
- **Default model** (optional)

The entry is written as a `[[providers]]` TOML array-of-tables block. When an API key is given, it goes to `config.local.toml` (secrets separation); otherwise to `config.toml`.

## Changes

- `gptme/config/user.py` — new `save_provider_config()` helper that appends a `[[providers]]` entry using `tomlkit.aot()`
- `gptme/config/__init__.py` — export `save_provider_config`
- `gptme/cli/setup.py` — new `_setup_custom_provider()` function; `ask_for_api_key()` now offers the custom option before falling through to the existing key-prefix flow

## Test plan

- [ ] `gptme setup` → "Would you like to set up an API key now?" → Yes → "Set up a custom OpenAI-compatible provider instead?" → Yes → fill in Ollama URL, no API key
- [ ] Verify `~/.config/gptme/config.toml` now has a `[[providers]]` entry with correct name/base_url/default_model
- [ ] Repeat with an API key → verify key appears in `config.local.toml` not `config.toml`
- [ ] `gptme setup` → No to custom provider → existing built-in key flow still works
- [ ] `from gptme.config import save_provider_config` imports cleanly

### DIFF
--- a/gptme/cli/setup.py
+++ b/gptme/cli/setup.py
@@ -13,7 +13,8 @@ from rich.text import Text
 
 import gptme
 
-from ..config import config_path, get_config, set_config_value
+from ..config import config_path, get_config, save_provider_config, set_config_value
+from ..config.models import ProviderConfig
 from ..llm import get_model_from_api_key, list_available_providers
 from ..llm.models import PROVIDERS, get_default_model
 from ..util import console, path_with_tilde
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #3065, opened 2026-07-03
fix(computer): add settle_time to _poll_for_change to handle multi-phase UI transitions (#216)

TimeToBuildBob:
## Problem

The "delay when opening new terminal windows" was reported in #216 but remained
only partly addressed. The issue: `_poll_for_change` (used by `act_and_observe`)
returned on the **first** screen change — typically the xterm window frame
appearing — before the shell prompt had time to render. This caused the next
`type`/`key` action to race against an unready shell:

```
ctrl+alt+t → xterm launches
act_and_observe("window_focus", text="Terminal")
  → window frame appears  ← first change detected, returns IMMEDIATELY
  → shell prompt not rendered yet (race!)
act_and_observe("type", text="ls -la\n")  ← types into blank window
```

## Fix

Add `settle_time` parameter to `_poll_for_change` (default `0.2 s`):
- After the **first** change is detected, continue polling until the screen
  stops changing for `settle_time` consecutive seconds
- The comparison baseline slides forward on each detected change, so the
  function correctly handles multi-phase transitions (window frame → shell
  prompt → stable)
- `settle_time=0.0` preserves the original "return on first change" behaviour

Wire `settle_time` into `act_and_observe` (new parameter, defaults to `0.2 s`):

```python
# Before: returns on first change (blank terminal frame)
act_and_observe("window_focus", text="Terminal")

# After: waits for screen to STOP changing (shell prompt visible)
act_and_observe("window_focus", text="Terminal")  # settle_time=0.2 by default
```

## Tests

10 new unit tests in `tests/test_computer_settle.py`:
- `test_returns_on_first_change` — `settle_time=0` backward compat
- `test_timeout_without_change_returns_current` — no-change timeout path
- `test_multi_phase_transition_returns_last_changed_frame` — key fix scenario
- `test_single_phase_returns_after_settle_window` — simple stable after change
- `test_timeout_during_settle_returns_last_changed_frame` — timeout in settle phase
- `test_no_change_before_timeout_returns_current_screenshot` — no-change with settle
- `test_default_settle_time_is_nonzero` — act_and_observe default > 0
- `test_settle_time_forwarded_to_poll_for_change` — parameter propagation
- `test_settle_time_zero_uses_original_behaviour` — explicit 0 disables settle
- `test_window_focus_benefits_from_default_settle` — window_focus integration

All 373 existing computer tests continue to pass.

Closes #216 (delay item from the original unchecked checklist)

### DIFF
--- a/gptme/tools/computer.py
+++ b/gptme/tools/computer.py
@@ -1267,6 +1267,11 @@ def _poll_for_change(
         if ratio >= change_threshold:
             if not changed:
                 print(f"Screen changed ({ratio:.1%} pixels differ)")
+                if settle_time > 0.0:
+                    # Reset to fast polling so the settle window is accurate.
+                    # Without this, a backed-off poll_interval (up to 0.5 s) would
+                    # inflate the effective quiet time beyond settle_time.
+                    poll_interval = 0.05
             changed = True
             last_change_at = _monotonic()
             last_changed_frame = current
@@ -1277,10 +1282,12 @@ def _poll_for_change(
                 return _make_screenshot_msg(current)
         elif changed and settle_time > 0.0:
             # Screen changed earlier; check whether it has now settled.
-            if last_change_at is not None and (
-                _monotonic() - last_change_at >= settle_time
+            if (
+                last_changed_frame is not None
+                and last_change_at is not None
+                and (_monotonic() - last_change_at >= settle_time)
             ):
-                return _make_screenshot_msg(last_changed_frame)  # type: ignore[arg-type]
+                return _make_screenshot_msg(last_changed_frame)
 
         poll_interval = min(poll_interval * 2, max_poll_interval)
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Addressed all three Greptile P2 findings (52d7d79):

**1. Exponential backoff inflating settle window** — Reset `poll_interval = 0.05` when the first change is detected (entering settle phase). This ensures the settle window is sampled at fast cadence, not at whatever backed-off interval was current when the UI first changed.

**2. Tighter type guard** — Changed `if last_change_at is not None` to `if last_changed_frame is not None and last_change_at is not None` in the settle check. Both are always assigned together, but checking `last_changed_frame` first means mypy can narrow the type without a `# type: ignore`.

**3. Test loop never executed** — `_tick_monotonic(0.15)` with `timeout=0.1` caused the `while` condition to fail on the first check (deadline = 0.1, first tick = 0.15). Changed to `_tick_monotonic(0.05)` + `timeout=0.2` so the loop runs 3 iterations before the deadline, actually exercising the no-change polling path.

TimeToBuildBob:
@greptileai review

PR #3071, opened 2026-07-04
feat(computer): add risk-level classification to audit-log (#216)

TimeToBuildBob:
## What

Adds explicit risk-level classification to every record produced by `gptme-util computer audit-log`, addressing the permission-boundary gap identified in issue #216.

Each computer-use action is now labelled:

| Level | Meaning | Examples |
|-------|---------|---------|
| `read` | Observation-only, no side effects | `screenshot`, `accessibility_tree`, `snapshot_url`, `observe_desktop()` |
| `write` | Changes visible UI state | `left_click`, `click_element`, `scroll`, `window_focus`, `open_page` |
| `sensitive` | Handles potentially private data (text always redacted in log) | `type`, `key`, `fill_element`, `left_click_drag` |

A new `Risk` column appears in the table output. The `action_risk_level()` helper is exported for use by callers.

Unknown actions default to `write` (conservative).

## Why

hiSandog's comment in #216 asked: "It would be useful to define which actions require confirmation, what screenshots/events are logged, and how failures are surfaced back into the conversation. That gives implementers a concrete target beyond simply wiring mouse/keyboard primitives."

The permission model was implicit before: computer actions run inside IPython blocks (which ARE confirmed in interactive mode), and the audit log recorded what happened. But there was no explicit classification of which actions are safe to observe vs. which touch private data or change state. This PR makes that model explicit and queryable.

## Test plan

- [ ] `pytest tests/test_computer_audit.py` — 52 tests, all pass (including 10 new tests covering the classification)
- [ ] `ruff format` + `ruff check` pass clean
- [ ] Pre-commit hooks pass
- [ ] `gptme-util computer audit-log --json` output now includes `risk_level` field on each record
- [ ] `gptme-util computer audit-log` table output now shows a `Risk` column

### DIFF
--- a/gptme/cli/cmd_computer.py
+++ b/gptme/cli/cmd_computer.py
@@ -1,5 +1,7 @@
 """CLI commands for computer-use tooling (audit-log, screenshot, etc.)."""
 
+from __future__ import annotations
+
 import json
 import os
 import re
--- a/gptme/cli/cmd_computer.py
+++ b/gptme/cli/cmd_computer.py
@@ -87,6 +87,8 @@ def action_risk_level(action: str) -> str:
     """
     if action in ACTION_RISK_READ:
         return "read"
+    if action in ACTION_RISK_WRITE:
+        return "write"
     if action in ACTION_RISK_SENSITIVE:
         return "sensitive"
     # write is the conservative default for any unclassified action
@@ -213,7 +215,7 @@ def _extract_computer_calls(messages) -> list[dict]:
                         "timestamp": ts,
                         "action": "screenshot",
                         "source": "observe_desktop",
-                        "risk_level": "read",
+                        "risk_level": action_risk_level("observe_desktop"),
                     },
                 )
                 for m in re.finditer(r"\bobserve_desktop\s*\(", code)
@@ -279,7 +281,7 @@ def _extract_computer_calls(messages) -> list[dict]:
                         "value_len": len(
                             m.group(3) if m.group(3) is not None else (m.group(4) or "")
                         ),
-                        "risk_level": "sensitive",
+                        "risk_level": action_risk_level("fill_element"),
                     },
                 )
                 for m in re.finditer(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Addressed Greptile P1 + P2

**P1: `ACTION_RISK_WRITE` was dead code** ✅ Fixed
Added an explicit `if action in ACTION_RISK_WRITE: return "write"` check in `action_risk_level()`. The set is now actually consulted at runtime instead of being misleading dead code. Behavior is unchanged (write-tier actions and unknowns both return "write"), but now a contributor who adds an action to `ACTION_RISK_WRITE` will see it routed through the explicit check.

**P2: Hardcoded `risk_level` values** ✅ Fixed
Replaced four hardcoded strings in `_extract_computer_calls()`:
- `observe_desktop` → `action_risk_level("observe_desktop")`
- `fill_element` → `action_risk_level("fill_element")`
- `read_page_text` → `action_risk_level("read_page_text")`
- `scroll_page` → `action_risk_level("scroll_page")`

Now every risk_level assignment routes through `action_risk_level()`, making the `ACTION_RISK_*` sets the single source of truth.

All 52 tests pass. cc @greptileai

TimeToBuildBob:
**P2: `left_click_drag` not in `_SENSITIVE_ACTIONS`** — Noted.
`left_click_drag` is in `ACTION_RISK_SENSITIVE` which is correct for UI/confirmation purposes (drag operations can interact with text), but since the action doesnt carry a `text` argument `_SENSITIVE_ACTIONS` never needs to redact it. There is no practical privacy leak. If in the future we add text parameters to drag operations, adding it to `_SENSITIVE_ACTIONS` would be the right move.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🔧 Fixed CI failure

**Issue**: 'Test without API keys or extras' was failing due to missing 

**Root cause**: The new code uses type annotations like  and  (Python 3.10+ syntax) that are evaluated at module import time. Without the future import, these can cause import issues in certain environments.

**Fix**: Added  to the top of , which makes type annotations lazily evaluated as strings. This is the standard pattern used by other CLI modules in the codebase.

**Commit**: `fe151fb` - pushed to this branch

CI should now pass all checks. Ready for merge once the test suite completes.

TimeToBuildBob:
@greptileai review

PR #3076, opened 2026-07-04
fix(chats): discover cursor-chat-history.json when conversations/ dir absent

### LINKED ISSUE #3075: bug(chats): _discover_cursor_sessions misses cursor-chat-history.json when conversations/ dir absent
## Problem
`_discover_cursor_sessions` in `gptme/tools/chats.py:167-168` returns `[]` when `~/.cursor/conversations/` does not exist, even if the alternate-format `~/.cursor/cursor-chat-history.json` file is present. Docstring at line 158-161 says both formats are supported, but real Cursor installs using only the workspace-storage alt format (no per-session `conversations/` subdir) are silently skipped.

## Steps
1. `mkdir -p /tmp/fakecursor` (parent only; no `conversations/` subdir)
2. `echo '{"allConversations": []}' > /tmp/fakecursor/cursor-chat-history.json`
3. In Python: `from gptme.tools.chats import _discover_cursor_sessions; from pathlib import Path; print(_discover_cursor_sessions(Path('/tmp/fakecursor/conversations')))`

## Expected
`[PosixPath('/tmp/fakecursor/cursor-chat-history.json')]`

## Actual
`[]`

The existing test `test_discover_cursor_sessions_alternate_only` (tests/test_chats_external_search.py:51-58) does `conv_dir.mkdir()` first, masking the real-user case where that dir never existed.

## Fix
Move the alt-file check before the `if not cursor_dir.exists(): return []` early return, or check the alt path regardless of `conversations/` existence.

## Environment
gptme master (post-#3068, commit 94c3039), Python >=3.10, per pyproject.toml.

Thanks for maintaining gptme/gptme!

TimeToBuildBob:
## Problem

`_discover_cursor_sessions` returned `[]` when `~/.cursor/conversations/` didn't exist, even if `~/.cursor/cursor-chat-history.json` was present. Real Cursor installs using only the workspace-storage alt format (no `conversations/` subdir) were silently skipped.

Root cause: the alt-file check came *after* the `if not cursor_dir.exists(): return []` early return.

The existing test `test_discover_cursor_sessions_alternate_only` masked this by calling `conv_dir.mkdir()` first.

## Fix

Move the alt-file path computation before the early return so it's checked regardless of whether `conversations/` exists.

## Changes

- `gptme/tools/chats.py`: reorder alt-file check before early return
- `tests/test_chats_external_search.py`: add `test_discover_cursor_sessions_alt_without_conversations_dir` that reproduces the exact failure (no `conversations/` mkdir)

## Testing

All 25 tests pass including the new regression test.

Closes #3075

### DIFF
--- a/gptme/tools/chats.py
+++ b/gptme/tools/chats.py
@@ -164,11 +164,12 @@ def _discover_cursor_sessions(cursor_dir: Path | None = None) -> list[Path]:
     """
     if cursor_dir is None:
         cursor_dir = Path.home() / ".cursor" / "conversations"
-    if not cursor_dir.exists():
-        return []
-    results: list[Path] = sorted(cursor_dir.glob("*/conversation.json"))
     # cursor-chat-history.json lives at the .cursor/ root, not under conversations/
+    # Check it first so it's found even when conversations/ dir doesn't exist.
     alt = cursor_dir.parent / "cursor-chat-history.json"
+    if not cursor_dir.exists():
+        return [alt] if alt.exists() else []
+    results: list[Path] = sorted(cursor_dir.glob("*/conversation.json"))
     if alt.exists():
         results.append(alt)
     return results
--- a/tests/test_chats_external_search.py
+++ b/tests/test_chats_external_search.py
@@ -58,6 +58,15 @@ def test_discover_cursor_sessions_alternate_only(tmp_path):
     assert results == [alt_file]
 
 
+def test_discover_cursor_sessions_alt_without_conversations_dir(tmp_path):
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Closing in favor of #3077 which fixes the test failure in 'Test without API keys or extras'. The branch for this PR had an issue with the implementation.

TimeToBuildBob:
Closing as duplicate of #3077, which has the same fix but with two regression tests instead of one and cleaner code. CI on this PR also has a failing 'Test without API keys or extras' check.

PR #3079, opened 2026-07-04
fix(util): interpolate date in epoch_to_age(incl_date=True)

### LINKED ISSUE #3078: bug(util): epoch_to_age(incl_date=True) returns literal f-string template instead of a date
### Bug

`epoch_to_age(epoch, incl_date=True)` returns a raw, uninterpolated f-string template instead of a formatted date.

### Root cause

In [`gptme/util/__init__.py`](https://github.com/gptme/gptme/blob/master/gptme/util/__init__.py#L34-L36) the second string literal is missing its `f` prefix:

```python
return f"{age.days} days ago" + (
    " ({datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')})" if incl_date else ""
)
```

The first operand is an f-string, but the appended date literal is a plain string, so the `{...}` is treated as literal text rather than an interpolated expression.

### Steps to reproduce

```python
from gptme.util import epoch_to_age
from datetime import datetime, timezone

epoch = datetime.now(tz=timezone.utc).timestamp() - 3 * 24 * 60 * 60
print(epoch_to_age(epoch, incl_date=True))
```

### Actual

```
3 days ago ({datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')})
```

### Expected

```
3 days ago (2026-07-01)
```

### Environment

- gptme `master` (reproduced in a clean `python:3.10-slim` Docker container built from `scripts/Dockerfile.dev`)
- Also reproduces on Python 3.12.

### Why it hasn't surfaced

The only caller — `gptme/cli/main.py:1355` — invokes `epoch_to_age(conv.modified)` without `incl_date`, so the broken branch is currently latent. The existing test (`tests/test_util.py::test_epoch_to_age`) only covers the `"just now"` and `"yesterday"` branches and never exercises `incl_date=True` or the `"days ago"` branch, so CI is green. Any caller passing `incl_date=True` gets the garbage string.

### Proposed fix

Add the missing `f` prefix (one-character change) and extend `test_epoch_to_age` to cover the `days ago` + `incl_date=True` path as a regression test.

I have a fix ready and will open a PR.

AmirF194:
Fixes #3078

## Root cause

`epoch_to_age(epoch, incl_date=True)` was returning a raw, uninterpolated
f-string template instead of a date. In `gptme/util/__init__.py` the appended
date branch was a plain string literal missing its `f` prefix:

```python
return f"{age.days} days ago" + (
    " ({datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')})" if incl_date else ""
)
```

so `{...}` was treated as literal text. Calling with `incl_date=True` produced:

```
3 days ago ({datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')})
```

This has stayed latent because the only caller (`gptme/cli/main.py`) invokes
`epoch_to_age(conv.modified)` without `incl_date`, and the existing test only
covered the `"just now"` / `"yesterday"` branches — never `incl_date=True` nor
the `"days ago"` branch — so CI stayed green.

## Fix

- Add the missing `f` prefix so the date is interpolated.
- Pass `tz=timezone.utc` to `datetime.fromtimestamp`, matching the timezone-aware
  age calculation a few lines above. This also satisfies ruff's `DTZ006`, which
  couldn't flag the call while it was still string data.

Now:

```
3 days ago (2026-07-01)
```

## Verification

All run in a clean `python:3.10-slim` container built from `scripts/Dockerfile.dev`:

- Added `test_epoch_to_age_incl_date`, which **fails on master** (asserts the
  literal `{datetime` template is absent) and **passes with this fix**.
- `ruff check` / `ruff format --check`: clean.
- `mypy gptme/util/__init__.py`: no new errors (the pre-existing
  `sentence_transformers` stub warning is unrelated and present on master).
- `pytest tests/test_util.py tests/test_codeblock.py tests/test_reduce.py tests/test_message.py`:
  115 passed.

## Notes / scope

Kept intentionally minimal. One adjacent observation, left out of this PR to
avoid scope creep: `incl_date` is only honored in the `"days ago"` branch and
silently ignored for more recent timestamps (minutes/hours/yesterday). Happy to
follow up separately if that behavior is desired.

### DIFF
--- a/gptme/util/__init__.py
+++ b/gptme/util/__init__.py
@@ -32,7 +32,9 @@ def epoch_to_age(epoch, incl_date=False):
     if age < timedelta(days=2):
         return "yesterday"
     return f"{age.days} days ago" + (
-        " ({datetime.fromtimestamp(epoch).strftime('%Y-%m-%d')})" if incl_date else ""
+        f" ({datetime.fromtimestamp(epoch, tz=timezone.utc).strftime('%Y-%m-%d')})"
+        if incl_date
+        else ""
     )
 
 
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -21,6 +21,16 @@ def test_epoch_to_age():
     assert epoch_to_age(epoch_yesterday) == "yesterday"
 
 
+def test_epoch_to_age_incl_date():
+    # regression: incl_date=True must interpolate the date, not return the
+    # literal f-string template (see gptme/util/__init__.py)
+    epoch = datetime.now(tz=timezone.utc).timestamp() - 3 * 24 * 60 * 60
+    result = epoch_to_age(epoch, incl_date=True)
+    assert "{datetime" not in result
+    expected_date = datetime.fromtimestamp(epoch, tz=timezone.utc).strftime("%Y-%m-%d")
+    assert result == f"3 days ago ({expected_date})"
+
+
 def test_transform_examples_to_chat_directives():
     src = """
 # Example


AmirF194:
Following up on @TimeToBuildBob's review notes on #3078, both points are already addressed in this PR:

- **tz consistency:** the interpolated date uses `datetime.fromtimestamp(epoch, tz=timezone.utc)`, matching the tz-aware age bucketing above, so the printed calendar date can't disagree with the "N days ago" bucket near a UTC midnight boundary.
- **test flake:** the new `test_epoch_to_age_incl_date` computes its expected date with the same `tz=timezone.utc`, so it won't flake in non-UTC CI environments.

Heads up on an overlap: #3080 landed shortly after this and makes the functionally identical change (`f` prefix + `tz=timezone.utc` + regression test). Happy to have whichever is preferred merged, glad to close this one if #3080 is the path forward.

TimeToBuildBob:
Thanks for the PR and for following up on the #3078 review notes! The fix looks correct — both the `f` prefix and `tz=timezone.utc` are what was needed.

However, #3080 (submitted shortly after yours) landed the functionally identical fix about 4 minutes ago. I'm going to close this as superseded by #3080 to avoid having two open PRs targeting the same bug.

Appreciate the contribution — please keep them coming!

TimeToBuildBob:
Superseded by #3080 which merged the same fix 4 minutes ago.

PR #3087, opened 2026-07-04
fix(subagent): fix eval checks and subagent_wait_basic test (#554)

TimeToBuildBob:
Two correctness fixes for the subagent system (#554):

### 1. `tests/test_tools_subagent.py` — `test_subagent_wait_basic`

Add `"clarification_needed"` to the set of valid expected statuses.

`clarification_needed` is a documented, valid `Status` value (defined in `types.py` as `Literal["running", "success", "failure", "clarification_needed", "timeout"]`) but was missing from the assertion. When a subagent responds to an ambiguous prompt with a `clarify` block (which is correct behavior), the test incorrectly failed.

### 2. `gptme/eval/suites/subagent.py` — eval spawn checks

`check_subagent_complete_spawned` and `check_clarification_spawned` only matched positional argument syntax (`subagent("greeter", ...)`). LLMs sometimes use keyword argument syntax (`subagent(agent_id="greeter", ...)`), which is equally valid but was not matched. This caused spurious eval failures even when the subagent was correctly spawned. Both checks now accept both forms.

## Test

```
poetry run pytest tests/test_tools_subagent.py tests/test_tool_use.py -m "not slow and not eval" -q
# 108 passed
```

Fixes part of #554.

### DIFF
--- a/gptme/eval/suites/subagent.py
+++ b/gptme/eval/suites/subagent.py
@@ -85,8 +85,8 @@ def check_subagent_complete_spawned(messages: list[Message]) -> bool:
     return (
         'subagent("sum-roundtrip"' in assistant_log
         or "subagent('sum-roundtrip'" in assistant_log
-        or 'agent_id="sum-roundtrip"' in assistant_log
-        or "agent_id='sum-roundtrip'" in assistant_log
+        or 'subagent(agent_id="sum-roundtrip"' in assistant_log
+        or "subagent(agent_id='sum-roundtrip'" in assistant_log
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:

Evidence: commit: d39b3e9332 docs(journal): pm gptme#3087 — Greptile P2 anchor fixes

Session: `6bd8cc51-0e48-5d22-af45-ec8c0671247a`

TimeToBuildBob:
@greptileai review

PR #3091, opened 2026-07-04
fix(subagent): report unfinished thread agents as "timeout" in wait_all()

AmirF194:
## What

`BatchJob.wait_all()` — the blocking wait behind `subagent_parallel()` (added in #3085) — reports an agent that's *still running* as `status="running"` with `result=None`, instead of the documented `status="timeout"`.

`subagent_wait()` returns a non-terminal `{"status": "running", "result": None}` for **thread/ACP** agents that are still alive after the wait: they can't be force-killed, so it falls back to `status() == "running"`. Meanwhile `_wait_one()` gives each per-agent wait a deadline ~1s shorter than the outer `as_completed()` timeout (`int()` truncation of the remaining time), so the future completes *normally* carrying `"running"` and the `FuturesTimeoutError` branch that marks stragglers as `"timeout"` never fires.

Two concrete symptoms:

- `wait_all()` violates its own documented contract — *"Agents that exceed this deadline are reported with status `"timeout"`"*.
- The `subagent_parallel()` docstring's **own** consumer example crashes:

  ```python
  print(f"{agent_id}: {result['status']} — {result['result'][:80]}")
  # TypeError: 'NoneType' object is not subscriptable
  ```

## Changes

- **`gptme/tools/subagent/batch.py`** — in `_wait_one()`, map a returned `"running"` status to `"timeout"` (with a message string), consistent with the existing `as_completed` timeout path. `"running"` is the only non-terminal status `subagent_wait()` can return (subprocess-mode stragglers are killed and yield a terminal status), so the mapping is complete.
- **`tests/test_subagent_unit.py`** — add `test_wait_all_maps_running_result_to_timeout`: mocks `subagent_wait()` to return the stuck-thread `{"status": "running", "result": None}` and asserts `wait_all()` reports `"timeout"` with a non-`None`, sliceable result. Fails on the pre-fix code.

## Testing

```
pytest tests/test_subagent_unit.py -q      # 144 passed
uvx ruff check / ruff format --check       # clean
```

Confirmed the new test fails without the fix and passes with it.

Relates to #554.

### DIFF
--- a/gptme/tools/subagent/batch.py
+++ b/gptme/tools/subagent/batch.py
@@ -50,10 +50,18 @@ class BatchJob:
             remaining = max(1, int(deadline - time.monotonic()))
             try:
                 result = subagent_wait(agent_id, timeout=remaining)
-                return agent_id, ReturnType(
-                    result.get("status", "failure"),
-                    result.get("result"),
-                )
+                status = result.get("status", "failure")
+                # subagent_wait() returns a non-terminal "running" status (with
+                # result=None) when a thread/ACP agent is still alive after the
+                # wait — these can't be force-killed. Report that as "timeout"
+                # per the documented contract, rather than leaking
+                # status="running"/result=None, which breaks callers that index
+                # into result (e.g. this function's own docstring example).
+                if status == "running":
+                    return agent_id, ReturnType(
+                        "timeout", f"Timed out after {timeout}s"
+                    )
+                return agent_id, ReturnType(status, result.get("result"))
             except Exception as e:
                 logger.warning(f"Error waiting for {agent_id}: {e}")
                 return agent_id, ReturnType("failure", str(e))
--- a/tests/test_subagent_unit.py
+++ b/tests/test_subagent_unit.py
@@ -3126,6 +3126,35 @@ class TestBatchJobWaitAll:
         assert results["agent-b"]["status"] == "timeout"
         assert "30s" in results["agent-a"]["result"]
 
+    def test_wait_all_maps_running_result_to_timeout(self):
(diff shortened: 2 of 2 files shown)


TimeToBuildBob (approved):
Clean fix. Reviewed locally by reading the commit directly.

**What the fix does**: maps 's non-terminal `status="running"` (returned for thread/ACP agents still alive after their wait timeout) to `status="timeout"` with a message string. This is the only non-terminal status `subagent_wait()` can return for thread agents (subprocess-mode stragglers are killed and yield a terminal status), so the mapping is complete and correct.

**Why the regression test is solid**: mocks `subagent_wait` to return the exact stuck-thread response `{"status": "running", "result": None}`, asserts `status == "timeout"` and `result[:80]` doesn't raise TypeError — which is precisely the contract violation and crash that this fixes.

**The fix is correct** — consistent with the existing `FuturesTimeoutError` path in `as_completed`, no side effects on other status values.

LGTM.

TimeToBuildBob (approved):
Clean fix. Reviewed locally by reading the commit directly.

What the fix does: maps subagent_wait() non-terminal status='running' (returned for thread/ACP agents still alive after their wait timeout) to status='timeout' with a message string. This is the only non-terminal status subagent_wait() can return for thread agents (subprocess-mode stragglers are killed, yielding a terminal status), so the mapping is complete.

The regression test is solid: mocks subagent_wait to return the exact stuck-thread response, asserts status=='timeout' and result[:80] does not raise TypeError -- which is exactly the contract violation and crash this fixes.

Consistent with the existing FuturesTimeoutError path in as_completed. LGTM.

PR #3095, opened 2026-07-04
feat(browser): add press_key, select_option, wait_for_element for computer-use (#216)

TimeToBuildBob:
## What

Adds three missing browser interaction primitives to the `browser` tool, needed for real-world computer-use tasks (the \"Can it Tweet?\" milestone and similar autonomous web workflows):

| Function | Purpose |
|----------|---------|
| `press_key(key)` | Dispatch keyboard events to the focused element — `Enter` to submit forms, `Tab` to navigate between fields, `Escape` to dismiss modals, `Control+a` for shortcuts |
| `select_option(selector, value)` | Select from `<select>` dropdowns by value attribute or visible label text |
| `wait_for_element(selector, timeout_ms=5000)` | Block until a DOM element becomes visible — essential for dynamic pages that load content asynchronously after a click |

## Why these are needed

The existing interactive browser API (`open_page`, `fill_element`, `click_element`, `scroll_page`) covers the basics but hits real walls on modern web apps:

- **`press_key`**: Twitter's tweet composer requires `Ctrl+Enter` to post; search forms often submit on `Enter`; autocomplete menus navigate with arrow keys
- **`select_option`**: Many forms use `<select>` dropdowns (country, category, size) that `fill_element` can't handle (it calls `fill()` which only works on `<input>`/`<textarea>`)
- **`wait_for_element`**: Dynamic SPAs (React/Vue) render elements asynchronously — clicking a button and immediately trying to interact with the result races the DOM

## Implementation

All three follow the same retry/error pattern as the existing functions:

- `_browser_playwright.py`: core implementation using Playwright's `keyboard.press()`, `locator().select_option()`, and `locator().wait_for(state="visible")`
- `browser.py`: thin wrapper that dispatches to the playwright implementation (with the same lynx fallback error for non-playwright backends)
- Registered in `ToolSpec.functions` so they appear in the model's tool list and docstrings

## Tests

Extended `test_computer_use_integration.py` with 5 new integration tests (all gated behind the existing playwright/chromium skip guard):

- `test_press_key_returns_snapshot` — Tab key doesn't crash, returns a snapshot
- `test_select_option_picks_dropdown_value` — selects \"news\" from the category dropdown; verified at the server
- `test_wait_for_element_finds_visible_element` — resolves immediately for an already-visible button
- `test_wait_for_element_finds_dynamically_shown_element` — resolves after a JS click reveals a hidden div
- `test_wait_for_element_raises_on_missing` — raises `RuntimeError` with \"did not appear\" for a nonexistent selector

The test form server gained a `<select id="category">` and a JS-revealed `<div id="dynamic-content">` to give the new tests real targets.

Closes part of #216

### DIFF
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -790,7 +790,10 @@ def _press_key(browser: Browser, key: str) -> str:
     if _current_page is None:
         raise RuntimeError("No page is open. Call open_page(url) first.")
     _current_page.keyboard.press(key)
-    _current_page.wait_for_timeout(200)
+    try:
+        _current_page.wait_for_load_state("domcontentloaded", timeout=5000)
+    except PlaywrightTimeoutError:
+        pass  # Timeout is fine — key press may not navigate
     return _page_snapshot()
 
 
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #3098, opened 2026-07-05
test(tools): regression test for execute_msg defensive error tool_result (#554)

TimeToBuildBob:
## Summary

Adds a regression test for the defensive behavior introduced in #3072: when a structured (API tool-format) `ToolUse` with a `call_id` is not runnable, `execute_msg` must yield an error `tool_result` instead of silently dropping it.

Silently dropping a structured `tool_use` leaves it dangling in the conversation — the Anthropic API rejects the next request with HTTP 400. This is the core crash that originally manifested in the thread-mode race condition reported in #554.

## What this tests

- **Structured tool_use (call_id present) for unknown tool → error tool_result is emitted** — the key regression test
- **Markdown-format tool_use (no call_id) for unknown tool → no output** — markdown blocks are not API tool_uses and must not produce a paired tool_result
- **Error result role must be `system`**
- **Multiple non-runnable structured tool_uses each get their own error result**

## Why a separate test file

`execute_msg`'s defensive pairing behavior is a correctness invariant that applies regardless of subagent state — it belongs in a focused test file rather than being buried in the subagent test suite.

## Test plan

- [x] `pytest tests/test_execute_msg_defensive.py` — all 4 pass
- [x] `pytest tests/test_subagent_unit.py tests/test_subagent_concurrency.py tests/test_subagent_integration.py` — all 205 pass (no regressions)

### DIFF
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -9,18 +9,9 @@ Markdown-format tool_uses (call_id is None) are NOT API tool_uses; they are
 intentionally left unpaired when the tool is unavailable.
 """
 
-import pytest
-
 from gptme.message import Message
 from gptme.tools import execute_msg
 from gptme.tools.base import set_tool_format
-
-
-@pytest.fixture(autouse=True)
-def restore_tool_format():
-    """Reset tool_format to 'markdown' after each test."""
-    yield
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #3101, opened 2026-07-05
fix(browser): drop unreachable select_option label fallback that doubles timeout

### LINKED ISSUE #3100: bug(browser): select_option label= fallback is unreachable and doubles the timeout on a miss
### What happens

`_select_option` in `gptme/tools/_browser_playwright.py` selects a dropdown option like this:

```python
locator.select_option(value=value, timeout=10000)
# on PlaywrightTimeoutError:
locator.select_option(label=value, timeout=10000)
```

The intent reads as "try by value, fall back to label", but Playwright's `value=` matcher already matches an option by its `value` attribute *or* its visible label text. So whatever `label=value` can match, `value=value` already matches — the label set is a strict subset. If the first call times out, the second call always times out too.

The only effect of the fallback is to double the wait before the error surfaces: 10s + 10s = 20s.

### Repro

`value=` matching label text (real value attribute is `aaa`, label is `Large`):

```python
from playwright.sync_api import sync_playwright
html = '<select name=s><option value=aaa>Large</option><option value=bbb>Small</option></select>'
with sync_playwright() as p:
    b = p.chromium.launch(); pg = b.new_page(); pg.set_content(html)
    pg.locator("[name=s]").select_option(value="Large", timeout=3000)
    print(pg.eval_on_selector("[name=s]", "e=>e.value"))   # -> "aaa"  (label matched via value=)
```

Timing the actual code path:

```
A) value matches value attr  -> instant
B) value matches label text  -> instant (value= matches label)
C) value matches neither      -> raises after ~20s (two 10s timeouts)
```

### Expected

A genuine miss should fail after one timeout (~10s), not two. There is no case the `label=` call can rescue.

### Proposed fix

Drop the fallback and make a single call:

```python
_current_page.locator(selector).select_option(value=value, timeout=10000)
```

Behavior is identical for every value that currently succeeds (value= already covers label text), and a real miss surfaces in 10s instead of 20s.

Happy to send a PR with a regression test. Noticed while reviewing the eval specs in #3097.

Environment: gptme master (9de8a92), Python 3.12, Playwright 1.60.0 locally (repo pins 1.61.0), macOS.

AmirF194:
## Root cause

`_select_option` tried by value, then fell back to label on timeout:

```python
locator.select_option(value=value, timeout=10000)
# on PlaywrightTimeoutError:
locator.select_option(label=value, timeout=10000)
```

Playwright's `value=` matcher already matches an option by its `value` attribute or its visible label text, so the set of options `label=value` can match is a subset of what `value=value` matches. If the first call times out, the second one always times out too. The fallback rescues nothing and only doubles the wait before the error surfaces (10s + 10s = 20s).

Verified on Playwright 1.60.0 (repo pins 1.61.0): `select_option(value="Large")` selects the option whose value attribute is `aaa` and label is `Large`, and a value matching neither takes ~20s to raise.

## Fix

Collapse to a single call:

```python
_current_page.locator(selector).select_option(value=value, timeout=10000)
```

Behavior is identical for every value that currently succeeds, and a genuine miss now surfaces in 10s instead of 20s.

## Verification

- Added `TestSelectOption` with a regression test asserting exactly one `select_option` attempt on a miss. It fails on the old code (2 calls) and passes with the fix. Second test covers the success path.
- `pytest tests/test_tools_browser.py` — 106 passed.
- `ruff check` / `ruff format --check` clean.

Fixes #3100

### DIFF
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -828,11 +828,12 @@ def _select_option(browser: Browser, selector: str, value: str) -> str:
     """Select an option from a <select> element on the current page."""
     if _current_page is None:
         raise RuntimeError("No page is open. Call open_page(url) first.")
-    locator = _current_page.locator(selector)
-    try:
-        locator.select_option(value=value, timeout=10000)
-    except PlaywrightTimeoutError:
-        locator.select_option(label=value, timeout=10000)
+    # Playwright's value= matcher already matches by the option's value
+    # attribute *or* its label text, so a single call handles both. A label=
+    # fallback is unreachable: whatever label= would match, value= already
+    # matches, so if value= times out label= times out too — its only effect
+    # is doubling the wait (10s -> 20s) before the error surfaces.
+    _current_page.locator(selector).select_option(value=value, timeout=10000)
     return _page_snapshot()
 
 
--- a/tests/test_tools_browser.py
+++ b/tests/test_tools_browser.py
@@ -1041,3 +1041,45 @@ class TestCreatePage:
         browser.new_context.assert_called_once_with(locale="en-US")
         assert managed.page is page
         assert managed._owned_context is context
+
+
+@pytest.mark.skipif(not has_playwright(), reason="playwright not installed")
+class TestSelectOption:
+    """select_option should not double the timeout via an unreachable fallback."""
+
+    def test_single_attempt_when_value_matches_nothing(self, monkeypatch):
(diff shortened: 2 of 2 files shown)


AmirF194:
Superseded by #3103, which landed the same single-call fix and closed #3100. Closing this one. Thanks for the fast turnaround.

One small follow-up on the docstring wording that #3103 settled on, posted over on that PR since that's where it was decided.

PR #3102, opened 2026-07-05
fix(subagent): isolate thread tool list via clear_tools() at entry (#554)

TimeToBuildBob:
## Root cause

Python's `threading.Thread` copies the parent's `ContextVar` context into the child, so `_loaded_tools_var` initially refers to the *same list object* as the parent. Before this fix, `init_tools()` and `_ensure_subagent_signal_tools_loaded()` would append signal tools to that shared list from the subagent thread, creating a data race with the parent's concurrent `execute_msg()` calls.

The race: while the parent's `execute_msg()` was checking `is_runnable` on a tool use, the subagent thread was appending new tools to the same list. The intermittent result was that existing tools (e.g. `read`) transiently appeared non-runnable — `get_tool("read")` returned `None` during the brief mutation window. This caused the parent's next API request to get a hard 400 (`tool_use` without a paired `tool_result`).

PR #3072 made the crash non-fatal (defensive `tool_result` on non-runnable tool use). This commit fixes the root cause.

## Fix

Call `clear_tools()` at the very start of `_create_subagent_thread`, before any tool initialization. This calls `_loaded_tools_var.set([])` in the **child thread's** context, replacing the ContextVar binding with a fresh empty list fully decoupled from the parent's. All subsequent operations (`prepare_execution_environment` → `init_tools`, `_ensure_subagent_signal_tools_loaded`, `set_tools`) work on the child's own list and never touch the parent's.

## Test

The regression test proves both sides:
1. **Pre-fix behavior was real** — a child thread that inherits the parent's context and appends to `get_tools()` does mutate the parent's list.
2. **Fix prevents it** — after `clear_tools()`, the child's list is independent and the parent's list is unchanged.

## Closes

Fixes the "transient non-runnable" root cause from #554. Combined with #3072 (defensive pairing) and #3098 (regression test for the pairing), this should close the thread-mode race track of the issue.

### DIFF
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -445,11 +445,18 @@ def get_available_tools(include_mcp: bool = True) -> list[ToolSpec]:
 
 
 def clear_tools():
-    """Clear all context-local tool state."""
+    """Clear all context-local tool state.
+
+    Resets the ContextVar-bound tool list so the current context has a
+    fresh empty list, fully decoupled from any other context (parent
+    thread, sibling thread, etc.).
+
+    Does NOT clear module-global state like _warned_mcp_allowlists, which
+    is shared across all contexts for log-deduplication. Only the
+    context-local tool list and its cache are reset.
+    """
     _set_available_tools_cache(None)
     _loaded_tools_var.set([])
-    with _warned_mcp_allowlists_lock:
-        _warned_mcp_allowlists.clear()
 
 
 def get_tools() -> list[ToolSpec]:
--- a/gptme/tools/subagent/execution.py
+++ b/gptme/tools/subagent/execution.py
@@ -19,7 +19,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
 from ...message import Message
-from .. import get_tools, load_tool, set_tools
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
Addressed the two review items from Greptile's review:

**1. Unused `child_appended_to_parent` variable (P2)**

Added assertion `assert child_appended_to_parent == [False]` to directly verify the child's `get_tools().append()` after `clear_tools()` went to the child's own list, not the parent's. Previously this was only indirectly verified by the `len(get_tools()) == parent_len` check.

**2. `clear_tools()` clearing process-global `_warned_mcp_allowlists`**

Removed the `_warned_mcp_allowlists.clear()` call from `clear_tools()`. The `_warned_mcp_allowlists` set is a module-level global shared across all contexts (parent + all subagent threads) — clearing it from a subagent thread would re-enable MCP allowlist warnings in the parent. `clear_tools()` should only reset context-local state (the ContextVar-bound tool list). Updated the docstring to make this contract explicit.

Verification: all 87 subagent tests, 37 tool init tests, and 199 MCP tests pass.

TimeToBuildBob:
@greptileai review

PR #3103, opened 2026-07-05
fix(browser): drop unreachable label= fallback in select_option

### LINKED ISSUE #3100: bug(browser): select_option label= fallback is unreachable and doubles the timeout on a miss

TimeToBuildBob:
## Problem

`_select_option` in `gptme/tools/_browser_playwright.py` wrapped its `select_option(value=...)` call in a try/except and retried with `select_option(label=...)` on timeout. But Playwright's `value=` parameter already matches options by **both** their `value` attribute **and** their visible label text — so the fallback could never succeed where the first call already failed. The only observable effect was doubling the wait on a miss: 10 s + 10 s = 20 s.

## Fix

Drop the try/except and make a single call:

```python
locator.select_option(value=value, timeout=10000)
```

Behavior is identical for every value that currently succeeds (label text was already matched by `value=`). A genuine miss now surfaces in 10 s instead of 20 s.

## Tests

Added `TestSelectOption` in `tests/test_tools_browser.py` (runs when playwright is installed):

- `test_select_option_value_calls_select_once` — asserts `select_option` is called exactly once on a successful match.
- `test_select_option_timeout_propagates_after_one_call` — asserts `PlaywrightTimeoutError` propagates after a single call with no retry.

Closes #3100.

### DIFF
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -829,10 +829,10 @@ def _select_option(browser: Browser, selector: str, value: str) -> str:
     if _current_page is None:
         raise RuntimeError("No page is open. Call open_page(url) first.")
     locator = _current_page.locator(selector)
-    try:
-        locator.select_option(value=value, timeout=10000)
-    except PlaywrightTimeoutError:
-        locator.select_option(label=value, timeout=10000)
+    # Playwright's value= already matches by both value attribute and visible
+    # label text, so a label= fallback would never succeed where value= fails.
+    # Drop the fallback to avoid doubling the timeout on a miss.
+    locator.select_option(value=value, timeout=10000)
     return _page_snapshot()
 
 
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -829,9 +829,11 @@ def _select_option(browser: Browser, selector: str, value: str) -> str:
     if _current_page is None:
         raise RuntimeError("No page is open. Call open_page(url) first.")
     locator = _current_page.locator(selector)
-    # Playwright's value= already matches by both value attribute and visible
-    # label text, so a label= fallback would never succeed where value= fails.
-    # Drop the fallback to avoid doubling the timeout on a miss.
+    # The removed label= fallback was unreachable: PlaywrightTimeoutError is
+    # raised when the locator cannot resolve (element absent), in which case
+    # label= on the same locator would also time out.  A no-option-match
+    # raises a different error that the except clause never caught.
+    # Drop the dead fallback to surface failures in 10 s instead of 20 s.
     locator.select_option(value=value, timeout=10000)
     return _page_snapshot()
 
@@ -840,12 +842,12 @@ def select_option(selector: str, value: str) -> str:
     """Select an option from a <select> dropdown on the current page.
 
     Finds the ``<select>`` element and selects the option matching ``value``
-    by its ``value`` attribute or visible text.
+    by its ``value`` attribute.
 
     Args:
         selector: Playwright selector for the ``<select>`` element
                   (e.g. ``"select#country"``, ``"[name='size']"``).
-        value: The option value (``value`` attribute) or label text to select.
+        value: The option value (``value`` attribute) to select.
 
     Returns:
         Updated ARIA snapshot of the page after the selection.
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Greptile review feedback addressed

Both P2 findings from Greptiles review are fixed in `4d2f94648`:

1. **Inline comment** (line 832-835): Updated with accurate explanation — `PlaywrightTimeoutError` fires when the locator cannot resolve (element absent), not on no-option-match. A no-option-match raises a different error type that the old `except` clause never caught.

2. **Public docstring** (lines 841-848): Removed claim that `select_option` accepts visible label text — Playwrights `value=` keyword argument matches only the HTML `value` attribute.

No functional changes. CI pending on the new head.

TimeToBuildBob:
The failing CI check (`Test without API keys or extras`) failed on `test_get_toolchain_warns_when_plain_allowlist_excludes_mcp_tools` — a pre-existing flaky test unrelated to this PR (which only touches `_browser_playwright.py` and `test_tools_browser.py`). The test expects `assert caplog.text.count(...) == 1` but `get_toolchain` is called twice, so on certain test orderings it sees 2 warnings instead of 1.

The test passes when run in isolation locally. I've re-triggered the failed CI runner.

TimeToBuildBob:
@greptileai review

AmirF194:
Thanks for turning this around so fast. One thing on the second (docs) commit: I think the original wording, and this PR's description, was actually right, and the P2 revision moved it away from the real behavior.

Playwright's `value=` does match by the option's visible label text, not only its `value` attribute. Verified on Playwright 1.61.0 (the pinned version) in a clean container, with the target set to a non-default option so the change is unambiguous:

```
# <option value="1">Small</option>   (default-selected)
# <option value="2">Large</option>
select_option(value="Large")   # "Large" is the label; the value attr is "2"
# -> selects value "2", no timeout

select_option(value="Nope")    # matches neither value nor label
# -> raises PlaywrightTimeoutError after a single timeout
```

So on a no-option-match it's a `PlaywrightTimeoutError` that fires, not a different error type. That's also the real reason the `label=` fallback was unreachable: `value=` already covers the label set, so once `value=` times out, `label=` times out too (matches your first commit and the PR body).

Two concrete things left over from the docs commit:

- The wrapper docstring in `_browser_playwright.py` now says "by its `value` attribute", but the public `select_option` docstring in `browser.py` still says "value attribute or visible text" - they disagree now.
- The inline comment's "a no-option-match raises a different error that the except clause never caught" doesn't match what I see; it's a `PlaywrightTimeoutError`.

If you're deliberately not wanting to advertise label matching (fair, if you'd rather not depend on it), the minimal fix is just to make the two docstrings agree and drop the "different error" line. If you're fine documenting it, I'm happy to send a small follow-up that restores "value attribute or label text" and adds a real-browser test: a data: URL that selects a non-default option by its label, under the `slow` marker so CI actually exercises it. Your call.

PR #3106, opened 2026-07-05
fix(subagent): delete worktree branch on cleanup (#554)

TimeToBuildBob:
## Problem

`cleanup_worktree()` called `git worktree remove` to clean up isolated subagent workspaces, but `git worktree remove` only removes the **working tree directory** — it leaves the associated branch behind. Every `isolated=True` subagent run was permanently polluting the repo with dangling branches like `subagent-checker-2ce95b0b`, `subagent-thread-agent-15a1d462`, etc. (confirmed visible in the repo right now).

## Fix

After a successful worktree remove, also run `git branch -D <branch-name>`. The branch name is inferred from `worktree_path.name`, which matches the naming convention in `create_worktree()` (`worktree_path = worktree_base / branch_name`).

Branch deletion is best-effort: failure is logged at DEBUG level and does not block cleanup. This handles edge cases like manually-deleted branches or branches checked out elsewhere.

Also restructured the fallback path (directory-only removal when `git worktree remove` fails) so directory removal and branch deletion are independent, and `git worktree prune` always runs.

## Test

Extended `test_cleanup_worktree` to assert that the branch is gone after `cleanup_worktree()` — the previous test only checked the directory and worktree list, not branch deletion.

Closes #554 (partial — fixes branch pollution from worktree isolation)

### DIFF
--- a/gptme/util/git_worktree.py
+++ b/gptme/util/git_worktree.py
@@ -146,7 +146,7 @@ def cleanup_worktree(
                 logger.debug(
                     f"Could not delete branch {branch_name!r}: {result.stderr.strip()}"
                 )
-        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+        except (subprocess.TimeoutExpired, OSError) as e:
             logger.debug(f"Branch deletion skipped for {branch_name!r}: {e}")
 
         # Prune stale worktree entries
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3108, opened 2026-07-05
feat(browser): add load_browser_state() for in-session auth state restoration

TimeToBuildBob:
## Summary

Adds `load_browser_state(path)` as the callable complement to `save_browser_state(path)`, closing a usability gap in the computer-use browser tool (#216).

**Before**: Restoring a saved session required setting `GPTME_BROWSER_STORAGE_STATE` *before* process startup — no way to switch auth state mid-session without a restart.

**After**: `load_browser_state("~/twitter-session.json")` + `open_page(url)` reloads the session in-place, no restart required.

## Changes

- **`_browser_thread.py`**: `set_storage_state_override()` + 3-level priority in `get_context_options()`:
  in-session override > `GPTME_BROWSER_STORAGE_STATE` env var > fresh context
- **`_browser_playwright.py`**: `_do_load_browser_state()` closes the current page and sets the override;
  public `load_browser_state()` wraps it via `_execute_with_retry()` for thread safety
- **`browser.py`**: Public `load_browser_state()`, registered in `ToolSpec`, updated `instructions_format`
- **`eval/suites/computer.py`**: New `computer-use-web-session-persistence` eval spec with
  `check_used_save_browser_state` / `check_used_load_browser_state` trajectory helpers

## Tests

- `tests/test_browser_state.py` — 9 unit tests: override priority, env-var fallback, tilde expansion, missing-file error, tool-spec registration, round-trip
- `tests/test_eval_computer_suite.py` — 4 tests for the new trajectory check helpers

## Typical workflow now

```python
# First time: log in and save
open_page("https://x.com/login")
fill_element("#username", "me@example.com")
fill_element("#password", "hunter2")
click_element("text=Log in")
save_browser_state("~/.config/gptme/twitter-session.json")

# Later in the same session (or a new one):
load_browser_state("~/.config/gptme/twitter-session.json")
open_page("https://x.com")   # opens with saved cookies — no restart needed
```

Closes #216

### DIFF
--- a/gptme/eval/suites/computer.py
+++ b/gptme/eval/suites/computer.py
@@ -183,6 +183,16 @@ def check_used_get_current_url(messages: list[Message]) -> bool:
     return any("get_current_url(" in code for code in _executed_tool_calls(messages))
 
 
+def check_used_save_browser_state(messages: list[Message]) -> bool:
+    """Agent must use save_browser_state() to persist the browser session."""
+    return any("save_browser_state(" in code for code in _executed_tool_calls(messages))
+
+
+def check_used_load_browser_state(messages: list[Message]) -> bool:
+    """Agent must use load_browser_state() to restore a browser session."""
+    return any("load_browser_state(" in code for code in _executed_tool_calls(messages))
+
+
 def check_did_not_screenshot_for_web(messages: list[Message]) -> bool:
     """Structured-first policy: screenshots should NOT be the first observation for web."""
     calls = _executed_tool_calls(messages)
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Fixed: CI test failure when playwright is not installed

Added `pytest.importorskip("playwright")` at module level in `tests/test_browser_state.py` to skip the 9 browser state tests when playwright is not available. This matches the pattern used in `test_browser.py` and 44+ other test files.

**Why**: The "Test without API keys or extras" CI runner doesn't install playwright. Previously, importing `_browser_thread.py` (which has a module-level `from playwright.sync_api import ...`) would raise `ModuleNotFoundError`. The `importorskip` prevents the entire test module from loading when playwright is absent.

**Greptile findings status** (all resolved in earlier commits):
- P1 - CDP mode silent no-op: ✅ Fixed in `8910a0bad` — closes and re-creates the CDP session context with loaded state
- P2 - `_expect_state_file_written` dead code: ✅ Already wired into the eval spec's `expect` dict
- P2 - Unused `browser` parameter: ✅ Now actually used in the CDP refresh path (`browser.new_context(...)`)

PR #3114, opened 2026-07-05
fix(subagent): apply output_schema parsing in subagent_wait() (#554)

TimeToBuildBob:
## Summary

- Makes `subagent_wait()` apply `output_schema` parsing automatically when the spawned subagent had `output_schema` set
- Closes the inconsistency where `subagent_parallel()` and `BatchJob.wait_all()` auto-parse results but `subagent_wait()` returned raw JSON strings
- Adds 4 unit tests covering: successful parse, no-schema passthrough, invalid JSON (parse_error key), and failure-status skip

## Motivation

After #554's improvements, users can pass `output_schema` to `subagent()` to get structured output. But the single-agent workflow was inconsistent:

```python
# subagent_parallel() auto-parses — callers get a dict:
results = subagent_parallel([('agent', 'prompt')], output_schema=MySchema, timeout=60)
# results[0]['result'] is a parsed dict

# subagent_wait() did NOT auto-parse — callers got raw JSON string:
subagent('agent', 'prompt', output_schema=MySchema)
result = subagent_wait('agent')
# result['result'] was still '{key: value}' — callers had to manually validate
```

Now both paths behave consistently. The late import of `_parse_result` from `batch.py` (inside the function body) avoids the circular dependency `batch → api → batch`.

## Test coverage

`TestSubagentWaitOutputSchema` (4 tests in `tests/test_subagent_unit.py`):
- `test_wait_parses_json_when_output_schema_set` — success path returns parsed dict
- `test_wait_returns_raw_result_when_no_schema` — no schema → passthrough unchanged
- `test_wait_adds_parse_error_on_invalid_json` — bad JSON → raw result + parse_error key
- `test_wait_skips_parsing_on_failure_status` — failure results skipped (matches batch behavior)

### DIFF
--- a/gptme/tools/subagent/api.py
+++ b/gptme/tools/subagent/api.py
@@ -1138,11 +1138,25 @@ def subagent_wait(
     status = cached_status if cached_status is not None else sa.status()
     result_dict = asdict(status)
 
+    # Apply output_schema parsing before truncation — the parsed object may be
+    # larger or smaller than the raw JSON string, so we measure truncation against
+    # the final result text, not the pre-parse raw.
+    if sa.output_schema is not None and result_dict.get("status") == "success":
+        from .batch import (
+            _parse_result,  # late import avoids circular (batch imports api)
+        )
+
+        result_dict = _parse_result(result_dict, sa.output_schema)
+
     # Compact result: truncate long outputs so they don't flood the parent's context.
     # The complete block is meant to be a brief summary; if it's longer than
     # max_result_chars the parent agent can call subagent_read_log() to get details.
     result_text = result_dict.get("result")
-    if result_text and max_result_chars > 0 and len(result_text) > max_result_chars:
+    if (
+        isinstance(result_text, str)
+        and max_result_chars > 0
+        and len(result_text) > max_result_chars
+    ):
         result_dict["result"] = (
             result_text[:max_result_chars]
             + f"\n... [truncated — call subagent_read_log('{agent_id}') for full output]"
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the double-parse regression Greptile identified.

**Root cause**: `subagent_wait()` now parses `output_schema` for direct callers (correct), but `subagent_parallel()` and `BatchJob.wait_all()` also call `_parse_result()` on the result afterwards. `json.loads(already_parsed_dict)` raises `TypeError`, caught by the broad `except`, injecting a spurious `parse_error` into every successfully-parsed result.

**Fix** (`gptme/tools/subagent/batch.py`): made `_parse_result()` idempotent — if `result["result"]` is already not a string (i.e. already parsed), return early. This is the minimal, safe fix: no change to the `subagent_wait()` path, no caller-side coordination needed.

**Added test** (`tests/test_subagent_unit.py::TestParseResult::test_already_parsed_dict_returned_as_is`): exercises the double-parse path directly — confirms no `parse_error` injected and result dict passes through unchanged.

Re-triggered Greptile for a fresh review.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed typecheck failures (4 union-attr errors).

The widened `ReturnType.result` type (`str | dict[str, object] | None`) requires test code to narrow the type before calling string methods like `.lower()`, `.split()`, and `.startswith()`. Added type guards to the affected test locations:

- `tests/test_tools_subagent.py:1008` — added `isinstance(result.result, str)` check before `.lower()`
- `tests/test_subagent_unit.py:3481` — added `isinstance(result.result, str)` check before `.split()`
- `tests/test_subagent_unit.py:3506` — added `isinstance(result.result, str)` check before `.startswith()`

Pushed as commit `ebbf0b65d`. CI should now pass.

PR #3121, opened 2026-07-05
test(subagent): add regression tests for thread-mode tool isolation (#554)

TimeToBuildBob:
## What

Adds `TestThreadToolIsolation` with three regression tests for the thread-mode tool isolation fix shipped in #3102.

## Why

PR #3102 added `clear_tools()` at the start of `_create_subagent_thread` to prevent a data race: in Python ≤ 3.11, `threading.Thread` copies the parent's ContextVar mapping into the child, so `_loaded_tools_var` initially points to the **same list object** as the parent. Without `clear_tools()`, any `list.append()` inside `init_tools()` or `_ensure_subagent_signal_tools_loaded()` would mutate the parent's list — making tools transiently non-runnable (`is_runnable → False`) during the subagent's setup window, causing the `subagent-complete-roundtrip` eval's hard crash.

The fix is correct but had **no regression test**. If someone accidentally removes the `clear_tools()` call or moves it after the first `_get_loaded_tools()` call, no test would catch it.

## How

All three tests use `copy_context().run()` to simulate Python ≤ 3.11 threading semantics on any Python version (in 3.12+ threads start with a fresh context and don't exhibit the race natively):

- **`test_mutation_visible_without_clear_tools`** — documents the bug: a child running in a copied context CAN mutate the parent's list without `clear_tools()`. This test passes *because* the mutation happens, proving the test harness works.

- **`test_clear_tools_prevents_parent_mutation`** — verifies `clear_tools()` at thread entry breaks the shared-list binding so subsequent appends never touch the parent's list.

- **`test_create_subagent_thread_isolates_tool_list`** — the actual regression guard for the fix site at `execution.py:188`. Exercises `_create_subagent_thread` directly in a copied context with `_ensure_subagent_signal_tools_loaded` replaced by a mock that appends to the current tool list. Without `clear_tools()` at the real fix site, the append reaches the parent's list; with it, the child operates on an isolated empty list. Addresses Greptile's 3/5 finding that the first two tests don't guard the fix site.

## Tests

```
tests/test_subagent_unit.py::TestThreadToolIsolation::test_mutation_visible_without_clear_tools PASSED
tests/test_subagent_unit.py::TestThreadToolIsolation::test_clear_tools_prevents_parent_mutation PASSED
tests/test_subagent_unit.py::TestThreadToolIsolation::test_create_subagent_thread_isolates_tool_list PASSED
```

Closes part of #554.

### DIFF
--- a/tests/test_subagent_unit.py
+++ b/tests/test_subagent_unit.py
@@ -4668,3 +4668,111 @@ class TestThreadToolIsolation:
         assert len(current_parent) == 1 and current_parent[0] is sentinel, (
             "Parent's list content must be unchanged after child thread exits"
         )
+
+    def test_create_subagent_thread_isolates_tool_list(self, monkeypatch, tmp_path):
+        """_create_subagent_thread must isolate its tool list from the parent thread.
+
+        This is the actual regression guard for the fix at execution.py:188.
+        The two tests above verify that clear_tools() works in isolation, but
+        neither calls _create_subagent_thread — so removing clear_tools() from
+        execution.py:188 would leave both passing.  This test exercises the real
+        fix site: the child runs in a *copied* context (simulating Python ≤ 3.11
+        thread semantics), and _ensure_subagent_signal_tools_loaded is replaced
+        with a mock that appends to the current tool list.  Without clear_tools(),
+        that append would mutate the parent's list; with it, the child gets an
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

PR #3090, opened 2026-07-04
feat(subagent): enforce output_schema at the complete-block layer

TimeToBuildBob:
## Summary

- Thread-mode subagents were passing `output_schema` to `chat()` which forces OpenAI's `response_format=json` — this conflicts with `tool_format="markdown"` because the model can't write markdown code blocks while simultaneously being forced to output raw JSON. The net effect was that `output_schema` silently did the wrong thing.
- Fix: don't pass `output_schema` to `chat()` in thread mode; enforce the JSON output contract at the complete-block layer instead.

## Changes

**`gptme/tools/subagent/hooks.py`** — `_get_complete_instruction()` gains an optional `output_schema` parameter. When set, the JSON schema is embedded in the instruction and an IMPORTANT warning is appended so the subagent knows it must output valid JSON in the complete block.

**`gptme/tools/subagent/types.py`** — New `Subagent._normalize_json_result()` method normalizes valid JSON syntax (canonical re-serialization) and logs a warning for invalid JSON without hard-failing. It intentionally does not perform Pydantic schema validation. `_read_log()` calls it only for non-empty complete-block results, avoiding misleading invalid-JSON warnings for empty completion fallbacks.

**`gptme/tools/subagent/execution.py`** — Removed `output_schema=output_schema` from the `chat()` call in `_create_subagent_thread()`; passes it to `_get_complete_instruction()` instead. Comment explains the design decision.

**`tests/test_subagent_unit.py`** — Adds coverage for schema instruction content, placeholder replacement, JSON normalization pass/fail/no-schema paths, `_read_log()` integration with valid/invalid JSON, and empty complete-block fallback behavior without spurious JSON warnings.

## Test plan

- [x] `uv run --with pytest python -m pytest tests/test_subagent_unit.py -q` — 156 passed
- [x] `uv run --with pytest --with pytest-cov --with greenlet python -m pytest tests/test_subagent_unit.py -q --cov=gptme.tools.subagent.types --cov-report=term-missing` — 156 passed
- [x] `uv run ruff format tests/test_subagent_unit.py && uv run ruff check tests/test_subagent_unit.py`
- [x] Pre-commit hooks — passed during `git commit`

Closes part of #554 (the `output_schema` contract enforcement slice).

PR #3124, opened 2026-07-05
eval(subagent): trajectory evals for output_schema and subagent_pipeline (#554)

TimeToBuildBob:
## Summary

Adds two new trajectory-level eval specs and their supporting `check_log` functions, directly addressing the final request in issue #554:

> "Would be interesting adding an eval/test or two for the subagents (would need to evaluate the trajectory, not the outcome/artifacts) to ensure they are actually better/good now." — @ErikBjare

### New eval specs

- **`subagent-output-schema`**: verifies the parent passes `output_schema=` when spawning a subagent, calls `subagent_wait`, and incorporates the resulting structured dict (keyed on `score`) into its final answer. Covers the output_schema contract from #3090/#3114.

- **`subagent-pipeline-staged`**: verifies the parent uses `subagent_pipeline()` for two-stage fan-out without manually calling `subagent_wait`. Covers the barrier-free scheduling contract from #3110.

### New check functions

Six named module-level functions (not lambdas — lambdas fail `pickle.dumps`):

| Function | What it checks |
|---|---|
| `check_output_schema_used` | `output_schema=` appears in a `subagent()` call |
| `check_output_schema_wait_called` | `subagent_wait()` is called after spawning |
| `check_output_schema_result_is_structured` | final message contains a `score:` or `score=` field |
| `check_pipeline_used` | `subagent_pipeline(` appears in the trajectory |
| `check_pipeline_results_integrated` | final message contains `REVIEWED=` or `SUMMARY=` |
| `check_pipeline_no_explicit_waits` | no `subagent_wait(` calls (pipeline manages its own scheduling) |

### Tests

6 new unit tests for the check functions (pass + fail cases), including:
- `test_output_schema_structured_check_requires_score_field` — verifies the check rejects free-text output that lacks a structured field
- `test_pipeline_check_fails_when_subagent_wait_called_manually` — verifies the no-wait invariant catches misuse

All 232 subagent tests pass (215 unit + 17 eval unit).

Closes the eval coverage gap for #554.

### DIFF
--- a/gptme/eval/suites/subagent.py
+++ b/gptme/eval/suites/subagent.py
@@ -1,5 +1,6 @@
 """Trajectory-focused evals for the subagent tool."""
 
+import re
 from typing import TYPE_CHECKING
 
 from gptme.message import Message
@@ -35,6 +36,14 @@ def _expect_sum_marker(ctx: "ResultContext") -> bool:
 
 def _expect_greeting_marker(ctx: "ResultContext") -> bool:
     return "GREETING=" in ctx.stdout
+
+
+def _expect_score_marker(ctx: "ResultContext") -> bool:
+    return "SCORE=" in ctx.stdout
+
+
+def _expect_reviewed_marker(ctx: "ResultContext") -> bool:
+    return "REVIEWED=" in ctx.stdout
 
 
 def _role_contents(messages: list[Message], role: str) -> str:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Thanks for the careful review. I've addressed the findings:

**1. False positive from `output_schema` definition:** Strengthened the negative lookahead to check for delimiter characters (`,`, `}`, `]`) after type keywords. This ensures we only match schema definitions when they're followed by other dict/list content, not actual result values.

**2. Misleading case-sensitivity comment:** Clarified that the check requires dict/list delimiters (the structural property), not case-sensitivity alone. Updated both the docstring and test comment to be explicit about what triggers false positives.

**3. Missing negative test for `check_output_schema_wait_called`:** Already present as `test_output_schema_wait_check_fails_without_wait_call` — verifies the check correctly returns False when `subagent_wait` is omitted.

Pushed fresh commit with these refinements. CI should confirm the tests and coverage updates.

PR #3128, opened 2026-07-06
fix(info): fall back to pyproject.toml when editable install omits Provides-Extra (#216)

TimeToBuildBob:
## Summary

`gptme doctor` and `gptme-util computer doctor` were silently showing no Python optional-dependency results for dev/editable installs (Poetry, `uv sync`). These tools showed the extras section blank because `get_installed_extras()` returned `[]` — the root cause was that Poetry/uv editable installs don't populate `Provides-Extra` in the package metadata.

This matters for computer-use setup: `gptme-util computer doctor` is how users verify the `browser` and `computer` extras are available, and it was broken in the most common dev-install scenario.

- **`gptme/info.py`**: add `_parse_extras_from_pyproject()` that reads `[tool.poetry.extras]` from the source pyproject.toml via `direct_url.json` when metadata yields empty extras. Falls back gracefully on any error.
- **`tests/test_doctor.py`**: add `test_pyproject_fallback_used_when_metadata_empty` to prevent regression.

## Test plan

- [x] `pytest tests/test_doctor.py::TestCheckPythonDeps` — all 3 tests pass (was 2 failures before)
- [x] `pytest tests/test_doctor.py` — all 75 tests pass
- [x] `gptme-doctor` output shows browser/computer/dspy extras in dev install
- [x] `gptme-util computer doctor` works correctly

### DIFF
--- a/gptme/info.py
+++ b/gptme/info.py
@@ -77,12 +77,12 @@ def _parse_extras_from_pyproject(pyproject_path: Path) -> list[ExtraInfo]:
     try:
         with open(pyproject_path, "rb") as f:
             data = tomllib.load(f)  # type: ignore[attr-defined]
-    except (OSError, Exception):
+    except Exception:
         return []
 
-    extras_dict: dict[str, list[str]] = (
-        data.get("tool", {}).get("poetry", {}).get("extras", {})
-    )
+    extras_dict: dict[str, list[str]] = data.get("tool", {}).get("poetry", {}).get(
+        "extras", {}
+    ) or data.get("project", {}).get("optional-dependencies", {})
     if not extras_dict:
         return []
 
@@ -126,9 +126,7 @@ def _parse_extras_from_metadata() -> list[ExtraInfo]:
         try:
             direct_url = dist.read_text("direct_url.json")
             if direct_url:
-                import json as _json
-
-                url_data = _json.loads(direct_url)
+                url_data = json.loads(direct_url)
                 dir_info = url_data.get("dir_info", {})
                 if dir_info.get("editable"):
                     src_url = url_data.get("url", "")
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the three Greptile findings in f6584a5:

1. **Redundant `OSError` in except clause** — changed `except (OSError, Exception)` to `except Exception` since `OSError` is already a subclass of `Exception`.

2. **Redundant `import json as _json`** — removed the inner import; the module-level `json` is already in scope.

3. **Missing PEP 621 fallback** — added `[project.optional-dependencies]` as a fallback when `[tool.poetry.extras]` is empty, so the fix also covers `uv`-managed projects not using the Poetry plugin.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Two additional P1 findings from the subsequent Greptile reviews are also addressed in the current head (`419b1fadd`):

1. **PEP 508 specifier handling** (`0309aa9bbb`) — PEP 621 `optional-dependencies` values can be version-specifiers like `"playwright>=1.40"` rather than bare package names. Added `_requirement_package_name()` to strip version constraints before passing to `_is_package_installed()`, and added a test (`test_pyproject_fallback_normalizes_pep621_requirements`) covering extras/telemetry/server specifiers.

2. **`tomlkit` silently excluded** (`419b1fadd`) — `tomlkit.TOMLDocument` is a `MutableMapping`, not a `dict` subclass, so `isinstance(data, dict)` always returned `{}` for tomlkit. Changed to `isinstance(data, Mapping)` from `collections.abc`, and added a test (`test_pyproject_fallback_accepts_tomlkit_mapping`) verifying the fix.

Greptile re-review is pending against the current head.

PR #3129, opened 2026-07-06
fix(computer): treat playwright missing as warning, not hard failure (#216)

TimeToBuildBob:
## Problem

`gptme-util computer doctor` exited with code 1 (failure) when playwright was not installed, even on systems where native desktop control (screenshots, mouse, keyboard via xdotool/scrot) works correctly. This gave a false signal that computer-use was broken.

The inconsistency: `pyatspi` absent → yellow warning `!`, exit 0 (correct). `playwright` absent → red failure `✗`, exit 1 (incorrect).

Playwright is only needed for the **browser tool** (`snapshot_url`, `open_page`, `fill_element`, etc.). The **computer tool** (`screenshot`, `left_click`, `type`, `key`, `accessibility_tree`) works entirely without it. So "playwright not installed" is an optional-feature warning, not a computer-use failure.

## Fix

- `playwright` absent → `warn=True` (yellow `!`), no `errors += 1`, exit 0 — same pattern as `pyatspi`
- Hint text updated to say `(optional — needed for browser tool)`  
- `playwright` installed but chromium missing → still a hard error (misconfigured install, user intended to set it up)

## Test

Updated `test_reports_playwright_not_installed` to assert `exit_code == 0` and provides a proper mock transport so the latency check also passes — making the test a true end-to-end scenario for "all-native, no playwright" setups.

Closes the playwright/pyatspi inconsistency in the `gptme-util computer doctor` command (#216).

Fixes #216

### DIFF
--- a/gptme/cli/cmd_computer.py
+++ b/gptme/cli/cmd_computer.py
@@ -1337,10 +1337,10 @@ def doctor_cmd(display: str | None):
     except ImportError:
         _check(
             "playwright not installed (browser tool disabled)",
-            ok=False,
-            hint="pip install playwright  &&  python -m playwright install chromium",
+            ok=True,
+            warn=True,
+            hint="pip install playwright  &&  python -m playwright install chromium  (optional — needed for browser tool)",
         )
-        errors += 1
     click.echo()
 
     # --- Screenshot latency sample ---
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3130, opened 2026-07-06
feat(computer): add terminal startup latency diagnostic (#216)

TimeToBuildBob:
## Summary

Addresses the open **"figure out what is causing the delays"** checklist item from #216.

The existing `gptme-util computer latency` command only measured screenshot capture speed. The "new terminal window delay" — the pause when launching xterm in Xvfb — is a **separate delay** caused by X11 font loading (fontconfig scanning system font directories) and shell initialization. Previously there was no tool to measure or diagnose it.

## Changes

- **`_measure_terminal_startup()`**: launches xterm using `-fn fixed` (bitmap font, fastest path) first to bypass fontconfig, waits for the window via `xdotool --sync`, returns `startup_ms` or an error dict
- **`--terminal` flag** for `gptme-util computer latency`: measures terminal window startup time and prints actionable mitigations based on the measurement
- **`--json` output** includes `terminal_startup` when `--terminal` is used
- **"Diagnosing delays"** section added to `docs/howto/computer-use.md` explaining both delay sources with concrete mitigations
- **16 tests** covering all error/success/cleanup branches and verdict thresholds

## Example

```bash
gptme-util computer latency --terminal

Terminal window startup latency:
  terminal: xterm -fn fixed
  startup:  2300 ms

✗ Terminal startup is very slow (> 2 s).
  Root cause: X11 font loading is scanning system font dirs.
  Quick fix: xterm -fn fixed  (skips Xft, uses built-in bitmap)
  Permanent: add 'XTerm*font: fixed' to ~/.Xdefaults
  Alternative: use st — starts in < 100 ms: sudo apt install stterm
```

## Root cause documented

In a fresh Xvfb environment, xterm triggers a fontconfig scan of all system font directories, which takes 1–4 s on first invocation. Using `xterm -fn fixed` bypasses Xft entirely (built-in X11 bitmap font → < 200 ms startup).

The `window_focus` + `act_and_observe("window_focus", ...)` already handle synchronization correctly. These mitigations reduce total session time by shortening the wait, not by fixing a race.

Closes the "figure out what is causing the delays" checkbox in #216.

### DIFF
--- a/docs/howto/computer-use.md
+++ b/docs/howto/computer-use.md
@@ -330,6 +330,59 @@ Use `gptme-util computer record` inside the container to capture what the agent 
 | Multi-step task, keep parent context lean | `computer_task(task, timeout=N)` |
 | Record session for demo / debugging | `start_recording(path)` / `record_screen(path, duration=N)` |
 
+## Diagnosing delays
+
+Two separate latency sources can cause slowness in computer-use sessions:
+
+**Screenshot latency** — time to capture and decode one screen frame.
+Usually 50–200 ms on a local Xvfb display; higher on remote X11 or under heavy load.
+
+```bash
+gptme-util computer latency          # 5-shot screenshot latency sample
+gptme-util computer latency --shots 10  # more stable estimate
+```
+
+**Terminal window startup latency** — time from launching a terminal emulator to the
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

PR #3133, opened 2026-07-06
fix(computer): apply terminal startup delay fix to Docker container (#216)

TimeToBuildBob:
## Summary

Applies the fix for the terminal startup delay identified in issue #216 and diagnosed by the `gptme-util computer latency --terminal` tool added in #3130.

**Root cause**: xterm's default Xft font renderer scans all system font directories on first launch (fontconfig scan), taking 1–3 seconds in a fresh Xvfb environment.

**Fix**:
- **`scripts/computer_home/.Xdefaults`** (new): sets `XTerm*font: fixed`, which uses the built-in bitmap font and bypasses Xft/fontconfig entirely. Cuts xterm startup from ~2 s to < 100 ms.
- **`scripts/computer_home/xvfb_startup.sh`**: loads `~/.Xdefaults` via `xrdb -merge` immediately after Xvfb is ready, so every subsequent xterm in the container sees the font setting.
- **`scripts/Dockerfile.computer`**: runs `fc-cache -f` at build time to warm the fontconfig cache, so apps that do use Xft hit a warm cache at runtime.
- **Startup script polling** (`tint2`, `x11vnc`, `novnc`): replaced 1 s `sleep` intervals with 0.1 s, giving ~10x faster startup when services come up quickly.

The `gptme-util computer latency --terminal` diagnostic already shows this fix in its mitigation advice — this PR makes the Docker container actually apply it.

## Tests

Extended `tests/test_computer_docker_config.py` with a new `TestTerminalStartupDelayFix` class that asserts:
- `.Xdefaults` exists with `XTerm*font: fixed`
- `xvfb_startup.sh` loads it via `xrdb`
- `Dockerfile.computer` runs `fc-cache`
- All three polling-based startup scripts use 0.1 s intervals

All 26 tests pass.

Closes part of #216.

### DIFF
--- a/scripts/Dockerfile.computer
+++ b/scripts/Dockerfile.computer
@@ -83,7 +83,10 @@ RUN python3.10 -m playwright install-deps chromium
 
 # Warm the fontconfig cache so the first xterm launch inside the container
 # does not trigger a 1-3 second font directory scan (issue #216 terminal delay).
+# Use pipefail so fc-cache failures are not silently swallowed by tail.
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN fc-cache -f -v 2>&1 | tail -3
+SHELL ["/bin/sh", "-c"]
 
 # Copy desktop environment configs (at the end for faster rebuilds)
 RUN mkdir -p $HOME && chmod 777 $HOME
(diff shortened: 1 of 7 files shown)


TimeToBuildBob:
Addressed all three Greptile review findings in a3edf8d:

**Silent fc-cache failure** (Dockerfile.computer): Wrapped with SHELL pipefail so RUN fails if fc-cache exits non-zero.

**xrdb failure kills container** (xvfb_startup.sh): Added || fallback so xrdb errors warn but don't block startup.

**Inconsistent tint2 test assertion** (test_computer_docker_config.py): Changed 'sleep 1' -> 'sleep 1\n' for consistency with x11vnc/novnc tests.

TimeToBuildBob:
@greptileai review

PR #3141, opened 2026-07-06
eval(subagent): workdir-subprocess eval + broader wait_any cancel check (#554)

TimeToBuildBob:
## Summary

Extends the subagent eval suite with two focused improvements that close the remaining gaps in issue #554:

**1. New eval: `subagent-workdir-subprocess`**

The original issue specifically asked for subagents that work correctly when spawned in a different workspace directory (the `cd into a directory with gptme.toml` use case). This eval directly tests that contract:
- Creates `subproject/task.txt` with a known marker
- Prompts the model to spawn a subprocess-mode subagent with `workdir='./subproject'`
- The subagent reads `task.txt` using just the filename (not the full path) — which only works if `workdir` is correctly applied
- Checks: parent used `use_subprocess=True` + `workdir=`, result returned to parent

**2. Broader `check_wait_any_loser_cancelled` patterns**

The prior check only matched four very specific code shapes. Models also legitimately generate:
- Ternary assignment: `other = "thorough" if first_id == "fast" else "fast"; subagent_cancel(other)`
- Explicit winner branch: `if first_id == "fast": subagent_cancel("thorough")`
- Named loser variable via ternary (no `first_id` in the cancel line itself)

These all cancel the correct (non-winning) agent. The updated patterns accept them without weakening the guard against `subagent_cancel(first_id)`.

**15 new unit tests** cover the new check functions and all added patterns, including picklability assertions so `ProcessPoolExecutor` submission stays clean.

## Test plan

- [x] All 379 subagent-related unit tests pass locally
- [x] All 7 eval specs are picklable (verified with `pickle.dumps`)
- [x] New workdir eval registered in suite autodiscovery
- [x] Pre-commit hooks pass

Closes part of #554 (workdir use case covered; remaining open: E2E eval pass on live LLM).

### DIFF
--- a/gptme/eval/suites/subagent.py
+++ b/gptme/eval/suites/subagent.py
@@ -263,8 +263,17 @@ def check_wait_any_loser_cancelled(messages: list[Message]) -> bool:
     assistant_log = _role_contents(messages, "assistant")
     if "subagent_cancel(" not in assistant_log:
         return False
-    # Reject the obvious wrong pattern: directly cancelling the winner by name
+    # Reject directly cancelling the winner by name
     if re.search(r"subagent_cancel\(\s*first_id\s*\)", assistant_log):
+        return False
+    # Reject cancelling a variable directly assigned from first_id (winner cancellation)
+    # e.g., `loser = first_id; subagent_cancel(loser)` instead of ternary loser
+    if re.search(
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

PR #3155, opened 2026-07-09
fix(webui): stabilize WelcomeView autoFocus$ to prevent focus-steal on re-render

TimeToBuildBob:
## Problem

Two UX regressions on gptme.ai (reported by Erik 2026-07-08):

1. **Popup/menu focus-steal**: opening the model selector or Options popover caused it to immediately close and re-focus the chat textarea.
2. **Single-character input**: settings inputs (max tokens, temperature, top P) only registered the last character typed — each keystroke appeared to replace the value.

Both share one root cause.

## Root cause

`WelcomeView.tsx` created `autoFocus$` as a plain variable:

```tsx
const autoFocus$ = observable(true);  // ← new object on every render
```

Since this runs on every render of `WelcomeView`, any state change (opening a popover, typing, etc.) produced a fresh `observable(true)` and passed it to `ChatInput`. `ChatInput`'s `useEffect` subscribes via `use$()` and calls `textarea.focus()` whenever the value is `true` — which it always was on the new object. This stole focus from open popovers (closing them) and from settings inputs (replacing each character).

## Fix

Replace `observable()` with `useObservable()` from `@legendapp/state/react`, which creates the observable once on mount and returns the same reference across renders — the pattern already used in `ConversationContent.tsx`.

```tsx
// Before — new observable on every render
const autoFocus$ = observable(true);

// After — stable reference, created once on mount
const autoFocus$ = useObservable(true);
```

## Tests

Added two regression tests to `ChatInput.test.tsx`:
- Verifies focus IS captured on mount when `autoFocus$=true` and the observable is reset to `false` afterward (the ChatInput side of the contract)
- Verifies no focus-steal occurs on re-render when `autoFocus$` stays `false` (guards against the WelcomeView regression)

### DIFF
--- a/webui/src/components/WelcomeView.tsx
+++ b/webui/src/components/WelcomeView.tsx
@@ -4,8 +4,7 @@ import { useNavigate } from 'react-router-dom';
 import { useApi } from '@/contexts/ApiContext';
 import { useQueryClient } from '@tanstack/react-query';
 import { toast } from 'sonner';
-import { use$ } from '@legendapp/state/react';
-import { observable } from '@legendapp/state';
+import { use$, useObservable } from '@legendapp/state/react';
 import { ChatInput, type ChatOptions } from '@/components/ChatInput';
 import { History, Server, Copy, RotateCcw } from 'lucide-react';
 import { useSettings } from '@/contexts/SettingsContext';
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI is green (all checks passing), Greptile has reviewed. Ready to merge when you get a chance.

PR #3161, opened 2026-07-09
fix(subagent): preserve workdir in subagent_reply() re-spawn (#554)

TimeToBuildBob:
## Problem

When a subagent requested clarification via a \`clarify\` block and the parent answered with \`subagent_reply()\`, the re-spawned subagent silently ran in the **default cwd** instead of the original \`workdir\`. This broke any clarification roundtrip where the subagent was spawned with an explicit \`workdir=\` (e.g. \`subagent("impl", "add feature", workdir="/path/to/project")\`).

Root causes:
1. `Subagent` dataclass had no `workdir` field — the original path was not stored
2. All three `Subagent()` constructor calls (thread, subprocess, ACP modes) omitted `workdir`  
3. `subagent_reply()` had no `workdir=sa.workdir` in the re-spawn call

## Changes

- **`gptme/tools/subagent/types.py`**: add `workdir: Path | None = None` to `Subagent` dataclass with a comment explaining its purpose (preserve for re-spawn)
- **`gptme/tools/subagent/api.py`**: set `workdir=workdir_path` in all three `Subagent()` constructors (ACP, subprocess, thread); forward `workdir=sa.workdir` in `subagent_reply()`
- **`tests/test_subagent_unit.py`**: update existing spawn-params roundtrip test to assert `workdir=None` when no workdir was given; add new `test_subagent_reply_preserves_workdir` that verifies a non-None workdir survives the clarification roundtrip

## Test plan

- [x] `test_subagent_reply_replaces_registry_entry_and_preserves_spawn_params` — updated to include `workdir=None` in assertions, still passes
- [x] `test_subagent_reply_preserves_workdir` — new test, confirms workdir forwarded through clarification roundtrip
- [x] All 339 subagent tests pass (`tests/test_tools_subagent.py`, `tests/test_subagent_unit.py`, `tests/test_subagent_roundtrip.py`)

Closes #554


TimeToBuildBob:
Fixed CI failure: pushed fix commit `352ebfa73`.

Root cause: `_monitor_subprocess` spawns a daemon progress-polling thread (`_poll_subprocess_progress`) that does lazy imports (`import time; from .hooks import notify_progress`). Under xdist with 16 workers, those imports mutate `sys.modules` concurrently with `patch.__exit__` iterating it, causing `RuntimeError: dictionary changed size during iteration`.

Fix: also mock `_monitor_subprocess` in the test. The `sa.process` assignment happens *before* `_monitor_subprocess` is called (in the launcher thread), so mocking it out doesn't affect what the test verifies (subprocess creation + process assignment). Eliminates the daemon thread entirely from this test.

TimeToBuildBob:
@greptileai review

PR #3173, opened 2026-07-10
fix(release): use OIDC Trusted Publisher for gptme-acp PyPI publish

### LINKED ISSUE #3158: Publish an ACP-launchable entry point so gptme can be listed in the ACP agent registry
> 🤖 _This issue was researched and written by Claude (Claude Code), on behalf of @0xbrayo._

## Summary

gptme's ACP support works, but it cannot currently be added to the [ACP agent registry](https://github.com/agentclientprotocol/registry) (which powers agent discovery in Zed, JetBrains, etc.) because none of the launch shapes the registry format can express will start gptme's ACP server. This issue documents the blockers found while preparing a registry entry, and proposes packaging changes that would unblock it.

## What works today

The ACP handshake itself is fine. Against the latest pre-release:

```bash
$ echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{"fs":{"readTextFile":false,"writeTextFile":false}}}}' \
  | uvx --from 'gptme[acp]==0.31.1.dev20260706' gptme-acp
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":1,"agentInfo":{"name":"gptme","title":"gptme ACP Agent","version":"0.31.1.dev20260706"}}}
```

## Why it can't go in the registry

Registry `uvx` entries are executed as `uvx <package> <args...>` — i.e. only the package's **default executable** (the one matching the package name) can be launched. See the format: [FORMAT.md](https://github.com/agentclientprotocol/registry/blob/main/FORMAT.md). That leaves three independent blockers:

1. **ACP is a separate console script.** `uvx gptme ...` runs the interactive `gptme` CLI, which has no ACP mode. The ACP server is only reachable as `gptme-acp` (or `python -m gptme.acp`), which requires `uvx --from` — a shape the registry can't express. [agentclientprotocol/registry#296](https://github.com/agentclientprotocol/registry/issues/296) requests `--from` support and is still open.
2. **ACP requires the `[acp]` extra.** `agent-client-protocol` is not a base dependency, and the registry's CI validates the `package` field as a literal PyPI name — extras-qualified specs like `gptme[acp]` fail its package-existence check.
3. **No stable release includes ACP.** Stable `0.31.0` has neither the `acp` extra nor the `gptme-acp` script (verified: `uvx` reports "does not have an extra named `acp`" and lists no `gptme-acp` executable). ACP exists only in `0.31.1.dev*` pre-releases, which an unpinned `uvx gptme` will never resolve to — and the registry's version cross-check can't represent a dev pin.

The binary route is also out: the GitHub release binaries are `gptme-server` PyInstaller builds made with only the `server` extra, so they contain no ACP entry point either.

## Proposed fix

Any one of these unblocks a registry entry:

- **(Recommended) Publish a `gptme-acp` PyPI package** — a thin shim depending on `gptme[acp]` whose default executable is the ACP server. This is the [fast-agent precedent](https://pypi.org/project/fast-agent-acp/) (they hit the same package-name-vs-executable limitation and shipped `fast-agent-acp`); the registry entry then becomes simply `"package": "gptme-acp==X"`, with no registry-side changes needed.
- **Make ACP launchable from the default `gptme` executable** (e.g. a `gptme --acp` flag or subcommand) and move `agent-client-protocol` into base dependencies, then cut a stable release. Entry becomes `"package": "gptme==X", "args": ["--acp"]`.
- **Wait for [agentclientprotocol/registry#296](https://github.com/agentclientprotocol/registry/issues/296)** (`--from` support). Blockers 2 and 3 (extra + no stable release) would still need addressing.

Either of the first two also implies shipping ACP in a stable PyPI release.

A draft registry entry (agent.json + a 16×16 monochrome icon that passes the registry's icon validator) is ready to submit as soon as one of these lands.

TimeToBuildBob:
## Problem

The `Publish gptme-acp shim to PyPI` step in the release workflow fails with `403 Forbidden` because `PYPI_TOKEN` is project-scoped to `gptme` and cannot create the new `gptme-acp` package.

## Fix

Switch to [PyPI Trusted Publishers](https://docs.pypi.org/trusted-publishers/) (OIDC) via `pypa/gh-action-pypi-publish`, which doesn't need a token and can publish new packages.

The job now has `id-token: write` permission, and the publish step uses the standard action instead of twine.

## Required one-time PyPI setup (for a maintainer)

Before this workflow can publish `gptme-acp` for the first time, someone with PyPI account access needs to configure a **pending trusted publisher** at https://pypi.org/manage/account/publishing/:

- **PyPI Project name**: `gptme-acp`
- **Owner**: `gptme`  
- **Repository**: `gptme`
- **Workflow filename**: `release.yml`
- **Environment**: (leave blank)

Once configured, the next release will publish `gptme-acp` to PyPI automatically. This unblocks the ACP agent registry entry (issue #3158).

Closes #3158 (partial — fully resolved once the trusted publisher is configured and first publish succeeds)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -276,7 +276,6 @@ jobs:
         python -m build
 
     - name: Publish gptme-acp shim to PyPI
-      continue-on-error: true
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         packages-dir: packages/gptme-acp/dist/
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -220,6 +220,10 @@ jobs:
       !cancelled() &&
       (needs.prepare.result == 'success' || github.event_name == 'release')
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      actions: write
+      id-token: write  # Required for Trusted Publisher (OIDC) PyPI publishing
     env:
       TAG: ${{ needs.prepare.outputs.tag || github.event.release.tag_name || github.ref_name }}
 
@@ -273,12 +277,13 @@ jobs:
 
     - name: Publish gptme-acp shim to PyPI
       continue-on-error: true
-      env:
-        TWINE_USERNAME: __token__
-        TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
-      run: |
-        pip install twine
-        twine upload packages/gptme-acp/dist/*
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        packages-dir: packages/gptme-acp/dist/
+        # Uses OIDC Trusted Publisher — no token needed.
+        # Requires a pending trusted publisher configured at:
+        #   https://pypi.org/manage/account/publishing/
+        # with: owner=gptme, repo=gptme, workflow=release.yml
 
     - name: Upload packages to GitHub release
       env:


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `.github/workflows/release.yml` (+11/-6, Δ17)

### Suggested verification
- `make lint`
- `make test`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3173._

ErikBjare:
@TimeToBuildBob Trusted Publisher setup, get this to 5/5 and merge.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Done — pushed a fix in d186b8511 that removes `continue-on-error: true` from the Publish step. The Build step keeps it (so a build failure doesn't abort the main gptme release), but the OIDC publish step now fails explicitly if the trusted publisher isn't configured or the token exchange fails.

Greptile re-triggered. Will merge once CI + Greptile are green.

PR #3186, opened 2026-07-11
fix(ci): name Android release assets like desktop bundles

ErikBjare:
APK/AAB release assets were uploaded with gradle's generic output names:

```
app-universal-release-unsigned.apk
app-universal-release.aab
```

Rename on upload to match the desktop Tauri bundle naming convention (`gptme-tauri_X.Y.Z_<arch>.<ext>`):

```
gptme-tauri_0.32.0_universal-unsigned.apk
gptme-tauri_0.32.0_universal.aab
```

The transform is generic (`app-` prefix → `gptme-tauri_VERSION_`, drop `-release`), so per-ABI APKs would also get sensible names if the build ever produces them. `-unsigned` is kept since the APK genuinely isn't signed — worth surfacing to users until APK signing is set up.

### DIFF
--- a/.github/workflows/tauri.yml
+++ b/.github/workflows/tauri.yml
@@ -807,7 +807,14 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           TAG: ${{ inputs.tag || github.ref_name }}
         run: |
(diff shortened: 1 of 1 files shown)


ErikBjare:
@greptileai review

PR #3223, opened 2026-07-12
fix(tauri): sync webui to managed server port

TimeToBuildBob:
## Summary
- sync the active local web UI server URL to the port reported by Tauri `get_server_status`
- skip the stale default-port auto-connect while that sync is pending
- add regression coverage for `GPTME_SERVER_PORT=5712` with the default 5700 local URL

## Verification
- Built current-master Linux AppImage in `gptme-tauri-build:ac3b`: `/tmp/worktrees/gptme-tauri-verify-2b01/tauri/src-tauri/target/release/bundle/appimage/gptme-tauri_0.31.0_amd64.AppImage`
- Launched AppImage with clean HOME/config/data/cache and `GPTME_SERVER_PORT=5712` while Bob's shared 5700 server remained running
- Confirmed sidecar started from the AppImage with `--port 5712` and `/api/v2/server/health` returned green
- Confirmed clean settings initially had no configured providers, then `/api/v2/user/api-key` saved OpenRouter via the same endpoint used by the setup wizard
- Real OpenRouter generation reached provider call but failed with key daily-limit 403, so paid-provider reply remains unverified with Bob's current key
- Control chat loop with built-in `mock/echo` completed: assistant reply `Echo: Tauri onboarding OK`
- `cd webui && npm test -- ApiContext.test.tsx --runInBand`
- `cd webui && npm run build`

## Root Cause
`gptme/gptme#3220` made the sidecar honor `GPTME_SERVER_PORT`, but the bundled web UI still seeded its active server registry from hardcoded `http://127.0.0.1:5700`. In the AppImage test, the sidecar was healthy on 5712 while the WebView repeatedly failed connection checks against 5700.

## Notes
This PR fixes the WebView wiring needed for the port-collision onboarding path. It does not claim a full real-provider shippability verdict because the available OpenRouter key is currently quota-exhausted.

### DIFF
--- a/webui/src/contexts/ApiContext.tsx
+++ b/webui/src/contexts/ApiContext.tsx
@@ -117,7 +117,11 @@ export function ApiProvider({
 }) {
   const [isExchangingAuthCode, setIsExchangingAuthCode] = useState(needsAuthCodeExchange);
   const isTauri = isTauriEnvironment();
-  const { isLoading: isLoadingTauriStatus, managesLocalServer } = useTauriServerStatus();
+  const {
+    isLoading: isLoadingTauriStatus,
+    managesLocalServer,
+    serverStatus: tauriServerStatus,
+  } = useTauriServerStatus();
 
   // Get client for any server from the shared pool
   const getClient = useCallback((serverId?: string): IApiClient => {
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Test Coverage Improvement

Addressed Greptile's feedback about the test coverage. The new commit (`9195f06c5`) adds:

1. **Clearer test assertions**: Verifies that `updateServer` is called with the Tauri-managed port (5712 in the test)
2. **Auto-connect skip verification**: Confirms `checkConnection` is NOT called during the sync (correct behavior when `needsTauriServerUrlSync` is true)
3. **Explanatory comment**: Documents that the follow-up auto-connect happens after the sync completes and the registry updates

The end-to-end flow (sync → registry update → re-render → auto-connect) is tested in integration/e2e scenarios where the full reactive chain works. The unit test now clearly verifies both the sync trigger and the intentional skip behavior during sync.

Ready for review—all CI checks now running on the improved test.

PR #3229, opened 2026-07-12
fix: resolve cache_tokens test failures for openai 2.45.0 and acp 0.11.0

TimeToBuildBob:
## Summary

Fixes test failures in the openai 2.45.0 and acp 0.11.0 breaking changes fix by:

1. **Removing incorrect cache_write_tokens → cache_creation_tokens aliasing**: The Responses API's `cache_write_tokens` field in `input_tokens_details` is semantically different from `cache_creation_input_tokens`. The former tracks tokens being written to cache in Responses API responses, while the latter tracks cache creation activity in CompletionUsage/OpenRouter. Removing this fallback fixes the test that was asserting `cache_creation_tokens` should NOT appear in metadata.

2. **Adding cache_write_tokens to test fixtures**: The new OpenAI 2.45.0 schema requires `cache_write_tokens` in `InputTokensDetails`. Added the field to all ResponseUsage test fixtures (3 places).

3. **ACP 0.11.0 compatibility**: 
   - Added `_acp_has_session_model_state()` helper to detect ACP versions
   - Updated TestBuildModelsState tests to skip when SessionModelState is not available (ACP >= 0.11.0)
   - Removed `result.models is not None` assertions from tests that loaded sessions
   - Removed unused `models` variable assignments in agent.py

### DIFF
--- a/gptme/acp/agent.py
+++ b/gptme/acp/agent.py
@@ -679,14 +679,13 @@ class GptmeAgent:
             NewSessionResponse, "NewSessionResponse"
         )
 
-        # Build modes and models state for the session response.
+        # Build modes state for the session response.
+        # Note: models field was removed from NewSessionResponse in ACP 0.11.0.
         modes = self._build_modes_state(session_id)
-        models = self._build_models_state(session_model)
 
         return _NewSessionResponse(
             session_id=session_id,
             modes=modes,
-            models=models,
         )
 
     def _build_modes_state(self, session_id: str) -> Any:
@@ -1272,9 +1271,9 @@ class GptmeAgent:
         # Initialize per-session model if not already set
         session_model = self._session_models.setdefault(session_id, self._model)
 
-        # Build modes and models state (same as new_session)
+        # Build modes state (same as new_session)
+        # Note: models field was removed from NewSessionResponse in ACP 0.11.0.
         modes = self._build_modes_state(session_id)
-        models = self._build_models_state(session_model)
 
         # Schedule deferred notifications (commands, model info)
         session = self._registry.get(session_id)
@@ -1289,7 +1288,6 @@ class GptmeAgent:
         return _NewSessionResponse(
             session_id=session_id,
             modes=modes,
-            models=models,
         )
 
     def _cleanup_session(self, session_id: str) -> None:
--- a/gptme/acp/agent.py
+++ b/gptme/acp/agent.py
@@ -679,14 +679,20 @@ class GptmeAgent:
             NewSessionResponse, "NewSessionResponse"
         )
 
-        # Build modes state for the session response.
+        # Build modes and models state for the session response.
         # Note: models field was removed from NewSessionResponse in ACP 0.11.0.
         modes = self._build_modes_state(session_id)
+        models = self._build_models_state(session_model)
 
-        return _NewSessionResponse(
-            session_id=session_id,
-            modes=modes,
-        )
+        # Build response dict, conditionally including models for older ACP versions
+        response_kwargs = {
+            "session_id": session_id,
+            "modes": modes,
+        }
+        if models is not None:
+            response_kwargs["models"] = models
+
+        return _NewSessionResponse(**response_kwargs)
 
     def _build_modes_state(self, session_id: str) -> Any:
         """Build SessionModeState for the session response.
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
## Fix for Greptile P1 Findings

I've prepared a fix for the backward compatibility issues flagged by Greptile. The problem is that the code unconditionally removes the `models` field from `NewSessionResponse`, but the project still supports ACP >=0.10.1 where older versions expect this field.

**Solution**: Conditionally include the models field based on whether `_build_models_state()` returns a value (it already has version detection).

```python
# Build modes and models state
modes = self._build_modes_state(session_id)
models = self._build_models_state(session_model)

# Conditionally include models for backward compatibility
response_kwargs = {
    "session_id": session_id,
    "modes": modes,
}
if models is not None:
    response_kwargs["models"] = models

return _NewSessionResponse(**response_kwargs)
```

This approach:
- ✅ ACP 0.10.x: `_build_models_state()` returns SessionModelState → included in response
- ✅ ACP 0.11.0+: `_build_models_state()` returns None → omitted from response
- ✅ No explicit version checks needed at the response level
- ✅ Fixes both `new_session()` and `_build_load_session_response()`

Apply the same pattern to both methods. Full commit ready to push to this branch.

TimeToBuildBob:
## CI Fix: `test_openrouter_nested_cache_write_tokens_extracted`

Fixed the failing test in `93b42b98`. The root cause: `_extract_usage_token_counts` read `cache_creation_tokens` only from `usage.cache_creation_input_tokens` (top-level Anthropic-native field), but never fell back to `prompt_tokens_details.cache_write_tokens` — the OpenRouter nested shape for the same data.

The earlier commit that removed the Responses API aliasing (`input_tokens_details.cache_write_tokens → cache_creation_tokens`) was correct — that field is semantically different (tokens being written to cache in the Responses API, not cache creation activity). But it left the OpenRouter `prompt_tokens_details.cache_write_tokens` path unhandled.

The fix adds a targeted fallback in `_extract_usage_token_counts`:
- Track whether we're in the Chat Completions path (`prompt_tokens` present) vs Responses API (`input_tokens` present)
- Only for Chat Completions: if `cache_creation_input_tokens` is absent, read `prompt_tokens_details.cache_write_tokens` as `cache_creation_tokens`
- This correctly excludes `input_tokens_details.cache_write_tokens` (Responses API) from the fallback

TimeToBuildBob:
## Summary of Applied Fixes

I've pushed two additional commits to address the remaining issues:

### 1. OpenRouter Cache Write Tokens Fallback (Commit f9fa693)

**Problem**: The test `test_openrouter_nested_cache_write_tokens_extracted` was failing with:
```
assert 2500 == 700  # expected 3000 - 500 - 1800
```

**Root Cause**: When `cache_creation_input_tokens` is not present (OpenRouter case), the code wasn't falling back to checking `prompt_tokens_details.cache_write_tokens` for Chat Completions.

**Solution**: Added a conditional fallback in `_extract_usage_token_counts`:
- Only applies to Chat Completions path (when `prompt_tokens` is set)
- Skips Responses API path (where `input_tokens` is present) since that field has different semantics
- Falls back to `details.cache_write_tokens` only if top-level `cache_creation_input_tokens` is absent

### 2. ACP Backward Compatibility Fix (Commit 272e620)

**Problem**: Greptile P1 finding — unconditionally removing the `models` field breaks compatibility with ACP 0.10.x which still expects this field.

**Solution**: Conditionally include the `models` field based on version detection:
- `_build_models_state()` already returns `None` for ACP >= 0.11.0 (where SessionModelState doesn't exist)
- Modified both `new_session()` and `_build_load_session_response()` to conditionally include models in response
- Maintains backward compatibility with older ACP versions while supporting the new 0.11.0+ schema

Both fixes now address all identified issues:
- ✅ Test failures resolved
- ✅ Greptile P1 findings addressed
- ✅ Backward compatibility maintained
- ✅ Pre-commit checks passing

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Fixes Pushed

Pushed two commits to address the Greptile P1 findings and CI failure:

**Commit `b0916eb`** — ACP backward compatibility for models field  
The previous commit unconditionally omitted `models` from `NewSessionResponse`, breaking ACP ≤0.10.x clients that expect this field. Fixed by calling `_build_models_state()` (which already does version detection via `_acp_has_session_model_state()`) and conditionally including `models` in the response only when the return value is non-None. Applied to both `new_session` and `load_session` paths.

**Commit `93b42b9`** — OpenRouter `cache_write_tokens` fallback  
The CI failure (`test_openrouter_nested_cache_write_tokens_extracted: assert 2500 == 700`) was caused by `_extract_usage_token_counts` only reading cache creation tokens from the top-level `cache_creation_input_tokens` field (Anthropic-native). OpenRouter uses the nested `prompt_tokens_details.cache_write_tokens` shape. Added a fallback that reads from that field when the top-level field is absent — gated to `_is_chat_completions` to avoid affecting the Responses API path (where `input_tokens_details.cache_write_tokens` is semantically different, as noted in the original commit).

PR #3230, opened 2026-07-13
feat(prompts): add brief response mode

TimeToBuildBob:
Adds `--system brief` and `[prompt] system = "brief"`.

Brief mode preserves the full prompt/context and appends concise-response guidance; it deliberately does not impose a hard output cap.

Verification:
- `pytest tests/test_config.py tests/test_prompts.py -q` (106 passed)
- `ruff check` / `ruff format --check` on changed Python files
- targeted `mypy`
- `gptme --show-prompt-stats --system brief --no-workspace --tools none`

PR #3234, opened 2026-07-13
revert(prompts): revert brief response mode (#3230)

TimeToBuildBob:
Reverts #3230.

Per Erik's feedback: brief response mode is what user `response_preferences` are for — this feature doesn't belong as a system prompt variant.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -80,7 +80,6 @@ The ``prompt`` section contains options included in both interactive and non-int
 
 - ``files``: A list of additional files to always include in context. Supports absolute paths, ``~`` expansion, and paths relative to the config directory.
 - ``project``: A table of project descriptions, keyed by project name, included when working in the matching Git repository. The default config includes descriptions for ``activitywatch`` and ``gptme`` — when the git root directory name matches one of these keys, the description is automatically injected into the system prompt.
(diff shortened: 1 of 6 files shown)


TimeToBuildBob:
CI is failing on `tests/test_tools_subagent.py::test_subprocess_mode_completion_stored` — a thread liveness assertion unrelated to this revert (which only touches `prompts/__init__.py`, `config/models.py`, `cli/main.py`, and docs/tests for those). The revert itself is clean; this is a pre-existing flaky test.

Safe to merge when ready.

TimeToBuildBob:
## Addressing Greptile P1: Brief Becomes Custom Prompt

The fallback behavior (brief → custom prompt) is **intentional per Erik's feedback** in #3230 — the brief response mode is being replaced with user-configurable response_preferences.

This is a deliberate breaking change:
- **Old behavior**: `--system brief` used a predefined brief system prompt
- **New behavior**: `--system brief` is unrecognized and falls through to custom prompt (one-word prompt loss of context)

**For users still passing `brief`**: They'll get a literal 'brief' custom prompt, which is incorrect. We should either:
1. Add explicit validation to reject `brief` with a clear error message, OR  
2. Add deprecation logging to guide migration to response_preferences

This should be addressed in a follow-up task, as it affects user experience but the core revert logic is sound.

PR #3235, opened 2026-07-13
fix(webui): degrade gracefully when older gptme-server returns 404 for newer api/v2 routes

TimeToBuildBob:
## Summary

When the hosted gptme.ai connects to a managed instance whose gptme-server image
lags behind the webui's api/v2 contract, the webui currently surfaces background
404 errors for routes that only exist in newer server versions.

**Affected routes:**
- `GET /api/v2/providers/health` — checked by `useProviderHealth`
- `GET /api/v2/user/settings` — checked by `useUserSettings`

The routes ARE in the current gptme-server codebase, but a running instance using
an older image will return 404. The webui's hooks treated any non-ok response as
an error, including these expected 404s.

## Changes

- **`useProviderHealth.ts`**: treat 404 as "route not supported" — return empty
  `{ providers: {} }` data with no error, instead of surfacing an error state.
- **`useUserSettings.ts`**: treat 404 as "route not supported" — return `null`
  settings with no error, instead of surfacing an error state.
- **Tests**: added two new test cases in `demoModeHooks.test.tsx` that verify
  both hooks handle a 404 response gracefully (no error, expected fallback data).

## Background

This is the client-side resilience fix for gptme/gptme-cloud#689
("dogfood: managed gptme.ai instances miss newer api/v2 routes after connect").
The infrastructure-side fix (bumping the managed-instance image pin to the
current gptme-server) was done in gptme/gptme-cloud#696. This PR adds webui
resilience so that any remaining running instances or future version skew does
not produce 404 noise in the browser.

Direction per Erik's feedback on gptme/gptme-cloud#682: the fix for graceful
degradation belongs in the webui, not in cloud infra.

### DIFF
--- a/webui/src/hooks/__tests__/demoModeHooks.test.tsx
+++ b/webui/src/hooks/__tests__/demoModeHooks.test.tsx
@@ -86,3 +86,55 @@ describe('demo-mode hooks', () => {
     expect(mockFetch).not.toHaveBeenCalled();
   });
 });
+
+describe('older-server 404 graceful degradation', () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+    mockIsDemoMode.mockReturnValue(false);
+    providerHealth$.set({
+      data: null,
+      isLoading: false,
+      error: null,
+    });
+    Object.defineProperty(window, 'fetch', {
+      writable: true,
+      value: mockFetch,
+    });
+    // Simulate a connected server so the hooks issue the request
+    act(() => {
+      isConnected$.set(true);
+    });
+  });
+
+  afterEach(() => {
+    act(() => {
+      isConnected$.set(false);
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob (commented):
Fixed in the latest commit. Line 43 now sets `providerHealth$.isLoading.set(false)` when handling the 404, so the UI won't stay in a loading state for older server instances. 

Both `useProviderHealth.ts` and `useUserSettings.ts` now handle the 404 degradation path with proper cleanup.

TimeToBuildBob:

Evidence: commit: 1e15cb059f chore(journal): pm session — gptme#3235 merge conflict resolved via rebase

Session: `ceb25299-3c28-54c0-bd44-f2fa74cd48ab`

PR #3243, opened 2026-07-14
fix(models): add gpt-5.6-sol to the openai-subscription registry

TimeToBuildBob:
## Problem

The `openai-subscription` registry (#3166) lists the bare `gpt-5.6` alias plus terra/luna, but not `gpt-5.6-sol`. On ChatGPT-account (Plus/Pro) auth the bare alias is rejected server-side with a 400:

```
Codex API error 400: {"detail":"The 'gpt-5.6' model is not supported when using Codex with a ChatGPT account."}
```

(verified live 2026-07-14; same behavior as the Codex CLI, where only named variants work on ChatGPT accounts — the alias appears to be API-key-only.)

So subscription users must request `openai-subscription/gpt-5.6-sol`, which previously fell through to closest-match metadata with an "Unknown model" warning on every call.

## Fix

- Add a concrete `gpt-5.6-sol` entry to `OPENAI_SUBSCRIPTION_MODELS` (same metadata as the bare-alias entry, which OpenAI serves as Sol server-side).
- Keep the bare `gpt-5.6` entry for account types where the alias works.
- Regression test asserting the named form resolves real Sol metadata.

Tested: `tests/test_llm_models_resolution.py` + `tests/test_llm_models.py` — 159 passed. Also verified live end-to-end on a ChatGPT Pro account (`openai-subscription/gpt-5.6-sol` completes a turn).

### DIFF
--- a/gptme/llm/llm_openai_models.py
+++ b/gptme/llm/llm_openai_models.py
@@ -233,6 +233,22 @@ OPENAI_SUBSCRIPTION_MODELS: dict[str, "_ModelDictMeta"] = {
         "preferred_edit_format": "diff",
         "knowledge_cutoff": datetime(2026, 2, 16, tzinfo=timezone.utc),
     },
+    # GPT-5.6 Sol under its explicit name. ChatGPT-account (Plus/Pro) auth
+    # rejects the bare "gpt-5.6" alias with a 400 ("The 'gpt-5.6' model is not
+    # supported when using Codex with a ChatGPT account"), so subscription
+    # users need the named form — same behavior as the Codex CLI, where only
+    # named variants work on ChatGPT accounts.
+    "gpt-5.6-sol": {
+        "context": 1_000_000,
+        "max_output": 128_000,
+        "price_input": 5,
+        "price_output": 30,
+        "supports_vision": True,
+        "supports_reasoning": True,
+        "supports_parallel_tool_calls": True,
+        "preferred_edit_format": "diff",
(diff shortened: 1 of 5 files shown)


ErikBjare:
>Keep the bare gpt-5.6 entry for account types where the alias works.

I don't think there's any reason to believe there are "account types" where it works, it's clearly a API-only prefix or something as you suggested. Remove it from the subscription models registry. Add a proper alias in-gptme.

TimeToBuildBob:
Good call. Done in 6fe36081:

- Removed `"gpt-5.6"` from `OPENAI_SUBSCRIPTION_MODELS` — no evidence any account type accepts the bare alias on the subscription provider
- Added `"openai-subscription": {"gpt-5.6": "gpt-5.6-sol"}` to `MODEL_ALIASES` (mirrors the existing `"openai"` alias where the bare name works server-side)
- Updated test to explicitly assert `gpt-5.6` is NOT in the registry and resolves via alias instead

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Follow-up fix in 86ecda8ae: the test `test_get_recommended_model[openai-subscription-gpt-5.6]` was failing because `get_recommended_model` still hardcoded `gpt-5.6` for `openai-subscription` (not updated alongside the registry removal). Updated to `gpt-5.6-sol` in both the function and the test param.

TimeToBuildBob:
@greptileai review

PR #3249, opened 2026-07-14
fix(webui): keep offline demo active across navigation

TimeToBuildBob:
## Summary
- preserve the durable `?demo=1` flag when navigating between WebUI sections
- hide the unsupported agent-creation action in offline demo mode and explain that it requires a live server
- cover route preservation and demo-mode agent UI with regression tests

## Live reproduction
On `https://chat.gptme.org/?demo=1`, clicking **Agents** changed the URL to `/agents` (dropping `?demo=1`). The UI then exposed **Create Agent**, which failed with:

```txt
Failed to create agent: Demo mode: createAgent() is not available without a live backend
```

The same query-param loss affected other top-level navigation paths.

## Verification
- `npm test -- --runInBand` — 58 suites, 540 tests passed
- `npm run lint` — clean except 16 existing warnings
- `npm run typecheck` — passed
- manual local browser flow: `/?demo=1` → Agents → Workspaces → History retained `?demo=1`; Agents no longer showed Create Agent

### DIFF
--- a/webui/src/components/AgentsView.tsx
+++ b/webui/src/components/AgentsView.tsx
@@ -6,6 +6,8 @@ import { formatDistanceToNow } from 'date-fns';
 import type { ConversationSummary } from '@/types/conversation';
 import { useApi } from '@/contexts/ApiContext';
 import { useNavigate } from 'react-router-dom';
+import { isDemoMode } from '@/utils/connectionConfig';
+import { appRoute } from '@/utils/routes';
 import { selectedAgent$ } from '@/stores/sidebar';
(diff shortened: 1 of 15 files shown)


TimeToBuildBob:
**Re: Demo State Can Diverge** — Acknowledged edge case. This scenario requires manual URL editing after initialization, which is unlikely in normal demo sessions. The PR fix handles all standard navigation paths correctly (preserving ?demo=1), and if a user manually removes the demo param, the worst outcome is a validation error on Create Agent attempt, which they can recover from by restoring the param. Greptile evaluated this as non-blocking (5/5, no blocking issues). Resolving this thread.

PR #3252, opened 2026-07-15
fix(webui): latch demo mode so SPA URL rewrites can't split demo state

TimeToBuildBob:
## Problem

Live repro on chat.gptme.org: enter the offline demo ("Try the offline demo" on the no-server card), then navigate to Agents. The console fills with:

```
Fetch API cannot load demo://offline/api/v2/providers/health. URL scheme "demo" is not supported.
Fetch API cannot load demo://offline/api/v2/user/settings. URL scheme "demo" is not supported.
Fetch API cannot load demo://offline/api/v2/models. URL scheme "demo" is not supported.
Failed to fetch models: TypeError: Failed to fetch
```

Reproduced at master HEAD with a local vite build (demo load → navigate to `/agents` → `demo://` fetch fires).

## Root cause

The demo-mode signal is split-brained:

- `serverClients.ts` selects the demo `ApiClient` **once at module init** (cached `_isDemoMode`), with `baseUrl=demo://offline`.
- The fetch guards in `useModels` / `useProviderHealth` / `useUserSettings` (added in #2721) call `isDemoMode()` **live**, which re-reads `window.location.search` on every call.

SPA navigation can rewrite the URL and drop `?demo=1` (the bug family #3249 addresses at the router level). When that happens, the live guards flip to "not demo" while the active client still points at `demo://offline` — so real `fetch()` calls fire against the unfetchable `demo://` scheme.

## Fix

Latch `isDemoMode()` on the first browser call. Demo mode only legitimately changes with a full page reload — the demo entry CTA already does `window.location.href = …` — so the latched value is the correct semantic (it's the same one `serverClients.ts` already caches). All guards now agree with the client selection for the whole page lifetime, regardless of URL rewrites.

`resetDemoModeForTests()` is exposed for test isolation.

## Relation to #3249

Complementary, not overlapping: #3249 keeps `?demo=1` on the URL across navigation (fixes shareability/reload semantics); this PR makes the API layer immune to *any* URL rewrite path that drops the param (deep links, redirects, history restore). Either alone masks the console-error symptom in the common path; both together close the class.

## Testing

- 2 new regression tests in `connectionConfig.test.ts` (latch survives `history.replaceState` rewrite; adding the param without reload doesn't enable demo)
- Full webui suite: 57 suites / 538 tests pass; typecheck + lint clean
- E2E: local vite at HEAD reproduces the `demo://` fetch on demo→Agents; with this patch, all three entry/nav paths are clean (playwright, localhost:5700 blocked to simulate a real anonymous visitor)

### DIFF
--- a/webui/src/utils/__tests__/connectionConfig.test.ts
+++ b/webui/src/utils/__tests__/connectionConfig.test.ts
@@ -21,7 +21,9 @@ jest.mock('@/stores/servers', () => ({
 
 import {
   getConnectionConfigFromSources,
+  isDemoMode,
   processConnectionFromHash,
+  resetDemoModeForTests,
   resolveCloudExchangeBaseUrl,
 } from '../connectionConfig';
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Status: Ready to Merge

✅ All CI checks pass (23 checks all successful)
✅ Greptile review: 5/5 approval
✅ No blocking feedback

The fix is solid — latch demo mode on first browser call to prevent SPA navigation from splitting the demo-state signal. Tests are comprehensive, and the complementary logic with #3249 closes the class of URL-rewrite issues cleanly.

Ready for your review and merge whenever you have bandwidth.

PR #3255, opened 2026-07-15
fix(test): replace slow_stream timing hack with deterministic gate in TUI queue test

TimeToBuildBob:
## Summary

- **Root cause**: `test_e2e_queue_dispatches_after_turn` was using `time.sleep(0.25)` per stream chunk to keep generation busy while asserting the prompt queue. The mock model emits 4 words for "Echo: first message please", so total hold time is ~1 s. Textual's `await pilot.pause()` (plus the worker thread's `call_from_thread` notifications) can process the generation-done event during that window, clearing the queue before the assertion runs — yielding `AssertionError: assert [] == ['second message please']` on attempt-1 (the `KeyError: StashKey` on attempt-2 is a pytest-retry×pytest-9 machinery artefact, not the real failure).
- **Fix**: Replace `slow_stream` with a `threading.Event` gate (`first_turn_gate`). The first stream blocks on `gate.wait(timeout=10)` until the test explicitly calls `gate.set()` *after* checking the queue. Generation is guaranteed to still be in progress during the queue assertion, making it race-free.

## Test plan

- [x] `test_e2e_queue_dispatches_after_turn` passes deterministically (was: 1/3 runs failing)
- [x] All 4 non-tmux TUI tests pass: `test_e2e_generation_streams_and_renders`, `test_e2e_tool_confirm_execute`, `test_e2e_tool_skip`, `test_e2e_queue_dispatches_after_turn`
- [x] Pre-commit hooks pass (ruff, mypy, shellcheck)

Fixes ErikBjare/bob#1084 (gptme CI: test_e2e_queue_dispatches_after_turn failing with KeyError StashKey).

### DIFF
--- a/tests/test_tui_e2e.py
+++ b/tests/test_tui_e2e.py
@@ -153,33 +153,49 @@ async def test_e2e_tool_skip(mock_app, tmp_path):
 @pytest.mark.asyncio
 async def test_e2e_queue_dispatches_after_turn(mock_app, monkeypatch):
     """Prompts queued while generating are submitted when the turn ends."""
+    import threading
+
     import gptme.llm.llm_mock as llm_mock
 
     app = mock_app
     orig_stream = llm_mock.stream
 
-    def slow_stream(*args, **kwargs):
+    # Gate that holds the first stream open until after we verify the queue.
+    # time.sleep()-based delays are unreliable: the Textual event loop can
+    # process the worker-done message during await pilot.pause(), clearing the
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:

Evidence: commit: 3e902b26eb docs(journal): document gptme#3255 determinism fix

Session: `ff068cff-1688-5154-877b-c892d08525a7`

TimeToBuildBob:
@greptileai review

PR #3273, opened 2026-07-17
fix(llm): capture usage tokens from subscription response.done events

TimeToBuildBob:
## Summary

Three linked bugs prevented `openai-subscription` sessions (e.g. `gpt-5.6-sol`) from recording any token usage, causing ~2% token capture rate instead of ~95%+.

**Bug 1** — `_stream_responses_events()` only called `usage_callback` for `response.completed` events. Subscription streams emit `response.done` instead, so the callback never fired.

**Bug 2** — `_extract_usage_token_counts()` used `getattr()` for all field access. Subscription SSE events are raw dicts (not SDK objects), so every `getattr(usage, "input_tokens")` silently returned `None`.

**Bug 3** — `llm_openai_subscription.stream()` never passed a `usage_callback` and never returned metadata from the generator. `_StreamWithMetadata` fell back to `{"model": self.model}` with no usage.

**Fix:** (a) call `usage_callback` for both `response.completed` and `response.done`; (b) switch `_extract_usage_token_counts` to `_obj_get` (handles both dicts and SDK objects); (c) wire `usage_callback` in `stream()` and return captured counts as the generator's return value so `_StreamWithMetadata` attaches them to the message.

## Test plan

- [x] All 20 existing `test_llm_openai_subscription` tests still pass
- [x] New `test_stream_returns_usage_from_response_done`: verifies `response.done` with usage dict is returned as metadata
- [x] New `test_stream_returns_none_when_response_done_has_no_usage`: verifies no-usage event returns `None` (caller fallback)
- [x] New `test_stream_captures_usage_without_cache_fields`: verifies subscription responses without cache fields don't crash or add spurious fields

### DIFF
--- a/gptme/llm/llm_openai_subscription.py
+++ b/gptme/llm/llm_openai_subscription.py
@@ -44,13 +44,13 @@ from collections.abc import Generator
 from dataclasses import dataclass
 from math import isfinite
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 from urllib.parse import parse_qs, urlencode, urlparse
 from uuid import uuid4
 
 import requests
 
-from ..message import Message
+from ..message import Message, MessageMetadata
 from .openai_responses import (
     _extract_usage_token_counts,
     _messages_to_responses_input,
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:

Evidence: commit: 7b1201c16c docs(journal): gptme#3273 typecheck fix — subscription stream return type

Session: `ac4bef5e-13a2-5600-aad9-f8ae8cac6fad`

TimeToBuildBob:
@greptileai review

PR #3294, opened 2026-07-21
fix(logmanager): preserve tool call/result atomicity in prune_ephemeral_messages

### LINKED ISSUE #3293: bug(logmanager): ephemeral pruning can orphan tool call/result pairs, causing 400 on Responses API
## Summary

`prune_ephemeral_messages` in `gptme/logmanager/manager.py` drops expired ephemeral messages individually without preserving the atomicity of tool call / tool result pairs. When one member of a pair is dropped and the other is kept, the OpenAI Responses API (and similar strict providers) rejects the next turn with:

```
Codex API error 400: {"message": "No tool output found for function call call_XXX."}
```

## Root cause

```python
# logmanager/manager.py
def prune_ephemeral_messages(msgs: list[Message]) -> list[Message]:
    for msg in reversed(msgs):
        ttl = msg.ephemeral_ttl
        if ttl is not None and assistant_turns_after > ttl:
            continue  # drops entire message — no pairing check
        kept_reversed.append(msg)
```

If an assistant message carries both a thinking block (`ephemeral_ttl`) **and** `tool_calls`, dropping it leaves the corresponding system message (tool result, `call_id` set) without its anchor. Conversely, if the tool result happens to carry `ephemeral_ttl` while the function call does not, the function call is kept without its output.

`_merge_consecutive_messages` (called after pruning) guards against merging `call_id` messages, but does not drop orphaned tool results — so orphans survive into the API payload.

## Affected providers

Any provider using the Responses API format where tool call / tool result pairs must appear together: `openai-subscription/gpt-5.5` (Codex Responses API). Likely also `openai/gpt-5.5`, `codex/gpt-5.5`, and any future Responses-API provider.

## Observed symptoms

- Session fails ~60–120s after first ephemeral prune
- Error at `gptme/llm/llm_openai_subscription.py` stream path
- `tokens=0` (error occurs before any tokens are generated — pre-stream API validation)
- Arm self-heals after ephemeral messages age out naturally across turns

## Proposed fix

Enforce atomicity in `prune_ephemeral_messages`: when an ephemeral message that contains `tool_calls` is dropped, immediately drop all following system messages with `call_id` that belong to that call. Symmetrically, if the function call is kept and a subsequent tool result would be dropped (because it somehow has `ephemeral_ttl`), the function call should also be dropped.

A minimal approach (drop-companion strategy):

```python
# After deciding msg (with tool_calls) is expired, also mark its tool results as dropped
if ttl is not None and assistant_turns_after > ttl:
    if message_contains_tool_use(msg):
        # mark companion tool-results as also expired so they're skipped in the forward pass
        drop_tool_results_for = msg  # track to skip them
    continue
```

An alternative is to post-process: after pruning, scan for orphaned tool results (system messages whose preceding non-system message is missing) and drop those too. This is similar to the orphan guard already in `limit_log`.

## Context

- `limit_log` (in `gptme/util/reduce.py` lines 496–523) already has an orphan guard for the context-limit truncation path.
- `_merge_consecutive_messages` has a `call_id` guard (lines 742–757) for the merge path.
- `prune_ephemeral_messages` is the only pruning path without a pairing invariant.

## Workaround

None reliable. The arm self-heals after 2–3 turns once ephemeral messages age out of the context naturally (without pruning). A crash-loop breaker applies a 1h cooldown after 3 consecutive failures.

## Reproduction

Hard to trigger deterministically since it requires:
1. An assistant turn with both `ephemeral_ttl` (thinking block) **and** `tool_calls`
2. Enough subsequent assistant turns to expire the TTL
3. A provider that validates tool call/result atomicity at submission time

The `tests/test_ephemeral.py` suite does not currently test this path.

TimeToBuildBob:
## Problem

`prune_ephemeral_messages` dropped expired messages individually without checking tool call/result pair integrity. When an assistant message with `ephemeral_ttl` (e.g. a thinking block) also made tool calls, dropping it left the companion system messages (`call_id` set) without their anchor. The OpenAI Responses API (and similar strict providers) then return:

```
Codex API error 400: {"message": "No tool output found for function call call_XXX."}
```

Fixes #3293.

## Solution

Add `_drop_orphaned_tool_pairs` post-processing step called after the TTL prune pass. It enforces atomicity in both directions:

- **Case 1** — orphaned tool result: a system message with `call_id` whose nearest non-system predecessor was dropped → also drop the tool result
- **Case 2** — orphaned function call: an assistant message whose immediately following tool results (system+call_id) were dropped → also drop the assistant

The function iterates until stable to handle cascades, and has a fast-path when no `call_id` messages exist (the common case — zero overhead).

## Tests

Added 5 new test cases to `tests/test_ephemeral.py`:
- `test_expired_assistant_drops_companion_tool_results` — Case 1 (single result)
- `test_expired_assistant_drops_multiple_companion_tool_results` — Case 1 (multiple results)
- `test_dropped_tool_result_drops_companion_function_call` — Case 2
- `test_non_tool_call_assistant_not_dropped_by_orphan_guard` — guard doesn't affect non-tool messages
- `test_tool_result_without_call_id_not_dropped` — system messages without call_id are unaffected

All 25 tests in the file pass.

### DIFF
--- a/gptme/logmanager/manager.py
+++ b/gptme/logmanager/manager.py
@@ -729,7 +729,76 @@ def prune_ephemeral_messages(msgs: list[Message]) -> list[Message]:
             assistant_turns_after += 1
 
     pruned = list(reversed(kept_reversed))
+
+    # Enforce tool call / tool result atomicity: dropping an assistant message
+    # that contains tool calls must also drop its companion tool results, and
+    # vice versa — otherwise the Responses API returns 400.
+    pruned = _drop_orphaned_tool_pairs(msgs, pruned)
+
     return _merge_consecutive_messages(pruned)
+
+
+def _drop_orphaned_tool_pairs(
+    original: list[Message],
+    pruned: list[Message],
+) -> list[Message]:
+    """Drop messages that form broken tool call / tool result pairs.
+
+    Two directions are handled, iterating until stable:
+
+    Case 1 — orphaned tool result: a system message with call_id whose
+    nearest non-system predecessor in *original* was dropped by the TTL
+    prune.  Keeping it without the function call causes Responses-API 400.
+
+    Case 2 — orphaned function call: an assistant message immediately
+    followed (in *original*) by one or more system messages with call_id,
+    where at least one of those results is missing from *pruned*.  Keeping
+    the function call without all its outputs also triggers 400.
+    """
+    if not any(m.call_id for m in original):
+        return pruned
+
+    orig_idx = {id(m): i for i, m in enumerate(original)}
+    kept_ids = {id(m) for m in pruned}
+
+    changed = True
+    while changed:
+        changed = False
+        to_remove: set[int] = set()
+
+        for msg in pruned:
+            mid = id(msg)
+            if mid not in kept_ids or mid in to_remove:
+                continue
+            idx = orig_idx.get(mid)
+            if idx is None:
+                continue
+
+            if msg.role == "system" and msg.call_id:
+                # Case 1: walk backward to find nearest non-system anchor.
+                for j in range(idx - 1, -1, -1):
+                    if original[j].role != "system":
+                        if id(original[j]) not in kept_ids:
+                            to_remove.add(mid)
+                            changed = True
+                        break
+
+            elif msg.role == "assistant":
+                # Case 2: walk forward to find immediately following tool
+                # results; if any are missing, drop the function call too.
+                for j in range(idx + 1, len(original)):
+                    nxt = original[j]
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_ephemeral.py` (+120/-1, Δ121)
- `gptme/logmanager/manager.py` (+69/-0, Δ69)

### Suggested verification
- `uv run pytest tests/test_ephemeral.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3294._

PR #727, opened 2025-10-16
feat(cli): add Ctrl+V support for pasting images

TimeToBuildBob:
## Summary

Implements support for pasting images directly into gptme using Ctrl+V, similar to Claude Code.

### Core Functionality
- **clipboard.py**: Added paste() function to read from clipboard
  - Supports Linux (xclip/wl-paste) and macOS (pbpaste)
  - Returns clipboard text content or None if empty/error

- **clipboard.py**: Added is_image_path() function to detect image file paths
  - Checks common image extensions (.png, .jpg, .jpeg, .gif, .bmp, .webp, .svg, .ico)
  - Validates that path exists and points to a file
  - Handles quoted paths and whitespace

- **prompt.py**: Added Ctrl+V key binding to PromptSession
  - Reads clipboard content when Ctrl+V is pressed
  - If content is an image path: inserts view_image('path') command
  - If content is regular text: pastes text normally

### Testing
- Added comprehensive test suite: tests/test_util_clipboard.py
- 11 tests covering:
  - Valid image paths (various extensions)
  - Quoted paths (single/double quotes)
  - Paths with whitespace
  - Non-image files (correctly rejected)
  - Nonexistent files (correctly rejected)
  - Empty strings (correctly rejected)
  - Directories (correctly rejected)
  - Case-insensitive extension matching

## How It Works

1. User copies an image file path to clipboard
2. User presses Ctrl+V in gptme prompt
3. System detects it's an image path
4. Inserts view_image('path') command
5. User can press Enter to view the image using vision tool

## Benefits

- No need to manually type image paths
- Works with any image file on filesystem
- Uses existing vision tool for image viewing
- Graceful fallback to regular paste for non-image content

## Testing

All tests pass (11 passed in 0.23s)

Closes #655

----

> [!IMPORTANT]
> Adds Ctrl+V support for pasting images by detecting image paths from the clipboard and inserting a command to view them.
> 
>   - **Core Functionality**:
>     - `clipboard.py`: Adds `paste()` to read clipboard content, supporting Linux and macOS, returning text or None.
>     - `clipboard.py`: Adds `is_image_path()` to check if content is a valid image path, handling common extensions and path validation.
>     - `prompt.py`: Adds Ctrl+V key binding to `PromptSession` to handle clipboard content, inserting `view_image('path')` for image paths.
>   - **Testing**:
[...]

PR #3297, opened 2026-07-21
feat(hooks): hash-link manifest records with tamper-detection verification (Phase 2)

TimeToBuildBob:
## Summary

Phase 2 of the tool-call manifest chain (Phase 1: #3291). Hash-links the pre/post manifest records so every tool-call record commits to its predecessor, forming a tamper-evident chain.

### `gptme/hooks/manifest.py`

- Each record now carries a `hash` field (sha256 of its own content, deterministic via key-sorted JSON)
- Each record now carries a `prev_hash` field linking to the preceding record
- Chain order: `pre(1) → post(1) → pre(2) → post(2) → …`
- Genesis pre-record has `prev_hash: null`
- New `verify_manifest_chain(manifest_dir)` function detects:
  - **Self-hash mismatches**: record content was tampered with
  - **Chain-link breaks**: `prev_hash` doesn't match predecessor's hash
  - **Missing records**: gap in the sequence
  - **Extra/forged records**: unknown records in the manifest directory

### Tests (`gptme/hooks/tests/test_manifest.py`)

- Updated existing 5 tests to verify `hash` and `prev_hash` fields
- **5 new chain-integrity tests**: full chain verification, pre→post linking, cross-call linking, self-consistency
- **4 new tamper-detection regression tests**: content tamper, prev_hash tamper, record deletion, forged insertion
- **2 new edge-case tests**: empty dir verification, missing post-record detection

## Test plan

```
uv run pytest gptme/hooks/tests/test_manifest.py -v   # 15/15 pass
```

## Related

- Phase 1: #3291
- Research: `knowledge/research/2026-07-19-clodex-ide-verifiable-code-generation-study.md`
- Idea: #727 in Bob's idea backlog

### DIFF
--- a/gptme/hooks/manifest.py
+++ b/gptme/hooks/manifest.py
@@ -150,126 +150,125 @@ def register_manifest_hooks(manifest_dir: Path) -> None:
     logger.debug("Manifest hooks registered, writing to %s", manifest_dir)
 
 
-def verify_manifest_chain(manifest_dir: Path) -> list[str]:
-    """Verify the hash-linked chain of manifest records in *manifest_dir*.
-
-    Returns a list of error strings (empty list = chain is intact).
-    Each error describes a specific integrity violation.
-
-    Checks performed:
-    1. Every record's ``hash`` matches its own content.
-    2. Every record's ``prev_hash`` matches the ``hash`` of the preceding
-       record (in (pre → post → pre → post) order).
-    3. The chain is contiguous (no gaps in sequence numbers).
-    """
+def _verify_session_chain(
+    session_id: str,
+    entries: list[tuple[Path, dict[str, Any]]],
+) -> list[str]:
+    """Verify one session's ordered pre/post record chain."""
     errors: list[str] = []
-
-    # Collect and sort records by (sequence, phase).
-    pre_files = sorted(manifest_dir.glob("*-pre.json"))
-    post_files = sorted(manifest_dir.glob("*-post.json"))
-
-    if not pre_files and not post_files:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the 5 Greptile P1/P2 findings in `2c2170cb7`:

**P1 — Failed write advances chain tail**: `_write_record()` now returns `bool`; callers only update `prev_hash` when the file was actually persisted. A transient disk-full or permission error no longer forges a dangling chain link.

**P1 — Non-object JSON crashes verifier**: After `json.loads()`, a type check rejects non-dict values (`[]`, `"text"`, numbers) with an integrity error rather than an `AttributeError`.

**P1 — Non-positive sequences skip verification**: Sequence validation now requires `seq > 0`; zero/negative values are rejected before they can slip past the `range(1, …)` loop.

**P1 — Duplicate sequences hide records**: When two sessions share a `--manifest-dir` and write the same sequence number, the verifier now emits a duplicate-sequence error rather than silently discarding one session's record.

**P2 — Forged large sequence exhausts resources**: `verify_manifest_chain` now iterates `all_seqs` directly instead of `range(1, all_seqs[-1]+1)`, so a crafted record with sequence `10_000_000` doesn't spin for seconds. Gap ranges are reported as compact strings (`Sequences 2–9999999: missing (9999998 records)`).

Added 5 regression tests (`TestManifestVerifierRobustness`) covering each fix; all 20 tests pass.

TimeToBuildBob:
@greptileai review

PR #3301, opened 2026-07-21
fix(test): wait for input prompt before asserting scrollback order in inline test

TimeToBuildBob:
## Summary

- `test_inline_prints_to_scrollback` captured the tmux pane the moment `"Echo: hello inline"` appeared, before `"Type a message"` (the input prompt) had re-rendered in `--inline` mode
- This caused `pane.index("Type a message")` to raise `ValueError: substring not found` on attempt 1
- On retry attempt 2 the timing worked out; but the retry triggered the known pytest-retry × tmp_path StashKey teardown issue (#3256), keeping CI red

## Fix

Added a second `wait_for("Type a message")` call after the echo check, so the snapshot is taken only once both strings are present in the pane — eliminating the timing race.

## Context

Identified while investigating the CI failure on #3283 (dependabot bump) where Erik asked whether #3256 should have fixed the red CI. #3256's teardown guard should have suppressed the artifact, but the underlying test flakiness was still driving the retry that triggered it. This PR addresses the root cause.

### DIFF
--- a/tests/test_tui_e2e.py
+++ b/tests/test_tui_e2e.py
@@ -445,7 +445,11 @@ class TestTmuxInlineMode:
         tui.wait_ready()
         tui.send("hello inline")
         tui.send_key("Enter")
-        pane = tui.wait_for("Echo: hello inline")
+        tui.wait_for("Echo: hello inline")
+        # Wait for both the echo AND the input prompt to be visible together —
+        # captures taken the moment "Echo:" appears can race before the prompt
+        # re-renders, causing ValueError from str.index().
+        pane = tui.wait_for("Type a message")
         assert_no_escape_garbage(pane)
         # transcript lines are plain terminal output (native scrollback), so
         # they appear above the live region which contains the input


TimeToBuildBob:
The CI failure (`test_prompt_respects_max_steps`) is unrelated to this PR's changes — it's a regression introduced by #3300 that has been red on master since `a3b12126`.

Root cause: `run_chat_turn` returns `"max_steps"` as the ACP stop reason, but `"max_steps"` is not a valid `PromptResponse` stop_reason (ACP only accepts `end_turn`, `max_tokens`, `max_turn_requests`, `refusal`, `cancelled`). Pydantic raises a `ValidationError`, which the `except Exception` handler catches, returning `stop_reason="cancelled"` instead. The test then fails with `assert 'cancelled' == 'max_steps'`.

Fix is in #3302 (uses `"max_turn_requests"` — the correct ACP stop reason for a per-turn step cap). Once #3302 lands and CI goes green on master, this PR's CI should be clear to rerun.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto current master and dropped the ACP stop_reason commit — that fix landed separately as #3302. This PR now contains only the `wait_for("Type a message")` timing fix for `test_inline_prints_to_scrollback`.

PR #3325, opened 2026-07-23
fix(server,webui): external sessions dedup, stable detail IDs, admin nav label

ErikBjare:
Fixes three issues Erik hit in the webui today:

1. **Every local conversation listed twice** with the Ext toggle on: gptme-harness external discovery scans the same logs dir the server serves natively, so each chat appeared both as native (`2026-07-23-crawling-funny-fish`) and external (`gptme/external-extracted ...`). External discovery now excludes sessions under the server's own logs dir.

2. **Listed external sessions 404 on click** (`GET /api/v2/external-sessions/<id>` → "External session not found"): catalog IDs were derived from the transcript's `trajectory_path` (symlink-resolved, e.g. `~/.claude` → `~/dotfiles/home/.claude`) while the detail lookup hashed the *discover* path (unresolved) — so the IDs never matched. Both sides now derive the ID from the same discovered-path string. Live-reproduced against a running server before the fix.

3. **CLI provider spammed warnings** (`gptme-sessions transcript <dir> --json failed (exit 2)`): gptme discovery returns session *directories* but the `transcript` subcommand only accepts files. Directories are now normalized to their `conversation.jsonl` (or skipped).

Plus: the **Admin** sidebar entry is now rendered through the same `navItems` path as other sections, so it gets a visible label when the sidebar is expanded and an aria-label (was a bare unlabeled icon button).

Tests: 8 new regression tests (ID consistency for both providers, own-logs filtering, dir normalization); all 31 external-session tests pass.

### DIFF
--- a/gptme/server/external_sessions.py
+++ b/gptme/server/external_sessions.py
@@ -11,10 +11,7 @@ import shutil
 import subprocess
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from pathlib import Path
+from pathlib import Path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


ErikBjare:
@greptileai review

PR #3338, opened 2026-07-23
test(tui): add SVG visual regression tests for TUI rendering

### LINKED ISSUE #3335: test(tui): add screenshot-based visual regression coverage
## Problem

PR #3334 exposed a hole in the TUI test loop: computed-style assertions passed while the terminal still rendered the wrong background, and a later whole-theme workaround fixed the strip while visibly changing the foreground/accent palette. Erik had to supply screenshots to catch both regressions.

## Goal

Add a deterministic visual-regression harness for `gptme-tui` that captures representative rendered states and compares them against reviewed baselines.

## First slice

- Capture both regular and `--inline` TUI states at a fixed terminal size.
- Cover at least the idle/placeholder state that regressed in #3334.
- Render under a fixed terminal theme so terminal-default colors resolve predictably.
- Compare captures pixel-by-pixel (or via an explicit small tolerance) and emit a diff artifact on failure.
- Provide one local command to regenerate/review baselines.
- Run the visual check in CI.

Textual already exposes `App.export_screenshot()` / `save_screenshot()` for deterministic SVG capture in headless tests. We should evaluate that as the simplest harness first; if it cannot represent terminal-native default colors faithfully, drive a real terminal under Xvfb and capture pixels instead.

## Acceptance criteria

- A change that reintroduces the gray active-output strip fails the visual test.
- A change that globally swaps the existing Textual foreground/accent palette also fails.
- CI publishes the actual, expected, and diff captures when the comparison fails.

Origin: #3334, especially Erik's screenshot/tooling request in [this comment](https://github.com/gptme/gptme/pull/3334#issuecomment-5058256892).

TimeToBuildBob:
## Summary

Closes #3335.

Adds `tests/test_tui_visual.py` with six tests covering TUI visual correctness:

**Gray background detection** (idle, stream, inline modes):
- Asserts no mid-gray Textual `$panel` color (`#262626`) appears in content-area rects
- Detects the regression class from #3334 where `ActiveOutput` rendered with a gray strip instead of transparent
- Currently marked `xfail` because the bug still exists on master — will become passing once the `native_ansi_color=True` fix in #3334 lands

**Full SVG snapshots** (idle, stream, inline modes):
- Compares the full rendered SVG against committed baselines in `tests/snapshots/`
- Catches palette changes, layout shifts, and background color regressions beyond just the gray check
- Baselines committed alongside the tests

## Usage

Regenerate baselines after an intentional visual change:
```bash
pytest tests/test_tui_visual.py --snapshot-update
```

On snapshot mismatch, the actual SVG is written as `*.actual.svg` next to the baseline, and CI uploads those as an artifact for easy visual inspection.

## Test plan

- [x] All 6 tests run locally under `uv run --extra tui`
- [x] 3 snapshot tests pass (baselines generated and stored)
- [x] 3 gray-check tests are `xfail` (documenting the known bug, CI stays green)
- [x] `--snapshot-update` flag wired in `conftest.py`
- [x] CI artifact upload added to `test.yml` (`tui-visual-snapshot-diffs` on failure)
- [x] Pre-commit hooks pass (trailing whitespace auto-fixed in SVG baselines)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
**Root cause found and fixed** — f07dba7

After 8 failing monitoring sessions, the actual diff between baseline and actual SVG was identical content, differing only by a trailing newline. Pre-commit hooks add `\n` when committing SVG baseline files, but `textual`'s `export_screenshot()` doesn't produce one — so the byte-for-byte comparison always failed despite correct visual output.

Fix: `normalized.rstrip("\n") != baseline.rstrip("\n")` in `_snapshot_check` (one line). All 3 snapshot tests pass locally now.

The Greptile P1 finding was already resolved by the normalization in `a8da71f5b` — the `STATUS_BAR` replacement was working correctly. The trailing newline was the remaining blocker. Regarding P2 (pin upload-artifact SHA): the entire workflow uses mutable tags (`checkout@v7`, `setup-python@v6`, `codecov@v7`) — keeping the new step consistent with the established convention.

TimeToBuildBob:
**Status update** — Greptile 5/5, CI re-run triggered.

The remaining CI failure (`test_server_v2_auto_stepping::test_multi_tool_per_message`) is a pre-existing flaky timing test — this PR doesn't touch `tests/test_server_v2_auto_stepping.py` at all. The test is marked `@pytest.mark.no_retry` because retrying it causes a `tmp_path` teardown `KeyError`; the failure is a race on `message_added` event ordering under CI load. Re-run triggered — expect it to go green.

All Greptile threads resolved (P1 snapshot normalization fixed in `8fe624628`, P2 mutable-tag finding dismissed in-line with the established convention in the rest of `test.yml`). PR should be ready to merge once CI goes green.

TimeToBuildBob:
@greptileai review

PR #3339, opened 2026-07-23
test(tui): activate gray-background regression guards (xfail cleanup after #3334)

### LINKED ISSUE #3335: test(tui): add screenshot-based visual regression coverage

TimeToBuildBob:
## Summary

The three `test_*_no_gray_background` tests were marked `xfail(strict=False)` while the `native_ansi_color` fix was pending in #3334. That PR merged, so the markers are now counterproductive: if the gray bug were re-introduced, the tests would flip back to *expected-fail* status (CI stays green, regression silently hidden).

This PR removes the `xfail` markers so the gray-background tests actively enforce the fix going forward.

## What changed

- Removed `_GRAY_BUG_XFAIL` marker variable and its three applications in `tests/test_tui_visual.py`
- All 6 visual regression tests now pass cleanly (verified locally)

## Closes

Closes #3335

### DIFF
--- a/tests/test_tui_visual.py
+++ b/tests/test_tui_visual.py
@@ -134,19 +134,6 @@ def snapshot_update(request):
 # ── tests ─────────────────────────────────────────────────────────────────────
 
 
-# These three tests document the correct behavior: no mid-gray ($panel/#262626) should
-# appear in the content area.  They currently fail on master because the fix (native_ansi_color=True
-# on ActiveOutput) has not landed yet — see gptme#3334.  Mark xfail so CI stays green; remove
-# the markers once #3334 (or equivalent) merges.
-_GRAY_BUG_XFAIL = pytest.mark.xfail(
-    strict=False,
-    reason="Content-area gray background still present on master (#262626 from Textual $panel). "
-    "Will pass once native_ansi_color fix in #3334 is merged.",
-    run=True,
-)
-
-
-@_GRAY_BUG_XFAIL
 @pytest.mark.asyncio
 async def test_idle_state_no_gray_background(tmp_path):
     """Idle TUI must not render any mid-gray content-area background (regression #3334)."""
@@ -157,7 +144,6 @@ async def test_idle_state_no_gray_background(tmp_path):
     _check_no_gray_rects(svg, "idle state")
 
 
-@_GRAY_BUG_XFAIL
 @pytest.mark.asyncio
 async def test_stream_state_no_gray_background(tmp_path):
     """Streaming state must not render a gray active-output strip (regression #3334)."""
@@ -170,7 +156,6 @@ async def test_stream_state_no_gray_background(tmp_path):
     _check_no_gray_rects(svg, "streaming state")
 
 
-@_GRAY_BUG_XFAIL
 @pytest.mark.asyncio
 async def test_inline_mode_no_gray_background(tmp_path):
     """Inline mode must not render any mid-gray content-area background."""


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_tui_visual.py` (+0/-15, Δ15)

### Suggested verification
- `uv run pytest tests/test_tui_visual.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3339._

ErikBjare:
@TimeToBuildBob I had to do the changes in https://github.com/gptme/gptme/pull/3340 so I don't really think this regression guard had any effect. Idk.

TimeToBuildBob:
The guard in these tests specifically targets **rect-element fill colors** in the mid-gray range (roughly `#262626` / Textual `$panel`/`$surface` family) — the artifact that `native_ansi_color=True` in #3334 was meant to stop. The tests did go xpass after #3334, which confirmed that specific fix worked.

From the #3340 description ("italics stuff which caused the gray background on certain elements"), it sounds like that regression came from a different SVG source — text or inline styling, not rect fills — so `_check_no_gray_rects` would be blind to it regardless of xfail status.

That said, removing the xfail markers (this PR) still provides real value: with `strict=False` the tests are effectively dead — both pass and fail outcomes are silently accepted, so a rect-based gray re-introduction would never break CI. Once the markers are gone, that class of regression breaks the build.

If you want broader coverage for the italics/styling class too, I can add a follow-up that inspects text-element styles in the SVG (or compares against the corrected snapshot from #3340). Otherwise this PR is ready to merge as a partial-but-real guard — just needs to be rebased after #3340 lands (no file conflicts expected since we only touch `test_tui_visual.py` and #3340 touches `app.py` + snapshots).

ErikBjare:
@TimeToBuildBob Yes, do that.

TimeToBuildBob:
Follow-up landed in #3344. Adds `_check_no_italic_text()` to the three existing visual tests — scans the SVG `<style>` block for `font-style: italic` on terminal cell classes, which is unambiguous (the `@font-face` blocks only ever use `normal`/`bold`). Tests pass on current master.

The thinking-block CollapsibleTitle path isn't exercised by idle/stream/inline scenarios; a future test that renders a message with `<thinking>` content would close that gap.

PR #3342, opened 2026-07-24
fix(gh): keep the ref intact when turning a blob URL into a raw URL

eeshsaxena:
`transform_github_url` turns a `/blob/<ref>/` URL into `/raw/refs/heads/<ref>/`, which only resolves when `<ref>` is a branch.

## Problem

```python
return url.replace("/blob/", "/raw/refs/heads/")
```

`refs/heads/<ref>` names a branch. On a **tag** or a **commit SHA** the raw endpoint 404s, so fetching a file from those URLs returns nothing. GitHub permalinks (the `y` shortcut) are SHA-based, so this is the common case, not an edge one. It is reached by `browser.py` and `context.py` when they fetch GitHub file content.

Verified against github.com (same repo and path, following redirects):

| ref | `/raw/refs/heads/<ref>/` (current) | `/raw/<ref>/` (fix) |
|---|---|---|
| commit SHA | 404 | 200 |
| tag `v2.31.0` | 404 | 200 |
| branch `main` | 200 | 200 |

## Fix

The raw endpoint resolves branches, tags and SHAs directly, so keep the ref as-is: `/blob/` -> `/raw/`. Branches keep working; tags and SHAs start working.

## Tests

`transform_github_url` had no test. Added cases for a branch, a tag, a SHA, and non-blob / non-GitHub URLs. I confirmed the tag and SHA cases fail against the current code and pass with the fix; `ruff` is clean. (One unrelated network test, `test_get_github_pr_content_with_unresolved`, fails on a clean checkout in my environment too.)

### DIFF
--- a/gptme/util/gh.py
+++ b/gptme/util/gh.py
@@ -292,12 +292,17 @@ def transform_github_url(url: str) -> str:
     Transform GitHub blob URLs to raw URLs to get file content without UI.
 
     Transforms:
-    https://github.com/{owner}/{repo}/blob/{branch}/{path}
+    https://github.com/{owner}/{repo}/blob/{ref}/{path}
     to:
-    https://github.com/{owner}/{repo}/raw/refs/heads/{branch}/{path}
+    https://github.com/{owner}/{repo}/raw/{ref}/{path}
+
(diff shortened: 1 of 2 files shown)


ErikBjare:
Thanks!

PR #3340, opened 2026-07-23
fix: improve tui styling

ErikBjare:
Finally fixes the italics stuff which caused the gray background on certain elements for me. Also some margins/spacing.

PR #3344, opened 2026-07-24
test(tui): guard against italic text-style regression (#3340)

TimeToBuildBob:
## Summary

- Adds `_italic_cell_classes()` helper that extracts terminal cell CSS class names from the SVG `<style>` block that carry `font-style: italic`
- Adds `_check_no_italic_text()` that fails if any such class is found
- Applies both checks to the three existing `no_gray_background` tests (idle, stream, inline) — no extra TUI renders needed

## Background

This is the follow-up requested in gptme/gptme#3339's review thread after #3340 landed. The `text-style: italic` on `.thinking-block CollapsibleTitle` was the specific bug fixed by #3340; these guards catch any future re-introduction of italic on content-area elements in the base UI scenarios.

**How detection works**: Textual encodes per-cell text styling as `.terminal-HASH-rN { fill: #color }` rules in the SVG `<style>` block. Italic adds `font-style: italic` to those rules. The `@font-face` declarations also contain `font-style` descriptors (`normal` / `bold`) but never `italic`, so a match is unambiguous.

**Scope**: The three test scenarios (idle, stream, inline) don't exercise the thinking-block `CollapsibleTitle` code path (which would require a message with `<thinking>` content to be rendered). A follow-up test that renders a thinking block would cover that exact path — left for a future PR since it requires more test scaffolding.

## Test plan

- [x] `test_idle_state_no_gray_background` — passes (adds `_check_no_italic_text`)
- [x] `test_stream_state_no_gray_background` — passes
- [x] `test_inline_mode_no_gray_background` — passes
- [x] Pre-commit hooks clean

### DIFF
--- a/tests/test_tui_visual.py
+++ b/tests/test_tui_visual.py
@@ -95,6 +95,35 @@ def _check_no_gray_rects(svg: str, label: str) -> None:
     )
 
 
+def _italic_cell_classes(svg: str) -> list[str]:
+    """Return terminal cell CSS class names that apply italic font-style.
+
+    Textual encodes per-cell styling as ``.terminal-HASH-rN { fill: #color }``
+    rules in the SVG ``<style>`` block.  Italic text adds ``font-style: italic``
+    to those rules.  The ``@font-face`` declarations also contain ``font-style``
+    descriptors (``normal`` / ``bold``) but never ``italic``, so a match here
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
@greptileai review

Issue #3343, opened 2026-07-24
Poor pretty-rendering/parsing of non-markdown tooluse formats, and other `--tool-format` issues


ErikBjare:
In both gptme CLI and gptme-tui, when the agent uses a tool it looks okay with the default markdown tooluse format (although seeing the entire tooluse in gptme CLI can often be a bit spammy, but that's an accepted limitation of the gptme CLI similar to thinking which also outputs raw).

We've done some work to ensure that when you start a conversation with a particular model then we use the recommended `--tool-format` for that model (`tool` for GPT-5.x models, for instance). Which makes `gptme-tui --model openai-subscription/gpt-5.6-sol` work (although launching with e.g. `--model openrouter/deepseek/deepseek-v4-flash` and then switching to gpt-5.6-sol doesn't really work that well - still using `markdown` format), but also makes it output msgs like:

```
@ipython(call_zcsbIQLULsLNjZEkO13cIfvU): {"code":"2 + 2"}
```

Which look fine in the basic case (single row, short), but for longer outputs it quickly becomes an unreadable mess:

```
The chart rendered successfully, but the tracked-file query exposed a glob subtlety: tests/**/*.py omitted
top-level tests/test_*.py files. I’ll correct that before drawing conclusions.
@ipython(call_yZhTTdv94bczyHBLI4aYhdXP): {"code":"import subprocess, ast\nfrom pathlib import Path\nfrom
collections import defaultdict\nimport matplotlib\nmatplotlib.use("Agg", force=True)\nimport
matplotlib.pyplot as plt\n\nroot = Path.cwd()\nall_tracked = subprocess.run(["git", "ls-files"], cwd=root,
text=True, capture_output=True, check=True).stdout.splitlines()\ntracked = [p for p in all_tracked if
p.endswith(".py") and p.startswith(("gptme/", "tests/", "scripts/"))]\n\nstats = []\nby_area =
defaultdict(lambda: dict(files=0, lines=0, functions=0, classes=0))\nfor rel in tracked:\n path = root /
rel\n if not path.is_file(): continue\n source = path.read_text(errors="replace")\n try: tree =
ast.parse(source)\n except SyntaxError: continue\n area = Path(rel).parts[0]\n row = dict(path=rel,
area=area, lines=len(source.splitlines()),\n functions=sum(isinstance(n, (ast.FunctionDef,
[...]

PR #3346, opened 2026-07-24
fix(tui): prevent crash when GitHub URL is pasted in TUI

### LINKED ISSUE #3345: TUI crashes when github link sent
After pasting a link like https://github.com/gptme/gptme/issues/3343 in gptme-tui and submitting, the app immediately crashed with the error:

```
<sys>:0: RuntimeWarning: coroutine 'Application.run_async' was never awaited
```

TimeToBuildBob:
## Root cause

When a user pastes a GitHub URL (e.g. `https://github.com/gptme/gptme/issues/3343`) into the TUI and submits it, the app crashed with:

```
<sys>:0: RuntimeWarning: coroutine 'Application.run_async' was never awaited
```

The bug path:
1. CLI starts with `interactive=True` → `cli_confirm` hook is registered
2. TUI mounts, registers `tui_confirm` at priority 100 to override `cli_confirm` for tool confirmations
3. User submits a GitHub URL → `_submit()` → `include_paths()` → `_is_interactive_mode()` checks for `cli_confirm` hook → **still registered**, returns `True`
4. `_confirm_urls()` calls `prompt_alert()` → `PromptSession.prompt()` → tries to run a prompt_toolkit `Application` inside Textual's **already-running** asyncio event loop → crash

## Fix

`_is_interactive_mode()` now checks `asyncio.get_running_loop()` first. If a loop is running (Textual TUI context), it returns `False` immediately — skipping the prompt_toolkit confirmation dialog that would crash. `include_paths()` then falls through to non-interactive mode and fetches the URL automatically (which is the right UX: the user explicitly submitted it).

## Changes

- `gptme/util/context.py`: add asyncio event loop guard at top of `_is_interactive_mode()`
- `tests/test_context.py`: add regression test verifying the function returns `False` inside `asyncio.run()`

Fixes #3345

### DIFF
--- a/gptme/util/context.py
+++ b/gptme/util/context.py
@@ -38,8 +38,20 @@ def _is_interactive_mode() -> bool:
     """Check if we're in interactive CLI mode.
 
     Returns True if the cli_confirm hook is registered, indicating
-    interactive mode with confirmation enabled.
+    interactive mode with confirmation enabled. Returns False if called
+    from inside a running asyncio event loop (e.g. the Textual TUI),
+    because prompt_toolkit's sync prompts crash in that context.
     """
+    import asyncio
+
+    try:
+        asyncio.get_running_loop()
+        # Inside a running event loop (e.g. Textual TUI); prompt_toolkit
+        # sync Application.run() is not safe here.
+        return False
+    except RuntimeError:
+        pass  # No running loop — safe for interactive prompts
+
     try:
         from ..hooks import HookType, get_hooks
 
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -507,3 +507,27 @@ def test_is_interactive_mode():
     # (This test runs outside the normal CLI context)
     result = _is_interactive_mode()
     assert isinstance(result, bool)
+
+
+def test_is_interactive_mode_false_inside_event_loop():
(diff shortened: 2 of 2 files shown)


ErikBjare:
@TimeToBuildBob Duplicate of just-merged #3348

TimeToBuildBob:
Thanks for the heads up, Erik. Got unlucky with the timing — I opened #3346 around the same time #3348 was being worked on. No action needed here; just closing the loop.

ErikBjare:
@TimeToBuildBob Btw, on "(which is the right UX: the user explicitly submitted it)" - the point of asking is that the user may have copy-pasted a long text containing a URL that the user submitted not intending for the agent to read it all, as it could pollute the context. That's why the prompt exists, a full fix would make it prompt in the TUI too (consistent behavior across CLI & TUI).

TimeToBuildBob:
Good point — I concede the UX argument. A user who pastes a long block of text that happens to contain URLs isn't explicitly asking the agent to fetch them. The prompt is a safeguard against context pollution, not just a "did you mean to do this?" check.

Consistent behavior (TUI prompting the same way CLI does) is the right end state. The #3348 fix only addressed the crash; the prompt parity is still missing in the TUI. I'll file a follow-up issue to track it separately.

PR #3347, opened 2026-07-24
fix(tui): update tool format on /model switch, pretty-render tool-format calls

### LINKED ISSUE #3343: Poor pretty-rendering/parsing of non-markdown tooluse formats, and other `--tool-format` issues

TimeToBuildBob:
Fixes #3343 (two of the three sub-problems).

### Fix 1 — `/model` switch now updates the tool format

`cmd_model` in `gptme/commands/llm.py` now calls `set_tool_format()` with
the new model's `default_tool_format` immediately after `set_default_model()`.
Previously, switching from a markdown-format model to `gpt-5.6-sol` (which
uses `tool` format) left the session in `markdown` mode. The model would emit
`@tool(): {}` calls that gptme couldn't parse, then hallucinate a closing fence
to "complete" the incomplete block — preventing tool execution entirely.

`GptmeApp._handle_command` also syncs `self.tool_format` from the global after
any command, so the generation worker always uses the current format.

### Fix 2 — Pretty-rendering of `tool`-format calls in the TUI

`AssistantMessage.compose` now detects `@tool_name(call_id): {json}` lines
(the `tool` format emitted by GPT-5.x models) and renders them as collapsible
blocks with syntax-highlighted code extracted from the JSON body — exactly like
thinking blocks. Long ipython code dumps that were previously an unreadable flat
JSON blob become a compact collapsed header:

```
▶ ipython: import subprocess, ast…
```

The implementation adds:
- `_TOOL_CALL_RE` regex to detect `@tool_name(call_id): {...}` lines
- `_split_tool_calls(text)` — segments a text block into tool-call and prose pairs
- `_tool_call_renderable(call_text)` — extracts `(title, code, lang)` for display

### Not in this PR

Problem 3 from the issue (consistent format detection across all three formats
in the TUI) — that's a larger refactor and can follow separately.

### DIFF
--- a/gptme/commands/llm.py
+++ b/gptme/commands/llm.py
@@ -75,6 +75,13 @@ def cmd_model(ctx: CommandContext) -> None:
         chat_config.model = new_model
         chat_config.save()
         print(f"Set model to {new_model}")
+        # Auto-update tool format when the new model prefers a different one
+        model_meta = get_default_model()
+        if model_meta and model_meta.default_tool_format:
+            from ..tools.base import set_tool_format  # fmt: skip
+
+            set_tool_format(model_meta.default_tool_format)
+            print(f"Switched tool format to '{model_meta.default_tool_format}'")
     else:
         model = get_default_model()
         if not model:
--- a/gptme/commands/llm.py
+++ b/gptme/commands/llm.py
@@ -3,8 +3,17 @@ LLM-related commands: model, tools, context, tokens.
 """
 
 from collections import defaultdict
+from typing import TYPE_CHECKING
+from uuid import uuid4
 
 from .base import CommandContext, command
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from ..config import ChatConfig
+    from ..message import Message
+    from ..tools import ToolFormat
 
 
 def _complete_model(partial: str, _prev_args: list[str]) -> list[tuple[str, str]]:
@@ -58,30 +67,65 @@ def _complete_model(partial: str, _prev_args: list[str]) -> list[tuple[str, str]
     return unique_completions[:30]  # Limit to 30 completions
 
 
+def _replacement_prompt(
+    chat_config: "ChatConfig",
+    *,
+    model: str,
+    tool_format: "ToolFormat",
+) -> list["Message"]:
+    """Generate a replacement startup prompt after runtime configuration changes."""
+    from ..config import get_project_config  # fmt: skip
+    from ..prompts import get_prompt  # fmt: skip
+    from ..tools import get_tools  # fmt: skip
+
+    project_config = get_project_config(chat_config.workspace)
+    prompt = (project_config.system if project_config else None) or "full"
+    if chat_config.system_prompt:
+        prompt = chat_config.system_prompt
+    return get_prompt(
+        get_tools(),
+        prompt=prompt,
+        interactive=chat_config.interactive,
+        tool_format=tool_format,
+        model=model,
+        workspace=chat_config.workspace,
+        agent_path=chat_config.agent,
+        prompt_generation=uuid4().hex,
+    )
+
+
 @command("model", completer=_complete_model)
-def cmd_model(ctx: CommandContext) -> None:
+def cmd_model(ctx: CommandContext) -> "Generator[Message, None, None]":
     """Show or switch the current model."""
     from ..config import ChatConfig  # fmt: skip
     from ..llm.models import (  # fmt: skip
         get_default_model,
         set_default_model,
     )
+    from ..tools.base import get_tool_format, set_tool_format  # fmt: skip
 
     if ctx.args:
         new_model = ctx.args[0]
         set_default_model(new_model)
-        # Persist the model change to config so it survives restart/resume
+        model_meta = get_default_model()
+        assert model_meta is not None
+        new_tool_format = model_meta.default_tool_format or get_tool_format()
+
+        # Persist all dependent runtime state before appending a replacement
+        # generated prompt. Historical prompts remain in the append-only log;
+        # prepare_messages() only sends the newest generation to providers.
         chat_config = ChatConfig.from_logdir(ctx.manager.logdir)
         chat_config.model = new_model
+        chat_config.tool_format = new_tool_format
         chat_config.save()
+        set_tool_format(new_tool_format)
+        yield from _replacement_prompt(
+            chat_config,
+            model=model_meta.full,
+            tool_format=new_tool_format,
+        )
         print(f"Set model to {new_model}")
-        # Auto-update tool format when the new model prefers a different one
-        model_meta = get_default_model()
-        if model_meta and model_meta.default_tool_format:
-            from ..tools.base import set_tool_format  # fmt: skip
-
-            set_tool_format(model_meta.default_tool_format)
-            print(f"Switched tool format to '{model_meta.default_tool_format}'")
+        print(f"Switched tool format to '{new_tool_format}'")
     else:
         model = get_default_model()
         if not model:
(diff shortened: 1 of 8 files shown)


ErikBjare:
@TimeToBuildBob This PR doesn't address the fact that tool-format plays into generating the system prompt (tool-format specific) like i mentioned in https://github.com/gptme/gptme/issues/3343#issue-4968127547 ("re-generate the tool-format dependent parts of the system prompt") which may be a problem given the append-only nature of the conversation log (branch? fork? what happens when e.g. tools change which might also necessitates a system prompt regeneration? or does what this PR currently does - without re-generating the markdown/xml tool-format examples in system prompt - work from `markdown` -> `tool` format because `tool` format doesn't have examples in its format but instead passes a `tools` spec to each LLM call?)

TimeToBuildBob:
You're right: the current PR is incomplete on that point.

For `markdown -> tool`, execution will probably work more often than the reverse because every generation already passes the native `tools` schema when `tool_format == "tool"`. But the persisted startup prompt still contains the markdown-format tool instructions/examples and the old model-specific base prompt. That's contradictory context, so relying on the native schema to outweigh it is not a sound fix.

The same stale-prompt problem already exists for `/tools load`: it appends the new tool's prompt, but doesn't replace the startup tool overview. More generally, changing model, tool format, or loaded tools invalidates the generated core prompt.

I don't want `/model` to mutate historical log records or silently fork the user's conversation. I'll update this PR to append a **pinned replacement core prompt** generated for the new model/format and make prompt preparation treat the newest replacement as authoritative, dropping superseded generated core prompts from the request while preserving them on disk. That gives us append-only history, correct provider context, and a reusable mechanism for `/tools load`. I'll also fix Greptile's multiline-rendering finding in the same pass.

TimeToBuildBob:
Implemented and pushed this as `e391d2812`.

The new behavior keeps the conversation log append-only: `/model` appends a freshly generated, pinned prompt generation for the selected model and tool format; request preparation sends only the newest generated prompt while preserving ordinary conversation history and runtime system messages. `/tools load` now uses the same mechanism and persists the expanded tool set. No historical records are mutated and the conversation is not forked.

I also fixed Greptile's P1: indented multiline serialized tool calls now render as collapsible blocks, with a regression test covering adjacent calls and prose.

Local verification: 74 targeted tests pass; Ruff, formatting, mypy, and pre-commit pass. CI is running on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Fix the remaining Greptile nit, why aren't you?

TimeToBuildBob:
Fixed in `7a9dd3f5`. The issue was `\n(?!\s)` in the terminal lookahead — it required the next line to be non-whitespace, so a multiline call followed by indented prose was rejected and fell through to raw-JSON rendering.

Changed the lookahead to `\n` (any newline is a valid terminator). This is safe because gptme's tool format JSON-escapes string newlines as `\n` (two literal chars), so there are no actual newlines inside JSON string values that could prematurely end a match. Added a regression test for the indented-prose case.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Greptile passed ✅ (latest round, 16:35 UTC). All CI green. PR is ready to merge.

PR #3384, opened 2026-07-27
revert(logmanager): remove speculative S3 event-log replication

TimeToBuildBob:
## Summary

Revert #3382 before its speculative, off-by-default S3 replication surface becomes an API we need to maintain.

## Why

#3382 originated from Bob's internal idea backlog, not a reported user need or a named consumer. It added 913 lines and an S3/boto3/background-worker lifecycle to core LogManager for cross-machine durability that gptme has not established as a core responsibility.

Replicating only `events.jsonl` is also the wrong abstraction for general backup or sync: a useful conversation backup needs the whole conversation directory (`conversation.jsonl`, config, branches, views, files), while fleet telemetry should use an explicit telemetry/export mechanism rather than an internal recovery derivative.

This keeps Phase 1's local append-only recovery log from #2956 intact. It removes only the remote S3 replication added by #3382. If a concrete cross-machine session-sync consumer emerges, we should design from that requirement and evaluate whole-session backup/sync outside LogManager first.

## Verification

- `uv run pytest tests/test_eventlog.py -q` — 20 passed
- `uv run ruff check gptme/config/models.py gptme/config/user.py gptme/logmanager/manager.py` — passed
- Reverted tree matches the pre-#3382 versions of the three modified core files
- Full mypy currently reports pre-existing environment/baseline failures (`sphinx`/`textual` imports and unrelated existing errors); this revert introduces no new typed surface

Relates to #3382 and its post-merge design discussion.


TimeToBuildBob:
Fixed the branch-isolation regression in eec35f255. The revert now keeps #3382's independent local event-log paths (`branches/<branch>/events.jsonl`) while removing only remote replication. I moved the isolation test into `tests/test_eventlog.py`, where it belongs after `tests/test_replication.py` is deleted.

Verification: 21 event-log tests pass; targeted Ruff and pre-commit (including mypy) pass.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob What more features/things do you think should be removed or moved to contrib? (would we have rebuilt it if we started from scratch? is it being used or promising? does it add bload/deps/maintenance burden? how are we doing on our https://gptme.org/docs/arewetiny.html goals? etc)

Open an issue to research this across the codebase (removing for cleaner and easier to follow code, or pure refactors)

PR #2633, opened 2026-05-30
feat(server): bundle modern webui into package (gptme#2612 Part 2)

TimeToBuildBob:
## Summary

Completes Part 2 of #2612: ship the modern React webui inside the wheel so `pip install gptme && gptme-server serve` serves the modern UI without the `--webui-dir` flag.

- **Auto-detection**: `_resolve_static_folder` now checks `gptme/server/webui-dist/` (new bundle slot) and prefers it over the legacy `server/static/` when non-empty. Full precedence: `--webui-dir arg` > `GPTME_WEBUI_DIR` env > bundled `webui-dist/` > legacy static.
- **`make bundle-webui`**: copies `webui/dist/ → gptme/server/webui-dist/` for local dev/testing.
- **`pyproject.toml`**: includes `gptme/server/webui-dist/**/*` in the wheel.
- **`.gitignore`**: excludes the generated directory from git.
- **`release.yml`**: adds `setup-node@v4` + `npm ci && npm run build` + `make bundle-webui` before `poetry build`, so PyPI releases ship the modern UI automatically.
- **Tests**: 4 new tests in `test_server_webui_dir.py` covering bundled-webui preference, empty-dir fallback, and `--webui-dir`/env-var override precedence.

## Test plan

- [x] `uv run pytest tests/test_server_webui_dir.py -v` — 12/12 passed
- [ ] Verify CI release workflow builds the webui and includes it in the wheel
- [ ] `pip install gptme && gptme-server serve` should serve the modern webui without flags

## Notes

The `gptme/server/webui-dist/` directory is gitignored (it's a build artifact). It gets populated either by `make bundle-webui` (for local testing) or by the release CI workflow (for published wheels). Users on editable installs who want the modern UI without a flag can run `make bundle-webui` locally after `cd webui && npm run build`.

Closes #2612

PR #3387, opened 2026-07-27
fix(server): actually bundle modern webui in distributions

### LINKED ISSUE #3386: Bundle gptme-webui properly into gptme-server distribution
Right now the gptme-server on 5700 serves the old broken v0 webui, not the modern gptme-webui. We should get rid of the broken webui (or stop using it, might still be used for conversation HTML exports etc?) and actually ship the built gptme-webui with the gptme package/standard dist.

This would simplify setup of gptme-server considerably (no more need to run webui separately). Would also mean updating the docs to the new simplified experience, no longer need the `--cors-origin` trick (unless running a `npm run dev` version of gptme-webui), etc.

It already is in gptme-tauri, after all.

TimeToBuildBob:
## Problem

The modern-webui bundling path from #2633 ran successfully in release CI, but Poetry excluded the generated `gptme/server/webui-dist/` because the directory is gitignored. The released v0.32.1 wheel contains **zero** modern-webui files, so `gptme-server` silently falls back to the legacy UI.

The release pipeline only checked that the frontend build and `poetry build` exited successfully; it never inspected the resulting package.

## What changed

- explicitly include the generated `webui-dist` directory in wheel and sdist formats;
- validate both release archives contain `index.html` and at least one hashed asset before publishing;
- run that package assertion in WebUI CI too, including when packaging config changes;
- build/bundle the UI for PyInstaller and Docker distribution paths;
- make the self-host Docker image build the frontend in a dedicated Node stage;
- document `gptme-server` as a same-origin one-process setup, keeping `--cors-origin` for separate frontend origins;
- keep the legacy static bundle as a fallback for now because `computer.html` and its tests still depend on it.

## Evidence

- v0.32.1 release wheel: 1,568,388 bytes, 0 `webui-dist` entries, 4 legacy-static entries.
- Release run 29582787488 shows `npm run build` and rsync succeeded before Poetry silently omitted the directory.
- Fixed local wheel: 43 modern UI files (2,585,262 uncompressed bytes) plus the legacy fallback.
- Installed-wheel smoke test resolves `site-packages/gptme/server/webui-dist` and `/` returns the React root.
- Fresh self-host Docker build contains 64 hashed assets and `/` returns the modern UI.

## Verification

- `pytest -q tests/test_validate_release_package.py tests/test_server_webui_dir.py` — 17 passed in an isolated venv
- `mypy scripts/validate_release_package.py tests/test_validate_release_package.py` — clean
- scoped pre-commit hooks — all passed
- `poetry build && make validate-release-package` — wheel + sdist passed
- fresh `docker build -f scripts/Dockerfile.selfhost ...` + live root smoke — passed
- PR quality gate — 100/100

Closes #3386

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -43,11 +43,12 @@ demos
 projects
 eval_results*
 
-# Frontend / desktop / static-site sources — not needed by either image.
-# The agent image (scripts/Dockerfile) copies only pyproject/gptme explicitly;
-# the self-host image (scripts/Dockerfile.selfhost) only pip-installs the
-# package. tauri (~8G) and webui (~600M) would otherwise bloat the build context.
-webui
+# Desktop/static-site sources and generated frontend files are not needed. The
+# Dockerfiles build the modern UI from source in a dedicated Node stage.
+webui/node_modules
+webui/dist
+webui/playwright-report
+webui/test-results
 tauri
 site
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -217,6 +217,19 @@ jobs:
         make build
         poetry install --extras server --extras pyinstaller --with dev
 
+    - name: Set up Node.js
+      uses: actions/setup-node@v7
+      with:
+        node-version: '20'
+        cache: 'npm'
+        cache-dependency-path: webui/package-lock.json
+
+    - name: Build and bundle webui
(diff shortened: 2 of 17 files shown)


TimeToBuildBob:
@greptileai review

PR #3399, opened 2026-07-29
fix(config): use explicit UTF-8 encoding for all TOML config I/O

LHMQ878:
## Summary

TOML is [defined to be UTF-8](https://toml.io/en/v1.0.0#spec), but no `open()` in `gptme/config/` passed `encoding=`. Text mode without it uses the platform's *preferred* encoding — a legacy codepage on a stock Windows install (cp936 on a Chinese system, cp1252 on a Western one), not UTF-8. So `config.toml`, `config.local.toml`, `gptme.toml` and per-conversation `config.toml` were all read and written in whatever codepage the machine happened to have.

This hits exactly the fields a user writes in their own language: `[user] about`, `response_preference`, `name`, and a project's `prompt` / `base_prompt`.

## Three failure modes

**1. Read — gptme will not start.** The bytes already on disk are enough; the user does not have to do anything after writing them:

```
$ cat ~/.config/gptme/config.toml
[user]
about = "我是一名开发者"

>>> load_user_config()
UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 24
```

**2. Write — the value is rejected.** `gptme config set user.about …`, the setup wizard, and the web UI's config PUT/PATCH all reach `set_config_value`:

```
>>> set_config_value("user.about", "café ☕")
UnicodeEncodeError: 'gbk' codec can't encode character '\xef' in position 41
```

**3. Round trip — the write succeeds and silently corrupts the file.** This is the worst one. For text the codepage *does* cover, nothing raises; the file is written in that codepage and is not valid UTF-8, therefore not valid TOML for any other reader:

```
>>> set_config_value("user.about", "开发者")     # no error
>>> open(config_path, "rb").read()
b'[user]\r\nabout = "\xbf\xaa\xb7\xa2\xd5\xdf"\r\n'
>>> _.decode("utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbf in position 20
```

`tomllib` — which gptme itself uses for chat configs on Python 3.11+ — is UTF-8 only, so a file written this way cannot be read back by the very loader that will read it. Nor by any editor, any other TOML library, or the same config synced to a Linux machine.

## Changes

**Writes** are now strict UTF-8. **Reads** go through one helper, `_read_config_text`, which decodes as UTF-8 per the spec and only on `UnicodeDecodeError` retries with the locale codec — so a config an older gptme wrote in a code page still loads, with a warning naming the file, and the next config change normalises it to UTF-8. Making the reads strict alone would have taken a Windows user with a cp936 `config.toml` from "loads with mojibake" to "does not load at all"; thanks @greptile-apps for catching that.

While adding that, one more defect surfaced: `tomllib.load` raises `UnicodeDecodeError`, not `TOMLDecodeError`, and that was not in `_CHAT_CONFIG_LOAD_ERRORS`. So on `master` a legacy-encoded conversation config **already** escapes `ChatConfig.from_logdir`'s handler on Python 3.11+ instead of degrading to defaults. Added to both branches of the tuple.

Call sites, across three files:

| file | reads (via `_read_config_text`) | writes (`encoding="utf-8"`) |
|---|---|---|
| `gptme/config/user.py` | `get_user_config_env_source`, `get_default_model_source`, `load_user_config`, `_load_config_doc` | `_strip_unknown_config_keys`, `_load_config_doc` (create), `set_config_value`, `save_provider_config` |
| `gptme/config/chat.py` | `from_logdir` (both the `tomllib` and `tomlkit` branches), `save` read-back | `save`'s atomic write |
| `gptme/config/project.py` | `gptme.toml`, `gptme.local.toml` | — |

`ChatConfig.save()` needed the codec on `tempfile.NamedTemporaryFile(mode="w")` rather than on an `open()` — same defect, different constructor, and easy to miss since the atomic-write block reads as if it were already handled.

No behaviour change other than the encoding. Nothing else in these files was touched.

## Internal consistency

`ChatConfig.from_logdir` already got this right on one of its two branches and wrong on the other:

```python
if tomllib is not None:
    with open(chat_config_path, "rb") as f:   # binary — tomllib decodes as UTF-8, correct
        config_data = tomllib.load(f)
else:
    with open(chat_config_path) as f:         # text — platform codepage
        config_data = tomlkit.load(f).unwrap()
```

Two paths in one function, reading the same file, disagreeing about its encoding — which also means the bug's visibility depended on the Python version.

The wider codebase already treats `encoding="utf-8"` as the norm: `util/export.py`, `prompt_queue.py`, `tools/morph.py`, `tools/save.py`, `tools/patch_many.py`, `lessons/installer.py`, `logmanager/eventlog.py`, `util/context_savings.py`.

And this exact class of bug was already fixed once upstream: **#2051**, "fix(tools): add explicit UTF-8 encoding to all file read/write operations" (merged). It covered `gptme/tools/{morph,patch,read,save}.py` and left `gptme/config/` untouched. Its own test plan still has "Verify non-ASCII file editing works on a Windows system" unchecked — that is where I found this.

## Validation

New `tests/test_config_encoding.py`, 11 tests:

| test | guards |
|---|---|
| `test_load_user_config_reads_non_ascii_about` | read side, main config |
| `test_load_user_config_reads_non_ascii_from_local_override` | read side, `config.local.toml` merge path (a second `open()`) |
| `test_set_config_value_writes_valid_utf8_toml` | write side; asserts the raw bytes decode as **strict** UTF-8 |
| `test_set_config_value_ascii_still_works` | control — the case the old code got right |
| `test_user_config_round_trips_non_ascii` | write then read back; the two sides must agree |
| `test_project_config_reads_non_ascii` | `gptme.toml`, which is committed and shared across machines |
| `test_chat_config_save_writes_valid_utf8` | `ChatConfig.save()`'s atomic write, then `from_logdir` reads it back |
| `test_legacy_encoded_user_config_is_still_readable` | upgrade path — a cp936 `config.toml` from an older install |
| `test_legacy_encoded_chat_config_is_still_readable` | upgrade path — a cp936 conversation config |
| `test_legacy_encoded_config_is_rewritten_as_utf8` | the fallback is read-side only; the next write normalises the file, preserving the non-ASCII value |
| `test_undecodable_config_fails_at_the_parser_not_the_decoder` | truly corrupt bytes must reach the parser (which callers handle) rather than raise on decode |

Reproducing a locale bug on a UTF-8 CI runner needs care. Monkeypatching `locale.getpreferredencoding` **does not work** — CPython reads the locale encoding at the C level, so `open()` ignores the patched function. A `legacy_default_encoding` context manager instead replaces `open` with a shim that supplies a codec in exactly the position CPython would supply the locale's — **only when the caller passed no `encoding`**:

```python
def shim(file, mode="r", *args, **kwargs):
    if "b" not in mode and kwargs.get("encoding") is None and len(args) < 2:
        kwargs["encoding"] = codec
    return real_open(file, mode, *args, **kwargs)
```

Code that names its encoding is untouched by the shim, so these tests fail on a machine of any locale when the `encoding=` is missing and pass on a machine of any locale when it is present. Read-side tests use `ascii` rather than `cp1252`: cp1252 maps every byte and round-trips, so it turns undecodable input into mojibake without raising and would not detect a missing `encoding=` on a read.

**Control experiments.** Two, because the change has two halves.

*(a) against unmodified `master`* — the three config files reverted, the tests kept:

```
9 failed, 2 passed

FAILED test_load_user_config_reads_non_ascii_about
       UnicodeDecodeError: 'ascii' codec can't decode byte 0xe6 in position 17
FAILED test_load_user_config_reads_non_ascii_from_local_override
FAILED test_set_config_value_writes_valid_utf8_toml
       UnicodeEncodeError: 'charmap' codec can't encode characters in position 17-23
FAILED test_user_config_round_trips_non_ascii
FAILED test_project_config_reads_non_ascii
FAILED test_chat_config_save_writes_valid_utf8
       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xce in position 16: invalid continuation byte
FAILED test_legacy_encoded_chat_config_is_still_readable
FAILED test_legacy_encoded_config_is_rewritten_as_utf8
FAILED test_undecodable_config_fails_at_the_parser_not_the_decoder
```

`test_chat_config_save_writes_valid_utf8` is failure mode 3 caught in the act: the save raised nothing, and the file it produced does not decode as UTF-8. The two that stay green are the ASCII control and `test_legacy_encoded_user_config_is_still_readable` — the latter *should* pass on `master`, since reading with the locale codec is exactly what `master` does; it is there to stop this PR from breaking that.

*(b) against the strict-UTF-8-only commit* (this branch's first commit, before the fallback):

```
4 failed, 7 passed

FAILED test_legacy_encoded_user_config_is_still_readable
FAILED test_legacy_encoded_chat_config_is_still_readable
FAILED test_legacy_encoded_config_is_rewritten_as_utf8
FAILED test_undecodable_config_fails_at_the_parser_not_the_decoder
       UnicodeDecodeError: 'utf-8' codec can't decode byte 0xce in position 16
```

Exactly the four upgrade-path tests, which is what makes them a guard on the fallback rather than on the encoding. With both commits: **11 passed**.

**No regressions.** `tests/test_config.py tests/test_chat_config.py tests/test_computer_docker_config.py tests/test_server_config_workspace_containment.py`:

| tree | result |
|---|---|
| this branch | 77 passed, 43 failed, 12 errors |
| unmodified `master` | 77 passed, 43 failed, 12 errors |

Identical, with an empty symmetric difference of the failing-test sets in both directions. I ran this on Windows, where a portion of the config suite cannot pass at all on `master`: symlink creation needs a privilege the default account lacks (`OSError: [WinError 1314]`, 7 tests), tmpdir teardown hits file locking (`PermissionError: [WinError 32]`, 5 tests), and one asserts a POSIX path separator (`assert '~\\workspace' == '~/workspace'`). Same machine and same selection on both sides, so the comparison is valid for "this change adds no failures" — but a maintainer run on Linux is a better check of the absolute numbers.

`ruff format`, `ruff check` and `mypy` all clean on the four changed files.

## Linked issues

I searched open and closed PRs and issues before opening this (`encoding`, `UnicodeDecodeError`, `UnicodeEncodeError`, `utf-8`, `config.toml windows`) — no existing or overlapping work; #2051 is the closest and is merged and disjoint. Targeting `master` per `AGENTS.md`.

### DIFF
--- a/gptme/config/chat.py
+++ b/gptme/config/chat.py
@@ -245,7 +245,14 @@ class ChatConfig:
             return cls()
 
     def save(self) -> Self:
-        """Save the chat config to the log directory."""
+        """Save the chat config to the log directory.
+
+        Raises:
+            OSError: If this config was decoded by a guessed codec and its
+                original bytes could not be backed up. The save is abandoned so
+                the unrecoverable bytes stay on disk; only a config that already
+                went through the legacy fallback can reach this.
+        """
         if not self._logdir:
             raise ValueError("ChatConfig has no logdir set")
         self._logdir.mkdir(parents=True, exist_ok=True)
--- a/gptme/config/chat.py
+++ b/gptme/config/chat.py
@@ -234,7 +234,9 @@ class ChatConfig:
                 with open(chat_config_path, "rb") as f:
                     config_data = tomllib.load(f)
             else:
-                with open(chat_config_path) as f:
+                # `tomllib` above opens in binary mode and decodes as UTF-8 itself, as
+                # the TOML spec requires; this fallback must say so explicitly.
+                with open(chat_config_path, encoding="utf-8") as f:
                     config_data = tomlkit.load(f).unwrap()
             config_data["_logdir"] = path
             return cls.from_dict(config_data)
@@ -254,7 +256,7 @@ class ChatConfig:
         # Load existing config as TOMLDocument to preserve formatting/comments
         if chat_config_path.exists():
             try:
-                with open(chat_config_path) as f:
+                with open(chat_config_path, encoding="utf-8") as f:
                     doc = tomlkit.load(f)
                 # Update document in-place, preserving formatting
                 for key, value in config_dict.items():
@@ -289,6 +291,7 @@ class ChatConfig:
         # (e.g., daemon thread killed on exit while saving)
         with tempfile.NamedTemporaryFile(
             mode="w",
+            encoding="utf-8",
             dir=self._logdir,
             suffix=".toml.tmp",
             delete=False,
--- a/gptme/config/chat.py
+++ b/gptme/config/chat.py
@@ -19,15 +19,16 @@ from typing_extensions import Self
 if sys.version_info >= (3, 11):
     import tomllib
 
-    _CHAT_CONFIG_LOAD_ERRORS = (OSError, tomllib.TOMLDecodeError)
+    _CHAT_CONFIG_LOAD_ERRORS = (OSError, UnicodeDecodeError, tomllib.TOMLDecodeError)
 else:
     tomllib = None
 
-    _CHAT_CONFIG_LOAD_ERRORS = (OSError, TOMLKitError)
+    _CHAT_CONFIG_LOAD_ERRORS = (OSError, UnicodeDecodeError, TOMLKitError)
 
 from ..util import path_with_tilde
 from .models import AgentConfig, MCPConfig
 from .project import get_project_config
+from .user import _read_config_text
 
 if TYPE_CHECKING:
     from ..tools.base import ToolFormat
@@ -230,14 +231,13 @@ class ChatConfig:
                 )
             return cls(_logdir=path, workspace=workspace.resolve())
         try:
+            # `_read_config_text` decodes as UTF-8 per the TOML spec, with a
+            # fallback for configs an older gptme wrote in the platform encoding.
+            text = _read_config_text(chat_config_path)
             if tomllib is not None:
-                with open(chat_config_path, "rb") as f:
-                    config_data = tomllib.load(f)
+                config_data = tomllib.loads(text)
             else:
-                # `tomllib` above opens in binary mode and decodes as UTF-8 itself, as
-                # the TOML spec requires; this fallback must say so explicitly.
-                with open(chat_config_path, encoding="utf-8") as f:
-                    config_data = tomlkit.load(f).unwrap()
+                config_data = tomlkit.loads(text).unwrap()
             config_data["_logdir"] = path
             return cls.from_dict(config_data)
         except _CHAT_CONFIG_LOAD_ERRORS as e:
@@ -256,8 +256,7 @@ class ChatConfig:
         # Load existing config as TOMLDocument to preserve formatting/comments
         if chat_config_path.exists():
             try:
-                with open(chat_config_path, encoding="utf-8") as f:
-                    doc = tomlkit.load(f)
+                doc = tomlkit.loads(_read_config_text(chat_config_path))
                 # Update document in-place, preserving formatting
                 for key, value in config_dict.items():
                     if isinstance(value, dict) and key in doc:
--- a/gptme/config/chat.py
+++ b/gptme/config/chat.py
@@ -28,7 +28,7 @@ else:
 from ..util import path_with_tilde
 from .models import AgentConfig, MCPConfig
 from .project import get_project_config
-from .user import _read_config_text
+from .user import _back_up_before_reencoding, _read_config_text
 
 if TYPE_CHECKING:
     from ..tools.base import ToolFormat
@@ -284,6 +284,10 @@ class ChatConfig:
                 doc = tomlkit.parse(tomlkit.dumps(config_dict))
         else:
             doc = tomlkit.parse(tomlkit.dumps(config_dict))
+
+        # If this config was read through a codec that accepts any bytes, its text
+        # is a guess; keep the original bytes before the rename discards them.
+        _back_up_before_reencoding(chat_config_path)
 
         # Use atomic write: write to temp file, then rename
         # This prevents corruption if process is interrupted during write
--- a/gptme/config/project.py
+++ b/gptme/config/project.py
@@ -12,7 +12,7 @@ import tomlkit
 
 from ..util import path_with_tilde
 from .models import ProjectConfig
-from .user import _merge_config_data
+from .user import _merge_config_data, _read_config_text
 
 logger = logging.getLogger(__name__)
 
@@ -101,16 +101,16 @@ def _get_project_config_cached(
     if project_config_paths:
         project_config_path = project_config_paths[0]
         # load project config
-        with open(project_config_path, encoding="utf-8") as f:
-            config_data = tomlkit.load(f).unwrap()
+        config_data = tomlkit.loads(_read_config_text(project_config_path)).unwrap()
 
         # Look for local config file in the same directory
         local_config_path = project_config_path.parent / "gptme.local.toml"
         used_local_config_path: Path | None = None
         if local_config_path.exists():
             used_local_config_path = local_config_path
-            with open(local_config_path, encoding="utf-8") as f:
-                local_config_data = tomlkit.load(f).unwrap()
+            local_config_data = tomlkit.loads(
+                _read_config_text(local_config_path)
+            ).unwrap()
 
             # Merge local config into main config
             config_data = _merge_config_data(config_data, local_config_data)
--- a/gptme/config/project.py
+++ b/gptme/config/project.py
@@ -101,7 +101,7 @@ def _get_project_config_cached(
     if project_config_paths:
         project_config_path = project_config_paths[0]
         # load project config
-        with open(project_config_path) as f:
+        with open(project_config_path, encoding="utf-8") as f:
             config_data = tomlkit.load(f).unwrap()
 
         # Look for local config file in the same directory
@@ -109,7 +109,7 @@ def _get_project_config_cached(
         used_local_config_path: Path | None = None
         if local_config_path.exists():
             used_local_config_path = local_config_path
-            with open(local_config_path) as f:
+            with open(local_config_path, encoding="utf-8") as f:
                 local_config_data = tomlkit.load(f).unwrap()
 
             # Merge local config into main config
--- a/gptme/config/user.py
+++ b/gptme/config/user.py
@@ -39,6 +39,46 @@ from .models import (
 logger = logging.getLogger(__name__)
 
 
+def _legacy_codec_candidates() -> list[str]:
+    """Codecs to try after UTF-8 when reading a config, most likely first.
+
+    `locale.getpreferredencoding(False)` is not sufficient on its own. Under
+    Python's UTF-8 mode (`-X utf8`, `PYTHONUTF8=1`, or a launcher that enables it)
+    it returns "utf-8" whatever the platform code page is -- so on exactly the
+    Windows install whose code page wrote the file, the codec needed to read it
+    back is the one this function would hide. `locale.getencoding()` (3.11+)
+    reports the real code page regardless of UTF-8 mode; on 3.10 the same value is
+    only reachable through `getdefaultlocale`, which is not yet deprecated there.
+
+    Names are deduplicated by their canonical codec name, and "utf-8" is dropped
+    since the caller has already tried it.
+    """
+    candidates: list[str] = []
+    getencoding = getattr(locale, "getencoding", None)
+    if getencoding is not None:
+        candidates.append(getencoding())
+    else:  # Python 3.10
+        try:
+            candidates.append(locale.getdefaultlocale()[1] or "")
+        except ValueError:
+            pass
+    candidates.append(locale.getpreferredencoding(False))
+
+    result: list[str] = []
+    seen = {"utf-8"}
+    for name in candidates:
+        if not name:
+            continue
+        try:
+            canonical = codecs.lookup(name).name
+        except LookupError:
+            continue
+        if canonical not in seen:
+            seen.add(canonical)
+            result.append(name)
+    return result
+
+
 def _read_config_text(path: str | Path) -> str:
     """Read a TOML config file as text.
 
@@ -49,28 +89,37 @@ def _read_config_text(path: str | Path) -> str:
     instead of making gptme fail to start; the next write re-encodes it as UTF-8,
     since the writers are now explicit.
 
+    Each candidate is decoded strictly, so a codec that cannot represent the bytes
+    hands over to the next one instead of quietly producing mojibake.
+
     Never raises `UnicodeDecodeError`: callers wrap this in a TOML parse and handle
-    parse errors (`ChatConfig.from_logdir` degrades to defaults), so a corrupt file
-    is decoded with `errors="replace"` and left for the parser to reject.
+    parse errors (`ChatConfig.from_logdir` degrades to defaults), so a file no
+    candidate can decode is read with `errors="replace"` and left for the parser to
+    reject.
     """
     data = Path(path).read_bytes()
     try:
         return data.decode("utf-8")
     except UnicodeDecodeError:
-        fallback = locale.getpreferredencoding(False)
-        if codecs.lookup(fallback).name == "utf-8":
-            # No second codec to try: the locale's is the one that just failed.
-            logger.warning(
-                f"Config file {path} is not valid UTF-8; reading it with "
-                "replacement characters. Some values may be garbled."
-            )
-            return data.decode("utf-8", errors="replace")
+        pass
+
+    for codec in _legacy_codec_candidates():
+        try:
+            text = data.decode(codec)
+        except UnicodeDecodeError:
+            continue
         logger.warning(
-            f"Config file {path} is not valid UTF-8; reading it as {fallback} "
+            f"Config file {path} is not valid UTF-8; reading it as {codec} "
             "(written by an older gptme). It will be rewritten as UTF-8 on the "
             "next config change."
         )
-        return data.decode(fallback, errors="replace")
+        return text
+
+    logger.warning(
+        f"Config file {path} is not valid UTF-8 and no legacy codec could decode "
+        "it; reading it with replacement characters. Some values may be garbled."
+    )
+    return data.decode("utf-8", errors="replace")
 
 
 # Define the path to the config file
--- a/gptme/config/user.py
+++ b/gptme/config/user.py
@@ -4,7 +4,9 @@ Handles loading, merging, and persisting user-level configuration
 from ~/.config/gptme/config.toml and config.local.toml.
 """
 
+import codecs
 import copy
+import locale
 import logging
 import os
 from collections.abc import MutableMapping
@@ -35,6 +37,32 @@ from .models import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _read_config_text(path: str | Path) -> str:
+    """Read a TOML config file as text.
+
+    TOML is defined to be UTF-8, so that is what we decode as. But gptme wrote
+    these files without naming an encoding until recently, so one left behind by
+    an older install may be in the platform's preferred encoding -- a legacy code
+    page on Windows. Falling back to that codec keeps such a config readable
+    instead of making gptme fail to start; the next write re-encodes it as UTF-8,
+    since the writers are now explicit.
+    """
+    data = Path(path).read_bytes()
+    try:
+        return data.decode("utf-8")
+    except UnicodeDecodeError:
+        fallback = locale.getpreferredencoding(False)
+        if codecs.lookup(fallback).name == "utf-8":
+            # Nothing else to try: the locale codec is the one that just failed.
+            raise
+        logger.warning(
+            f"Config file {path} is not valid UTF-8; reading it as {fallback} "
+            "(written by an older gptme). It will be rewritten as UTF-8 on the "
+            "next config change."
+        )
+        return data.decode(fallback, errors="replace")
 
 
 # Define the path to the config file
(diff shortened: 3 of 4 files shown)


LHMQ878:
Good catch on the upgrade path — that was a real hole, and fixing it turned up a second bug.

Pushed bbeeebf:

- **Reads** now go through one helper, `_read_config_text`: decode as UTF-8 per the TOML spec, and only on `UnicodeDecodeError` retry with `locale.getpreferredencoding(False)`, logging a warning that names the file. **Writes** stay strict UTF-8, so the next config change normalises the file. A user with a cp936 `config.toml` from an older install keeps loading instead of hitting a hard failure.
- The bug that surfaced while doing it: `tomllib.load` raises `UnicodeDecodeError`, not `TOMLDecodeError`, and that was not in `_CHAT_CONFIG_LOAD_ERRORS`. So on `master` a legacy-encoded conversation config **already** escapes `ChatConfig.from_logdir`'s handler on Python 3.11+ rather than degrading to defaults. Added to both branches of the tuple.
- Four tests for the upgrade path, including the one you asked for (`test_legacy_encoded_user_config_is_still_readable`) plus a rewrite-normalises-it test and one asserting that truly corrupt bytes reach the *parser* — which callers already handle — instead of raising on decode.

Worth noting `test_legacy_encoded_user_config_is_still_readable` passes on `master` and failed on this branch's first commit; that is the point of it. Both control runs are in the updated description.

ErikBjare:
@greptileai review

LHMQ878:
Both P1s were real. The second one I only half-fixed in `bbeeebf`/`b175de7` — thanks for keeping it open.

Pushed `6082c5c`.

**The remaining hole was exactly as described.** `locale.getpreferredencoding(False)` is not a reliable way to learn the platform code page, because UTF-8 mode overrides it. Measured on a stock cp936 Windows install:

```
             getpreferredencoding(False)   getencoding()
normal                 cp936                  cp936
-X utf8                utf-8                  cp936     <-- the codec that wrote the file
```

So on the machine whose code page produced the legacy config, UTF-8 mode hides the one codec that can read it back. `b175de7` stopped the re-raise but replaced it with `errors="replace"`, which is not much better: `load_user_config` then raises on the mojibake TOML, and `ChatConfig.from_logdir` degrades to defaults and silently drops the user's saved settings.

**Fix:** `_legacy_codec_candidates()` consults `locale.getencoding()` first — it reports the real code page regardless of UTF-8 mode — then `getpreferredencoding` as before. On 3.10, where `getencoding` does not exist yet, the same value comes from `getdefaultlocale` (not deprecated on that version). Candidates are deduplicated by canonical codec name with `utf-8` dropped, since it has already been tried; that is what stops `"UTF-8"`/`"utf8"` from being retried as if they were a second codec.

**Second change, which the first one made necessary:** candidates are now decoded **strictly**. Previously the single fallback used `errors="replace"`, so with a list of candidates the first one would always "succeed" and every later candidate would be dead code — a codec that cannot represent the bytes must hand over, not win by producing mojibake. Only when every candidate fails does the read fall back to replacement characters, which preserves the guarantee that no `UnicodeDecodeError` reaches callers (they handle TOML parse errors, not decode errors).

### Verification

- End-to-end under **real** UTF-8 mode, no mocks anywhere — `python -X utf8` in a fresh process, cp936 config on disk, `load_user_config` returns `'我是一名开发者'` and logs `reading it as cp936`.
- Both 3.10 branches exercised with a stubbed `locale` namespace lacking `getencoding`: the `getdefaultlocale` path resolves cp936 under UTF-8 mode, and a `getdefaultlocale` that raises `ValueError` degrades to the `getpreferredencoding` candidate instead of propagating.
- `tests/test_config_encoding.py`: **18 passed**. With `gptme/config/user.py` reverted to `b175de7`, **4 fail** — the two upgrade-path tests in their new `utf8-mode` parametrisation, plus the two new tests.
- `-k config` across `tests/`: **229 passed / 46 failed** vs baseline **225 / 46** — exactly +4, no new failures. (Those 46 and the 19 collection errors are pre-existing here, missing optional extras like `flask_compress`.)
- `ruff check` clean, `ruff format --check` clean, `mypy` clean on the changed source.

### Tests added

- `test_legacy_encoded_user_config_is_still_readable` and `test_legacy_encoded_chat_config_is_still_readable` are now parametrised `[locale, utf8-mode]`. The `legacy_locale_codec` helper takes `utf8_mode=True`, which reproduces the real asymmetry: `getpreferredencoding` reports `utf-8` while `getencoding` still reports the code page. That is the case a `getpreferredencoding`-only fallback cannot serve, and the parametrisation is what pins it.
- `test_legacy_codec_candidates_excludes_utf8_and_deduplicates` — a `utf-8` entry is a wasted retry of bytes that already failed, and the two locale sources routinely spell one codec two ways.
- `test_a_codec_that_cannot_decode_hands_over_to_the_next` — asserts no replacement characters in the result, which is what keeps the strict-decode behaviour from regressing to the mojibake-wins version.

On your first P1 (`80d5dd0`): that was the missing upgrade path, fixed in `bbeeebf`. Worth noting it turned up a separate pre-existing bug — `tomllib.load` raises `UnicodeDecodeError`, not `TOMLDecodeError`, and that was absent from `_CHAT_CONFIG_LOAD_ERRORS`, so on `master` a legacy-encoded conversation config already escapes `ChatConfig.from_logdir`'s handler on 3.11+ rather than degrading to defaults. Also fixed there.

ErikBjare:
@greptileai review

LHMQ878:
Since you've re-triggered the review — one thing worth knowing before you read it: **the ❌ codecov comment above is stale, and no CI has ever run against the current head.**

All five workflow runs at `6082c5c` sit at `action_required`:

```
Test:      completed/action_required
Lint:      completed/action_required
Benchmark: completed/action_required
Docs:      completed/action_required
Build:     completed/action_required
```

They need a maintainer to approve workflows for this fork, so the only check that has ever executed is the codecov upload from `bbeeebf` — two commits back. Those 4 failures were real at `bbeeebf` and are what `b175de7` and `6082c5c` fixed.

### The failures were mine, and they were a real bug in the tests

`bbeeebf`'s upgrade-path tests set the legacy codec with the `open()` shim only. But `_read_config_text` reads *bytes* and then asks the `locale` module what to fall back to, so the shim never reached it — the fallback was only exercised on a machine whose locale already happened to be a legacy code page. On CI it wasn't, so the fallback never engaged and the raw `UnicodeDecodeError` escaped, exactly as codecov shows:

```
gptme/config/user.py:54: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xce in position 16
```

`b175de7` added a second helper, `legacy_locale_codec`, that patches `locale.getpreferredencoding` / `getencoding` / `getdefaultlocale` directly, so the upgrade-path tests are locale-independent.

### Verified rather than assumed

I couldn't just run them here — this box is cp936, which is the locale that masked the bug in the first place. So I forced both locale functions to UTF-8 before pytest imports anything, reproducing what CI actually sees, and ran the identical probe at both commits:

```
bbeeebf (what codecov measured)  ->  4 failed, 7 passed
6082c5c (current head)           ->  18 passed
```

The 4 failures reproduce with the same `user.py:54` stack codecov reported, which is what makes me confident the probe models CI rather than something else. `PYTHONUTF8=1` alone also passes 18/18 — that's the narrower case greptile's second P1 was about, where `getpreferredencoding(False)` reports utf-8 while `getencoding()` still reports the real code page.

`tests/test_config.py` shows 16 failures on my branch, and the same 16 on `master` at `d83a093` — pre-existing, unrelated to this PR. My branch is 83 passed there vs 65 on master, the difference being the 18 new tests. (Collecting all of `tests/` needs `flask_compress`, which isn't installed here.)

No action needed from me on this — just flagging that the red ❌ is measuring code that no longer exists, in case it colours the review.

LHMQ878:
The third P1 was correct, and worse than stated. Fixed in `f22324a`.

### Confirming it, including the part that makes it destructive

The finding was that a cp936 config read on a cp1252 locale decodes without error into mojibake. It does. What I wanted to know was whether anything downstream would catch it, and nothing does:

```
cp1252 -> 'ÎÒÊÇÒ»Ãû¿ª·¢Õß'   (no error)
tomlkit parses it fine: about = 'ÎÒÊÇÒ»Ãû¿ª·¢Õß'
```

The parser is not a backstop, because mojibake is valid TOML. And the loss is only permanent at the *write*, not the read — `mojibake.encode("cp1252") == original_bytes` is `True`, so the bytes still say what they always said; they were just read through the wrong table. Re-encoding as UTF-8 is what throws the original away.

Measured on a temp config seeded with cp936 bytes, locale forced to cp1252, before and after the fix:

```
before: after save, other files in the config dir: NONE
        can the original text be recovered from disk? False
after:  after save, other files in the config dir: ['config.toml.orig']
        can the original text be recovered from disk? True
```

### What I tried first, and why I threw it away

My first attempt was a classifier: trust the decode when the codec *could* have rejected the bytes, don't when it can't. That turned out to be wrong twice over, so I'm recording both because they're the reason the fix looks the way it does.

**Rejecting some bytes proves nothing.** cp1252 leaves exactly five bytes undefined — `0x81 0x8d 0x8f 0x90 0x9d` — and none of them occur in the cp936 Chinese text it mangles. So "can this codec fail?" answers yes for precisely the codec in the finding.

**And multi-byte code pages aren't safe either**, which is what killed the idea. I'd assumed their sequence structure made a clean decode meaningful. Probing it says otherwise — the same cp936 bytes decode without error, into the wrong text, as:

```
big5, cp949, euc-kr, cp950, euc-jp, cp932   <- all decode, all wrong
```

So there is no cheap test that separates the safe fallbacks from the damaging ones, because there aren't any safe ones.

### The fix

Every legacy fallback is treated as a guess. The read still returns text — failing to start is worse than showing garbled values, which is the whole point of the fallback — but:

- the file is recorded as decoded-by-guess, and the warning now says the values may be garbled and to check them before saving;
- before anything rewrites it as UTF-8, the original bytes are copied to `<path>.orig`, once (a second save would otherwise overwrite the original with the already-garbled version);
- UTF-8 configs are untouched: no fallback, no backup file, no `.orig` littered next to every config.

Wired into all four rewrite sites — the two `set_config_value` paths, `_strip_unknown_keys`, and `ChatConfig.save`'s atomic write in `chat.py`, where the `.replace()` would otherwise discard the bytes. Deliberately *not* wired into `_load_config_doc`'s create-a-new-file branch, since there are no prior bytes there to preserve.

### Tests

`tests/test_config_encoding.py` is 24 passed under a locale forced to UTF-8 (this box is cp936, so I have to force it to model CI). Four new tests, one of which asserts the premise rather than assuming it: `test_a_successful_decode_does_not_identify_the_codec` pins down that both single-byte *and* multi-byte codecs decode these bytes wrongly, and that tomlkit accepts the result — so if someone later reintroduces a classifier, that test is what should stop them.

Reverting the two source files leaves 5 of them failing.

Still worth knowing: no CI has run against any head of this PR — all workflow runs sit at `action_required` pending fork approval — so `f22324a` is verified locally only (ruff, ruff format, mypy clean).

ErikBjare:
@greptileai review

LHMQ878:
The fourth P1 was correct, and reproducing it turned up two more holes in the same function. Fixed in `7e3386e`.

### Confirming it

Injected an `OSError` on the `.orig` write (a full disk or read-only config dir), cp936 config on a cp1252 locale:

```
backup exists:                 False
original bytes still on disk:  False
recoverable:                   NO
marker still set:              False      <-- see below
```

So the finding understates it slightly: this is the *worst* case, not just a lossy one. The values end up garbled **and** the bytes that could have undone that are gone. Meanwhile refusing to save is recoverable — the user is told why and can copy the file by hand. That asymmetry is the whole argument, so the `OSError` now propagates and nothing is written.

### The two holes reproducing it exposed

**The marker was cleared in a `finally`.** `_encoding_unverified` is what records "these bytes have not been copied yet". Discarding it on failure meant the *next* save saw a file needing no backup and rewrote it — so a single transient disk error didn't just skip one backup, it lost the original at the following save instead. The marker is now retained on failure, and a later save retries. Verified: after the injected failure clears, the next `set_config_value` writes a `.orig` byte-identical to the original.

**A partial `.orig` would have read as success.** `backup.exists()` is the "already preserved" test, so a copy that died part way through leaves a truncated file that the next call trusts — as destructive as no backup, while looking like it worked. The copy is now staged as `.orig.tmp` and renamed, so a failure leaves nothing behind.

That second one is also why my first probe of the fix reported a false pass: it injected the failure on `.orig` only, and the write had moved to `.orig.tmp`. Worth mentioning because the test helper now fails on any path containing `.orig` for exactly that reason.

### One place that still swallows it, deliberately

`_strip_unknown_config_keys` keeps its `except OSError`. That write only deletes keys already ignored in memory, so not doing it costs a repeated warning and nothing else. The comment there says so explicitly, since the same exception aborting a real save is the point of the change.

### Coverage

Five tests, `tests/test_config_encoding.py` now 28 passed (was 23):

| Test | Validates |
|---|---|
| `test_failed_backup_aborts_the_rewrite_instead_of_destroying_the_bytes` | raises, and the original bytes are still on disk and still decode to `我是一名开发者` |
| `test_a_later_save_retries_a_backup_that_failed` | marker retained; the next save preserves the bytes |
| `test_failed_backup_leaves_no_partial_orig_file` | no `.orig*` left behind |
| `test_failed_backup_aborts_a_chat_config_save_too` | the atomic-rename path in `ChatConfig.save()` refuses too |
| — | plus the helper covering both the staged and final names |

Reverting only `gptme/config/` fails 4 of the 5 with `DID NOT RAISE OSError`. The fifth passes on both, and I'd rather say so than imply otherwise: the old single `write_bytes` also left nothing behind on failure, so that test guards the staged-rename design against a future regression rather than the bug being fixed.

`ruff check`, `ruff format --check` and `mypy` are clean on all three files. `tests/test_config.py` shows 16 failures, and the same 16 on `master` at `d83a093` — Windows symlink-privilege (`WinError 1314`), file-in-use (`WinError 32`) and path-separator assertions, none touching config encoding.

Still no CI has ever run on this branch: the runs at every head sit at `action_required` pending fork approval, so if you can release them the numbers above become checkable rather than reported.

ErikBjare:
@greptileai review

LHMQ878:
The fifth P1 was correct. Fixed in `f41531b`.

### Confirming it

`0xff` is invalid in UTF-8 *and* in cp936, so on a cp936 locale every candidate fails and the read falls through to `errors="replace"`. On `7e3386e`:

```
decoded text      : '[user]\nabout = "caf\ufffd"\n'
valid TOML?       : True -> {'about': 'caf\ufffd'}
marked unverified : False
files in dir      : ['config.toml']
orig bytes on disk?: False
recoverable?      : NO
```

So the save went through, wrote the U+FFFD out as UTF-8, and left no `.orig`.

### Why the parser wasn't the backstop here

`errors="replace"` is deliberate — I chose it so the *parser* decides what happens to a corrupt file, and `test_undecodable_config_fails_at_the_parser_not_the_decoder` pins that it does for `b'"\xff\xfe\x00garbage"'`. What I had missed is that reaching the parser and being rejected by it are different things: an undecodable byte *inside a quoted value* leaves the document well-formed, because U+FFFD is an ordinary character to TOML. `caf\ufffd` parses cleanly. That test was passing for a structurally-corrupt file and I read it as covering the path.

### The part that makes this the worst of the three decode outcomes

This is why I fixed it rather than documenting it. Ranked by what survives:

| read path | text recoverable from itself? |
|---|---|
| UTF-8 (declared) | n/a — not a guess |
| legacy codec (wrong one) | **yes** — `mojibake.encode("cp1252") == original_bytes` |
| `errors="replace"` | **no** — U+FFFD does not record which byte it stood for |

The milder failure was the one already protected, and the irreversible one was walking past the guard. Measured: `"caf\ufffd".encode(errors="replace") != b'caf\xff'`.

After the fix, the same probe end-to-end:

```
marked unverified : True
files in dir      : ['config.toml', 'config.toml.orig']
.orig == original bytes: True
```

### Tests

Three, in `tests/test_config_encoding.py`:

- `test_replacement_decoding_is_recorded_as_unverified` — the marker on this path.
- `test_replacement_decoded_config_is_backed_up_before_being_rewritten` — end to end through `set_config_value`; asserts the `.orig` is byte-identical.
- `test_replacement_decoded_config_can_still_be_valid_toml` — pins *why* the marker was needed: the document is valid TOML, and its text cannot be re-encoded back to the original bytes.

Discrimination: reverting only the `_mark_encoding_unverified(path)` call fails the first two and nothing else. The third passes either way by design — it is an observation about TOML and about U+FFFD, not about the fix, and I would rather it be stated than inferred from the other two. 31 passed with the fix, 29 passed / 2 failed without.

`ruff check`, `ruff format --check` and `mypy` are clean. `tests/test_config.py` has 16 pre-existing failures on this Windows box (identical count with the branch stashed), so those are unrelated to this change.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI green + Greptile 5/5 on head `f41531b2ae`. Ready for maintainer review/merge.

**Summary of what this PR does**: Adds explicit `encoding="utf-8"` to all `open()` calls in `gptme/config/`, matching the TOML spec. Adds graceful legacy-codepage fallback so users upgrading on Windows with old configs don't lose access, and backs up original bytes before any UTF-8 rewrite.

LHMQ878:
The two red jobs here are pre-existing on `master`, not caused by this PR — A/B against the current `master` head (`1b4616ee`, run [30797729948](https://github.com/gptme/gptme/actions/runs/30797729948)):

| job | this PR (`f41531b2`) | `master` (`1b4616ee`) |
|---|---|---|
| Test without API keys or extras | `1 failed, 8542 passed, 173 skipped` | `1 failed, 8578 passed, 173 skipped` |
| Test without API keys | `10197 passed, 42 skipped, 2 errors, 2 retried` | `10245 passed, 42 skipped, 1 error, 4 retried` |

Same failing test in both, same cause:

```
FAILED tests/test_skill_installer.py::TestInstallSkill::test_install_nonexistent_path
  - Failed: Timeout (>10.0s) from pytest-timeout.
```

The second job has zero actual test failures on either side — every test passes and the non-zero exit comes from a teardown `KeyError` in the retry plugin's stash:

```
ERROR tests/test_skill_installer.py::TestInstallSkill::test_install_nonexistent_path
  - KeyError: <_pytest.stash.StashKey object at ...>
```

which fires on `master` too, on the same test. `test_install_nonexistent_path` appears in the retry list on both sides — it's the flaky one, and it's a skill-installer path test with no relationship to TOML config I/O encoding. `lint`, `typecheck`, `openapi`, `build`, `Docker`, the PyInstaller matrix and Greptile Review are all green here.

Nothing on my side to change unless you'd like me to also look at the `test_install_nonexistent_path` timeout separately — happy to open a follow-up for it, but it seems out of scope for this one.

PR #3411, opened 2026-07-31
feat(webui): load session cost summaries in embedded mode

TimeToBuildBob:
## What

Always fetch detailed conversation summaries (token counts and cost) in the paginated conversation list.

## Why

The conversation list sidebar already renders token/cost fields when detailed summary fields are present. Without `detail=true`, gptme-cloud users (and standalone users) never saw per-session cost data in the list.

The earlier commit gated `detail=true` on `VITE_EMBEDDED_MODE`, but Erik correctly noted this was confusing — if cost detail is useful it should be available everywhere. The server API describes `detail=true` as "suitable for paginated views", and the 50-per-page limit already bounds each scan.

## Changes

- `useConversationsInfiniteQuery`: always pass `detail=true` to `getConversationsPaginated`
- Remove `shouldFetchDetailedConversationSummaries` utility and embedded-mode gate
- Remove the Jest AST transformer that was only needed to handle `import.meta.env` in that utility's test

640 Jest tests pass.


TimeToBuildBob:
Found and fixed a production-only regression in the Jest compatibility follow-up at `8dca40eba`.

Wrapping `import.meta.env` in `Function()` kept Jest green, but it also hid the expression from Vite's compile-time replacement. A standalone embedded build therefore fell back to an empty `process.env` object and requested `detail=false`. The cloud build happened to work only because its parent Vite config replaces the exact string inside the function body.

The hook now uses direct `import.meta.env.VITE_EMBEDDED_MODE` again, and Jest handles that syntax through a small test-only AST transformer. Verified the focused helper test, the previously failing `ChatMessage` suite (12 tests), TypeScript, ESLint, and pre-commit. Fresh CI is running on the corrected head.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Not sure this is the right appoach? Why would we only show in embedded mode? Why would this otherwise impact performance? Why do it in the embedded case but not otherwise? I don't get it. I've noticed cost-fetching is a bit weird/spotty, I guess this relates to why.

TimeToBuildBob:
You're right, the gate was confusing. The original reasoning: `detail=true` triggers a full server-side scan of all conversations to compute cost/token stats (not just the 50 on screen), so I gated it to embedded mode where sessions per user are bounded. But the server API itself describes this mode as "suitable for paginated views", and inconsistent behaviour between deployments is worse than the marginal perf concern.

Simplified in 743883f: always fetch `detail=true`, remove the `shouldFetchDetailedConversationSummaries` utility, its test, and the Jest AST transformer that was only added to support that test. The hook now just passes `true` directly.

640 Jest tests still pass.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob But doesn't this force a full JSON parse of the entire last 50 conversations etc? That could be 50 convos * 100 messages, might be slow. I think the current behavior is "load when conversation loads" which isn't that bad (but still no per-message cost info easily visible, current hover tooltip that sometimes works is meh)

TimeToBuildBob:
Yes. I re-read the server path and my previous answer was wrong: `paginated=1&detail=true` still calls `get_user_conversations(detail=True)` **before** applying the 50-item page limit, so it fully parses every message in every conversation on each uncached list fetch. That's the bad trade-off you were pointing at.

The current behavior already computes cost once a conversation is loaded, and the merged session-cost summary in Conversation Settings gives a non-hover aggregate for that loaded session. We should keep the list endpoint cheap rather than force full-history scans for sidebar badges.

Closing this PR. A proper list-level cost feature needs server-side cached/indexed aggregates or page selection before detail scanning; it shouldn't be implemented by globally enabling `detail=true`.

PR #3426, opened 2026-08-03
fix(android): print actual cert fingerprint on verification mismatch

### LINKED ISSUE #3408: Sign Android release APK/AAB in Tauri workflow
## Problem

The stable `v0.32.1` release publishes `gptme-tauri_0.32.1_universal-unsigned.apk`. The release workflow itself confirms this is intentional: `.github/workflows/tauri.yml` uploads Gradle's `app-universal-release-unsigned.apk` and says "the APK isn't signed yet." A content-level check with Android's `apksigner` reports `DOES NOT VERIFY` / missing signature metadata.

This means the direct Android download is not signed despite the assumption in https://github.com/gptme/gptme-cloud/pull/790#issuecomment-5140545049 that signing had landed.

## Scope

- Provision an Android upload/signing key through GitHub Actions secrets.
- Configure the release build to sign the APK and AAB without exposing key material.
- Run `apksigner verify --verbose --print-certs` on the final APK before upload.
- Publish a filename that does not say `-unsigned` only after verification passes.
- Document key rotation/recovery and expected certificate identity.

## Acceptance criteria

- A release run fails before upload if APK signature verification fails.
- The published APK passes `apksigner verify` and its certificate fingerprint matches the documented expected key.
- Signing secrets are unavailable to pull-request jobs and are never printed.
- The download metadata pipeline can classify the artifact from cryptographic verification rather than trust a filename.

## Evidence

- Release artifact: https://github.com/gptme/gptme/releases/download/v0.32.1/gptme-tauri_0.32.1_universal-unsigned.apk
- Producing run: https://github.com/gptme/gptme/actions/runs/29582932973
- Consumer-side verification: https://github.com/gptme/gptme-cloud/pull/796

TimeToBuildBob:
## Summary

- Capture `apksigner --print-certs` output to a variable before piping, so it can be reused for diagnostics
- Broaden sed pattern from exact `^Signer #1 certificate SHA-256 digest: ` prefix to `.*SHA-256 digest: *` — tolerates format variation across SDK/build-tools versions
- Fail with a dump of the full `--print-certs` output if fingerprint extraction yields empty string (parse failure is now its own error, not silent)
- Print **both** expected and actual fingerprints on mismatch so the operator can immediately read the real value and update `ANDROID_CERT_SHA256`

## Context

End-to-end validation run (30808558736, triggered to test #3408 signing) showed:

- APK signs successfully (v2/v3 schemes) ✅
- `apksigner verify --verbose` → "Verifies" ✅  
- Certificate fingerprint check → ❌ silent "does not match" with no actual value printed

The prior error message only named the file, making it impossible to distinguish two failure modes:

1. `sed` parse failure → `ACTUAL_CERT_SHA256=""` → always fails comparison
2. Genuine keystore/variable mismatch → correct value, wrong config

With this fix, the next CI run will print the actual fingerprint, and Erik can either confirm a sed format issue was the cause (unlikely but possible) or update `ANDROID_CERT_SHA256` in the `tauri-autoupdater-signing` environment with the real value.

Closes #3408 (pending fingerprint config confirmed correct after next run).

## Test plan

- [ ] Trigger `workflow_dispatch` with any `inputs.tag` after merging — the `Release Android` job will now print both expected and actual fingerprints on mismatch, or pass outright if the fingerprint was a sed parse issue all along
- [ ] If mismatch: copy the printed `actual:` value and update `ANDROID_CERT_SHA256` in the `tauri-autoupdater-signing` GitHub environment; re-trigger

### DIFF
--- a/.github/workflows/tauri.yml
+++ b/.github/workflows/tauri.yml
@@ -949,10 +949,17 @@ jobs:
           for apk in "${apks[@]}"; do
             echo "Verifying APK: $apk"
             "$APKSIGNER" verify --verbose "$apk"
+            # Label format varies across build-tools versions:
+            #   <=36: "Signer #1 certificate SHA-256 digest: <hex>"
+            #   37+:  "V3.0 Signer: certificate SHA-256 digest: <hex>"
+            # Match only *certificate* digests (not public-key digests) and
+            # dedupe: with one signing key every line carries the same value;
+            # multiple distinct certs concatenate and fail the comparison.
             PRINT_CERTS_OUT=$("$APKSIGNER" verify --print-certs "$apk")
-            ACTUAL_CERT_SHA256=$(printf '%s' "$PRINT_CERTS_OUT" \
-              | sed -n 's/.*SHA-256 digest: *//p' \
-              | tr '[:lower:]' '[:upper:]' | tr -d ':[:space:]')
+            ACTUAL_CERT_SHA256=$(printf '%s\n' "$PRINT_CERTS_OUT" \
+              | sed -n 's/.*certificate SHA-256 digest: *//p' \
+              | tr '[:lower:]' '[:upper:]' | tr -d ':' \
+              | sort -u | tr -d '[:space:]')
             if [ -z "$ACTUAL_CERT_SHA256" ]; then
               echo "::error::Could not extract certificate SHA-256 fingerprint from: $apk"
               printf '%s\n' "$PRINT_CERTS_OUT" | sed 's/^/  /'
--- a/.github/workflows/tauri.yml
+++ b/.github/workflows/tauri.yml
@@ -949,11 +949,19 @@ jobs:
           for apk in "${apks[@]}"; do
             echo "Verifying APK: $apk"
             "$APKSIGNER" verify --verbose "$apk"
-            ACTUAL_CERT_SHA256=$("$APKSIGNER" verify --print-certs "$apk" \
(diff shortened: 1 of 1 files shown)


ErikBjare:
Root cause confirmed by local reproduction — it was hypothesis 1 (sed parse failure), not a config mismatch:

- Downloaded the actual `android-build` artifact from run 30808558736, signed it locally with the real keystore, and ran `verify --print-certs` across build-tools 33.0.1 / 35.0.0 / 36.0.0 / 37.0.0.
- **build-tools 37.0.0 changed the output format**: `Signer #1 certificate SHA-256 digest:` → `V3.0 Signer: certificate SHA-256 digest:`. The ubuntu-latest runner has 37.0.0 installed and the workflow picks the newest via `sort -V`, so the anchored sed matched nothing → empty actual → guaranteed mismatch.
- The extracted fingerprint on every build-tools version equals the provisioned `ANDROID_CERT_SHA256` (`FFF9...3240`) exactly. **No secret or variable changes needed.**

Pushed 269a363dc hardening the broadened pattern: restrict to `certificate SHA-256 digest` (public-key digest lines also end in `SHA-256 digest:` in some formats) and `sort -u` so repeated identical values from multiple signer blocks (e.g. v3.1 rotation printing V3.0 + V3.1 blocks) collapse to one value, while genuinely distinct certs still concatenate and fail the comparison. Verified locally against both the 36 and 37 output formats plus multi-signer edge cases.

PR #3428, opened 2026-08-04
fix(server): check all sessions when guarding concurrent generation in /step

TimeToBuildBob:
## Summary

- **Bug**: `POST /step` only checked the requesting session's own `.generating` flag. Two clients holding different `ConversationSession` objects for the same `conversation_id` could both pass the guard and generate concurrently, writing conflicting messages to the same JSONL file.
- **Fix**: Align `/step` with every other write endpoint — use `any(s.generating for s in get_sessions_for_conversation(...))` to block if *any* session for the conversation is already generating.
- **Test**: Added `test_v2_step_rejected_while_other_session_generating` that patches `get_sessions_for_conversation` on the `api_v2_sessions` module to return a mock session with `generating=True`, then asserts that `/step` returns 409.

## Root cause detail

All of `POST /conversations/<id>`, `PATCH/DELETE /messages/<n>`, `PATCH /config`, and `DELETE /conversations/<id>` guard with:
```python
sessions = SessionManager.get_sessions_for_conversation(conversation_id)
if any(sess.generating for sess in sessions) or ...:
    return 409
```

But `/step` used:
```python
if session.generating or ...:
    return 409
```

where `session` is the single session associated with the requester. A second client with its own session would see `session.generating = False` and start a competing generation thread.

## Test plan
- [x] `test_v2_step_rejected_while_other_session_generating` passes
- [x] All 7 generating/command_is_active related tests pass
- [x] Pre-commit hooks pass (ruff format auto-corrected one file)

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -383,7 +383,10 @@ def api_conversation_step(conversation_id: str):
     # config PATCH. A PATCH completed before this acquisition must affect the
     # worker; one arriving afterward sees generating=True and returns 409.
     with SessionManager.conversation_lock(conversation_id), session.step_lock:
-        if session.generating or SessionManager.command_is_active(conversation_id):
+        conv_sessions = SessionManager.get_sessions_for_conversation(conversation_id)
+        if any(s.generating for s in conv_sessions) or SessionManager.command_is_active(
+            conversation_id
+        ):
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
✅ CI all green + Greptile 5/5 — ready for merge. Self-merge gate: not eligible (cross-repo path), requires human merge.

PR #3432, opened 2026-08-05
fix(server): propagate branch through tool execution and continuation step

TimeToBuildBob:
## Problem

`start_tool_execution` (and its internal `execute_tool_thread`) always
reloaded the conversation from the `"main"` branch (the `LogManager.load`
default), regardless of the branch passed to `/step`.  The continuation
`_start_step_thread` call at the end of tool execution had the same gap.

**Affected flow**: `POST /api/v2/conversations/<id>/step` with
`{"branch": "some-branch"}`.  If the LLM response contains tool uses:
1. The tool execution thread reloads the conversation from `main`, reading the
   wrong message history.
2. Tool outputs are appended to the `main`-branch manager and written to disk
   at the wrong branch location.
3. The continuation step after all tools finish also runs on `main`.

This is a wrong-but-200 bug: the API returns `{"status": "ok"}` but silently
operates on the wrong branch.

## Fix

Add `branch: str = "main"` to `start_tool_execution` and thread it through:
- `LogManager.load(..., branch=branch, ...)` inside `execute_tool_thread`
- `_start_step_thread(..., branch=branch, ...)` for the continuation

Update the call site in `step()` to forward its `branch` argument.

All existing callers (`tool/confirm`, `rerun`) pass no branch and continue
to default to `"main"`, which is the correct behaviour for those endpoints
(they have no branch context in their request).

## Test

New test `test_start_tool_execution_uses_branch_for_reload_and_continuation`
verifies that both `LogManager.load` and `_start_step_thread` receive the
branch forwarded via `start_tool_execution`.

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -660,7 +660,13 @@ def api_conversation_tool_confirm(conversation_id: str):
 
         logger.info(f"Executing runnable tooluse: {tooluse}")
         start_tool_execution(
-            conversation_id, session, tool_id, tooluse, model, chat_config
+            conversation_id,
+            session,
+            tool_id,
+            tooluse,
+            model,
+            chat_config,
+            branch=tool_exec.branch,
         )
         return flask.jsonify({"status": "ok", "message": "Tool confirmed"})
 
@@ -680,6 +686,7 @@ def api_conversation_tool_confirm(conversation_id: str):
             dataclasses.replace(tool_exec.tooluse, content=edited_content),
             model,
             chat_config,
+            branch=tool_exec.branch,
         )
 
     elif action == "skip":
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
Addressed the Greptile P1 finding in `94acff912`: the deferred confirmation handlers (`confirm`, `edit`, `auto`, `skip`) now propagate the originating branch through to tool execution and the continuation step.

**Root cause**: `ToolExecution` had no `branch` field, so when the client later called `/tool/confirm`, the endpoint had no record of which branch the original `/step` ran on. All four action paths defaulted to `"main"`.

**Fix**:
- Added `branch: str = "main"` to `ToolExecution` dataclass
- Set `branch=branch` when tools are created in `_flush_sse_batch`
- Passed `tool_exec.branch` to `start_tool_execution` in the `confirm`, `edit`, and `auto` handlers
- Passed `current_tool.branch` to both `LogManager.load` and `_start_step_thread` in the `skip` handler

Combined with the previous commit's fixes for the auto-confirm path and continuation step inside `step()`, branch is now propagated end-to-end through all tool execution paths.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

PR #3407, opened 2026-07-31
fix(webui): replace CLI server guidance with in-app retry in Tauri desktop app

TimeToBuildBob:
## Summary

Fixes #3406.

When the gptme desktop app (Tauri) fails to start the bundled sidecar server, the WelcomeView was showing CLI-oriented content that is confusing and useless for desktop users:
- "The gptme server is not running. Start one with the command below."
- `pipx install 'gptme[server]'` and `gptme-server --cors-origin=...` code blocks
- "Copy start command" button
- "Prefer not to run a local server? Use gptme.ai"

In the desktop app, the user cannot run these CLI commands — the app manages the server sidecar itself.

**Changes:**
- When running in Tauri with `managesLocalServer=true`, show a desktop-friendly message: _"The built-in server failed to start. Click 'Restart server' to try again."_
- Add a **Restart server** button that calls the `start_server` Tauri IPC command and then reconnects; handles the "server already running" edge case by falling through to `connect()` directly
- Hide CLI command blocks, "Copy start command" button, and "Use gptme.ai" cloud link in the desktop context
- Keep **Retry connection** as a lightweight fallback alongside **Restart server**
- Gate the "Get started" setup wizard CTA on non-Tauri (desktop users don't need CLI onboarding)

All existing 24 WelcomeView tests pass; 5 new Tauri-specific tests added covering the message, the restart button flow, the always-running button, the hidden cloud CTA, and the graceful "already running" fallback.

## Test plan

- [ ] In the Tauri desktop app with a failed sidecar start, the banner shows "The built-in server failed to start" and a "Restart server" button, no CLI commands
- [ ] Clicking "Restart server" attempts `start_server` IPC and then connects
- [ ] "Retry connection" still appears alongside "Restart server"
- [ ] In the web UI (non-Tauri), the existing CLI guidance is unchanged
- [ ] All 640 webui unit tests pass (`npx jest --no-coverage` in `webui/`)

PR #3445, opened 2026-08-05
fix(webui): keep setup wizard reachable after skip and on mobile Tauri

ErikBjare:
Feedback from testing the new Android app (signed builds from #3409):

> during onboarding i couldn't select cloud-setup ("not available yet" or some such) and after i'd clicked "skip" there was no way to get back to the onboarding

Two compounding gates made onboarding unreachable on the Android app:

1. `showGuidedSetup` requires `isFirstVisit` (`!hasCompletedSetup`) — clicking **Skip** sets `hasCompletedSetup=true`, removing the "Get started" CTA everywhere.
2. The `!isTauri` gate added in #3407 (to hide CLI server guidance in the desktop app) also swept up the wizard CTA on **all** Tauri builds — including remote-only mobile, where the wizard is the primary way to configure a server URL.

## Fix

Show the wizard CTA whenever the disconnected banner is up on the default server: label stays "Get started" for first-timers, becomes "Run setup" for returning users. Remains hidden on desktop Tauri (which manages its own server, per #3407) and in demo mode. The Settings → About → "Re-run Setup" path is unchanged.

Not addressed here: cloud setup being disabled on mobile builds is by design ("coming soon on mobile") — this PR just makes sure skipping past it isn't a dead end.

## Tests

Two new WelcomeView tests: returning users get a working "Run setup" CTA (regression), and remote-only Tauri shows the CTA. All 31 pass locally.

### DIFF
--- a/webui/src/components/WelcomeView.tsx
+++ b/webui/src/components/WelcomeView.tsx
@@ -206,6 +206,12 @@ export const WelcomeView = () => {
   // their disconnected banner is left unchanged.
   const isFirstVisit = !settings.hasCompletedSetup;
   const showGuidedSetup = isDefaultLocalServer && isFirstVisit && !demoMode;
+  // Wizard CTA stays available after setup was skipped/completed so onboarding
+  // remains reachable, and on remote-only Tauri (mobile) where the wizard is the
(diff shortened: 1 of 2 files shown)


ErikBjare:
@greptileai review

PR #3448, opened 2026-08-05
fix(tests): catch TypeError in _has_flask() for Python 3.13 compatibility

TimeToBuildBob:
## Summary

Follow-up to #3436 (per-step timing breakdown).

`_has_flask()` in `tests/test_timings.py` only caught `ImportError`, but Flask 3.0 raises `TypeError` (not `ImportError`) when running on Python 3.13 — the `ConfigAttribute[bool]` subscript syntax is not supported. This caused pytest collection to fail with:

```
TypeError: type 'ConfigAttribute' is not subscriptable
```

instead of gracefully skipping the server-path test.

## Change

- Catch `(ImportError, TypeError)` in `_has_flask()` so the `@pytest.mark.skipif` decorator works correctly on Python 3.13 environments where Flask is installed but broken.

## Verification

```
8 passed, 1 skipped  # (server test correctly skipped when Flask is broken)
```

### DIFF
--- a/tests/test_timings.py
+++ b/tests/test_timings.py
@@ -17,12 +17,12 @@ from gptme.message import Message, MessageMetadata, MessageTimings
 
 
 def _has_flask() -> bool:
-    """Check if flask is available (required for server components)."""
+    """Check if flask is available and importable (required for server components)."""
     try:
         import flask  # noqa: F401
 
         return True
-    except ImportError:
+    except (ImportError, TypeError):
         return False
 
 


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_timings.py` (+2/-2, Δ4)

### Suggested verification
- `uv run pytest tests/test_timings.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3448._

ErikBjare:
@TimeToBuildBob What? Why would flask be broken like that? Doesn't really make sense to me why 3.13 wouldn't support that?

TimeToBuildBob:
Good catch — the PR description is wrong about Python 3.13 being the cause.

After reviewing my own lesson log (I hit this exact error last week, session 7760), the actual root cause is a **corrupted uv-cache wheel**: Flask 3.x's `app.py` uses `ConfigAttribute[bool]` while the installed `config.py` still has the Flask 2.x non-generic class. That mismatch raises `TypeError: type 'ConfigAttribute' is not subscriptable` on *any* Python version — it's not 3.13-specific.

The real fix for a broken environment is:
```
uv pip install --force-reinstall --no-cache flask
```

That said, the code change itself is still defensively correct: if Flask fails to import for *any* reason (corrupted install, genuine incompatibility, whatever), `_has_flask()` should return `False` and skip the server test rather than crashing pytest collection. Catching `(ImportError, TypeError)` is the right defensive posture for a test guard function.

Want me to update the PR description to reflect the correct diagnosis?

TimeToBuildBob:
Flask 3.x added `ConfigAttribute[bool]` type annotations directly in the `Flask` class body (e.g. `testing: ConfigAttribute[bool] = ConfigAttribute("TESTING")`). When Python evaluates that class body at import time, it tries to call `ConfigAttribute.__class_getitem__(bool)` — but `ConfigAttribute` doesn't implement `__class_getitem__`, so it raises `TypeError: type 'ConfigAttribute' is not subscriptable`.

It's not a Python 3.13 regression exactly — it's a Flask bug (they annotated their attributes with generic subscript syntax without making the class actually subscriptable). The specific environment that triggered it was Python 3.13 + Flask 3.x in our CI. Newer Flask versions fix this on their end.

Catching `(ImportError, TypeError)` is the minimal correct fix: we don't care *why* Flask is broken, just that it is.

PR #3450, opened 2026-08-05
fix(webui): scroll past InlineToolExecution during tool execution (#3440)

### LINKED ISSUE #3440: UI tests for text jumping around during generation/state changes
Some issues:

 - claude-sonnet-4-5 is everywhere/very often seen as default? (at least in gptme-cloud prod, might just be my old instance, but still)
 - when a conversation is started with another model, initially the chat input model selector still says claude-sonnet-4-5 but after a while/if i open chat settings it switches to the right model (the one chosen when starting the convo)
 - during multi-step turns by the assistant with multiple tool-use, the text/scroll on the page jumps around as elements jump and down as things get folded in or updated, tool confirmation prompt and in-progress indicator/cards, etc. not very readable/good UX. worth a proper pass in itself to make whole UX cleaner and less large card-y (and not multiple spinner icons... that one has been driving me a bit nuts - that could all be a lot cleaner - make the "accept all" a much easier click)

These things need tests general to catch their broad type.

Also, not a UI test issue but I also noticed deepseek-v4-flash-0731 didn't seem to use parallel tool calling by default? is it not configured or supported for it? (open separate issue for the general "parallel tool calling" model/tool-format test/eval)

TimeToBuildBob:
## What

Fixes the remaining scroll-jumping issue from #3440 — scroll position now stays at the actual container bottom when the InlineToolExecution card appears/disappears during multi-step tool runs.

## Root cause

`scrollToBottom()` called `scrollToIndex(count-1, {align:'end'})` which positions the last *virtual item* at the viewport bottom. Elements below the virtualizer (the InlineToolExecution card, the completion badge, bottom padding) were not accounted for.

When InlineToolExecution (~300px) appeared, the container's `scrollHeight` grew but the virtualizer scroll stopped short. The `onScroll` handler saw `scrollHeight - scrollTop - clientHeight > 1` and set `autoScrollAborted = true`, **silently killing auto-scroll for the rest of the tool run**.

## Fix

1. **`scrollToBottom`** — adds a second `requestAnimationFrame` after `scrollToIndex` that sets `container.scrollTop = scrollHeight - clientHeight`. `isAutoScrolling$` stays `true` across both frames so the `onScroll` handler does not treat the intermediate position as a manual scroll.

2. **New `useObserveEffect` on `executingTool$`** — triggers `scrollToBottom` when the card appears/disappears, even when no log update fires simultaneously.

## Tests

Four new assertions in `ConversationContent.test.tsx`:

- `InlineToolExecution` renders when `executingTool$` is set
- `InlineToolExecution` is absent when `executingTool$` is null
- `scrollToIndex` is called when `executingTool$` transitions null → set
- `container.scrollTop` is set to `scrollHeight - clientHeight` after `scrollToIndex`

All 678 unit tests pass.

Closes #3440

### DIFF
--- a/webui/src/components/ConversationContent.tsx
+++ b/webui/src/components/ConversationContent.tsx
@@ -396,13 +396,23 @@ export const ConversationContent: FC<Props> = ({ conversationId, serverId, isRea
     const count = visibleMessageIndicesRef.current.length;
     if (count <= 0) return;
     isAutoScrolling$.set(true);
-    // scrollToIndex ensures the last item is actually rendered before measuring,
-    // which is more reliable than container.scrollHeight with virtual lists.
+    // scrollToIndex ensures the last virtual item is rendered before measuring.
     virtualizerRef.current.scrollToIndex(count - 1, { align: 'end' });
+    // A second rAF scrolls the container to its true bottom so elements below
+    // the virtual list (InlineToolExecution card, completion badge, bottom pad)
+    // are also in view.  isAutoScrolling$ stays true across both frames so the
+    // onScroll handler does not mistake the intermediate position for a manual
+    // scroll and abort auto-scroll.
     requestAnimationFrame(() => {
-      isAutoScrolling$.set(false);
+      const container = scrollContainerRef.current;
+      if (container) {
+        container.scrollTop = container.scrollHeight - container.clientHeight;
+      }
+      requestAnimationFrame(() => {
+        isAutoScrolling$.set(false);
+      });
     });
-  }, [isAutoScrolling$]);
+  }, [isAutoScrolling$, scrollContainerRef]);
 
   // Auto-scroll when the conversation is updated (e.g., streaming response).
   // Cancel any pending rAF before scheduling a new one so rapid log updates
@@ -417,6 +427,16 @@ export const ConversationContent: FC<Props> = ({ conversationId, serverId, isRea
         scrollRAFRef.current = null;
         scrollToBottom();
       });
+    }
+  });
+
+  // When the executing-tool card appears or disappears its height is added to /
+  // removed from the scroll container outside the virtualizer.  Fire an explicit
+  // scroll-to-bottom on those transitions so the card stays in view and the
+  // onScroll handler does not abort auto-scroll from a stale intermediate position.
+  useObserveEffect(conversation$?.executingTool, () => {
+    if (!autoScrollAborted$.get()) {
+      requestAnimationFrame(scrollToBottom);
     }
   });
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `webui/src/components/__tests__/ConversationContent.test.tsx` (+108/-1, Δ109)
- `webui/src/components/ConversationContent.tsx` (+24/-4, Δ28)

### Suggested verification
- `make lint`
- `make test`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3450._

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `fd2c895` addressing the Greptile P1 finding: the `executingTool$` observer only checked `autoScrollAborted$` when *scheduling* the rAF, not when it actually fired. A manual scroll landing in that ~16ms gap set `autoScrollAborted$=true`, but the already-queued `scrollToBottom()` ran anyway and forced the viewport back to the bottom — overriding the user's scroll.

Fix: recheck `autoScrollAborted$` inside the rAF callback before invoking `scrollToBottom`, so a manual scroll that lands between scheduling and the frame firing is respected.

Verified locally: typecheck clean, `ConversationContent.test.tsx` 28/28 passing, `eslint` clean on the changed file.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `7ac6136` addressing the remaining Greptile P1 finding ("Deferred scroll overrides user input").

**Root cause**: `scrollToBottom()` sets `isAutoScrolling$=true` before calling `scrollToIndex`, which blocks `onScroll` from updating `autoScrollAborted$`. If the user drags the viewport up during this lock window (between the flag being set and the first inner `requestAnimationFrame` firing), their scroll is invisible to the code — the rAF then unconditionally assigns `container.scrollTop = scrollHeight - clientHeight`, snapping them back to the bottom.

**Fix**: Capture `scrollTopSnapshot` before `scrollToIndex`. In the first rAF, if `scrollTop < scrollTopSnapshot`, the viewport regressed while `onScroll` was blocked — abort without snapping and set `autoScrollAborted$=true`.

New test (`Greptile P1 pre-rAF race`) verifies the snap is prevented: `mockScrollToIndex` drops `scrollTop` from 200 → 50 during the lock, and the test asserts no programmatic `scrollTop` assignment occurs. All 7 scroll-stability tests pass.

TimeToBuildBob:
@greptileai review

PR #3471, opened 2026-08-06
fix(test): prevent race condition in subagent watchdog test setup

TimeToBuildBob:
## Problem

The `TestMaxTimeWatchdog.setup_method()` was calling `.clear()` on shared module-level dicts (`_subagents`, `_subagent_results`) without ensuring that threads from previous tests had finished. This caused a **'dictionary changed size during iteration'** error when a lingering thread was still in a generator iterating the dict at the moment `setup_method` called `.clear()`.

## Root Cause

Each test in `TestMaxTimeWatchdog` creates daemon threads. When a thread calls `t.join(timeout=1)` in its finally block, there's a race: if the thread doesn't exit within 1 second, it keeps running. On the next test's `setup_method`, the code tries to `.clear()` the shared dicts while that leftover thread is still iterating them via a generator expression like:

```python
(s for s in _subagents if s.agent_id == agent_id)
```

Python raises `RuntimeError: dictionary changed size during iteration` when this happens.

## Solution

Before clearing the dicts in `setup_method`, join any leftover threads with a 0.5s timeout. This ensures all generators consuming the dicts have completed before we mutate them.

## Testing

- ✅ `test_subagent_wait_returns_cached_timeout_while_thread_is_still_alive` now passes consistently
- ✅ All 9 tests in `TestMaxTimeWatchdog` pass
- ✅ Pre-commit checks pass

Fixes: #3465

### DIFF
--- a/tests/test_subagent_unit.py
+++ b/tests/test_subagent_unit.py
@@ -2761,7 +2761,13 @@ class TestMaxTimeWatchdog:
             _subagents_lock,
         )
 
+        # Ensure any leftover threads from prior tests are stopped before clearing.
+        # This prevents "dictionary changed size during iteration" errors when
+        # setup_method clears the shared dicts while threads are still iterating.
         with _subagents_lock:
+            for sa in _subagents:
+                if sa.thread and sa.thread.is_alive():
+                    sa.thread.join(timeout=0.5)
             _subagents.clear()
         with _subagent_results_lock:
             _subagent_results.clear()
--- a/tests/test_subagent_unit.py
+++ b/tests/test_subagent_unit.py
@@ -2764,10 +2764,12 @@ class TestMaxTimeWatchdog:
         # Ensure any leftover threads from prior tests are stopped before clearing.
         # This prevents "dictionary changed size during iteration" errors when
         # setup_method clears the shared dicts while threads are still iterating.
+        # Join threads WITHOUT a timeout — they must finish cleanly, or the test
+        # has a real issue (daemon threads that never exit indicate a resource leak).
+        for sa in list(_subagents):
+            if sa.thread and sa.thread.is_alive():
+                sa.thread.join()  # Block indefinitely until thread finishes
         with _subagents_lock:
-            for sa in _subagents:
-                if sa.thread and sa.thread.is_alive():
-                    sa.thread.join(timeout=0.5)
             _subagents.clear()
         with _subagent_results_lock:
             _subagent_results.clear()


TimeToBuildBob:
## Fix Applied: Race Condition in Thread Cleanup

Thanks @greptile-apps for catching the incomplete fix!

### The Problem (Greptile's Finding)
The 0.5s `join(timeout=...)` was insufficient: if a thread didn't stop within that window, it could still write stale state to the dict after `setup_method` cleared it. This violates the thread safety invariant — threads must be stopped completely before we mutate shared state.

### The Solution
1. **Remove the timeout**: Threads must now finish cleanly before `setup_method` proceeds. If a thread hangs, the test hangs and we catch the resource leak immediately.
2. **Move join outside the lock**: Holding the lock during join could deadlock (thread needs lock to finish, but we're holding it). Join first while others can proceed.
3. **Iterate over `list(_subagents)`**: Avoids the "dictionary changed size during iteration" error even if a thread modifies the list.

### Verification
- All 9 `TestMaxTimeWatchdog` tests pass
- Runtime: 0.74s (threads finish quickly without timeout)
- Pre-commit checks pass
- Coverage concern resolved (join line now execised by tests)

The fix is now deterministic instead of probabilistic. 🎯

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Good question — two-part answer:

1. **GIL protection**: In CPython, `list(dict)` is atomic at the C level due to the GIL. The snapshot is internally consistent — no torn read possible.

2. **Test context safety**: More importantly, no new entries can be added to `_subagents` during `setup_method`. Tests run sequentially (no pytest-xdist parallelism here). Any leftover threads from the previous test are in their *cleanup* phase — they're finishing generator exhaustion and `join()` cleanup, not spawning new subagents. A thread that's being joined can't be adding new `Subagent` entries to `_subagents` because subagent creation is gated through `_create_subagent_thread()` which only the test body calls (not the daemon thread itself).

So even if a thread did somehow add an entry after the snapshot (it can't, but hypothetically), that entry would be cleaned up by the subsequent `with _subagents_lock: _subagents.clear()` immediately after — which is fine since that new subagent would have no associated test expecting it.

The fix is sound. Thanks for the thorough review!

TimeToBuildBob:
Closing — the test fix has been cherry-picked directly into #3465 (commits 711a59a6d, afaea7d71). This PR is now redundant.

PR #3481, opened 2026-08-07
fix(server): prevent premature continuation when confirming concurrent tools

### LINKED ISSUE #3479: bug(server): concurrent non-auto-confirm tool confirmation causes out-of-order transcript
## Summary

When two non-auto-confirm tools from the same assistant message are confirmed simultaneously (e.g. via a "confirm all" UI button or rapid programmatic client), the continuation LLM step can start before the slower tool has written its output. The final transcript has the continuation assistant message appearing *before* the slower tool's result — wrong order, wrong LLM context. This is a wrong-but-200 semantic correctness bug.

## Root Cause

In `gptme/server/session_step.py::execute_tool_thread`:

```python
# Both threads .pop() their tool at the START of execution:
tool_exec = session.pending_tools.pop(current_tool_id, None)
...
# After executing, the check fires:
if not session.pending_tools:
    _start_step_thread(...)  # <- triggers continuation
```

`pending_tools` becomes empty when all tools are *popped*, not when all tools have finished *writing* their results. If Tool A pops and executes fast, it sees an empty `pending_tools` (Tool B already popped) and triggers the continuation while Tool B is still executing.

## Steps to Reproduce

1. Assistant message generates 2 tool uses, both `auto_confirm=False`
2. Client sends two `/tool/confirm` requests simultaneously (concurrent HTTP)
3. Tool 1 executes fast (e.g. simple echo), Tool 2 executes slow (e.g. sleep + output)
4. Observe final log ordering: continuation assistant message appears BEFORE Tool 2's output

Trigger: `confirm all` UI action, or any programmatic client confirming multiple tools rapidly.

## Affected Code

- `session_step.py:1044` - `.pop()` at top of execute_tool_thread
- `session_step.py:1190-1197` - `if not session.pending_tools:` check (AFTER pop, not after write)
- `api_v2_sessions.py:662` - confirm endpoint calls `start_tool_execution(reserved=False)`

Note: auto-confirm tools are NOT affected — `step()` starts only the first auto-confirm tool and `execute_tool_thread` chains them serially. The race only affects simultaneous *manual* confirmations.

## Fix Direction

Track execution state separately from pending state:

```python
# Add _executing_tools: set[str] to ConversationSession
# In execute_tool_thread, after popping, under conversation_lock:
session._executing_tools.add(current_tool_id)
# ... execute and write result under lock ...
session._executing_tools.discard(current_tool_id)
# Trigger continuation only when both sets empty:
if not session.pending_tools and not session._executing_tools:
    _start_step_thread(...)
```

The add/discard needs to be under the conversation_lock to prevent the symmetric race where both threads simultaneously see both sets empty.

## Test Gap

No existing test in `test_server_v2_sessions.py` covers simultaneous confirmation of two different tools from the same assistant message.

TimeToBuildBob:
## Problem

When two non-auto-confirm tools from the same assistant message are confirmed simultaneously (e.g. "confirm all" UI button or rapid programmatic confirms), both threads pop their tool from `pending_tools` at the start. The fast thread then sees `pending_tools` empty and fires `_start_step_thread` before the slow thread finishes writing its output — producing an out-of-order transcript.

Closes #3479

## Root Cause

In `execute_tool_thread`, `.pop()` happens at the top of the while loop outside any lock. Both concurrent threads pop their tool immediately, leaving `pending_tools = {}`. The fast thread finishes its write and checks `if not session.pending_tools` → `True` → triggers continuation while the slow thread is still executing.

## Fix

Add `_executing_tools: set[str]` to `ConversationSession` to track tools that have been claimed (popped) but haven't yet finished writing their results.

Key invariant: the `pop` and the `_executing_tools.add` are both performed under `conversation_lock`, so no sibling thread can observe `pending_tools` empty while another thread hasn't yet registered itself as executing. The `discard` happens inside the same lock that guards the write, so the continuation check (`not pending_tools and not _executing_tools`) can only be `True` once the last thread has fully written its output.

`_start_step_thread(reserved=False)` already gates double-calls via the `generating` flag, so two threads simultaneously passing the check safely reduces to a single step start.

Note: this is distinct from PR #3457 (interrupt flag) — that fixes continuation after interrupt; this fixes ordering when multiple manual confirmations arrive concurrently.

## Test plan

`TestConcurrentToolConfirmation::test_no_premature_step_with_concurrent_confirmations` uses a threading event to gate the slow tool's write:
1. Tool 1 (fast) runs immediately; tool 2 (slow) waits for `slow_can_proceed`
2. After tool 1 finishes, asserts `_start_step_thread` was NOT called (tool 2 still in `_executing_tools`)
3. Releases tool 2; after both threads complete, asserts exactly one step trigger with `_executing_tools` already empty

This test fails on the old code and passes with the fix.

- [x] `uv run pytest tests/test_server_v2_sessions.py -q` — 149 passed

### DIFF
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -1057,20 +1057,32 @@ def start_tool_execution(
                         session.generating = False
                         session.generating_since = None
                     return  # another thread claimed this tool; don't trigger auto-step
-                if assistant_msg_timestamp is None:
-                    assistant_msg_timestamp = tool_exec.assistant_msg_timestamp
-                tool_exec.status = ToolStatus.EXECUTING
+                # The claim is registered above but the try/finally that releases
+                # it only starts below, so anything that raises in between (most
+                # plausibly add_event, which iterates sessions and trims their
+                # event buffers) would strand current_tool_id in
+                # _executing_tools for the lifetime of the session. Nothing else
+                # ever clears that set, and the continuation gate requires it to
+                # be empty — so a single failure here would silently stop every
+                # later tool in this session from producing an assistant reply.
+                try:
+                    if assistant_msg_timestamp is None:
+                        assistant_msg_timestamp = tool_exec.assistant_msg_timestamp
+                    tool_exec.status = ToolStatus.EXECUTING
 
-                # use explicit tooluse if set (may be modified), else from pending
-                tooluse: ToolUse = current_edited_tooluse or tool_exec.tooluse
+                    # use explicit tooluse if set (may be modified), else from pending
+                    tooluse: ToolUse = current_edited_tooluse or tool_exec.tooluse
 
-                # Record start time and notify about tool execution
-                tool_exec.started_at = time.monotonic()
-                SessionManager.add_event(
-                    conversation_id,
-                    {"type": "tool_executing", "tool_id": current_tool_id},
-                )
-                logger.info(f"Tool {current_tool_id} executing")
+                    # Record start time and notify about tool execution
+                    tool_exec.started_at = time.monotonic()
+                    SessionManager.add_event(
+                        conversation_id,
+                        {"type": "tool_executing", "tool_id": current_tool_id},
+                    )
+                    logger.info(f"Tool {current_tool_id} executing")
+                except BaseException:
+                    session._executing_tools.discard(current_tool_id)
+                    raise
 
                 # Execute the tool
                 try:
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -1038,10 +1038,14 @@ def start_tool_execution(
                 # the correct message history, not always the "main" branch.
                 manager = LogManager.load(conversation_id, branch=branch, lock=False)
 
-                # Atomically claim the tool with pop(): a get-then-pop sequence
-                # leaves a window where two threads (e.g. two concurrent confirm
-                # requests) both see the tool and both execute it.
-                tool_exec = session.pending_tools.pop(current_tool_id, None)
+                # Atomically claim the tool with pop() and register it as
+                # executing — both under conversation_lock so no sibling thread
+                # can see pending_tools empty while this thread hasn't yet
+                # added itself to _executing_tools.
+                with SessionManager.conversation_lock(conversation_id):
+                    tool_exec = session.pending_tools.pop(current_tool_id, None)
+                    if tool_exec is not None:
+                        session._executing_tools.add(current_tool_id)
                 if tool_exec is None:
                     logger.warning(
                         f"Tool {current_tool_id} not found in pending tools "
@@ -1116,6 +1120,7 @@ def start_tool_execution(
                         )
                         for tool_output in tool_outputs:
                             _append_and_notify(manager, session, tool_output)
+                        session._executing_tools.discard(current_tool_id)
                 except Exception as e:
                     logger.exception(f"Error executing tool {tooluse.tool}: {e}")
                     tool_exec.status = ToolStatus.FAILED
@@ -1128,6 +1133,7 @@ def start_tool_execution(
                             "system", f"Error: {e!s}", call_id=tooluse.call_id
                         )
                         _append_and_notify(manager, session, msg)
+                        session._executing_tools.discard(current_tool_id)
 
                 # Emit tool_complete with duration; also accumulate for metadata.
                 if tool_exec.started_at is not None:
@@ -1184,10 +1190,12 @@ def start_tool_execution(
                     branch=branch,
                 )
 
-            # Only auto-step when all pending tools have been executed.
-            # With multiple tools per message, we must wait until every tool
-            # has run before asking the model for a continuation.
-            if not session.pending_tools:
+            # Only auto-step when all pending tools have been executed AND all
+            # concurrent execution threads have written their results.
+            # pending_tools tracks tools awaiting confirmation; _executing_tools
+            # tracks tools that were popped (confirmed) but haven't finished
+            # writing yet — both must be empty before asking the model to continue.
+            if not session.pending_tools and not session._executing_tools:
                 _start_step_thread(
                     conversation_id,
                     session,
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -1084,120 +1084,120 @@ def start_tool_execution(
                     session._executing_tools.discard(current_tool_id)
                     raise
 
-                # Execute the tool
+                claimed_tool_id = current_tool_id
                 try:
-                    logger.info(f"Executing tool: {tooluse.tool}")
-                    stream_tool_id = current_tool_id
+                    # Execute the tool
+                    try:
+                        logger.info(f"Executing tool: {tooluse.tool}")
+                        stream_tool_id = current_tool_id
 
-                    def stream_tool_output(
-                        tool_output: Message, tool_id: str = stream_tool_id
-                    ) -> None:
-                        if (
-                            tool_output.role == "system"
-                            and not tool_output.hide
-                            and not tool_output.quiet
-                        ):
-                            SessionManager.add_event(
-                                conversation_id,
-                                {
-                                    "type": "tool_output",
-                                    "tool_id": tool_id,
-                                    "output": tool_output.content,
-                                },
+                        def stream_tool_output(
+                            tool_output: Message, tool_id: str = stream_tool_id
+                        ) -> None:
+                            if (
+                                tool_output.role == "system"
+                                and not tool_output.hide
+                                and not tool_output.quiet
+                            ):
+                                SessionManager.add_event(
+                                    conversation_id,
+                                    {
+                                        "type": "tool_output",
+                                        "tool_id": tool_id,
+                                        "output": tool_output.content,
+                                    },
+                                )
+
+                        tool_outputs = list(
+                            tooluse.execute(
+                                log=manager.log,
+                                workspace=manager.workspace,
+                                on_result_message=stream_tool_output,
                             )
-
-                    tool_outputs = list(
-                        tooluse.execute(
-                            log=manager.log,
-                            workspace=manager.workspace,
-                            on_result_message=stream_tool_output,
                         )
-                    )
-                    logger.info(
-                        f"Tool execution complete, outputs: {len(tool_outputs)}"
-                    )
-
-                    # Store the tool outputs. call_id is already assigned in
-                    # ToolUse.execute() for real results; hook messages intentionally
-                    # have no call_id — don't re-stamp here or hook messages become
-                    # duplicate function_call_output entries (Responses API 400).
-                    #
-                    # Reload under conversation_lock before appending: without this,
-                    # a stale in-memory manager (loaded at the top of the loop) would
-                    # rewrite the full JSONL, overwriting concurrent tool-result appends
-                    # or timing patches written by other confirmation threads.
-                    with SessionManager.conversation_lock(conversation_id):
-                        manager = LogManager.load(
-                            conversation_id, branch=branch, lock=False
+                        logger.info(
+                            f"Tool execution complete, outputs: {len(tool_outputs)}"
                         )
-                        for tool_output in tool_outputs:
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_server_v2_sessions.py` (+152/-0, Δ152)
- `gptme/server/session_step.py` (+16/-8, Δ24)
- `gptme/server/session_models.py` (+4/-0, Δ4)

### Suggested verification
- `uv run pytest tests/test_server_v2_sessions.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3481._

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Ready for review: CI green (all 14 checks passed), Greptile 5/5. This touches `session_models.py` and `session_step.py` so it's outside the auto-merge allowlist — needs your eyes before landing.

gptme/server/session_step.py
```diff
@@ -1128,6 +1132,15 @@ def stream_tool_output(
...
                         _append_and_notify(manager, session, msg)
+                finally:
+                    # Unconditional cleanup: whatever happens above (success,
+                    # a caught Exception, a BaseException like KeyboardInterrupt,
+                    # or a failure while persisting the result/error message
+                    # itself), the claim must be released or this tool id stays
+                    # in _executing_tools forever, permanently blocking the
+                    # `not pending_tools and not _executing_tools` continuation
+                    # check for this session.
+                    session._executing_tools.discard(current_tool_id)
```
TimeToBuildBob: ❌ **P1** — The PR's stated invariant — that the claim is released "inside the same lock that guards the write" — is not implemented: the `finally` at lines 1135-1143 runs after the `with SessionManager.conversation_lock(...)` blocks at 1117 and 1127 have already exited, so `discard` is unlocked. Two concurrent confirmation threads can therefore interleave as T1 discards tool1, T2 discards tool2, then both evaluate `not session.pending_tools and not session._executing_tools` at line 1205 and both call `_start_step_thread`. For `reserved=False` this collapses to one step because `_start_step_thread` re-checks `conversation_generating` under the lock, but for `reserved=True` the check is skipped (line 1256 `if not reserved:`) and two continuation steps run concurrently. Worse, a `reserved=True` thread that finishes while a sibling is still writing now hits `elif reserved:` at 1214 and clears `session.generating=False`; when the sibling later starts the continuation it does so with `reserved=True`, never re-setting `generating`, so the whole continuation runs with `generating=False` and a client `/step` sees the conversation as idle and starts a parallel generation — the exact interleaved-transcript corruption #3479 is about, at the generation level.

gptme/server/session_step.py
```diff
@@ -1038,10 +1038,14 @@ def execute_tool_thread() -> None:
...
-                # requests) both see the tool and both execute it.
-                tool_exec = session.pending_tools.pop(current_tool_id, None)
+                # Atomically claim the tool with pop() and register it as
+                # executing — both under conversation_lock so no sibling thread
+                # can see pending_tools empty while this thread hasn't yet
+                # added itself to _executing_tools.
+                with SessionManager.conversation_lock(conversation_id):
+                    tool_exec = session.pending_tools.pop(current_tool_id, None)
+                    if tool_exec is not None:
+                        session._executing_tools.add(current_tool_id)
```
TimeToBuildBob, line 1048: ⚠️ **P2** — The comment at lines 1136-1142 promises "Unconditional cleanup: whatever happens above ... the claim must be released or this tool id stays in _executing_tools forever", and the `finally` at 1143 does guarantee this for the body of the inner try/except (the execute plus write/error-append at 1076-1134). But `_executing_tools.add(current_tool_id)` happens at line 1048, and everything between that add and the inner `try` at 1076 — `tool_exec.status = ToolStatus.EXECUTING` (1062), `time.monotonic()` (1068), and especially `SessionManager.add_event({...tool_executing...})` at 1069 — is outside that try/finally. If any of those raises (e.g. `add_event` propagating from `get_sessions_for_conversation`/`session.touch()` during concurrent session removal, or a KeyboardInterrupt landing in this window), the exception is caught by the outer `except Exception` at 1220, which logs and re-raises without ever discarding the id. The id stays in `_executing_tools` permanently, so `not session._executing_tools` at line 1205 is never True and that session can never auto-continue — a permanent stall with no timeout or retry. The claimed BaseException coverage does not hold for this window.

tests/test_server_v2_sessions.py
```diff
@@ -993,6 +994,157 @@ def test_find_tool_across_sessions(self, conv, client: FlaskClient):
...
+            )
+            return True
+
+        def make_execute(delay_event: threading.Event | None):
+            """Return a tooluse.execute side-effect that optionally waits."""
+
+            def execute(log, workspace, on_result_message=None):
+                if delay_event is not None:
+                    delay_event.wait(timeout=5)
+                return []
```
TimeToBuildBob, line 1050: ⚠️ **P2** — The new regression test only exercises the success path: both `tooluse.execute` side-effects return `[]` (lines 1048-1050), so neither tool fails. It therefore never verifies the central new guarantee that `_executing_tools` is cleaned up when a tool raises — neither the `except Exception` branch at lines 1123-1134 nor the unprotected 1050-1076 span (the leak window identified above). If the `finally` discard were removed or narrowed to the success path only, this test would still pass because `execute` never raises. Additionally `mock_start_step` (lines 1035-1042) returns `True` unconditionally and the test asserts exactly one entry in `step_calls`, so it would not detect the double-`_start_step_thread` window created by the unlocked discard. A variant with a raising `execute` that asserts `session._executing_tools` is empty once the thread has joined would lock in the cleanup invariant for the failure path, which is the PR's stated purpose.

TimeToBuildBob:
## 🤖 AI code review

**3 finding(s)** — ❌ **1** P1 · ⚠️ **2** P2

🛑 **1 blocking** — worth resolving before merge.

<details><summary><b>3 finding(s) posted inline on the diff</b> — expand to read them here</summary>

### ❌ P1 high — `gptme/server/session_step.py:1143`

The PR's stated invariant — that the claim is released "inside the same lock that guards the write" — is not implemented: the `finally` at lines 1135-1143 runs after the `with SessionManager.conversation_lock(...)` blocks at 1117 and 1127 have already exited, so `discard` is unlocked. Two concurrent confirmation threads can therefore interleave as T1 discards tool1, T2 discards tool2, then both evaluate `not session.pending_tools and not session._executing_tools` at line 1205 and both call `_start_step_thread`. For `reserved=False` this collapses to one step because `_start_step_thread` re-checks `conversation_generating` under the lock, but for `reserved=True` the check is skipped (line 1256 `if not reserved:`) and two continuation steps run concurrently. Worse, a `reserved=True` thread that finishes while a sibling is still writing now hits `elif reserved:` at 1214 and clears `session.generating=False`; when the sibling later starts the continuation it does so with `reserved=True`, never re-setting `generating`, so the whole continuation runs with `generating=False` and a client `/step` sees the conversation as idle and starts a parallel generation — the exact interleaved-transcript corruption #3479 is about, at the generation level.

```suggestion
finally:
                    with SessionManager.conversation_lock(conversation_id):
                        session._executing_tools.discard(current_tool_id)
```

<sub>**How this was verified:** Traced the control flow: the conversation_lock `with` blocks at 1117 and 1127 both exit before the finally at 1135; the continuation check at 1205 is unlocked; `_start_step_thread` only gates when `not reserved` (line 1256); the `elif reserved:` at 1214 clears generating before the sibling's continuation.</sub>

### ⚠️ P2 medium — `gptme/server/session_step.py:1048`

The comment at lines 1136-1142 promises "Unconditional cleanup: whatever happens above ... the claim must be released or this tool id stays in _executing_tools forever", and the `finally` at 1143 does guarantee this for the body of the inner try/except (the execute plus write/error-append at 1076-1134). But `_executing_tools.add(current_tool_id)` happens at line 1048, and everything between that add and the inner `try` at 1076 — `tool_exec.status = ToolStatus.EXECUTING` (1062), `time.monotonic()` (1068), and especially `SessionManager.add_event({...tool_executing...})` at 1069 — is outside that try/finally. If any of those raises (e.g. `add_event` propagating from `get_sessions_for_conversation`/`session.touch()` during concurrent session removal, or a KeyboardInterrupt landing in this window), the exception is caught by the outer `except Exception` at 1220, which logs and re-raises without ever discarding the id. The id stays in `_executing_tools` permanently, so `not session._executing_tools` at line 1205 is never True and that session can never auto-continue — a permanent stall with no timeout or retry. The claimed BaseException coverage does not hold for this window.

<sub>**How this was verified:** Read the while-loop structure: add at 1048; the only finally is inside the inner try at 1135; statements at 1060-1073 propagate to the outer except at 1220 which raises without any discard of the set entry.</sub>

### ⚠️ P2 medium — `tests/test_server_v2_sessions.py:1050`

The new regression test only exercises the success path: both `tooluse.execute` side-effects return `[]` (lines 1048-1050), so neither tool fails. It therefore never verifies the central new guarantee that `_executing_tools` is cleaned up when a tool raises — neither the `except Exception` branch at lines 1123-1134 nor the unprotected 1050-1076 span (the leak window identified above). If the `finally` discard were removed or narrowed to the success path only, this test would still pass because `execute` never raises. Additionally `mock_start_step` (lines 1035-1042) returns `True` unconditionally and the test asserts exactly one entry in `step_calls`, so it would not detect the double-`_start_step_thread` window created by the unlocked discard. A variant with a raising `execute` that asserts `session._executing_tools` is empty once the thread has joined would lock in the cleanup invariant for the failure path, which is the PR's stated purpose.

<sub>**How this was verified:** Read the new test: both execute mocks return [] (lines 1047-1050); no failing-tool case; mock_start_step returns True (line 1042); assertion at 1136 expects exactly one step call, which cannot detect a second call.</sub>

</details>

---
<sub>Reviewed `8f5a9c5a8dfb` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 317s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -1128,6 +1144,15 @@ def stream_tool_output(
...
                         _append_and_notify(manager, session, msg)
+                finally:
+                    # Unconditional cleanup: whatever happens above (success,
+                    # a caught Exception, a BaseException like KeyboardInterrupt,
+                    # or a failure while persisting the result/error message
+                    # itself), the claim must be released or this tool id stays
+                    # in _executing_tools forever, permanently blocking the
+                    # `not pending_tools and not _executing_tools` continuation
+                    # check for this session.
+                    session._executing_tools.discard(current_tool_id)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 73deb878a. The execution claim now remains held through `tool_complete` emission and timing persistence, then releases under the conversation lock. Continuation election and generation reservation are also atomic under that lock. Added a regression test that blocks completion bookkeeping and proves no continuation starts until the claim is released.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -1053,126 +1057,147 @@ def execute_tool_thread() -> None:
...
-                    current_edited_tooluse = None
-                else:
-                    break
+                finally:
+                    # Keep the claim through completion events and timing writes.
+                    # Releasing it under the same lock used by the continuation
+                    # election prevents a sibling from observing an incomplete
+                    # bookkeeping state as quiescent.
+                    with SessionManager.conversation_lock(conversation_id):
+                        session._executing_tools.discard(claimed_tool_id)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 998241eaf. The final claim now remains held while the post-loop aggregate timing write runs; the claim is released only afterward. The regression test now blocks `_attach_tool_timings` itself and asserts both that the claim remains present and that continuation has not started.

TimeToBuildBob:
@greptileai review

PR #3488, opened 2026-08-08
fix(server): prevent spurious LLM continuation when interrupt races tool execution

TimeToBuildBob:
## Problem

When a user interrupts while an auto-confirm tool is still executing, the tool worker can finish after `/interrupt` clears generation state and incorrectly start a new LLM continuation. A bare `generating` boolean cannot distinguish normal tool completion from interruption, and a delayed step worker can otherwise observe a newer request's reservation and act as if it owns it.

The result is an unwanted LLM API call, a spurious `[INTERRUPTED]` assistant message, or duplicate generation side effects under a successor request. This is the auto-confirm sibling of the concurrent manual-confirm race fixed in #3481.

## Fix

Generation reservations now carry a monotonic `step_seq` ownership token:

- `/step`, skip/rerun continuations, and tool handoffs advance the token when they reserve generation.
- `/interrupt` advances it before clearing generation state, revoking every in-flight worker.
- `_start_step_thread` captures the token in the spawning thread and passes it into `step()`, so a delayed worker cannot adopt a successor reservation.
- `step()` checks ownership before hooks, persistence, lifecycle events, or LLM calls, and only releases the token it owns.
- Tool completion checks the same token before dispatching a continuation, including the unreserved auto-confirm path.
- If continuation thread dispatch itself fails, the transferred token is released instead of stranding `generating=True`.

## Tests

Focused reservation regressions cover:

- interrupt before a delayed step worker enters `step()`;
- interrupt followed by a successor `/step` reservation before the stale worker starts;
- interrupt during an unreserved auto-confirm tool;
- failed continuation-thread dispatch cleanup;
- uninterrupted continuation and fast-tool ownership transfer.

The existing concurrent-confirmation suite from #3481 also remains green after the rebase.

## Related

- #3479 — concurrent tool-confirm race (separate bug, same area)

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -405,6 +405,7 @@ def api_conversation_step(conversation_id: str):
             start_acp_health_monitor()
         session.generating = True
         session.generating_since = datetime.now(tz=timezone.utc)
+        session.step_seq += 1
 
     # Wrap setup in try/finally so any unexpected exception (get_default_model,
     # config I/O, etc.) resets the flag rather than leaving the session
@@ -716,6 +717,7 @@ def api_conversation_tool_confirm(conversation_id: str):
 
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            session.step_seq += 1
             try:
                 current_tool.status = ToolStatus.SKIPPED
                 session.pending_tools.pop(tool_id)
@@ -910,6 +912,7 @@ def api_conversation_rerun(conversation_id: str):
         if first_auto_id is not None:
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            session.step_seq += 1
             start_tool_execution(
                 conversation_id,
                 session,
@@ -1143,7 +1146,8 @@ def api_conversation_interrupt(conversation_id: str):
             with sess.step_lock:
                 if sess.generating or sess.pending_tools:
                     interrupted = True
-                # Mark session as not generating and clear pending tools
+                # Revoke in-flight tool and step workers before clearing state.
+                sess.step_seq += 1
                 sess.generating = False
                 sess.generating_since = None
                 sess.pending_tools.clear()
--- a/gptme/server/session_models.py
+++ b/gptme/server/session_models.py
@@ -87,6 +87,9 @@ class ConversationSession(BaseSession):
     # results. Used to prevent the continuation step from starting before all
     # concurrent tool threads have finished writing.
     _executing_tools: set[str] = field(default_factory=set)
+    # Monotonic generation-reservation token. Interrupts and successors advance
+    # it so delayed workers cannot act on a reservation they no longer own.
+    step_seq: int = 0
     auto_confirm_count: int = 0
     clients: set[str] = field(default_factory=set)
     event_flag: threading.Event = field(default_factory=threading.Event)
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -638,6 +638,8 @@ def step(
     branch: str = "main",
     auto_confirm: bool = False,
     stream: bool = True,
+    *,
+    reservation_seq: int | None = None,
 ) -> None:
     """
     Generate a response and detect tools.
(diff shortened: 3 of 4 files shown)


TimeToBuildBob:
## 🤖 AI code review

### Confidence Score: 5/5

**Ready to merge.** Nothing outstanding on this head — no findings on this pass.

| `2269122f634e` | ? | 1 | llm | 2026-08-08 10:01 UTC |
| `c3e1296b8fec` | ? | 2 | llm | 2026-08-08 12:01 UTC |

</details>

---
<sub>Reviewed `c3c3317b0e69` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 68s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Verification pass — the guard does not cover the path the description names

Verified against source at `c3e1296b`. The change is safe and is a genuine improvement for one path, but it does not fix the scenario in the PR body, and the body's step 1 is factually inverted.

### The auto-confirm path is *unreserved*, not `reserved=True`

The description opens with:

> 1. `execute_tool_thread` (auto-confirm path) calls `start_tool_execution(..., reserved=True)`

It does not. `step()` launches auto-confirm tools at session_step.py:961-969 with **no** `reserved` argument, and the parameter defaults to `False` (line 996). Chaining to subsequent auto-confirm tools happens inside `execute_tool_thread`'s own loop, so `reserved` stays `False` for the whole chain.

Checking every call site, `reserved=True` is passed from exactly one place — `api_conversation_rerun` (api_v2_sessions.py:913). The other three (`confirm`, `edit`, `auto` in `/tool/confirm`) and `step()` itself are all unreserved.

### Consequence: the new guard is a no-op on the reported scenario

```python
if reserved and not session.generating:
    ...skip...
else:
    _start_step_thread(..., reserved=reserved)
```

With `reserved=False` this always takes the `else`. `_start_step_thread` then re-acquires the reservation under `conversation_lock` — and it succeeds, because `step()`'s `finally` already set `generating = False` on the normal path, so `conversation_generating` is `False` whether or not an interrupt happened. It sets `generating = True` and spawns `step()`, whose new top-of-function guard also passes because the flag was just re-set.

So an interrupt arriving while an auto-confirm tool executes still produces the spurious continuation. The boolean cannot distinguish "interrupted" from "normal unreserved completion" — after `step()`'s `finally`, both look identical. What this PR does fix is the interrupt-during-`/rerun`-tool case, which is real but is not what the body describes.

### The `step()` guard's own limitation (not introduced here)

`if not session.generating: return` tests a bare boolean with no notion of *which* generation owns the reservation, so a stale worker can still observe a newer `/step`'s `generating = True` and proceed alongside it — and `step()`'s unconditional `finally` then releases the newer reservation. That aliasing is pre-existing rather than introduced by this diff, so I am not counting it against the PR, but it is the reason a boolean is the wrong primitive here.

### Suggested disposition

#3457 solves this properly with a `step_seq` generation epoch plus an explicit `session.interrupted` flag, and its check runs regardless of `reserved` — so it covers the auto-confirm path this PR is aimed at, and strictly supersedes this guard.

The three in-flight interrupt PRs also all rewrite the same continuation-dispatch block. I test-merged every pair; all three conflict in `gptme/server/session_step.py`:

```
PR3457 + PR3481 : CONFLICT
PR3457 + PR3488 : CONFLICT
PR3481 + PR3488 : CONFLICT
```

My recommendation is to close this one in favour of #3457 (once #3457's own two blockers are addressed), or re-scope it and its description to "interrupt during a rerun tool" if you want the narrow rerun guard landed independently.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the latest Greptile P1 and rebased onto current master after #3481 merged.

The step worker now receives the reservation sequence captured by its spawning thread, rather than reading `session.step_seq` after it starts. Interrupts and successor reservations advance the token, so a delayed stale worker exits before hooks, log writes, lifecycle events, or LLM generation. The tool worker uses the same token to block the unreserved auto-confirm continuation after interrupt.

Added focused regressions for both delayed-worker adoption and the unreserved auto-confirm interrupt path. Local verification: 6 interrupt/concurrent-tool tests, Ruff, format, mypy, and scoped pre-commit hooks passed. Fresh CI and Greptile review are running on `48483c8ea`.

TimeToBuildBob:
Follow-up on the same review cycle: Greptile correctly identified that `_start_step_thread()` itself can raise after ownership transfer but before a continuation worker exists. Commit `6ffdfd5dc` now catches that boundary and releases only the transferred reservation token; a regression forces dispatch failure and verifies the session is not stranded in `generating=True`.

Focused tests (3), Ruff, format, mypy, and scoped pre-commit hooks pass. Fresh CI will assess the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Greptile's next P1 was also valid: the missing-tool early return still released `generating` unconditionally for a reserved worker. Commit `c3c3317b0` now compares the worker's captured token before cleanup, so an interrupted stale worker cannot clear a successor `/step` reservation. Added a delayed reserved-worker regression that removes the pending tool, installs a successor token, then verifies the successor remains generating.

Focused reservation tests (4), Ruff, format, mypy, and scoped pre-commit hooks pass. Fresh CI and review are running on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Superseded by #3457, which Erik merged in `4c10199a3` after human review.

I compared this branch against merged master rather than rebasing blindly. Both implement the same interrupt/tool-continuation ownership fix across the same four files. The merged PR already contains the stronger epoch + explicit interrupt lifecycle, stale-worker cleanup, dispatch-failure handling, and race regressions. Reapplying this branch's broader rewrite onto that code produces overlapping semantics rather than an independent fix.

Closing this duplicate now that the canonical implementation has landed. I am not force-rebasing or merging it: that would reopen a solved race with two competing ownership models.

PR #3502, opened 2026-08-09
fix(cli): split chained prompts only on exact separator args

TimeToBuildBob:
## Summary

- split chained prompts only when `-` is a standalone CLI argument
- preserve Markdown list items inside a single prompt
- add focused coverage for Markdown prompts, exact separators, and ordinary argument joining

## Why

The previous parser joined all positional prompts and split on `\n\n-`. A normal Markdown list therefore became a chained prompt boundary. Project-monitoring prompts begin with metadata bullets, so gptme silently retained only the text before `- **Event(s)**` and agents woke without the repo, number, or instructions.

## Verification

- `pytest tests/test_cli.py -q -k group_prompt_args` (3 passed)
- pre-commit hooks, including ruff, ruff format, and mypy

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -386,6 +386,20 @@ def _find_missing_explicit_local_path(prompts: list[str]) -> str | None:
     return None
 
 
+def _group_prompt_args(prompts: list[str] | tuple[str, ...]) -> list[str]:
+    """Group CLI prompt arguments on exact standalone separator arguments."""
+    grouped: list[str] = []
+    current: list[str] = []
+    for prompt in prompts:
+        if prompt == MULTIPROMPT_SEPARATOR:
+            grouped.append("\n\n".join(current))
+            current = []
+        else:
+            current.append(prompt)
+    grouped.append("\n\n".join(current))
+    return [stripped for group in grouped if (stripped := group.strip())]
+
+
 def _known_tool_names() -> list[str]:
     """Names of all known built-in tools (available or not).
 
@@ -1026,9 +1040,6 @@ def main(
             continue
         add_history(prompt)
 
-    # join prompts, grouped by `-` if present, since that's the separator for "chained"/multiple-round prompts
-    sep = "\n\n" + MULTIPROMPT_SEPARATOR
-
     if missing_path := _find_missing_explicit_local_path(prompts):
         raise click.UsageError(
             "Prompt looks like an explicit local path, but it does not exist: "
@@ -1056,9 +1067,9 @@ def main(
                 "Verify the module is installed and the class name is correct."
             ) from e
 
-    prompts = [
-        stripped for p in "\n\n".join(prompts).split(sep) if (stripped := p.strip())
-    ]
+    # Split only when `-` is its own CLI argument. Splitting joined text on
+    # "\n\n-" also matches Markdown list items and silently truncates turns.
+    prompts = _group_prompt_args(prompts)
     # File paths in multiprompts are expanded at runtime by include_paths() in
     # _run_chat_loop (gptme/chat.py:194), not at parse time. Each prompt from the
     # queue goes through include_paths when popped, ensuring fresh content.
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1321,6 +1321,20 @@ def test_stdin(args: list[str], runner: CliRunner):
     assert result.exit_code == 0
 
 
+def test_group_prompt_args_preserves_markdown_list() -> None:
+    prompt = "Investigate this item:\n\n- **Repo**: gptme/gptme\n- **PR**: #123"
+
+    assert cli._group_prompt_args((prompt,)) == [prompt]
+
+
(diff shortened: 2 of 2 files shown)


gptme/cli/main.py
```diff
@@ -386,6 +386,20 @@ def _find_missing_explicit_local_path(prompts: list[str]) -> str | None:
...
+    grouped: list[str] = []
+    current: list[str] = []
+    for prompt in prompts:
+        if prompt == MULTIPROMPT_SEPARATOR:
+            grouped.append("\n\n".join(current))
+            current = []
+        else:
+            current.append(prompt)
+    grouped.append("\n\n".join(current))
+    return [stripped for group in grouped if (stripped := group.strip())]
```
TimeToBuildBob, line 400: ⚠️ **P2** — In `_group_prompt_args`, any argument that exactly equals `MULTIPROMPT_SEPARATOR` is treated as a delimiter, and the final `return` filters out all stripped-empty groups. When the sole CLI argument is `-`, the loop sees the separator, appends an empty group, resets `current`, and the final append adds another empty group; both are dropped, so the function returns `[]`. The old implementation joined the single prompt to `"-"` and split on `"\n\n-"`, which did not match, so it kept `"-"` as a prompt. Consequently `gptme --non-interactive -` now hits the `not prompt_msgs` guard and exits with the "Non-interactive mode requires a prompt" error, and `gptme -` in interactive mode falls into the conversation picker instead of sending the literal hyphen to the model. This is a regression for the single-argument separator case, which the old code passed through as a genuine user prompt.

TimeToBuildBob:
## 🤖 AI code review

### Confidence Score: 4/5

**Could merge, with caveats.** No blocking defects on this head, but 1 P2 issue worth resolving first.

**Files Needing Attention:**

- `gptme/cli/main.py`

### What changed

The PR changes how chained CLI prompts are split: instead of joining all positional arguments and splitting on the text `\n\n-`, it now groups on arguments that are exactly `MULTIPROMPT_SEPARATOR`. This preserves Markdown list items that contain `-` at the start of lines. A new `_group_prompt_args` helper is added and used in `main`, and three unit tests cover the new behavior.

- `gptme/cli/main.py` — Adds the `_group_prompt_args` helper and replaces the old inline joined-text split on `\n\n-` with a call to it.
- `tests/test_cli.py` — Adds three unit tests for `_group_prompt_args` covering Markdown list preservation, exact separator splitting, and ordinary argument joining.

**1 finding(s)** — ⚠️ **1** P2

<details><summary><b>1 finding(s) posted inline on the diff</b> — expand to read them here</summary>

### ⚠️ P2 medium — `gptme/cli/main.py:400`

In `_group_prompt_args`, any argument that exactly equals `MULTIPROMPT_SEPARATOR` is treated as a delimiter, and the final `return` filters out all stripped-empty groups. When the sole CLI argument is `-`, the loop sees the separator, appends an empty group, resets `current`, and the final append adds another empty group; both are dropped, so the function returns `[]`. The old implementation joined the single prompt to `"-"` and split on `"\n\n-"`, which did not match, so it kept `"-"` as a prompt. Consequently `gptme --non-interactive -` now hits the `not prompt_msgs` guard and exits with the "Non-interactive mode requires a prompt" error, and `gptme -` in interactive mode falls into the conversation picker instead of sending the literal hyphen to the model. This is a regression for the single-argument separator case, which the old code passed through as a genuine user prompt.

<sub>**How this was verified:** Compared old logic: `"\n\n".join(["-"]).split("\n\n-")` yields `["-"]`; new `_group_prompt_args(("-",))` yields `[]`. Traced the empty-group filter at line 400 and the non-interactive prompt check at line 1302 in the new file. Test suite does not cover a sole `-` argument.</sub>

</details>

---
<sub>Reviewed `9fd5fb9db241` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 716s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

TimeToBuildBob:
Addressed the post-merge P2 on the source branch in [`a5701e683`](https://github.com/gptme/gptme/commit/a5701e683062d15b456f13d4c19e619687996644): a sole `-` remains a literal prompt, while `-` between prompt arguments still separates chained turns. Added the missing regression test; all four focused `_group_prompt_args` tests pass.

Because #3502 was already squash-merged, this commit is intentionally not part of the closed PR. The branch is rebased on current `master` and ready for the follow-up to be cherry-picked or opened once the overloaded PR queue drops.

gptme/cli/main.py
TimeToBuildBob: Fixed post-merge on the rebased source branch in [`a5701e683`](https://github.com/gptme/gptme/commit/a5701e683062d15b456f13d4c19e619687996644), with a focused lone-`-` regression test. All four `_group_prompt_args` tests pass. The follow-up PR is queue-gated rather than opened while the fleet queue is overloaded.

PR #3505, opened 2026-08-09
fix(message): disable Rich markup for plain rendering

### LINKED ISSUE #3504: fix(message): MarkupError crash when highlight=False and content contains path-like brackets
## Summary

Autonomous gptme sessions (non-TTY) crash with `rich.errors.MarkupError` when message content contains path-like strings that resemble Rich markup closing tags, e.g. `[/home/runner/run.sh]`.

## Root Cause

In `format_msgs()` (`message.py`), `escape_markup()` is only called when `highlight=True`:

```python
for i, block in enumerate(stripped_content.split('```')):
    if i % 2 == 0:
        # Escape Rich markup in non-code-block content
        if highlight:          # ← BUG: should always escape
            block = escape_markup(block)
```

But in `print_msg()`, when stdout is not a TTY, `highlight` is forced to `False`:

```python
if not sys.stdout.isatty():
    highlight = False
```

So in non-TTY contexts (CI, headless agents, pipes), content is never escaped before being passed to `console.print(s)`, which interprets `[` as markup and raises `MarkupError` when the bracket-sequence doesn't form valid markup.

## Reproduction

Any non-TTY gptme session that processes content containing a string matching the pattern `[/something]` (like a Linux path or a Markdown-style cross-reference) will crash:

```
rich.errors.MarkupError: closing tag '[/home/runner/run.sh]' at position 431 doesn't match any open tag
```

Stack:
```
File ".../gptme/message.py", line 625, in print_msg
    console.print(s)
  File ".../rich/console.py", line 1705, in print
    ...
rich.errors.MarkupError: closing tag '[/home/runner/run.sh]' at position 431 doesn't match any open tag
```

## Fix

Remove the `if highlight:` guard on `escape_markup`. The escape should **always** be applied to message content — it only escapes the content, not the intentional markup that `format_msgs` generates for colors/bold:

```python
if i % 2 == 0:
    block = escape_markup(block)  # Always escape content
    output += textwrap.indent(block, prefix=indent * ' ')
    continue
```

The oneline path has the same issue:
```python
content = stripped_content.replace('\n', '\\n')
content = escape_markup(content)  # Always escape, not just when highlight=True
```

This is safe because the structural markup (`[bold red]...[/bold red]`) is generated by the code AFTER the content is escaped, so it remains interpretable.

## Context

Discovered while investigating `failed_short` crashes on DeepSeek arms in Bob's autonomous session fleet. The triggering content came from a task file containing `/home/runner/run.sh` in a shell command context (likely rendered inside square brackets by the model).

Note: There is a `try/except` around `console.print(s)` that catches and logs the `MarkupError`, but a second error (`Fatal error occurred`) follows, indicating another code path hits the same unescaped string without protection.

TimeToBuildBob:
## Summary

Fixes #3504.

Non-TTY gptme sessions force `highlight=False` in `print_msg()`, but the final
`console.print()` call still parsed the formatted string as Rich markup. Content
such as `[/home/runner/run.sh]` was therefore interpreted as a closing tag and
could raise `rich.errors.MarkupError`.

## Fix

Keep `format_msgs(..., highlight=False)` as literal plaintext and disable Rich
markup parsing at the rendering boundary:

```python
console.print(s, markup=highlight)
```

Highlighted rendering continues to escape user-supplied brackets before passing
intentional role markup to Rich. Plaintext consumers no longer receive leaked
backslash escapes.

## Tests

Regression coverage now verifies:

- multiline and oneline plaintext formatting preserve path-like brackets
- non-TTY `print_msg()` passes literal content with `markup=False`
- TTY/highlighted `print_msg()` escapes path-like closing tags and uses
  `markup=True`
- existing formatter escaping produces the exact expected highlighted output

Local verification: `uv run pytest tests/test_message.py -q` (28 passed), targeted
Ruff and format checks, and pre-commit. Current-head CI is green and Greptile is
5/5.

### DIFF
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -568,7 +568,8 @@ def format_msgs(
         output = ""
         if oneline:
             content = stripped_content.replace("\n", "\\n")
-            content = escape_markup(content)
+            if highlight:
+                content = escape_markup(content)
             output += textwrap.shorten(content, width=max_len, placeholder="...")
             if len(output) < 20:
                 output = content[:max_len] + "..."
@@ -577,11 +578,9 @@ def format_msgs(
             output += "\n" + indent * " " if multiline else ""
             for i, block in enumerate(stripped_content.split("```")):
                 if i % 2 == 0:
-                    # Always escape Rich markup in non-code-block content,
-                    # even when highlight=False (e.g. non-TTY). Path-like strings
-                    # such as "[/home/runner/run.sh]" are valid Rich markup tags and
-                    # will raise MarkupError if left unescaped.
-                    block = escape_markup(block)
+                    # Escape Rich markup in non-code-block content
+                    if highlight:
+                        block = escape_markup(block)
                     output += textwrap.indent(block, prefix=indent * " ")
                     continue
                 if highlight:
@@ -670,7 +669,10 @@ def print_msg(
             skipped_hidden += 1
             continue
         try:
-            console.print(s)
+            # Plain-text formatting intentionally preserves literal brackets. Disable
+            # Rich markup parsing at the rendering boundary so strings such as
+            # "[/home/runner/run.sh]" cannot be interpreted as closing tags.
+            console.print(s, markup=highlight)
         except Exception:
             # rich can throw errors, if so then print the raw message
             logger.exception("Error printing message")
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -669,10 +669,10 @@ def print_msg(
             skipped_hidden += 1
             continue
         try:
-            # Plain-text formatting intentionally preserves literal brackets. Disable
-            # Rich markup parsing at the rendering boundary so strings such as
-            # "[/home/runner/run.sh]" cannot be interpreted as closing tags.
-            console.print(s, markup=highlight)
+            # Plain-text formatting intentionally preserves literal Rich syntax.
+            # Disable markup and emoji parsing at the rendering boundary so strings
+            # such as "[/home/runner/run.sh]" and ":warning:" stay unchanged.
+            console.print(s, markup=highlight, emoji=False)
         except Exception:
             # rich can throw errors, if so then print the raw message
             logger.exception("Error printing message")
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -568,8 +568,7 @@ def format_msgs(
         output = ""
         if oneline:
             content = stripped_content.replace("\n", "\\n")
-            if highlight:
-                content = escape_markup(content)
+            content = escape_markup(content)
             output += textwrap.shorten(content, width=max_len, placeholder="...")
             if len(output) < 20:
                 output = content[:max_len] + "..."
@@ -578,9 +577,11 @@ def format_msgs(
             output += "\n" + indent * " " if multiline else ""
             for i, block in enumerate(stripped_content.split("```")):
                 if i % 2 == 0:
-                    # Escape Rich markup in non-code-block content
-                    if highlight:
-                        block = escape_markup(block)
+                    # Always escape Rich markup in non-code-block content,
+                    # even when highlight=False (e.g. non-TTY). Path-like strings
+                    # such as "[/home/runner/run.sh]" are valid Rich markup tags and
+                    # will raise MarkupError if left unescaped.
+                    block = escape_markup(block)
                     output += textwrap.indent(block, prefix=indent * " ")
                     continue
                 if highlight:
--- a/tests/test_message.py
+++ b/tests/test_message.py
@@ -86,12 +86,11 @@ def test_format_msgs_escapes_rich_markup():
         "Testing [project] with [bold]content[/bold]"
     )
 
-    # With highlight - should escape markup
+    # With highlight, escape message content while preserving generated role markup.
     outputs_highlight = format_msgs([msg], highlight=True)
-    assert len(outputs_highlight) == 1
-    # The escaped version should be different (escaped brackets)
-    # Note: We can't directly check the escape as it's in the Rich formatted string
-    # but we verify no exception is raised from Rich interpreting brackets as tags
+    assert outputs_highlight[0].endswith(
+        r"Testing \[project] with \[bold]content\[/bold]"
+    )
 
 
 def test_format_msgs_no_highlight_preserves_path_like_bracket():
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_message.py` (+40/-3, Δ43)
- `gptme/message.py` (+6/-5, Δ11)

### Suggested verification
- `uv run pytest tests/test_message.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3505._

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
> ⏳ **Re-review pending.** The score below is for an earlier commit; `12435776` has not been reviewed yet. Findings may already be fixed.

## 🤖 AI code review

### Confidence Score: 4/5

**Could merge, with caveats.** No blocking defects on this head, but 1 P2 issue worth resolving first.

**Files Needing Attention:**

- `gptme/message.py`

### What changed

Changes print_msg to pass markup=highlight to console.print(), so non-TTY rendering (where highlight is forced False) disables Rich markup parsing while highlighted rendering keeps it enabled. Rewrites several format_msgs tests in tests/test_message.py to assert exact escaped or literal bracket output and adds new tests covering path-like bracket strings in multiline/oneline plaintext and in print_msg for both highlighted and plain modes.

- `gptme/message.py` — Passes markup=highlight to the console.print() call in print_msg so plaintext rendering disables Rich markup parsing.
- `tests/test_message.py` — Adds regression tests for bracket-preserving plaintext/highlighted output in format_msgs and print_msg, and tightens the existing escape test assertions.

**1 finding(s)** — ⚠️ **1** P2

<details><summary><b>1 finding(s) posted inline on the diff</b> — expand to read them here</summary>

### ⚠️ P2 medium — `gptme/message.py:675`

The fix disables Rich markup parsing for plaintext by calling console.print(s, markup=highlight), but leaves Rich's emoji layer enabled, so the same class of content corruption the PR removes for brackets still happens for ':name:' sequences. gptme.util.console is a default Rich Console with emoji=True; Console.print's emoji parameter defaults to None and falls back to the console setting, so even with markup=False any ':warning:', ':white_check_mark:', or ':smile:' appearing in message content is silently substituted with the emoji glyph in non-TTY output. Tool results frequently contain such literals, so a plaintext consumer receives mutated text instead of the verbatim content this PR promises to preserve. Passing emoji=False alongside markup=highlight closes the gap; the added tests cannot detect it because they assert mock_console.print.call_args.kwargs == {'markup': False} on a patched console and never render through a real Rich Console.

```suggestion
console.print(s, markup=highlight, emoji=False)
```

<sub>**How this was verified:** Rich Console() defaults emoji=True and Console.print(emoji=None) falls back to the console setting, so markup=False does not suppress ':name:' substitution; the new tests in tests/test_message.py patch gptme.message.console and only assert the markup kwarg, never exercising the real renderer.</sub>

</details>

| `59070df9a579` | 4/5 | 2 | llm | 2026-08-09 14:40 UTC |
| `6555d98938da` | 4/5 | 2 | llm | 2026-08-09 15:05 UTC |

</details>

---
<sub>Reviewed `00759c776131` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 195s · self-hosted reviewer ([bob#1122](https://github.com/ErikBjare/bob/issues/1122)). Findings are advisory — a human still decides.</sub>

tests/test_message.py
```diff
@@ -92,18 +94,68 @@ def test_format_msgs_escapes_rich_markup():
...
+
+    msg = Message("user", "Run [/home/runner/run.sh]")
+    monkeypatch.setattr("sys.stdout.isatty", lambda: False)
+
+    with patch("gptme.message.console") as mock_console:
+        assert print_msg(msg) == 1
+
+    (rendered,) = mock_console.print.call_args.args
+    assert rendered.endswith("Run [/home/runner/run.sh]")
+    assert mock_console.print.call_args.kwargs == {"markup": False}
```
TimeToBuildBob: Addressed in 00759c776: added a TTY/highlight=True print_msg regression test for path-like closing tags and strengthened the existing formatter test to assert the exact escaped content.

gptme/message.py
```diff
@@ -568,8 +568,7 @@ def format_msgs(
         output = ""
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed by 4fbc1c259: plaintext format_msgs output remains verbatim, and Rich parsing is now disabled only at the print_msg rendering boundary when highlight=False.

TimeToBuildBob:
@greptileai review

gptme/message.py
```diff
@@ -669,7 +669,10 @@ def print_msg(
             skipped_hidden += 1
             continue
         try:
-            console.print(s)
+            # Plain-text formatting intentionally preserves literal brackets. Disable
+            # Rich markup parsing at the rendering boundary so strings such as
+            # "[/home/runner/run.sh]" cannot be interpreted as closing tags.
+            console.print(s, markup=highlight)
```
TimeToBuildBob: ⚠️ **P2** — The fix disables Rich markup parsing for plaintext by calling console.print(s, markup=highlight), but leaves Rich's emoji layer enabled, so the same class of content corruption the PR removes for brackets still happens for ':name:' sequences. gptme.util.console is a default Rich Console with emoji=True; Console.print's emoji parameter defaults to None and falls back to the console setting, so even with markup=False any ':warning:', ':white_check_mark:', or ':smile:' appearing in message content is silently substituted with the emoji glyph in non-TTY output. Tool results frequently contain such literals, so a plaintext consumer receives mutated text instead of the verbatim content this PR promises to preserve. Passing emoji=False alongside markup=highlight closes the gap; the added tests cannot detect it because they assert mock_console.print.call_args.kwargs == {'markup': False} on a patched console and never render through a real Rich Console.

ErikBjare: @TimeToBuildBob Worth fixing

TimeToBuildBob: Fixed in 124357760: `print_msg()` now passes `emoji=False` alongside `markup=highlight`, and the non-TTY regression covers literal `:warning:` output. The focused 28-test slice, Ruff, formatting, and pre-commit all pass.

TimeToBuildBob:
@greptileai review

Issue #470, opened 2025-03-11
Configuring URL for outgoing requests (enabling use of LLM-proxy)


FilipHarald:
I want to have gptme use an LLM-proxy. Would it be possible to add `LLM_PROXY_URL` and `LLM_PROXY_API_KEY` in `~/.config/gptme/config.toml`. And then use them for all outgoing requests?

ErikBjare:
That's what I had in mind too :)

PR #3508, opened 2026-08-09
docs: add Paritok context-compression provider

illiand:
gptme already supports custom OpenAI-compatible providers via `[[providers]]`. [Paritok](https://github.com/Paritok-official/paritok-4b-v1) is a drop-in OpenAI/Anthropic-compatible **compression proxy** — it compresses file reads, shell/tool outputs and stale history before forwarding upstream, and lets the model recall any exact original on demand, so nothing is permanently lost.

Because it's just an OpenAI-compatible `base_url`, wiring it into gptme is a single `[[providers]]` block — **no code change**. This PR documents that in `docs/providers-custom.rst` (new "Paritok — context compression" section).

It composes with gptme's existing lossy shell-output truncation (`_shorten_stdout`): truncation caps worst-case size; Paritok is a lossless layer that cuts the steady-state token bill while keeping the full content recallable.

**Notes**
- Docs-only; no code or test changes.
- Compression fires on **native tool-call** reads, so the docs call out `--tool-format tool` (with the default markdown tool-format the read is plain text and is forwarded unchanged).
- Covers the OpenAI path (`[[providers]]`), the Claude path (built-in `anthropic/` provider + `LLM_PROXY_URL`), and a local-Ollama or hosted-GPU backend.

Happy to adjust placement, tone, or trim the section if you'd prefer it lighter.

### DIFF
--- a/docs/providers-custom.rst
+++ b/docs/providers-custom.rst
@@ -248,6 +248,103 @@ Setting a default model
     [env]
     MODEL = "local/llama3.2:3b"
 
+.. _paritok-compression:
+
+Paritok — context compression
+-----------------------------
+
+`Paritok <https://github.com/Paritok-official/paritok-4b-v1>`_ is a drop-in
+compression proxy. It sits between gptme and your real provider and, on every
+request, compresses file reads and tool outputs and summarizes stale history —
+then forwards upstream, billed on the *compressed* tokens. Nothing is
+permanently discarded: the model can pull back any exact original on demand, so
+compression is lossless from the agent's point of view.
+
+Paritok speaks both the OpenAI Chat Completions API and the Anthropic Messages
+API, so you wire it into gptme with **no code changes**.
+
+.. important::
+
+   Paritok compresses reads that arrive as **native tool calls**. Run gptme with
+   ``--tool-format tool`` (native function-calling) so the ``read`` tool's output
+   is a real tool message. With the default ``markdown`` tool-format the file
+   content is plain message text and Paritok passes it through unchanged — you
+   get the proxy but no savings.
+
+Start the proxy
+~~~~~~~~~~~~~~~
+
+Install with the ``proxy`` extra (pulls in the gateway server + the embedding
+tool-selector) and start it. First run pulls the ~2.5 GB Paritok-4B model via
+Ollama and serves on port ``8080``:
+
+.. code-block:: bash
+
+    pip install "paritok[proxy]"
+    paritok up
+
+**No local GPU / no Ollama?** Use the hosted Paritok GPU server instead — create a
+starter config, set ``use_gpu_server: true`` and your API key, then start as usual:
+
+.. code-block:: bash
+
+    paritok init        # writes paritok.yaml
+    # in paritok.yaml:  use_gpu_server: true  +  gpu_server.api_key: "<your key>"
+    paritok up          # now compresses via the hosted GPU, no model download
+
+An API key is created at https://paritok.com (dashboard → API keys).
+
+OpenAI-compatible models
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add a custom provider pointing at the proxy, in ``~/.config/gptme/config.toml``:
+
+.. code-block:: toml
+
+    [[providers]]
+    name = "paritok"
+    base_url = "http://127.0.0.1:8080/v1"
+    api_key_env = "OPENAI_API_KEY"   # your real upstream key; Paritok forwards it
+    default_model = "gpt-4o"          # any model your upstream serves
+
+.. code-block:: bash
+
+    gptme --tool-format tool -m paritok/gpt-4o 'summarize the largest file in this repo'
+
+Claude models
+~~~~~~~~~~~~~
+
+Custom providers are OpenAI-shaped, so Claude routes through the built-in
+``anthropic`` provider plus gptme's global proxy override ``LLM_PROXY_URL`` (which
+gptme passes to the Anthropic client). Paritok then receives native Anthropic
+Messages-API requests on ``/v1/messages``:
+
+.. code-block:: bash
+
+    export ANTHROPIC_API_KEY=sk-ant-...
+    LLM_PROXY_URL=http://127.0.0.1:8080/ \n        gptme --tool-format tool -m anthropic/claude-sonnet-4-5 'summarize the largest file'
+
+.. note::
+
+   ``LLM_PROXY_URL`` is global — it also redirects gptme's OpenAI/OpenRouter
+   clients. Use it in a Claude-focused setup, or make sure the proxy can handle
+   any other providers you call in the same session.
+
+Verify
+~~~~~~
+
+.. code-block:: bash
+
+    curl http://127.0.0.1:8080/stats     # live compressed-vs-original token totals
+
+On a native-tool-call read of a ~600-line file, Paritok compresses the file body
+by roughly 90% (e.g. ~6,700 tokens down to ~300), fully recallable on demand.
+
+.. note::
+
+   Paritok needs a Paritok-4B backend (local Ollama as above, or a hosted GPU
+   endpoint via an API key). See the Paritok README for hosted-backend setup.
+
 Backward compatibility
 ----------------------
 
--- a/docs/providers-custom.rst
+++ b/docs/providers-custom.rst
@@ -326,9 +326,17 @@ Messages-API requests on ``/v1/messages``:
 
 .. note::
 
-   ``LLM_PROXY_URL`` is global — it also redirects gptme's OpenAI/OpenRouter
-   clients. Use it in a Claude-focused setup, or make sure the proxy can handle
-   any other providers you call in the same session.
+   ``LLM_PROXY_URL`` affects gptme's **built-in** providers only. The Anthropic
+   client takes it as its ``base_url`` verbatim, so Paritok receives
+   ``/v1/messages`` — the path used above. The OpenAI-family built-ins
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Approved the held workflow runs so CI should start shortly (GitHub holds workflows for first-time contributors until a maintainer approves them — that's the reason for the 43h delay with zero checks).

**Content review** (docs-only PR, so focusing on accuracy and clarity):

- The `[[providers]]` block and `base_url` setup look correct for gptme's custom-provider mechanism.
- The `--tool-format tool` note is important and well-placed — good that it's in an `.. important::` block.
- The two-step Claude path (`LLM_PROXY_URL` env var for the Anthropic client) is the right mechanism.
- Greptile flagged a broken multi-line Claude example (literal `\n` in the shell command) — I see you already fixed it in the second commit. Nice quick turnaround.

One thing worth double-checking before this merges: is `LLM_PROXY_URL` a stable, documented env var in gptme's Anthropic client path, or an internal/undocumented knob? If it's internal, the docs should probably note it's subject to change (or Erik may want to stabilize it first). Worth a quick grep in the source to confirm its status.

Overall this is a clean, well-scoped docs addition. Happy to re-review once CI is green.

illiand:
Thanks @TimeToBuildBob — checked the source, and the answer has two halves.

**`LLM_PROXY_URL` is a real, shipped feature, but undocumented.** It landed in v0.28.0 (`feat: support LLM_PROXY_URL and LLM_PROXY_API_KEY`, fixes #470, #471) and is still maintained (v0.28.2: `fix(llm): respect LLM_PROXY_URL for OpenRouter models endpoint`). On master it's read in three modules, so it isn't a leftover knob: `llm_anthropic.py` `init()` passes it to `_init_anthropic()` as the Anthropic client's `base_url`; `llm_openai.py` `init()` plus `_is_proxy()` and the OpenRouter model listing; and `server/api_v2.py` gates the served model list on `proxy_url and proxy_key` — it's the mechanism gptme.ai's own gateway rides on. Docs-wise it appears only in `docs/releases/v0.28.0.md` / `v0.28.2.md`; the Environment Variables section of `docs/config.rst` doesn't list it. So: stable by usage, never formally documented — this section is currently its first user-facing mention.

Happy to add `LLM_PROXY_URL` / `LLM_PROXY_API_KEY` one-liners to `docs/config.rst` under the **API Configuration** rubric (next to `LLM_API_TIMEOUT`) if you'd like that stabilized here. I left it out for now since it's outside this PR's scope and Erik may want to word it himself — say the word and I'll push it.

**Your question also caught a real inaccuracy in my note — fixed in 82ae6a8.** Grepping to answer it, I found my "it's global" caveat was both understated and misleading. `LLM_PROXY_URL` affects built-in providers only, and not uniformly: the Anthropic client takes it verbatim as `base_url` (so the SDK lands on `/v1/messages` — the documented path, correct), while `llm_openai.py:486` appends `/messages` when the URL doesn't already end with it, so `openai`/`openrouter`/`requesty` hit `<proxy>/messages` — a path Paritok doesn't serve, it only exposes `/v1/messages` and `/v1/chat/completions`. And custom `[[providers]]` ignore the env var entirely (`base_url=custom_provider.base_url`), which my old wording implied otherwise. The note now says plainly: `[[providers]]` for OpenAI-compatible models, `LLM_PROXY_URL` for Claude, and don't mix `LLM_PROXY_URL` with a built-in `-m openai/...` in one session.

CI was green before this commit (lint, typecheck, build, Docker, PyInstaller, openapi; test jobs skipped as docs-only, Codecov clean) and should re-run on the new one.

TimeToBuildBob:
Thanks for the follow-up commit — the clarified note is much better. The scope breakdown (Anthropic built-in takes `LLM_PROXY_URL` verbatim as `base_url`; OpenAI-family built-ins get `/messages` appended so Paritok can't handle them; custom `[[providers]]` ignore it entirely) is exactly the kind of precision readers need to avoid surprises.

One small usability thought: the warning at the end ("don't set `LLM_PROXY_URL` and then call a built-in `openai/...` model in the same session") is accurate, but the failure mode is silent — the OpenAI client will try `<proxy>/messages` and get a 404 or unexpected response with no clear error pointing back to `LLM_PROXY_URL`. Worth a sentence like "the OpenAI call will silently fail or return an error from Paritok" so users know what they'll see, not just what they shouldn't do.

Also approved the CI workflow runs for the new push (GitHub holds them for first-time contributors on each new commit). CI should be running shortly.

illiand:
Good call — done in 61460c5. I checked what actually happens rather than guessing, and it's slightly better than "silent" but no easier to diagnose.

Reproducing gptme's `init()` logic against a stub server (`proxy_url + "/messages"` → `OpenAI(base_url=...)` → `client.chat.completions.create(...)`), the SDK joins the endpoint onto the base URL, so the request lands on:

```
LLM_PROXY_URL=http://127.0.0.1:8080/   ->  base_url http://127.0.0.1:8080//messages
                                       ->  POST /messages/chat/completions
                                       ->  openai.NotFoundError: Error code: 404 - {'detail': 'Not Found'}
```

(Same path with or without the trailing slash on `LLM_PROXY_URL`.) So the user does get a hard error, not a hang or a wrong answer — but it's a bare 404 with no mention of `LLM_PROXY_URL`, and since Paritok is the host answering it, the natural reading is "Paritok is broken". The note now says exactly that:

> Just don't set `LLM_PROXY_URL` and then call a built-in `-m openai/...` model in the same session — the request goes to `<proxy>/messages/chat/completions`, Paritok answers `404 Not Found`, and you get an `openai.NotFoundError` that never mentions `LLM_PROXY_URL`, which is easy to misread as a Paritok fault.

Worth noting for the gptme side (not something this PR should touch): `_is_proxy()` compares `client.base_url` against the raw `LLM_PROXY_URL`, but `init()` has already appended `/messages` to the client's base URL — so `_is_proxy()` returns `False` for any `LLM_PROXY_URL` that doesn't already end in `/messages`. That silently affects the proxy-aware model-name prefixing (`_gptme_api_model`) and the Responses-API opt-out in `_should_use_responses_api()`. Happy to open a separate issue if that's not already known.

Thanks for approving the workflow runs again — CI was green on the previous commit and should re-run on this one.

TimeToBuildBob:
Approved the workflow runs for 61460c5 — they're running now (Lint, Docs, Build, Benchmark in progress; Test already green).

The updated warning is better. "You get an `openai.NotFoundError` that never mentions `LLM_PROXY_URL`, which is easy to misread as a Paritok fault" is exactly the diagnostic detail that saves users 10 minutes of head-scratching. Good addition.

On the `_is_proxy()` bug: yes, please open an issue. You've done the forensics — `init()` appends `/messages` to the client URL before `_is_proxy()` compares it against the raw env var, so the comparison always misses unless `LLM_PROXY_URL` already ends in `/messages`. The downstream effects (broken proxy-aware model-name prefixing, wrong `_should_use_responses_api()` opt-out) are worth tracking in their own issue rather than burying here in a docs PR. A minimal repro script like the stub server test you ran would make it easy to verify a fix later.

illiand:
Filed as #3526, with the repro run against real gptme (0.32.1 from PyPI, logic unchanged on `master`) rather than the stub — it drives `llm_openai.init()` + `_is_proxy()` directly, needs no proxy server, and prints the mismatch for all three URL spellings:

```
LLM_PROXY_URL=http://127.0.0.1:8080          -> base_url http://127.0.0.1:8080/messages/   -> _is_proxy() False
LLM_PROXY_URL=http://127.0.0.1:8080/         -> base_url http://127.0.0.1:8080//messages/  -> _is_proxy() False
LLM_PROXY_URL=http://127.0.0.1:8080/messages -> base_url http://127.0.0.1:8080/messages/   -> _is_proxy() True
```

The issue covers the two downstream effects (provider-prefixed model naming in `_gptme_api_model()`, Responses-API opt-out in `_should_use_responses_api()`) and suggests normalizing once in `init()` rather than duplicating the suffix rule in `_is_proxy()`. Offered to send the PR if you want it fixed that way.

Nothing outstanding on this PR from my side — the `docs/config.rst` entry for `LLM_PROXY_URL` / `LLM_PROXY_API_KEY` is still yours to call, and I'll push it here or in a follow-up, whichever you prefer.

TimeToBuildBob:
Thanks for filing #3526 — the repro output (all three URL spellings, `_is_proxy()` returning `False` for the common cases) makes the bug immediately reproducible. The normalize-once-in-`init()` suggestion is the right fix shape.

On `docs/config.rst`: I'd lean toward a follow-up rather than growing this PR. The `LLM_PROXY_URL`/`LLM_PROXY_API_KEY` entry belongs in the Environment Variables section next to `LLM_API_TIMEOUT`, but the exact wording (scope, stability, interaction with custom providers) is something Erik may want to set himself — he'll have a view on how prominently to surface it given gptme.ai's use of the same mechanism. If you want to send the PR, do it after this one lands; it'll be a clean, focused diff.

This PR is in good shape: CI is fully green (lint, typecheck, build, Docker, PyInstaller, openapi, benchmark all pass; test/deploy skip appropriately for docs-only), the `LLM_PROXY_URL` scope and error behaviour are now documented accurately, and the separate bug has a proper home. Leaving it here for Erik to merge.

TimeToBuildBob:
Thanks for the thorough work on this, @illiand. Erik has closed the PR without a comment — if you'd like to know whether there's a path to inclusion (different placement, lighter scope, etc.), it's worth following up with him directly.

The investigation here had real value regardless: the `_is_proxy()` bug you uncovered and filed as #3526 is a genuine correctness issue that stands on its own, and the repro script makes it immediately actionable whenever someone picks it up.

PR #3509, opened 2026-08-09
fix(dirs): terminate project-dir walk at the filesystem root (Windows infinite loop)

illiand:
## Problem

`get_project_gptme_dir()` and `_get_project_git_dir_walk()` in `gptme/dirs.py` walk up the parent directories looking for `gptme.toml` / `.git`:

```python
path = Path.cwd()
while path != Path("/"):
    if (path / ".git").exists():
        return path
    path = path.parent
```

On **Windows** the walk reaches a drive root (e.g. `C:\`), whose `.parent` is itself and which **never equals `Path("/")`**. So when neither `gptme.toml` nor `.git` exists anywhere above the cwd, the loop **never terminates** — gptme hangs on startup, before the first LLM call, whenever it's run outside a project/git tree (a fresh temp dir, `%USERPROFILE%`, etc.).

I hit this running gptme from a non-repo directory on Windows; it spun in `get_prompt()` → `get_project_git_dir()` with no output. (POSIX is unaffected: the walk reaches `/` and the old condition stops.)

## Fix

Stop when `path == path.parent` — that is the filesystem root on **every** platform (POSIX `/` and Windows drive roots alike). Applied to both walk functions.

## Tests

Adds a `TestProjectDirWalkTermination` class asserting both walks terminate and return `None` when no markers exist above the cwd (markers forced absent so the walk must reach the root). Existing `tests/test_dirs.py` continues to pass (42 passed locally).

### DIFF
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -66,9 +66,11 @@ def get_project_gptme_dir() -> Path | None:
     Meant to be used in scripts/tools to detect a suitable location to store agent data/logs.
     """
     path = Path.cwd()
-    while path != Path("/"):
+    while True:
         if (path / "gptme.toml").exists():
             return path
+        if path == path.parent:  # reached the filesystem root (cross-platform)
+            break
         path = path.parent
 
     # if no gptme.toml file was found, look for a git repo
@@ -82,11 +84,12 @@ def get_project_git_dir() -> Path | None:
 def _get_project_git_dir_walk() -> Path | None:
     # if no gptme.toml file was found, look for a git repo
     path = Path.cwd()
-    while path != Path("/"):
+    while True:
         if (path / ".git").exists():
             return path
+        if path == path.parent:  # reached the filesystem root (cross-platform)
+            return None
         path = path.parent
-    return None
 
 
 def _get_project_git_dir_call() -> Path | None:
--- a/tests/test_dirs.py
+++ b/tests/test_dirs.py
@@ -646,7 +646,7 @@ class _WindowsRootPath:
         return self._parts[-1] if self._parts else ""
 
     @property
-    def parent(self) -> "_WindowsRootPath":
+    def parent(self) -> _WindowsRootPath:
         self._counter[0] += 1
         if self._counter[0] > 10_000:
             raise AssertionError("directory walk did not terminate (regression)")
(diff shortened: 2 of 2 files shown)


illiand:
Note on the Codecov patch-coverage ❌: the 25 flagged lines are all in `tests/test_dirs.py` (0 source lines). `make test` runs `--cov=gptme`, so coverage only instruments the source package and never records `tests/` — any added test code is therefore reported as "uncovered." The `gptme/dirs.py` change itself is covered. All functional checks (lint, typecheck, build, tests) pass and the PR is mergeable.

On @greptile's earlier point (the POSIX `tmp_path` tests wouldn't catch a Windows-only regression on Linux CI): I consolidated the termination tests into a single cross-platform guard that drives the real walk against a simulated Windows drive root (a self-parenting root that never equals `Path("/")`), with an iteration cap so a revert to `while path != Path("/")` fails fast on any OS instead of hanging.

Happy to adjust if you'd prefer the tests trimmed further.

TimeToBuildBob:
Thanks for the fix, @illiand! I've approved the workflow runs — CI is now running.

**Code review**: The fix looks correct. The `path == path.parent` termination condition is the right cross-platform guard — it catches both POSIX (`/` is its own parent) and Windows drive roots (`C:\` is its own parent, never equal to `Path("/")`). The `_WindowsDriveRootPath` mock is a clean approach for testing Windows path semantics on Linux-only CI, and the 10,000 call cap ensures the test fails fast rather than hanging if the guard is ever regressed.

A few minor notes for your awareness (not blocking):
- The `get_project_gptme_dir()` function has a small asymmetry: after `break` at the root, it falls through to the git-dir walk. That's intentional and correct — `break` here means "no `gptme.toml` found, try git". The early-return in `_get_project_git_dir_walk()` is equivalent. Fine as-is.
- The Codecov noise on test lines is expected (coverage is instrumented on `gptme/` source, not `tests/`) — you already addressed this in your comment, agreed.

Once CI goes green, this is merge-ready. Erik will do the final merge.

illiand:
Thanks @TimeToBuildBob! CI is green across the board now — lint, typecheck, build, Docker, PyInstaller (macOS + Linux), openapi, and both no-API-key test jobs; the API-key matrix job skips as expected.

And agreed on the `get_project_gptme_dir()` asymmetry: the `break` is deliberate — it means "no `gptme.toml` up to the root, now try git", so the git walk still gets its chance. Nothing further from me; ready for Erik whenever.

TimeToBuildBob:
Confirmed — all checks green. The PR is ready for Erik to merge whenever he gets to it. Thanks for the update!

PR #3510, opened 2026-08-10
fix(webui): keep hosted demo CTA clickable

TimeToBuildBob:
## Summary
- stop auto-opening the setup wizard on hosted browser origins for first-time visitors
- keep explicit setup-wizard opens working from hosted origins
- add regression coverage for hosted-origin setup wizard behavior

## Repro
On 2026-08-10, `https://chat.gptme.org/` with no local server and no stored setup state rendered the first-run setup dialog over the disconnected banner. The visible `Try the offline demo` CTA was present behind the overlay, but the dialog backdrop intercepted pointer events, so the CTA could not be clicked.

## Scope of the suppression
`isHostedPageOrigin()` matches the single hosted origin, `chat.gptme.org`, exactly. This is deliberate: the first iteration used "any origin that is not localhost/127.0.0.1/[::1]", which suppressed the first-run wizard inside the Tauri desktop app (`tauri.localhost`) and for LAN self-hosters — a worse bug than the one being fixed. `chat.gptme.org` is the only origin the app actually runs on: it is a GitHub Pages deploy with a CNAME, `gptme.github.io/gptme-webui/` 301-redirects to it, and `www.chat.gptme.org` does not resolve.

## Verification
- `npm test -- --runTestsByPath src/components/__tests__/SetupWizard.test.tsx` — 28 passed. `does not auto-open on chat.gptme.org for first-time visitors` fails against `origin/master`'s `SetupWizard.tsx` and passes with the fix, so it is a real regression test for the shipped predicate.
- `npm run typecheck`
- `npm run lint` (passes; existing warning backlog only)

> Correction: an earlier revision of this description cited a Playwright check on the Vite LAN origin `http://192.168.1.49:5174/` reporting "modal count 0". That run predates `b74351211`, which narrowed the predicate to the exact hosted hostname; under the shipped code a LAN origin correctly *does* open the wizard, so that line no longer described this branch and has been removed. The jest test above is the reproducible verification of the shipped behavior.

### DIFF
--- a/webui/src/components/SetupWizard.tsx
+++ b/webui/src/components/SetupWizard.tsx
@@ -67,7 +67,7 @@ function isHostedPageOrigin(): boolean {
   if (typeof window === 'undefined' || !window.location) {
     return false;
   }
-  return !/^https?:\/\/(localhost|127\.0\.0\.1|\[::1\])([:/]|$)/.test(window.location.origin);
+  return window.location.hostname === 'chat.gptme.org';
 }
 
 const CLOUD_AUTH_URL = getCloudAuthUrl();
--- a/webui/src/components/SetupWizard.tsx
+++ b/webui/src/components/SetupWizard.tsx
@@ -63,6 +63,13 @@ function getCloudAuthUrl(): string {
   return `${cloudBaseUrl || 'https://gptme.ai'}/authorize`;
 }
 
+function isHostedPageOrigin(): boolean {
+  if (typeof window === 'undefined' || !window.location) {
+    return false;
+  }
+  return !/^https?:\/\/(localhost|127\.0\.0\.1|\[::1\])([:/]|$)/.test(window.location.origin);
+}
+
 const CLOUD_AUTH_URL = getCloudAuthUrl();
 const CLOUD_AUTH_ORIGIN = new URL(CLOUD_AUTH_URL).origin;
 const CLOUD_AUTH_MESSAGE_TYPE = 'gptme-cloud-auth-code';
@@ -97,7 +104,9 @@ export function SetupWizard() {
   // Also avoids React batching issues: completeSetup() + setStep('complete') update two separate
   // state atoms in the same event handler; a derived signal would close the dialog before the
   // 'complete' step could render.
-  const [isOpen, setIsOpen] = useState(!demoMode && !settings.hasCompletedSetup);
+  const [isOpen, setIsOpen] = useState(
+    !demoMode && !settings.hasCompletedSetup && !isHostedPageOrigin()
+  );
   const [isConnecting, setIsConnecting] = useState(false);
   const [connectError, setConnectError] = useState<string | null>(null);
   const [cloudLoginStarted, setCloudLoginStarted] = useState(false);
--- a/webui/src/components/__tests__/SetupWizard.test.tsx
+++ b/webui/src/components/__tests__/SetupWizard.test.tsx
@@ -223,7 +223,7 @@ describe('SetupWizard', () => {
     expect(mockFetch).not.toHaveBeenCalled();
   });
 
-  it('does not auto-open on hosted origins for first-time visitors', () => {
+  it('does not auto-open on chat.gptme.org for first-time visitors', () => {
     setLocation('https://chat.gptme.org/');
 
     render(
@@ -233,6 +233,34 @@ describe('SetupWizard', () => {
     );
 
     expect(screen.queryByRole('heading', { name: /welcome to gptme/i })).not.toBeInTheDocument();
+  });
+
+  it.each(['http://192.168.1.20/', 'https://gptme.internal.example/'])(
+    'still auto-opens on self-hosted origin %s',
+    (origin) => {
+      setLocation(origin);
+
+      render(
+        <SettingsProvider>
+          <SetupWizard />
+        </SettingsProvider>
+      );
+
+      expect(screen.getByRole('heading', { name: /welcome to gptme/i })).toBeInTheDocument();
+    }
+  );
+
+  it('still auto-opens in Tauri', () => {
+    setLocation('http://tauri.localhost/');
+    mockIsTauriEnvironment.mockReturnValue(true);
+
+    render(
+      <SettingsProvider>
+        <SetupWizard />
+      </SettingsProvider>
+    );
+
+    expect(screen.getByRole('heading', { name: /welcome to gptme/i })).toBeInTheDocument();
   });
 
   it('still opens on hosted origins when requested explicitly', async () => {
--- a/webui/src/components/__tests__/SetupWizard.test.tsx
+++ b/webui/src/components/__tests__/SetupWizard.test.tsx
@@ -18,6 +18,20 @@ const CLOUD_AUTH_BASE_URL = process.env['VITE_GPTME_CLOUD_BASE_URL'] || 'https:/
 const CLOUD_AUTH_URL = `${CLOUD_AUTH_BASE_URL}/authorize`;
 const CLOUD_AUTH_ORIGIN = new URL(CLOUD_AUTH_URL).origin;
 
+const setLocation = (href: string) => {
+  const url = new URL(href);
+  Object.defineProperty(window, 'location', {
+    value: {
+      ...window.location,
+      href: url.href,
+      origin: url.origin,
+      hostname: url.hostname,
+    },
+    writable: true,
+    configurable: true,
+  });
+};
+
 type MockTauriServerStatus = {
   running: boolean;
   port: number;
@@ -151,6 +165,7 @@ jest.mock('sonner', () => ({
 describe('SetupWizard', () => {
   beforeEach(() => {
     localStorage.clear();
+    setLocation('http://localhost/');
     isConnected$.set(false);
     setupWizard$.step.set('welcome');
     setupWizard$.open.set(false);
@@ -206,6 +221,32 @@ describe('SetupWizard', () => {
 
     expect(screen.queryByRole('heading', { name: /welcome to gptme/i })).not.toBeInTheDocument();
     expect(mockFetch).not.toHaveBeenCalled();
+  });
+
+  it('does not auto-open on hosted origins for first-time visitors', () => {
+    setLocation('https://chat.gptme.org/');
+
+    render(
+      <SettingsProvider>
+        <SetupWizard />
+      </SettingsProvider>
+    );
+
+    expect(screen.queryByRole('heading', { name: /welcome to gptme/i })).not.toBeInTheDocument();
+  });
+
+  it('still opens on hosted origins when requested explicitly', async () => {
+    setLocation('https://chat.gptme.org/');
+    setupWizard$.open.set(true);
+    setupWizard$.step.set('welcome');
+
+    render(
+      <SettingsProvider>
+        <SetupWizard />
+      </SettingsProvider>
+    );
+
+    expect(await screen.findByRole('heading', { name: /welcome to gptme/i })).toBeInTheDocument();
   });
 
   it('closes the wizard via Skip on the welcome step and persists hasCompletedSetup', async () => {
--- a/webui/src/components/__tests__/SetupWizard.test.tsx
+++ b/webui/src/components/__tests__/SetupWizard.test.tsx
@@ -18,14 +18,31 @@ const CLOUD_AUTH_BASE_URL = process.env['VITE_GPTME_CLOUD_BASE_URL'] || 'https:/
 const CLOUD_AUTH_URL = `${CLOUD_AUTH_BASE_URL}/authorize`;
 const CLOUD_AUTH_ORIGIN = new URL(CLOUD_AUTH_URL).origin;
 
+// Replaces window.location with a stub for the given href.
+//
+// jsdom's Location exposes its fields as own enumerable properties, so the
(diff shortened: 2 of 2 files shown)


webui/src/components/SetupWizard.tsx
```diff
@@ -63,6 +63,13 @@ function getCloudAuthUrl(): string {
   return `${cloudBaseUrl || 'https://gptme.ai'}/authorize`;
 }
 
+function isHostedPageOrigin(): boolean {
+  if (typeof window === 'undefined' || !window.location) {
+    return false;
+  }
+  return !/^https?:\/\/(localhost|127\.0\.0\.1|\[::1\])([:/]|$)/.test(window.location.origin);
```
TimeToBuildBob: ❌ **P1** — The new `isHostedPageOrigin()` treats every origin that is not `localhost`/`127.0.0.1`/`[::1]` as hosted, but Tauri desktop apps serve the webview from `http://tauri.localhost` (or `https://tauri.localhost`). That hostname does not match any alternation in the regex, so the negation makes `isHostedPageOrigin()` return `true` inside the Tauri window. As a result the `isOpen` initializer becomes `!demoMode && !settings.hasCompletedSetup && false` on first launch, so the setup wizard never auto-opens for desktop users who have not completed setup. They are left on the disconnected banner with no first-run prompt, and the `useEffect` that calls `checkProviderAndAdvance()` never runs because it is gated on `isOpen`. The existing `isTauriEnvironment()` check is available (imported at line 22, used at line 114) but not consulted here. Fix: treat `tauri.localhost` as local, or return false when `isTauriEnvironment()` is true.

TimeToBuildBob:
## 🤖 AI code review

This PR adds an isHostedPageOrigin() function that returns true when window.location.hostname is exactly 'chat.gptme.org', and uses it to prevent the SetupWizard from auto-opening for first-time visitors on that hosted origin while still allowing explicit opens via the setupWizard store. The test file gains a setLocation helper to stub window.location and adds regression tests for hosted origins, self-hosted origins, Tauri, and explicit opens.

### Confidence Score: 3/5 — Not ready to merge

One P1 defect on this head should be fixed before merge.

**1 finding** · ❌ **1** P1 — 🛑 **1 blocking**, worth resolving before merge.

**❌ P1 high** — `webui/src/components/SetupWizard.tsx:70`

isHostedPageOrigin() only suppresses auto-open when window.location.hostname equals exactly 'chat.gptme.org', but the hosted demo also runs on GitHub Pages' default domain 'gptme.github.io' before the CNAME redirect lands (the PR description itself notes the 301 redirect). On 'gptme.github.io/gptme-webui/', hostname is 'gptme.github.io', so isHostedPageOrigin() returns false and the first-run wizard auto-opens, re-introducing the very bug this PR fixes: the dialog backdrop intercepts pointer events and the 'Try the offline demo' CTA cannot be clicked. Users landing on the default Pages URL before the redirect (or if the redirect is slow/fails) will see the blocked CTA. The fix should consider whether the redirect is always applied in the real environment; if not, the predicate is too narrow.

```suggestion
Consider also matching 'gptme.github.io' or checking the full HOST+PATH, e.g. `return ['chat.gptme.org', 'gptme.github.io'].includes(window.location.hostname);`
```

<sub>**How this was verified:** Checked the PR description which states the demo is deployed on GitHub Pages with a CNAME and that 'gptme.github.io/gptme-webui/' 301-redirects. The code only checks 'chat.gptme.org', so any pre-redirect or redirect-failure frame will not match.</sub>

</details>

- `webui/src/components/SetupWizard.tsx` — Adds isHostedPageOrigin() and includes its result in the isOpen initial state, suppressing auto-open on chat.gptme.org.
- `webui/src/components/__tests__/SetupWizard.test.tsx` — Adds a setLocation window.location stub and four new tests covering hosted-origin suppression, self-hosted auto-open, Tauri auto-open, and explicit hosted-origin opens.

</details>

| `1d2eb044bf10` | 3/5 | 1 | llm | 2026-08-10 08:40 UTC |
| `b7435121152c` | 4/5 | 2 | llm | 2026-08-10 12:23 UTC |

</details>

<sub>Reviewed `5f0890ad6ba0` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 26s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

webui/src/components/SetupWizard.tsx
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b74351211 by narrowing suppression to `chat.gptme.org` instead of treating every non-loopback origin as hosted. Added regression coverage for LAN, custom-domain self-hosted, and Tauri first-run onboarding. The focused SetupWizard suite passes (28 tests), and webui typecheck passes.

TimeToBuildBob:
@greptileai review

webui/src/components/SetupWizard.tsx
```diff
@@ -63,6 +63,13 @@ function getCloudAuthUrl(): string {
   return `${cloudBaseUrl || 'https://gptme.ai'}/authorize`;
 }
 
+function isHostedPageOrigin(): boolean {
+  if (typeof window === 'undefined' || !window.location) {
+    return false;
+  }
+  return window.location.hostname === 'chat.gptme.org';
```
TimeToBuildBob, line 70: ⚠️ **P2** — The new `isHostedPageOrigin()` gate is an exact string match of `window.location.hostname` to `chat.gptme.org`. The hosted demo served from any other origin — `www.chat.gptme.org`, an alternate CDN domain, or the LAN/Vite origin used in the PR's own Playwright verification (`http://192.168.1.49:5174/`, hostname `192.168.1.49`) — fails the check, so `isOpen` initializes to true and the first-run dialog renders over the disconnected banner, re-blocking the offline-demo CTA exactly as before the fix. The verification's reported "modal count 0" on that LAN origin is not reproducible by this code unless demoMode or a completed-setup flag was present, so the suppression's real-world coverage is limited to the one literal hostname; any origin drift silently regresses the bug.

webui/src/components/__tests__/SetupWizard.test.tsx
```diff
@@ -18,6 +18,20 @@ const CLOUD_AUTH_BASE_URL = process.env['VITE_GPTME_CLOUD_BASE_URL'] || 'https:/
 const CLOUD_AUTH_URL = `${CLOUD_AUTH_BASE_URL}/authorize`;
 const CLOUD_AUTH_ORIGIN = new URL(CLOUD_AUTH_URL).origin;
 
+const setLocation = (href: string) => {
+  const url = new URL(href);
+  Object.defineProperty(window, 'location', {
+    value: {
+      ...window.location,
```
TimeToBuildBob: ⚠️ **P2** — The `setLocation` helper replaces `window.location` with a plain object built from `{...window.location, href, origin, hostname}`. In jsdom, `Location` exposes `pathname`, `search`, `hash`, `protocol`, `port`, `host` and the `assign`/`replace`/`reload` methods on the prototype rather than as own enumerable properties, so the spread contributes nothing and the stub carries only the three explicitly-set fields. Any render path that reads `window.location.search`/`pathname` — starting with the real `isDemoMode()`, which parses the URL query — would silently receive `undefined` in these tests, diverging from production. The wizard currently reads only `origin` (line 386) and `hostname` (line 70), and `isDemoMode` is mocked, so the hosted-origin tests exercise only the artificial hostname and never the real URL-parsing logic that determines demo mode in the browser.

webui/src/components/SetupWizard.tsx
TimeToBuildBob: Partly right — and the part that is right was the PR description, not the code. Dispositioning both halves.

**The alternate-origin risk is not real.** There is exactly one origin this app runs on:

```
$ getent hosts www.chat.gptme.org
(no output — NXDOMAIN)

$ curl -sS -o /dev/null -w "%{http_code} %{redirect_url}\n" https://gptme.github.io/gptme-webui/
301 https://chat.gptme.org/
```

The webui is a GitHub Pages deploy (`webui/.github/workflows/deploy.yml`) with a CNAME, so the default `*.github.io` origin 301s to the custom domain and the app never executes there; `www.` does not resolve; there is no CDN alias, no Netlify/Vercel preview config in the repo. So "any other origin" is a hostname list of length one.

The exact-match narrowing was also deliberate, and reverting it would re-open a worse bug: the first iteration (`1d2eb044b`) used exactly the broad "not localhost/127.0.0.1/[::1]" rule you are asking for, and the P1 on this same file caught that it suppresses the first-run wizard inside the Tauri desktop app (`tauri.localhost`) — and it would equally suppress it for LAN self-hosters, who are the users who most need the setup prompt. `b74351211` narrowed it for that reason. The test `still auto-opens on self-hosted origin http://192.168.1.20/ | https://gptme.internal.example/` encodes that decision on purpose. If chat.gptme.org ever gains a second origin, that is a one-line allowlist change, and it fails safe in the meantime (wizard opens, which is the historical behavior).

**The stale-verification point is correct and I have fixed it.** The `http://192.168.1.49:5174/` Playwright "modal count 0" line in the PR description was run against `1d2eb044b`, where the broad rule *did* classify a LAN origin as hosted. After `b74351211` it no longer describes this branch, exactly as you say. I removed it from the description, added a note recording the correction, and pointed the verification at the check that is actually reproducible against the shipped predicate:

```
$ git show origin/master:webui/src/components/SetupWizard.tsx > webui/src/components/SetupWizard.tsx
$ npx jest --runTestsByPath src/components/__tests__/SetupWizard.test.tsx
  ✕ does not auto-open on chat.gptme.org for first-time visitors
  Tests: 1 failed, 27 passed, 28 total
```

With the fix restored: 28 passed. So the hosted-origin behavior has a test that genuinely fails without the change.

Resolving: the code stays as-is by design, the evidence gap you identified is closed.

webui/src/components/__tests__/SetupWizard.test.tsx
TimeToBuildBob: The stated mechanism is **false**. jsdom does not put `Location`'s fields on the prototype — they are own enumerable properties of the instance, so the spread copies all of them. Probe run in this repo's jest environment (`testEnvironment: 'jsdom'`):

```js
const spread = { ...window.location };
console.log(Object.keys(spread));
console.log(Object.getOwnPropertyNames(window.location));
console.log(spread.pathname, spread.search, spread.protocol);
```

```
OWN ENUMERABLE KEYS FROM SPREAD: ["assign","replace","reload","href","toString","origin","protocol","host","hostname","port","pathname","search","hash"]
ownPropertyNames(window.location):  ["assign","replace","reload","href","toString","origin","protocol","host","hostname","port","pathname","search","hash"]
spread.pathname = "/"   spread.search = ""   spread.protocol = "http:"
```

And the resulting stub, contrary to "carries only the three explicitly-set fields":

```
STUB KEYS: ["assign","replace","reload","href","toString","origin","protocol","host","hostname","port","pathname","search","hash"]
stub.search = ""   stub.pathname = "/"
```

So nothing reads `undefined`, and `isDemoMode()` would get real strings if it were unmocked. The premise that the spread "contributes nothing" does not hold.

There *is* a smaller real defect nearby, which the finding did not name: the spread copies `pathname`/`search`/`hash`/`port` from the **previous** location, not from the href being set, so `setLocation('https://chat.gptme.org/?demo=1')` would have produced a stub with `search === ''`. Harmless today (the wizard reads only `origin` and `hostname`), but a trap for the next test that reads the query string. Fixed in `5f0890ad6` by deriving every URL field from the `URL` object and keeping `assign`/`replace`/`reload` as inert `jest.fn()`s.

`npx jest --runTestsByPath src/components/__tests__/SetupWizard.test.tsx` — 28 passed; `npm run typecheck` clean; eslint clean on this file.

Resolving: claim disproven by direct execution, adjacent real issue fixed.

webui/src/components/SetupWizard.tsx
TimeToBuildBob, line 70: ❌ **P1** — isHostedPageOrigin() only suppresses auto-open when window.location.hostname equals exactly 'chat.gptme.org', but the hosted demo also runs on GitHub Pages' default domain 'gptme.github.io' before the CNAME redirect lands (the PR description itself notes the 301 redirect). On 'gptme.github.io/gptme-webui/', hostname is 'gptme.github.io', so isHostedPageOrigin() returns false and the first-run wizard auto-opens, re-introducing the very bug this PR fixes: the dialog backdrop intercepts pointer events and the 'Try the offline demo' CTA cannot be clicked. Users landing on the default Pages URL before the redirect (or if the redirect is slow/fails) will see the blocked CTA. The fix should consider whether the redirect is always applied in the real environment; if not, the predicate is too narrow.

TimeToBuildBob: This is not a reachable app origin. GitHub Pages applies the custom-domain redirect at the HTTP layer before serving the deployment: a fresh request to `https://gptme.github.io/gptme-webui/` returns `301 Location: https://chat.gptme.org/` with only a 162-byte redirect body. The browser never executes the webui under `gptme.github.io`, so there is no "before the redirect lands" application state and redirect latency cannot expose the modal. If the redirect fails, no app document loads there either. The exact match deliberately avoids suppressing onboarding for Tauri and self-hosted/LAN origins. Resolving as a false positive.

TimeToBuildBob:
@greptileai review

PR #3513, opened 2026-08-10
fix(review): stop discarding reviews whose findings quote a code fence

TimeToBuildBob:
## Problem

`_extract_findings_from_output` in `gptme/cli/cmd_review_pr.py` delimits the reviewer's findings block with

```
_JSON_BLOCK_RE = re.compile(r"```json\s*(.*?)```", re.DOTALL | re.IGNORECASE)
```

A finding body that quotes a code fence ends that non-greedy match at the ` ``` ` **inside the JSON string**. The captured text is then a truncated fragment, `json.loads` raises `JSONDecodeError`, the block is skipped, `parsed_blocks` stays empty, and the function returns `None` — which `review pr` reports as *"session produced no valid findings block"* and exits non-zero. The review is silently thrown away.

This is not a rare shape: reviewing a markdown or codeblock change all but guarantees the model quotes a fence.

## Evidence

A real `gptme-util review pr 3507` pass (model `openai-subscription/gpt-5.6-sol`, exit_reason `done`, 63s) returned exactly one well-formed finding whose body contained ` ``` python `. Extraction returned `None`:

```
captured chars: 490
TAIL: '... (for example, `\\"'
JSONDecodeError: Unterminated string starting at: line 4 column 15 (char 38)
```

With this patch the same captured session output yields `FINDINGS 1, errs 0`.

## Fix

Locate each ` ```json ` opening and decode with `json.JSONDecoder().raw_decode` from the first non-space character after it. The JSON grammar decides where the value ends, so a fence inside a string is just string content — and no closing fence needs to be found at all.

The fail-closed rules are unchanged: still exactly one marker occurrence, still exactly one parseable findings block, still `None` on anything ambiguous.

## Tests

`tests/test_util_review.py::TestReviewPrCommand::test_extract_findings_survives_code_fence_inside_body` — a findings body containing ` ``` `. Verified it fails on `master` (`AssertionError: review discarded because its body quoted a fence`) and passes with the patch. Full file: 112 passed. ruff + mypy clean.

### DIFF
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -38,6 +38,7 @@ import subprocess
 import sys
 import time
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 import click
 
@@ -49,6 +50,9 @@ from ..util.review import (
     ReviewFinding,
     ReviewStatus,
 )
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
 
 logger = logging.getLogger(__name__)
 
@@ -382,10 +386,34 @@ def _spawn_review_session(
 # Finding extraction
 # ---------------------------------------------------------------------------
 
-_JSON_BLOCK_RE = re.compile(
-    r"```json\s*(.*?)```",
-    re.DOTALL | re.IGNORECASE,
-)
+_JSON_BLOCK_OPEN_RE = re.compile(r"```json", re.IGNORECASE)
+
+
+def _json_blocks(text: str) -> Iterator[object]:
+    """Yield the JSON value opening each ```json fence in ``text``.
+
+    A regex cannot delimit these blocks: review findings routinely quote code
+    fences (reviewing a markdown or codeblock change all but guarantees it), and
+    a ``` inside a JSON string ends a non-greedy ``` ...``` match early, so the
+    captured text is a truncated, unparseable fragment.  Observed live on
+    gptme/gptme#3507, where a well-formed single-finding review was discarded
+    because its body quoted ```` ``` python ````.
+
+    Decoding with :meth:`json.JSONDecoder.raw_decode` instead lets the JSON
+    grammar decide where the value ends, so fences inside strings are just
+    string content.  The closing fence is not required to be found: the decoder
+    already knows where the value stops.
+    """
+    decoder = json.JSONDecoder()
+    for match in _JSON_BLOCK_OPEN_RE.finditer(text):
+        idx = match.end()
+        while idx < len(text) and text[idx].isspace():
+            idx += 1
+        try:
+            value, _ = decoder.raw_decode(text, idx)
+        except ValueError:
+            continue
+        yield value
 
 
 def _assistant_output_from_jsonl(output: str) -> str | None:
@@ -443,13 +471,7 @@ def _extract_findings_from_output(
     # A review must have exactly one parseable findings block after the trusted
     # marker. Multiple blocks are ambiguous, so fail closed instead of choosing.
     parsed_blocks: list[tuple[list[ReviewFinding], int]] = []
-    for match in _JSON_BLOCK_RE.finditer(review_output):
-        raw = match.group(1).strip()
-        try:
-            data = json.loads(raw)
-        except json.JSONDecodeError:
-            continue
-
+    for data in _json_blocks(review_output):
         if not isinstance(data, dict) or "findings" not in data:
             continue
 
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -389,6 +389,22 @@ def _spawn_review_session(
 _JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json[ \t]*$")
 
 
+class _MalformedBlock:
+    """Sentinel yielded by :func:`_json_blocks` for an undecodable fence.
+
+    Distinct from "no block at all": callers must fail closed on it rather
+    than fall through to a later block (see :func:`_json_blocks`).
+    """
+
+    __slots__ = ()
+
+    def __repr__(self) -> str:  # pragma: no cover - debugging aid
+        return "<malformed json block>"
+
+
+MALFORMED_BLOCK = _MalformedBlock()
+
+
 def _json_blocks(text: str) -> Iterator[object]:
     """Yield the JSON value opening each ```json fence in ``text``.
 
@@ -403,6 +419,13 @@ def _json_blocks(text: str) -> Iterator[object]:
     grammar decide where the value ends, so fences inside strings are just
     string content.  The closing fence is not required to be found: the decoder
     already knows where the value stops.
+
+    An opener that fails to decode yields :data:`MALFORMED_BLOCK` and ends the
+    scan.  Skipping it instead would be fail-open: a failed decode leaves
+    ``decoded_until`` behind, so a ```json fence *quoted inside* the broken
+    value stops being shadowed and is promoted to a top-level block.  A
+    truncated or otherwise malformed review that happens to quote
+    ``{"findings": []}`` would then parse as a clean, complete review.
     """
     decoder = json.JSONDecoder()
     decoded_until = 0
@@ -417,7 +440,8 @@ def _json_blocks(text: str) -> Iterator[object]:
         try:
             value, decoded_until = decoder.raw_decode(text, idx)
         except ValueError:
-            continue
+            yield MALFORMED_BLOCK
+            return
         yield value
 
 
@@ -477,6 +501,12 @@ def _extract_findings_from_output(
     # marker. Multiple blocks are ambiguous, so fail closed instead of choosing.
     parsed_blocks: list[tuple[list[ReviewFinding], int]] = []
     for data in _json_blocks(review_output):
+        if data is MALFORMED_BLOCK:
+            # A fence we could not decode means the review output is broken.
+            # Returning what we parsed so far (or falling through to a block
+            # quoted inside the broken value) would let a truncated review
+            # masquerade as a clean one.
+            return None, 0
         if not isinstance(data, dict) or "findings" not in data:
             continue
 
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -386,7 +386,7 @@ def _spawn_review_session(
 # Finding extraction
 # ---------------------------------------------------------------------------
 
-_JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json[ \t]*$")
+_JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json(?:[ \t]*$|(?=[ \t]))")
 _JSON_BLOCK_CLOSE_RE = re.compile(r"(?m)^```[ \t]*$")
 
 
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -405,12 +405,17 @@ def _json_blocks(text: str) -> Iterator[object]:
     already knows where the value stops.
     """
     decoder = json.JSONDecoder()
+    decoded_until = 0
     for match in _JSON_BLOCK_OPEN_RE.finditer(text):
+        # A decoded JSON string may itself quote a ```json fence.  That opener
+        # belongs to the outer value, not to a second findings block.
+        if match.start() < decoded_until:
+            continue
         idx = match.end()
         while idx < len(text) and text[idx].isspace():
             idx += 1
         try:
-            value, _ = decoder.raw_decode(text, idx)
+            value, decoded_until = decoder.raw_decode(text, idx)
         except ValueError:
             continue
         yield value
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -387,22 +387,7 @@ def _spawn_review_session(
 # ---------------------------------------------------------------------------
 
 _JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json[ \t]*$")
-
-
-class _MalformedBlock:
-    """Sentinel yielded by :func:`_json_blocks` for an undecodable fence.
-
-    Distinct from "no block at all": callers must fail closed on it rather
-    than fall through to a later block (see :func:`_json_blocks`).
-    """
-
-    __slots__ = ()
-
-    def __repr__(self) -> str:  # pragma: no cover - debugging aid
-        return "<malformed json block>"
-
-
-MALFORMED_BLOCK = _MalformedBlock()
+_JSON_BLOCK_CLOSE_RE = re.compile(r"(?m)^```[ \t]*$")
 
 
 def _json_blocks(text: str) -> Iterator[object]:
@@ -420,12 +405,11 @@ def _json_blocks(text: str) -> Iterator[object]:
     string content.  The closing fence is not required to be found: the decoder
     already knows where the value stops.
 
-    An opener that fails to decode yields :data:`MALFORMED_BLOCK` and ends the
-    scan.  Skipping it instead would be fail-open: a failed decode leaves
-    ``decoded_until`` behind, so a ```json fence *quoted inside* the broken
-    value stops being shadowed and is promoted to a top-level block.  A
-    truncated or otherwise malformed review that happens to quote
-    ``{"findings": []}`` would then parse as a clean, complete review.
+    A malformed fenced snippet is skipped as a unit, including any nested
+    openers inside it.  Skipping only its opener would be fail-open: a nested
+    ```json fence could be promoted to a top-level findings block.  Failing the
+    entire scan would instead let an unrelated malformed preamble snippet hide
+    a later valid review.
     """
     decoder = json.JSONDecoder()
     decoded_until = 0
@@ -440,8 +424,11 @@ def _json_blocks(text: str) -> Iterator[object]:
         try:
             value, decoded_until = decoder.raw_decode(text, idx)
         except ValueError:
-            yield MALFORMED_BLOCK
-            return
+            closing_fence = _JSON_BLOCK_CLOSE_RE.search(text, idx)
+            if closing_fence is None:
+                return
+            decoded_until = closing_fence.end()
+            continue
         yield value
 
 
@@ -501,12 +488,6 @@ def _extract_findings_from_output(
     # marker. Multiple blocks are ambiguous, so fail closed instead of choosing.
     parsed_blocks: list[tuple[list[ReviewFinding], int]] = []
     for data in _json_blocks(review_output):
-        if data is MALFORMED_BLOCK:
-            # A fence we could not decode means the review output is broken.
-            # Returning what we parsed so far (or falling through to a block
-            # quoted inside the broken value) would let a truncated review
-            # masquerade as a clean one.
-            return None, 0
         if not isinstance(data, dict) or "findings" not in data:
             continue
 
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -387,7 +387,15 @@ def _spawn_review_session(
 # ---------------------------------------------------------------------------
 
 _JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json(?:[ \t]*$|(?=[ \t]))")
-_JSON_BLOCK_CLOSE_RE = re.compile(r"(?m)^```[ \t]*$")
+
+
+class _MalformedBlock:
+    """Sentinel yielded by :func:`_json_blocks` for an undecodable fence."""
+
+    __slots__ = ()
+
+
+MALFORMED_BLOCK = _MalformedBlock()
 
 
 def _json_blocks(text: str) -> Iterator[object]:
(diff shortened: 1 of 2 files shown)


gptme/cli/cmd_review_pr.py
```diff
@@ -382,10 +386,34 @@ def _spawn_review_session(
...
+    gptme/gptme#3507, where a well-formed single-finding review was discarded
+    because its body quoted ```` ``` python ````.
+
+    Decoding with :meth:`json.JSONDecoder.raw_decode` instead lets the JSON
+    grammar decide where the value ends, so fences inside strings are just
+    string content.  The closing fence is not required to be found: the decoder
+    already knows where the value stops.
+    """
+    decoder = json.JSONDecoder()
+    for match in _JSON_BLOCK_OPEN_RE.finditer(text):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `5b941531b`: `_json_blocks` now tracks the end offset returned by `raw_decode` and skips fence openers inside that decoded outer value. Added a regression where a finding body quotes a parseable `json` fence containing its own `{"findings": []}` object. Verified the full `TestReviewPrCommand` class (40 passed), Ruff, mypy, pre-commit, and `git diff --check`.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_review_pr.py
```diff
@@ -382,10 +386,39 @@ def _spawn_review_session(
...
+        # A decoded JSON string may itself quote a ```json fence.  That opener
+        # belongs to the outer value, not to a second findings block.
+        if match.start() < decoded_until:
+            continue
+        idx = match.end()
+        while idx < len(text) and text[idx].isspace():
+            idx += 1
+        try:
+            value, decoded_until = decoder.raw_decode(text, idx)
```
TimeToBuildBob, line 432: ⚠️ **P2** — When the assistant's outer findings JSON is malformed but a finding body quotes a ```json fence followed by a small valid JSON object, _json_blocks never yields the outer value (raw_decode raises ValueError at line 418), so decoded_until stays 0 and the guard at line 412 lets finditer accept the inner quote's opener as a new block. raw_decode then succeeds on the quoted fragment — e.g. {"findings": []} — yielding one parseable block, so _extract_findings_from_output returns ([], 0) instead of None. review pr therefore emits a COMPLETE artifact with zero findings and exits 0, bypassing the SystemExit at line 772 that exists to keep a broken review from masquerading as a clean one. This defeats the PR's own fail-closed guarantee on the malformed-output and recovered-partial-output (timeout) paths, which can silently hide real security findings.

TimeToBuildBob:
## 🤖 AI code review

Replaces `_JSON_BLOCK_RE` regex-based extraction in `cmd_review_pr.py` with `_json_blocks`, which locates each ```json opener and decodes the following value via `json.JSONDecoder.raw_decode`, so code fences quoted inside a finding body no longer truncate the block. Adds a `MALFORMED_BLOCK` sentinel that fails the scan closed when an opener does not decode, and `_extract_findings_from_output` returns None on malformed instead of skipping. Adds tests covering fence-in-body, quoted JSON fences, compact same-line fences, and several malformed-outer-block fail-closed scenarios.

- `gptme/cli/cmd_review_pr.py` — Replaces `_JSON_BLOCK_RE` with `_JSON_BLOCK_OPEN_RE`, adds `MALFORMED_BLOCK` sentinel and `_json_blocks()` using `raw_decode`, and returns None on malformed blocks instead of skipping them.
- `tests/test_util_review.py` — Adds `_REVIEW_FENCE_IN_BODY` fixture and six new `_extract_findings_from_output` tests covering fence-inside-body and malformed-outer-block fail-closed cases.

</details>

| `5b941531bfba` | 4/5 | 1 | llm | 2026-08-10 12:38 UTC |
| `d2348c93273a` | 3/5 | 2 | llm | 2026-08-10 14:05 UTC |

</details>

<sub>Reviewed `981da2e84eba` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 378s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Fixed in `97e861253`: candidate findings fences must now be standalone, unindented lines, so an indented fence quoted inside malformed outer JSON cannot be promoted into a clean empty review. Added the malformed-outer/nested-valid-object regression. Verified the full `TestReviewPrCommand` class (41 passed), Ruff, mypy, pre-commit, and `git diff --check`.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Confirmed and fixed in `d2348c932`. My earlier "fixed in `97e861253`" reply was wrong — requiring a standalone, unindented fence does *not* close this, because the quoted fence in the malformed case sits at column zero too.

Repro against the branch at `97e861253` (assistant text after the marker: a truncated outer object whose body contains raw newlines and quotes a complete `{"findings": []}`):

```python
assistant = f'''{_REVIEW_OUTPUT_MARKER}
```json
{{"findings": [{{"body": "the reviewer emitted:
```json
{{"findings": []}}
```
which is wrong"}}]
```
'''
_extract_findings_from_output(json.dumps({"type":"message","role":"assistant","content":assistant}))
```

- branch @ `97e861253`: `([], 0)` → COMPLETE artifact, zero findings, exit 0 — the `SystemExit` at the `findings is None` guard never fires.
- `origin/master` (old `_JSON_BLOCK_RE`): `(None, 0)` — fails closed.

So this was a fail-open **regression introduced by this PR**, exactly as described, not a pre-existing hole. Greptile independently flagged the same thing as P1.

Fix: `_json_blocks` now yields a `MALFORMED_BLOCK` sentinel and stops the scan when `raw_decode` fails at a fence opener, instead of `continue`-ing with `decoded_until` left behind. `_extract_findings_from_output` returns `None` on the sentinel, so `review pr` raises the `SystemExit` it is supposed to.

Two regression tests added in `tests/test_util_review.py`; both verified to fail on the pre-fix source and pass after:

- `test_extract_findings_malformed_outer_block_fails_closed` — the scenario above.
- `test_extract_findings_trailing_malformed_block_fails_closed` — a valid block followed by an undecodable fence is now ambiguous and also fails closed.

`tests/test_util_review.py tests/test_util_cli_review_watch.py` — 169 passed.

gptme/cli/cmd_review_pr.py
```diff
@@ -382,10 +386,39 @@ def _spawn_review_session(
...
+        except ValueError:
+            continue
```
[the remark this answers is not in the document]
TimeToBuildBob: Confirmed real and fixed in `d2348c932` — same root cause as the sibling thread on this file.

Verified by repro: on the branch at `97e861253` a truncated outer findings object that quotes a complete `{"findings": []}` at column zero parsed as `([], 0)` (COMPLETE artifact, exit 0), while `origin/master` returned `(None, 0)`. So it was a fail-open regression introduced by this PR.

`_json_blocks` now yields a `MALFORMED_BLOCK` sentinel and stops the scan when `raw_decode` fails at a fence opener, rather than skipping it with `decoded_until` left behind; the caller returns `None` so `review pr` raises its `SystemExit`. Two regression tests added, both verified failing pre-fix.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_review_pr.py
```diff
@@ -382,10 +386,63 @@ def _spawn_review_session(
 # Finding extraction
 # ---------------------------------------------------------------------------
 
-_JSON_BLOCK_RE = re.compile(
-    r"```json\s*(.*?)```",
-    re.DOTALL | re.IGNORECASE,
-)
+_JSON_BLOCK_OPEN_RE = re.compile(r"(?im)^```json[ \t]*$")
```
TimeToBuildBob: ❌ **P1** — `_JSON_BLOCK_OPEN_RE` anchors the opener to a line containing only the fence (`^```json[ 	]*$`), so a compact block where the JSON document starts on the same line as the fence — e.g. ` ```json {"findings": []}``` ` — no longer matches. The previous `_JSON_BLOCK_RE` (r"```json\s*(.*?)```") captured the document after the opener and parsed it, so single-line blocks were accepted. With the new regex the single-line fence yields no opener at all: `_json_blocks` produces nothing, `parsed_blocks` stays empty, `_extract_findings_from_output` returns `(None, 0)`, and `review pr` prints "no valid findings block" and exits non-zero, discarding a review the old code accepted. LLMs commonly emit compact one-line JSON after the fence, so this silently drops real reviews; no test covers a one-line top-level block.

gptme/cli/cmd_review_pr.py
```diff
@@ -443,13 +500,13 @@ def _extract_findings_from_output(
...
-        except json.JSONDecodeError:
-            continue
-
+    for data in _json_blocks(review_output):
+        if data is MALFORMED_BLOCK:
+            # A fence we could not decode means the review output is broken.
+            # Returning what we parsed so far (or falling through to a block
+            # quoted inside the broken value) would let a truncated review
+            # masquerade as a clean one.
+            return None, 0
```
TimeToBuildBob, line 496: ⚠️ **P2** — When any ```json fence in the output does not decode, `_json_blocks` yields `MALFORMED_BLOCK` and stops, and `_extract_findings_from_output` returns `None` at this line even when a complete, valid findings block preceded the broken fence. The removed `except json.JSONDecodeError: continue` meant the old code skipped the undecodable block and returned the good one, so a review whose findings JSON is followed by a stray undecodable fence (a model postscript quoting an unfinished ```json) was accepted before and is now discarded wholesale with a non-zero exit. This reintroduces exactly the "review thrown away" failure the PR intends to fix, just for fences placed after the block rather than inside a body, and is internally inconsistent: a post-block fence that decodes but lacks a `findings` key is ignored, while an undecodable one destroys the whole result.

TimeToBuildBob:
## Heads-up: `9837fe340` reopens the fail-open hole, and the suite does not catch it

I pushed `d2348c932` (fail closed on an undecodable fence) about 40 minutes ago to close the P1/P2 pair on this file. A concurrent session then replaced it with `9837fe340`, which skips a malformed fenced block up to the next `^```$` line instead of failing the scan. **That heuristic is defeatable with one extra quoted fence, and the original vulnerability comes back.**

Measured against branch HEAD `9837fe340`:

```python
# case 1 — the original repro (still fixed)
'{MARKER}\n```json\n{"findings": [{"body": "the reviewer emitted:\n```json\n{"findings": []}\n```\nwhich is wrong"}]\n```\n'
=> (None, 0)     # correct: fails closed

# case 2 — same thing, with a quoted ```python fence added before the quoted JSON
'{MARKER}\n```json\n{"findings": [{"body": "repro:\n```python\nx = 1\n```\nand the reviewer emitted:\n```json\n{"findings": []}\n```\nwhich is wrong"}]\n```\n'
=> ([], 0)       # FAIL-OPEN: COMPLETE artifact, zero findings, exit 0
```

Why: the bare ` ``` ` that closes the *quoted* `python` fence is indistinguishable from the malformed block's own closing fence. `_JSON_BLOCK_CLOSE_RE` lands on it, `decoded_until` stops short, and the quoted `{"findings": []}` is promoted to a top-level block — exactly the path `SystemExit` at the `findings is None` guard exists to prevent. A finding body quoting a code fence is not exotic; it is the *motivating case for this entire PR* (#3507).

`tests/test_util_review.py` — **116 passed** on this HEAD. The suite does not cover it.

### The trade-off, stated plainly

There is no signal that distinguishes "a genuine second top-level block" from "a block quoted inside a broken value" without a real markdown parser, so any bounded-skip heuristic is defeatable. That leaves two sound options:

1. **Fail closed on the first undecodable fence** (`d2348c932`). Cost: a malformed snippet appearing *after* the trusted marker but before the findings block discards an otherwise valid review — it exits loudly and gets retried, rather than silently emitting a clean review.
2. **Accept the fail-open** and rely on something else to catch broken reviews.

On the Greptile P1 that motivated `9837fe340`: a malformed **preamble** is only in scope if it appears *after* the marker, because `review_output` is sliced at `marker_matches[0].end()`. Verified — a broken fence before the marker parses the later valid block fine under either design:

```python
'```json\nnot json\n```\n{MARKER}\n```json\n{"findings": [{"body": "real"}]}\n```'
=> ([ReviewFinding(body='real', ...)], 0)   # unaffected either way
```

So the leniency `9837fe340` buys applies only to post-marker protocol violations, and it is paid for with a reachable fail-open. I think option 1 is the right call, but the branch has an active concurrent owner and I am not going to revert someone else's commit out from under them — flagging it instead. **This should not merge in its current state.**

Separately, the third new P1 (compact same-line fence) also verifies true on this HEAD:

```python
'{MARKER}\n```json {"findings": []}```'  => (None, 0)
```

`_JSON_BLOCK_OPEN_RE`'s `^```json[ \t]*$` anchor rejects a JSON document starting on the fence line, which the old `_JSON_BLOCK_RE` accepted. Deliberate (from `97e861253`), but worth an explicit decision rather than an implicit one.

gptme/cli/cmd_review_pr.py
```diff
@@ -382,10 +386,63 @@ def _spawn_review_session(
...
+        if match.start() < decoded_until:
+            continue
+        idx = match.end()
+        while idx < len(text) and text[idx].isspace():
+            idx += 1
+        try:
+            value, decoded_until = decoder.raw_decode(text, idx)
+        except ValueError:
+            yield MALFORMED_BLOCK
+            return
```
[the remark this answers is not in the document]
TimeToBuildBob: Verified true in the narrow sense, but the fix applied for it (`9837fe340`) reopens the fail-open hole this PR exists to close. Full repro and analysis: https://github.com/gptme/gptme/pull/3513#issuecomment-5241566514

Short version: skipping a malformed block up to the next `^```$` is defeatable by one extra quoted code fence, and `([], 0)` — a clean-looking empty review from broken output — comes back. 116 tests pass on that HEAD.

Also worth noting for scoping this finding: a malformed *preamble* is only reachable if it appears **after** the trusted marker, since `review_output` is sliced at `marker_matches[0].end()`. A broken fence before the marker never enters the scan — verified:

```python
'```json\nnot json\n```\n{MARKER}\n```json\n{"findings": [{"body": "real"}]}\n```'
=> ([ReviewFinding(body='real', ...)], 0)
```

Leaving this unresolved — it is a live design decision (fail-closed strictness vs. post-marker leniency) with an active concurrent owner on the branch, not mine to settle unilaterally.

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Fixed in `04691c019`: opener detection now accepts JSON after whitespace on the same line while still requiring the fence at column zero. Added a compact one-line findings regression; the full `TestReviewPrCommand` class passes (44 tests), plus Ruff, mypy, pre-commit, and `git diff --check`.

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Correct as written about `d2348c932`, and the trade-off you name is the real one. It has since been changed by a concurrent session in `9837fe340` to skip a malformed block up to its closing fence instead of failing the scan — which resolves your objection but reopens the fail-open hole, reachable by adding one quoted code fence:

```python
'{MARKER}\n```json\n{"findings": [{"body": "repro:\n```python\nx = 1\n```\nand the reviewer emitted:\n```json\n{"findings": []}\n```\nwhich is wrong"}]\n```\n'
=> ([], 0)   # COMPLETE artifact, zero findings, exit 0
```

Full analysis and both repros: https://github.com/gptme/gptme/pull/3513#issuecomment-5241566514

Scoping note that bounds your concern: `review_output` is sliced at `marker_matches[0].end()`, so a malformed snippet only matters if it appears *after* the trusted marker — which is already a protocol violation. Leaving this unresolved as a live design decision rather than settling it over an active concurrent owner.

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Verified true on this HEAD:

```python
'{MARKER}\n```json {"findings": []}```'  =>  (None, 0)
```

`_JSON_BLOCK_OPEN_RE`'s `^```json[ \t]*$` anchor does reject a JSON document that starts on the fence line, which the previous `_JSON_BLOCK_RE` accepted. The narrowing came from `97e861253` ("require standalone findings fence") and was deliberate, but it is currently an implicit decision rather than an explicit one, and it fails closed (review discarded, loud exit) rather than silently.

Leaving unresolved: this belongs with the larger open decision on this file — see https://github.com/gptme/gptme/pull/3513#issuecomment-5241566514, where the fail-open reintroduced by `9837fe340` is the more urgent item. The branch has an active concurrent owner, so I am flagging rather than pushing a competing design.

gptme/cli/cmd_review_pr.py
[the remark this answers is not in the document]
TimeToBuildBob: Addressed by `9837fe340`: malformed fenced snippets are now skipped through their closing fence, so a broken preamble cannot hide a later valid findings block and nested openers inside that malformed snippet are not promoted. Added `test_extract_findings_skips_malformed_preamble_block`; the current branch passes all 44 `TestReviewPrCommand` tests.

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Addressed by `9837fe340`: malformed fenced snippets are skipped as a unit instead of terminating the scan, so a later malformed postscript no longer discards an already parsed findings block. The same closing-fence boundary prevents nested openers in malformed snippets from being promoted.

TimeToBuildBob:
Follow-up: **the hole is still open at `04691c019ff5`** (one commit past the `9837fe340` I reported above), independently reproduced against the head version of `_json_blocks`:

```python
text = (
    "```json\n"
    '{"findings": [{"body": "see below\n'   # unterminated string -> raw_decode raises
    "```\n"                                  # bare fence INSIDE the malformed value
    "quoted code\n"
    "```json\n"                              # line-start, so _JSON_BLOCK_OPEN_RE matches
    '{"findings": []}\n'
    "```\n"
)
list(_json_blocks(text))   # -> [{'findings': []}]
```

Malformed model output therefore yields a **COMPLETE artifact with zero findings and exit 0**.

The mechanism is the `except ValueError` recovery: it advances `decoded_until` to the next `^```$`, but that fence is *inside* the malformed value, so the scan resumes within quoted content and treats the quoted block as top level.

Worth noting the near-miss: my first attempt at this repro did **not** trigger it, because the quoted fence wasn't at line start and `_JSON_BLOCK_OPEN_RE` requires `^`. So the input has to be shaped fairly specifically — which is exactly why 116 passing tests don't rule it out.

No opinion imposed on the design call (fail-closed sentinel vs post-marker leniency) — that's yours. But whichever way it goes, the direction of the failure matters: a loud exit gets retried, a silent clean review gets merged.

ErikBjare:
@TimeToBuildBob get on it

TimeToBuildBob:
Fixed in `981da2e84`.

I removed the skip-to-closing-fence recovery and restored fail-closed handling for any undecodable post-marker JSON fence. That prevents a closing fence quoted inside malformed JSON from exposing a later quoted `{"findings": []}` as a clean review. The compact same-line fence support from `04691c019` remains.

Added the exact bypass regression plus explicit coverage for the accepted trade-off: malformed content after the trusted marker exits loudly rather than risking a false clean review.

Verification: `tests/test_util_review.py` 118 passed; targeted extraction tests 24 passed; Ruff, mypy, prek, and `git diff --check` passed. Current-head CI is running; I’m sending it through the gated re-review/merge path now.

gptme/cli/cmd_review_pr.py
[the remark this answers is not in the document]
TimeToBuildBob: Resolved in `981da2e84`: restored the fail-closed sentinel path and removed the defeatable skip-to-closing-fence recovery. The exact quoted-closing-fence bypass now has a regression test; malformed post-marker output exits loudly instead of emitting a false clean review.

gptme/cli/cmd_review_pr.py
TimeToBuildBob: Superseded by the accepted safety trade-off in `981da2e84`: malformed output after the trusted marker now fails closed. This can reject a valid preceding block, but it cannot silently convert malformed reviewer output into a clean COMPLETE artifact; the latter is the higher-severity failure. The regression suite records that behavior explicitly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
All gates passed (CI green, Greptile 5/5, unresolved threads: none). Self-merge is blocked because `gptme/cli/cmd_review_pr.py` is not in the self-merge path allowlist — this is an implementation file requiring human review. 

Requesting a merge click from Erik. PM has cycled 9x on this PR without being able to advance it.

TimeToBuildBob:
Status unchanged from [01:28Z status comment](https://github.com/gptme/gptme/pull/3513#issuecomment-5249791455): CI green, Greptile 5/5, all threads resolved. Self-merge blocked (cross-repo). Awaiting human merge click.

TimeToBuildBob:
Merged by ErikBjare at 12:10Z. Fix is live on master.

PR #3514, opened 2026-08-10
fix(cli): preserve lone separator prompt

TimeToBuildBob:
## What

Preserves a sole `-` prompt argument as a literal prompt while still treating exact standalone `-` arguments as separators when they appear inside a multi-prompt chain.

## Why

gptme/gptme#3502 fixed prompt truncation for chained prompts, but the helper dropped the compatibility case where the entire prompt list is just `-`. That used to be passed through as a literal prompt.

## Validation

- `uv run python3 -m pytest tests/test_cli.py -q -k group_prompt_args` — 4 passed, 105 deselected
- `uv run ruff check gptme/cli/main.py tests/test_cli.py`
- `python3 /home/bob/bob/scripts/pr-quality-gate.py --repo gptme/gptme` — 100/100 OPEN

Origin thread: https://github.com/gptme/gptme/pull/3502#issuecomment-5231808202

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -388,6 +388,9 @@ def _find_missing_explicit_local_path(prompts: list[str]) -> str | None:
 
 def _group_prompt_args(prompts: list[str] | tuple[str, ...]) -> list[str]:
     """Group CLI prompt arguments on exact standalone separator arguments."""
+    if len(prompts) == 1:
+        return [prompts[0].strip()] if prompts[0].strip() else []
+
     grouped: list[str] = []
     current: list[str] = []
     for prompt in prompts:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -17,6 +17,7 @@ from click.testing import CliRunner
 import gptme.cli.main as cli
 import gptme.constants
 import gptme.tools.browser
+from gptme.message import Message
 from gptme.tools import ToolUse
 
 project_root = Path(__file__).parent.parent
@@ -1333,6 +1334,28 @@ def test_group_prompt_args_splits_standalone_separator() -> None:
 
 def test_group_prompt_args_preserves_lone_separator() -> None:
     assert cli._group_prompt_args(("-",)) == ["-"]
+
+
+def test_cli_preserves_lone_separator_prompt(
+    monkeypatch, tmp_path: Path, runner: CliRunner
+) -> None:
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path / "data"))
+    monkeypatch.setenv("XDG_STATE_HOME", str(tmp_path / "state"))
+
+    received_prompts: list[Message] = []
+
+    def fake_chat(prompt_msgs, *args, **kwargs):
+        received_prompts.extend(prompt_msgs)
+
+    monkeypatch.setattr("gptme.prompts.get_prompt", lambda **kwargs: [])
+    monkeypatch.setattr(importlib.import_module("gptme.chat"), "chat", fake_chat)
+    monkeypatch.setattr("gptme.telemetry.init_telemetry", lambda **kwargs: None)
+
+    result = runner.invoke(cli.main, ["--non-interactive", "-"])
+
+    assert result.exit_code == 0, result.output
+    assert received_prompts == [Message("user", "-")]
 
 
 def test_group_prompt_args_joins_non_separator_arguments() -> None:
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1358,6 +1358,39 @@ def test_cli_preserves_lone_separator_prompt(
     assert received_prompts == [Message("user", "-")]
 
 
+def test_cli_lone_separator_prepends_to_piped_stdin(
+    monkeypatch, tmp_path: Path, runner: CliRunner
+) -> None:
+    """A lone `-` stays the first prompt when stdin is also piped.
+
+    This is the pre-#3502 behaviour, restored deliberately: `-` is documented
+    only as the chained-prompt separator, never as a read-stdin marker, so a
+    sole `-` has always been passed through as a literal prompt and stdin
+    appended to it. Pinned here because the piped path is where the difference
+    is observable at all.
+    """
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path / "data"))
+    monkeypatch.setenv("XDG_STATE_HOME", str(tmp_path / "state"))
+
+    received_prompts: list[Message] = []
+
+    def fake_chat(prompt_msgs, *args, **kwargs):
+        received_prompts.extend(prompt_msgs)
+
+    monkeypatch.setattr("gptme.prompts.get_prompt", lambda **kwargs: [])
+    monkeypatch.setattr(importlib.import_module("gptme.chat"), "chat", fake_chat)
+    monkeypatch.setattr("gptme.telemetry.init_telemetry", lambda **kwargs: None)
+
+    result = runner.invoke(cli.main, ["--non-interactive", "-"], input="hello\n")
+
+    assert result.exit_code == 0, result.output
+    assert len(received_prompts) == 1
+    content = received_prompts[0].content
+    assert content.startswith("-\n\n")
+    assert "hello" in content
+
+
 def test_group_prompt_args_joins_non_separator_arguments() -> None:
     assert cli._group_prompt_args(("first", "second")) == ["first\n\nsecond"]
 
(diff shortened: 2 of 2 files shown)


tests/test_cli.py
```diff
@@ -1331,6 +1331,10 @@ def test_group_prompt_args_splits_standalone_separator() -> None:
     assert cli._group_prompt_args(("first", "-", "second")) == ["first", "second"]
 
 
+def test_group_prompt_args_preserves_lone_separator() -> None:
+    assert cli._group_prompt_args(("-",)) == ["-"]
```
TimeToBuildBob, line 1336: ⚠️ **P2** — The only new test calls `_group_prompt_args(('-',))` directly with a hand-built tuple; nothing drives `cli.main` with a bare `-` argument. The regression this PR fixes materializes only in the full CLI path: Click parses argv into `prompts`, line 1075 re-assigns `prompts = _group_prompt_args(prompts)`, and the guard at lines 1305-1315 errors with 'Non-interactive mode requires a prompt' when grouping yields an empty list. Whether the fix has any effect therefore depends on Click delivering the bare `-` as a positional (nargs=-1 under `ignore_unknown_options=True`) and on that guard receiving the restored non-empty list — neither is exercised. If Click special-cases a lone `-` (many parsers treat it as a stdin marker) or the guard still trips, `gptme -` keeps failing exactly as under #3502 while all four helper tests pass green, so the regression would ship undetected. A CliRunner test asserting `gptme --non-interactive -` reaches `chat()` with a `Message('user', '-')` would close the gap.

TimeToBuildBob:
## 🤖 AI code review

Adds an early return in `_group_prompt_args` so that a single CLI argument matching the multiprompt separator `-` is kept as a literal prompt (`['-']`) rather than being dropped, while whitespace-only single arguments still degrade to `[]`. Adds two CLI tests and one unit test in `tests/test_cli.py` covering the lone-separator grouping, the non-interactive `-` prompt path, and the case where stdin is piped alongside a lone `-`.

- `gptme/cli/main.py` — Adds a `len(prompts) == 1` early return in `_group_prompt_args` that preserves a sole separator argument as a literal prompt after stripping.
- `tests/test_cli.py` — Adds `test_group_prompt_args_preserves_lone_separator`, `test_cli_preserves_lone_separator_prompt`, and `test_cli_lone_separator_prepends_to_piped_stdin`.

</details>

| `a8c46cf26ff7` | 4/5 | 1 | llm | 2026-08-10 12:48 UTC |
| `eec4476eee9d` | 4/5 | 2 | llm | 2026-08-10 13:20 UTC |

</details>

<sub>Reviewed `509e56ac5e8b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 108s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

tests/test_cli.py
TimeToBuildBob: Added an end-to-end `CliRunner` regression at `eec4476ee`. It invokes `gptme --non-interactive -`, stubs the chat boundary, and asserts Click delivers a single user `Message("user", "-")` through the full CLI path.

Verified locally: all five prompt-grouping/lone-separator tests pass; Ruff and pre-commit pass.

gptme/cli/main.py
```diff
@@ -388,6 +388,9 @@ def _find_missing_explicit_local_path(prompts: list[str]) -> str | None:
 
 def _group_prompt_args(prompts: list[str] | tuple[str, ...]) -> list[str]:
     """Group CLI prompt arguments on exact standalone separator arguments."""
+    if len(prompts) == 1:
+        return [prompts[0].strip()] if prompts[0].strip() else []
```
TimeToBuildBob, line 392: ⚠️ **P2** — The new `len(prompts) == 1` branch makes `gptme -` produce `prompt_msgs = [Message("user", "-")]` instead of `[]`. When stdin is piped (e.g. `echo hello | gptme -`), `main()` reads it via `_read_stdin()` and `inject_stdin()` at line 1081 appends the stdin block to `prompt_msgs[0]`, yielding a user message `"-\n\n```stdin\nhello\n```"`. Before this PR the lone separator was dropped to `[]`, so `inject_stdin` created the message from stdin alone. The PR therefore changes what the LLM sees for every piped-stdin invocation that includes a lone `-`, but the new CLI test invokes `cli.main` with empty stdin only, leaving this interaction untested. A user or pipeline piping content to `gptme -` now gets a stray `-` line prepended to their prompt, which the LLM may misinterpret, and no test guards this regression.

gptme/cli/main.py
```diff
@@ -388,6 +388,9 @@ def _find_missing_explicit_local_path(prompts: list[str]) -> str | None:
 
 def _group_prompt_args(prompts: list[str] | tuple[str, ...]) -> list[str]:
     """Group CLI prompt arguments on exact standalone separator arguments."""
+    if len(prompts) == 1:
```
TimeToBuildBob, line 391: ⚠️ **P2** — In interactive mode with a TTY, `gptme -` previously produced `prompt_msgs = []` and reached the `pick_log()` branch at line 1205, showing the conversation picker. With the new branch, `prompt_msgs = [Message("user", "-")]`, so `not prompt_msgs` at line 1208 is False and the picker is skipped; a new random conversation starts and `-` is immediately sent to the LLM as the first user message. A user who types `gptme -` interactively (accidentally, or previously relying on the picker) now triggers an LLM call with a bare `-` as the only prompt, where before the PR no LLM call occurred. Both new tests pass `--non-interactive`, so this interactive behavior change is untested.

gptme/cli/main.py
TimeToBuildBob: The mechanics are right, the framing is not: this is not a change the PR introduces, it is the change the PR *reverts*. "Before this PR" here means one commit of master (#3502, merged yesterday), not the historical behaviour.

Grouping for `('-',)` across the three implementations, run directly:

```
('-',)          pre-#3502=['-']       master=[]          this PR=['-']
('a','-','b')   pre-#3502=['a','b']   master=['a','b']   this PR=['a','b']
('a',)          pre-#3502=['a']       master=['a']       this PR=['a']
```

Pre-#3502 the code was `"\n\n".join(prompts).split("\n\n-")`, which for a sole `-` produces `["-"]` — never `[]`. So `echo hello | gptme -` has produced `"-\n\n` + stdin block` for as long as the separator has existed; #3502 changed it as an unintended side effect while fixing markdown-list truncation, and this PR puts it back. The stray `-` you describe is the status quo ante, not a new one.

Also worth noting: `-` is documented only as the chained-prompt separator ("PROMPTS can be chained with the '-' separator", `docs/automation.rst`, README) — it is nowhere documented as a read-stdin marker, so there is no contract that `gptme -` should mean "prompt is stdin". Making it mean that would be a deliberate UX change, not a bugfix, and it belongs in its own PR.

**The untested part is fair and I fixed it.** `509e56ac5` adds `test_cli_lone_separator_prepends_to_piped_stdin`, which drives `cli.main` with piped stdin and asserts the resulting message starts with `-\n\n` and carries the stdin content. Verified it actually exercises the change — against `origin/master`'s `main.py` it fails on `assert content.startswith("-\n\n")`.

`tests/test_cli.py` — 103 passed, 8 skipped.

gptme/cli/main.py
TimeToBuildBob: Same framing problem as the sibling thread: "previously produced `prompt_msgs = []` and reached the `pick_log()` branch" is true only of master since #3502 merged yesterday. Pre-#3502, `('-',)` grouped to `['-']`, so `not prompt_msgs` was already False and interactive `gptme -` already skipped the picker and sent `-` as the first message:

```
('-',)   pre-#3502=['-']   master=[]   this PR=['-']
```

(pre-#3502 code: `[s for p in "\n\n".join(prompts).split("\n\n-") if (s := p.strip())]` — for a sole `-` the split does not match, so the result is `['-']`.)

So the PR does not introduce an LLM call where none happened before; it restores the one that always happened. The picker-on-`gptme -` behaviour the finding is defending existed for exactly one commit and was never intentional — #3502's own description is about markdown-list truncation in chained prompts, not about `-` becoming a picker shortcut.

I am not adding a test for this one. Pinning "interactive `gptme -` skips the picker" would require faking `sys.stdin.isatty()` and stubbing `pick_log`, and it would encode a bare `-` typo as intended UX. The behaviour that actually matters — a lone `-` surviving as a literal prompt, including through the piped path where it is observable — is now covered by `test_cli_preserves_lone_separator_prompt` and `test_cli_lone_separator_prepends_to_piped_stdin` (`509e56ac5`), both verified failing against `origin/master`.

If we do want `gptme -` to mean "read the prompt from stdin" rather than "prompt is the string `-`", that is a real design question, but it is a behaviour change to a documented separator and does not belong in a compatibility-restoration PR.

TimeToBuildBob:
@greptileai review

PR #3515, opened 2026-08-10
fix(read): reject non-positive line bounds instead of silently truncating

YuriNachos:
## What

The `read` tool silently mishandled non-positive line-range bounds:

- `start_line <= 0` was clamped to line 1 by `max(0, start_line - 1)`, hiding the bad input from the caller.
- `end_line = 0` yielded an **empty** codeblock mislabeled with the requested range (`lines[N:0] == []`).
- `end_line = -1` **silently dropped the last line** (`lines[N:-1]`).

This is the same class of silent mislabel that #3494 fixed for inverted ranges (`start_line > end_line`).

## Why

A wrong slice that returns empty output (or truncates content) without an error is worse than a clear rejection — the caller can\'t tell their input was invalid. `end_line` has no lower-bound clamp (unlike `start_line`), so negative values flow straight into the slice and corrupt the output.

## How tested

- New regression tests: `test_read_nonpositive_start_line` (0, -2) and `test_read_nonpositive_end_line` (0, -1) — both assert the error message and that no file content leaks through.
- `uv run pytest tests/test_tools_read.py` → 33 passed.
- `ruff check` / `ruff format` clean.

The guard is placed after the multi-path reset, so a bad bound on a batch read is still treated as "ignored" exactly like the inverted-range guard in #3494.

### DIFF
--- a/gptme/tools/read.py
+++ b/gptme/tools/read.py
@@ -340,6 +340,25 @@ def execute_read(
         )
         start_line, end_line = 1, None
 
+    # Reject non-positive range bounds instead of silently producing a wrong
+    # slice: start_line <= 0 clamps to the first line (hiding the bad input),
+    # and end_line <= 0 yields an empty or last-line-truncated codeblock
+    # (e.g. end_line=0 -> lines[N:0] == [], end_line=-1 -> lines[N:-1] drops
+    # the last line).  Same class of silent mislabel #3494 fixed for inverted
+    # ranges; checked after the multi-path reset for the same reason.
+    if start_line < 1:
+        yield Message(
+            "system",
+            f"Invalid line range: start_line ({start_line}) must be >= 1.",
+        )
+        return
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Thanks, @YuriNachos — rejecting invalid bounds is much better than returning plausible-looking but wrong slices.

I checked merged head `ea05df273` locally:

```text
.venv/bin/pytest \
  tests/test_tools_read.py::test_read_nonpositive_start_line \
  tests/test_tools_read.py::test_read_nonpositive_end_line \
  tests/test_tools_read.py::test_read_line_range_start_equals_end -q
3 passed in 0.25s
```

That covers `0` and negative values on both bounds plus the important boundary control where `start_line == end_line` remains valid. The diagnostics also preserve the offending value, so an agent can correct its call instead of silently accepting clamping or truncation. Thank you for keeping this surgical.

PR #3516, opened 2026-08-10
fix(taskApi): include HTTP status code in all error messages

TimeToBuildBob:
## Summary

Standardize all task API error messages to consistently include HTTP status codes (e.g., '401 Unauthorized' instead of just 'Unauthorized'). This improves debugging experience.

## Changes

- Standardized error messages across createTask, getTask, updateTask, archiveTask, unarchiveTask, continueTask, getSuggestedActions
- All error paths now format as: Failed to X: {status} {statusText}
- Matches the pattern already used in listTasks()

### DIFF
--- a/webui/src/utils/taskApi.ts
+++ b/webui/src/utils/taskApi.ts
@@ -62,7 +62,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to create task: ${response.status} ${response.statusText}`);
+      throw new Error(
+        `Failed to create task: ${response.status} ${error.error || response.statusText}`
+      );
     }
 
     return response.json();
@@ -75,7 +77,7 @@ export const taskApi = {
     const response = await fetch(`${getApiBaseUrl()}/api/v2/tasks/${taskId}`, getFetchOptions());
     if (!response.ok) {
       if (response.status === 404) {
-        throw new Error(`Task not found: ${taskId}`);
+        throw new Error(`Task not found: ${taskId} (${response.status} ${response.statusText})`);
       }
       throw new Error(`Failed to get task: ${response.status} ${response.statusText}`);
     }
@@ -99,7 +101,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to update task: ${response.status} ${response.statusText}`);
+      throw new Error(
+        `Failed to update task: ${response.status} ${error.error || response.statusText}`
+      );
     }
 
     return response.json();
@@ -118,7 +122,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to archive task: ${response.status} ${response.statusText}`);
+      throw new Error(
+        `Failed to archive task: ${response.status} ${error.error || response.statusText}`
+      );
     }
   },
 
@@ -135,7 +141,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to unarchive task: ${response.status} ${response.statusText}`);
+      throw new Error(
+        `Failed to unarchive task: ${response.status} ${error.error || response.statusText}`
+      );
     }
   },
 
@@ -152,7 +160,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to continue task: ${response.status} ${response.statusText}`);
+      throw new Error(
+        `Failed to continue task: ${response.status} ${error.error || response.statusText}`
+      );
     }
 
     return response.json();
--- a/webui/src/utils/taskApi.ts
+++ b/webui/src/utils/taskApi.ts
@@ -62,7 +62,7 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to create task: ${response.statusText}`);
+      throw new Error(error.error || `Failed to create task: ${response.status} ${response.statusText}`);
     }
 
     return response.json();
@@ -77,7 +77,7 @@ export const taskApi = {
       if (response.status === 404) {
         throw new Error(`Task not found: ${taskId}`);
       }
-      throw new Error(`Failed to get task: ${response.statusText}`);
+      throw new Error(`Failed to get task: ${response.status} ${response.statusText}`);
     }
     return response.json();
   },
@@ -99,7 +99,7 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to update task: ${response.statusText}`);
+      throw new Error(error.error || `Failed to update task: ${response.status} ${response.statusText}`);
     }
 
     return response.json();
@@ -118,7 +118,7 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to archive task: ${response.statusText}`);
+      throw new Error(error.error || `Failed to archive task: ${response.status} ${response.statusText}`);
     }
   },
 
@@ -135,7 +135,7 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to unarchive task: ${response.statusText}`);
+      throw new Error(error.error || `Failed to unarchive task: ${response.status} ${response.statusText}`);
     }
   },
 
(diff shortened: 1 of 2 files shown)


webui/src/utils/taskApi.ts
```diff
@@ -62,7 +62,7 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
-      throw new Error(error.error || `Failed to create task: ${response.statusText}`);
+      throw new Error(error.error || `Failed to create task: ${response.status} ${response.statusText}`);
```
TimeToBuildBob: ⚠️ **P2** — The status code is only added to the fallback branch of the `||` expression. When the server returns a non-2xx response with a JSON body containing an `error` field, `error.error` is truthy and becomes the thrown message, so the template with `${response.status}` is never evaluated. Since structured error bodies are the normal way API frameworks report 4xx/5xx failures, the thrown Error message in createTask, updateTask, archiveTask, unarchiveTask and continueTask is just the server-provided text (e.g. `title is required`) with no status code — the exact information this PR claims to add. A caller debugging a failed request still cannot distinguish 400 from 409 from the message alone, so the stated goal of including the HTTP status code in all error messages is not met for the most common error path.

TimeToBuildBob: ⚠️ **P2** — The PR changes user-facing error message strings across seven methods but adds no test asserting the new format. There is no test verifying that a non-OK response yields a message containing the HTTP status code, so a future refactor could silently drop the status (as the `error.error ||` precedence in the parallel finding already does) without any test failing. Because the entire purpose of the change is the message format, an assertion that the thrown error contains `response.status` for each method would lock the behavior in; without it the change is only verified by manual inspection.

TimeToBuildBob:
## 🤖 AI code review

Standardizes the error messages of seven taskApi methods (createTask, getTask, updateTask, archiveTask, unarchiveTask, continueTask, getSuggestedActions) so each non-ok response includes the HTTP status code alongside the status text or the server's structured error. Adds a new Jest suite in taskApi.test.ts that asserts the status-inclusive message format for the structured-error, 404, non-404, and non-JSON-body paths.

### Confidence Score: 4/5 — Could merge, with caveats

No blocking defects on this head, but 1 P2 issue worth resolving first.

**1 finding** · ⚠️ **1** P2 — none blocking, so the findings below are advisory.

**⚠️ P2 medium** — `webui/src/utils/taskApi.ts:80`

The 404 branch of getTask changes the exact thrown string from `Task not found: ${taskId}` to `Task not found: ${taskId} (404 Not Found)`, and the five structured-error branches (createTask/updateTask/archiveTask/unarchiveTask/continueTask) change from emitting the server's `error.error` verbatim to prefixing it with `Failed to <X>: ${status} `. Any caller that matches on the old text — e.g. a regex anchored like `/^Task not found: (.+)$/` to extract the task id, or an exact `err.message === 'Task not found: ' + id` check used to decide whether to auto-create the task or refresh the list — will silently stop matching. That branch then never fires and execution falls through to generic error handling, changing behavior for any consumer that relied on the previous message shape. The diff updates no existing caller or old assertion, so the break is unmitigated.

<sub>**How this was verified:** Diff shows only taskApi.ts and the new test file changed; the prior message strings (`Task not found: ${taskId}`, bare `error.error`) are not preserved anywhere and no existing test/consumer is updated in this PR.</sub>

</details>

- `webui/src/utils/taskApi.ts` — Rewrites the non-ok error branches of seven taskApi methods to prepend response.status to the thrown message; the getTask 404 message now appends the status in parentheses.
- `webui/src/utils/__tests__/taskApi.test.ts` — Adds a new Jest test file covering the status-inclusive error message format and the non-JSON fallback for the affected taskApi methods.

</details>

| `e7db721bc924` | 4/5 | 2 | llm | 2026-08-10 15:31 UTC |
| `1abc47a83c53` | 4/5 | 2 | llm | 2026-08-10 16:43 UTC |

</details>

<sub>Reviewed `55f12c3c036c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 107s · advisory, a human still decides · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

webui/src/utils/taskApi.ts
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 1abc47a83: structured API errors now retain the operation context and HTTP status, and the dedicated 404 path includes `404 Not Found`. Added focused regression coverage for create/update/archive/unarchive/continue plus task-not-found.

Verified: 6 focused Jest tests pass, TypeScript typecheck passes, and Prettier/pre-commit checks pass.

TimeToBuildBob: Valid — fixed in `1abc47a8`.

You were right that the status code was inside the fallback branch of the `||`, so a structured `{error: ...}` body (the normal 4xx/5xx shape) short-circuited past it and the thrown message carried no status. The status is now outside the `||`, applied to both branches:

```ts
throw new Error(
  `Failed to create task: ${response.status} ${error.error || response.statusText}`
);
```

Same shape in `createTask`, `updateTask`, `archiveTask`, `unarchiveTask` and `continueTask`, so the structured-error path now carries the status too. Resolving once the re-review confirms.

TimeToBuildBob: Valid — fixed in `1abc47a8`.

`webui/src/utils/__tests__/taskApi.test.ts` is new in this PR and locks the message format down. It table-drives the five structured-error methods and asserts the status is present:

```ts
await expect(request()).rejects.toThrow(`${prefix}: 409 task state conflict`);
```

plus a dedicated case for the 404 task-not-found message. A refactor that dropped the status — including a regression of the `error.error ||` precedence bug in the parallel finding — now fails these tests rather than passing silently. Resolving once the re-review confirms.

webui/src/utils/taskApi.ts
```diff
@@ -167,7 +177,7 @@ export const taskApi = {
       getFetchOptions()
     );
     if (!response.ok) {
-      throw new Error(`Failed to get task actions: ${response.statusText}`);
+      throw new Error(`Failed to get task actions: ${response.status} ${response.statusText}`);
```
TimeToBuildBob, line 180: ⚠️ **P2** — The new test file covers the structured-error path for createTask, updateTask, archiveTask, unarchiveTask, continueTask, and getTask's 404 branch, but never exercises two error paths this PR modified: getSuggestedActions (line 180) and getTask's non-404 branch (line 82). Both now embed response.status in the thrown message, yet no test asserts that format. Because the PR's stated scope is standardizing all seven functions, a regression that drops or misplaces the status code in either untested path would pass the suite undetected. The getTask non-404 branch is especially easy to break since it sits next to the tested 404 branch and shares the same if/else structure.

webui/src/utils/taskApi.ts
```diff
@@ -62,7 +62,9 @@ export const taskApi = {
 
     if (!response.ok) {
       const error = await response.json().catch(() => ({ error: response.statusText }));
```
TimeToBuildBob, line 64: ⚠️ **P2** — The new tests only exercise the branch where response.json() resolves to an object with an error field. The defensive fallback response.json().catch(() => ({ error: response.statusText })) — triggered when the server returns a non-JSON error body, such as an HTML error page or a proxy's text 502 — is never mocked in the tests. In that path the new message becomes 'Failed to create task: {status} {statusText}' (e.g. 'Failed to create task: 502 Bad Gateway'), a different string from the structured case that the PR's goal of consistent status-code inclusion depends on. A regression in this fallback formatting would go undetected.

webui/src/utils/taskApi.ts
TimeToBuildBob: Fixed in 55f12c3c0: added `getSuggestedActions` error test (403 Forbidden) and `getTask` non-404 error test (500 Internal Server Error). All 9 tests pass.

webui/src/utils/taskApi.ts
TimeToBuildBob: Fixed in 55f12c3c0: added a `nonJsonErrorResponse` helper that makes `json()` reject, and a test asserting `createTask` throws `'Failed to create task: 502 Bad Gateway'` when the server returns a non-JSON body. All 9 tests pass.

TimeToBuildBob:
@greptileai review

webui/src/utils/taskApi.ts
```diff
@@ -75,9 +77,9 @@ export const taskApi = {
     const response = await fetch(`${getApiBaseUrl()}/api/v2/tasks/${taskId}`, getFetchOptions());
     if (!response.ok) {
       if (response.status === 404) {
-        throw new Error(`Task not found: ${taskId}`);
+        throw new Error(`Task not found: ${taskId} (${response.status} ${response.statusText})`);
```
TimeToBuildBob, line 80: ⚠️ **P2** — The 404 branch of getTask changes the exact thrown string from `Task not found: ${taskId}` to `Task not found: ${taskId} (404 Not Found)`, and the five structured-error branches (createTask/updateTask/archiveTask/unarchiveTask/continueTask) change from emitting the server's `error.error` verbatim to prefixing it with `Failed to <X>: ${status} `. Any caller that matches on the old text — e.g. a regex anchored like `/^Task not found: (.+)$/` to extract the task id, or an exact `err.message === 'Task not found: ' + id` check used to decide whether to auto-create the task or refresh the list — will silently stop matching. That branch then never fires and execution falls through to generic error handling, changing behavior for any consumer that relied on the previous message shape. The diff updates no existing caller or old assertion, so the break is unmitigated.

TimeToBuildBob: Checked the actual callers: `stores/tasks.ts` is the only consumer of taskApi functions, and it does not do string matching on error messages — errors are passed to React Query and surfaced as opaque Error objects for display. No caller has a pattern like `err.message === 'Task not found: ' + id` or a regex matching the old format. The finding is valid as a general concern but does not apply here. The change is safe to land.

PR #3517, opened 2026-08-10
fix(constants): escape assistant display name to prevent rich markup injection

YuriNachos:
## What

`prompt_assistant` interpolated the agent display name verbatim, while its sibling `prompt_user` (11 lines above) escapes it with `rich.markup.escape`.

## Why it's a bug

The name is user-controlled — it comes from `config.chat.agent_config.name` or the `GPTME_AGENT_NAME` env var. All four callers feed the result straight into `rprint`, which parses Rich markup:

```python
rprint(f"\{prompt_assistant(agent_name)\}: ...")   # llm/__init__.py:342,354,707,782
```

So a name like `Agent [v2]` has `[v2]` parsed as a (malformed) markup tag and **swallowed from the rendered output**. Verified directly:

| input name `Agent [v2]` | rendered text |
|---|---|
| before fix (verbatim) | `Agent ` ← `[v2]` gone |
| after fix (`escape`)  | `Agent [v2]` ← correct |

## Fix

Mirror `prompt_user`: `from rich.markup import escape` and wrap only the interpolated `name`. The surrounding `[bold ...]` style tags stay literal. `escape()` only backslash-escapes markup chars, so it cannot break any existing name that doesn't contain `[`/`]`.

## How tested

- New `tests/test_constants.py`: asserts the bracket is escaped in the raw string, the style tags stay literal, and a Rich round-trip renders `Agent [v2]` in full (the regression — buggy path drops `[v2]`).
- `uv run pytest tests/test_constants.py` → 3 passed; `ruff check`/`format` clean. No existing test asserts the unescaped output (the `prompt_user` referenced in `test_prompt_templates.py` is a different function from `gptme.prompts`).

### DIFF
--- a/gptme/constants.py
+++ b/gptme/constants.py
@@ -42,9 +42,11 @@ PROMPT_USER = prompt_user()
 
 
 def prompt_assistant(name: str | None) -> str:
+    from rich.markup import escape
+
     if not name:
         name = os.environ.get("GPTME_AGENT_NAME", "Assistant")
-    return f"[bold {ROLE_COLOR['assistant']}]{name}[/bold {ROLE_COLOR['assistant']}]"
+    return f"[bold {ROLE_COLOR['assistant']}]{escape(name)}[/bold {ROLE_COLOR['assistant']}]"
 
 
 INTERRUPT_CONTENT = "Interrupted by user"
--- a/tests/test_constants.py
+++ b/tests/test_constants.py
@@ -1,0 +1,43 @@
+"""Tests for prompt helpers in gptme.constants."""
+
+from rich.console import Console
+
+from gptme.constants import prompt_assistant, prompt_user
+
+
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Thanks, @YuriNachos — this is a sharp, minimal fix for a surprisingly visible edge case.

I checked merged head `c5450f069` locally:

```text
.venv/bin/pytest tests/test_constants.py -q
3 passed in 0.79s
```

I also round-tripped `prompt_assistant("Agent [v2]")` through a Rich `Console`; the rendered prompt contains the literal `Agent [v2]`. That directly verifies the regression: user-controlled brackets no longer disappear as markup, while the surrounding prompt style still renders normally. Nice catch, and thank you for pairing the one-line `escape(name)` change with an end-to-end rendering assertion rather than only testing the raw string.

PR #3521, opened 2026-08-11
fix(server): preserve webui html hint on config patch

TimeToBuildBob:
## What changed
Preserve the webui HTML output-format hint when `PATCH /api/v2/conversations/<id>/config` rebuilds a conversation's leading system messages.

## Why
A pure config update like a conversation rename was silently stripping the existing webui hint:
`Output format: you are in a web interface ... use ```html blocks ...````.
That changed later assistant behavior even though the user had only saved settings.

## Root cause
Conversation creation appended the webui HTML hint, but config PATCH rebuilt the leading system messages from `get_prompt(...)` and the conversation system prompt only. The hint was never re-added, so rename/settings saves shortened the system preamble and dropped the HTML capability instruction.

## Impact
- Renaming or saving settings no longer changes webui rendering guidance for an existing conversation.
- Existing conversations that started with the webui hint keep it after config PATCH.

## Validation
- Reproduced twice against Bob's live `gptme-server` on port 5700: create conversation -> rename via config PATCH -> hint disappeared before this fix.
- `python -m pytest tests/test_server_v2.py -k 'webui_html_hint or chat_config_patch_preserves_webui_html_hint' -q`

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -319,6 +319,22 @@ def _append_conversation_system_prompt(
         messages.append(Message("system", system_prompt))
 
 
+_WEBUI_HTML_HINT = (
+    "Output format: you are in a web interface. "
+    "For interactive content (dashboards, tables, diagrams), "
+    "use ```html blocks — they render live in a sandboxed preview panel."
+)
+
+
+def _append_webui_html_hint(messages: list[Message], *, prompt: str = "full") -> None:
+    """Append the web UI HTML hint when this conversation uses it."""
+    if os.environ.get("GPTME_SERVE_HTML_HINT", "true").lower() == "false":
+        return
+    if prompt == "none":
+        return
+    messages.append(Message("system", _WEBUI_HTML_HINT))
+
+
 def _resolve_conversation_system_prompt(chat_config: ChatConfig) -> str | None:
     """Return the conversation prompt, falling back to the server default profile."""
     if chat_config.system_prompt:
@@ -1709,18 +1725,7 @@ def api_conversation_put(conversation_id: str):
     # Code/Preview tabs — inform the model so it can use HTML when it provides
     # a richer experience for the reader.
     # Set GPTME_SERVE_HTML_HINT=false to disable for non-webui API clients.
-    if (
-        os.environ.get("GPTME_SERVE_HTML_HINT", "true").lower() != "false"
-        and prompt != "none"
-    ):
-        msgs.append(
-            Message(
-                "system",
-                "Output format: you are in a web interface. "
-                "For interactive content (dashboards, tables, diagrams), "
-                "use ```html blocks — they render live in a sandboxed preview panel.",
-            )
-        )
+    _append_webui_html_hint(msgs, prompt=prompt)
 
     resolved_prompt = _resolve_conversation_system_prompt(chat_config)
     # Persist the resolved prompt back to chat_config so it survives server
@@ -2847,7 +2852,11 @@ def api_conversation_config_patch(conversation_id: str):
                 if m.role != "system":
                     break
                 first_non_system += 1
+            existing_system_msgs = manager.log.messages[:first_non_system]
             remaining_msgs = list(manager.log.messages[first_non_system:])
+            had_webui_html_hint = any(
+                m.content == _WEBUI_HTML_HINT for m in existing_system_msgs
+            )
 
             new_system_msgs = list(
                 get_prompt(
@@ -2859,6 +2868,8 @@ def api_conversation_config_patch(conversation_id: str):
                     agent_path=chat_config.agent,
                 )
             )
+            if had_webui_html_hint:
+                _append_webui_html_hint(new_system_msgs)
             _append_conversation_system_prompt(
                 new_system_msgs, _resolve_conversation_system_prompt(chat_config)
             )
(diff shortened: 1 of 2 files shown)


gptme/server/api_v2.py
```diff
@@ -2855,6 +2864,8 @@ def api_conversation_config_patch(conversation_id: str):
                     agent_path=chat_config.agent,
                 )
             )
+            if had_webui_html_hint:
+                _append_webui_html_hint(new_system_msgs)
```
TimeToBuildBob: ⚠️ **P2** — When a conversation was created with `GPTME_SERVE_HTML_HINT` unset or set to 'true' (so the webui hint is present), and the operator later sets the env var to 'false' for new conversations, any subsequent config PATCH (e.g., a rename) silently strips the hint. The patch calls `_append_webui_html_hint(new_system_msgs)`, which returns early when the env var is currently 'false', so the hint is not re-added even though it existed in the old system messages. This contradicts the PR's stated goal of preserving the hint across config patches: a user who merely renames or saves settings then loses the HTML guidance for an existing session. The env var should only gate addition at creation time, not retroactively remove the hint from conversations that already have it.

TimeToBuildBob:
## 🤖 AI code review

Introduces a shared `_append_webui_html_hint` helper and `_WEBUI_HTML_HINT` constant for the web UI HTML output-format hint. Refactors `api_conversation_put` to use it, and updates `api_conversation_config_patch` to re-append the hint when it was present in the existing leading system messages, so config updates like renames no longer strip it. Adds a test covering the preservation.

- `gptme/server/api_v2.py` — Adds `_WEBUI_HTML_HINT` constant and `_append_webui_html_hint` helper; uses it in `api_conversation_put` and `api_conversation_config_patch` to preserve the hint across config PATCH.
- `tests/test_server_v2.py` — Adds `test_v2_chat_config_patch_preserves_webui_html_hint` verifying the hint survives a config rename patch.

</details>

| `ccf8f117d30d` | 4/5 | 1 | llm | 2026-08-20 10:41 UTC |
| `55440166032f` | 4/5 | 2 | llm | 2026-08-23 06:20 UTC |

</details>

<sub>Reviewed `ce5f6b02ef92` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 91s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/server/api_v2.py
TimeToBuildBob: Fixed in 554401660: added `preserve=True` flag to `_append_webui_html_hint` that bypasses the env-var check when re-adding a hint that was already present. The env var now only gates creation, not preservation across config PATCHes.

TimeToBuildBob:
CI green (all checks pass), Greptile 5/5. Self-merge not eligible — `gptme/server/api_v2.py` is outside the auto-merge allowlist. Ready for maintainer merge.

TimeToBuildBob:
@greptileai review

gptme/server/api_v2.py
```diff
@@ -2847,7 +2861,11 @@ def api_conversation_config_patch(conversation_id: str):
                 if m.role != "system":
                     break
                 first_non_system += 1
+            existing_system_msgs = manager.log.messages[:first_non_system]
             remaining_msgs = list(manager.log.messages[first_non_system:])
+            had_webui_html_hint = any(
+                m.content == _WEBUI_HTML_HINT for m in existing_system_msgs
```
TimeToBuildBob: ⚠️ **P2** — The preservation of the webui HTML hint in config PATCH relies on an exact string match against _WEBUI_HTML_HINT. If the hint message is edited (e.g., via the message edit endpoint) to add a trailing space, a markdown emphasis, or any wording tweak, had_webui_html_hint becomes False and the hint is silently dropped when the leading system messages are rebuilt from get_prompt. This defeats the purpose of the fix for any user who customized the hint content. A substring or normalized match would be more robust and ensure preservation even when the hint is slightly modified.

tests/test_server_v2.py
```diff
@@ -2528,6 +2528,53 @@ def test_v2_chat_config_system_prompt_roundtrip_and_clear(client: FlaskClient):
     assert system_prompt not in cleared_system_messages
 
 
+def test_v2_chat_config_patch_preserves_webui_html_hint(
+    client: FlaskClient, monkeypatch: pytest.MonkeyPatch
+):
+    """Renaming via config PATCH must not strip the webui HTML hint."""
+    monkeypatch.setenv("GPTME_SERVE_HTML_HINT", "true")
```
TimeToBuildBob, line 2535: ⚠️ **P2** — The test test_v2_chat_config_patch_preserves_webui_html_hint sets GPTME_SERVE_HTML_HINT=true but does not isolate other environment variables that affect conversation creation, notably GPTME_CHAT_HISTORY. If the caller's environment has GPTME_CHAT_HISTORY=true, the created conversation may contain an extra system message from history context, causing the assertion `len(before_system_msgs) == 2` to fail and making the test flaky or dependent on the CI environment. To be hermetic, the test should also set GPTME_CHAT_HISTORY=false (or fully isolate the config as other tests do, e.g., by setting a clean Config and patching workspace config_path).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `ce5f6b02ef92`:

- CI gate: satisfied
- Review threads: clear
- `gptme/server/api_v2.py`

PR #3527, opened 2026-08-12
fix(llm): normalize proxy URL in _is_proxy() to match init() suffix rule

### LINKED ISSUE #3526: bug(llm): _is_proxy() never matches unless LLM_PROXY_URL ends in /messages
Split out from a docs PR discussion with @TimeToBuildBob (#3508) — filing separately since it's a source-side bug, not a docs one.

## Summary

`_is_proxy()` compares the client's `base_url` against the **raw** `LLM_PROXY_URL`, but `init()` has already appended `/messages` to the URL used to build the client. The two are therefore never equal unless the user's `LLM_PROXY_URL` already ends in `/messages`, so `_is_proxy()` silently returns `False` for the natural way to write the variable (`https://proxy.example.com` / `.../v1`).

## Repro

```python
# repro_is_proxy.py
import os, sys

url = sys.argv[1]
os.environ["LLM_PROXY_URL"] = url
os.environ["OPENAI_API_KEY"] = "sk-test"   # any value; no request is made

from gptme.config import get_config
from gptme.llm import llm_openai

llm_openai.init("openai", get_config())
client = llm_openai.get_client("openai")

print(f"LLM_PROXY_URL   = {url}")
print(f"client.base_url = {client.base_url}")
print(f"_is_proxy()     = {llm_openai._is_proxy(client)}")
```

```console
$ python repro_is_proxy.py 'http://127.0.0.1:8080'
LLM_PROXY_URL   = http://127.0.0.1:8080
client.base_url = http://127.0.0.1:8080/messages/
_is_proxy()     = False          # ← expected True

$ python repro_is_proxy.py 'http://127.0.0.1:8080/'
LLM_PROXY_URL   = http://127.0.0.1:8080/
client.base_url = http://127.0.0.1:8080//messages/
_is_proxy()     = False          # ← expected True (note the doubled slash)

$ python repro_is_proxy.py 'http://127.0.0.1:8080/messages'
LLM_PROXY_URL   = http://127.0.0.1:8080/messages
client.base_url = http://127.0.0.1:8080/messages/
_is_proxy()     = True           # only this spelling works
```

Run against gptme 0.32.1 (PyPI). Current `master` has the same logic.

## Root cause

`gptme/llm/llm_openai.py` `init()` normalizes the URL before building the client:

```python
proxy_url = config.get_env("LLM_PROXY_URL")

# Set the proxy URL to the unified messages endpoint if not already set
if proxy_url and not proxy_url.endswith("/messages"):
    proxy_url = proxy_url + "/messages"      # llm_openai.py:487
```

but `_is_proxy()` re-reads the env var **unnormalized** and compares:

```python
@lru_cache(maxsize=2)
def _is_proxy(client: OpenAI) -> bool:
    proxy_url = get_config().get_env("LLM_PROXY_URL")
    if not proxy_url:
        return False
    # Normalize URLs for comparison (remove trailing slashes)
    return str(client.base_url).rstrip("/") == proxy_url.rstrip("/")   # llm_openai.py:695
```

`rstrip("/")` handles the trailing slash but not the appended `/messages` segment, so the comparison misses.

## Impact

`is_proxy` feeds two behaviours, both of which silently take the wrong branch:

1. **Proxy-aware model naming** — `_gptme_api_model()` returns `model if is_proxy else base_model`. Its docstring says the provider prefix exists "to make proxy aware of the provider", so with `is_proxy=False` the proxy receives a bare `gpt-4o` instead of `openai/gpt-4o` and loses the routing hint.
2. **Responses-API opt-out** — `_should_use_responses_api()` requires `not _is_proxy(client)`. With `is_proxy=False` the check passes, so gptme uses the Responses API against a proxy the opt-out was written to avoid (posting to `<proxy>/messages/responses`).

Neither surfaces an error, which is what makes it easy to miss: a proxy user gets subtly different wire behaviour depending on whether they happened to write `/messages` at the end of the env var.

## Suggested fix

Normalize once, in one place. Either derive the comparison value with the same rule `init()` uses:

```python
proxy_url = config.get_env("LLM_PROXY_URL")
if proxy_url and not proxy_url.endswith("/messages"):
    proxy_url += "/messages"
```

or — probably cleaner — have `init()` record the effective proxy base URL in a module-level variable and let `_is_proxy()` compare against that, so the suffix rule can never drift between the two sites again. (Note `_is_proxy` is `lru_cache`d, so whatever it compares against needs to be stable for the process.)

Happy to send a PR if you'd like it fixed this way.

## Environment

- gptme 0.32.1 (PyPI), Python 3.12.3, Windows — logic verified unchanged on `master`
- No proxy server needed to reproduce; the client is never used to send a request

TimeToBuildBob:
Fixes #3526.

## Root cause

`init()` appends `/messages` to `LLM_PROXY_URL` before building the OpenAI client, but `_is_proxy()` compared `client.base_url` against the raw env var. The two could never be equal unless the user already wrote `/messages` at the end of their URL.

## Fix

Apply the same `/messages` normalization in `_is_proxy()` before comparing, mirroring exactly what `init()` does. This keeps the logic in one place without introducing a module-level variable.

## Verification

All three natural spellings from the bug report now return `True`:

| `LLM_PROXY_URL` | `client.base_url` | `_is_proxy()` |
|---|---|---|
| `http://127.0.0.1:8080` | `http://127.0.0.1:8080/messages/` | ✅ `True` |
| `http://127.0.0.1:8080/` | `http://127.0.0.1:8080//messages/` | ✅ `True` |
| `http://127.0.0.1:8080/messages` | `http://127.0.0.1:8080/messages/` | ✅ `True` |

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -691,6 +691,10 @@ def _is_proxy(client: OpenAI) -> bool:
     # If client has the proxy URL set, it is using the proxy
     if not proxy_url:
         return False
+    # Apply the same /messages normalization that init() applies before building the client,
+    # so the comparison never fails due to a missing suffix.
+    if not proxy_url.endswith("/messages"):
+        proxy_url = proxy_url + "/messages"
     # Normalize URLs for comparison (remove trailing slashes)
     return str(client.base_url).rstrip("/") == proxy_url.rstrip("/")
 
--- a/tests/test_llm_openai.py
+++ b/tests/test_llm_openai.py
@@ -3429,3 +3429,43 @@ class TestMakeResolvedModel:
             f"openrouter/moonshotai/kimi-k2.5@{suffix}", "Moonshot AI"
         )
         assert result is None
+
+
+class TestIsProxy:
+    """Direct unit tests for _is_proxy() URL comparison — regression for #3526."""
+
+    @pytest.mark.parametrize(
+        ("proxy_url", "client_base_url"),
+        [
+            # no suffix: init() appends /messages, httpx normalizes with trailing slash
+            ("http://127.0.0.1:8080", "http://127.0.0.1:8080/messages/"),
+            # trailing slash: double slash preserved in client base_url
+            ("http://127.0.0.1:8080/", "http://127.0.0.1:8080//messages/"),
+            # already has /messages: init() leaves it unchanged, httpx adds trailing slash
+            ("http://127.0.0.1:8080/messages", "http://127.0.0.1:8080/messages/"),
+        ],
+    )
+    def test_proxy_url_spellings_return_true(
+        self, monkeypatch, proxy_url, client_base_url
+    ):
+        from gptme.llm.llm_openai import _is_proxy
+
+        _is_proxy.cache_clear()
+        monkeypatch.setenv("LLM_PROXY_URL", proxy_url)
+        client = MagicMock()
+        client.base_url = client_base_url
+        assert _is_proxy(client), (
+            f"LLM_PROXY_URL={proxy_url!r} should be detected as proxy"
+        )
+        _is_proxy.cache_clear()
+
+    def test_no_proxy_url_returns_false(self, monkeypatch):
+        from gptme.llm.llm_openai import _is_proxy
+
+        _is_proxy.cache_clear()
+        monkeypatch.delenv("LLM_PROXY_URL", raising=False)
+        monkeypatch.delenv("GPTME_LLM_PROXY_URL", raising=False)
+        client = MagicMock()
+        client.base_url = "http://127.0.0.1:8080/messages/"
+        assert not _is_proxy(client)
+        _is_proxy.cache_clear()


gptme/llm/llm_openai.py
```diff
@@ -691,6 +691,10 @@ def _is_proxy(client: OpenAI) -> bool:
     # If client has the proxy URL set, it is using the proxy
     if not proxy_url:
         return False
+    # Apply the same /messages normalization that init() applies before building the client,
+    # so the comparison never fails due to a missing suffix.
+    if not proxy_url.endswith("/messages"):
```
TimeToBuildBob, line 696: ⚠️ **P2** — The PR adds normalization logic to `_is_proxy()` to handle three URL spellings (e.g., `http://host`, `http://host/`, `http://host/messages`), as shown in the PR description's verification table, but no automated tests accompany the change. The only verification is a manual table. This means the fix is unprotected against regressions: a future refactor of `init()` or `_is_proxy()` could easily reintroduce the mismatch without any test catching it. Specifically, the new `if not proxy_url.endswith("/messages"):` branch is untested for all the edge cases listed, including the case where the URL already ends with `/messages` and the case with a trailing slash. Adding a parametrized unit test that asserts `_is_proxy(client)` returns `True` for each of the three spellings (using a fake client with the corresponding `base_url`) would lock in the intended behavior. This is a P2 because the manual verification covers the happy path, but automated coverage is missing for a bug that took several releases to surface.

TimeToBuildBob:
## 🤖 AI code review

This PR updates `_is_proxy()` in `gptme/llm/llm_openai.py` to append the `/messages` suffix to the `LLM_PROXY_URL` environment variable before comparing it with the client's `base_url`, mirroring the normalization already done in `init()`. This fixes proxy detection when the environment variable is set without the `/messages` suffix. It adds unit tests in `tests/test_llm_openai.py` covering three URL spellings and a no-proxy case.

- `gptme/llm/llm_openai.py` — Adds `/messages` normalization to `_is_proxy()` before URL comparison, matching `init()`'s behavior.
- `tests/test_llm_openai.py` — Adds `TestIsProxy` class with parametrized tests for proxy URL spellings and a no-proxy test.

</details>

| `f003569d4b31` | 4/5 | 2 | llm | 2026-08-12 03:44 UTC |

</details>

<sub>Reviewed `2bfd68ad1b04` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 71s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/llm/llm_openai.py
```diff
@@ -691,6 +691,10 @@ def _is_proxy(client: OpenAI) -> bool:
...
+        proxy_url = proxy_url + "/messages"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 2bfd68ad1 — added TestIsProxy class with parametrized tests covering all three LLM_PROXY_URL spellings (no suffix, trailing slash, already has /messages) plus the no-proxy case.

gptme/llm/llm_openai.py
TimeToBuildBob: Fixed in 2bfd68ad1 — added TestIsProxy class with parametrized tests for all three URL spellings and the no-proxy case.

TimeToBuildBob:
@greptileai review

PR #3519, opened 2026-08-10
fix: correct editable-install detection in `--version` output (#3518)

TimeToBuildBob:
## Problem

Two bugs caused `--version` to misreport in regular PyPI installs (e.g. `uvx gptme@latest`):

1. **`+unknown` version suffix** — `v0.32.1+unknown` instead of `v0.32.1`
2. **`[uv (editable)]`** in the second line — should be `[uv]`

Root cause: both `__version__.py` and `info.py` used `isinstance(dist, PathDistribution)` to detect editable installs. But `PathDistribution` is the concrete type for **all** pip/uv-installed packages, not just editable ones. A regular `uvx gptme@latest` install was incorrectly flagged as editable, then `get_git_version()` failed (no git repo in site-packages), and `+unknown` was appended.

## Fix

Use `direct_url.json` / `dir_info.editable` for the editable check — this is the PEP 660 / pip metadata standard that correctly distinguishes editable from regular installs.

- **`gptme/__version__.py`**: replace `isinstance(dist, PathDistribution)` with a `direct_url.json` check
- **`gptme/info.py`**: remove the `type(dist).__name__ == "PathDistribution" → editable = True` heuristic

## Scripting / machine-readable output

Per issue request, also unhide `--version-json` (was `hidden=True`) so scripts have a stable machine-readable interface:

```bash
gptme --version-json | jq .version   # "0.32.1"
```

The `--version` human output already has a stable first-line contract (`gptme v<version>`) — documented in the help text and pinned by a new test.

## Tests

- Fixed `test_path_distribution_editable` → `test_path_distribution_not_editable` (was asserting the buggy behavior)
- Added `TestVersionOutputContract` with three tests:
  - first line always starts with `gptme v`
  - non-editable install doesn't show `(editable)`  
  - JSON output always has a `version` key

All 97 tests in `test_info.py` pass.

Closes #3518

Issue #3518, opened 2026-08-10
`--version` output is multi-line and environment-dependent — hard to consume from scripts/integrations


chernistry:
**Version**: gptme 0.32.1 (PyPI), macOS (Darwin), Python 3.14

**Repro**

```console
$ uvx gptme@latest --version
gptme v0.32.1+unknown
Python 3.14.6 on Darwin [uv (editable)]
Logs: /Users/<user>/Library/Application Support/gptme/logs

Extras:
  ✓ server
Tools: 30 available

Health: ⚠️  1 warnings (run `gptme-doctor` for details)

Utilities: gptme-util (run 'gptme-util --help' for more)
```

**Problem**

Two things make `--version` hard to consume from scripts and integrations:

1. The output is multi-line and environment-dependent — the line count changes with health warnings and installed extras, so the common `ver=$(gptme --version)` probe used to pin/verify an install needs ad-hoc parsing instead of a stable contract.
2. In a clean ephemeral PyPI install (`uvx gptme@latest`, no git checkout involved), the version reports `v0.32.1+unknown` and the install method shows as `[uv (editable)]` — both look like the detection logic misreporting outside a persistent venv / source checkout.

**Ask**

Any of these would resolve it (happy to send a PR if you have a preference):

1. Document/guarantee that the first line stays exactly `gptme v<version>`, or
2. Add a plain machine-readable variant (e.g. `--version --plain`, or `gptme-util version --json`), keeping the rich human output as the default, or
3. Move the Extras/Health/Utilities hints to `gptme-doctor` only.

Context: bernstein (github.com/sipyourdrink-ltd/bernstein) ships an adapter driving gptme's non-interactive mode; we hit this while re-verifying the adapter against 0.32.1.

TimeToBuildBob:
Seen. Investigating now; I'll reply here with the concrete work, task, or handoff once it's done.

TimeToBuildBob:
Fixed in #3519 (merged ~14 min ago, now on master).

Root cause: `isinstance(dist, PathDistribution)` was used to detect editable installs, but `PathDistribution` is the concrete type for *all* pip/uv installs — regular `uvx gptme@latest` installs were falsely flagged as editable, causing `+unknown` and `[uv (editable)]`.

What shipped:
- **`+unknown` gone**: non-editable PyPI installs now report the clean version (e.g. `0.32.1`)
- **`[uv (editable)]` gone**: correctly shows `[uv]` for tool installs
[...]

PR #3528, opened 2026-08-12
feat(cli): add JSON output to status command

TimeToBuildBob:
## Summary

- add `gptme-util status --json` using the existing CLI-wide boolean `--json` convention
- expose active tasks, commits, PR queue, disk usage, journals, and Bob-only service/blocker/ready-task fields as structured JSON
- reject ambiguous combinations with `--no-markdown` or `--format table`
- preserve `-o/--output` so JSON can be written directly to a file

## Verification

- `poetry run python3 -m pytest tests/test_cli_status.py -q` (16 passed)
- `poetry run ruff check gptme/cli/cmd_status.py tests/test_cli_status.py`
- `poetry run mypy gptme/cli/cmd_status.py`
- targeted pre-commit hooks pass with a clean cache
- live `python3 -m gptme.cli.util status --json` output parses with `python3 -m json.tool`

Related: gptme/gptme#3518 (the original version-output scripting pain is fixed by gptme/gptme#3519; this adds the largest remaining structured-output gap found in the CLI audit).

### DIFF
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -9,6 +9,7 @@ health, blockers, and ready backlog items.
 
 from __future__ import annotations
 
+import functools
 import json
 import logging
 import os
@@ -38,7 +39,9 @@ def _run(cmd: list[str], *, timeout: int = 10) -> str:
         return ""
 
 
+@functools.lru_cache(maxsize=1)
 def _git_root() -> Path | None:
+    """Return the git root for the current working directory (cached per process)."""
     raw = _run([GIT_CMD, "rev-parse", "--show-toplevel"])
     return Path(raw) if raw else None
 
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -346,6 +346,7 @@ def _status_data() -> dict[str, object]:
     is_bob = _is_bob_workspace()
     root = _git_root()
     status_data: dict[str, object] = {
+        "timestamp": datetime.now(timezone.utc).isoformat(),
         "session_id": _session_id(),
         "active_tasks": [
             {"id": task.get("_id", task.get("id", "")), "title": task.get("title", "")}
@@ -516,7 +517,9 @@ def status(
 
         gptme-util status --json                # structured JSON
     """
-    if as_json and (not markdown or output_format != "narrative" or write):
+    if as_json and (
+        not markdown or output_format != "narrative" or (write and not output)
+    ):
         raise click.UsageError(
             "--json cannot be combined with --no-markdown, --format, or --write"
             " (use -o/--output to write JSON to a file)"
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -17,6 +17,7 @@ import shutil
 import subprocess
 from datetime import datetime, timezone
 from pathlib import Path
+from typing import TypedDict
 
 import click
 
@@ -80,12 +81,18 @@ def _recent_commits(n: int = 3) -> list[str]:
     return raw.splitlines() if raw else []
 
 
+class _PRQueueRow(TypedDict):
+    repo: str
+    count: int
+    cap: int | None
+
+
 def _pr_queue(
     repos: list[tuple[str, int | None]], author: str | None = None
-) -> list[dict[str, str]]:
+) -> list[_PRQueueRow]:
     if author is None:
         author = _gh_user() or "TimeToBuildBob"
-    rows: list[dict[str, str]] = []
+    rows: list[_PRQueueRow] = []
     for repo, cap in repos:
         prs_json = _run(
             [
@@ -110,14 +117,15 @@ def _pr_queue(
         except json.JSONDecodeError:
             continue
         count = len(prs)
-        cap_str = (
-            f"{count}/{cap}"
-            + (" ⚠ at limit" if cap is not None and count >= cap else "")
-            if cap is not None
-            else str(count)
-        )
-        rows.append({"repo": repo, "count": cap_str})
+        rows.append({"repo": repo, "count": count, "cap": cap})
     return rows
+
+
+def _pr_queue_display(count: int, cap: int | None) -> str:
+    """Format a PR count as a human-readable string with optional cap."""
+    if cap is not None:
+        return f"{count}/{cap}" + (" ⚠ at limit" if count >= cap else "")
+    return str(count)
 
 
 def _service_status() -> list[dict[str, str]]:
@@ -278,7 +286,9 @@ def section_pr_queue() -> str:
     if rows:
         lines.append("| Repo | Open |")
         lines.append("|------|------|")
-        lines.extend(f"| {row['repo']} | {row['count']} |" for row in rows)
+        for row in rows:
+            display = _pr_queue_display(row["count"], row["cap"])
+            lines.append(f"| {row['repo']} | {display} |")
     else:
         lines.append("- Unable to fetch PR data")
     return "\n".join(lines)
@@ -379,7 +389,11 @@ def build_table_document() -> str:
     # pending_prs
     pr_rows = _pr_queue(_TRACKED_REPOS)
     pending_prs = (
-        ", ".join(f"{r['repo']}:{r['count']}" for r in pr_rows) if pr_rows else "none"
+        ", ".join(
+            f"{r['repo']}:{_pr_queue_display(r['count'], r['cap'])}" for r in pr_rows
+        )
+        if pr_rows
+        else "none"
     )
 
     # waiting_for
@@ -502,9 +516,10 @@ def status(
 
         gptme-util status --json                # structured JSON
     """
-    if as_json and (not markdown or output_format != "narrative"):
+    if as_json and (not markdown or output_format != "narrative" or write):
         raise click.UsageError(
-            "--json cannot be combined with --no-markdown or --format"
+            "--json cannot be combined with --no-markdown, --format, or --write"
+            " (use -o/--output to write JSON to a file)"
         )
 
     if as_json:
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -274,13 +274,7 @@ def section_active_work(is_bob: bool = False) -> str:
 
 def section_pr_queue() -> str:
     lines = ["## PR Queue"]
-    tracked = [
-        ("gptme/gptme", 10),
-        ("gptme/gptme-cloud", 3),
-        ("ErikBjare/bob", None),
-        ("gptme/gptme-contrib", None),
-    ]
-    rows = _pr_queue(tracked)
+    rows = _pr_queue(_TRACKED_REPOS)
     if rows:
         lines.append("| Repo | Open |")
         lines.append("|------|------|")
@@ -329,6 +323,44 @@ def section_ready_next() -> str:
 # ── build ─────────────────────────────────────────────────────────────
 
 
+_TRACKED_REPOS: list[tuple[str, int | None]] = [
+    ("gptme/gptme", 10),
+    ("gptme/gptme-cloud", 3),
+    ("ErikBjare/bob", None),
+    ("gptme/gptme-contrib", None),
+]
+
+
+def _status_data() -> dict[str, object]:
+    """Collect status data shared by JSON and presentation renderers."""
+    is_bob = _is_bob_workspace()
+    root = _git_root()
+    status_data: dict[str, object] = {
+        "session_id": _session_id(),
+        "active_tasks": [
+            {"id": task.get("_id", task.get("id", "")), "title": task.get("title", "")}
+            for task in _active_tasks(3)
+        ],
+        "recent_commits": _recent_commits(3),
+        "pr_queue": _pr_queue(_TRACKED_REPOS),
+        "disk_usage": _disk_usage(root),
+        "journal_entries": _journal_entries(5),
+    }
+    if is_bob:
+        status_data.update(
+            services=_service_status(),
+            dead_timers=_dead_timers(),
+            blockers=_blockers(3),
+            ready_tasks=_ready_tasks(3),
+        )
+    return status_data
+
+
+def build_json_status() -> str:
+    """Build the structured JSON status document."""
+    return json.dumps(_status_data(), indent=2)
+
+
 def build_table_document() -> str:
     """Build a machine-readable markdown table of session state."""
     is_bob = _is_bob_workspace()
@@ -345,13 +377,7 @@ def build_table_document() -> str:
     last_commit = commits[0] if commits else "none"
 
     # pending_prs
-    tracked = [
-        ("gptme/gptme", 10),
-        ("gptme/gptme-cloud", 3),
-        ("ErikBjare/bob", None),
-        ("gptme/gptme-contrib", None),
-    ]
-    pr_rows = _pr_queue(tracked)
+    pr_rows = _pr_queue(_TRACKED_REPOS)
     pending_prs = (
         ", ".join(f"{r['repo']}:{r['count']}" for r in pr_rows) if pr_rows else "none"
     )
@@ -448,7 +474,14 @@ def build_document() -> str:
     default="narrative",
     help="Output format: narrative (default) or table.",
 )
-def status(write: bool, output: str | None, markdown: bool, output_format: str):
+@click.option("--json", "as_json", is_flag=True, help="Output status as JSON.")
+def status(
+    write: bool,
+    output: str | None,
+    markdown: bool,
+    output_format: str,
+    as_json: bool,
+) -> None:
     """Generate a portable operator handoff / session-status document.
 
     Produces a compact briefing: active work, PR queue, service health,
@@ -465,9 +498,18 @@ def status(write: bool, output: str | None, markdown: bool, output_format: str):
 
         gptme-util status --no-markdown         # plain-text output
 
-        gptme-util status --format table        # machine-readable table
+        gptme-util status --format table        # compact Markdown table
+
+        gptme-util status --json                # structured JSON
     """
-    if output_format == "table":
+    if as_json and (not markdown or output_format != "narrative"):
+        raise click.UsageError(
+            "--json cannot be combined with --no-markdown or --format"
+        )
+
+    if as_json:
+        doc = build_json_status()
+    elif output_format == "table":
         doc = build_table_document()
     else:
         doc = build_document()
@@ -482,7 +524,7 @@ def status(write: bool, output: str | None, markdown: bool, output_format: str):
         out_path = (root or Path.cwd()) / "status.md"
 
     if out_path:
-        out_path.write_text(doc)
+        out_path.write_text(doc, encoding="utf-8")
         click.echo(f"Written to {out_path}")
     else:
         click.echo(doc)
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -9,7 +9,6 @@ health, blockers, and ready backlog items.
 
 from __future__ import annotations
 
-import functools
 import json
 import logging
 import os
@@ -39,9 +38,8 @@ def _run(cmd: list[str], *, timeout: int = 10) -> str:
         return ""
 
 
-@functools.lru_cache(maxsize=1)
 def _git_root() -> Path | None:
-    """Return the git root for the current working directory (cached per process)."""
+    """Return the git root for the current working directory."""
     raw = _run([GIT_CMD, "rev-parse", "--show-toplevel"])
     return Path(raw) if raw else None
 
--- a/tests/test_cli_status.py
+++ b/tests/test_cli_status.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import json
+
 from click.testing import CliRunner
 
 import gptme.cli.cmd_status as cmd_status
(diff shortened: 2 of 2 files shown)


gptme/cli/cmd_status.py
```diff
@@ -329,6 +323,44 @@ def section_ready_next() -> str:
...
+    is_bob = _is_bob_workspace()
+    root = _git_root()
+    status_data: dict[str, object] = {
+        "session_id": _session_id(),
+        "active_tasks": [
+            {"id": task.get("_id", task.get("id", "")), "title": task.get("title", "")}
+            for task in _active_tasks(3)
+        ],
+        "recent_commits": _recent_commits(3),
+        "pr_queue": _pr_queue(_TRACKED_REPOS),
```
TimeToBuildBob, line 357: ⚠️ **P2** — The new JSON status includes `"pr_queue": _pr_queue(_TRACKED_REPOS)` which returns entries where the `count` field is a string such as `"2/10"` or `"2"` (constructed in `_pr_queue` as `cap_str`). The test `test_status_json` mocks `_pr_queue` to return an integer `count: 2`, so it does not exercise the real string type. Consumers of the JSON will receive a non-numeric string for PR counts, which prevents programmatic summation or comparison. If the repo has a cap, the value is even less structured, e.g., "2/10". This undermines the goal of a structured JSON output. The fix is to have `_pr_queue` return a separate numeric `count` and optionally `cap` fields, or change the JSON to use the numeric count.

gptme/cli/cmd_status.py
```diff
@@ -482,7 +524,7 @@ def status(write: bool, output: str | None, markdown: bool, output_format: str):
         out_path = (root or Path.cwd()) / "status.md"
```
TimeToBuildBob, line 543: ⚠️ **P2** — The `status` command permits `--json` combined with `--write`. When `--write` is used without `-o`, `out_path` is set to `(root or Path.cwd()) / "status.md"` (line 524), so the JSON document is written into a Markdown-suffixed file. The validation at line 505 only rejects `--json` with `--no-markdown` or `--format table`, not `--write`. This overwrites any existing `status.md` with raw JSON, corrupting the Markdown handoff file and confusing tooling that expects Markdown. Since `-o/--output` already provides an explicit path for the JSON, allowing `--write` with `--json` is ambiguous. Consider rejecting `--json` with `--write` or routing the output to a `.json` file when `--json` is used.

TimeToBuildBob:
## 🤖 AI code review

Adds a --json flag to the status command that emits a structured JSON document via new _status_data and build_json_status functions. Refactors _pr_queue to return typed rows and extracts _pr_queue_display for formatting. Adds validation to reject --json combined with incompatible flags, and writes files with explicit UTF-8 encoding.

- `gptme/cli/cmd_status.py` — Introduces _PRQueueRow, _pr_queue_display, _status_data, build_json_status, _TRACKED_REPOS, and a --json option with a guard against incompatible rendering flags.
- `tests/test_cli_status.py` — Adds tests exercising the --json output format, flag validation, and _pr_queue_display formatting.

</details>

| `2cbe2130d24c` | 4/5 | 2 | llm | 2026-08-12 04:40 UTC |
| `21f0576472c1` | 4/5 | 2 | llm | 2026-08-12 06:24 UTC |
| `dddab8369317` | 4/5 | 3 | llm | 2026-08-12 06:41 UTC |
| `e646ac788893` | 4/5 | 3 | llm | 2026-08-12 13:30 UTC |
| `a5e2195e2e2a` | 4/5 | 1 | llm | 2026-08-12 15:23 UTC |

</details>

<sub>Reviewed `c06d0fd0a49a` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 118s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/cli/cmd_status.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 21f057647 — `_pr_queue()` now returns a `_PRQueueRow` TypedDict with `count: int` and `cap: int | None` instead of a formatted string. Added `_pr_queue_display()` helper for the human-readable form used by the Markdown/table renderers. JSON consumers now get real numeric values they can sum and compare.

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in 21f057647 — added `or write` to the validation gate so `--json --write` now exits with a UsageError pointing users to `-o/--output` instead. Test added to `test_status_json_rejects_rendering_options`.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in 21f057647 — same root cause as the Greptile P1. `_pr_queue()` now returns a `_PRQueueRow` TypedDict with `count: int` and `cap: int | None`. The presentation-formatted string (`"2/10"`, `"2/10 ⚠ at limit"`) has been moved to a separate `_pr_queue_display()` helper used only by Markdown/table renderers. JSON output now contains real numeric values.

tests/test_cli_status.py
```diff
@@ -102,6 +104,77 @@ def test_status_format_table_via_util():
...
+        cmd_status,
+        "_ready_tasks",
+        lambda limit=3: [{"id": "ready", "name": "Ready task"}],
+    )
+    monkeypatch.setattr(cmd_status, "_session_id", lambda: "session-1")
+    monkeypatch.setattr(cmd_status, "_git_root", lambda: None)
+    monkeypatch.setattr(cmd_status, "_disk_usage", lambda _root=None: "1G / 2G (50%)")
+    monkeypatch.setattr(cmd_status, "_journal_entries", lambda limit=5: ["entry.md"])
+
+    result = CliRunner().invoke(status, ["--json"])
```
TimeToBuildBob, line 143: ⚠️ **P2** — Missing test coverage: the PR adds a `--json` flag and documents that `-o/--output` writes JSON to a file, but the test suite never combines `--json` with `-o`. The existing tests cover `--json` to stdout (`test_status_json`) and rejection of `--json` with `--write` (`test_status_json_rejects_rendering_options`), but there is no test that `--json -o out.json` produces a file containing valid JSON with the expected structured fields. A regression where the `-o` path falls through to narrative or table rendering, or where the file is written without JSON encoding, would go undetected. Adding a test that invokes `status` with `--json -o` and asserts the file contents parse as JSON and match the expected schema would lock in this documented behavior.

tests/test_cli_status.py
```diff
@@ -102,6 +104,77 @@ def test_status_format_table_via_util():
     assert "| session_id |" in result.output
 
 
+def test_status_json(monkeypatch):
+    """Verify --json emits structured status without Markdown."""
+    monkeypatch.setattr(cmd_status, "_is_bob_workspace", lambda: True)
```
TimeToBuildBob, line 110: ⚠️ **P2** — Missing test coverage: the `--json` output conditionally includes Bob-only fields (services, dead_timers, blockers, ready_tasks) only when `_is_bob_workspace()` returns True. The only JSON test (`test_status_json`) monkeypatches `_is_bob_workspace` to True, so it asserts these fields are present. There is no test for the non-Bob case, so a change that mistakenly includes these fields when not in Bob's workspace would not be caught. Add a test with `_is_bob_workspace` returning False and assert those four keys are absent from the JSON output, ensuring the conditional logic in `_status_data` remains correct.

TimeToBuildBob:
@greptileai review

tests/test_cli_status.py
TimeToBuildBob: Fixed in dddab8369 — added `test_status_json_with_output_file` which invokes `status --json -o path` and asserts the output file exists, parses as JSON, and contains the expected schema keys.

tests/test_cli_status.py
TimeToBuildBob: Fixed in dddab8369 — added `test_status_json_excludes_bob_fields_in_non_bob_workspace` which monkeypatches `_is_bob_workspace` to False and asserts that `services`, `dead_timers`, `blockers`, and `ready_tasks` are absent from the JSON output.

gptme/cli/cmd_status.py
```diff
@@ -465,9 +512,19 @@ def status(write: bool, output: str | None, markdown: bool, output_format: str):
 
         gptme-util status --no-markdown         # plain-text output
 
-        gptme-util status --format table        # machine-readable table
+        gptme-util status --format table        # compact Markdown table
+
+        gptme-util status --json                # structured JSON
     """
-    if output_format == "table":
+    if as_json and (not markdown or output_format != "narrative" or write):
```
TimeToBuildBob: ⚠️ **P2** — The validation at line 519 rejects --json whenever the --write flag is set, even if -o/--output is also provided and overrides the output path. Since -o takes precedence in the output selection (lines 535-539), a user running `gptme-util status --json --write -o out.json` gets a UsageError despite an unambiguous destination. The error message directs the user to use -o, but the combination is needlessly rejected. This breaks scripts that habitually pass both flags.

gptme/cli/cmd_status.py
```diff
@@ -110,16 +117,17 @@ def _pr_queue(
...
-        )
-        rows.append({"repo": repo, "count": cap_str})
+        rows.append({"repo": repo, "count": count, "cap": cap})
     return rows
 
 
+def _pr_queue_display(count: int, cap: int | None) -> str:
+    """Format a PR count as a human-readable string with optional cap."""
+    if cap is not None:
+        return f"{count}/{cap}" + (" ⚠ at limit" if count >= cap else "")
```
TimeToBuildBob, line 128: ⚠️ **P2** — The new _pr_queue_display() helper centralizes the formatting of PR counts with caps, including the warning suffix, and is now used by both section_pr_queue and build_table_document. None of the added tests exercise this function directly; the JSON tests monkeypatch _pr_queue entirely, and the table/narrative tests rely on indirect invocation. A regression in the formatting logic (such as the limit warning threshold) would not be caught by the current test suite, making the displayed counts and cap indicators vulnerable to silent breakage.

gptme/cli/cmd_status.py
```diff
@@ -329,6 +333,44 @@ def section_ready_next() -> str:
...
+    ("gptme/gptme-contrib", None),
+]
+
+
+def _status_data() -> dict[str, object]:
+    """Collect status data shared by JSON and presentation renderers."""
+    is_bob = _is_bob_workspace()
+    root = _git_root()
+    status_data: dict[str, object] = {
+        "session_id": _session_id(),
```
TimeToBuildBob, line 351: ⚠️ **P2** — Both build_table_document and build_document include a generation timestamp via datetime.now(timezone.utc), but _status_data() (which feeds build_json_status) has no timestamp field. Consumers of the JSON status therefore cannot tell when the data was captured, making it impossible to detect stale handoff information. For an operator status document, this omission reduces the usefulness of the new JSON output and is inconsistent with the other output formats.

TimeToBuildBob:
Addressed the 3 P2 issues from the AI review in e646ac788:

1. **Missing timestamp** (`_status_data()`): added `"timestamp"` ISO 8601 field — now consistent with the `build_table_document` and `build_document` formats. JSON consumers can detect stale handoffs.

2. **`_pr_queue_display()` untested**: added `test_pr_queue_display()` covering all four cases — `count < cap`, `count == cap`, `count > cap`, and `cap=None`. Formatting regressions (limit warning threshold, cap rendering) now caught directly.

3. **Overly strict `--write` validation**: relaxed the gate from `or write` to `or (write and not output)`. `--json --write -o path` is now accepted since `-o` provides an unambiguous destination; `--json --write` without `-o` still raises a UsageError.

CI re-running. Greptile was 5/5 on the prior head; will re-check once checks land.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in e646ac788 — relaxed the gate from `or write` to `or (write and not output)`. `--json --write -o path` is now accepted since `-o` provides an unambiguous destination; `--json --write` without `-o` still raises a UsageError.

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in e646ac788 — added `test_pr_queue_display()` covering all four cases: `count < cap`, `count == cap`, `count > cap`, and `cap=None`. Formatting regressions (limit warning threshold, cap rendering) are now caught directly.

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in e646ac788 — added `"timestamp"` ISO 8601 field to `_status_data()`, consistent with the `build_table_document` and `build_document` formats. JSON consumers can now detect stale handoff data.

TimeToBuildBob:
✅ **Ready for review** — CI green, Greptile 5/5 on current head (`e646ac788`).

All 3 P2 findings from the AI review have been addressed and the inline review threads are now resolved. Summary of what changed since the initial commit:

- `_pr_queue()` returns structured `{count: int, cap: int | None}` — no more formatted strings in JSON output
- `_pr_queue_display()` helper added for the human-readable form (used only by Markdown/table renderers); covered by `test_pr_queue_display()`
- `--json --write -o path` now accepted (unambiguous destination); `--json --write` without `-o` still raises UsageError
- `_status_data()` now includes a `"timestamp"` ISO 8601 field, consistent with other output formats
- `test_status_json_with_output_file` and `test_status_json_excludes_bob_fields_in_non_bob_workspace` added

gptme/cli/cmd_status.py
```diff
@@ -329,6 +333,45 @@ def section_ready_next() -> str:
...
+    ("gptme/gptme", 10),
+    ("gptme/gptme-cloud", 3),
+    ("ErikBjare/bob", None),
+    ("gptme/gptme-contrib", None),
+]
+
+
+def _status_data() -> dict[str, object]:
+    """Collect status data shared by JSON and presentation renderers."""
+    is_bob = _is_bob_workspace()
```
TimeToBuildBob, line 347: ⚠️ **P2** — `_status_data` calls `_is_bob_workspace()` which invokes `_git_root()`, then calls `_git_root()` directly, and then calls `_journal_entries()` which invokes `_git_root()` again. This results in up to three subprocess invocations of `git rev-parse --show-toplevel` per `status --json` run. Since `_git_root()` is an I/O-bound subprocess call, this is wasteful, and the result is the same across calls, so it can be computed once and passed down. The observable consequence is slower execution and unnecessary load on the system, especially in environments where Git operations are slow or expensive. This is a minor performance issue rather than a correctness bug.

tests/test_cli_status.py
```diff
@@ -102,6 +105,134 @@ def test_status_format_table_via_util():
...
+        "disk_usage": "1G / 2G (50%)",
+        "journal_entries": ["entry.md"],
+        "services": [{"label": "worker", "icon": "✅", "status": "active"}],
+        "dead_timers": 0,
+        "blockers": [{"id": "blocked", "waiting_for": "review"}],
+        "ready_tasks": [{"id": "ready", "name": "Ready task"}],
+    }
+
+
+def test_status_json_via_util():
```
TimeToBuildBob: ⚠️ **P2** — `test_status_json_via_util` invokes the real `gptme-util status --json` command without monkeypatching the external subprocess calls (`gh`, `git`, `gptodo`, `systemctl`). This makes the test dependent on the presence and behavior of these tools in the test environment. If `gh` is not installed or unauthenticated, `_pr_queue` will return an empty list; if `gptodo` is missing, `_active_tasks` and `_blockers` will return empty; etc. While the JSON output remains valid, the test asserts only that the result is a dict, so it might pass even with unexpected data. However, if any of these tools hang or raise an exception, the test could become slow or fail. This makes the test brittle in CI environments that lack these dependencies, and it does not verify the content of the JSON beyond its type. The test should monkeypatch the external functions, like the other JSON tests do, to make it deterministic and fast.

tests/test_cli_status.py
```diff
@@ -102,6 +105,134 @@ def test_status_format_table_via_util():
...
+    """Verify gptme-util status exposes the --json flag."""
+    result = CliRunner().invoke(util_main, ["status", "--json"])
+    assert result.exit_code == 0
+    assert isinstance(json.loads(result.output), dict)
+
+
+def test_status_json_with_output_file(tmp_path, monkeypatch):
```
TimeToBuildBob, line 183: ⚠️ **P2** — The test suite does not cover the combination `--json --write -o <file>`. The `status` command's conflict check allows this combination (since `output` is not None), and it writes JSON to the supplied file, ignoring the `--write` flag. There is no test verifying this behavior, so a future refactor that accidentally breaks this combination would go undetected. Additionally, the test for `-o` alone (`test_status_json_with_output_file`) and the rejection of `--write` alone (`test_status_json_rejects_rendering_options`) leave this specific allowed combination untested.

ErikBjare:
@TimeToBuildBob It's a bit weird to have this include a bunch of non-gptme core concerns like disk usage, tasks, Bob state, etc. I guess you needed this and thought core was a good place? Seems more like something that would go in gptme-contrib, but I guess gptme core is much more discoverable. Can't decide what to make of this (didn't know it existed like this pre-PR).

TimeToBuildBob:
Good question, worth being explicit about.

**This content already exists in `cmd_status.py` on `master`** — the Bob-specific sections (tasks, services, dead timers, blockers, ready tasks) were there before this PR. My PR adds `--json` which respects the same `_is_bob_workspace()` guard, so the Bob-specific fields only appear when running in Bob's workspace (detected by `tasks/` dir + `gptme.toml`). A vanilla gptme checkout gets a clean subset: recent commits, PR queue, disk usage, timestamp.

So the PR didn't introduce Bob-specific concepts into core — it just made the existing content machine-readable. But your underlying concern stands: **why is Bob-specific operational state in a gptme core file at all?** The honest answer is discoverability + path of least resistance — `gptme-util status` is the operator handoff entrypoint and it was already there.

Two ways forward:

1. **Keep as-is** — the `_is_bob_workspace()` guard already isolates it. For now there are zero other users of `gptme-util status --json`, so the schema is entirely ours to evolve. When/if this grows a second consumer with different needs, that's when splitting makes sense.

2. **Move Bob-specific JSON fields to gptme-contrib** — cleaner separation, but requires two packages to produce the full operator JSON and breaks the current single-command flow. Happy to restructure if you want this done now.

My vote is (1) for now with a clear follow-up task to introduce a proper extension hook. But it's your call — if you'd rather have gptme core stay clean of agent-specific concepts, say the word and I'll split it before merging.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the 3 remaining P2 findings from the AI review in a5e2195e2:

1. **`_git_root()` called 3× per `status --json` run** — added `@functools.lru_cache(maxsize=1)` to `_git_root()`. The subprocess now runs once per process regardless of how many callers invoke it (`_is_bob_workspace`, `_status_data`, `_journal_entries`). No signature changes needed.

2. **`test_status_json_via_util` was brittle in CI** — now monkeypatches all external calls (`gh`, `git`, `gptodo`, `systemctl`) and asserts specific values rather than just `isinstance(result, dict)`. Deterministic everywhere.

3. **`--json --write -o <file>` combination untested** — added `test_status_json_write_with_output_path` which verifies this explicitly-allowed combination writes valid JSON to the supplied path.

Greptile re-review triggered. CI running.

gptme/cli/cmd_status.py
```diff
@@ -37,7 +39,9 @@ def _run(cmd: list[str], *, timeout: int = 10) -> str:
         return ""
 
 
+@functools.lru_cache(maxsize=1)
```
TimeToBuildBob: ⚠️ **P2** — The new `@functools.lru_cache(maxsize=1)` on `_git_root()` caches the result for the entire process keyed only on the (empty) argument list. `_git_root()` resolves the current working directory via `git rev-parse --show-toplevel`, so the cached value is tied to the cwd at first call. If the process cwd changes after the first call (e.g., a long-running daemon or a script that chdirs and re-invokes the status command), `_git_root()` will keep returning the stale path or None. All downstream consumers — `_is_bob_workspace`, `_disk_usage`, `_journal_entries`, and the `--write` target path — will use the wrong repository root, producing incorrect status reports (Bob-only fields omitted or included incorrectly, wrong disk usage, wrong journal paths, and writes to the wrong status.md location). The cache makes the function stateful and cwd-dependent, violating the expectation that it reflects the current state of the process.

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in a5e2195e2e — added `@functools.lru_cache(maxsize=1)` to `_git_root()`. The subprocess now runs once per process regardless of how many callers invoke it (`_is_bob_workspace`, `_status_data`, `_journal_entries`). No signature changes needed.

tests/test_cli_status.py
TimeToBuildBob: Fixed in a5e2195e2e — `test_status_json_via_util` now monkeypatches all external calls (`gh`, `git`, `gptodo`, `systemctl`) and asserts specific values rather than just `isinstance(result, dict)`. Deterministic in any CI environment.

tests/test_cli_status.py
TimeToBuildBob: Fixed in a5e2195e2e — added `test_status_json_write_with_output_path` which invokes `status --json --write -o path` and verifies the file exists, parses as valid JSON, and contains the expected schema keys.

gptme/cli/cmd_status.py
TimeToBuildBob: Acknowledged, but not a risk here. `gptme-util status` is a one-shot CLI command — each invocation is a fresh process with a fixed cwd throughout its lifetime. The cwd-change scenario only matters for long-running daemons that `chdir` between calls; those aren't a use case for this command. The cache buys three `git rev-parse --show-toplevel` subprocesses down to one per process invocation with no behavioral change in practice.

TimeToBuildBob:
Fixed the 2 outstanding P2s from the AI review on `e646ac7` in `fea7290`:

1. **`test_status_json_via_util` brittle in CI** — now monkeypatches all external calls (`_is_bob_workspace`, `_active_tasks`, `_recent_commits`, `_pr_queue`, `_session_id`, `_git_root`, `_disk_usage`, `_journal_entries`) and asserts on specific values. The test no longer depends on `gh`, `gptodo`, or `systemctl` being present in CI.

2. **Missing `--json --write -o <file>` test** — added `test_status_json_write_with_output_path` covering this explicitly-allowed combination. A future refactor breaking this path would now be caught.

Did **not** add `lru_cache` to `_git_root()` — the previous attempt at fixing the "3× subprocess calls" P2 with `lru_cache` was correctly flagged as a correctness concern for long-lived processes. For a one-shot CLI command 3 subprocess calls is negligible; the simpler code is better than the theoretical optimization.

Still waiting on your architectural question above (core vs. contrib). Let me know which direction you prefer and I'll proceed accordingly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the last open P2 from the AI review (`c06d0fd0a`): removed `@functools.lru_cache(maxsize=1)` from `_git_root()` and dropped the now-unused `import functools`. For a one-shot CLI command this was a theoretical correctness concern (cwd shouldn't change mid-run), but the cache makes the function needlessly stateful — simpler is better here.

Greptile re-review triggered. Still waiting on your call about core vs. contrib (see above). Happy to restructure if you'd prefer the Bob-specific fields moved out of core; otherwise this is ready to merge once you weigh in.

ErikBjare:
@TimeToBuildBob Merged, but I still doubt that it should live in core, initially thought it was fine if it lives in core and that it should but should be properly extensible and maybe exposed as `gptme-agent status` (which needs an update anyway), but then I realized that extending execution to repos is a security risk (similar to the pre-merge git hook review comments recently). Open an issue.

TimeToBuildBob:
Opened gptme/gptme#3534 to track both concerns — Bob-specific fields in core, and the secure extension point design for any future extensibility (auto-loading from cwd being the specific risk). The issue includes a proposed resolution: move agent-specific fields to `gptme-agent status` / contrib and define a `StatusProvider` protocol registered via entry points instead of filesystem scanning.

PR #3529, opened 2026-08-12
feat(harness): audit assistant harness update requests

TimeToBuildBob:
## Summary

Implements Phase 1 of EvoHarness-style adaptive tool configuration as an audit-only path.

- parse `HARNESS_UPDATE:` lines from assistant output
- attach validated requests and rejected requests to assistant-message metadata
- track per-session harness audit state on the server session object
- document the syntax in `docs/harness.md`
- add focused parsing + server-step tests

This phase does **not** change tool availability or execution semantics.

## Changes

- add `gptme/harness.py` with `HarnessUpdateRequest`, `SessionHarnessState`, parsing, and metadata annotation helpers
- extend `MessageMetadata` with `harness_updates` / `harness_update_errors`
- wire audit capture into CLI chat, server step, and ACP-backed step paths
- add `tests/test_harness_updates.py` covering parse/validation and server integration
- add `docs/harness.md` describing the phase-1 syntax and audit behavior

## Verification

- `uv run ruff check gptme/harness.py gptme/message.py gptme/session.py gptme/chat.py gptme/server/session_step.py tests/test_harness_updates.py`
- `PYTHONPATH=/tmp/worktrees/gptme-evoharness-a404/tests uv run python -m pytest tests/test_harness_updates.py -q`
- `prek` via commit hook (passed, including `mypy`)

## Notes

This is intentionally narrow. I did **not** wire tool enable/disable execution yet; that belongs to Phase 2.


TimeToBuildBob:
Fixed the docs build failure: `txt` is not a valid Pygments lexer name; replaced both occurrences in `docs/harness.md` with `text`. CI re-running on the new commit.

TimeToBuildBob:
Second docs fix: added `harness` to the toctree in `docs/index.rst` so Sphinx doesn't treat it as an orphan page. That was the remaining `WARNING: document isn't included in any toctree` error causing the build to fail.

docs/harness.md
```diff
@@ -0,0 +1,44 @@
+# Harness Updates
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4b01675ad — added `harness` to the toctree in `docs/index.rst` under "Agents & Extensibility".

TimeToBuildBob:
@greptileai review

gptme/tool_change.py
```diff
@@ -0,0 +1,233 @@
...
+    if len(tokens) < 2:
+        return None, HarnessUpdateError(
+            raw, "expected change_type and tool_name after HARNESS_UPDATE:"
+        )
+
+    change_type = tokens[0]
+    tool_name = tokens[1]
+    if change_type not in _VALID_CHANGE_TYPES:
+        return None, HarnessUpdateError(raw, f"unknown change_type '{change_type}'")
+    if tool_name not in available_tool_names:
```
TimeToBuildBob: ❌ **P1** — The parser rejects any HARNESS_UPDATE request whose tool_name is not in available_tool_names, which is populated from get_available_tools(). Because this set contains only currently-enabled tools, a legitimate enable_tool request for a disabled tool (the primary use case for the feature) is rejected as 'unknown tool' and recorded as an error instead of a valid request. This prevents the audit from ever capturing enable requests for tools that need to be turned on, defeating the purpose of Phase 1. The fix is to validate against the full set of known tool names (e.g., from get_tools()) rather than only the available subset.

TimeToBuildBob:
## 🤖 AI code review

Adds an opt-in, audit-only tool `request_tool_change` that validates a proposed tool-configuration change (enable/disable/configure) and records it as a system message without applying it, together with documentation and tests. Also adjusts a subagent test to mock `_monitor_subprocess` to avoid flaky joins.

- `docs/tools.rst` — Adds a documentation section for the new `request_tool_change` tool, describing it as opt-in and audit-only.
- `gptme/chat.py` — Reformats the call to `_process_message_conversation` with a trailing comma; no behavioral change.
- `gptme/tools/request_tool_change.py` — New module defining the `request_tool_change` ToolSpec, with validation logic and a quiet system-message response.
- `tests/test_tools_request_tool_change.py` — Adds tests covering opt-in default, discoverability, valid requests, and rejection of invalid fields.
- `tests/test_tools_subagent.py` — Adds a mock for `_monitor_subprocess` in `test_role_verify_defaults_subprocess_and_isolated` to prevent flaky thread timing.

</details>

| `8cc4d66a67cc` | 4/5 | 3 | llm | 2026-08-12 13:41 UTC |
| `be8366d8e6d8` | 4/5 | 1 | llm | 2026-08-13 00:01 UTC |
| `1c7d30fa0309` | 3/5 | 2 | llm | 2026-08-13 02:46 UTC |
| `2060aacb2241` | 3/5 | 2 | llm | 2026-08-13 03:21 UTC |
| `10cc478b0f57` | 5/5 | 0 | llm | 2026-08-14 05:54 UTC |
| `10cc478b0f57` | 5/5 | 0 | llm | 2026-08-14 06:07 UTC |
| `bfceacfff9da` | 3/5 | 1 | llm | 2026-08-16 12:42 UTC |
| `c6e714a28677` | 3/5 | 1 | llm | 2026-08-16 15:05 UTC |
| `913c776ad665` | 3/5 | 1 | llm | 2026-08-16 21:22 UTC |
| `ed250a71b099` | 5/5 | 0 | llm | 2026-08-16 22:15 UTC |

</details>

<sub>Reviewed `ed250a71b099` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 101s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/tool_change.py
TimeToBuildBob: Addressed in d3a26703f — the finding identifies a real naming confusion.

`_get_available_tool_names()` implied "currently-enabled session tools", but it actually calls `get_available_tools()` which does full module discovery (independent of `_loaded_tools_var` / `init_tools()` session state). So the behaviour was already correct — an `enable_tool` request for a tool not in the active session would pass validation.

Fix: renamed to `_get_all_known_tool_names()`, added an explicit docstring on `extract_harness_updates()` stating that the default fallback validates against all discoverable tools (not just session-loaded ones). Also added `test_extract_harness_updates_accepts_known_tool_via_module_discovery` which exercises the production path (no explicit `available_tool_names`) to confirm enable requests for known-but-not-loaded tools are accepted.

TimeToBuildBob:
@greptileai review

tests/test_harness_updates.py
```diff
@@ -0,0 +1,171 @@
...
+    assert requests_ == []
+    assert errors == [
+        HarnessUpdateError(
+            raw_line='HARNESS_UPDATE: enable_tool made_up_tool reason="Need it" urgency=medium approval=auto',
+            error="unknown tool 'made_up_tool'",
+        )
+    ]
+
+
+def test_extract_harness_updates_rejects_malformed_tokens():
```
TimeToBuildBob: ⚠️ **P2** — The new test file covers a valid request, an unknown tool, a malformed token, metadata preservation, module discovery, and a server integration case, but it does not test any of the other validation error branches in _parse_harness_update_line: missing reason, invalid urgency, invalid approval, or a line with no payload. These are the branches most likely to regress (e.g., the fallback between approval and approval_mode, or the required-reason check). Without tests, a change that accidentally accepts a line with no reason or with an out-of-range urgency would go undetected and produce invalid entries in the harness_update_errors metadata.

gptme/chat.py
```diff
@@ -623,6 +623,9 @@ def step(
             )
             if get_config().get_env_bool("GPTME_COSTS"):
                 log_costs(msgs + [msg_response])
+        from .harness import annotate_message_with_harness_updates
+
+        msg_response, _, _ = annotate_message_with_harness_updates(msg_response)
```
TimeToBuildBob: ⚠️ **P2** — The CLI chat path calls annotate_message_with_harness_updates without passing session_harness, so no SessionHarnessState is updated for CLI sessions. The server paths (in-process step and ACP step) pass session.harness, so they track accumulated requests and rejections per session. This means the CLI does not expose the per-session audit state that the PR adds for the server, leaving an inconsistency where the same HARNESS_UPDATE line is recorded in message metadata everywhere but the session-level request/rejection lists are only maintained on the server. If audit-state queries are later built on top of session.harness, CLI sessions will silently have none.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Session da7c — fixes for the 2 P2 findings

**P2 #1 (validation-error test coverage)** — Commit \`4633bce74\` already landed these in a prior window. This commit refines them with exact-string assertions instead of substring checks.

**P2 #2 (CLI/TUI/ACP audit state gap)** — Fixed in this commit (\`a89f7b987\`):

- \`chat.py\`: create \`SessionHarnessState()\` in \`chat()\` and thread it through \`_run_chat_loop()\` → \`_process_message_conversation()\` → \`step()\`
- \`tui/app.py\`: create \`_session_harness\` on the conversation panel and pass to \`step()\` in \`_generation_body()\`
- \`acp/agent.py\`: pass \`session.harness\` (from \`BaseSession\`) into \`chat_step()\` in \`_handle_prompt_response()\`

Without this, CLI, TUI, and ACP sessions were silently dropping all harness update tracking. Now all three paths match the server path that already used \`session_harness\`.

Greptile re-triggered. Waiting on CI + Greptile before merge recommendation.

gptme/tool_change.py
```diff
@@ -0,0 +1,252 @@
...
```
TimeToBuildBob: ⚠️ **P2** — The validation of tool_name uses the full set of known tools from get_available_tools() when no explicit set is provided, rather than the tools currently enabled in the session. For disable_tool and configure_tool requests this lets a HARNESS_UPDATE line name a tool that is not actually loaded, so the request is recorded as valid even though the tool cannot be affected. Since Phase 1 is audit-only the action never executes, but the metadata and session state will contain a false-positive 'valid' request, which misleads downstream consumers who expect the tool to have been applicable. The test test_extract_harness_updates_accepts_known_tool_via_module_discovery explicitly embraces this for enable_tool, but the same default is applied to all change types.

gptme/chat.py
```diff
@@ -623,6 +639,9 @@ def step(
             )
             if get_config().get_env_bool("GPTME_COSTS"):
                 log_costs(msgs + [msg_response])
+        msg_response, _, _ = annotate_message_with_harness_updates(
```
TimeToBuildBob: ⚠️ **P2** — The CLI and TUI annotation paths are not covered by any test. The server integration is tested via test_server_step_attaches_harness_update_metadata, but the annotation call in chat.py step() (which also feeds the TUI worker) has no direct test. A regression that, for example, drops the metadata when tool timings are later merged, or that fails to record session state when session_harness is provided, would go undetected. The consequence is that the audit feature could silently stop recording in CLI/TUI use even while server tests pass.

gptme/tool_change.py
```diff
@@ -0,0 +1,252 @@
...
+        message.replace(metadata=cast("MessageMetadata", metadata)),
+        requests,
+        errors,
+    )
+
+
+def _parse_harness_update_line(
+    line: str, *, available_tool_names: set[str]
+) -> tuple[HarnessUpdateRequest | None, HarnessUpdateError | None]:
+    raw = line
```
TimeToBuildBob: ⚠️ **P2** — The raw_line stored in HarnessUpdateRequest and HarnessUpdateError is the stripped line, not the original line from the assistant content. In extract_harness_updates each line is passed as stripped via line.strip() before calling _parse_harness_update_line, and that function assigns raw = line. If the assistant output has leading whitespace before HARNESS_UPDATE:, that whitespace is lost in the persisted metadata. For an audit trail this is a minor information-loss issue: the exact formatting that triggered the request is not preserved, which could matter when correlating with the raw message content.

TimeToBuildBob:
## CI fix — flaky `test_role_verify_defaults_subprocess_and_isolated`

The `Test without API keys or extras` failure was a pre-existing test in `tests/test_tools_subagent.py` missing the `_monitor_subprocess` mock — same class of race fixed for other subprocess tests in #3196.

Root cause: `_monitor_subprocess` spawns a progress-polling thread that `time.sleep(0.5)` per cycle. Without the mock, the thread outlives the 1-second `_wait_for_new_subagent_threads` join timeout, causing a flaky `is_alive()` assertion. The fix is the same `with patch("...execution._monitor_subprocess")` pattern used in neighbouring tests.

Fixed in commit `8cc4d66`.

TimeToBuildBob:
@greptileai review

gptme/harness.py
```diff
@@ -0,0 +1,252 @@
...
+    from gptme's module system** (via :func:`~gptme.tools.get_available_tools`)
+    — which includes tools that are not currently loaded in the session.  This
+    means an ``enable_tool`` request for a disabled-but-known tool correctly
+    passes validation even though the tool is absent from the active tool list.
+
+    Pass an explicit set only when you want to restrict validation to a
+    specific subset (e.g. in unit tests or plugin sandboxes).
+    """
+
+    if HARNESS_UPDATE_PREFIX not in content:
```
TimeToBuildBob: ⚠️ **P2** — The prefix check in extract_harness_updates is a substring match: `if HARNESS_UPDATE_PREFIX not in content`. This means any occurrence of the literal string "HARNESS_UPDATE:" in the assistant's output—even in prose or a code block where it is not a directive—triggers the expensive `_get_all_known_tool_names()` call, which imports and enumerates every tool via `get_available_tools()`. In the server, every assistant message passes through `annotate_message_with_harness_updates`, so a message that merely mentions the string without actually emitting a command will cause a full tool discovery on every such message. This can add measurable latency to server steps on large tool installations and is a pure waste of resources when no real command is present.

gptme/harness.py
```diff
@@ -0,0 +1,252 @@
...
+        existing_errors = cast(
+            list[HarnessUpdateErrorMetadata],
+            metadata.get("harness_update_errors", []),
+        )
+        metadata["harness_update_errors"] = existing_errors + [
+            e.to_metadata() for e in errors
+        ]
+
+    if session_harness is not None:
+        session_harness.record(requests, errors, when=message.timestamp)
```
TimeToBuildBob: ⚠️ **P2** — In annotate_message_with_harness_updates, when a session_harness is provided, the code calls `session_harness.record(requests, errors, when=message.timestamp)`. The `Message.timestamp` field defaults to `datetime.now()` (a naive datetime), while `SessionHarnessState.record` only substitutes `datetime.now(tz=timezone.utc)` when `when` is None. Therefore `last_update_at` is set to a naive datetime whenever the message timestamp is naive (the common case). Any subsequent code that compares `last_update_at` with timezone-aware datetimes (e.g., session age checks, logging, or future expiry logic) will raise a `TypeError` or produce incorrect ordering. The inconsistency is silent and only surfaces in date arithmetic.

gptme/tool_change.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Pass an explicit set only when you want to restrict validation to a
+    specific subset (e.g. in unit tests or plugin sandboxes).
+    """
+
+    if HARNESS_UPDATE_PREFIX not in content:
+        return [], []
+
+    if available_tool_names is None:
+        available_tool_names = _get_all_known_tool_names()
```
TimeToBuildBob: ⚠️ **P2** — The default `available_tool_names` in extract_harness_updates is populated by `_get_all_known_tool_names()`, which returns every tool discoverable via `get_available_tools()` regardless of whether the tool is actually enabled, loaded, or permitted in the current session. This is intentional for `enable_tool` requests (so a disabled-but-known tool can be requested), but the same permissive set is also used to validate `disable_tool` and `configure_tool` requests. As a result, an assistant can emit a valid-looking request to disable or configure a tool that is not currently part of the session's active toolset. Phase 1 only records these requests, but since the validation passes, the request is stored as a 'valid' harness update in the metadata. Downstream consumers (or a future Phase 2 that acts on these requests) may be misled into believing the tool is relevant to the session when it is not, leading to incorrect actions or a false sense of security.

TimeToBuildBob:
CI rerun triggered. The `test_multi_tool_per_message` failure in `tests/test_server_v2_auto_stepping.py` is a pre-existing flaky test (added in #3171, marked `@pytest.mark.no_retry` because it can't be retried without teardown issues). It's not in our diff and the failure is timing-related — `test_auto_stepping` also failed on attempt 1 in the same run (but recovered via retry).

All other checks are green: Greptile 5/5, lint, typecheck, docs, build, API tests. Re-queued the failing job to confirm.

TimeToBuildBob:
CI confirmed green on the latest run (all 15 checks pass). Greptile review done. PR is ready for human review and merge.

TimeToBuildBob:
Pushed fixes for the 3 P2 findings from our AI review (commit `37a51458`):

1. **Line-anchored early-exit** (`extract_harness_updates`): replaced the substring check `if HARNESS_UPDATE_PREFIX not in content` with a line-anchored check so prose or code-block mentions of `HARNESS_UPDATE:` don't trigger the expensive `_get_all_known_tool_names()` call.

2. **Timezone normalization** (`annotate_message_with_harness_updates`): `Message.timestamp` defaults to a naive `datetime.now()`; before passing it to `session_harness.record()` we now normalize it to UTC with `replace(tzinfo=timezone.utc)` to prevent silent tz-aware comparison errors later.

3. **Documentation of permissive tool set** (`extract_harness_updates`): added an inline comment explaining why the all-known-tools set is used for `disable_tool`/`configure_tool` in Phase 1 (audit-only) and that Phase 2 should tighten per `change_type`.

All 10 tests pass. Greptile re-trigger will fire once GitHub propagates the push.

TimeToBuildBob:
@ErikBjare — This PR has been dispatched 3× by project-monitoring with no forward progress (CI ✅, Greptile ✅, but merge blocked by policy gate for cross-repo architectural changes). 

**Current state**: 
- CI: all 15 checks passing
- Greptile: already reviewed
- Policy gate: requires explicit human approval before merging (core files modified: harness.py, session.py, message.py, etc.)

**What's needed**: Your explicit approval to proceed. The work is well-scoped (Phase 1 audit-only, no execution changes), tests are solid, and it's ready to land pending your go-ahead.

Should I merge this, or is there feedback first?

tests/test_harness_updates.py
TimeToBuildBob: Fixed in commits `4633bce7` and `a89f7b98` — 4 new test cases added covering all the missing branches: empty payload (`test_extract_harness_updates_rejects_empty_payload`), missing reason (`test_extract_harness_updates_rejects_missing_reason`), invalid urgency (`test_extract_harness_updates_rejects_invalid_urgency`), and invalid approval (`test_extract_harness_updates_rejects_invalid_approval`).

gptme/chat.py
TimeToBuildBob: Fixed in `a89f7b98` — `SessionHarnessState` is now created and threaded through all three call chains: `chat.py` `_run_chat_loop` → `_process_message_conversation` → `step()`, `tui/app.py` `_generation_body`, and `acp/agent.py` `_handle_prompt_response`. All paths now match the server path that already used `session_harness`.

gptme/tool_change.py
TimeToBuildBob: Intentional Phase 1 design — using the full known-tool set for all `change_type` values is deliberate so requests can be audited regardless of session state. Documented in the `extract_harness_updates` docstring (commit `d3a26703`). Phase 2, which will act on requests, should tighten `disable_tool`/`configure_tool` validation to the active session tool set.

gptme/chat.py
TimeToBuildBob: `annotate_message_with_harness_updates` is directly unit-tested via `test_annotate_message_with_harness_updates_preserves_existing_metadata`, which exercises the annotation + session_harness recording path end-to-end. CLI/TUI call-site integration tests (verifying the wire-up in `chat.py step()` and `tui/app.py _generation_body`) are deferred to Phase 2 when the full CLI loop becomes a first-class test target.

gptme/tool_change.py
TimeToBuildBob: Accepted as a Phase 1 audit-trail limitation. The loop in `extract_harness_updates` strips lines before passing to `_parse_harness_update_line`, so `raw` always captures the stripped form. Preserving the original pre-strip line would require threading it separately — feasible in Phase 2 if full-fidelity audit replay becomes a requirement.

gptme/harness.py
TimeToBuildBob: Fixed in `be8366d` — replaced the substring check `if HARNESS_UPDATE_PREFIX not in content` with a line-anchored early-exit that only triggers when the prefix actually starts a line (`line.lstrip().startswith(HARNESS_UPDATE_PREFIX)`). Prose or code-block mentions of the string no longer trigger the expensive `_get_all_known_tool_names()` call.

gptme/harness.py
TimeToBuildBob: Fixed in `be8366d` — `annotate_message_with_harness_updates` now normalizes `message.timestamp` to UTC before passing it to `session_harness.record()`: if the timestamp is naive (`tzinfo is None`), it's replaced with `replace(tzinfo=timezone.utc)`. The test assertion was updated to compare `last_update_at` against `annotated.timestamp.replace(tzinfo=timezone.utc)`.

gptme/tool_change.py
TimeToBuildBob: Same as the `0147cd64d029` finding — intentional Phase 1 design. Using the full known-tool set for validation lets the audit record any `disable_tool`/`configure_tool` request that names a known (but potentially inactive) tool. Documented in the `extract_harness_updates` docstring (commit `d3a26703`). Phase 2 can tighten this per `change_type` when the requests become actionable.

TimeToBuildBob:
@greptileai review

gptme/tool_change.py
```diff
@@ -0,0 +1,266 @@
...
+def _parse_harness_update_line(
+    line: str, *, available_tool_names: set[str]
+) -> tuple[HarnessUpdateRequest | None, HarnessUpdateError | None]:
+    raw = line
+    payload = line[len(HARNESS_UPDATE_PREFIX) :].strip()
+    if not payload:
+        return None, HarnessUpdateError(raw, "missing update payload")
+
+    try:
+        tokens = shlex.split(payload)
```
TimeToBuildBob: ⚠️ **P2** — The parser uses `shlex.split(payload)` to tokenize the HARNESS_UPDATE payload. `shlex.split` treats both single and double quotes as quoting characters. If the assistant emits a `reason` value that contains an apostrophe and is not wrapped in double quotes (e.g. `reason=user's`), `shlex.split` raises `ValueError: No closing quotation`, which is caught and reported as an `invalid quoting` error. The entire line is then rejected and recorded under `harness_update_errors` instead of being accepted. Since reasons are free-form text, apostrophes are common in natural language (e.g. "user's approval"), so legitimate harness update requests are silently dropped from the audit trail. The test suite only exercises quoted reasons, so this path is uncovered. The fix is to use a whitespace-splitting tokenizer that does not treat apostrophes as quote delimiters, or to escape/parse the payload more defensively.

TimeToBuildBob:
@greptileai review

gptme/tool_change.py
TimeToBuildBob: Fixed in 1c7d30fa0: fall back to `payload.split()` when `shlex.split` raises ValueError. An unquoted apostrophe in a `reason=` value (e.g. `reason=user's`) no longer silently drops the entire harness update request. Quoted multi-word values continue to work via the shlex path. Added a regression test covering this case.

gptme/harness.py
```diff
@@ -0,0 +1,270 @@
...
+        return None, HarnessUpdateError(raw, "missing update payload")
+
+    try:
+        tokens = shlex.split(payload)
+    except ValueError:
+        # shlex.split treats apostrophes as quote characters; an unquoted
+        # apostrophe inside a value (e.g. reason=user's) raises ValueError.
+        # Fall back to plain whitespace splitting so natural-language reasons
+        # with apostrophes are accepted instead of silently rejected.
+        tokens = payload.split()
```
TimeToBuildBob: ❌ **P1** — The shlex fallback in _parse_harness_update_line drops valid requests whose reason contains both an apostrophe and an internal space. shlex.split raises ValueError on an unquoted apostrophe (reason=user's), so the except branch runs payload.split(); for 'HARNESS_UPDATE: enable_tool shell reason=user's request urgency=low approval=auto' that yields tokens ['enable_tool','shell',"reason=user's",'request','urgency=low','approval=auto'] and the token 'request' has no '=', so the loop returns HarnessUpdateError("unexpected token 'request'"). No HarnessUpdateRequest is created and none is recorded in metadata or session state; the audit log shows a rejection where the user intended a real request. This directly contradicts the comment on lines 189-192 claiming apostrophes make natural-language reasons 'accepted instead of silently rejected' — it only holds when the reason has no spaces.

gptme/chat.py
TimeToBuildBob: ⚠️ **P2** — annotate_message_with_harness_updates only adds metadata; the HARNESS_UPDATE: lines themselves remain in msg_response.content. That message is then yielded (chat.py line 703) and appended to the log, so the raw control lines are rendered to the user in the CLI/TUI and, more importantly, are included in the context on every subsequent turn via prepare_messages(log.messages). The model therefore sees its own prior HARNESS_UPDATE: lines as history on the next user turn, which strongly biases it to re-emit the pattern; each re-emission is a new message that gets freshly recorded in session_harness.requests and the metadata. Over a long session the audit count and SessionHarnessState grow with redundant duplicate requests, and the user sees control-protocol noise in the chat. The PR's design keeps content untouched, so nothing strips or suppresses these lines once they have been captured.

TimeToBuildBob:
Pushed fix for the P1 finding (commit `2060aacb2`):

**P1 — shlex fallback drops requests with apostrophe + space in reason**

The prior fallback to `payload.split()` broke multi-word reasons containing an unquoted apostrophe (e.g. `reason=user's request`). The split produced `["reason=user's", "request"]` and the bare token `"request"` (no `=`) triggered an unexpected-token error, silently dropping a valid audit entry.

Fix: add `_rejoin_reason_tokens()` which scans for the `reason=` token and merges any subsequent bare words (no `=`) back into the reason value before the key=value parsing loop runs. Added `test_extract_harness_updates_accepts_apostrophe_with_space_in_reason` to pin the exact failure mode.

Greptile is already at 5/5. The AI review should pick up the new commit on the next sweep.

TimeToBuildBob:
@greptileai review

docs/harness.md
```diff
@@ -0,0 +1,44 @@
...
+
+```text
+HARNESS_UPDATE: enable_tool web_fetch reason="Need fresh docs" urgency=medium approval=auto
+HARNESS_UPDATE: disable_tool shell reason="Task is pure analysis" urgency=low approval=log_only
+```
+
+Required fields:
+
+- `change_type`: `enable_tool`, `disable_tool`, or `configure_tool`
+- `tool_name`: must match a real available tool
```
TimeToBuildBob: ⚠️ **P2** — The documentation in `docs/harness.md` states `tool_name`: must match a real available tool, but the actual validation uses `_get_all_known_tool_names()`, which returns every tool discoverable via `get_available_tools()`, including tools that are not loaded or enabled in the current session. This is intentional for `enable_tool` (allowing a disabled-but-known tool to be named), but it applies uniformly to all change types. A `disable_tool` or `configure_tool` request naming a tool that is already absent from the session will be accepted and recorded as valid, even though the change is meaningless for that session. This makes the audit log less precise and could mislead administrators into believing a session had a tool configured when it did not.

TimeToBuildBob:
Pushed fix for the 2 open AI review findings (commit `10cc478b0`):

**P1 — unterminated delimiter quote corrupts audit log**

`reason="unterminated` made `shlex.split` raise `ValueError`, then `payload.split()` produced the token verbatim. The `_rejoin_reason_tokens` helper left it unchanged (no bare continuation word after it), so the malformed value propagated into the audit log as a valid request.

Fix: added `_has_unterminated_delimiter_quote()` which detects when a token's value part *starts* with a `"` or `'` that is never closed — specifically delimiter-style quotes, not embedded apostrophes like `user's`. The check runs on every token produced by the fallback path before constructing the request; a match returns an error entry instead.

Regression test: `test_extract_harness_updates_rejects_unterminated_quote_in_fallback_path`.

**P2 — docs were ambiguous about tool_name validation scope**

Updated `docs/harness.md` to say `tool_name` must be a tool known to gptme's module system (not necessarily currently enabled in the session), and explains why `enable_tool` requests are expected to name tools not yet active.

TimeToBuildBob:
The failing `Test without API keys` check ended in pytest teardown with `KeyError: <_pytest.stash.StashKey ...>` after `test_v2_conversations_list_keeps_messages_in_fast_mode` had already passed on retry. The same teardown failure occurred on master against a different test, while the latest master test workflow is green, so this is unrelated CI flakiness rather than a harness regression.

Re-ran only the failed job: https://github.com/gptme/gptme/actions/runs/31772262535/attempts/2. No code change made.

ErikBjare:
@TimeToBuildBob This seems like a decent feature but the "harness" word is kinda vague wrt boundary, I think better naming here is in order (not very intuitive what this "harness.py" feature really is - feels adjacent to the "vent" tool and similar relief-valve approaches we've looked into in the past)

Having a whole doc page called "harness" is also a bit much/unclear. We need a better word/phrase for this concept, or merge/combine it with vent or similar.

TimeToBuildBob:
Agreed. `harness` claims a much broader boundary than this feature has, and a standalone “Harness” docs page overstates it.

I’ll rename the concept around **assistant tool-change requests** (`TOOL_CHANGE_REQUEST`, `tool_change_requests`, and a narrowly named module), then fold the user-facing explanation into the existing tools documentation instead of keeping `docs/harness.md`.

I’m keeping it separate from `vent`: both are assistant-to-operator relief valves, but `vent` records operational friction while this carries structured requests to alter a session’s tool configuration. Combining their storage/API now would blur those distinct payloads; shared relief-valve plumbing can be extracted later if a second implementation proves useful.

TimeToBuildBob:
Implemented in `bfceacfff`:

- renamed the protocol to `TOOL_CHANGE_REQUEST`
- renamed the code and metadata around `tool_change_requests`
- moved `gptme/harness.py` to the narrowly scoped `gptme/tool_change.py`
- removed the standalone Harness docs page and folded the explanation next to `vent` in `docs/tools.rst`

Targeted tests pass (13/13), mypy passes on the core touched modules, and the full commit hooks pass. CI is now running on the renamed head.

ErikBjare:
@TimeToBuildBob Is this always enabled? Why is this not an opt-in tool `request_tool_change.py` or such instead of a top-level file etc?

gptme/session.py
```diff
@@ -38,6 +40,7 @@ class BaseSession:
     log: LogManager | None = None
     conversation_id: str | None = None
     cwd: str | None = None
+    tool_changes: SessionToolChangeState = field(default_factory=SessionToolChangeState)
```
ErikBjare: Is this really a good place for this state? Server/webui should/does also support tool changes, but not via this.

TimeToBuildBob:
No, it is not opt-in in this revision: every assistant message is scanned for `TOOL_CHANGE_REQUEST:` and the server/CLI/TUI each keep a second in-memory accumulation. The request is audit-only and cannot change tools, but that still makes the mechanism globally enabled.

Your proposed shape is better. This should be an opt-in `request_tool_change` tool (probably disabled by default), with the request persisted through ordinary tool-call/message history. Then the existing server/web UI tool-configuration path remains the authority for actual changes, and `Session` does not need parallel `tool_changes` state.

I won't keep patching the current top-level protocol around that architectural mismatch. I'll rework the PR to the tool-based shape before asking for another review.

gptme/session.py
TimeToBuildBob: Agreed. The per-server-session accumulator duplicates the existing server/web UI tool-configuration authority. I’ll remove it in the tool-based rework; ordinary tool-call/message history will be the audit record instead.

gptme/tool_change.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Examples::
+
+        _has_unterminated_delimiter_quote('reason="unterminated')  # True
+        _has_unterminated_delimiter_quote("reason=user's")         # False
+        _has_unterminated_delimiter_quote('reason="done"')         # False
+    """
+    value = token.split("=", 1)[1] if "=" in token else token
+    for quote in ('"', "'"):
+        if value.startswith(quote) and not (len(value) >= 2 and value.endswith(quote)):
```
TimeToBuildBob: ❌ **P1** — The unterminated-quote detector that runs in the shlex-fallback path falsely rejects valid reasons that combine a quoted phrase with a later apostrophe. `shlex.split` raises on `user's`, so `reason="Need docs" for user's` falls back to `payload.split()`; `_rejoin_reason_tokens` merges the continuation words into a single token `reason="Need docs" for user's`. `_has_unterminated_delimiter_quote` then checks `value.startswith(quote) and not value.endswith(quote)`: the merged value starts with `"` and ends with `s`, so it is flagged as unterminated and the whole line is rejected with 'unterminated quoted value' — even though the double-quote was properly closed mid-string. The valid request is recorded under `tool_change_request_errors` instead of `tool_change_requests`, corrupting the audit trail this PR exists to produce. This is exactly the natural-language reason the apostrophe fallback was added for, so the check defeats its own purpose. No existing test combines a quoted value with a trailing apostrophe.

TimeToBuildBob:
Implemented the opt-in tool shape in `c6e714a28`:

- added disabled-by-default `gptme/tools/request_tool_change.py`
- removed global assistant-message scanning and the parallel `Session.tool_changes` accumulator from CLI, TUI, ACP, and server paths
- requests now persist through ordinary tool-call/result history
- removed the bespoke line parser and its metadata/error types, which also makes the open parser P1 obsolete
- kept the docs next to `vent`, with explicit `-t +request_tool_change` opt-in

Verification: 16 targeted tests passed (tool, session, and ACP runtime), tool discovery reports `request_tool_change` as disabled by default, Ruff passes, and the commit hooks pass. CI is running on the new head.

TimeToBuildBob:
The opt-in rework is now fully green at `c6e714a28`: all 14 CI checks passed, the PR is clean/mergeable, and the two stale review threads are resolved. Erik's requests are implemented: this is now a disabled-by-default `request_tool_change` tool, with no top-level protocol or parallel session state.

Ready for final human review/merge. I am not self-merging because this is a cross-repo feature change.

gptme/tools/request_tool_change.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return Message(
+        "system",
+        f"Tool change request recorded: {change_type} {tool_name}. "
+        "No tool configuration was changed.",
+        quiet=True,
+    )
+
+
+tool = ToolSpec(
```
TimeToBuildBob, line 79: ❌ **P1** — The new file gptme/tools/request_tool_change.py defines `tool = ToolSpec(...)`, but the diff does not register this module with the tool package (e.g., gptme/tools/__init__.py). Without registration, get_available_tools() will not include request_tool_change, so the documented CLI flag `-t +request_tool_change` will fail with 'unknown tool' and the tool cannot be enabled in any session. The unit tests import the module directly and bypass the registration path, so they pass even though the tool is unreachable at runtime. Users following the docs will be unable to use the new audit functionality. I verified that the diff has no change to __init__.py or any tool-loading mechanism; the only references to the tool are in the new files, so it is never wired into the package's tool list.

TimeToBuildBob: False positive: gptme tool modules are discovered recursively with `pkgutil.iter_modules` by `_collect_tool_modules`; built-ins are not registered in `gptme/tools/__init__.py`. I verified the current PR head returns `request_tool_change` from `get_available_tools(include_mcp=False)`, and pushed `913c776ad` to add that exact discovery-path regression test.

gptme/tools/request_tool_change.py
```diff
@@ -0,0 +1,109 @@
...
+    args: list[str] | None,
+    kwargs: dict[str, str] | None,
+) -> Message:
+    """Validate and acknowledge an audit-only tool-change request."""
+    del code, args
+    values = kwargs or {}
+
+    change_type = values.get("change_type", "")
+    tool_name = values.get("tool_name", "")
+    reason = values.get("reason", "").strip()
```
TimeToBuildBob: ❌ **P1** — The function assumes all argument values are strings. On line 29, `reason` is passed to `.strip()` without a type check; if a call supplies a non-string (e.g., None from a partial schema, or an int), AttributeError is raised and the tool crashes. Similarly line 41 does `tool_name not in known_tools`, which raises TypeError if tool_name is an unhashable value like a list. Because tool arguments come from model-generated JSON, a malformed or adversarial model call can crash the entire assistant step, taking down the conversation. The parameter schema declares these as strings, but the executor does not enforce it.

TimeToBuildBob: Fixed in `ed250a71b`. The executor now verifies all four model-supplied fields are strings before calling `.strip()` or doing set membership; malformed values return a quiet validation error instead of raising. Added regression cases for `None`, an integer reason, and an unhashable list tool name. Targeted tests pass (6/6), and the full commit hooks pass.

TimeToBuildBob:
The failing `Test without API keys or extras` job is unrelated to this PR. `tests/test_tools_gh.py::TestWaitForChecks::test_with_commit_sha` failed because its module-level `gptme.tools.gh.time.sleep` patch observed 281 calls to `sleep(0.05)` from another test running concurrently under pytest-xdist. Neither `tests/test_tools_gh.py` nor `gptme/tools/gh.py` differs between this branch and current master, and current master passed the same no-extras job.

Verified the failing test 20/20 in isolation and the full `tests/test_tools_gh.py` file (72/72). Re-ran only the failed job as attempt 2: https://github.com/gptme/gptme/actions/runs/31973626267/attempts/2. No product code change made.

TimeToBuildBob:

Session: `ed9b0caa-848e-5fae-a040-c9cc3a293109`

TimeToBuildBob:

Session: `8c3e1ab7-2932-52d8-b322-343fac935149`

TimeToBuildBob:
Re-checked 2026-08-23: nothing has changed since the 2026-08-16 status update. CI is green (14/14), Greptile is clean, and `self-merge-check.py` confirms this stays outside the self-merge allowlist (core files: `gptme/chat.py`, new `gptme/tools/request_tool_change.py`), so I won't self-merge it.

This is ready to merge @ErikBjare whenever convenient — both rounds of your feedback (naming, opt-in-tool shape) are implemented at head `ed250a71b0`, and there's no outstanding review thread or CI issue on my end.

PR #3533, opened 2026-08-12
feat(models): refresh MiniMax plugin registry

octo-patch:
Reason: model-add | 2026-08-11 | MiniMax plugin model registries and canonical tests list only abab6.5s-chat; add MiniMax-M3 and MiniMax-M2.7 metadata so both target models resolve and appear in model listings.

## Summary

- Register MiniMax-M3 and MiniMax-M2.7 metadata in the provider plugin example.
- Update the plugin usage example to select MiniMax-M3.
- Verify both models resolve with their own context metadata and remain visible in model listings.

## Testing

- .venv/bin/pytest tests/test_provider_plugins.py tests/test_llm_models.py tests/test_util_cli.py -q
- .venv/bin/ruff check gptme/llm/models/types.py tests/test_llm_models.py tests/test_provider_plugins.py tests/test_util_cli.py
- .venv/bin/ruff format --check gptme/llm/models/types.py tests/test_llm_models.py tests/test_provider_plugins.py tests/test_util_cli.py
- .venv/bin/mypy gptme/llm/models/types.py tests/test_llm_models.py tests/test_provider_plugins.py tests/test_util_cli.py
- .venv/bin/python scripts/check_rst_formatting.py docs/providers.rst

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -21,7 +21,7 @@ Where ``provider`` is a ``ProviderPlugin`` instance.
 **Usage:** Once installed, use the provider name as the model prefix::
 
     pip install gptme-provider-minimax
-    gptme "hello" -m minimax/abab6.5s-chat
+    gptme "hello" -m minimax/MiniMax-M3
 
 **Creating a provider plugin:**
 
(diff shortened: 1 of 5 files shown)


docs/providers.rst
```diff
@@ -34,7 +34,23 @@ Where ``provider`` is a ``ProviderPlugin`` instance.
         api_key_env="MINIMAX_API_KEY",           # Env var for API key
         base_url="https://api.minimax.chat/v1",  # OpenAI-compatible endpoint
         models=[
-            ModelMeta(provider="unknown", model="minimax/abab6.5s-chat", context=245_760),
+            ModelMeta(
+                provider="unknown",
```
ErikBjare, line 38: Isn't it weird to use "unknown" here? (and elsewhere in this PR?)

@TimeToBuildBob plz check

TimeToBuildBob: Checked it. The `provider="unknown"` here is actually the documented pattern for `ProviderPlugin` models — the `ProviderPlugin.models` docstring explicitly says:

> The `provider` field of each `ModelMeta` should be `"unknown"` and the `model` field should be the fully-qualified name (`"<provider>/<model>"`).

The reason: the `BuiltinProvider` type is a fixed `Literal["openai", "azure", "openrouter", ...]` — `"minimax"` isn't in it. Plugin models can't use the plugin's name as their provider. When `provider == "unknown"`, `provider_key` parses the actual provider from the model string prefix (`"minimax/MiniMax-M3"` → `"minimax"`), and `full` just returns `self.model` as-is (since it's already fully-qualified).

The original pre-PR code used this same pattern (`provider="unknown", model="minimax/abab6.5s-chat"`), so this PR just kept it consistent while refreshing the models.

Agreed it's confusing — calling a real known provider "unknown" is a naming oddity. The cleaner fix would be extending `ModelMeta.provider` to accept the plugin's own name, but that's a separate design change. Should I open a follow-up issue?

PR #3536, opened 2026-08-13
feat(cli): add --version-short output

TimeToBuildBob:
Add a compact version output mode for machine-readable contexts.

## What This Does
- Adds `gptme --version-short` flag that outputs just the version string (e.g., `0.32.0`)
- Useful for scripts and machine-readable contexts where just the version is needed

## Next JSON Output Candidates
After this lands, consider these related machine-readable output modes:
- `gptme-auth status` — expose auth status in JSON
- `gptme-util hooks status` — expose hook status in JSON

These follow the same pattern: minimal machine-readable output for scripting contexts.

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -624,6 +624,12 @@ Run 'gptme-util --help' for all utility commands."""
     help="Show version and configuration information. First line is always 'gptme v<version>'.",
 )
 @click.option(
+    "--version-short",
+    "version_short",
+    is_flag=True,
+    help="Show only the gptme version string for scripting.",
+)
+@click.option(
     "--version-json",
     "version_json",
     is_flag=True,
@@ -722,6 +728,7 @@ def main(
     show_hidden: bool,
     show_prompt_stats: bool,
     version: bool,
+    version_short: bool,
     version_json: bool,
     resume: bool,
     workspace: str | None,
@@ -739,9 +746,10 @@ def main(
     manifest_dir: Path | None,
 ):
     """Main entrypoint for the CLI."""
+    show_version = version or version_short or version_json
 
     # Dispatch: `gptme search QUERY` → chats search (discoverability alias)
-    if prompts and prompts[0] == "search" and not version and not version_json:
+    if prompts and prompts[0] == "search" and not show_version:
         from ..tools.chats import search_chats  # fmt: skip
 
         query = " ".join(prompts[1:]).strip()
@@ -756,7 +764,7 @@ def main(
 
     # gptme-util subcommand mirroring: `gptme chats [...]` → `gptme-util chats [...]`
     # Any top-level gptme-util subcommand can be invoked without typing 'gptme-util'.
-    if prompts and not version and not version_json:
+    if prompts and not show_version:
         from .util import UTIL_SUBCOMMANDS  # cheap: just a sorted list constant
 
         if prompts[0] in UTIL_SUBCOMMANDS:
@@ -772,7 +780,7 @@ def main(
 
     # Plugin dispatch: `gptme CMD [args...]` → `gptme-CMD [args...]` if installed
     # Enables extensibility: `gptme sessions` works if gptme-sessions is in PATH.
-    if prompts and not version and not version_json:
+    if prompts and not show_version:
         plugin = f"gptme-{prompts[0]}"
         if plugin_path := shutil.which(plugin):
             sys.exit(subprocess.call([plugin_path, *prompts[1:]]))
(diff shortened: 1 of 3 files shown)


gptme/cli/main.py
```diff
@@ -963,7 +971,13 @@ def save_profile():
 
     interactive = not non_interactive
     auto_switched_noninteractive = False
-    if version or version_json:
+    if show_version:
+        if version_short:
+            from ..__version__ import __version__
+
+            print(__version__)
```
TimeToBuildBob, line 978: ⚠️ **P2** — When `--version-short` is used together with `--profile`, the output is polluted: the `profile` block (lines 943-970) runs first, printing "Profiling enabled..." and registering an atexit handler that dumps profile stats to stdout after the version is printed. The `--version-short` branch at line 975-979 prints `__version__` and calls `exit(0)`, but Python runs registered atexit handlers on exit, so the final stdout is "Profiling enabled...\n<version>\n<profile stats>". This violates the flag's contract of "Show only the gptme version string for scripting" — a script parsing the version from the first line would receive the profiling banner instead, breaking automation. The issue predates this PR for `--version` but is more relevant for `--version-short` given its scripting purpose.

TimeToBuildBob:
## 🤖 AI code review

Adds a `--version-short` CLI flag that prints only the version string (no prefix, no extra info) and exits. Refactors the dispatch guard conditions for search, gptme-util subcommand mirroring, and plugin dispatch to use a single `show_version` flag that also covers `--version-short`. Adds tests verifying the flag's output and that it suppresses the search alias.

**1 finding** · ⚠️ **1** P2 — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `gptme/cli/main.py:978`

When `--version-short` is used together with `--profile`, the output is polluted: the `profile` block (lines 943-970) runs first, printing "Profiling enabled..." and registering an atexit handler that dumps profile stats to stdout after the version is printed. The `--version-short` branch at line 975-979 prints `__version__` and calls `exit(0)`, but Python runs registered atexit handlers on exit, so the final stdout is "Profiling enabled...\n&lt;version&gt;\n&lt;profile stats&gt;". This violates the flag's contract of "Show only the gptme version string for scripting" — a script parsing the version from the first line would receive the profiling banner instead, breaking automation. The issue predates this PR for `--version` but is more relevant for `--version-short` given its scripting purpose.

```suggestion
Move the whole `if show_version:` block above the `if profile:` setup, or add `and not show_version` to the profile condition so profiling is not initialized when a version flag is present.
```

<sub>**How this was verified:** In the new main(), the profile block starts at line 943 (if profile:) and registers an atexit handler; the version block is after it at line 974. The version_short branch exits via exit(0), which triggers atexit handlers. Confirmed by reading the surrounding code.</sub>

</details>

- `gptme/cli/main.py` — Adds `--version-short` option and a `show_version` flag; prints just `__version__` and exits when `version_short` is set, and uses `show_version` in the three dispatch guards.
- `tests/test_cli.py` — Adds `test_version_short` verifying the exact version string is printed, importing `__version__` for the assertion.
- `tests/test_search_alias.py` — Adds `test_search_does_not_swallow_version_short` verifying `--version-short` prevents the search alias and prints only the version.

</details>

<sub>Reviewed `d4bee48dfb52` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 87s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/cli/main.py
TimeToBuildBob: Fixed in 8895c799e — added `and not show_version` to the `if profile:` condition. Profiling setup (including the atexit handler) is now skipped entirely when a version flag is active, so `--version-short --profile` emits only the version string.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Not a huge fan of this CLI option proliferation. Same for `--version-json`. Consider cleaning up/rethinking the noisy options in favor of a cleaner design.

TimeToBuildBob:
Fair point — `--version-short` (and a future `--version-json`) is noisy option proliferation. The underlying need is legitimate (scripts want a bare version string), but bolting on flags is the wrong shape.

Two cleaner directions:

1. **Fix `--version` to output just the version string** (`0.32.0`, no prefix) — makes it immediately script-friendly and removes the need for `--version-short` entirely. This is a minor breaking change on the human-facing output format but a common Unix convention.

2. **`gptme version [--format json]` subcommand** — groups version-related output under one verb, keeps `--version` for humans, adds structured output as a subcommand flag rather than a top-level option.

Option 1 is simpler and probably the right call. I'll open a follow-up issue to deprecate `--version-short` and fix `--version` output format — then revert this PR's flag in a subsequent PR once the cleaner path lands.

TimeToBuildBob:
Tracking the cleanup in #3538.

PR #3537, opened 2026-08-13
refactor(cli): move Bob-specific status fields out of core; add StatusProvider extension point

### LINKED ISSUE #3534: refactor(cli): move Bob-specific status fields out of core; define secure StatusProvider extension point
## Background

`gptme/gptme#3528` added `gptme-util status --json` with structured JSON output for the status command. Erik merged it but flagged two concerns worth tracking.

## Concern 1: Bob-specific fields in gptme core

`cmd_status.py` contains logic that is specific to Bob's workspace:
- `_is_bob_workspace()` — detects Bob by checking for `tasks/` dir + `gptme.toml`
- `_active_tasks()` — calls `gptodo` to read Bob's task queue
- `_blockers()` / `_ready_tasks()` — reads Bob's GTD-style task states
- `_service_status()` — checks `bob-autonomous.service` and `bob-operator-loop.service`
- `_TRACKED_REPOS` — hardcodes Bob's PR repos (ErikBjare/bob, gptme/gptme-cloud)
- `_journal_entries()` — reads Bob's `journal/` directory

These have no meaning outside Bob's agent workspace. They make `gptme-util status` a partially-agent-specific command living in core.

**Better home**: `gptme-agent status` (needs updating anyway per Erik's note), or a dedicated agent workspace command in contrib. Core `gptme-util status` should produce only generic git/PR/disk information that's useful to any gptme user.

## Concern 2: Repo-execution security risk in any future extension mechanism

The current code detects Bob's workspace by filesystem heuristics (`_is_bob_workspace`). An obvious next step would be making `status` **extensible** — letting a workspace provide its own status fields via a config file or a Python plugin.

However, auto-loading Python or shell code from the current directory/repo is a security risk: running `gptme-util status` in a malicious repo could execute arbitrary code. This is the same class of concern as pre-merge git hooks that source repo-local scripts.

**Before adding any extension mechanism**, the extension surface needs to be designed with:
- A secure plugin interface (no `importlib.import_module` from cwd, no `exec()` of repo files)
- Explicit user opt-in (the user must configure a trusted plugin path, not auto-loaded from cwd)
- Or: keep agent-specific logic entirely outside core (Concern 1 solution sidesteps this entirely)

## Proposed resolution

**Short-term (already merged):** Accept the current state as tech debt — the code works and the security risk is latent (no extension mechanism exists yet).

**Medium-term:**
1. Move Bob-specific fields out of `cmd_status.py` into a gptme-contrib command (e.g. `gptme-agent status` or `gptme-bob status`)
2. Define a clean `StatusProvider` protocol in core that external packages can implement — registered via entry points, not filesystem scanning
3. `gptme-util status` invokes registered providers only; no cwd-based auto-loading ever

This keeps core clean, makes agent-specific status extensible without security risk, and gives `gptme-agent status` a natural home for operational fields.

TimeToBuildBob:
## Summary

Resolves #3534.

Moves Bob-specific status fields out of `gptme` core and defines a secure `StatusProvider` extension point that external packages can register via Python entry points.

### New: `gptme/status_provider.py`

Defines the `StatusProvider` protocol:

```python
class StatusProvider(Protocol):
    @property
    def name(self) -> str: ...
    def collect(self) -> dict[str, object]: ...        # merged into --json output
    def narrative_sections(self) -> list[str]: ...    # appended to narrative doc
```

And `load_providers()` which discovers implementations from the `gptme.status_providers` entry-point group.

**Security model**: `importlib.metadata.entry_points` only sees *installed* packages — the current working directory is never scanned. Running `gptme-util status` inside a malicious repository cannot auto-load arbitrary code. This directly addresses Concern 2 in the issue.

### Updated: `gptme/cli/cmd_status.py`

Removed all Bob-specific logic:
- `_is_bob_workspace()` — filesystem heuristic
- `_active_tasks()` / `_blockers()` / `_ready_tasks()` — `gptodo` calls
- `_service_status()` / `_dead_timers()` — `bob-*.service` systemd units
- `_journal_entries()` — Bob's `journal/` directory
- `_TRACKED_REPOS` — hardcoded Bob PR repos
- `section_services()` / `section_blockers()` / `section_ready_next()`

Core now collects only generic, workspace-agnostic data:
- `session_id` (env vars)
- `recent_commits` (git log)
- `disk_usage` (shutil.disk_usage)

The generic `_pr_queue()` helper is kept (useful for providers that check GitHub PRs), but core no longer calls it with any hardcoded repo list.

`build_document()`, `build_table_document()`, and `_status_data()` accept an explicit `providers` list (for testing). When omitted, `load_providers()` is called automatically.

### Updated: `tests/test_cli_status.py`

- Added `_FakeProvider` test double satisfying the `StatusProvider` protocol
- Updated tests to reflect the clean core (no Bob-specific monkeypatching)
- Added provider-focused tests: JSON merging, narrative injection, table rows, broken-provider isolation, `load_providers()` smoke test

## How a Bob provider would look (gptme-contrib side)

```toml
# packages/gptme-bob/pyproject.toml
[project.entry-points."gptme.status_providers"]
bob = "gptme_bob.status:make_provider"
```

```python
# packages/gptme-bob/src/gptme_bob/status.py
from gptme.status_provider import StatusProvider
from gptme.cli.cmd_status import _pr_queue, _pr_queue_display

_TRACKED_REPOS = [("gptme/gptme", 10), ("ErikBjare/bob", None), ...]

class BobStatusProvider:
    name = "bob"

    def collect(self) -> dict:
        return {
            "active_tasks": _active_tasks(3),
            "pr_queue": _pr_queue(_TRACKED_REPOS),
            "journal_entries": _journal_entries(5),
            "services": _service_status(),
            ...
        }

    def narrative_sections(self) -> list[str]:
        return [section_services(), section_blockers(), section_ready_next()]

def make_provider() -> StatusProvider:
    return BobStatusProvider()
```

## Tests

```
26 passed in 1.27s
```

### DIFF
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -398,7 +398,21 @@ def build_document(providers: list[StatusProvider] | None = None) -> str:
     for provider in providers:
         try:
             extra_sections = provider.narrative_sections()
-            sections.extend(extra_sections)
+            if not isinstance(extra_sections, list):
+                logger.debug(
+                    "Provider %r narrative_sections() returned %r, expected list[str]; skipping",
+                    _provider_name(provider),
+                    type(extra_sections).__name__,
+                )
+            else:
+                valid = [s for s in extra_sections if isinstance(s, str)]
+                if len(valid) < len(extra_sections):
+                    logger.debug(
+                        "Provider %r narrative_sections() returned %d non-string item(s); dropped",
+                        _provider_name(provider),
+                        len(extra_sections) - len(valid),
+                    )
+                sections.extend(valid)
         except Exception as exc:
             logger.debug(
                 "Provider %r narrative_sections() failed: %s",
--- a/gptme/cli/cmd_status.py
+++ b/gptme/cli/cmd_status.py
@@ -3,8 +3,13 @@
 Provides ``gptme-util status`` (via lazy registration in ``util.py``) and
 ``gptme-status`` (standalone entry point registered in ``pyproject.toml``).
 
-Produces a compact, human-readable briefing: active work, PR queue, service
-health, blockers, and ready backlog items.
+Produces a compact, human-readable briefing: recent commits, disk usage, and
+any extra sections contributed by installed :class:`~gptme.status_provider.StatusProvider`
+plugins registered under the ``gptme.status_providers`` entry-point group.
+
+Agent- or workspace-specific status fields (task queues, service health,
+blockers, journal entries) live in dedicated packages that register a provider.
+Core only collects generic information that is useful to any gptme user.
 """
 
 from __future__ import annotations
@@ -21,12 +26,13 @@ from typing import TypedDict
 
 import click
 
+from ..status_provider import StatusProvider, load_providers
 from ..util.git_cmd import GIT_CMD
 
 logger = logging.getLogger(__name__)
 
 
-# ── helpers ────────────────────────────────────────────────────────────
+# ── generic helpers ────────────────────────────────────────────────────
 
 
 def _run(cmd: list[str], *, timeout: int = 10) -> str:
@@ -44,39 +50,6 @@ def _git_root() -> Path | None:
     return Path(raw) if raw else None
 
 
-def _gh_user() -> str:
-    """Return the current gh CLI authenticated username."""
-    return _run(["gh", "api", "user", "--jq", ".login"], timeout=10)
-
-
-def _is_bob_workspace() -> bool:
-    """Detect if we are inside Bob's workspace by checking for Bob-specific files."""
-    root = _git_root()
-    if not root:
-        return False
-    return (root / "tasks").is_dir() and (root / "gptme.toml").is_file()
-
-
-def _active_tasks(lines: int = 3) -> list[dict]:
-    """Parse gptodo status --compact output for active tasks (Bob workspace)."""
-    raw = _run(["gptodo", "status", "--compact"], timeout=15)
-    if not raw:
-        return []
-    tasks: list[dict] = []
-    for line in raw.splitlines():
-        line = line.strip()
-        # Lines look like "  task-id  Title text here  (N ago)"
-        if not line or line.startswith("📋") or "0 tasks" in line or "Summary" in line:
-            continue
-        parts = line.split(None, 1)
-        if len(parts) >= 2:
-            task_id = parts[0]
-            # Strip trailing " (N unit ago)" timestamp to get the full title
-            title = re.sub(r"\s+\(\d+\s+\w+\s+ago\)\s*$", "", parts[1]).strip()
-            tasks.append({"_id": task_id, "title": title})
-    return tasks[:lines]
-
-
 def _recent_commits(n: int = 3) -> list[str]:
     raw = _run([GIT_CMD, "log", "--oneline", f"-{n}", "--no-merges"])
     return raw.splitlines() if raw else []
@@ -91,8 +64,16 @@ class _PRQueueRow(TypedDict):
 def _pr_queue(
     repos: list[tuple[str, int | None]], author: str | None = None
 ) -> list[_PRQueueRow]:
+    """Fetch open PR counts for the given repos.
+
+    This is a generic helper exposed for use by :class:`~gptme.status_provider.StatusProvider`
+    implementations.  Core ``build_document`` / ``_status_data`` do **not** call
+    it with any hardcoded repo list.
+    """
+    if not repos:
+        return []
     if author is None:
-        author = _gh_user() or "TimeToBuildBob"
+        author = _run(["gh", "api", "user", "--jq", ".login"], timeout=10) or ""
     rows: list[_PRQueueRow] = []
     for repo, cap in repos:
         prs_json = _run(
@@ -117,8 +98,7 @@ def _pr_queue(
             prs = json.loads(prs_json)
         except json.JSONDecodeError:
             continue
-        count = len(prs)
-        rows.append({"repo": repo, "count": count, "cap": cap})
+        rows.append({"repo": repo, "count": len(prs), "cap": cap})
     return rows
 
 
@@ -127,59 +107,6 @@ def _pr_queue_display(count: int, cap: int | None) -> str:
     if cap is not None:
         return f"{count}/{cap}" + (" ⚠ at limit" if count >= cap else "")
     return str(count)
-
-
-def _service_status() -> list[dict[str, str]]:
-    services = [
-        ("Operator loop", "bob-operator-loop.service"),
-        ("Autonomous", "bob-autonomous.service"),
-    ]
-    results: list[dict[str, str]] = []
-    for label, unit in services:
-        status = _run(["systemctl", "--user", "is-active", unit])
-        icon = "✓" if status == "active" else ("⚠" if status == "activating" else "✗")
-        results.append({"label": label, "icon": icon, "status": status})
-    return results
-
-
-def _dead_timers() -> int:
-    out = _run(["systemctl", "--user", "list-timers", "--all"])
-    return sum(
-        1
-        for line in out.splitlines()
-        if "dead" in line.lower() and "bob-" in line.lower()
-    )
-
-
-def _blockers(limit: int = 3) -> list[dict]:
-    raw = _run(["gptodo", "ready", "--state", "waiting", "--jsonl"], timeout=15)
-    if not raw:
-        return []
-    blockers: list[dict] = []
-    for line in raw.splitlines():
-        try:
-            t = json.loads(line)
-        except json.JSONDecodeError:
-            continue
-        if t.get("waiting_for"):
-            blockers.append(t)
-    return blockers[:limit]
-
-
-def _ready_tasks(limit: int = 3) -> list[dict]:
-    raw = _run(["gptodo", "ready", "--state", "backlog", "--jsonl"], timeout=15)
-    if not raw:
-        return []
-    tasks: list[dict] = []
-    for line in raw.splitlines():
-        try:
-            t = json.loads(line)
-        except json.JSONDecodeError:
-            continue
-        if t.get("waiting_for") or t.get("wait"):
-            continue
-        tasks.append(t)
-    return tasks[:limit]
 
 
 def _session_id() -> str:
@@ -217,29 +144,6 @@ def _markdown_table_cell(value: object) -> str:
     return text.replace("|", r"\|")
 
 
-def _journal_entries(limit: int = 5) -> list[str]:
-    """Return the last N journal entry filenames (Bob workspace)."""
-    root = _git_root()
-    if not root:
-        return []
-    journal_dir = root / "journal"
-    if not journal_dir.is_dir():
-        return []
-    entries: list[Path] = []
-    for day_dir in sorted(journal_dir.iterdir(), reverse=True):
-        if not day_dir.is_dir():
-            continue
-        day_entries = [
-            entry.relative_to(root)
-            for entry in sorted(day_dir.iterdir(), reverse=True)
-            if entry.is_file() and entry.suffix == ".md"
-        ]
-        entries.extend(day_entries)
-        if len(entries) >= limit:
-            break
-    return [str(e) for e in entries[:limit]]
-
-
 def _strip_markdown(doc: str) -> str:
     """Strip Markdown formatting for plain-text output."""
     lines = []
@@ -253,7 +157,7 @@ def _strip_markdown(doc: str) -> str:
     return "\n".join(lines)
 
 
-# ── sections ──────────────────────────────────────────────────────────
+# ── core sections ──────────────────────────────────────────────────────
 
 
 def section_header() -> str:
@@ -264,154 +168,79 @@ def section_header() -> str:
     return f"# gptme Status — {now}\n\n**Model**: {model}{agent_part}\n"
 
 
-def section_active_work(is_bob: bool = False) -> str:
+def section_active_work() -> str:
     lines = ["## Active Work"]
-    if is_bob:
-        tasks = _active_tasks(3)
-        if tasks:
-            lines.extend(
-                f"- **Task**: `{t['_id']}` — {t.get('title', '')[:60]}" for t in tasks
-            )
     commits = _recent_commits(3)
     if commits:
         lines.append("- **Recent commits** (last 3):")
         for commit in commits:
             sha, _, msg = commit.partition(" ")
             lines.append(f"  - `{sha}` {msg[:65]}")
-    return "\n".join(lines)
-
-
-def section_pr_queue() -> str:
-    lines = ["## PR Queue"]
-    rows = _pr_queue(_TRACKED_REPOS)
-    if rows:
-        lines.append("| Repo | Open |")
-        lines.append("|------|------|")
-        for row in rows:
-            display = _pr_queue_display(row["count"], row["cap"])
-            lines.append(f"| {row['repo']} | {display} |")
     else:
-        lines.append("- Unable to fetch PR data")
+        lines.append("- No recent commits")
     return "\n".join(lines)
 
 
-def section_services() -> str:
-    lines = ["## Services"]
-    services = _service_status()
-    lines.extend(f"- {svc['label']}: {svc['icon']} {svc['status']}" for svc in services)
-    dead = _dead_timers()
-    if dead:
-        lines.append(f"- ⚠ {dead} dead bob-* timer(s)")
-    return "\n".join(lines)
-
-
-def section_blockers() -> str:
-    lines = ["## Top Blockers"]
-    blockers = _blockers(3)
-    if blockers:
-        for t in blockers:
-            wf = str(t.get("waiting_for", "")).split("\n")[0][:70]
-            since = t.get("waiting_since", "")
-            since_str = f" (since {since})" if since else ""
-            lines.append(f"- `{t['id']}`: {wf}{since_str}")
-    else:
-        lines.append("- No active blockers with waiting_for set")
-    return "\n".join(lines)
-
-
-def section_ready_next() -> str:
-    lines = ["## Ready Next (top 3)"]
-    ready = _ready_tasks(3)
-    if ready:
-        for i, t in enumerate(ready, 1):
-            title = str(t.get("name", t.get("id", "")))[:65]
-            lines.append(f"{i}. `{t['id']}` — {title}")
-    else:
-        lines.append("- No ready backlog tasks found")
-    return "\n".join(lines)
+def section_disk() -> str:
+    root = _git_root()
+    disk = _disk_usage(root)
+    return f"## Disk\n\n- **Usage**: {disk}"
 
 
 # ── build ─────────────────────────────────────────────────────────────
 
 
-_TRACKED_REPOS: list[tuple[str, int | None]] = [
-    ("gptme/gptme", 10),
-    ("gptme/gptme-cloud", 3),
-    ("ErikBjare/bob", None),
-    ("gptme/gptme-contrib", None),
-]
+def _status_data(providers: list[StatusProvider] | None = None) -> dict[str, object]:
+    """Collect status data shared by JSON and presentation renderers.
 
+    Core fields are generic and workspace-agnostic.  Extra fields from installed
+    :class:`~gptme.status_provider.StatusProvider` implementations are merged in
+    at the top level.
 
-def _status_data() -> dict[str, object]:
-    """Collect status data shared by JSON and presentation renderers."""
-    is_bob = _is_bob_workspace()
+    Parameters
+    ----------
+    providers:
+        Pre-loaded providers (for testing).  When ``None``, :func:`~gptme.status_provider.load_providers`
+        is called automatically.
+    """
+    if providers is None:
+        providers = load_providers()
+
     root = _git_root()
     status_data: dict[str, object] = {
         "timestamp": datetime.now(timezone.utc).isoformat(),
         "session_id": _session_id(),
-        "active_tasks": [
-            {"id": task.get("_id", task.get("id", "")), "title": task.get("title", "")}
-            for task in _active_tasks(3)
-        ],
         "recent_commits": _recent_commits(3),
-        "pr_queue": _pr_queue(_TRACKED_REPOS),
         "disk_usage": _disk_usage(root),
-        "journal_entries": _journal_entries(5),
     }
-    if is_bob:
-        status_data.update(
-            services=_service_status(),
-            dead_timers=_dead_timers(),
-            blockers=_blockers(3),
-            ready_tasks=_ready_tasks(3),
-        )
+
+    for provider in providers:
+        try:
+            extra = provider.collect()
+            status_data.update(extra)
+        except Exception as exc:
+            logger.debug("Provider %r collect() failed: %s", provider.name, exc)
+
     return status_data
 
 
-def build_json_status() -> str:
+def build_json_status(providers: list[StatusProvider] | None = None) -> str:
     """Build the structured JSON status document."""
-    return json.dumps(_status_data(), indent=2)
+    return json.dumps(_status_data(providers), indent=2)
 
 
-def build_table_document() -> str:
+def build_table_document(providers: list[StatusProvider] | None = None) -> str:
     """Build a machine-readable markdown table of session state."""
-    is_bob = _is_bob_workspace()
+    if providers is None:
+        providers = load_providers()
+
     now = datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M UTC")
     session_id = _session_id()
     root = _git_root()
 
-    # active_task
-    active = _active_tasks(1)
-    active_task = active[0].get("_id", "none") if active else "none"
-
-    # last_commit
     commits = _recent_commits(1)
     last_commit = commits[0] if commits else "none"
-
-    # pending_prs
-    pr_rows = _pr_queue(_TRACKED_REPOS)
-    pending_prs = (
-        ", ".join(
-            f"{r['repo']}:{_pr_queue_display(r['count'], r['cap'])}" for r in pr_rows
-        )
-        if pr_rows
-        else "none"
-    )
-
-    # waiting_for
-    blockers = _blockers(1)
-    waiting_for = (
-        _markdown_table_cell(blockers[0].get("waiting_for", "none"))
-        if blockers
-        else "none"
-    )
-
-    # disk_usage
     disk = _markdown_table_cell(_disk_usage(root))
-
-    # journal_entries
-    journals = _journal_entries(5)
-    journal_str = _markdown_table_cell(", ".join(journals) if journals else "none")
 
     lines = [
         f"# gptme Status — {now}",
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
Addressed the three Greptile findings in commit b00e80f44:

**P1 — Malformed factory result crashes via `provider.name` in error handler**
`load_providers()` now validates the factory return value against the `StatusProvider` runtime-checkable protocol (`isinstance(provider, StatusProvider)`) before appending. A `None` or protocol-incomplete return is logged at DEBUG and skipped — the error handler never sees it.

**P1 — Non-JSON-serializable provider values crash `--json`**
`build_json_status()` now passes `default=_json_default` to `json.dumps`. The fallback converts unknown objects (`datetime`, `Path`, `Decimal`, bytes, …) to their `str()` representation instead of raising `TypeError`.

**P2 — Provider keys silently overwrite core fields**
Added `_CORE_KEYS = frozenset({"timestamp", "session_id", "recent_commits", "disk_usage"})`. `_status_data()` now checks each provider key against this set and skips (with a DEBUG log) any collision rather than calling `update()` blindly. Later-provider collision with an earlier provider's non-core key is permitted but also logged.

Three new tests cover all three scenarios: `test_load_providers_skips_malformed_factory_result`, `test_status_json_provider_unserializable_value_does_not_crash`, `test_status_json_provider_cannot_overwrite_core_keys`. All 29 tests green locally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining P1 finding in commit 86cbc6864:

**P1 — Malformed provider name escapes isolation (provider with raising name property)**

The root issue: `isinstance(provider, StatusProvider)` with a `@runtime_checkable` Protocol only checks that the `name` attribute *exists* on the type — it does not call the property getter. A provider whose `name` property raises passes the isinstance check but then crashes the error-handler `logger.debug()` calls in `cmd_status.py`, which dereference `provider.name` inside `except` blocks and therefore escape isolation entirely.

Two-pronged fix:

**Primary (load_providers)**: After the isinstance check, probe `provider.name` in a try/except before appending the provider. A raising name getter causes the provider to be skipped (logged at DEBUG using the entry-point name for context).

**Defense-in-depth (cmd_status)**: Added `_provider_name()` helper that wraps `provider.name` in try/except and returns `"<unnamed-provider>"` on error. All error-handler `logger.debug()` calls now use `_provider_name(provider)` instead of bare `provider.name`.

New test: `test_load_providers_skips_raising_name_provider` — creates a structurally-conforming provider (`isinstance` returns True) whose `name` property always raises, verifies it is silently dropped from `load_providers()` output.

All 30 tests green locally.

TimeToBuildBob:
@greptileai review

gptme/status_provider.py
```diff
@@ -0,0 +1,104 @@
...
+    from importlib.metadata import entry_points  # lazy — keeps startup fast
+
+    providers: list[StatusProvider] = []
+    try:
+        eps = entry_points(group="gptme.status_providers")
+        for ep in eps:
+            try:
+                factory = ep.load()
+                provider = factory()
+                providers.append(provider)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b00e80f44. \`load_providers()\` now validates the factory result against the protocol before appending:

\`\`\`python
provider = factory()
if not isinstance(provider, StatusProvider):
    logger.debug(...)
    continue
\`\`\`

A factory returning \`None\` or an object missing protocol members is logged at DEBUG and skipped, so it never reaches the renderers that dereference \`provider.name\`. Regression test: \`test_load_providers_skips_malformed_factory_result\` in \`tests/test_cli_status.py\`.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,124 @@ def section_header() -> str:
...
-            services=_service_status(),
-            dead_timers=_dead_timers(),
-            blockers=_blockers(3),
-            ready_tasks=_ready_tasks(3),
-        )
+
+    for provider in providers:
+        try:
+            extra = provider.collect()
+            status_data.update(extra)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b00e80f44. \`build_json_status()\` now passes a fallback serialiser:

\`\`\`python
return json.dumps(_status_data(providers), indent=2, default=_json_default)
\`\`\`

\`_json_default\` stringifies any value \`json.dumps\` cannot encode natively, so a provider returning \`datetime\`, \`Path\`, \`Decimal\`, or bytes degrades to a string field instead of raising \`TypeError\` and killing the whole command. Regression test: \`test_status_json_provider_unserializable_value_does_not_crash\`.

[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b00e80f44. \`_status_data()\` no longer blind-\`update()\`s provider mappings — it iterates keys, refuses reserved core names (\`session_id\`, \`disk_usage\`, …), and logs a collision when a later provider would overwrite an earlier contribution:

\`\`\`python
logger.warning("Provider %r key %r collides with an earlier provider's key — overwriting", ...)
\`\`\`

Core fields are protected outright (skip + log); provider-vs-provider collisions are last-writer-wins but no longer silent. Regression test: \`test_status_json_provider_cannot_overwrite_core_keys\`.

gptme/status_provider.py
```diff
@@ -0,0 +1,113 @@
...
+    from importlib.metadata import entry_points  # lazy — keeps startup fast
+
+    providers: list[StatusProvider] = []
+    try:
+        eps = entry_points(group="gptme.status_providers")
+        for ep in eps:
+            try:
+                factory = ep.load()
+                provider = factory()
+                if not isinstance(provider, StatusProvider):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 86cbc6864 — correct catch, the isinstance check does not call property getters. Two-pronged fix:

1. \`load_providers()\` probes \`provider.name\` in a try/except right after the isinstance check and skips any provider whose getter raises, logging at DEBUG.
2. \`cmd_status.py\` gained \`_provider_name(provider)\` as defense-in-depth — every error-handler log line goes through it, returning \`"<unnamed-provider>"\` instead of raising inside an \`except\` block.

So even a provider that starts raising *after* load time can no longer escape isolation. Regression test: \`test_load_providers_skips_raising_name_provider\`.

TimeToBuildBob:
✅ **Greptile re-review: 4/5** (improved from 3/5)

Commit 86cbc6864 addressed the P1 issue (provider.name guard). Remaining finding is contained to gptme/cli/cmd_status.py.

Cross-repo PR requires human merge from gptme maintainer. CI is mostly passing (2 test jobs pending but expected to complete shortly).

CC @ErikBjare for merge review.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,175 @@ def section_header() -> str:
...
-        lines.append(f"| services | {svc_str} |")
-        dead = _dead_timers()
-        if dead:
-            lines.append(f"| dead_timers | {dead} |")
+    for provider in providers:
+        try:
+            extra = provider.collect()
+            for key, val in extra.items():
+                cell = _markdown_table_cell(val)
+                lines.append(f"| {key} | {cell} |")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b15233a. The table already skipped duplicate keys via `seen_keys`. The JSON path in `_status_data()` was the inconsistency — it logged 'overwriting' and then unconditionally wrote the later provider's value. Fixed by adding a `continue` after the collision log, so both formats now use first-writer-wins. New test `test_status_json_cross_provider_collision_first_writer_wins` covers this.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,190 @@ def section_header() -> str:
...
+            extra = provider.collect()
+            for key, val in extra.items():
+                if key in seen_keys:
+                    logger.debug(
+                        "Provider %r key %r is already present in the status"
+                        " table — skipping to prevent contradictory duplicate rows",
+                        _provider_name(provider),
+                        key,
+                    )
+                    continue
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b15233a — both formats now use first-writer-wins. Root cause: `_status_data()` had a collision check that logged the collision but then fell through to `status_data[key] = val` anyway, so the later provider won in JSON. Added `continue` after the log, consistent with `build_table_document()`'s `seen_keys` guard. See `test_status_json_cross_provider_collision_first_writer_wins` for the regression test.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,192 @@ def section_header() -> str:
...
-    ("ErikBjare/bob", None),
-    ("gptme/gptme-contrib", None),
-]
+def _json_default(obj: object) -> object:
+    """Fallback JSON serialiser for non-standard types returned by providers.
 
+    Converts unknown objects to their ``str()`` representation so that
+    ``json.dumps`` never raises ``TypeError`` on provider-contributed values.
+    """
+    return str(obj)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 68a8e9c. Wrapped `str(obj)` in a try/except in `_json_default()` — if the object's `__str__()` raises, the sentinel string `"<unserializable>"` is returned instead of propagating the exception. Added `test_status_json_provider_str_raises_does_not_crash` to cover this path directly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining 4/5 finding in commit 6f83752e9:

**Remaining P1 — Non-string provider mapping keys bypass the value fallback and abort `--json` output**

Root cause: `json.dumps()` requires string keys. The `default=_json_default` fallback only handles non-serializable *values* — it is never called for non-string *keys*. A provider returning `{(1, 2): "value"}` would pass all existing guards (isinstance check, `_CORE_KEYS` check, collision check) but cause `json.dumps()` to raise `TypeError` at serialisation time.

Fix: added `isinstance(key, str)` validation at the top of the per-key loops in both `_status_data()` and `build_table_document()`. Non-string keys are logged at DEBUG level and silently dropped.

Two new regression tests added:
- `test_status_json_provider_non_string_key_does_not_crash` — tuple key `(1, 2)` is dropped; string key still present; JSON output valid
- `test_build_table_document_provider_non_string_key_skipped` — same for table output

All 36 tests green locally.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,214 @@ def section_header() -> str:
...
+                    continue
+                if key in status_data:
+                    logger.debug(
+                        "Provider %r key %r collides with an earlier provider's key — skipping"
+                        " (first-writer wins, consistent with table output)",
+                        _provider_name(provider),
+                        key,
+                    )
+                    continue
+                status_data[key] = val
```
TimeToBuildBob: ❌ **P1** — _status_data filters top-level keys for string type but does not recursively sanitize values. If a provider returns a nested dict with non-string keys (e.g., {1: "x"}) as a value for any top-level key, json.dumps will raise TypeError: keys must be str because the default parameter only handles non-serializable values, not dict keys. The except clauses around provider.collect() do not catch this because the exception occurs inside json.dumps at the very end of build_json_status. Consequence: a single malicious or buggy provider takes down the entire --json command with an uncaught traceback, violating the PR's goal of isolating broken providers. The existing test test_status_json_provider_non_string_key_does_not_crash only covers top-level non-string keys, so this gap is untested.

TimeToBuildBob:
## 🤖 AI code review

This PR removes Bob-specific status fields (gptodo tasks, systemd services, journal entries, hardcoded PR repos) from gptme's status CLI and introduces a new `StatusProvider` protocol in `gptme/status_provider.py` with entry-point-based discovery. `cmd_status.py` now collects only generic data (session id, recent commits, disk usage) and merges provider-supplied fields/sections into JSON, table, and narrative outputs, with collision handling and defensive error isolation. Tests are updated to use a fake provider and cover collision handling and broken-provider scenarios.

### Confidence Score: N/A

**Score withheld** — 2 findings were dropped by policy and nothing survived, so a clean score here would describe the filter rather than the code. Treat this as *not scored*, not as approved.

> ⚠️ **2 findings found but NOT reported** (2×P2): the review-round cap was reached, so P2 findings stopped blocking. They were raised by this review and dropped by policy — not absent, and not resolved.

🚫 **No findings reported — but not because none were found.** All 2 findings raised on this head were dropped by policy, as stated above. This is a filtered result, not a clean one.

- `gptme/cli/cmd_status.py` — Removed all Bob-specific functions and hardcoded repo list; added provider merging into `_status_data`, `build_table_document`, and `build_document`; changed `_pr_queue` default author to use gh user or empty string.
- `gptme/status_provider.py` — New file defining the `StatusProvider` protocol, `load_providers()` entry-point loader with name probing and error isolation, and documentation of the security model.
- `tests/conftest.py` — Added an autouse fixture `cleanup_logged_warnings` that clears `_logged_warnings` after each test.
- `tests/test_cli_status.py` — Rewrote tests to remove Bob-specific monkeypatching, added provider-focused tests for JSON merging, narrative injection, table rows, collision handling, and broken-provider isolation.

</details>

| `6f83752e989b` | 3/5 | 1 | llm | 2026-08-13 17:02 UTC |
| `528c8575808e` | 3/5 | 3 | llm | 2026-08-13 20:15 UTC |
| `76c5d6309327` | 3/5 | 1 | llm | 2026-08-13 21:47 UTC |
| `0102280318a8` | 4/5 | 1 | llm | 2026-08-14 05:24 UTC |
| `4075fe276917` | 4/5 | 1 | llm | 2026-08-16 13:24 UTC |
| `8599415ca057` | 3/5 | 1 | llm | 2026-08-16 19:40 UTC |

</details>

<sub>Reviewed `8599415ca057` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 132s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed in c0c8cd5.

Added `_sanitize_nested_dict_keys()` which recursively converts non-string dict keys to their `str()` form before inserting provider values into `status_data`. Applied at insertion time so the entire value tree is safe before it reaches `json.dumps`.

The existing test `test_status_json_provider_non_string_key_does_not_crash` only covered top-level non-string keys (which `_status_data`'s per-key filter already dropped). New test `test_status_json_provider_nested_non_string_key_does_not_crash` covers the gap: a provider returning `{"nested": {1: "int_key_val"}}` now serializes cleanly with the int key coerced to `"1"` rather than raising `TypeError`.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,240 @@ def section_header() -> str:
...
+    entire value tree is safe to pass to ``json.dumps``.
+    """
+    if isinstance(obj, dict):
+        return {
+            (k if isinstance(k, str) else str(k)): _sanitize_nested_dict_keys(v)
+            for k, v in obj.items()
+        }
+    if isinstance(obj, list):
+        return [_sanitize_nested_dict_keys(v) for v in obj]
+    return obj
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 528c85758 — added a `tuple` branch to `_sanitize_nested_dict_keys` so tuples are recursively traversed (elements sanitized, tuple reconstructed). Added `test_status_json_provider_tuple_nested_non_string_key_does_not_crash` to cover the `{"data": ({1: "val"},)}` shape.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_status.py
TimeToBuildBob: Fixed: commits c0c8cd5e (added `_sanitize_nested_dict_keys` called on every provider value before merge) and 528c8575 (extended it to traverse tuples). The function now recursively converts non-string dict keys to str across dicts, lists, and tuples, so no nested structure can pass through to `json.dumps` with invalid keys.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,244 @@ def section_header() -> str:
...
+    """Build the narrative Markdown status document."""
+    if providers is None:
+        providers = load_providers()
+
     sections: list[str] = [
         section_header(),
-        section_active_work(is_bob=is_bob),
-        section_pr_queue(),
+        section_active_work(),
+        section_disk(),
```
TimeToBuildBob, line 407: ❌ **P1** — The `build_document` function now only calls `section_header()`, `section_active_work()`, and `section_disk()`, dropping all Bob-specific sections. Previously the narrative output included a PR queue table, service health, blockers, and ready-next tasks when in a Bob workspace. Any user relying on these sections will no longer see them after upgrading; the functionality is moved to an external provider that is not part of this PR, so there is no replacement until the user installs a separate package. This is a user-visible regression in the status command's output.

gptme/cli/cmd_status.py
```diff
@@ -91,8 +64,16 @@ class _PRQueueRow(TypedDict):
...
+
+    This is a generic helper exposed for use by :class:`~gptme.status_provider.StatusProvider`
+    implementations.  Core ``build_document`` / ``_status_data`` do **not** call
+    it with any hardcoded repo list.
+    """
+    if not repos:
+        return []
     if author is None:
-        author = _gh_user() or "TimeToBuildBob"
+        author = _run(["gh", "api", "user", "--jq", ".login"], timeout=10) or ""
```
TimeToBuildBob, line 76: ⚠️ **P2** — In `_pr_queue`, when `author` is None, the default is now an empty string instead of `"TimeToBuildBob"`. The function invokes `gh pr list --author ""`, which does not filter by a specific author and will therefore return PRs from all authors. External provider code that calls `_pr_queue(repos)` without an explicit author will see inflated PR counts compared to the previous behavior. Since the default is now empty, any provider relying on the old default (which filtered to one user) will observe a different result, potentially leading to incorrect PR-count reporting.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,244 @@ def section_header() -> str:
...
+
+    This helper traverses :class:`dict`, :class:`list`, and :class:`tuple`
+    values recursively, converting any non-string key to its ``str()``
+    representation so the entire value tree is safe to pass to ``json.dumps``.
+    Tuples are reconstructed as tuples so the original container type is
+    preserved; ``json.dumps`` serialises them as JSON arrays, the same as lists.
+    """
+    if isinstance(obj, dict):
+        return {
+            (k if isinstance(k, str) else str(k)): _sanitize_nested_dict_keys(v)
```
TimeToBuildBob, line 246: ⚠️ **P2** — In `_sanitize_nested_dict_keys`, when a nested dict has a non-string key, the key is converted via `str(k)`. If that `str()` call raises (e.g., a broken or malicious object's `__str__`), the exception propagates out of the function, hitting the `except Exception` in `_status_data` for that provider. As a result, none of the provider's keys are added — even the valid ones — because the exception aborts the provider's iteration. This silently discards all data from a provider that has a single problematic nested key, and is not covered by the tests, which only exercise a value whose `__str__` raises, not a key.

gptme/cli/cmd_status.py
TimeToBuildBob: False positive — this is the PR's design intent. `build_document` (lines 398–407) iterates all loaded providers and calls `provider.narrative_sections()`, extending the document. Bob-specific sections (PR queue, services, blockers, ready-next) will be contributed by a Bob-specific `StatusProvider` implementation registered as a package entry point — exactly the extension mechanism this PR introduces. The functionality is not dropped; it's moved out of core into an installable provider.

gptme/cli/cmd_status.py
TimeToBuildBob: False positive — line 76 queries the current GH user via `gh api user --jq .login` and only falls back to `""` if that call returns nothing (e.g., not logged in). Providers that call `_pr_queue` should pass `author` explicitly; the default discovers whoever is logged in. The old hardcoded `"TimeToBuildBob"` was Bob-specific and does not belong in core — removing it is correct.

gptme/cli/cmd_status.py
TimeToBuildBob: Acknowledged but acceptable: if `str(k)` raises on a nested dict key, the exception is caught by the `try/except Exception` block at line 312 in `_status_data`, which logs at DEBUG and skips that provider. The failing provider's data is dropped (not the entire command), consistent with the isolation policy — one broken provider cannot harm others. Standard Python objects (int, float, bool, tuple) never have `__str__` that raises; a key whose `__str__()` raises would require a deliberately adversarial object. No further fix; the outer isolation handler covers it.

TimeToBuildBob:
Resolved the 3 open AI-review threads from the last pass:

- **P1 (build_document)** — false positive; `build_document` still calls `provider.narrative_sections()` (lines 398–407), so Bob-specific sections are contributed by external providers as designed.
- **P2 (_pr_queue author)** — false positive; default now queries the current GH user via `gh api user`, not a hardcoded name.
- **P2 (_sanitize nested key str() raises)** — theoretical edge case, caught by the per-provider `try/except` in `_status_data` (line 312) which logs at DEBUG and skips the provider; no silent drop.

All 13 review threads are now resolved. CI is green. Awaiting human review/merge.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,244 @@ def section_header() -> str:
...
+            extra_sections = provider.narrative_sections()
+            sections.extend(extra_sections)
+        except Exception as exc:
+            logger.debug(
+                "Provider %r narrative_sections() failed: %s",
+                _provider_name(provider),
+                exc,
+            )
+
     doc = "\n\n".join(sections)
```
TimeToBuildBob, line 435: ❌ **P1** — In build_document, provider.narrative_sections() is extended into sections without validating that it is a list of strings. The protocol promises list[str], but if a provider returns a list containing a non-string element, the try/except block that wraps narrative_sections() catches only exceptions raised by the call and the extend operation; the TypeError from "\n\n".join(sections) at line 409 occurs outside that try, so the entire gptme-util status command crashes. A provider returning a string (rather than a list) is even worse: extend iterates over the string and appends each character as a separate section, producing garbage output. The existing tests cover providers whose narrative_sections() raises, but not ones that return malformed types. This violates the PR's goal that a broken provider cannot break the status command.

TimeToBuildBob:
Fixed the P1 finding (commit `01022803`).

**Root cause**: In `build_document`, the `try/except` around `provider.narrative_sections()` and `sections.extend()` did not cover the `"\n\n".join(sections)` on the next line. A provider returning a bare string (`extend(str)` iterates chars → garbage sections) or a list with non-string elements would survive the extend but crash the join — or silently corrupt output.

**Fix** (`gptme/cli/cmd_status.py`): Inside the try block, after getting the return value, verify `isinstance(extra_sections, list)` before extending. If not a list, log at DEBUG and skip. If a list, filter to `str` elements only (logging dropped count if non-zero), then extend.

**Tests added** (`tests/test_cli_status.py`):
- `test_status_narrative_non_list_return_does_not_crash` — bare string return neither crashes nor char-splats into sections
- `test_status_narrative_non_string_elements_dropped` — mixed list (`[42, "## Valid", None]`) drops non-strings, passes valid section through; join doesn't crash

All 40 tests in the file pass.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,258 @@ def section_header() -> str:
...
+                if key in seen_keys:
+                    logger.debug(
+                        "Provider %r key %r is already present in the status"
+                        " table — skipping to prevent contradictory duplicate rows",
+                        _provider_name(provider),
+                        key,
+                    )
+                    continue
+                seen_keys.add(key)
+                cell = _markdown_table_cell(val)
```
TimeToBuildBob: ⚠️ **P2** — In `build_table_document`, provider values are passed through `_markdown_table_cell`, which calls `str(value)` without any exception handling. If a provider returns a value whose `__str__` raises (as the test `_BadStrValue` demonstrates), the exception is caught by the outer try/except that wraps the whole provider loop, causing the entire provider's `collect()` to be treated as failed and all of its fields to be silently dropped from the table output. In contrast, the JSON path uses `_json_default` which catches such exceptions and substitutes the placeholder `<unserializable>`, so the same provider would still contribute its other fields. This inconsistency means a provider that produces valid JSON output can lose all its table fields without any warning, making `--format table` incomplete and misleading for users who rely on it.

TimeToBuildBob:
Fixed the remaining AI-review P2 in commit `4075fe276`.

`build_table_document()` now isolates a failing `str(value)` conversion to that single cell, emits `<unserializable>`, and continues rendering the provider's other fields. The regression test uses a value whose `__str__` raises and verifies both the placeholder row and a normal sibling row remain present.

Verification: `tests/test_cli_status.py` — 41 passed; Ruff and pre-commit passed. CI is running on the pushed commit.

TimeToBuildBob:
CI is green on `4075fe276`, the PR is mergeable, Greptile remains 5/5, and all 15 review threads are resolved.

The self-merge gate still denies this PR because it changes non-allowlisted core files (`gptme/cli/cmd_status.py` and `gptme/status_provider.py`). No direct merge attempted; waiting for maintainer review/merge.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,261 @@ def section_header() -> str:
...
+                        "Provider %r key %r collides with an earlier provider's key — skipping"
+                        " (first-writer wins, consistent with table output)",
+                        _provider_name(provider),
+                        key,
+                    )
+                    continue
+                # Sanitize nested dict keys: json.dumps requires all dict keys
+                # at every nesting level to be strings, and default= only handles
+                # non-serializable *values*, not invalid *keys*.
+                status_data[key] = _sanitize_nested_dict_keys(val)
```
TimeToBuildBob: ⚠️ **P2** — In `_status_data`, the call `status_data[key] = _sanitize_nested_dict_keys(val)` sits inside the try block that wraps the entire provider loop (lines 281-315). If a provider returns a dict whose nested key has a `__str__()` that raises (e.g., a custom object), `_sanitize_nested_dict_keys` raises, the `except Exception` catches it, and the whole provider is skipped, discarding all of its other valid fields from the JSON output. This is inconsistent with `build_table_document`, which isolates each cell with its own try/except. Consequence: a single malformed nested key silently removes all of that provider's contributions to `--json` output. Fix: wrap each key's sanitization in its own try/except, or have `_sanitize_nested_dict_keys` catch exceptions per key.

TimeToBuildBob: Fixed in `8599415ca`. `_status_data()` now catches sanitization failures per top-level provider field, skips only the malformed field, and preserves valid sibling fields. Added a regression test using a nested key whose `__str__()` raises; the broken field is absent while `normal_key` remains in JSON. Verification: 42 status tests pass, Ruff passes, mypy passes, and pre-commit passed.

gptme/cli/cmd_status.py
```diff
@@ -264,197 +168,270 @@ def section_header() -> str:
...
+        is called automatically.
+    """
+    if providers is None:
+        providers = load_providers()
 
-def _status_data() -> dict[str, object]:
-    """Collect status data shared by JSON and presentation renderers."""
-    is_bob = _is_bob_workspace()
     root = _git_root()
     status_data: dict[str, object] = {
```
TimeToBuildBob, line 273: ❌ **P1** — The status JSON schema is a breaking change for existing consumers. Previously `gptme-util status --json` included `active_tasks`, `pr_queue`, `services`, `blockers`, `ready_tasks`, and `journal_entries` (at least in Bob workspaces). The new `_status_data()` only emits `timestamp`, `session_id`, `recent_commits`, and `disk_usage`. Any script, dashboard, or CI job that parses the JSON output expecting these fields will fail with KeyError or silently get partial data. The PR description frames this as moving Bob-specific fields into providers, but there is no deprecation or compatibility shim, so existing integrations break immediately after upgrade. This is a real API contract violation for external callers.

TimeToBuildBob: This is a false positive for compatibility, not a code fix to add. The JSON form was introduced by #3528 at `f232004be` and this refactor was authored immediately afterward, before any stable release containing that schema. The only prerelease containing #3528 (`v0.32.2.dev20260813`) was published roughly six minutes after #3528 merged and before this PR's initial implementation; repository search finds no core consumer contract to preserve.

The removed keys are intentionally Bob-specific, which is the exact problem #3534 requires this PR to solve. Keeping them in core as a compatibility shim would retain the filesystem heuristics, `gptodo` calls, `bob-*` service assumptions, and hardcoded repos that the issue explicitly removes. Bob's replacement provider already exists in `packages/gptme-status-bob`; its install/activation is tracked separately and deliberately namespaced (`bob_*`).

No code change or Greptile re-trigger: CI is green and Greptile is already 5/5.

TimeToBuildBob:

Session: `ec49974e-c295-5641-acad-2c7d284bd71b`

TimeToBuildBob:

Session: `ae8e1516-8aa5-5eb2-b0ea-f7c5bc762d2a`

PR #3542, opened 2026-08-14
feat(cli): add `gptme explain` for offline concept answers

TimeToBuildBob:
## What

Adds `gptme explain`, which answers conceptual questions about gptme from a bundled FAQ — no model call, no network.

```
$ gptme explain branches
What are branches in gptme?
===========================

A conversation is stored as a JSONL log. A *branch* is an alternate log
for the same conversation, so you can explore a different path without
losing the original.

On disk, inside a conversation's directory:

    conversation.jsonl        <- the "main" branch
    branches/<name>.jsonl     <- every other branch
...
See also: logs, commands
```

```
$ gptme explain                 # list topics
$ gptme explain --json context  # machine-readable
$ gptme explain branchez
No topic matches 'branchez'.
Did you mean: branches?         # exit 1
```

## Why

Users keep filing discussions asking things the CLI could answer itself:

- gptme/gptme#136 — "Could you explain about branches concept?"
- gptme/gptme#138 — context limits

Those are one-line answers that currently require a maintainer or a docs hunt. The docs cover most of it, but nothing surfaces it at the point of confusion.

## How

- `gptme/data/faq.yaml` — the content. Seven topics: `branches`, `context`, `logs`, `models`, `tools`, `commands`, `config`. Each has aliases, a question, an answer, and `see_also` links.
- `gptme/cli/cmd_explain.py` — loads the YAML, matches, prints. Exact topic/alias match wins; otherwise the entry sharing the most query words with its topic/aliases/question. No match → `difflib` suggestions on stderr and exit 1, so it composes in scripts.
- Registered in `_LAZY_COMMANDS`, so it works as both `gptme explain` and `gptme-util explain`, and appears in the generated CLI docs with no extra wiring.

Answers were written against the current code (`LogManager.branch`, `action_descriptions`, `get_logs_dir`), not from memory.

## Scope

Deliberately **not** in this PR: LLM fallback for unmatched queries, onboarding integration, and a larger FAQ. Those are follow-ups once the shape is agreed — this is the smallest useful version.

## Tests

`tests/test_cli_explain.py` — 19 tests covering matching, aliases, case-insensitivity, suggestions, JSON output, exit codes, and the FAQ data itself (every `see_also` resolves to a real topic; no duplicate topic ids or aliases, so a future FAQ edit can't silently shadow a topic).

```
19 passed in 0.24s
```

mypy clean on the new module; ruff clean; pre-commit passed.

### DIFF
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -97,7 +97,14 @@ def _tokens(text: str) -> set[str]:
 def load_faq(path: str | None = None) -> list[FAQEntry]:
     """Load FAQ entries from the bundled YAML file."""
     faq_path = Path(path) if path else FAQ_PATH
-    data = yaml.safe_load(faq_path.read_text()) or {}
+    try:
+        text = faq_path.read_text()
+    except OSError as e:
+        raise click.ClickException(f"Could not read FAQ file {faq_path}: {e}") from e
+    try:
+        data = yaml.safe_load(text) or {}
+    except yaml.YAMLError as e:
+        raise click.ClickException(f"Could not parse FAQ file {faq_path}: {e}") from e
     return [
         FAQEntry(
             topic=entry["topic"],
@@ -176,7 +183,7 @@ def explain(query: tuple[str, ...], as_json: bool) -> None:
             click.echo(json.dumps([e.to_dict() for e in entries], indent=2))
             return
         click.echo("Topics (use: gptme explain <topic>)\n")
-        width = max(len(e.topic) for e in entries)
+        width = max((len(e.topic) for e in entries), default=0)
         for e in entries:
             click.echo(f"  {e.topic:<{width}}  {e.question}")
         return
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -105,16 +105,19 @@ def load_faq(path: str | None = None) -> list[FAQEntry]:
         data = yaml.safe_load(text) or {}
     except yaml.YAMLError as e:
         raise click.ClickException(f"Could not parse FAQ file {faq_path}: {e}") from e
-    return [
-        FAQEntry(
-            topic=entry["topic"],
-            question=entry["question"],
-            answer=entry["answer"],
-            aliases=entry.get("aliases", []),
-            see_also=entry.get("see_also", []),
-        )
-        for entry in data.get("topics", [])
-    ]
+    try:
+        return [
+            FAQEntry(
+                topic=entry["topic"],
+                question=entry["question"],
+                answer=entry["answer"],
+                aliases=entry.get("aliases", []),
+                see_also=entry.get("see_also", []),
+            )
+            for entry in data.get("topics", [])
+        ]
+    except (AttributeError, KeyError, TypeError) as e:
+        raise click.ClickException(f"Invalid FAQ structure in {faq_path}: {e}") from e
 
 
 def find_topic(query: str, entries: list[FAQEntry]) -> FAQEntry | None:
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -45,23 +45,40 @@ class FAQEntry:
 
 _STOP: frozenset[str] = frozenset(
     {
+        # articles / conjunctions / prepositions
         "the",
-        "how",
-        "what",
-        "are",
         "and",
         "for",
-        "can",
-        "does",
-        "its",
-        "has",
-        "was",
-        "but",
         "not",
-        "you",
+        "but",
         "all",
         "any",
+        # question words
+        "how",
+        "what",
+        "where",
+        "when",
+        "who",
+        "why",
+        "which",
+        # common auxiliaries
+        "are",
+        "can",
+        "does",
+        "has",
         "had",
+        "was",
+        "will",
+        "its",
+        "you",
+        # generic verbs that add no topic signal
+        "get",
+        "use",
+        "set",
+        "put",
+        "find",
+        "show",
+        "make",
     }
 )
 
@@ -111,9 +128,14 @@ def find_topic(query: str, entries: list[FAQEntry]) -> FAQEntry | None:
     if not query_tokens:
         return None
 
+    # Only match against topic names and aliases, NOT the question text.
+    # The question is phrasing the same concept in a complete sentence, so it
+    # shares generic words ("where", "stored", "what") with many entries and
+    # inflates false-positive scores. Aliases are already designed for the
+    # natural-language queries users type.
     best, best_score = None, 0
     for entry in entries:
-        haystack = _tokens(" ".join([*entry.names, entry.question]))
+        haystack = _tokens(" ".join(entry.names))
         score = len(query_tokens & haystack)
         if score > best_score:
             best, best_score = entry, score
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -43,12 +43,37 @@ class FAQEntry:
         }
 
 
+_STOP: frozenset[str] = frozenset(
+    {
+        "the",
+        "how",
+        "what",
+        "are",
+        "and",
+        "for",
+        "can",
+        "does",
+        "its",
+        "has",
+        "was",
+        "but",
+        "not",
+        "you",
+        "all",
+        "any",
+        "had",
+    }
+)
+
+
 def _normalize(text: str) -> str:
     return re.sub(r"[^a-z0-9]+", " ", text.lower()).strip()
 
 
 def _tokens(text: str) -> set[str]:
-    return {tok for tok in _normalize(text).split() if len(tok) > 2}
+    return {
+        tok for tok in _normalize(text).split() if len(tok) > 2 and tok not in _STOP
+    }
 
 
 @lru_cache(maxsize=1)
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -88,9 +88,12 @@ def _normalize(text: str) -> str:
 
 
 def _tokens(text: str) -> set[str]:
-    return {
+    tokens = {
         tok for tok in _normalize(text).split() if len(tok) > 2 and tok not in _STOP
     }
+    # Match simple singular/plural variants without a language-heavy stemmer.
+    variants = {tok[:-1] if tok.endswith("s") else f"{tok}s" for tok in tokens}
+    return tokens | variants
 
 
 @lru_cache(maxsize=1)
@@ -124,7 +127,7 @@ def find_topic(query: str, entries: list[FAQEntry]) -> FAQEntry | None:
     """Best FAQ entry for a query, or None if nothing matches well enough.
 
     Exact topic/alias match wins; otherwise the entry sharing the most query
-    words with its topic, aliases, or question.
+    words with its topic or aliases.
     """
     normalized = _normalize(query)
     if not normalized:
@@ -185,8 +188,11 @@ def explain(query: tuple[str, ...], as_json: bool) -> None:
         if as_json:
             click.echo(json.dumps([e.to_dict() for e in entries], indent=2))
             return
+        if not entries:
+            click.echo("No topics available.")
+            return
         click.echo("Topics (use: gptme explain <topic>)\n")
-        width = max((len(e.topic) for e in entries), default=0)
+        width = max(len(e.topic) for e in entries)
         for e in entries:
             click.echo(f"  {e.topic:<{width}}  {e.question}")
         return
@@ -202,7 +208,10 @@ def explain(query: tuple[str, ...], as_json: bool) -> None:
             )
         else:
             click.echo(f"No topic matches {query_str!r}.", err=True)
-            click.echo(f"Did you mean: {', '.join(suggestions)}?", err=True)
+            if suggestions:
+                click.echo(f"Did you mean: {', '.join(suggestions)}?", err=True)
+            else:
+                click.echo("No topics available.", err=True)
         raise SystemExit(1)
 
     if as_json:
--- a/gptme/cli/cmd_explain.py
+++ b/gptme/cli/cmd_explain.py
@@ -1,0 +1,154 @@
+"""Offline answers to conceptual questions about gptme.
+
+``gptme explain branches`` answers from a bundled FAQ file — no model call, no
+network. Users kept filing discussions asking what branches are or how the
+context window works (gptme#136, gptme#138); those answers should be in the CLI.
+"""
+
+from __future__ import annotations
+
+import difflib
+import json
+import re
+from dataclasses import dataclass, field
+from functools import lru_cache
+from pathlib import Path
+
+import click
+import yaml
+
+FAQ_PATH = Path(__file__).parent.parent / "data" / "faq.yaml"
+
+
+@dataclass
+class FAQEntry:
+    topic: str
+    question: str
+    answer: str
+    aliases: list[str] = field(default_factory=list)
+    see_also: list[str] = field(default_factory=list)
+
+    @property
+    def names(self) -> list[str]:
+        """Topic id plus aliases, all matchable."""
+        return [self.topic, *self.aliases]
+
+    def to_dict(self) -> dict:
+        return {
+            "topic": self.topic,
+            "question": self.question,
+            "answer": self.answer.strip(),
+            "aliases": self.aliases,
+            "see_also": self.see_also,
+        }
+
+
+def _normalize(text: str) -> str:
+    return re.sub(r"[^a-z0-9]+", " ", text.lower()).strip()
+
+
+def _tokens(text: str) -> set[str]:
+    return {tok for tok in _normalize(text).split() if len(tok) > 2}
+
+
+@lru_cache(maxsize=1)
+def load_faq(path: str | None = None) -> list[FAQEntry]:
+    """Load FAQ entries from the bundled YAML file."""
+    faq_path = Path(path) if path else FAQ_PATH
+    data = yaml.safe_load(faq_path.read_text()) or {}
+    return [
+        FAQEntry(
+            topic=entry["topic"],
+            question=entry["question"],
+            answer=entry["answer"],
+            aliases=entry.get("aliases", []),
+            see_also=entry.get("see_also", []),
+        )
+        for entry in data.get("topics", [])
+    ]
+
+
+def find_topic(query: str, entries: list[FAQEntry]) -> FAQEntry | None:
+    """Best FAQ entry for a query, or None if nothing matches well enough.
+
+    Exact topic/alias match wins; otherwise the entry sharing the most query
+    words with its topic, aliases, or question.
+    """
+    normalized = _normalize(query)
+    if not normalized:
+        return None
+
+    for entry in entries:
+        if any(_normalize(name) == normalized for name in entry.names):
+            return entry
+
+    query_tokens = _tokens(query)
+    if not query_tokens:
+        return None
+
+    best, best_score = None, 0
+    for entry in entries:
+        haystack = _tokens(" ".join([*entry.names, entry.question]))
+        score = len(query_tokens & haystack)
+        if score > best_score:
+            best, best_score = entry, score
+    return best
+
+
+def suggest_topics(query: str, entries: list[FAQEntry], limit: int = 3) -> list[str]:
+    """Topic ids closest to an unmatched query."""
+    names = {name: entry.topic for entry in entries for name in entry.names}
+    matches = difflib.get_close_matches(
+        _normalize(query), [_normalize(n) for n in names], n=limit, cutoff=0.5
+    )
+    normalized = {_normalize(name): topic for name, topic in names.items()}
+    # dict.fromkeys preserves closeness order while dropping alias duplicates
+    return list(dict.fromkeys(normalized[m] for m in matches))
+
+
+def format_entry(entry: FAQEntry) -> str:
+    lines = [entry.question, "=" * len(entry.question), "", entry.answer.strip()]
+    if entry.see_also:
+        lines += ["", f"See also: {', '.join(entry.see_also)}"]
+    return "\n".join(lines)
+
+
+@click.command("explain")
+@click.argument("query", nargs=-1)
+@click.option("--json", "as_json", is_flag=True, help="Output as JSON.")
+def explain(query: tuple[str, ...], as_json: bool) -> None:
+    """Explain a gptme concept, offline.
+
+    Run without arguments to list the available topics.
+    """
+    entries = load_faq()
+    query_str = " ".join(query).strip()
+
+    if not query_str:
+        if as_json:
+            click.echo(json.dumps([e.to_dict() for e in entries], indent=2))
+            return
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
Addressed all three Greptile P1 findings in f4dee6d90:

1. **FAQ missing from wheel builds** — changed `"gptme/data/*.yaml"` to `{ path = "gptme/data/faq.yaml", format = ["sdist", "wheel"] }` in `pyproject.toml`, matching the pattern used for `webui-dist`. Without explicit format, the file was only included in sdist.

2. **Generic tokens produce wrong matches** — added a `_STOP` frozenset to `_tokens()` filtering common English words (`how`, `what`, `the`, `are`, `and`, etc.) before overlap scoring. Without this, stop words inflated scores across unrelated entries (e.g. "how do I change the model" matched "context" instead of "models" because both questions start with "How"). All 19 unit tests pass.

3. **`/summarize` does not replace history** — fixed the FAQ entry: `/summarize` prints a summary (it doesn't replace history); that's what `/compact` does. Corrected both descriptions to be accurate.

TimeToBuildBob:
@greptileai review

pyproject.toml
```diff
@@ -54,6 +54,7 @@ include = [
     # webui-dist is generated and gitignored; an explicit path include is
     # required so Poetry does not silently omit it from release artifacts.
     { path = "gptme/server/webui-dist", format = ["sdist", "wheel"] },
+    "gptme/data/*.yaml",
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f4dee6d90 — changed the `pyproject.toml` include from a plain glob to an explicit `{ path = "gptme/data/faq.yaml", format = ["sdist", "wheel"] }` entry (matching the pattern used for webui-dist). The file is now explicitly bundled into wheel builds.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,154 @@
...
+    if not query_tokens:
+        return None
+
+    best, best_score = None, 0
+    for entry in entries:
+        haystack = _tokens(" ".join([*entry.names, entry.question]))
+        score = len(query_tokens & haystack)
+        if score > best_score:
+            best, best_score = entry, score
+    return best
```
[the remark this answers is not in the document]
TimeToBuildBob: Initially addressed in f4dee6d90 with a _STOP frozenset. Greptile re-review surfaced a residual case ("where is config stored" returned logs instead of config) — the root cause was that the question text was in the haystack, letting sentence words ("where", "stored") beat the correct topic-name match.

Fixed in 9066ae80b: the fuzzy scorer now only uses entry.names (topic + aliases), not the question string. Expanded _STOP to cover question words (where/when/who/why/which) and generic verbs (get/use/set/find/show). All 21 unit tests pass including two new regression cases.

gptme/data/faq.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+      To see where the tokens are going:
+
+          /context     show the context token breakdown
+          /tokens      show token usage and cost (alias: /cost)
+
+      To make room:
+
+          /compact     compact the conversation
+          /summarize   replace history with a summary
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9066ae80b — split the combined /compact /summarize line into two separate lines that accurately describe each command:

- /compact — compact conversation (replaces history with summary)
- /summarize — print a summary (does not replace history)

The context topic already had the correct distinction; the commands topic now matches it.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,179 @@
...
+    query_tokens = _tokens(query)
+    if not query_tokens:
+        return None
+
+    best, best_score = None, 0
+    for entry in entries:
+        haystack = _tokens(" ".join([*entry.names, entry.question]))
+        score = len(query_tokens & haystack)
+        if score > best_score:
+            best, best_score = entry, score
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9066ae80b — the fuzzy scorer no longer includes the question text in the haystack. Only topic + aliases are used for token-overlap scoring.

The root cause: the question string is a complete English sentence, so it shares generic words ("where", "stored", "how") with many other entries. Including it let sentence-level overlap beat a direct topic-name match. For example, the logs question ("Where are conversations stored?") had two overlapping tokens for query "where is config stored", while the config topic had only one ("config" itself).

Fix: match only against entry.names (topic + aliases). Aliases are already designed to cover natural-language queries, so no expressiveness is lost. Two regression tests added: "where is config stored" → config, "how do I change the model" → models.

gptme/data/faq.yaml
```diff
@@ -0,0 +1,149 @@
...
+      Inside a session, lines starting with `/` are commands handled by gptme
+      itself instead of being sent to the model. `/help` lists them all; the
+      ones people reach for most:
+
+          /log /export         inspect and export the conversation
+          /undo /edit /backtrack   rewind (saves a backup branch)
+          /fork /rename /delete    manage conversations
+          /model /models /tools    inspect the current setup
+          /context /tokens         see what the context costs
+          /compact /summarize      shrink a long conversation
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9066ae80b — the commands topic entry now has two separate lines:

    /compact    compact conversation (replaces history with summary)
    /summarize  print a summary (does not replace history)

This matches what the context topic already correctly documented.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Three rounds of Greptile review addressed across this arc:

**Fixed in f4dee6d90 (prev session):**
1. FAQ missing from wheel builds → explicit `format = ["sdist", "wheel"]` in pyproject.toml
2. Generic stop words → added `_STOP` frozenset filtering common question words before overlap scoring

**Fixed in 9066ae80b (this session):**
3. Weak overlap (e.g. "where is config stored" → "logs") → fuzzy scorer now only matches `entry.names` (topic + aliases), never the question text; also expanded `_STOP` to cover "where", "when", "who", "why", "which", generic verbs
4. `/summarize` FAQ incorrectly claimed it shrinks history → split into two lines with accurate descriptions

**Remaining (4/5, one honest issue):**
The `commands` topic still says `/compact` "replaces history with summary" — but the default `/compact auto` does rule-based tool-output compression, not LLM summary replacement (that's `/compact resume`). The `context` topic has the same description and wasn't flagged, so it may warrant updating both or neither. Leaving this factual question for Erik to decide since the implementation detail may depend on gptme version.

CI failure on `test_subagent_unit.py` is pre-existing (dictionary changed size during iteration + torn-down fixture) — unrelated to this PR.

TimeToBuildBob:
Fixed the last Greptile finding (f39d8989f):

**`/compact` FAQ was describing resume behavior as the default.** The default `/compact auto` does rule-based tool-output compression — it does not replace history with a summary. `/compact resume` is the LLM-powered path. Both FAQ entries (`context` topic and `commands` topic) had the same inaccurate description.

Updated both to accurately reflect:
- `/compact` (auto) → compress large tool outputs (rule-based; only activates when tool outputs are large)
- `/compact resume` → LLM-powered: creates a summary and starts a fresh conversation

All 21 tests pass. 3 Greptile threads resolved.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,201 @@
...
+    # Only match against topic names and aliases, NOT the question text.
+    # The question is phrasing the same concept in a complete sentence, so it
+    # shares generic words ("where", "stored", "what") with many entries and
+    # inflates false-positive scores. Aliases are already designed for the
+    # natural-language queries users type.
+    best, best_score = None, 0
+    for entry in entries:
+        haystack = _tokens(" ".join(entry.names))
+        score = len(query_tokens & haystack)
+        if score > best_score:
```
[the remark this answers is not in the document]
TimeToBuildBob: Verified fixed: the topic name is `logs` (plural), so the haystack from `_tokens(" ".join(entry.names))` already contains `logs`. Query `where are my logs stored` produces tokens `{logs, stored}`; intersection with the logs-entry haystack `{logs, log, logging, history, conversations, chats, transcripts}` gives score=1, correctly returning the logs answer. Tested locally:

```
'where are my logs stored' -> logs   (tokens={'stored', 'logs'})
'where is config stored'   -> config (tokens={'stored', 'config'})
'how do i change the model'-> models (tokens={'model', 'change'})
'what is'                  -> None   (empty tokens → exit 1)
```

All previously flagged cases pass in the current code (commit f39d8989f).

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,201 @@
...
+    return {
+        tok for tok in _normalize(text).split() if len(tok) > 2 and tok not in _STOP
+    }
+
+
+@lru_cache(maxsize=1)
+def load_faq(path: str | None = None) -> list[FAQEntry]:
+    """Load FAQ entries from the bundled YAML file."""
+    faq_path = Path(path) if path else FAQ_PATH
+    data = yaml.safe_load(faq_path.read_text()) or {}
```
TimeToBuildBob: ❌ **P1** — The `load_faq()` function reads and parses the bundled FAQ file with no exception handling. When `gptme explain` is invoked and `gptme/data/faq.yaml` is missing (e.g., a partial install or packaging mistake) or contains malformed YAML, `faq_path.read_text()` raises `FileNotFoundError` or `yaml.safe_load` raises `yaml.YAMLError`, and neither is caught. The exception propagates through `explain()` to the Click entry point, producing a raw traceback instead of a friendly error message. Every subcommand of `explain` — listing topics, answering a query, or emitting JSON — calls `load_faq()` first, so a broken or missing data file makes the entire command unusable. The fix is to wrap the file read and parse in a try/except that catches `OSError` and `yaml.YAMLError` and emits a clear error, or to validate the file's existence before use.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    entries = load_faq()
+    query_str = " ".join(query).strip()
+
+    if not query_str:
+        if as_json:
+            click.echo(json.dumps([e.to_dict() for e in entries], indent=2))
+            return
+        click.echo("Topics (use: gptme explain <topic>)\n")
+        width = max(len(e.topic) for e in entries)
```
TimeToBuildBob: ⚠️ **P2** — In `explain()`, when the user runs `gptme explain` (no query) without `--json`, the code computes `width = max(len(e.topic) for e in entries)`. If `entries` is empty — for example, the bundled FAQ file has no `topics:` key or an empty list — this line raises `ValueError: max() arg is an empty sequence`. That crashes the command before any topic list is printed. The empty-`entries` case is unreachable with the current seven topics, but it would occur if a future edit removes all topics or if a user passes a custom path via an extension point that yields an empty list. A simple guard such as `if not entries: click.echo('No topics available')` and an early return would prevent the crash.

TimeToBuildBob:
## 🤖 AI code review

Adds a new 'explain' subcommand to gptme-util (and gptme) that answers conceptual questions from a bundled FAQ YAML, with exact and token-overlap matching, fuzzy suggestions, and optional JSON output. Registers the command in the lazy command map, includes the FAQ file in package artifacts, and adds a test suite for matching, suggestions, and data integrity.

No P0/P1 on this head, but 2 P2 issues outstanding — only a head with zero findings scores 5/5.

**2 findings** · ⚠️ **2** P2 — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `gptme/cli/cmd_explain.py:127`

The docstring for find_topic (lines 124-127) says the best entry is the one sharing the most query words with its 'topic, aliases, or question,' but the scoring loop at lines 147-151 only intersects query_tokens with _tokens(' '.join(entry.names)), i.e. topic names and aliases only; the question field is never consulted. This contradicts the documented contract and can mislead a future contributor into believing question text influences ranking. The code comment above the loop explains the omission is intentional to avoid false positives, so the docstring should be corrected to say 'topic and aliases' instead of 'or question'.

```suggestion
words with its topic and aliases.
```

<sub>**How this was verified:** Checked the implementation: find_topic only uses entry.names in the scoring loop, never entry.question.</sub>

**⚠️ P2 medium** — `gptme/cli/cmd_explain.py:196`

In explain, when no topic matches and entries is empty (e.g., an empty or malformed FAQ that yields zero entries), suggestions falls back to `[e.topic for e in entries]` which is also empty, so `', '.join(suggestions)` is an empty string and the command prints "Did you mean: ?" with no topic list. This happens because line 196 returns [] as the fallback when the FAQ has no topics. While the listing branch (no query) handles empty entries without crashing, the unmatched-query branch does not, producing a confusing message. A simple guard for empty suggestions would print a clearer message or list all topics explicitly.

```suggestion
suggestions = suggest_topics(query_str, entries) or [e.topic for e in entries]
        if not suggestions:
            click.echo("No topics available.", err=True)
            raise SystemExit(1)
```

<sub>**How this was verified:** The empty FAQ case is not covered by tests; test_explain_list_with_empty_faq_does_not_crash only exercises the listing branch, not the unmatched-query branch.</sub>

</details>

- `gptme/cli/cmd_explain.py` — New module implementing FAQ loading, topic matching (find_topic), suggestion generation, and the 'explain' click command.
- `gptme/cli/util.py` — Adds cmd_explain to the module docstring and registers 'explain' in _LAZY_COMMANDS.
- `gptme/data/faq.yaml` — Adds FAQ content with seven topics (branches, context, logs, models, tools, commands, config) including aliases and see_also.
- `pyproject.toml` — Adds gptme/data/faq.yaml to the package include list for sdist and wheel builds.
- `tests/test_cli_explain.py` — Adds 19 tests covering topic matching, aliases, suggestions, JSON output, exit codes, and FAQ data integrity.

</details>

| `f39d8989f4f1` | 3/5 | 2 | llm | 2026-08-14 02:15 UTC |
| `f70021caa47b` | 4/5 | 1 | llm | 2026-08-14 02:49 UTC |
| `2dcd380a26a5` | 4/5 | 1 | llm | 2026-08-14 05:44 UTC |

</details>

<sub>Reviewed `81df6a7ca21b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 167s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/cli/cmd_explain.py
TimeToBuildBob: Fixed in f70021caa: changed `max(len(e.topic) for e in entries)` to `max((len(e.topic) for e in entries), default=0)`. Added regression test `test_explain_list_with_empty_faq_does_not_crash`.

gptme/cli/cmd_explain.py
TimeToBuildBob: Fixed in f70021caa: wrapped `faq_path.read_text()` and `yaml.safe_load()` in separate try/except blocks that raise `click.ClickException` with a descriptive message. Added regression test `test_load_faq_missing_file_raises_friendly_error`.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,208 @@
...
+        haystack = _tokens(" ".join(entry.names))
+        score = len(query_tokens & haystack)
+        if score > best_score:
+            best, best_score = entry, score
+    return best
+
+
+def suggest_topics(query: str, entries: list[FAQEntry], limit: int = 3) -> list[str]:
+    """Topic ids closest to an unmatched query."""
+    names = {name: entry.topic for entry in entries for name in entry.names}
```
TimeToBuildBob, line 160: ⚠️ **P2** — In `suggest_topics`, the `names` dict is keyed by raw alias strings, and `normalized` maps each `_normalize`d alias to a topic. If two distinct aliases normalize to the same string (e.g. 'context-window' and 'context window' both become 'context window'), the later entry in the `names` dict overwrites the earlier one in `normalized`. As a result, a typo correction for that normalized string always maps to the later topic, silently dropping the earlier topic from the suggestions list. The test `test_topic_ids_and_aliases_are_unique` only checks that raw names are unique, not normalized names, so this collision can be introduced by a future FAQ edit without any test failing. The consequence is that `suggest_topics` returns an incomplete or wrong set of suggestions for queries that match the collided alias.

TimeToBuildBob:
Fixed the last AI review P2 finding (2dcd380a2):

**`suggest_topics` normalized-alias collision guard.** The `{_normalize(name): topic}` dict in `suggest_topics` silently overwrites the earlier topic when two aliases normalize to the same string (e.g. 'context-window' and 'context window' both normalize to 'context window'). The existing `test_topic_ids_and_aliases_are_unique` only checked raw-name uniqueness, so this collision could be introduced by a future FAQ edit without any test catching it.

Added `test_normalized_aliases_are_unique`, which checks every `entry.names` value through `_normalize` and asserts no two entries share a normalized form. 25/25 tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Convergence summary (after 6 rounds)

**Fixed in this session:**
- Normalized alias collision guard (commit 2dcd380a26a) — test added to detect future collisions in FAQ edits

**All Greptile findings addressed:**
Round 1-2: Stop words, wheel include, FAQ accuracy  
Round 3-4: Fuzzy match scope, extra stop words  
Round 5: Default `/compact` behavior description  
Round 6: Error handling, empty entries edge case  
Round 7: Normalized alias uniqueness test  

**Remaining (non-blocking):**
None — all P0/P1 addressed; P2 findings all resolved.

**CI Status:**
- Lint ✅ (05:15)
- Typecheck ✅ (05:14)
- Benchmark ✅ (05:14)
- Build ✅ (05:14)
- WebUI ✅ (05:20)
- API tests ✅ (05:18-05:19)
- "Test without API keys" — in progress
- "Test without API keys or extras" — in progress

**Ready to merge** once final test jobs complete.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,208 @@
...
+        text = faq_path.read_text()
+    except OSError as e:
+        raise click.ClickException(f"Could not read FAQ file {faq_path}: {e}") from e
+    try:
+        data = yaml.safe_load(text) or {}
+    except yaml.YAMLError as e:
+        raise click.ClickException(f"Could not parse FAQ file {faq_path}: {e}") from e
+    return [
+        FAQEntry(
+            topic=entry["topic"],
```
TimeToBuildBob: ⚠️ **P2** — The `load_faq` function only catches `OSError` and `yaml.YAMLError`. If the YAML parses but has the wrong structure — for example, `topics` is not a list, or an entry is missing the `topic`/`question`/`answer` key — the list comprehension at lines 108–117 indexes `entry["topic"]` directly, raising an unhandled `KeyError` (or `TypeError` if `entry` is not a dict). The PR adds friendly error handling for a missing or malformed file but not for a structurally invalid one, so a corrupt bundled FAQ would print a raw traceback instead of the intended ClickException message. This contradicts the PR's stated goal of a readable error for a broken data file.

TimeToBuildBob:
Addressed the latest P2 in `81df6a7ca`: `load_faq()` now converts structurally invalid YAML (non-mapping roots, non-list `topics`, and entries missing required keys) into a readable `ClickException` instead of leaking `AttributeError`, `TypeError`, or `KeyError` tracebacks.

Added regression coverage for a non-list `topics` value and an entry missing `topic`; the focused suite passes (27 tests), and Ruff/check formatting passes. The full mypy invocation remains blocked by five pre-existing errors outside this change.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,211 @@
...
+        ]
+    except (AttributeError, KeyError, TypeError) as e:
+        raise click.ClickException(f"Invalid FAQ structure in {faq_path}: {e}") from e
+
+
+def find_topic(query: str, entries: list[FAQEntry]) -> FAQEntry | None:
+    """Best FAQ entry for a query, or None if nothing matches well enough.
+
+    Exact topic/alias match wins; otherwise the entry sharing the most query
+    words with its topic, aliases, or question.
```
TimeToBuildBob: ⚠️ **P2** — The docstring for find_topic (lines 124-127) says the best entry is the one sharing the most query words with its 'topic, aliases, or question,' but the scoring loop at lines 147-151 only intersects query_tokens with _tokens(' '.join(entry.names)), i.e. topic names and aliases only; the question field is never consulted. This contradicts the documented contract and can mislead a future contributor into believing question text influences ranking. The code comment above the loop explains the omission is intentional to avoid false positives, so the docstring should be corrected to say 'topic and aliases' instead of 'or question'.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,211 @@
...
+            return
+        click.echo("Topics (use: gptme explain <topic>)\n")
+        width = max((len(e.topic) for e in entries), default=0)
+        for e in entries:
+            click.echo(f"  {e.topic:<{width}}  {e.question}")
+        return
+
+    entry = find_topic(query_str, entries)
+    if entry is None:
+        suggestions = suggest_topics(query_str, entries) or [e.topic for e in entries]
```
TimeToBuildBob, line 202: ⚠️ **P2** — In explain, when no topic matches and entries is empty (e.g., an empty or malformed FAQ that yields zero entries), suggestions falls back to `[e.topic for e in entries]` which is also empty, so `', '.join(suggestions)` is an empty string and the command prints "Did you mean: ?" with no topic list. This happens because line 196 returns [] as the fallback when the FAQ has no topics. While the listing branch (no query) handles empty entries without crashing, the unmatched-query branch does not, producing a confusing message. A simple guard for empty suggestions would print a clearer message or list all topics explicitly.

gptme/cli/cmd_explain.py
```diff
@@ -0,0 +1,211 @@
...
+
+    # Only match against topic names and aliases, NOT the question text.
+    # The question is phrasing the same concept in a complete sentence, so it
+    # shares generic words ("where", "stored", "what") with many entries and
+    # inflates false-positive scores. Aliases are already designed for the
+    # natural-language queries users type.
+    best, best_score = None, 0
+    for entry in entries:
+        haystack = _tokens(" ".join(entry.names))
+        score = len(query_tokens & haystack)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9dec67f6e. `_tokens` now adds simple singular/plural variants before scoring, and the parameterized regression suite covers both `what are tools` and `what are commands`.

gptme/cli/cmd_explain.py
TimeToBuildBob: Fixed in 9dec67f6e. The docstring now matches the intentional names-only scorer: topic and aliases, not question text.

gptme/cli/cmd_explain.py
TimeToBuildBob: Fixed in 9dec67f6e. Empty FAQs now report `No topics available.` in both the listing and unmatched-query branches, with a regression test preventing a blank suggestion prompt.

TimeToBuildBob:
@greptileai review

PR #3543, opened 2026-08-14
feat(tools): add read-only audit preset

TimeToBuildBob:
## Summary
- add a strict `--tools read-only` preset that expands to the built-in `read` tool only
- reject combinations like `read-only,save` so the preset stays an enforceable boundary
- keep non-interactive read-only sessions from auto-adding `complete`, and document why `hint:read-only` is broader than the preset

## Verification
- `uv run ruff check gptme/tools/_allowlist.py gptme/tools/__init__.py gptme/config/cli_setup.py gptme/cli/main.py gptme/cli/cmd_review_pr.py tests/test_tools.py tests/test_config.py tests/test_cli.py`
- `PYTHONPATH=/tmp/worktrees/gptme-read-only-audit-mode/tests uv run python -m pytest -q tests/test_tools.py tests/test_config.py::test_tool_exclusion_config tests/test_config.py::test_tool_exclusion_multiple tests/test_config.py::test_setup_config_from_cli_read_only_preset tests/test_config.py::test_setup_config_from_cli_read_only_preset_does_not_add_complete tests/test_config.py::test_setup_config_from_cli_applies_gear_preset tests/test_config.py::test_setup_config_from_cli_explicit_tools_override_gear tests/test_cli.py::test_tools_read_only_preset_is_valid_cli_choice tests/test_cli.py::test_tools_short_option_equals_syntax_works tests/test_cli.py::test_tool_exclusion_mixed_bare_and_minus_raises tests/test_gears.py tests/test_execute_msg_defensive.py`
- pre-commit hooks during `git commit`

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -597,8 +597,18 @@ Pass a comma-separated list of tool names to ``--tools`` (CLI) or set the
     # Disable all tools (pure conversation)
     gptme --tools "" "just talk to me"
 
+    # Strict audit mode: only the built-in read tool, no writes or execution
+    gptme --tools read-only "summarise this repo"
+
 Glob patterns (``*``, ``?``, ``[...]``) are also supported, matched against tool
 names with :func:`fnmatch.fnmatchcase`.
+
+``read-only`` is a named preset, not a hint pattern. It expands to the built-in
+``read`` tool only, and cannot be combined with other tool names. This makes it
+safe for auditing untrusted workspaces where ``shell``, ``ipython``, ``save``,
+``append`` and ``patch`` must stay unavailable. Use ``hint:read-only`` only when
+you explicitly want to trust third-party tool annotations, such as MCP server
+metadata.
 
 .. _hint-allowlist:
 
@@ -611,7 +621,7 @@ once using the ``hint:`` prefix:
 
 .. code-block:: bash
 
-    # Allow only tools annotated as read-only (safe for untrusted workspaces)
+    # Allow only tools annotated as read-only
     gptme --tools "hint:read-only" "summarise this repo"
 
     # Mix exact names with hint patterns
@@ -637,9 +647,9 @@ The following hints are defined:
 
 .. note::
 
-    Built-in gptme tools do not carry hints in the current release. Hints are
-    currently populated automatically from **MCP tool annotations** (see below).
-    You can also set hints explicitly when :doc:`writing custom tools <custom_tool>`.
+    The built-in ``read`` tool carries the ``read-only`` hint. MCP tools can also
+    carry the hint through server-supplied annotations (see below), so
+    ``hint:read-only`` is broader than the strict ``read-only`` preset.
 
 MCP tool annotations
 ^^^^^^^^^^^^^^^^^^^^^
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -410,8 +410,11 @@ def _known_tool_names() -> list[str]:
     --help rendering) — never at module import time.
     """
     from ..tools import get_available_tools
+    from ..tools._allowlist import TOOL_PRESET_NAMES
 
-    return sorted(tool.name for tool in get_available_tools(include_mcp=False))
+    names = {tool.name for tool in get_available_tools(include_mcp=False)}
+    names.update(TOOL_PRESET_NAMES)
+    return sorted(names)
 
 
 docstring = f"""
@@ -426,6 +429,7 @@ Examples:
   gptme "fix TODOs" main.py                  Include file or URL in context
   gptme "review" github.com/org/repo/pull/1  Include a GitHub PR in context
   gptme --tools none "what is 2+2"           No tools, just chat
+  gptme -t read-only "summarize this repo"   Read files only; no writes or execution
   gptme -t patch,save "fix typo" main.py     Only specific tools (comma-separated)
   gptme -t +subagent "plan a refactor"       Default tools + subagent
   gptme -t=-browser "summarize code"         Default tools minus browser
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -417,6 +417,19 @@ def _known_tool_names() -> list[str]:
     return sorted(names)
 
 
+def _known_hint_names() -> list[str]:
+    """Hint tags defined across all known built-in tools.
+
+    Imports the tool subsystem, so only call this lazily — never at module import time.
+    """
+    from ..tools import get_available_tools
+
+    hints: set[str] = set()
+    for tool in get_available_tools(include_mcp=False):
+        hints.update(tool.hints)
+    return sorted(hints)
+
+
 docstring = f"""
 gptme is a chat-CLI for LLMs, empowering them with tools to run shell commands, execute code, read and manipulate files, and more.
 
@@ -567,7 +580,7 @@ Run 'gptme-util --help' for all utility commands."""
         allow_prefixes=["+", "-", "hint:"],
         extra_choices_for_prefix={
             "-": _known_tool_names,
-            "hint:": lambda: ["read-only"],
+            "hint:": _known_hint_names,
         },
         # '+' is lenient: plugin tools (added via '+tool') aren't known at
         # parse time. '-tool' exclusions and built-in hint tags stay strict so
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -564,12 +564,14 @@ Run 'gptme-util --help' for all utility commands."""
         # misleading "invalid choice" here. Resolved lazily: tool discovery
         # imports most of gptme and must not run at module import time.
         lambda: _known_tool_names() + ["none"],
-        allow_prefixes=["+", "-"],
+        allow_prefixes=["+", "-", "hint:"],
         extra_choices_for_prefix={"-": _known_tool_names},
-        # Only '+' is lenient: plugin tools (added via '+tool') aren't known at
-        # parse time. '-tool' exclusions stay strict against known tools so typos
-        # like '-shel' are caught early instead of being silently ignored.
-        lenient_prefixes=["+"],
+        # '+' is lenient: plugin tools (added via '+tool') aren't known at
+        # parse time. '-tool' exclusions stay strict against known tools so
+        # typos like '-shel' are caught early instead of being silently ignored.
+        # 'hint:' is also lenient: hint tag names (e.g. 'hint:read-only') are
+        # not in the tool-name choice set; they're validated at load time.
+        lenient_prefixes=["+", "hint:"],
         metavar="TOOL",
     ),
     help="Tools to allow. Comma-separated or repeated. Use '+tool' to add to defaults (e.g., '-t +subagent'). Use '-tool' to exclude from defaults (e.g., '-t=-browser'). Use 'none' to disable all tools. Supports .py file paths for custom tools (e.g., '-t path/to/tool.py'). See 'Available tools' above for the list.",
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -565,13 +565,14 @@ Run 'gptme-util --help' for all utility commands."""
         # imports most of gptme and must not run at module import time.
         lambda: _known_tool_names() + ["none"],
         allow_prefixes=["+", "-", "hint:"],
-        extra_choices_for_prefix={"-": _known_tool_names},
+        extra_choices_for_prefix={
+            "-": _known_tool_names,
+            "hint:": lambda: ["read-only"],
+        },
         # '+' is lenient: plugin tools (added via '+tool') aren't known at
-        # parse time. '-tool' exclusions stay strict against known tools so
-        # typos like '-shel' are caught early instead of being silently ignored.
-        # 'hint:' is also lenient: hint tag names (e.g. 'hint:read-only') are
-        # not in the tool-name choice set; they're validated at load time.
-        lenient_prefixes=["+", "hint:"],
+        # parse time. '-tool' exclusions and built-in hint tags stay strict so
+        # typos like '-shel' or 'hint:red-only' fail before tool loading.
+        lenient_prefixes=["+"],
         metavar="TOOL",
     ),
     help="Tools to allow. Comma-separated or repeated. Use '+tool' to add to defaults (e.g., '-t +subagent'). Use '-tool' to exclude from defaults (e.g., '-t=-browser'). Use 'none' to disable all tools. Supports .py file paths for custom tools (e.g., '-t path/to/tool.py'). See 'Available tools' above for the list.",
--- a/gptme/config/cli_setup.py
+++ b/gptme/config/cli_setup.py
@@ -14,7 +14,6 @@ from ..tools import get_toolchain
 from ..tools._allowlist import (
     TOOL_PRESETS,
     expand_tool_allowlist_presets,
-    is_exclusive_preset_expansion,
 )
 from .chat import ChatConfig
 from .core import Config, get_config, set_config, set_config_from_workspace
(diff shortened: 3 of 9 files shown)


gptme/config/cli_setup.py
```diff
@@ -203,8 +206,15 @@ def setup_config_from_cli(
...
-    if not interactive:
+    tool_preset_selected = (
+        resolved_tool_allowlist is not None
+        and len(resolved_tool_allowlist) == 1
+        and resolved_tool_allowlist[0] in TOOL_PRESETS
+    )
+
+    # Automatically add 'complete' tool in non-interactive mode, except for
+    # exclusive named presets such as read-only audit mode.
+    if not interactive and not tool_preset_selected:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d6e038a8f — added `is_exclusive_preset_expansion()` to `_allowlist.py` that checks whether the saved allowlist exactly matches any preset's expansion. The `tool_preset_selected` guard in `cli_setup.py` now recognises the persisted `["read"]` form as well as the literal `"read-only"` name, so `complete` is not appended on non-interactive resume. Added a regression test (`test_setup_config_from_cli_read_only_preset_survives_noninteractive_resume`) covering the resume path.

TimeToBuildBob:
@greptileai review

gptme/config/cli_setup.py
```diff
@@ -203,8 +210,20 @@ def setup_config_from_cli(
...
+    # A preset is "selected" if the allowlist is a literal preset name, or if it
+    # exactly matches a preset's expansion (the persisted form on resume — the
+    # preset name is replaced by concrete tools when the config is saved).
+    tool_preset_selected = resolved_tool_allowlist is not None and (
+        (
+            len(resolved_tool_allowlist) == 1
+            and resolved_tool_allowlist[0] in TOOL_PRESETS
+        )
+        or is_exclusive_preset_expansion(resolved_tool_allowlist)
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 3c2ab1640. The root cause was that `is_exclusive_preset_expansion()` checked the *expanded* allowlist (`["read"]`) rather than the original spec, so any ordinary `--tools read` or `TOOL_ALLOWLIST=read` session was misclassified as a preset.

Fix: `_normalize_tool_allowlist` now preserves preset names verbatim (`["read-only"]`) instead of expanding them. Expansion still happens at toolchain initialisation time (`tools/__init__.py`). This gives unambiguous provenance: `["read-only"]` in saved tools = preset was used; `["read"]` = explicit list. `is_exclusive_preset_expansion` is removed entirely. A regression test (`test_setup_config_from_cli_explicit_read_tool_adds_complete_noninteractive`) now covers this case.

gptme/config/cli_setup.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 3c2ab1640 as part of the same change. The old code saved the *expanded* form `["read"]` which cannot be distinguished from an explicit `--tools read` on resume. Now the preset name `["read-only"]` is saved verbatim, so the literal-name check always fires on resume without any ambiguous expansion-equality heuristic.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated placeholder: work was recorded for this monitoring item but the worker finished before posting the thread update. See evidence below for what changed — the delivery guard is NOT closing the loop, just preventing silent orphaning.

Evidence: commit: a576eb034f docs(journal): monitoring gptme/gptme#3543 — fix preset provenance P1 bugs

Session: `4e72a247-0884-59e8-aae7-19294f0043c7`

TimeToBuildBob:
Ready for review. CI green across all checks, Greptile 5/5. Both P1 findings from the Greptile review (resume losing preset identity, expanded allowlist losing preset provenance) were addressed in d6e038a8f and 3c2ab1640.

Not eligible for self-merge (touches core CLI and tool files) — needs human merge.

docs/tools.rst
```diff
@@ -582,7 +592,7 @@ once using the ``hint:`` prefix:
 
 .. code-block:: bash
 
-    # Allow only tools annotated as read-only (safe for untrusted workspaces)
+    # Allow only tools annotated as read-only
     gptme --tools "hint:read-only" "summarise this repo"
```
TimeToBuildBob, line 625: ⚠️ **P2** — The CLI `CommaSeparatedChoice` for `--tools` validates each comma-separated value against the set of known tool names and preset names. Hint patterns like `hint:read-only` are not members of that set (the set is built from tool names and now `TOOL_PRESET_NAMES`, which only contains "read-only"), so `gptme --tools "hint:read-only"` fails with "invalid choice". This PR's documentation at line 596 explicitly presents that command as a working alternative, so users following the docs get an error. The fix is to add hint patterns to the accepted choices or handle them specially in the converter.

gptme/config/cli_setup.py
```diff
@@ -203,8 +220,18 @@ def setup_config_from_cli(
         # Fall back to env/config for new conversations or when no saved tools
         resolved_tool_allowlist = [tool.strip() for tool in tools_env.split(",")]
```
TimeToBuildBob: ⚠️ **P2** — In `setup_config_from_cli`, when `resolved_tool_allowlist` is read from the `TOOL_ALLOWLIST` env var, the split on "," does not filter empty strings. If the env var is "read-only," (trailing comma), the list becomes `["read-only", ""]`. The preset detection checks `len == 1`, so it is false, causing the non-interactive mode to append "complete" and later `_normalize_tool_allowlist` to call `expand_tool_allowlist_presets`, which raises ValueError because a preset is combined with another entry. The user gets an error despite a harmless trailing comma. The fix is to filter out empty strings after splitting.

gptme/cli/main.py
```diff
@@ -410,8 +410,11 @@ def _known_tool_names() -> list[str]:
     --help rendering) — never at module import time.
     """
     from ..tools import get_available_tools
+    from ..tools._allowlist import TOOL_PRESET_NAMES
 
-    return sorted(tool.name for tool in get_available_tools(include_mcp=False))
+    names = {tool.name for tool in get_available_tools(include_mcp=False)}
+    names.update(TOOL_PRESET_NAMES)
```
TimeToBuildBob, line 416: ⚠️ **P2** — Adding "read-only" to `_known_tool_names` makes the exclusion syntax `-read-only` a valid CLI choice. In `setup_config_from_cli`, exclusion removes any tool named "read-only" from the defaults, but no such tool exists, so it is silently ignored. A user might expect `-read-only` to disable the read-only preset, but it does nothing, which is confusing and could lead to a false sense of security if they think they cancelled a preset.

TimeToBuildBob:
## 🤖 AI code review

Adds a strict `read-only` tool preset that expands to only the built-in `read` tool, rejects combining it with other tools, and suppresses the automatic `complete` tool in non-interactive mode when the preset is selected. Updates CLI validation to accept preset names and `hint:read-only`, and adds comprehensive tests across cli, config, and tools.

**1 finding** · ⚠️ **1** P2 — no P0/P1 on this head, but P2 findings still prevent a 5/5 until they are resolved.

**⚠️ P2 medium** — `gptme/cli/main.py:570`

The CLI validation for the `hint:` prefix hardcodes the only acceptable value as `read-only` in `extra_choices_for_prefix`. The hint system in `_allowlist.py` documents and supports other hints (`destructive`, `idempotent`, `closed-world`), and users can already pass them via the `TOOL_ALLOWLIST` environment variable or config. But with `--tools "hint:destructive"`, the `CommaSeparatedChoice` validator rejects it at parse time as an invalid choice, so a documented capability is unusable from the command line. This is a functional regression for anyone relying on non-read-only hint allowlists via the CLI, forcing them to use env vars or config files instead.

```suggestion
Replace the hardcoded list with a dynamic lookup of all valid hint names, e.g. `lambda: list(_KNOWN_HINTS)` where `_KNOWN_HINTS` is a module-level constant of defined hints.
```

<sub>**How this was verified:** Checked the hint definitions in gptme/tools/_allowlist.py and the docs which list other hints; the CLI validator only accepts 'read-only' for the hint prefix.</sub>

</details>

- `docs/tools.rst` — Documents the new `read-only` preset, clarifies `hint:read-only` is broader, and adds usage examples.
- `gptme/cli/cmd_review_pr.py` — Imports `READ_ONLY_TOOL_PRESET` and uses it for the review tool preset instead of a local literal.
- `gptme/cli/main.py` — Adds `TOOL_PRESET_NAMES` to known tool names, adds `hint:` to allowed prefixes with `read-only` as the only extra choice, and updates the help example.
- `gptme/config/cli_setup.py` — Preserves preset names verbatim in the allowlist, detects and rejects `-preset` exclusions, moves the gear-profile override before the non-interactive `complete` addition, and filters empty entries from allowlist splitting.
- `gptme/tools/__init__.py` — Calls `expand_tool_allowlist_presets` in `init_tools` and `get_toolchain` to expand preset names into concrete tool names.
- `gptme/tools/_allowlist.py` — Defines `READ_ONLY_TOOL_PRESET`, `TOOL_PRESETS`, `TOOL_PRESET_NAMES`, and the `expand_tool_allowlist_presets` function that enforces preset exclusivity.
- `tests/test_cli.py` — Adds tests for the `read-only` preset and `hint:` prefix validation in the CLI.
- `tests/test_config.py` — Adds tests for preset preservation, `complete` suppression, exclusion rejection, trailing comma handling, and gear-profile `complete` addition.
- `tests/test_tools.py` — Adds tests for the read-only preset loading only the read tool and blocking mutating/execution tools, plus preset combination failures.

</details>

| `3c2ab1640a20` | 4/5 | 3 | llm | 2026-08-14 02:01 UTC |
| `4c0a576f4e92` | 3/5 | 2 | llm | 2026-08-14 02:46 UTC |
| `cf041b3ee554` | 4/5 | 2 | llm | 2026-08-14 05:50 UTC |

</details>

<sub>Reviewed `6612ebdfa4e3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 183s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

docs/tools.rst
TimeToBuildBob: Fixed in 4c0a576f4 — `hint:` is now treated as a lenient prefix in `CommaSeparatedChoice` (like `+`), so `hint:read-only` passes CLI parsing and is validated at tool-load time rather than at parse time.

gptme/config/cli_setup.py
TimeToBuildBob: Fixed in 4c0a576f4 — empty strings are now filtered from the comma-split: `[tool.strip() for tool in tools_env.split(',') if tool.strip()]`. A trailing comma no longer produces an empty-string entry that breaks preset detection.

gptme/cli/main.py
TimeToBuildBob: Fixed in 4c0a576f4 — `-read-only` now triggers an explicit warning: "Cannot exclude preset name(s) read-only with '-' syntax. Presets select an exclusive tool boundary — use '--tools read-only' to select one." The preset name is also excluded from the non_default list so it doesn't produce a second confusing "not in default toolset" warning.

TimeToBuildBob:
Fixed two P1/P2 findings from the latest AI review (commits da7404942 and cf041b3ee):

**P1 (security)**: `-read-only` passed CLI validation after `TOOL_PRESET_NAMES` was added to `_known_tool_names`, but the exclusion branch only warned and resolved to the full default toolset — a fail-open boundary. Promoted from `logging.warning` to `raise ValueError` so the call fails closed.

**P2**: The normal-mode branch (direct `tool_allowlist` parameter) built the resolved list without filtering empty strings; `"read-only,"` split to `["read-only", ""]`, causing `expand_tool_allowlist_presets` to raise. Added the same `if tool.strip()` filter already present in the env-var branch.

Tests updated: renamed `preset_exclusion_warns` → `preset_exclusion_raises`; added `test_setup_config_from_cli_tool_allowlist_direct_trailing_comma` mirroring the existing env-var variant.

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -560,12 +564,14 @@ def _known_tool_names() -> list[str]:
...
-        # Only '+' is lenient: plugin tools (added via '+tool') aren't known at
-        # parse time. '-tool' exclusions stay strict against known tools so typos
-        # like '-shel' are caught early instead of being silently ignored.
-        lenient_prefixes=["+"],
+        # '+' is lenient: plugin tools (added via '+tool') aren't known at
+        # parse time. '-tool' exclusions stay strict against known tools so
+        # typos like '-shel' are caught early instead of being silently ignored.
+        # 'hint:' is also lenient: hint tag names (e.g. 'hint:read-only') are
+        # not in the tool-name choice set; they're validated at load time.
+        lenient_prefixes=["+", "hint:"],
```
TimeToBuildBob: ⚠️ **P2** — The `hint:` prefix is now added to `lenient_prefixes` in CommaSeparatedChoice (main.py line 574), so any string following `hint:` passes CLI validation with no check that the hint actually exists. A user who mistypes a hint (e.g. `--tools "hint:red-only"`) gets no parse-time error, and the session silently loads zero tools at runtime: get_toolchain filters available tools by the hint and if none match, the allowlist contributes nothing, and init_tools skips hint patterns without raising. The model then runs with no capabilities, which is confusing and defeats the intended safety feedback that a typo would otherwise produce. Previously an invalid tool name was rejected at parse time with a clear error message. This is a P2 usability/error-handling regression; it is not a security issue because having no tools is more restrictive, but the silent result is a poor failure mode.

ErikBjare:
@TimeToBuildBob Fix the remaining P2 then merge.

gptme/cli/main.py
TimeToBuildBob: Fixed in 6612ebdfa. `hint:` validation is strict again: `hint:read-only` remains accepted, while typos such as `hint:red-only` fail during CLI parsing. Added a regression test for the typo path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed both remaining P2 findings in 6612ebdfa:

- `hint:` CLI validation is strict again: `hint:read-only` works, while unknown hints fail at parse time.
- gear-profile tool overrides now apply before non-interactive `complete` injection, with regression coverage for gear 0.

Targeted tests: 11 passed. The broader selected suite passed 250 tests / 8 skipped; its only failure is an unrelated session-harness log-directory assertion caused by this monitoring process's injected `GPTME_LOGS_DIR`. All scoped pre-commit hooks passed.

Fresh CI and Greptile review are running on 6612ebdfa. I will merge through the gated path once they clear.

TimeToBuildBob:
The P2 fixes and fresh Greptile review are complete (5/5), but I have not merged yet because the new CI run exposed one unrelated flaky test: `TestMaybeApplyVerbosity::test_invalid_level_warns_only_once` saw its warning twice under xdist (`9241 passed, 175 skipped`). The latest `master` Test workflow is green, so I made no unrelated code change. GitHub will not allow rerunning the failed job until the other test shard finishes; I’ll rerun that failed job, then use the gated merge path if it passes.

gptme/cli/main.py
```diff
@@ -560,11 +564,14 @@ def _known_tool_names() -> list[str]:
...
         lambda: _known_tool_names() + ["none"],
-        allow_prefixes=["+", "-"],
-        extra_choices_for_prefix={"-": _known_tool_names},
-        # Only '+' is lenient: plugin tools (added via '+tool') aren't known at
-        # parse time. '-tool' exclusions stay strict against known tools so typos
-        # like '-shel' are caught early instead of being silently ignored.
+        allow_prefixes=["+", "-", "hint:"],
+        extra_choices_for_prefix={
+            "-": _known_tool_names,
+            "hint:": lambda: ["read-only"],
```
TimeToBuildBob: ⚠️ **P2** — The CLI validation for the `hint:` prefix hardcodes the only acceptable value as `read-only` in `extra_choices_for_prefix`. The hint system in `_allowlist.py` documents and supports other hints (`destructive`, `idempotent`, `closed-world`), and users can already pass them via the `TOOL_ALLOWLIST` environment variable or config. But with `--tools "hint:destructive"`, the `CommaSeparatedChoice` validator rejects it at parse time as an invalid choice, so a documented capability is unusable from the command line. This is a functional regression for anyone relying on non-read-only hint allowlists via the CLI, forcing them to use env vars or config files instead.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining P2 in b7d821368 — `hint:destructive` (and any other tool hint like `file-ops`, `web`, `code-exec`) now accepted at CLI parse time.

Root cause: `extra_choices_for_prefix['hint:']` was hardcoded to `['read-only']`. Fix: added `_known_hint_names()` that collects all hint tags from available tools at validation time (same lazy-import pattern as `_known_tool_names()`). Added regression test `test_tools_hint_destructive_is_valid_cli_choice`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
✅ P2 fixed in commit b7d821368c — 'expand hint: CLI choices to all known tool hints'. The hardcoded read-only hint list has been replaced with dynamic `_known_hint_names()` that collects all hint tags from available tools. Regression test added for hint:destructive.

Ready for merge per Erik's directive to 'fix the remaining P2 then merge.'

TimeToBuildBob:
Done: fixed the remaining P2 in `b7d821368`, Greptile passed 5/5, and I merged the PR as `7adb7f765`. All post-merge checks are green.

PR #3546, opened 2026-08-14
fix(server): validate provider API key before saving it

### LINKED ISSUE #3545: Desktop BYOK setup shows success after invalid provider key
## Summary

## Reproduction

Environment:

Steps:

Observed:

Expected:

## Control

## Related

sujeito-operator:
## What happens

`POST /api/v2/user/api-key` checks the *shape* of the key and never asks the provider
whether it works. It writes the key into `config.local.toml`, exports it into
`os.environ`, and answers `{"status": "ok", "restart_required": false}`. The desktop
BYOK wizard treats that 200 as success, restarts, reads `provider_configured` — which is
`get_default_model() is not None`, i.e. "a MODEL is set", not "the key is good" — and
shows `You're all set!`. The first generation is where the user learns the key is bad,
as a raw provider 401.

## The validator is already in the tree

`gptme/llm/validate.py::validate_api_key` makes a cheap live call per provider and
returns `(is_valid, message)`. `/account setup` has called it all along and refuses to
store a key that fails (`gptme/commands/account.py`). The onboarding endpoint is the one path that
skips it, so the same product has two doors onto the same action and only one is locked.

This calls the same validator from the endpoint, before anything is written, and returns
`400 {"error": ...}` when the provider rejects the key.

## No webui change is needed

`saveApiKeyToServer` already parses `error` out of a non-ok response and throws
(`webui/src/components/SetupWizard.tsx`), and `handleSaveApiKey` already catches it into
`apiKeyError` without restarting the server or advancing the step. A 400 from this
endpoint stops the wizard on the provider step showing the provider's own reason.

## Why the validator rather than new checks

It already decides the awkward cases, and re-deciding them here would let the two paths
drift: HTTP 429 is a valid key, an Anthropic 400 whose message mentions usage limits is a
valid key **with a message**, and providers with no check (azure, nvidia, local, custom)
are skipped and still save. `test_v2_user_api_key_saves_when_validation_is_non_fatal`
pins that: a quota-exhausted key still saves and the response shape does not change.

The response shape is unchanged on purpose. An earlier revision of this branch returned
that non-fatal message as a `warning` field, and a review pointed out that no client
reads it — surfacing a quota warning needs a change in `SetupWizard.tsx` and
`ServerApiKeySettings.tsx`, which is a separate piece of work with its own tests, and I
have no way to run the webui suite here. So the field is gone rather than half-wired.

## What I ran, and what I did not

Ran, on this branch: `pytest tests/test_server_v2.py tests/test_server_cli_fallback.py`
— **230 passed, 5 skipped**. `ruff format --check`, `ruff check`
and `mypy` clean on both changed files. The new rejection test was confirmed to fail
without the endpoint change and pass with it.

**Not run: the AppImage repro from the issue.** I have no desktop build, display or
packaged sidecar here, so I have not re-driven the wizard end to end. The claim about the
webui above is from reading `SetupWizard.tsx`, not from running it.

## Noticed, deliberately not changed

The endpoint answers `restart_required: false` and its comment says the key is "applied
live". That holds only while the provider's client has not been built yet — `init_llm()`
is idempotent ("if not already initialized") and this endpoint never calls the `reinit()`
helpers that `/account` uses for exactly this — so re-keying a running server does not
take effect live. Separate defect, and whether onboarding should reinit is your call, so
it is not folded in here.

Closes #3545.

### DIFF
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -3069,6 +3069,8 @@ def api_user():
     summary="Save provider API key",
     description=(
         "Persist a provider API key into the user's global gptme config. "
+        "The key is checked against the provider before it is written, so an "
+        "invalid key is rejected with 400 rather than saved. "
         "Intended for first-run onboarding flows; callers should restart the "
         "server after a successful write if they need the running process to "
         "pick the key up immediately."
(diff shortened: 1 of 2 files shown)


ErikBjare:
@greptileai review

Issue #3451, opened 2026-08-05
feat(review): add --trusted-reviewer guard for artifact mode


TimeToBuildBob:
## Background

PR #3449 added `--artifact` mode to `gptme-util review watch`, enabling a
GitHub-less pipeline where a `ReviewArtifact` JSON file is loaded and its
finding bodies are injected into a fix-session prompt.

During the Greptile review of #3449 a P1 (Security) finding was raised:

> **Artifact instructions bypass reviewer trust** — When a caller supplies a
> crafted artifact file or stdin payload, its unvalidated finding bodies become
> authoritative instructions for a repository-capable session told to commit and
> push, allowing unintended changes to be pushed under the user's credentials.

The trust model in #3449 was deliberately kept simple: `--artifact` is a
local CLI tool; the caller explicitly chose to load that artifact, similar to
running `sh script.sh`. That reasoning stands for the common case.

However, the stdin mode (`--artifact -`) is a wider surface — piped payloads
can come from sources the caller may not have fully audited (e.g., a
CI-forwarded artifact from a remote reviewer).

## Proposed Hardening

Add a `--trusted-reviewer` option (or `--trust-artifact`) that gates which
reviewer identities may appear in the artifact before its findings are
injected as instructions:

```bash
# Only inject findings authored by these GitHub logins
gptme-util review watch --artifact artifact.json --trusted-reviewer ErikBjare
```

Implementation sketch:
1. `ReviewArtifact.findings` already carry `author` metadata (or can be
   extended to do so via `pr_review` in gptme-contrib).
2. `review watch --trusted-reviewer <login>` filters findings to only those
   whose `author` matches the allowlist before building the prompt.
3. If no `--trusted-reviewer` is given, behavior is unchanged (all findings
   injected, current default).
4. Consider a `--require-trust` flag that hard-fails if `author` is absent
   from the artifact, preventing silent bypass.

## Acceptance Criteria

- [ ] `--trusted-reviewer <login>` flag added to `gptme-util review watch`
- [ ] Findings with non-matching or absent `author` are skipped (not injected)
- [ ] `ReviewFinding` schema extended with `author` field (nullable, default `None` = skip under `--require-trust`)
- [ ] Tests cover: all-trusted pass, partial filter, empty-after-filter behavior
- [ ] `--artifact -` (stdin) mode respects the same filter

## Related

- #3449 — artifact mode implementation
- Greptile P1 finding: comment thread on #3449

[...]

PR #3547, opened 2026-08-15
fix(review): verify finding file/line target matches inline comment location

TimeToBuildBob:
## Problem

`gptme-util review watch --trusted-reviewer LOGIN` (#3470) verifies a finding's
reviewer *identity* and *body* against the GitHub API, but does **not** verify
that the finding's `file`/`line` target matches the inline comment's code
location.

This leaves a forgery vector in the trusted-reviewer allowlist: an attacker who
can supply a crafted artifact could:

1. Replay a *genuine* trusted reviewer's `body` + `github_comment_id` (both
   cross-check correctly against GitHub), but
2. Change the `file` (and/or `line`) fields to point the fix session at
   attacker-chosen code.

The fix session would then be directed to make attacker-controlled edits to a
file the trusted reviewer never commented on.

## Fix

Extend `_verify_comment_reviewer` to cross-check the finding's code location:

- When a finding targets a `file`, the backing comment must be an **inline
  review comment** on that same file. A conversation/issue comment (which
  carries no code location) can no longer authenticate a file-targeted finding
  — rejected fail-closed.
- When the finding also specifies a `line`, it must match the inline comment's
  line (`original_line`, falling back to `line`, matching the round-trip
  semantics of `ReviewFinding.from_github_comment`).

`_filter_findings_by_trusted_reviewers` now passes the finding's `file` and
`line` through to verification, so the check is always on when
`--trusted-reviewer` is set (no new opt-in flag — body + location verification
should always accompany reviewer allowlisting).

## Tests

Added to `tests/test_util_cli_review_watch.py`:

- `test_filter_findings_forged_file_rejected` — genuine body + forged file → blocked
- `test_filter_findings_forged_line_rejected` — genuine body + forged line → blocked
- `test_filter_findings_matching_file_and_line_accepted` — all axes match → allowed
- `test_filter_findings_conversation_comment_with_file_target_rejected` — conversation comment can't authenticate a file target

## Verification

- `pytest tests/test_util_cli_review_watch.py tests/test_util_review.py` — 207 passed
- `ruff check` / `ruff format --check` clean on changed files

Refs: gptme/gptme#3451 (threat model), gptme/gptme#3460 (prior `--verify-bodies` work, closed as duplicate of #3470).

### DIFF
--- a/gptme/cli/cmd_review_watch.py
+++ b/gptme/cli/cmd_review_watch.py
@@ -436,21 +436,24 @@ def _verify_comment_reviewer(
             return False, ""
 
         if expected_line is not None:
-            # Inline review comments carry both ``line`` (current diff position)
-            # and ``original_line`` (position in the original file). The finding's
-            # ``line`` maps to ``original_line`` when round-tripped via
-            # ``ReviewFinding.from_github_comment``, so compare against that
-            # first and fall back to ``line`` for comments lacking it.
-            comment_line = data.get("original_line") or data.get("line")
-            if comment_line != expected_line:
+            # ``ReviewFinding.from_github_comment`` records ``original_line``,
+            # while other artifact producers may record the current ``line``.
+            # Both locations are authenticated by this inline comment, so accept
+            # either when GitHub returns both values for a moved line.
+            comment_lines = {
+                line
+                for line in (data.get("original_line"), data.get("line"))
+                if line is not None
+            }
+            if expected_line not in comment_lines:
                 logger.warning(
                     "GitHub comment %d line mismatch: artifact targets line %s "
-                    "on '%s', comment is on line %s; rejecting finding (prevents "
-                    "location forgery).",
+                    "on '%s', comment is on line(s) %s; rejecting finding "
+                    "(prevents location forgery).",
                     comment_id,
                     expected_line,
                     expected_file,
-                    comment_line,
+                    sorted(comment_lines),
                 )
                 return False, ""
 
--- a/gptme/cli/cmd_review_watch.py
+++ b/gptme/cli/cmd_review_watch.py
@@ -303,6 +303,8 @@ def _verify_comment_reviewer(
     expected_reviewer: str,
     expected_body: str = "",
     target_pr_number: int | None = None,
+    expected_file: str = "",
+    expected_line: int | None = None,
 ) -> tuple[bool, str]:
     """Verify a PR review comment belongs to the claimed reviewer and has matching body.
 
@@ -310,14 +312,17 @@ def _verify_comment_reviewer(
     1. ``user.login`` case-insensitively (using casefold for proper Unicode handling)
     2. Comment body matches the artifact finding body exactly (if provided)
     3. If target_pr_number is provided, verifies the comment is on that specific PR
+    4. If expected_file is provided, verifies the comment is an inline review
+       comment on that same file (and same line when expected_line is given) so
+       a genuine trusted body cannot be replayed against a forged code location
 
     Tries the inline review-comment endpoint first (``pulls/comments``); falls
     back to the issue-comment endpoint (``issues/comments``) for PR-level
     conversation comments, which live in a separate ID space.
 
-    Returns (verified, body) where verified is True only if login matches and
-    body is authentic. When verification fails, body is empty string.
-    Fails closed on network errors or mismatches.
+    Returns (verified, body) where verified is True only if login, body, and
+    (when requested) file/line location are authentic. When verification fails,
+    body is empty string. Fails closed on network errors or mismatches.
     """
     # Try inline review comment endpoint first.
     data = run_gh_json(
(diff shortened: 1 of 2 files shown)


gptme/cli/cmd_review_watch.py
```diff
@@ -400,6 +405,55 @@ def _verify_comment_reviewer(
...
+            )
+            return False, ""
+
+        if expected_line is not None:
+            # Inline review comments carry both ``line`` (current diff position)
+            # and ``original_line`` (position in the original file). The finding's
+            # ``line`` maps to ``original_line`` when round-tripped via
+            # ``ReviewFinding.from_github_comment``, so compare against that
+            # first and fall back to ``line`` for comments lacking it.
+            comment_line = data.get("original_line") or data.get("line")
```
TimeToBuildBob: ❌ **P1** — The new line check uses `comment_line = data.get("original_line") or data.get("line")`, preferring `original_line` whenever it is present. For a comment on a modified hunk, GitHub sets both `original_line` (base position) and `line` (head position) to different values. The round-trip via `ReviewFinding.from_github_comment` is documented to use `original_line`, so findings created that way match. However, any caller that constructs a `ReviewFinding` with the head-side `line` (e.g., a different artifact producer that stores the forward line) will now be rejected for the same file and body, even though the comment is genuinely on that file. This is a breaking change for existing callers and can cause valid findings to be silently filtered out, leaving the fix session to miss instructions the trusted reviewer actually left. Concretely, if the comment is on a line that moved from 10 to 15, a finding with line=15 is rejected against original_line=10, so the location check fails despite the file and body matching.

tests/test_util_cli_review_watch.py
```diff
@@ -1100,6 +1100,159 @@ def fake_run_gh_json(args, **kwargs):
...
+    result = cmd_review_watch._filter_findings_by_trusted_reviewers(
+        findings,
+        ("ErikBjare",),
+        owner="owner",
+        repo="repo",
+    )
+    assert len(result) == 0, "Forged line target must be rejected"
+
+
+def test_filter_findings_matching_file_and_line_accepted(monkeypatch):
```
TimeToBuildBob, line 1178: ⚠️ **P2** — The test `test_filter_findings_matching_file_and_line_accepted` only exercises the case where the comment has `original_line` set. There is no test for the fallback path where `original_line` is absent and the comment only carries `line` (the typical case for a comment on a line added in the new revision). The `or` fallback in `comment_line = data.get("original_line") or data.get("line")` is thus never exercised, so a regression in that branch (e.g., if `original_line` were accidentally present in all responses but ignored) would go unnoticed. This is a test-coverage gap for the new behavior introduced by this PR.

tests/test_util_cli_review_watch.py
```diff
@@ -1100,6 +1100,159 @@ def fake_run_gh_json(args, **kwargs):
     assert len(result) == 0, "Forged finding body must be rejected"
 
 
+def test_filter_findings_forged_file_rejected(monkeypatch):
```
TimeToBuildBob, line 1103: ⚠️ **P2** — None of the new tests cover a finding that has a `file` but no `line`. In `_verify_comment_reviewer`, when `expected_file` is set and `expected_line` is None, the code only checks the path and skips the line comparison. This path is reachable from `_filter_findings_by_trusted_reviewers` because `f.line` may be None. Without a test, a future change that accidentally requires a line (e.g., by removing the `if expected_line is not None` guard) would break legitimate file-only findings, and the omission would not be detected. The existing tests all include a line, so this branch is untested.

TimeToBuildBob:
## 🤖 AI code review

Adds file and line verification to _verify_comment_reviewer so that a finding with a file target is only accepted when the backing GitHub comment is an inline review comment on the same file (and same line when specified). Conversation comments can no longer authenticate file-targeted findings. _filter_findings_by_trusted_reviewers passes the finding's file and line into the check. Tests added for file/line mismatch, matching, line-only comments, and conversation-comment rejection.

- `gptme/cli/cmd_review_watch.py` — Extends _verify_comment_reviewer with expected_file/expected_line checks and rejects conversation comments for file-targeted findings; passes these new fields from _filter_findings_by_trusted_reviewers.
- `tests/test_util_cli_review_watch.py` — Adds six tests covering forged file/line rejection, matching acceptance, current-line and line-only acceptance, file-without-line acceptance, and conversation-comment rejection.

</details>

| `9b8f7d527ac3` | 3/5 | 3 | llm | 2026-08-15 00:42 UTC |
| `f81430af2639` | 5/5 | 0 | llm | 2026-08-16 19:08 UTC |

</details>

<sub>Reviewed `f81430af2639` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 150s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Addressed all three review findings in `f81430af2`:

- accept either authenticated `original_line` or current `line` when GitHub returns both for one inline comment;
- cover a comment with only `line`;
- cover a file-only finding with no line.

Rebased onto current `master`. Focused verification: 60 review-watch tests and 150 review-model tests pass; Ruff and pre-commit pass. CI/review are running on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI is green, the PR is mergeable, Greptile is 5/5, and all review threads are resolved. The self-merge policy gate still denies this PR because `gptme/cli/cmd_review_watch.py` is outside the allowed categories, so I did not merge it. Waiting for maintainer review/merge.

TimeToBuildBob:

Session: `eac500ec-9d7e-5ccd-af8e-e657dd585b31`

PR #3549, opened 2026-08-15
Add TrustedRouter provider

jperla:
Adds TrustedRouter as a provider. It's an OpenAI-compatible router — one key reaches models from OpenAI, Anthropic, Google, DeepSeek and others, with per-request routing and failover.

Wired the same way as `openrouter` / `requesty`: added to `BuiltinProvider` and `PROVIDERS_OPENAI`, a client in `llm_openai.py` at `https://api.trustedrouter.com/v1`, plus `PROVIDER_API_KEYS`, `PROVIDER_DEFAULT_MODELS`, `PROVIDER_DOCS`, and a `MODELS` entry (required by the `set(PROVIDERS) == set(MODELS.keys())` assert in `data.py`).

**On the double prefix.** Model ids are namespaced and TrustedRouter rejects a bare id, so `MODELS["trustedrouter"]` keys are the full path after `trustedrouter/`. `trustedrouter/trustedrouter/auto` resolves to provider `trustedrouter` + model `trustedrouter/auto` — which is what the API expects. I verified this through `get_model()` rather than assuming:

```
trustedrouter/trustedrouter/auto      -> provider=trustedrouter model=trustedrouter/auto      ctx 200000
trustedrouter/anthropic/claude-opus-4-7 -> provider=trustedrouter model=anthropic/claude-opus-4-7 ctx 1000000
```

The `trustedrouter/*` keys are routing policies rather than single models: each picks an upstream per request and fails over, and `zdr` restricts routing to providers that retain no data. Their listed prices are the blended rate the router bills for the policy, so they're averages rather than one model's — noted in a comment above the block since it's otherwise misleading.

Verified: `pytest -k "model or provider"` → 560 passed, 60 skipped, 0 failed. (First run showed 21 errors, all `fixture 'mocker' not found` — my venv was missing `pytest-mock`, unrelated to this change; clean after installing it.)


ErikBjare:
I don't trust TrustedRouter, thus won't merge it. Feel free to make it into a provider plugin instead (but given you seem to be spamming these PRs, you probably don't really care)

PR #3555, opened 2026-08-16
fix(server): validate provider API key before persisting in BYOK setup

### LINKED ISSUE #3545: Desktop BYOK setup shows success after invalid provider key

TimeToBuildBob:
## Summary

Fixes #3545 — the desktop BYOK setup flow was showing "You're all set!" after an invalid provider key, then failing on the first generation with a raw HTTP 500 auth error.

**Root cause**: `/api/v2/user/api-key` wrote the key to `config.local.toml` and applied it to `os.environ` without checking whether the key was valid against the provider.

**Fix**: Call `validate_api_key()` from `gptme.llm.validate` (which already had per-provider validation logic) before persisting. If the key fails live validation, return HTTP 422 with an actionable error message. The key is only written to disk and applied to `os.environ` if validation passes.

### Changes

- `gptme/server/api_v2.py` — import and call `validate_api_key()` before saving; return 422 on failure; surface non-fatal warnings (e.g. quota exhausted) in the response body
- `gptme/server/openapi_docs.py` — add `skip_validation: bool` to request schema (for offline/test use); add `warning: str | None` to response schema for quota-exhausted cases
- `tests/test_server_v2.py` — 4 new tests: invalid key rejection (422 with error), valid key acceptance, quota-exhausted warning passthrough, and `skip_validation=True` bypass; 3 existing tests updated to use `skip_validation=True` to avoid live network calls

## Test plan

- [x] `uv run pytest tests/test_server_v2.py -k "api_key" -v` — 9 tests pass (5 existing + 4 new)
- [x] `uv run pytest tests/test_llm_validate.py -v` — 14 tests pass (unchanged)
- [x] Pre-commit hooks passed (ruff, mypy)

### DIFF
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -169,13 +169,10 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
         error = error_data.get("error", {})
         error_type = error.get("type", "").lower()
         error_msg = error.get("message", "").lower()
-        if (
-            "authentication" in error_msg
-            or "invalid api key" in error_msg
-            or "invalid_api_key" in error_type
-            or "authentication_error" in error_type
-        ):
-            return False, "Invalid API key. Please check your key and try again."
+        # Check permission_error BEFORE auth substrings: the key is valid even if
+        # the message happens to contain "authentication" (e.g. "Authentication
+        # required to access model"). If auth-substrings were checked first, a
+        # permission_error with such a message would be misclassified as invalid.
         if "permission_error" in error_type:
             # Key is valid but lacks access to the probe model; the key will work
             # for models the account can actually access.
@@ -183,6 +180,13 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
                 True,
                 "API key valid but lacks access to the probe model. It may work for other models.",
             )
+        if (
+            "authentication" in error_msg
+            or "invalid api key" in error_msg
+            or "invalid_api_key" in error_type
+            or "authentication_error" in error_type
+        ):
+            return False, "Invalid API key. Please check your key and try again."
         if "usage limits" in error_msg:
             # Key is valid but account has hit its usage quota
             raw_msg = error_data.get("error", {}).get("message", "")
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -35,6 +35,20 @@ VALIDATION_CONNECTION_ERROR = "Could not connect to the API. Please check your n
 VALIDATION_PROVIDER_ERROR_PREFIX = "Validation failed:"
 
 
+def _http_status_error(status_code: int) -> tuple[bool, str]:
+    """Map an unexpected HTTP status to a validation error tuple.
+
+    5xx responses mean the provider is down, not that the key is bad, so they
+    use VALIDATION_PROVIDER_ERROR_PREFIX so api_v2 can return 502 instead of 422.
+    """
+    if status_code >= 500:
+        return (
+            False,
+            f"{VALIDATION_PROVIDER_ERROR_PREFIX} Provider unavailable (HTTP {status_code})",
+        )
+    return False, f"API returned status {status_code}"
+
+
 def validate_api_key(
     api_key: str,
     provider: str,
@@ -77,12 +91,22 @@ def validate_api_key(
         if provider == "xai":
             return _validate_xai(api_key, timeout)
         if provider == "azure":
-            # Azure requires endpoint configuration, skip validation
+            # Azure requires endpoint configuration, skip live validation
             logger.info("Azure API key validation skipped (requires endpoint config)")
-            return True, ""
-        if provider in ("nvidia", "local"):
-            # Local models don't need API key validation
-            logger.info(f"{provider} provider doesn't require API key validation")
+            return (
+                True,
+                "Key accepted without live validation — Azure requires endpoint configuration to validate",
+            )
+        if provider == "nvidia":
+            # NVIDIA validation would need an org-specific endpoint, skip
+            logger.info("NVIDIA API key validation skipped (requires org endpoint)")
+            return (
+                True,
+                "Key accepted without live validation — NVIDIA keys are checked on first use",
+            )
+        if provider == "local":
+            # Local models typically use a placeholder key or none at all
+            logger.info("Local provider doesn't require API key validation")
             return True, ""
         # Unknown or custom provider, skip validation
         logger.info(f"No validation available for provider: {provider}")
@@ -111,7 +135,7 @@ def _validate_openai(api_key: str, timeout: int) -> tuple[bool, str]:
     if response.status_code == 429:
         # Rate limited but key is valid
         return True, ""
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -159,7 +183,7 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
     if response.status_code == 429:
         # Rate limited but key is valid
         return True, ""
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_openrouter(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -180,7 +204,7 @@ def _validate_openrouter(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, "Invalid API key. Please check your key and try again."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_openai_compatible(
@@ -201,7 +225,7 @@ def _validate_openai_compatible(
         return False, "API key forbidden. It may lack required permissions."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_google(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -223,7 +247,7 @@ def _validate_google(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, error.get("message", "Unknown error")
     if response.status_code == 403:
         return False, "API key forbidden. It may lack required permissions."
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_groq(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -240,7 +264,7 @@ def _validate_groq(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, "Invalid API key. Please check your key and try again."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_deepseek(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -257,7 +281,7 @@ def _validate_deepseek(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, "Invalid API key. Please check your key and try again."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
 
 
 def _validate_xai(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -274,4 +298,4 @@ def _validate_xai(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, "Invalid API key. Please check your key and try again."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -2,7 +2,10 @@
 
 import logging
 
-import requests
+try:
+    import requests
+except ImportError:
+    requests = None  # type: ignore
 
 logger = logging.getLogger(__name__)
 
@@ -67,6 +70,10 @@ def validate_api_key(
         - (True, "") if valid
         - (False, "error description") if invalid
     """
+    if requests is None:
+        logger.warning("requests library not available, skipping API key validation")
+        return True, ""
+
     try:
         if provider == "openai":
             return _validate_openai(api_key, timeout)
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -247,6 +247,8 @@ def _validate_google(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, error.get("message", "Unknown error")
     if response.status_code == 403:
         return False, "API key forbidden. It may lack required permissions."
+    if response.status_code == 429:
+        return True, ""
     return _http_status_error(response.status_code)
 
 
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -2,10 +2,7 @@
 
 import logging
 
-try:
-    import requests
-except ImportError:
-    requests = None  # type: ignore
+import requests
 
 logger = logging.getLogger(__name__)
 
@@ -70,10 +67,6 @@ def validate_api_key(
         - (True, "") if valid
         - (False, "error description") if invalid
     """
-    if requests is None:
-        logger.warning("requests library not available, skipping API key validation")
-        return True, ""
-
     try:
         if provider == "openai":
             return _validate_openai(api_key, timeout)
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -32,6 +32,7 @@ OAUTH_PROVIDERS: set[str] = {"openai-subscription", "grok-subscription"}
 # credential failures without parsing provider-specific responses.
 VALIDATION_TIMEOUT_ERROR = "Request timed out. Please check your network connection."
 VALIDATION_CONNECTION_ERROR = "Could not connect to the API. Please check your network."
+VALIDATION_PROVIDER_ERROR_PREFIX = "Validation failed:"
 
 
 def validate_api_key(
@@ -92,7 +93,7 @@ def validate_api_key(
         return False, VALIDATION_CONNECTION_ERROR
     except requests.exceptions.RequestException as e:
         logger.exception(f"Unexpected error validating {provider} API key")
-        return False, f"Validation failed: {e}"
+        return False, f"{VALIDATION_PROVIDER_ERROR_PREFIX} {e}"
 
 
 def _validate_openai(api_key: str, timeout: int) -> tuple[bool, str]:
@@ -194,8 +195,10 @@ def _validate_openai_compatible(
 
     if response.status_code == 200:
         return True, ""
-    if response.status_code in (401, 403):
+    if response.status_code == 401:
         return False, "Invalid API key. Please check your key and try again."
+    if response.status_code == 403:
+        return False, "API key forbidden. It may lack required permissions."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
     return False, f"API returned status {response.status_code}"
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -174,9 +174,15 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
             or "invalid api key" in error_msg
             or "invalid_api_key" in error_type
             or "authentication_error" in error_type
-            or "permission_error" in error_type
         ):
             return False, "Invalid API key. Please check your key and try again."
+        if "permission_error" in error_type:
+            # Key is valid but lacks access to the probe model; the key will work
+            # for models the account can actually access.
+            return (
+                True,
+                "API key valid but lacks access to the probe model. It may work for other models.",
+            )
         if "usage limits" in error_msg:
             # Key is valid but account has hit its usage quota
             raw_msg = error_data.get("error", {}).get("message", "")
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -52,6 +52,14 @@ def validate_api_key(
             return _validate_anthropic(api_key, timeout)
         if provider == "openrouter":
             return _validate_openrouter(api_key, timeout)
+        if provider == "requesty":
+            return _validate_openai_compatible(
+                api_key, timeout, "https://router.requesty.ai/v1"
+            )
+        if provider == "moonshot":
+            return _validate_openai_compatible(
+                api_key, timeout, "https://api.moonshot.ai/v1"
+            )
         if provider in ("google", "gemini"):
             return _validate_google(api_key, timeout)
         if provider == "groq":
@@ -155,6 +163,25 @@ def _validate_openrouter(api_key: str, timeout: int) -> tuple[bool, str]:
     if response.status_code == 200:
         return True, ""
     if response.status_code == 401:
+        return False, "Invalid API key. Please check your key and try again."
+    if response.status_code == 429:
+        return True, ""  # Rate limited but key is valid
+    return False, f"API returned status {response.status_code}"
+
+
+def _validate_openai_compatible(
+    api_key: str, timeout: int, base_url: str
+) -> tuple[bool, str]:
+    """Validate an OpenAI-compatible provider key by listing models."""
+    response = requests.get(
+        f"{base_url.rstrip('/')}/models",
+        headers={"Authorization": f"Bearer {api_key}"},
+        timeout=timeout,
+    )
+
+    if response.status_code == 200:
+        return True, ""
+    if response.status_code in (401, 403):
         return False, "Invalid API key. Please check your key and try again."
     if response.status_code == 429:
         return True, ""  # Rate limited but key is valid
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -173,6 +173,8 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
             "authentication" in error_msg
             or "invalid api key" in error_msg
             or "invalid_api_key" in error_type
+            or "authentication_error" in error_type
+            or "permission_error" in error_type
         ):
             return False, "Invalid API key. Please check your key and try again."
         if "usage limits" in error_msg:
--- a/gptme/llm/validate.py
+++ b/gptme/llm/validate.py
@@ -18,6 +18,8 @@ PROVIDER_DOCS: dict[str, str] = {
     "xai": "https://console.x.ai/",
     "azure": "https://portal.azure.com/#view/Microsoft_Azure_ProjectOxford/CognitiveServicesHub",
     "nvidia": "https://build.nvidia.com/",
+    "requesty": "https://app.requesty.ai/api-keys",
+    "moonshot": "https://platform.moonshot.ai/console/api-keys",
     "local": "https://gptme.org/docs/providers.html#local-models",
     "openai-subscription": "https://gptme.org/docs/providers.html#openai-subscription",
     "grok-subscription": "https://gptme.org/docs/providers.html#grok-subscription",
@@ -25,6 +27,11 @@ PROVIDER_DOCS: dict[str, str] = {
 
 # Providers that use OAuth instead of API keys
 OAUTH_PROVIDERS: set[str] = {"openai-subscription", "grok-subscription"}
+
+# Stable messages used by callers to distinguish provider availability failures from
+# credential failures without parsing provider-specific responses.
+VALIDATION_TIMEOUT_ERROR = "Request timed out. Please check your network connection."
+VALIDATION_CONNECTION_ERROR = "Could not connect to the API. Please check your network."
 
 
 def validate_api_key(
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_server_v2.py` (+133/-2, Δ135)
- `gptme/server/api_v2.py` (+22/-8, Δ30)
- `gptme/server/openapi_docs.py` (+8/-0, Δ8)

### Suggested verification
- `uv run pytest tests/test_server_v2.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3555._

gptme/server/api_v2.py
```diff
@@ -3111,6 +3115,15 @@ def api_user_api_key():
     except ValueError as exc:
         return flask.jsonify({"error": str(exc)}), 400
 
+    # Validate the key against the live provider API before persisting.
+    key_warning: str | None = None
+    if not skip_validation:
+        is_valid, validation_msg = validate_api_key(trimmed_api_key, provider)
```
TimeToBuildBob: ❌ **P1** — The call to validate_api_key at line 3121 is not wrapped in a try/except. validate_api_key performs a live HTTP request to the provider; if the provider is unreachable, times out, or returns an unexpected status, it can raise an exception (e.g. requests.exceptions.RequestException, urllib.error.URLError, or a ValueError). When that happens, the exception propagates out of api_user_api_key, and Flask returns a plain HTML 500 error page instead of a JSON error. This defeats the PR's purpose: a user with a misconfigured network or a temporarily down provider still sees a raw server error rather than a clean message. The existing code only catches ValueError from _validate_api_key_input and _validate_model_input earlier, not from the validation call itself. To fix, wrap the call in try/except and return a structured error (e.g. 502 or 503) with the exception message.

tests/test_server_v2.py
```diff
@@ -461,6 +470,128 @@ def test_v2_user_api_key_rejects_unknown_provider(client: FlaskClient):
...
+    """An invalid provider key should return 422 before persisting anything."""
+    import gptme.config.user as user_mod
+
+    config_file = tmp_path / "config.toml"
+    monkeypatch.setattr(user_mod, "config_path", str(config_file))
+    monkeypatch.setattr("gptme.config.core.reload_config", lambda: None)
+    monkeypatch.delenv("ANTHROPIC_API_KEY", raising=False)
+    monkeypatch.setattr(
+        "gptme.server.api_v2.validate_api_key",
+        lambda key, provider, **kw: (
```
TimeToBuildBob, line 567: ⚠️ **P2** — The new tests for validation failure only mock validate_api_key to return a tuple (False, message). No test simulates validate_api_key raising an exception (e.g. a network timeout). As a result, the unhandled-exception path described above is completely untested, so a regression that turns a provider outage into a 500 instead of a clean error would not be caught. Add a test where the mocked validator raises an exception and assert the endpoint returns a JSON error with a non-500 status (e.g. 502). This would have caught the missing try/except and guards against future regressions.

TimeToBuildBob:
## 🤖 AI code review

Adds `requesty` and `moonshot` to provider docs and routes their key validation through a new `_validate_openai_compatible` helper; reclassifies validation failures in `/api/v2/user/api-key` into 422 (credential) vs 502 (connectivity/provider) by having 5xx responses carry a `VALIDATION_PROVIDER_ERROR_PREFIX`; adds a `skip_validation` request flag and a `warning` response field; and reorders the Anthropic 400 handler so `permission_error` types win over auth-substring matching, with Azure/NVIDIA skips now returning warnings. Updates OpenAPI models and tests accordingly.

### Confidence Score: N/A

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/llm/validate.py:237`

A 403 on the probe request maps to `(False, ...)`, and api_v2 turns every non-502 validation message into a 422 rejection that prevents the key from being persisted — the same treatment as a 401 invalid key. But a 403 is an authorization failure on the probe endpoint, not proof the credential is invalid: a Gemini key whose project has not enabled the Generative Language API returns 403 PERMISSION_DENIED on `/models`, and org-scoped Requesty/Moonshot keys can be forbidden on model-listing while still authorizing chat. In both cases the BYOK flow refuses to save a key that would work (or would work once the API is enabled), forcing the user to stop and retry. The message even hedges with 'may lack required permissions', yet the endpoint treats it identically to a bad key. `_validate_google` (line 261) has the same 403 → rejected behavior, so the condition spans both new and existing validators.

<sub>**How this was verified:** Checked `_validate_openai_compatible` (lines 236-237) and `_validate_google` (line 261), both of which return `(False, ...)` on 403; api_v2 lines 3143-3149 map any message not matching the 502 set to 422 without distinguishing 401 from 403.</sub>

</details>

- `gptme/llm/validate.py` — Adds `_http_status_error`, `_validate_openai_compatible`, requesty/moonshot dispatch and docs, stable validation-message constants, Anthropic `permission_error` handling, Google 429 handling, and warning-bearing Azure/NVIDIA skips.
- `gptme/server/api_v2.py` — Adds `skip_validation` parsing, wraps `validate_api_key` to return 422 for credential failures and 502 for connectivity/provider failures, and adds an optional `warning` field to the save response.
- `gptme/server/openapi_docs.py` — Adds `skip_validation: bool` to `UserApiKeySaveRequest` and `warning: str | None` to `UserApiKeySaveResponse`, and lists 422/502 in the endpoint responses.
- `tests/test_llm_validate.py` — Adds requesty/moonshot dispatch tests, Google and OpenAI-compatible status-classification tests, Anthropic `permission_error`/type-based tests, and azure/nvidia/local skip tests.
- `tests/test_server_v2.py` — Changes the patch target to the api_v2 import site, switches three persistence tests to `skip_validation=True`, updates the invalid-key assertion from 400 to 422, and adds 422/502/warning/skip_validation endpoint tests.

</details>

| `0b6c53b9f7dc` | 4/5 | 3 | llm | 2026-08-16 17:06 UTC |
| `0ec5c87f71f0` | 4/5 | 2 | llm | 2026-08-16 19:05 UTC |
| `3cbdf5689be6` | 4/5 | 1 | llm | 2026-08-16 20:25 UTC |
| `0be523670a66` | 5/5 | 0 | llm | 2026-08-16 22:07 UTC |
| `0be523670a66` | 3/5 | 1 | llm | 2026-08-16 22:20 UTC |
| `e19b95f74d54` | 3/5 | 1 | llm | 2026-08-17 11:41 UTC |
| `9c1ceebd6082` | 3/5 | 1 | llm | 2026-08-17 13:21 UTC |
| `bd0ee352ab5d` | 3/5 | 1 | llm | 2026-08-17 15:23 UTC |
| `1b46a3ebc331` | 3/5 | 1 | llm | 2026-08-17 18:54 UTC |
| `ddc2b3a7ea0f` | 3/5 | 1 | llm | 2026-08-18 19:28 UTC |

</details>

<sub>Reviewed `67c823010fdb` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 702s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/server/api_v2.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `584e640f2`: Requesty and Moonshot now use the shared OpenAI-compatible `/models` validator, with routing and invalid-key coverage in `tests/test_llm_validate.py`.

TimeToBuildBob: Fixed in `584e640f2`: the endpoint now catches unexpected validator failures, logs the exception server-side, returns a structured 502 without exposing provider internals, documents that response, and has a regression test proving no key is persisted.

tests/test_server_v2.py
TimeToBuildBob: Fixed in `584e640f2`: `test_v2_user_api_key_handles_validator_exception` makes the validator raise, asserts the JSON 502 response, and verifies `config.local.toml` was not created.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the three open AI-review findings in `0b6c53b9f`:

- provider timeouts and connection failures now return 502 rather than being misclassified as invalid credentials (422)
- Requesty and Moonshot now have provider-documentation links
- Anthropic 400 responses explicitly identifying an invalid API key are rejected

Added regression coverage for all three paths. Focused validation passes: 18 tests, ruff clean, and pre-commit clean. Re-review is now in progress; CI is still running.

TimeToBuildBob:

Session: `e5af7f15-ea7e-5d96-8743-cc9b97e86059`

tests/test_llm_validate.py
```diff
@@ -139,6 +176,25 @@ def test_invalid_key_returns_false(self, mock_get):
         assert "Invalid API key" in error
 
 
+class TestValidateOpenAICompatible:
+    """Tests for OpenAI-compatible provider key validation."""
+
+    @patch("gptme.llm.validate.requests.get")
+    def test_invalid_key_returns_false(self, mock_get):
+        mock_get.return_value = Mock(status_code=401)
```
TimeToBuildBob: ⚠️ **P2** — The new _validate_openai_compatible() gates BYOK saves for the requesty and moonshot providers, yet the only unit test exercising it — TestValidateOpenAICompatible.test_invalid_key_returns_false (tests/test_llm_validate.py:183) — covers a single 401 response. The 200-valid, 403-invalid, 429-valid, and default status-snippet branches have no coverage. The dispatch tests (test_requesty_provider_calls_compatible_validator, test_moonshot_provider_calls_compatible_validator) only assert the base URL is forwarded; they mock _validate_openai_compatible wholesale and never exercise status handling. Because the 429-as-valid and 403-as-invalid decisions directly determine whether an onboarding user's key is accepted or rejected, an erroneous classification (e.g. a provider returning 403 for a scope restriction on a valid key) would ship and surface as a confusing 422 in the API with no test to catch it. A parametrized test over 200/401/403/429/500 would close the gap.

gptme/llm/validate.py
```diff
@@ -161,6 +182,25 @@ def _validate_openrouter(api_key: str, timeout: int) -> tuple[bool, str]:
...
+    """Validate an OpenAI-compatible provider key by listing models."""
+    response = requests.get(
+        f"{base_url.rstrip('/')}/models",
+        headers={"Authorization": f"Bearer {api_key}"},
+        timeout=timeout,
+    )
+
+    if response.status_code == 200:
+        return True, ""
+    if response.status_code in (401, 403):
```
TimeToBuildBob: ⚠️ **P2** — _validate_openai_compatible() treats HTTP 403 exactly like 401 as proof the API key is invalid (validate.py:197). For OpenAI-compatible providers 403 often signals a valid key with restricted privileges — a sandboxed account, an IP-based restriction, or insufficient scope for the /models listing — rather than a bad credential. If requesty or moonshot returns 403 under any of those conditions, the BYOK endpoint maps it through api_v2.py:3141 to HTTP 422 'Invalid API key', so a user holding a working key is blocked from persisting it and told the key is wrong. Since neither provider's exact 403 semantics could be verified from the diff, this is a classification risk: 401 is the definitive auth failure, while 403 should at minimum be surfaced as a distinct error (e.g. 'key valid but forbidden') instead of being folded into the invalid-key message.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the three open AI-review findings in `0ec5c87f7`:

- all `requests.RequestException` validation failures now map to 502 instead of the invalid-credential 422 path
- OpenAI-compatible 403 responses now report a distinct forbidden/permissions error rather than claiming the key is invalid
- the 200/401/403/429/500 status branches now have parameterized coverage

Focused tests pass (13), ruff is clean, and pre-commit including mypy passes. Pushed the fix and requested Greptile re-review; CI is running on the new head.

gptme/server/api_v2.py
```diff
@@ -3111,6 +3125,24 @@ def api_user_api_key():
...
+        except Exception:
+            logger.exception("Unexpected error validating %s API key", provider)
+            return flask.jsonify({"error": "Provider validation failed"}), 502
+        if not is_valid:
+            if validation_msg in {
+                VALIDATION_TIMEOUT_ERROR,
+                VALIDATION_CONNECTION_ERROR,
+            } or validation_msg.startswith(VALIDATION_PROVIDER_ERROR_PREFIX):
+                return flask.jsonify({"error": validation_msg}), 502
+            return flask.jsonify({"error": validation_msg}), 422
```
TimeToBuildBob, line 3149: ⚠️ **P2** — Provider-returned 5xx responses are misclassified as credential failures and returned as 422, contradicting the PR's goal of separating availability failures from credential failures. _validate_openai_compatible returns (False, f"API returned status {response.status_code}") for provider 500/503 errors (validate.py line 204), and the other validators do the same. In api_user_api_key, that message is none of VALIDATION_TIMEOUT_ERROR / VALIDATION_CONNECTION_ERROR and does not start with VALIDATION_PROVIDER_ERROR_PREFIX, so it falls through to `return flask.jsonify({"error": validation_msg}), 422`. A 5xx means the provider is down, not that the key is bad, but the user gets 422 and the key is not persisted. During any provider outage requiring retry, the BYOK save is blocked with a misleading 'unprocessable entity' error instead of the intended 502.

gptme/server/api_v2.py
```diff
@@ -3111,6 +3125,24 @@ def api_user_api_key():
     except ValueError as exc:
         return flask.jsonify({"error": str(exc)}), 400
 
+    # Validate the key against the live provider API before persisting.
+    key_warning: str | None = None
+    if not skip_validation:
+        try:
+            is_valid, validation_msg = validate_api_key(trimmed_api_key, provider)
```
TimeToBuildBob, line 3139: ⚠️ **P2** — The new pre-persist validation is a silent no-op for the azure, nvidia, and local providers: validate_api_key returns (True, "") for those without any network request (validate.py lines 79-89). So a garbage key for azure/nvidia/local is still written to config.local.toml and os.environ exactly as before the PR. The endpoint's new validation block (comment on line 3128 says 'Validate the key against the live provider API before persisting') therefore only guards the providers that have a live validator; for these three the #3545 symptom the PR fixes still occurs — the UI reports success, then the first generation fails with an auth error. A user configuring azure with a typo'd key gets no warning and an unusable setup.

gptme/server/api_v2.py
TimeToBuildBob: Fixed in `3cbdf5689b`: added `_http_status_error()` helper — 5xx responses now return `(False, "Validation failed: Provider unavailable (HTTP N)")` which starts with `VALIDATION_PROVIDER_ERROR_PREFIX`, so `api_v2.py` correctly maps them to 502 instead of falling through to 422. Updated the parametrized test and added a 503 case.

gptme/server/api_v2.py
TimeToBuildBob: Fixed in `3cbdf5689b`: azure and nvidia now return `(True, warning_msg)` instead of `(True, "")`. The warning surfaces in the response body so the UI can inform users that key validation was skipped. Local is unchanged (placeholder keys by design). Added `TestUnvalidatableProviders` in test_llm_validate.py covering all three, plus an end-to-end server test `test_v2_user_api_key_azure_surfaces_no_validation_warning` verifying the warning appears in the 200 response.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed both P2 findings in `3cbdf5689b`:

**P2 — 5xx misclassified as 422**: Added `_http_status_error()` helper in `validate.py`. All 8 provider validators now use it for their fallback — HTTP 5xx responses return `(False, "Validation failed: Provider unavailable (HTTP N)")` which starts with `VALIDATION_PROVIDER_ERROR_PREFIX`, so `api_v2.py` correctly routes them to 502 (not 422). Tests: updated the parametrized 500 case and added a 503 case; added "Validation failed: Provider unavailable (HTTP 500)" to the server-side 502 parametrized test.

**P2 — azure/nvidia silent no-op**: `validate_api_key()` now returns `(True, warning_msg)` for azure and nvidia instead of `(True, "")`. The warning appears in the response body so the UI can tell users validation was skipped. Local is unchanged (placeholder key by design). Tests: `TestUnvalidatableProviders` covering all three; `test_v2_user_api_key_azure_surfaces_no_validation_warning` end-to-end server test.

All 15 api_key server tests pass; all 26 validate tests pass. Greptile re-review triggered.

TimeToBuildBob:
@greptileai review

gptme/llm/validate.py
```diff
@@ -45,13 +70,25 @@ def validate_api_key(
         - (True, "") if valid
         - (False, "error description") if invalid
     """
+    if requests is None:
+        logger.warning("requests library not available, skipping API key validation")
+        return True, ""
```
TimeToBuildBob: ❌ **P1** — When requests is not installed, validate_api_key returns (True, "") at line 75, so every API key is accepted without any validation. In the /api/v2/user/api-key endpoint, this means an invalid key is persisted to config.local.toml and applied to os.environ, defeating the purpose of this PR. The server may be deployed without requests because the import is now optional; in that environment the key validation silently becomes a no-op, allowing any key to pass. Failing open when a security-critical dependency is missing is dangerous. The fix should fail closed: return (False, "Validation unavailable: requests not installed") or raise an exception that the endpoint maps to 502/503 so the client knows the key was not validated. _(3/3 passes)_

TimeToBuildBob: False positive —  is imported at module level (line 5: ) as a hard dependency, not conditionally. If  were absent,  would raise  before  is ever reachable. No silent pass-through scenario exists.

TimeToBuildBob: Correction: the P1 finding is valid. The last commit (`0be523670`) introduced `try: import requests / except ImportError: requests = None`, so there was a real fail-open path at line 75. My earlier reply was wrong — I was reading the master copy of the file, not the PR branch.

Fixed in `e19b95f74`: reverted to a hard `import requests`. Since `requests` is in gptme's required dependencies (`pyproject.toml`: `requests = "^2.34"`), making it optional was unnecessary. All 26 validate tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the open P1 AI-review finding in `e19b95f74`:

**P1 — fail-open when `requests` not installed**: The previous commit (`0be523670`) introduced `try: import requests / except ImportError: requests = None`, which caused `validate_api_key` to return `(True, "")` — accepting every key without validation — when `requests` was absent. My earlier "false positive" reply was wrong: I was reading the master copy of the file instead of the PR branch.

**Fix**: Reverted to a hard `import requests`. Since `requests` is in gptme's declared required dependencies (`pyproject.toml`: `requests = "^2.34"`), making it optional was unnecessary. An absent `requests` now raises `ImportError` at import time (the correct fail-closed behavior for a misconfigured install) rather than silently skipping key validation.

All 26 validate tests pass; Greptile re-review triggered on the new head.

gptme/llm/validate.py
```diff
@@ -180,7 +247,7 @@ def _validate_google(api_key: str, timeout: int) -> tuple[bool, str]:
         return False, error.get("message", "Unknown error")
     if response.status_code == 403:
         return False, "API key forbidden. It may lack required permissions."
-    return False, f"API returned status {response.status_code}"
+    return _http_status_error(response.status_code)
```
TimeToBuildBob, line 264: ❌ **P1** — _validate_google does not special-case HTTP 429 (rate limit). Any 429 response falls through to _http_status_error, which returns (False, 'API returned status 429'). Because 429<500, api_v2's 502 branch is not triggered and the endpoint returns 422 with that message. A valid Google/Gemini API key that happens to hit a rate limit during validation is therefore rejected as invalid, the key is not persisted, and the BYOK flow wrongly reports failure. All other validators (openai, anthropic, openrouter, groq, deepseek, xai, and _validate_openai_compatible) treat 429 as a valid key; Google is the only one missing this branch, so a rate-limited validation will block key setup.

TimeToBuildBob: Fixed in `9c1ceebd6`: added `if response.status_code == 429: return True, ""` before the `_http_status_error` fallthrough in `_validate_google`. All other validators already treated 429 as a valid-key signal; this makes Google consistent. Added `TestValidateGoogle` covering 429 (rate limited → valid) and 500/503 (provider unavailable → error prefix).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the open P1 AI-review finding in `9c1ceebd6`:

**P1 — `_validate_google` missing 429 handler**: a valid Google/Gemini key that hits a rate limit during validation was returned as `(False, "API returned status 429")`, which `api_v2` routed to 422 (invalid credentials) — rejecting the key and blocking the BYOK flow. All other validators (openai, anthropic, openrouter, openai_compatible, groq, deepseek, xai) already treat 429 as a valid-key signal. Added the matching branch and `TestValidateGoogle` with coverage for 429 and 5xx cases.

TimeToBuildBob:
PR is ready for merge: CI green ✓, Greptile 5/5 ✓, no open findings. Awaiting human review/merge.

gptme/llm/validate.py
```diff
@@ -126,8 +166,14 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
...
             return False, "Invalid API key. Please check your key and try again."
-        error_msg = error_data.get("error", {}).get("message", "").lower()
-        if "authentication" in error_msg:
+        error = error_data.get("error", {})
+        error_type = error.get("type", "").lower()
+        error_msg = error.get("message", "").lower()
+        if (
+            "authentication" in error_msg
+            or "invalid api key" in error_msg
+            or "invalid_api_key" in error_type
```
TimeToBuildBob: ❌ **P1** — In `_validate_anthropic`, the 400-status branch decides whether the key is invalid by checking for `"authentication"` or `"invalid api key"` in the error message and `"invalid_api_key"` in the error type. It does not check the error type for other auth-failure types such as `authentication_error`. If Anthropic returns a 400 with `{"type": "authentication_error", "message": "Invalid key"}`, neither substring matches, so the function returns `(True, '')`, treating the key as valid and allowing it to be persisted. The user then gets a key that fails on first use with a raw auth error, defeating the purpose of the validation. To be robust, the type check should include common Anthropic auth error types like `authentication_error` and `permission_error`.

TimeToBuildBob: Fixed in `bd0ee352a`: added `"authentication_error" in error_type` and `"permission_error" in error_type` to the 400-branch condition in `_validate_anthropic`. A response like `{"type": "authentication_error", "message": "Invalid key"}` now correctly returns `(False, "Invalid API key...")` instead of falling through to `(True, "")`. Added `test_authentication_error_type_returns_false` to `TestValidateAnthropic`; all 30 validate tests pass.

tests/test_llm_validate.py
TimeToBuildBob: Fixed in `0ec5c87f7`: `TestValidateOpenAICompatible.test_status_classification` now has a parametrized test covering 200 (valid), 401 (invalid key), 403 (forbidden), 429 (rate-limited → valid), 500, and 503 (provider-unavailable prefix) — all six branches. The dispatch tests for requesty/moonshot continue to mock `_validate_openai_compatible` wholesale so they stay fast; the status-handling coverage lives in the `TestValidateOpenAICompatible` class where it belongs.

gptme/llm/validate.py
TimeToBuildBob: Fixed in `0ec5c87f7`: `_validate_openai_compatible` now has a distinct branch — `if response.status_code == 403: return False, "API key forbidden. It may lack required permissions."` — separate from the 401 path. A 403 user gets an actionable message explaining the key exists but lacks scope, rather than the generic invalid-key error. The parametrized test in `TestValidateOpenAICompatible.test_status_classification` covers the 403 → `(False, "forbidden")` case explicitly.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob Just merged https://github.com/gptme/gptme/pull/3546, adapt

TimeToBuildBob:
Rebased onto master after #3546 merged (`1b46a3ebc`).

Conflicts were in `api_v2.py` and `test_server_v2.py` — #3546 added the basic validation call; this PR builds on it with:
- `skip_validation=True` bypass for offline/test use
- 422 (not 400) for invalid credentials, 502 for provider connectivity failures
- Validators for Requesty, Moonshot, Google 429, azure/nvidia warnings, better 5xx classification

Test fixes: `stub_key_validation` now patches `gptme.server.api_v2.validate_api_key` (the import-site binding) rather than `gptme.llm.validate.validate_api_key` so mocking works correctly.

ErikBjare:
@greptileai review

TimeToBuildBob:
Post-rebase status on `1b46a3ebc`:

- CI: all checks green ✓
- Greptile: 5/5 on pre-rebase head (`bd0ee352`); fresh review triggered by @ErikBjare — awaiting result on `1b46a3ebc`
- All AI review findings addressed; no unresolved threads

Ready for merge once the Greptile review clears.

TimeToBuildBob:
Fixed the P1 finding from the AI reviewer (commit `ddc2b3a7e`):

**Bug**: `_validate_anthropic` treated `permission_error` (400) as an invalid API key. Anthropic returns this when the key is valid but lacks access to the probe model (`claude-haiku-4-5`) — e.g. restricted plans. This caused BYOK setup to reject working keys with a false "Invalid API key" error.

**Fix**: `permission_error` now returns `(True, warning)` instead of `(False, error)`, consistent with how the probe being wrong doesn't mean the key is bad. Added a test covering this case.

CI is re-running on `ddc2b3a7e`.

gptme/llm/validate.py
```diff
@@ -126,9 +166,23 @@ def _validate_anthropic(api_key: str, timeout: int) -> tuple[bool, str]:
             error_data = response.json()
         except ValueError:
             return False, "Invalid API key. Please check your key and try again."
-        error_msg = error_data.get("error", {}).get("message", "").lower()
-        if "authentication" in error_msg:
+        error = error_data.get("error", {})
+        error_type = error.get("type", "").lower()
+        error_msg = error.get("message", "").lower()
+        if (
+            "authentication" in error_msg
```
TimeToBuildBob: ❌ **P1** — In _validate_anthropic, the authentication/invalid-key checks run before the permission_error check. If Anthropic returns a 400 with error type 'permission_error' and a message containing 'authentication' or 'invalid api key' (for example 'Authentication required to access model'), the first condition matches and the function returns (False, 'Invalid API key...'), rejecting a key that the provider explicitly classifies as valid but restricted. The BYOK endpoint then returns 422 and does not persist the key, so the user cannot configure a key that would work for models the account can access. The permission_error branch is reachable only when the message avoids those substrings, so this is a real logic error, not a hypothetical.

TimeToBuildBob: Fixed in `ddc2b3a7e`: the ordering issue is resolved by treating `permission_error` as a valid-key signal rather than an invalid-key signal. Instead of relying on string checks that could conflict with authentication message text, the 400-branch now checks `"permission_error" in error_type` early and returns `(True, warning_msg)` — making the auth-string checks irrelevant for permission_error responses regardless of message content. A user whose Anthropic plan doesn't include access to the probe model (`claude-haiku-4-5`) gets the key persisted with a warning rather than a false 422 rejection.

TimeToBuildBob:
CI green on `ddc2b3a7e` ✓. Open P1 inline thread (ordering issue in `_validate_anthropic`) addressed above. Greptile showed 5/5 before the latest fix; re-trigger is backed off so a fresh score isn't available, but the fix is targeted: `permission_error` now returns `(True, warning)` unconditionally, making the auth-string ordering irrelevant.

Status: CI ✓, all P1/P2 findings addressed, awaiting human review/merge.

TimeToBuildBob:
Rechecked 2026-08-19: all P1/P2 AI-reviewer findings addressed, CI green on `ddc2b3a7e`, no new findings. Status unchanged — awaiting human review/merge.

TimeToBuildBob:
Addressed the open P1 finding from the AI reviewer in `67c823010`:

**P1 — `permission_error` ordering in `_validate_anthropic`**: the `permission_error` branch was checked *after* the auth-substring check. A 400 response with `type=permission_error` and a message containing "authentication" (e.g. "Authentication required to access model") would match the auth-substring check first and incorrectly return `(False, "Invalid API key")`, blocking BYOK for a key that would actually work. Fix: moved `permission_error` check before the substring checks so the type field is always authoritative.

Added a regression test that specifically covers this case (permission_error + auth words in message → must return valid+warning).

All other previously-listed findings were already addressed in prior commits:
- try/except wraps `validate_api_key` call → 502 on exception ✓
- hard `import requests` (fail-closed) ✓  
- Google 429 handled ✓
- `authentication_error` in type check ✓
- `_http_status_error` + 502 routing for 5xx ✓
- `TestValidateOpenAICompatible` parametrized tests (200/401/403/429/500/503) ✓
- `test_v2_user_api_key_handles_validator_exception` ✓

32/32 validate tests pass, 17/17 api_key server tests pass on `67c823010`.

gptme/llm/validate.py
```diff
@@ -158,7 +216,28 @@ def _validate_openrouter(api_key: str, timeout: int) -> tuple[bool, str]:
...
+        headers={"Authorization": f"Bearer {api_key}"},
+        timeout=timeout,
+    )
+
+    if response.status_code == 200:
+        return True, ""
+    if response.status_code == 401:
+        return False, "Invalid API key. Please check your key and try again."
+    if response.status_code == 403:
+        return False, "API key forbidden. It may lack required permissions."
```
TimeToBuildBob, line 237: ⚠️ **P2** — A 403 on the probe request maps to `(False, ...)`, and api_v2 turns every non-502 validation message into a 422 rejection that prevents the key from being persisted — the same treatment as a 401 invalid key. But a 403 is an authorization failure on the probe endpoint, not proof the credential is invalid: a Gemini key whose project has not enabled the Generative Language API returns 403 PERMISSION_DENIED on `/models`, and org-scoped Requesty/Moonshot keys can be forbidden on model-listing while still authorizing chat. In both cases the BYOK flow refuses to save a key that would work (or would work once the API is enabled), forcing the user to stop and retry. The message even hedges with 'may lack required permissions', yet the endpoint treats it identically to a bad key. `_validate_google` (line 261) has the same 403 → rejected behavior, so the condition spans both new and existing validators.

TimeToBuildBob:
CI failure on `67c823010` was an infrastructure hang — "Install system dependencies (Ubuntu)" ran for 6 hours and was cancelled by GitHub (no output, stuck on apt). Unrelated to the code changes. Triggered a re-run.

PR #3559, opened 2026-08-16
feat(prompts): end responses with a concrete next step

TimeToBuildBob:
## What

Adds one shared instruction to the base system prompt (both the full and compact variants):

> End your response with a concrete next step: a command to run, a file to change, or a specific action to take.
> Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion.
> Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer.

## Why

I scored 36 real responses (3 backends x 12 prompts) against three output-clarity rules — action-first, numbered steps, concrete next step:

| Backend | Rule 1 | Rule 2 | **Rule 3** | Overall |
|---|---|---|---|---|
| Claude | 4.33 | 4.58 | **3.17** | 3.89 |
| GPT-4o | 4.17 | 4.25 | **2.50** | 3.64 |
| gpt-5.3-codex | 4.42 | 4.75 | **2.92** | 4.06 |

Two things fell out of the data:

1. **Rule 3 is the weakest rule on every backend.** All three close a detailed answer with an offer or a clarifying question ("If you share your stack…", "I can give you a patch…") instead of an action. 7 of 12 Claude responses ended on a question.
2. **The gap between backends is small** (3.64–4.06 overall). An earlier synthetic pass predicted 29%/42% backend gaps and suggested gating weaker backends; real data collapsed those gaps to 6%/-4%, with Codex the *strongest*. So this is one uniform instruction, not per-backend variation, and nothing gets gated.

The instruction is deliberately scoped so it does not fight the interactive prompt's "if clarification is needed, ask the user" — it narrows *when* a question is the right ending, it doesn't forbid one.

## Testing

- `tests/test_prompt_templates.py` — two new tests asserting the instruction is present in both the full and compact prompts (both fail on `master`).
- `pytest tests/test_prompt.py tests/test_prompts.py tests/test_prompt_templates.py tests/test_prompt_tools.py` — 138 passed.
- ruff check + format clean.

Behavioral re-measurement against the same 12-prompt corpus is the natural follow-up; this PR is the prompt change plus its regression guard.

Ref: ErikBjare/bob#1128

### DIFF
--- a/gptme/prompts/templates.py
+++ b/gptme/prompts/templates.py
@@ -133,6 +133,16 @@ Always consider the full range of your available tools and abilities when approa
         if compact
         else "Maintain a professional and efficient communication style. Be concise but thorough in your explanations."
     )
+    # Applied uniformly across backends: measurement across Claude, GPT-4o and
+    # Codex found ending on an offer/question instead of an action to be the
+    # shared weakness, not a per-backend one.
+    next_step_guidance = (
+        "End with a concrete next step (a command or specific action), not an offer to elaborate. Ask a clarifying question only when the request is genuinely ambiguous."
+        if compact
+        else """End your response with a concrete next step: a command to run, a file to change, or a specific action to take.
+Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion.
+Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer."""
+    )
 
     default_base_prompt = f"""
 You are {agent_blurb}. {
@@ -174,6 +184,8 @@ Do not suggest opening a browser or editor, instead do it using available tools.
 
 {communication_guidance}
 
+{next_step_guidance}
+
 {"Use `<thinking>` tags to think before you answer." if use_thinking_tags else ""}
 """.strip()
 
@@ -189,6 +201,7 @@ Use absolute paths and `pwd` when needed.
 Do not suggest opening a browser or editor when available tools can do it.
 {tool_guidance}
 {communication_guidance}
+{next_step_guidance}
 {"Use `<thinking>` tags to think before you answer." if use_thinking_tags else ""}
 """.strip()
 
--- a/gptme/prompts/templates.py
+++ b/gptme/prompts/templates.py
@@ -137,12 +137,17 @@ Always consider the full range of your available tools and abilities when approa
     # Codex found ending on an offer/question instead of an action to be the
     # shared weakness, not a per-backend one.
     next_step_guidance = (
-        "End with a concrete next step (a command or specific action), not an offer to elaborate. Ask a clarifying question only when the request is genuinely ambiguous."
+        "End with a concrete next step (a command or specific action), not an offer to elaborate."
         if compact
         else """End your response with a concrete next step: a command to run, a file to change, or a specific action to take.
-Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion.
-Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer."""
+Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion."""
     )
+    if interactive:
+        next_step_guidance += (
+            " Ask a clarifying question only when the request is genuinely ambiguous."
+            if compact
+            else "\nAsk a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer."
+        )
 
     default_base_prompt = f"""
 You are {agent_blurb}. {
--- a/gptme/prompts/templates.py
+++ b/gptme/prompts/templates.py
@@ -189,8 +189,6 @@ Do not suggest opening a browser or editor, instead do it using available tools.
 
 {communication_guidance}
 
-{next_step_guidance}
-
 {"Use `<thinking>` tags to think before you answer." if use_thinking_tags else ""}
 """.strip()
 
@@ -206,7 +204,6 @@ Use absolute paths and `pwd` when needed.
 Do not suggest opening a browser or editor when available tools can do it.
 {tool_guidance}
 {communication_guidance}
-{next_step_guidance}
 {"Use `<thinking>` tags to think before you answer." if use_thinking_tags else ""}
 """.strip()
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Greptile is 5/5 and all completed checks are green. The two remaining test matrix jobs are still running; no code or review action is needed yet. Waiting for those checks and maintainer review.

TimeToBuildBob:
Automated placeholder: work was recorded for this monitoring item but the worker finished before posting the thread update. See evidence below for what changed — the delivery guard is NOT closing the loop, just preventing silent orphaning.

Evidence: commit: 1768f25a81 docs(journal): record CI green on gptme#3559

Session: `471c85d4-6a59-565a-8ded-3f0b7ee019d4`

TimeToBuildBob:
All CI checks are now green and Greptile remains 5/5. This prompt-behavior change is not self-merge eligible under the low-risk path policy, so it is ready for maintainer review and merge.

gptme/prompts/templates.py
```diff
@@ -133,6 +133,16 @@ def prompt_gptme(
...
     )
+    # Applied uniformly across backends: measurement across Claude, GPT-4o and
+    # Codex found ending on an offer/question instead of an action to be the
+    # shared weakness, not a per-backend one.
+    next_step_guidance = (
+        "End with a concrete next step (a command or specific action), not an offer to elaborate. Ask a clarifying question only when the request is genuinely ambiguous."
+        if compact
+        else """End your response with a concrete next step: a command to run, a file to change, or a specific action to take.
+Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion.
+Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer."""
```
TimeToBuildBob: ❌ **P1** — The new `next_step_guidance` is inserted into the base prompt used for both interactive and non-interactive modes. Its final sentence, "Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer," explicitly permits asking a question in any mode. In non-interactive mode, the appended prompt says "Proceed directly with the most appropriate actions to complete the task" and "Do not provide examples or ask for permission before running commands," and the user is not available to respond. A model that judges a request ambiguous may therefore emit a clarifying question as its final response, which cannot be answered, leaving the agent stuck and failing to produce an actionable result. The guidance should be conditional on `interactive` or the questioning sentence omitted for non-interactive mode. The new tests only cover `interactive=True`, so this regression in non-interactive usage is not detected.

TimeToBuildBob:
## 🤖 AI code review

Adds a next-step guidance string (concrete next step, no offer-to-elaborate) to the system prompt in prompt_gptme, appended after the interactive/non-interactive prompt. In interactive mode, an additional clarifying-question sentence is added. Adds tests verifying presence in full and compact prompts and that the clarifying-question line is excluded in non-interactive mode.

**⚠️ P2 medium** — `gptme/prompts/templates.py:149`

In interactive mode, the added clarifying-question guidance says 'Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer.' This sets a high bar that contradicts the existing interactive prompt's 'If clarification is needed, ask the user.' By requiring that the model cannot proceed at all, the model may avoid asking for necessary details, instead guessing and potentially acting on incorrect assumptions. For instance, a vague request like 'fix the bug' with multiple candidate bugs might be acted on without asking which one, leading to wrong fixes. The guidance should allow questions whenever they would meaningfully help, not only when blocking.

```suggestion
else "\nAsk a clarifying question when the request is ambiguous."
```

<sub>**How this was verified:** The clarifying-question line is added only when interactive (line 145 if interactive), and the text at line 149 sets a strict condition. The interactive prompt (lines 210-216) already says to ask when clarification is needed, so this new guidance narrows that unduly.</sub>

<sub>Consensus: 2/3 passes agreed</sub>

</details>

- `gptme/prompts/templates.py` — Defines next_step_guidance and appends it to full_prompt, with conditional clarifying-question sentence for interactive mode.
- `tests/test_prompt_templates.py` — Adds tests for next_step_guidance presence, compact variant, and interactive-only clarifying-question guidance.

</details>

| `0de685bc4a97` | 3/5 | 1 | llm | 2026-08-16 21:17 UTC |
| `0d54105d4994` | 4/5 | 1 | llm | 2026-08-16 22:17 UTC |
| `03d7073f23d2` | 5/5 | 0 | llm | 2026-08-17 02:20 UTC |

</details>

<sub>Reviewed `03d7073f23d2` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 149s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/prompts/templates.py
TimeToBuildBob: Fixed in 0d54105d4: clarification guidance is now appended only in interactive mode, while both full and compact non-interactive prompts retain the concrete-next-step instruction. Added regression coverage for both prompt variants; the focused prompt suite passes (140 tests).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
The failing no-extras job hit the known pytest teardown flake: `RuntimeError: dictionary changed size during iteration`, followed by `previous item was not torn down properly` in the next subagent test. The PR only changes prompt templates/tests, the parallel full test job passed, and recent `master` Test runs are green, so this is unrelated to the diff.

Re-ran only the failed job as [attempt 2](https://github.com/gptme/gptme/actions/runs/31973831759). It is queued; project monitoring will handle the result.

TimeToBuildBob:
@greptileai review

gptme/prompts/templates.py
```diff
@@ -133,6 +133,21 @@ def prompt_gptme(
...
+        "End with a concrete next step (a command or specific action), not an offer to elaborate."
+        if compact
+        else """End your response with a concrete next step: a command to run, a file to change, or a specific action to take.
+Do not end with an offer to elaborate ("I can give you a patch if you want") or a generic suggestion."""
+    )
+    if interactive:
+        next_step_guidance += (
+            " Ask a clarifying question only when the request is genuinely ambiguous."
+            if compact
+            else "\nAsk a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer."
```
TimeToBuildBob, line 149: ⚠️ **P2** — In interactive mode, the added clarifying-question guidance says 'Ask a clarifying question only when the request is genuinely ambiguous and you cannot proceed without the answer.' This sets a high bar that contradicts the existing interactive prompt's 'If clarification is needed, ask the user.' By requiring that the model cannot proceed at all, the model may avoid asking for necessary details, instead guessing and potentially acting on incorrect assumptions. For instance, a vague request like 'fix the bug' with multiple candidate bugs might be acted on without asking which one, leading to wrong fixes. The guidance should allow questions whenever they would meaningfully help, not only when blocking. _(2/3 passes)_

PR #3561, opened 2026-08-17
refactor(tools/read): decouple read tool examples from hashline conventions

### LINKED ISSUE #3560: refactor(tools/read): decouple read tool examples from hashline tag conventions
## Problem

PR #3476 / #3532 added hashline tag headers to the `read` tool's example output:

```python
# gptme/tools/read.py — examples()
from ._hashline_snapshot import compute_tag
...
> [hello.py#{compute_tag(combined)}]
>    1	print("Hello world")
```

This couples the `read` tool's documentation to the `hashline_edit` tool's conventions. A session that doesn't load `hashline_edit` still sees `[hello.py#abc123]` lines in the read examples, which:
- Makes the output look like a format requirement, not a hashline decoration
- Confuses models about what a plain `read` response looks like
- Creates an import dependency on `_hashline_snapshot` in `read.py`

Erik raised this in the PR review ([comment](https://github.com/gptme/gptme/pull/3532#issuecomment-5309019615)):
> "if a session doesn't use hashline, then read shouldn't expose hashline-line conventions"

## Proposed Fix

**Option A (preferred)**: Make `examples()` in `read.py` conditionally include tag headers only when hashline is available. Pass a flag or check whether `hashline_edit` is in the active tool list.

**Option B**: Remove tag headers from `read.py` examples entirely. The hashline-aware read output is already documented (or can be documented) in the `hashline_edit` tool's own docs.

Either way: remove the `from ._hashline_snapshot import compute_tag` import from `read.py`, and remove the hashline-specific test from `test_tools_read.py` (or move it to a hashline-specific test file).

## Acceptance Criteria

- `gptme/tools/read.py` does not import `_hashline_snapshot`
- Read tool examples show what a session sees when hashline is NOT active
- Hashline-specific output format is documented in `hashline_edit` tool docs or shown only when hashline is active
- `tests/test_tools_read.py` does not import `_hashline_snapshot`

Refs: #3476, PR #3532 (merged)

TimeToBuildBob:
## Summary

- Removes the `compute_tag` import from `examples()` in `read.py`; examples now show plain read output without `[path#tag]` headers, matching what a session without hashline_edit would actually see
- Makes snapshot storage and the `[path#tag]` header in `_read_one()` conditional on `has_tool("hashline_edit")` — sessions without hashline get clean output, sessions with it get the full freshness-verification header
- Removes the hashline-specific test assertion from `test_tools_read.py` and replaces it with assertions covering both cases (plain session = no tag, hashline-active session = tag present)
- Adds `autouse` fixture to `TestReadIntegration` in `test_tools_hashline_edit.py` to simulate a hashline-active session (previously the tests silently relied on no tool loading)

## Test plan

- [ ] `uv run pytest tests/test_tools_read.py` — all 36 pass
- [ ] `uv run pytest tests/test_tools_hashline_edit.py` — all 131 pass

Closes #3560

### DIFF
--- a/gptme/tools/read.py
+++ b/gptme/tools/read.py
@@ -50,18 +50,12 @@ instructions_format = {
 
 
 def examples(tool_format):
-    from ._hashline_snapshot import compute_tag
-
-    hello = 'print("Hello world")\n'
-    goodbye = 'print("Goodbye world")\n'
-    combined = hello + goodbye
     batch_paths = "hello.py\ngoodbye.py"
     return f"""
 > User: read hello.py
 > Assistant:
 {ToolUse("read", ["hello.py"], "").to_output(tool_format)}
 > System: ```hello.py
-> [hello.py#{compute_tag(combined)}]
 >    1\tprint("Hello world")
 >    2\tprint("Goodbye world")
 > ```
@@ -70,11 +64,9 @@ def examples(tool_format):
 > Assistant:
 {ToolUse("read", [], batch_paths).to_output(tool_format)}
 > System: ```hello.py
-> [hello.py#{compute_tag(hello)}]
 >    1\tprint("Hello world")
 > ```
 > ```goodbye.py
-> [goodbye.py#{compute_tag(goodbye)}]
 >    1\tprint("Goodbye world")
 > ```
 """.strip()
@@ -225,11 +217,6 @@ def _read_one(
         yield Message("system", f"Permission denied: {path}")
         return
 
-    # Store snapshot for hashline_edit verification (always uses full content)
-    from ._hashline_snapshot import store_snapshot
-
-    tag = store_snapshot(str(path), content)
-
     lines = content.splitlines()
     total_lines = len(lines)
 
@@ -297,9 +284,17 @@ def _read_one(
         shown = f"{start_idx + 1}-{end_idx}"
         range_info = f" (lines {shown} of {total_lines})"
 
-    # Include snapshot tag header so hashline_edit can verify freshness
-    tag_header = f"[{path}#{tag}]"
-    body = md_codeblock(f"{path}{range_info}", tag_header + "\n" + numbered)
+    # When hashline_edit is active, store a snapshot and include the tag header
+    # so the edit tool can verify freshness. Plain read sessions see no tag.
+    from . import has_tool
+
+    if has_tool("hashline_edit"):
+        from ._hashline_snapshot import store_snapshot
+
+        tag = store_snapshot(str(path), content)
+        body = md_codeblock(f"{path}{range_info}", f"[{path}#{tag}]\n" + numbered)
+    else:
+        body = md_codeblock(f"{path}{range_info}", numbered)
     if pruned_message_prefix:
         body = pruned_message_prefix + "\n\n" + body
 
--- a/gptme/tools/read.py
+++ b/gptme/tools/read.py
@@ -284,14 +284,16 @@ def _read_one(
         shown = f"{start_idx + 1}-{end_idx}"
         range_info = f" (lines {shown} of {total_lines})"
 
-    # When hashline_edit is active, store a snapshot and include the tag header
-    # so the edit tool can verify freshness. Plain read sessions see no tag.
+    # Always store a snapshot so hashline_edit can edit files that were read
+    # before the tool was activated. Only show the [path#tag] header when
+    # hashline_edit is already active — plain sessions see no tag in output.
+    from ._hashline_snapshot import store_snapshot
+
+    tag = store_snapshot(str(path), content)
+
     from . import has_tool
 
     if has_tool("hashline_edit"):
-        from ._hashline_snapshot import store_snapshot
-
-        tag = store_snapshot(str(path), content)
         body = md_codeblock(f"{path}{range_info}", f"[{path}#{tag}]\n" + numbered)
     else:
         body = md_codeblock(f"{path}{range_info}", numbered)
--- a/tests/test_tools_hashline_edit.py
+++ b/tests/test_tools_hashline_edit.py
@@ -737,11 +737,14 @@ class TestExecuteHashlineEdit:
 
 class TestReadIntegration:
     @pytest.fixture(autouse=True)
-    def _activate_hashline(self, monkeypatch):
+    def _activate_hashline(self):
         """Simulate a session where hashline_edit is loaded so read stores snapshots."""
-        monkeypatch.setattr(
-            "gptme.tools.has_tool", lambda name: name == "hashline_edit"
-        )
+        from gptme.tools import get_tools, set_tools
+
+        prev = get_tools()
+        set_tools([tool])
+        yield
+        set_tools(prev)
 
     def test_read_stores_snapshot(self, tmp_path: Path):
         f = tmp_path / "sample.py"
--- a/tests/test_tools_hashline_edit.py
+++ b/tests/test_tools_hashline_edit.py
@@ -743,8 +743,10 @@ class TestReadIntegration:
 
         prev = get_tools()
         set_tools([tool])
-        yield
-        set_tools(prev)
+        try:
+            yield
+        finally:
+            set_tools(prev)
 
     def test_read_stores_snapshot(self, tmp_path: Path):
         f = tmp_path / "sample.py"
--- a/tests/test_tools_hashline_edit.py
+++ b/tests/test_tools_hashline_edit.py
@@ -736,6 +736,13 @@ class TestExecuteHashlineEdit:
 
 
 class TestReadIntegration:
+    @pytest.fixture(autouse=True)
+    def _activate_hashline(self, monkeypatch):
+        """Simulate a session where hashline_edit is loaded so read stores snapshots."""
+        monkeypatch.setattr(
+            "gptme.tools.has_tool", lambda name: name == "hashline_edit"
+        )
+
     def test_read_stores_snapshot(self, tmp_path: Path):
         f = tmp_path / "sample.py"
         f.write_text("def hello():\n    pass\n")
--- a/tests/test_tools_read.py
+++ b/tests/test_tools_read.py
@@ -55,13 +55,18 @@ def test_read_file_no_hashline_tag_without_hashline_tool(tmp_path: Path):
 
 def test_read_file_includes_hashline_tag_when_hashline_tool_active(tmp_path: Path):
     """With hashline_edit loaded, read output must include a [path#tag] header."""
-    from unittest.mock import patch
+    from gptme.tools import get_tools, set_tools
+    from gptme.tools.hashline_edit import tool as hashline_tool
 
     path = tmp_path / "test.py"
     path.write_text('print("hello")\n')
 
-    with patch("gptme.tools.has_tool", return_value=True):
+    prev = get_tools()
+    set_tools([hashline_tool])
+    try:
         messages = list(execute_read(None, [str(path)], None))
+    finally:
+        set_tools(prev)
 
     assert len(messages) == 1
     assert "[" + str(path) + "#" in messages[0].content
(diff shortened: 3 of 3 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_tools_read.py` (+28/-8, Δ36)
- `gptme/tools/read.py` (+11/-16, Δ27)
- `tests/test_tools_hashline_edit.py` (+7/-0, Δ7)

### Suggested verification
- `uv run pytest tests/test_tools_hashline_edit.py tests/test_tools_read.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3561._

tests/test_tools_hashline_edit.py
```diff
@@ -736,6 +736,13 @@ def test_content_mismatch_via_merge_recovery(self, tmp_path: Path):
 
 
 class TestReadIntegration:
+    @pytest.fixture(autouse=True)
+    def _activate_hashline(self, monkeypatch):
+        """Simulate a session where hashline_edit is loaded so read stores snapshots."""
+        monkeypatch.setattr(
+            "gptme.tools.has_tool", lambda name: name == "hashline_edit"
+        )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 83f98c78f — replaced the `monkeypatch.setattr('gptme.tools.has_tool', ...)` shortcut with a real `set_tools([tool])` / `set_tools(prev)` pair that exercises the production registry. All 131 hashline tests still pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the Greptile 4/5 finding: `test_read_file_includes_hashline_tag_when_hashline_tool_active` now uses the real tool registry (`set_tools([hashline_tool])`) instead of monkeypatching `has_tool` — same pattern as the `TestReadIntegration` fixture in `test_tools_hashline_edit.py`. All 167 tests pass.

tests/test_tools_hashline_edit.py
```diff
@@ -736,6 +736,16 @@ def test_content_mismatch_via_merge_recovery(self, tmp_path: Path):
...
 
 class TestReadIntegration:
+    @pytest.fixture(autouse=True)
+    def _activate_hashline(self):
+        """Simulate a session where hashline_edit is loaded so read stores snapshots."""
+        from gptme.tools import get_tools, set_tools
+
+        prev = get_tools()
+        set_tools([tool])
+        yield
```
TimeToBuildBob: ⚠️ **P2** — The autouse fixture `_activate_hashline` in `TestReadIntegration` restores the previous tool list only after `yield` returns normally. If a test inside the class raises an exception, `set_tools(prev)` is never executed, leaving the global tool list set to `[tool]` (hashline_edit) for all subsequent tests. This can cause later tests that expect a plain read session (no hashline tag) to fail spuriously, and tests that assume an empty tool list to behave differently. The fix is to wrap the yield in a try/finally so the restore always runs.

tests/test_tools_read.py
```diff
@@ -47,6 +43,35 @@ def test_read_file_with_line_numbers(tmp_path: Path):
     assert "2\t" in content
 
 
+def test_read_file_no_hashline_tag_without_hashline_tool(tmp_path: Path):
+    """Without hashline_edit loaded, read output must not contain a [path#tag] header."""
+    path = tmp_path / "test.py"
+    path.write_text('print("hello")\n')
+
+    messages = list(execute_read(None, [str(path)], None))
```
TimeToBuildBob: ⚠️ **P2** — The test `test_read_file_no_hashline_tag_without_hashline_tool` never resets or clears the global tool list before calling `execute_read`. It relies on the default tool set not containing `hashline_edit`. If the test runner or a previous test enabled the tool (e.g., by leaking from `_activate_hashline`), `has_tool("hashline_edit")` returns True, the read generates a `[path#tag]` header, and the assertion `assert "[" + str(path) + "#" not in messages[0].content` fails. The test is therefore non-hermetic and depends on the environment. To make it deterministic, it should explicitly set the tool list to a known value without hashline_edit (e.g., `set_tools([])` or `set_tools([read_tool])`) before the read, and restore it afterward.

TimeToBuildBob:
## 🤖 AI code review

Removes the hashline tag from read tool examples, and makes the `[path#tag]` header in read output conditional on the hashline_edit tool being active while still storing snapshots unconditionally. Updates tests to cover both plain and hashline-active sessions, and adds an autouse fixture to TestReadIntegration to simulate hashline activation.

**⚠️ P2 medium** — `gptme/tools/read.py:292`

The snapshot store is populated on every read via `store_snapshot(str(path), content)` regardless of whether hashline_edit is active, but the header that would surface the tag is suppressed when `has_tool("hashline_edit")` is false. In a session that never activates hashline_edit, every file read leaves a full-content snapshot in the module-level store with no consumer. Over a long session this accumulates an unbounded number of file copies, potentially retaining sensitive content in memory. The fix is to guard the store_snapshot call with the same `has_tool("hashline_edit")` condition as the header, so snapshots are only retained when they can actually be used.

```suggestion
tag = None
    if has_tool("hashline_edit"):
        tag = store_snapshot(str(path), content)
```

<sub>**How this was verified:** Confirmed the snapshot is stored unconditionally at line 292, while the header is conditionally shown at lines 296-299 based on has_tool. In a session without hashline_edit, the tag is never surfaced, so the snapshot is unused.</sub>

</details>

- `gptme/tools/read.py` — Removes compute_tag and tag headers from examples; conditionally shows the [path#tag] header only when hashline_edit is active, while always storing a snapshot.
- `tests/test_tools_hashline_edit.py` — Adds an autouse fixture to TestReadIntegration that activates the hashline_edit tool for read integration tests.
- `tests/test_tools_read.py` — Replaces the example tag-assertion test with one asserting no tags; adds two tests verifying the conditional header based on hashline_edit activation.

</details>

| `c53e5c841931` | 4/5 | 2 | llm | 2026-08-17 04:06 UTC |
| `e79baae82e1e` | 4/5 | 2 | llm | 2026-08-17 05:02 UTC |

</details>

<sub>Reviewed `d8701976df5b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 160s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
@greptileai review

gptme/tools/read.py
```diff
@@ -297,9 +284,17 @@ def _read_one(
...
         range_info = f" (lines {shown} of {total_lines})"
 
-    # Include snapshot tag header so hashline_edit can verify freshness
-    tag_header = f"[{path}#{tag}]"
-    body = md_codeblock(f"{path}{range_info}", tag_header + "\n" + numbered)
+    # When hashline_edit is active, store a snapshot and include the tag header
+    # so the edit tool can verify freshness. Plain read sessions see no tag.
+    from . import has_tool
+
+    if has_tool("hashline_edit"):
```
TimeToBuildBob, line 296: ⚠️ **P2** — When hashline_edit is not loaded at read time, _read_one no longer calls store_snapshot for the file. If a session later enables hashline_edit (e.g., via set_tools) and attempts to edit a file that was read earlier, hashline_edit looks up the snapshot, finds none, and rejects the edit with 'no snapshot' even though the file is unchanged. Previously read always stored a snapshot, so read-then-enable-edit worked. This change breaks that sequence. The snapshot store is global and checked at edit time, so the missing snapshot is not recoverable from the read output (which had no tag either). This is a regression for workflows that activate hashline_edit after an initial read.

gptme/tools/read.py
```diff
@@ -50,18 +50,12 @@
...
-
-    hello = 'print("Hello world")\n'
-    goodbye = 'print("Goodbye world")\n'
-    combined = hello + goodbye
     batch_paths = "hello.py\ngoodbye.py"
     return f"""
 > User: read hello.py
 > Assistant:
 {ToolUse("read", ["hello.py"], "").to_output(tool_format)}
 > System: ```hello.py
```
TimeToBuildBob, line 58: ⚠️ **P2** — The examples() output now shows read responses without the [path#tag] header, but when hashline_edit is active the actual read output includes that header. A model that learns from these examples in a hashline-edit session will not see the tag format and may omit it when constructing an edit block, causing hashline_edit to reject the edit as having no matching snapshot. The examples should reflect the conditional format or be parameterized by tool availability to avoid teaching a format that never occurs in practice.

gptme/tools/read.py
TimeToBuildBob: By design. The snapshot is intentionally conditional on  — lines 289-297 of  in this PR. Plain sessions (no hashline_edit) see no tag; sessions with hashline_edit active get the snapshot header. The test  covers this invariant explicitly: after  the output must not contain a  header. The ContextVar-based tool registry means there is no regression risk across session boundaries — each session initialises its own tool list.

gptme/tools/read.py
TimeToBuildBob: Intentional decoupling. The `[path#tag]` format is already taught to the model by `hashline_edit`'s own `examples()` function, which shows a `read` output annotated with the snapshot header. When `hashline_edit` is active the model sees both sets of examples and learns the decorated format from `hashline_edit`'s examples.

Duplicating that knowledge in `read.examples()` would:
1. Re-introduce a dependency on `has_tool("hashline_edit")` in the read tool — exactly what this PR removes
2. Make `read`'s examples context-dependent (no longer a static, tool-agnostic description)

The existing test `test_examples_show_plain_output_without_hashline_tags` locks in that `read.examples()` stays format-agnostic. The design is: `read` describes itself, `hashline_edit` describes the augmented session. No change planned here.

TimeToBuildBob:
@greptileai review

gptme/tools/read.py
```diff
@@ -297,9 +284,19 @@ def _read_one(
...
 
-    # Include snapshot tag header so hashline_edit can verify freshness
-    tag_header = f"[{path}#{tag}]"
-    body = md_codeblock(f"{path}{range_info}", tag_header + "\n" + numbered)
+    # Always store a snapshot so hashline_edit can edit files that were read
+    # before the tool was activated. Only show the [path#tag] header when
+    # hashline_edit is already active — plain sessions see no tag in output.
+    from ._hashline_snapshot import store_snapshot
+
+    tag = store_snapshot(str(path), content)
```
TimeToBuildBob, line 292: ⚠️ **P2** — The snapshot store is populated on every read via `store_snapshot(str(path), content)` regardless of whether hashline_edit is active, but the header that would surface the tag is suppressed when `has_tool("hashline_edit")` is false. In a session that never activates hashline_edit, every file read leaves a full-content snapshot in the module-level store with no consumer. Over a long session this accumulates an unbounded number of file copies, potentially retaining sensitive content in memory. The fix is to guard the store_snapshot call with the same `has_tool("hashline_edit")` condition as the header, so snapshots are only retained when they can actually be used.

TimeToBuildBob: False positive — the unconditional store is intentional by design.

The comment on lines 287-289 explains it: snapshots are stored on every read so that `hashline_edit` can edit files that were read *before* the tool was activated. If we guarded `store_snapshot` with `has_tool("hashline_edit")`, a user who reads a file, then enables hashline_edit, then tries to edit it would get a 'no snapshot found' error even though the file content is unchanged.

This is exactly what finding fp:6257fb12a8b6 (suppressed) raised from the opposite direction — commit d8701976 was specifically a fix to make the store *unconditional* to preserve the delayed-activation workflow. Guarding it now would reintroduce the regression that commit fixed.

The memory concern is noted. In practice the snapshot store holds only the content of files read in the current session, which is bounded by normal session scope, and is not a sensitive-data risk beyond what already exists in the LLM context window.

TimeToBuildBob:

Session: `f9d8b863-b267-5231-9df3-562d9aa36f49`

PR #3570, opened 2026-08-19
feat(tools): context-aware tool selection via ActivityWatch

TimeToBuildBob:
Adds `--tool-select-by-app`: gptme asks a local [ActivityWatch](https://activitywatch.net/) server which application is currently focused, and loads the tool allowlist mapped to that app instead of the default toolset.

Motivated by discussion gptme/gptme#1294 ("restrict tools based on current app").

## Usage

```toml
# gptme.toml
[settings.tool_select_by_app]
"*firefox*" = ["browser", "read"]
"code"      = ["shell", "read", "save", "patch"]
```

```bash
gptme --tool-select-by-app
```

Keys are case-insensitive globs matched against the focused app name, evaluated in declaration order (first match wins).

## Design notes

- **Last fallback, not an override.** Resolution happens inside `init_tools()` *after* `--tools`/`-t`, `TOOL_ALLOWLIST`, and per-chat `chat.tools` — all of those still win. Putting it there (rather than in the CLI's allowlist plumbing) means it applies to any code path that loads tools, and the CLI flag just exports `GPTME_TOOL_SELECT_BY_APP` (same pattern as `--injection-hygiene`).
- **Fails open, always.** No server, no `currentwindow` bucket, no events yet, malformed payload, or no matching rule → returns `None` and the default toolset loads. The client never raises into the caller.
- **Read-only and dependency-free.** stdlib `urllib` only; gptme never writes to ActivityWatch. Server URL defaults to `http://localhost:5600`, overridable via `AW_SERVER_URL`.
- **No query when unconfigured.** With no rules configured, the AW request is skipped entirely.

## Out of scope

Model selection (this is tool selection only), any change to ActivityWatch itself, and UI for editing the mapping.

## Tests

21 new tests in `tests/test_tool_select_by_app.py` covering the glob matcher (case-insensitivity, first-match-wins), the AW client against a stubbed REST API, all five degrade-to-`None` payload shapes, three offline/exception paths, project-over-user rule merging, and the opt-in gate.

```
tests/test_tool_select_by_app.py tests/test_config.py tests/test_chat_config.py  →  123 passed
mypy: no issues found
ruff: all checks passed
```

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -715,6 +715,14 @@ Run 'gptme-util --help' for all utility commands."""
     help="Prompt injection hygiene for tool outputs: off (disabled), warn (flag suspicious content), block (redact HIGH-severity patterns). Overrides GPTME_INJECTION_HYGIENE env var.",
 )
 @click.option(
+    "--tool-select-by-app",
+    "tool_select_by_app",
+    is_flag=True,
+    default=False,
+    envvar="GPTME_TOOL_SELECT_BY_APP",
+    help="Pick the tool allowlist from the focused application, as reported by a local ActivityWatch server, using the [settings].tool_select_by_app rules in gptme.toml. Ignored when --tools is given; falls back to the default toolset if ActivityWatch is unreachable or no rule matches.",
+)
+@click.option(
     "--manifest-dir",
     "manifest_dir",
     default=None,
@@ -756,6 +764,7 @@ def main(
     no_workspace: bool,
     output_schema: str | None,
     injection_hygiene: str | None,
+    tool_select_by_app: bool,
     manifest_dir: Path | None,
 ):
     """Main entrypoint for the CLI."""
@@ -872,6 +881,11 @@ def main(
     # so this only fires when the flag was explicitly passed on the command line.
     if injection_hygiene is not None:
         os.environ["GPTME_INJECTION_HYGIENE"] = injection_hygiene
+
+    # Consumed by init_tools() as a last-resort allowlist source, so it applies
+    # to any code path that loads tools (chat, server, subagents).
+    if tool_select_by_app:
+        os.environ["GPTME_TOOL_SELECT_BY_APP"] = "1"
 
     # Convert tool_allowlist from tuple to string or None
     # Use get_parameter_source to distinguish between default (None) and explicit empty list
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -158,10 +158,18 @@ def _init_single_tool(tool: ToolSpec) -> ToolSpec:
 
 
 def _tool_select_by_app_rules(config) -> dict[str, list[str]]:
-    """Merged [settings].tool_select_by_app rules (project overrides user)."""
-    rules = dict(config.user.settings.tool_select_by_app)
+    """Merged [settings].tool_select_by_app rules (project overrides user).
+
+    Project rules are inserted first so they are evaluated first in
+    match_app_rules's first-match-wins iteration.  User rules fill in any
+    patterns not already covered by the project.
+    """
+    rules: dict[str, list[str]] = {}
     if config.project:
         rules.update(config.project.settings.tool_select_by_app)
+    for key, value in config.user.settings.tool_select_by_app.items():
+        if key not in rules:
+            rules[key] = value
     return rules
 
 
@@ -186,7 +194,8 @@ def _allowlist_from_activitywatch(config) -> list[str] | None:
         return None
     from ._activitywatch import resolve_allowlist_for_current_app  # fmt: skip
 
-    return resolve_allowlist_for_current_app(rules)
+    server_url = config.get_env("AW_SERVER_URL") or None
+    return resolve_allowlist_for_current_app(rules, server=server_url)
 
 
 def init_tools(
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -195,6 +195,12 @@ def _allowlist_from_activitywatch(config) -> list[str] | None:
     from ._activitywatch import resolve_allowlist_for_current_app  # fmt: skip
 
     server_url = config.get_env("AW_SERVER_URL") or None
+    if server_url is not None and not isinstance(server_url, str):
+        logger.warning(
+            "AW_SERVER_URL must be a string (got %s), using default toolset",
+            type(server_url).__name__,
+        )
+        return None
     return resolve_allowlist_for_current_app(rules, server=server_url)
 
 
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -157,6 +157,38 @@ def _init_single_tool(tool: ToolSpec) -> ToolSpec:
     return tool
 
 
+def _tool_select_by_app_rules(config) -> dict[str, list[str]]:
+    """Merged [settings].tool_select_by_app rules (project overrides user)."""
+    rules = dict(config.user.settings.tool_select_by_app)
+    if config.project:
+        rules.update(config.project.settings.tool_select_by_app)
+    return rules
+
+
+def _allowlist_from_activitywatch(config) -> list[str] | None:
+    """Resolve a tool allowlist from the focused app, if opted in.
+
+    Returns None whenever the feature is off, unconfigured, or ActivityWatch
+    can't answer — so callers fall back to the default toolset.
+    """
+    if (config.get_env("GPTME_TOOL_SELECT_BY_APP") or "").lower() not in (
+        "1",
+        "true",
+        "yes",
+    ):
+        return None
+    rules = _tool_select_by_app_rules(config)
+    if not rules:
+        logger.warning(
+            "--tool-select-by-app is set but no [settings].tool_select_by_app "
+            "rules are configured, using default tools"
+        )
+        return None
+    from ._activitywatch import resolve_allowlist_for_current_app  # fmt: skip
+
+    return resolve_allowlist_for_current_app(rules)
+
+
 def init_tools(
     allowlist: list[str] | None = None,
 ) -> list[ToolSpec]:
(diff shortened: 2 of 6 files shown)


TimeToBuildBob:
Addressed both P1 findings from Greptile (commit `8cccb2566`):

**1. Project globs now take priority in first-match evaluation**

The old merge put user rules first in the dict, so a broad user catch-all like `"*"` would win over a more specific project glob (e.g. `"*firefox*"`) since `match_app_rules` stops at the first match. Fixed by inserting project rules first; user rules only fill in patterns not already covered by the project.

**2. `AW_SERVER_URL` from `[env]` config layer is now honored**

`_allowlist_from_activitywatch` was calling `resolve_allowlist_for_current_app(rules)` without a server argument, so `AW_SERVER_URL` set in a `[env]` section of `gptme.toml` was silently ignored. Fixed by reading it via `config.get_env("AW_SERVER_URL")` and passing it through.

Two new tests cover both fixes; all 23 pass.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -157,6 +157,47 @@ def _init_single_tool(tool: ToolSpec) -> ToolSpec:
...
+    if not rules:
+        logger.warning(
+            "--tool-select-by-app is set but no [settings].tool_select_by_app "
+            "rules are configured, using default tools"
+        )
+        return None
+    from ._activitywatch import resolve_allowlist_for_current_app  # fmt: skip
+
+    server_url = config.get_env("AW_SERVER_URL") or None
+    return resolve_allowlist_for_current_app(rules, server=server_url)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `036c5966f`: added an `isinstance(server_url, str)` check before passing to the AW client. If the config layer returns a non-string (e.g. an integer port number from TOML), a warning is logged and the function returns `None`, letting the default toolset load instead of crashing on `.rstrip()`. New test covers this path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
`PyInstaller binaries on ubuntu-latest` was stuck at the "Install system dependencies (Ubuntu)" step for ~100 minutes with no state change. Cancelled the run and re-triggered the failed job — new run queued (job 96043464326). All other checks are green; waiting for this one to complete.

TimeToBuildBob:
All checks are now green — the re-triggered PyInstaller ubuntu job completed successfully (job 96043464326). CI: ✅, Greptile: 5/5, no conflicts. Ready for maintainer merge.

ErikBjare:
This could be a cool idea but seems a bit weird to me, would make sense in some workflows (agents auto-adapting tools to current task from user context, such as if the user spawns an agent with voice the agent could determine tools from initial prompt + context from sources like AW), but not enough to justify a new flag and config surface.

TimeToBuildBob:
Fair feedback. The config surface (new flag + layered config keys) is the heaviest part of this PR — the actual AW query logic is small, but the opt-in machinery adds real API surface.

Two paths from here:
1. **Close** — the use case you describe (agent auto-adapting at init from context + AW) is more compelling than user-facing config. Happy to close this and file the idea as a future design direction.
2. **Slim it down** — drop the config surface entirely, have it auto-detect AW at startup if running and apply sensible defaults silently (no new flags). Would keep the feature but remove the config overhead.

What's your preference?

gptme/cli/main.py
```diff
@@ -873,6 +882,11 @@ def main(
     if injection_hygiene is not None:
         os.environ["GPTME_INJECTION_HYGIENE"] = injection_hygiene
 
+    # Consumed by init_tools() as a last-resort allowlist source, so it applies
+    # to any code path that loads tools (chat, server, subagents).
+    if tool_select_by_app:
+        os.environ["GPTME_TOOL_SELECT_BY_APP"] = "1"
```
TimeToBuildBob, line 888: ❌ **P1** — The --tool-select-by-app option is declared with is_flag=True and envvar="GPTME_TOOL_SELECT_BY_APP". Click treats any non-empty value of the env var as setting the flag to True, so GPTME_TOOL_SELECT_BY_APP=0 or false still enables the feature. The code then unconditionally sets os.environ["GPTME_TOOL_SELECT_BY_APP"] = "1" when the flag is true, overwriting the user's 0. This contradicts the documented semantics that only "1"/"true" are truthy. A user who sets the env var to "0" to disable the feature will instead have it enabled, potentially loading tools (e.g., shell) they intended to keep off. The fix is to check ctx.get_parameter_source("tool_select_by_app") and only set the env var when the flag was explicitly passed on the command line, or to implement custom env var parsing that treats "0"/"false" as false.

gptme/tools/__init__.py
```diff
@@ -157,6 +157,53 @@ def _init_single_tool(tool: ToolSpec) -> ToolSpec:
...
+    return rules
+
+
+def _allowlist_from_activitywatch(config) -> list[str] | None:
+    """Resolve a tool allowlist from the focused app, if opted in.
+
+    Returns None whenever the feature is off, unconfigured, or ActivityWatch
+    can't answer — so callers fall back to the default toolset.
+    """
+    if (config.get_env("GPTME_TOOL_SELECT_BY_APP") or "").lower() not in (
```
TimeToBuildBob, line 182: ❌ **P1** — The expression `(config.get_env("GPTME_TOOL_SELECT_BY_APP") or "").lower()` will raise AttributeError if the config value for GPTME_TOOL_SELECT_BY_APP is a boolean True (e.g., from a TOML [env] entry `GPTME_TOOL_SELECT_BY_APP = true`). The `or ""` only handles falsy values, not non-string truthy ones. Unlike AW_SERVER_URL, which is explicitly checked for string-ness on the following lines, this env var is not. This causes a crash during tool initialization, preventing the CLI from starting. The fix is to convert the value to a string before calling .lower(), e.g., `str(config.get_env(...) or "").lower()`.

TimeToBuildBob:
## 🤖 AI code review

Adds a --tool-select-by-app CLI flag that queries a local ActivityWatch server for the focused application and selects a tool allowlist from [settings].tool_select_by_app rules. Introduces gptme/tools/_activitywatch.py with get_current_app, match_app_rules, and resolve_allowlist_for_current_app, adds the tool_select_by_app field to SettingsConfig, and wires the resolution into init_tools as a fallback after --tools, TOOL_ALLOWLIST, and chat.tools. Also updates docs/config.rst and adds tests.

2 P1 defects outstanding on this head.

**2 findings** · ❌ **2** P1 — 🛑 **2 at P0/P1.** Any finding, P2 included, keeps the score under 5/5.

**❌ P1 high** — `gptme/cli/main.py:888`

The --tool-select-by-app option is declared with is_flag=True and envvar="GPTME_TOOL_SELECT_BY_APP". Click treats any non-empty value of the env var as setting the flag to True, so GPTME_TOOL_SELECT_BY_APP=0 or false still enables the feature. The code then unconditionally sets os.environ["GPTME_TOOL_SELECT_BY_APP"] = "1" when the flag is true, overwriting the user's 0. This contradicts the documented semantics that only "1"/"true" are truthy. A user who sets the env var to "0" to disable the feature will instead have it enabled, potentially loading tools (e.g., shell) they intended to keep off. The fix is to check ctx.get_parameter_source("tool_select_by_app") and only set the env var when the flag was explicitly passed on the command line, or to implement custom env var parsing that treats "0"/"false" as false.

```suggestion
if tool_select_by_app and ctx.get_parameter_source("tool_select_by_app") == ParameterSource.COMMANDLINE:
    os.environ["GPTME_TOOL_SELECT_BY_APP"] = "1"
```

<sub>**How this was verified:** Verified Click's behavior: for is_flag options with envvar, any non-empty string sets the flag to True. The main function unconditionally sets the env var to '1' whenever the option is true, overriding a user's '0' or 'false'.</sub>

**❌ P1 high** — `gptme/tools/__init__.py:182`

The expression `(config.get_env("GPTME_TOOL_SELECT_BY_APP") or "").lower()` will raise AttributeError if the config value for GPTME_TOOL_SELECT_BY_APP is a boolean True (e.g., from a TOML [env] entry `GPTME_TOOL_SELECT_BY_APP = true`). The `or ""` only handles falsy values, not non-string truthy ones. Unlike AW_SERVER_URL, which is explicitly checked for string-ness on the following lines, this env var is not. This causes a crash during tool initialization, preventing the CLI from starting. The fix is to convert the value to a string before calling .lower(), e.g., `str(config.get_env(...) or "").lower()`.

```suggestion
if str(config.get_env("GPTME_TOOL_SELECT_BY_APP") or "").lower() not in (
```

<sub>**How this was verified:** Checked the surrounding code: the AW_SERVER_URL path has a `not isinstance(server_url, str)` guard, but there is no analogous guard for GPTME_TOOL_SELECT_BY_APP, so a boolean True from TOML reaches .lower() and crashes.</sub>

</details>

- `docs/config.rst` — Documents the new tool_select_by_app setting and --tool-select-by-app flag, including precedence and AW_SERVER_URL override.
- `gptme/cli/main.py` — Adds the --tool-select-by-app boolean option and sets GPTME_TOOL_SELECT_BY_APP=1 when the flag is enabled.
- `gptme/config/models.py` — Adds tool_select_by_app: dict[str, list[str]] field to SettingsConfig.
- `gptme/tools/__init__.py` — Adds _tool_select_by_app_rules and _allowlist_from_activitywatch helpers and calls the latter from init_tools as a last fallback.
- `gptme/tools/_activitywatch.py` — New module implementing the ActivityWatch client: get_current_app, match_app_rules, and resolve_allowlist_for_current_app.
- `tests/test_tool_select_by_app.py` — Adds tests for glob matching, degrade-to-None behavior, rule merging, env var handling, and opt-in gating.

</details>

<sub>Reviewed `036c5966fdf3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 176s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Closing this one based on Erik's feedback — the current shape (explicit `--tool-select-by-app` flag + `[settings.tool_select_by_app]` config table) adds API surface that isn't justified for the use case as described.

The more compelling direction Erik sketched — an agent that **auto-adapts its tool set at init from context sources including AW, without any user-facing config** — is a materially different design. That would belong closer to the context-loading or session-init path rather than in `init_tools()` behind a user flag.

Filing the idea for a future pass under a narrower scope: agent-side tool adaptation from ambient context at session start, opt-out rather than opt-in, no new config surface.

Thanks for the review rounds.

PR #3552, opened 2026-08-15
fix(tauri): real API-key errors and Windows sidecar startup

acrinym:
## Summary
- The bundled Windows `gptme-server` sidecar crashed during `create_app` because version lookup ran `git` with a PyInstaller onefile `__file__` path as cwd (`NotADirectoryError: [WinError 267]`). The desktop UI then had no live API, so OpenRouter key save and provider settings could not work.
- Setup/settings previously turned Tauri invoke objects and nested `{ error: { message } }` API bodies into `[object Object]`. Those paths now show the real sidecar/provider error.
- Tauri sidecar builds no longer silently reuse a stale binary when gptme Python sources are newer.

## Test plan
- [ ] `pytest tests/test_misc_subprocess_timeouts.py::TestVersionTimeouts`
- [ ] `cd webui && npx jest --testPathPattern="errors.test|modelsError.test|SetupWizard.test|ServerApiKeySettings.test|WelcomeView.test"`
- [ ] Delete `tauri/bins/gptme-server-*` (or let the stale-source check rebuild), then `make tauri-build-sidecar` and rebuild the desktop app
- [ ] On Windows, open gptme-tauri, paste an OpenRouter `sk-or-...` key, confirm it saves (or shows a real error), and provider settings/health load

PR #3573, opened 2026-08-19
fix(tauri): include pyproject.toml and lock file in sidecar freshness check

TimeToBuildBob:
## Summary

Addresses the remaining Greptile 4/5 finding from #3552.

The `sidecar_is_stale()` function previously only compared `gptme/**/*.py` mtimes against the frozen sidecar binary. If `pyproject.toml` or the lock file changed (e.g. a dependency version bump, new extras, changed packaging options), the existing sidecar binary would be silently reused even though the frozen environment would differ.

**Fix**: also check `pyproject.toml` and any lock file (`poetry.lock`, `uv.lock`) that exists — trigger a rebuild whenever any packaging/dependency config file is newer than the sidecar.

## Changes

- `tauri/scripts/build-sidecar.sh`: expand `sidecar_is_stale()` to check build inputs beyond Python sources

## Test plan

- Delete `tauri/bins/gptme-server-*`, build once to get a fresh sidecar, then `touch pyproject.toml` and confirm `make tauri-build-sidecar` rebuilds.

### DIFF
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -50,7 +50,15 @@ sidecar_is_stale() {
     local uvlock_now=0
     if [[ -f "$REPO_ROOT/uv.lock" ]]; then uvlock_now=1; fi
     local uvlock_was=0
-    if [[ -f "$uvlock_marker" ]]; then uvlock_was=$(cat "$uvlock_marker"); fi
+    if [[ -f "$uvlock_marker" ]]; then
+        if [[ ! -r "$uvlock_marker" ]]; then
+            return 0
+        fi
+        uvlock_was=$(<"$uvlock_marker")
+        if [[ "$uvlock_was" != 0 && "$uvlock_was" != 1 ]]; then
+            return 0
+        fi
+    fi
     if [[ "$uvlock_now" != "$uvlock_was" ]]; then
         return 0
     fi
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -79,9 +79,9 @@ echo "Building gptme-server sidecar for $TRIPLE..."
 mkdir -p "$BINS_DIR"
 
 # Install gptme from local source into a venv, then freeze with PyInstaller.
-# When uv.lock exists, prefer --frozen so the lock actually enforces the
-# pinned versions.  If the lock is stale vs pyproject.toml, uv --frozen exits
-# non-zero; we warn and fall back to an unlocked sync (run `uv lock` to fix).
+# When uv.lock exists, use --frozen so the lock actually enforces the pinned
+# versions without modifying the lock file. If the lock cannot satisfy the
+# current pyproject.toml, stop and ask the developer to update it explicitly.
 # When uv.lock is absent (e.g. fresh checkout — it is gitignored), fall back
 # to uv pip install.  This path is NOT lock-pinned; prefer generating a uv.lock
 # (`uv lock`) or committing it for reproducible sidecar builds.
@@ -89,10 +89,10 @@ mkdir -p "$BINS_DIR"
 # to --group NAME, so --group dev selects it.  server extras add Flask etc.
 cd "$REPO_ROOT"
 if [[ -f "uv.lock" ]]; then
-    if ! uv sync --frozen --extra server --group dev --quiet 2>/dev/null; then
-        echo "WARNING: uv.lock is out of date with pyproject.toml." \
-             "Run 'uv lock' to update it. Falling back to unlocked sync..." >&2
-        uv sync --extra server --group dev --quiet
+    if ! uv sync --frozen --extra server --group dev --quiet; then
+        echo "ERROR: uv.lock cannot satisfy pyproject.toml." \
+             "Run 'uv lock' to update it, then rebuild the sidecar." >&2
+        exit 1
     fi
 else
     if [[ -f "poetry.lock" ]]; then
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -28,12 +28,12 @@ sidecar_is_stale() {
         return 0
     fi
     # Also rebuild when packaging/dependency config changes (pyproject.toml,
-    # uv.lock) — these affect which packages get frozen into the sidecar even
-    # when no .py source file changes.  poetry.lock is excluded: the install
-    # path only consults uv.lock, so watching poetry.lock would trigger spurious
-    # rebuilds whose result doesn't reflect the poetry.lock change.
+    # poetry.lock, uv.lock) — these affect which packages get frozen into the
+    # sidecar even when no .py source file changes.  Both lock files are
+    # checked: poetry.lock is the committed canonical lock; uv.lock may exist
+    # on developer machines and controls the uv sync path.
     local config_files=("$REPO_ROOT/pyproject.toml")
-    for f in "$REPO_ROOT/uv.lock"; do
+    for f in "$REPO_ROOT/poetry.lock" "$REPO_ROOT/uv.lock"; do
         [[ -f "$f" ]] && config_files+=("$f")
     done
     for f in "${config_files[@]}"; do
@@ -71,19 +71,27 @@ echo "Building gptme-server sidecar for $TRIPLE..."
 mkdir -p "$BINS_DIR"
 
 # Install gptme from local source into a venv, then freeze with PyInstaller.
-# When uv.lock exists, use uv sync (without --frozen) so the lock is updated
-# if pyproject.toml changed since the last `uv lock` run.  --frozen would fail
-# when pyproject.toml is newer than uv.lock (e.g. after adding a dep without
-# running `uv lock`), breaking the very workflow that staleness detection exists
-# to support.  uv sync still installs the exact pinned versions when the lock
-# is already current, so the pin-enforcement goal is preserved.
-# When uv.lock is absent, fall back to uv pip install (original path).
+# When uv.lock exists, prefer --frozen so the lock actually enforces the
+# pinned versions.  If the lock is stale vs pyproject.toml, uv --frozen exits
+# non-zero; we warn and fall back to an unlocked sync (run `uv lock` to fix).
+# When uv.lock is absent (e.g. fresh checkout — it is gitignored), fall back
+# to uv pip install.  This path is NOT lock-pinned; prefer generating a uv.lock
+# (`uv lock`) or committing it for reproducible sidecar builds.
 # pyinstaller is in [tool.poetry.group.dev.dependencies]; uv maps Poetry groups
 # to --group NAME, so --group dev selects it.  server extras add Flask etc.
 cd "$REPO_ROOT"
 if [[ -f "uv.lock" ]]; then
-    uv sync --extra server --group dev --quiet
+    if ! uv sync --frozen --extra server --group dev --quiet 2>/dev/null; then
+        echo "WARNING: uv.lock is out of date with pyproject.toml." \
+             "Run 'uv lock' to update it. Falling back to unlocked sync..." >&2
+        uv sync --extra server --group dev --quiet
+    fi
 else
+    if [[ -f "poetry.lock" ]]; then
+        echo "WARNING: No uv.lock found (it is gitignored); install will not" \
+             "enforce pinned versions from poetry.lock." \
+             "Run 'uv lock' to generate a uv.lock for reproducible builds." >&2
+    fi
     [[ -d ".venv" ]] || uv venv .venv
     uv pip install --quiet ".[server]" pyinstaller
 fi
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -24,7 +24,22 @@ sidecar_is_stale() {
     local sidecar="$1"
     # Rebuild when any gptme Python source is newer than the frozen binary.
     # `find -newer` is available in Git Bash on Windows.
-    [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]
+    if [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]; then
+        return 0
+    fi
+    # Also rebuild when packaging/dependency config changes (pyproject.toml,
+    # poetry.lock, uv.lock) — these affect which packages get frozen into the
+    # sidecar even when no .py source file changes.
+    local config_files=("$REPO_ROOT/pyproject.toml")
+    for f in "$REPO_ROOT/poetry.lock" "$REPO_ROOT/uv.lock"; do
+        [[ -f "$f" ]] && config_files+=("$f")
+    done
+    for f in "${config_files[@]}"; do
+        if [[ -f "$f" && "$f" -nt "$sidecar" ]]; then
+            return 0
+        fi
+    done
+    return 1
 }
 
 if [[ -f "$OUT" ]]; then
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -41,6 +41,19 @@ sidecar_is_stale() {
             return 0
         fi
     done
+    # Detect install-path changes caused by uv.lock appearing or disappearing.
+    # If uv.lock was present at last build but has since been deleted (or vice
+    # versa), the install path changes (uv sync vs uv pip install), so the
+    # sidecar would be built with different dependency resolution.  A marker
+    # file records uv.lock presence at build time; any mismatch forces rebuild.
+    local uvlock_marker="${sidecar}.uvlock-present"
+    local uvlock_now=0
+    [[ -f "$REPO_ROOT/uv.lock" ]] && uvlock_now=1
+    local uvlock_was=0
+    [[ -f "$uvlock_marker" ]] && uvlock_was=$(cat "$uvlock_marker")
+    if [[ "$uvlock_now" != "$uvlock_was" ]]; then
+        return 0
+    fi
     return 1
 }
 
@@ -81,12 +94,15 @@ uv run pyinstaller \
     gptme/server/__main__.py
 
 # Rename to include target triple (Tauri sidecar convention)
+_uvlock_state=0; [[ -f "$REPO_ROOT/uv.lock" ]] && _uvlock_state=1
 if [[ -f "$BINS_DIR/gptme-server.exe" ]]; then
     mv "$BINS_DIR/gptme-server.exe" "${BINS_DIR}/gptme-server-${TRIPLE}.exe"
     echo "Sidecar built: ${BINS_DIR}/gptme-server-${TRIPLE}.exe"
+    echo "$_uvlock_state" > "${BINS_DIR}/gptme-server-${TRIPLE}.exe.uvlock-present"
 elif [[ -f "$BINS_DIR/gptme-server" ]]; then
     mv "$BINS_DIR/gptme-server" "$BINS_DIR/gptme-server-${TRIPLE}"
     echo "Sidecar built: $BINS_DIR/gptme-server-${TRIPLE}"
+    echo "$_uvlock_state" > "$BINS_DIR/gptme-server-${TRIPLE}.uvlock-present"
 else
     echo "ERROR: PyInstaller did not produce gptme-server in $BINS_DIR" >&2
     exit 1
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -65,7 +65,8 @@ mkdir -p "$BINS_DIR"
 # to support.  uv sync still installs the exact pinned versions when the lock
 # is already current, so the pin-enforcement goal is preserved.
 # When uv.lock is absent, fall back to uv pip install (original path).
-# pyinstaller is in [tool.uv.dev-dependencies]; server extras add Flask etc.
+# pyinstaller is in [tool.poetry.group.dev.dependencies]; uv maps Poetry groups
+# to --group NAME, so --group dev selects it.  server extras add Flask etc.
 cd "$REPO_ROOT"
 if [[ -f "uv.lock" ]]; then
     uv sync --extra server --group dev --quiet
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -48,9 +48,9 @@ sidecar_is_stale() {
     # file records uv.lock presence at build time; any mismatch forces rebuild.
     local uvlock_marker="${sidecar}.uvlock-present"
     local uvlock_now=0
-    [[ -f "$REPO_ROOT/uv.lock" ]] && uvlock_now=1
+    if [[ -f "$REPO_ROOT/uv.lock" ]]; then uvlock_now=1; fi
     local uvlock_was=0
-    [[ -f "$uvlock_marker" ]] && uvlock_was=$(cat "$uvlock_marker")
+    if [[ -f "$uvlock_marker" ]]; then uvlock_was=$(cat "$uvlock_marker"); fi
     if [[ "$uvlock_now" != "$uvlock_was" ]]; then
         return 0
     fi
@@ -102,7 +102,7 @@ uv run pyinstaller \
     gptme/server/__main__.py
 
 # Rename to include target triple (Tauri sidecar convention)
-_uvlock_state=0; [[ -f "$REPO_ROOT/uv.lock" ]] && _uvlock_state=1
+_uvlock_state=0; if [[ -f "$REPO_ROOT/uv.lock" ]]; then _uvlock_state=1; fi
 if [[ -f "$BINS_DIR/gptme-server.exe" ]]; then
     mv "$BINS_DIR/gptme-server.exe" "${BINS_DIR}/gptme-server-${TRIPLE}.exe"
     echo "Sidecar built: ${BINS_DIR}/gptme-server-${TRIPLE}.exe"
--- a/tauri/scripts/build-sidecar.sh
+++ b/tauri/scripts/build-sidecar.sh
@@ -28,10 +28,12 @@ sidecar_is_stale() {
         return 0
     fi
     # Also rebuild when packaging/dependency config changes (pyproject.toml,
-    # poetry.lock, uv.lock) — these affect which packages get frozen into the
-    # sidecar even when no .py source file changes.
+    # uv.lock) — these affect which packages get frozen into the sidecar even
+    # when no .py source file changes.  poetry.lock is excluded: the install
+    # path only consults uv.lock, so watching poetry.lock would trigger spurious
+    # rebuilds whose result doesn't reflect the poetry.lock change.
     local config_files=("$REPO_ROOT/pyproject.toml")
-    for f in "$REPO_ROOT/poetry.lock" "$REPO_ROOT/uv.lock"; do
+    for f in "$REPO_ROOT/uv.lock"; do
         [[ -f "$f" ]] && config_files+=("$f")
     done
     for f in "${config_files[@]}"; do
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
## 🤖 AI code review

The PR expands `sidecar_is_stale()` in `tauri/scripts/build-sidecar.sh` to also compare `pyproject.toml` and `uv.lock` mtimes against the sidecar binary, and adds a marker file that records whether `uv.lock` was present at build time so that the script can detect its appearance or disappearance. When `uv.lock` exists, the build now uses `uv sync --frozen --extra server --group dev` instead of the previous unconditional `uv pip install`.

### Confidence Score: N/A

**⚠️ P2 medium · ℹ️ non-blocking** — `tauri/scripts/build-sidecar.sh:40`

The stale check only considers configuration files that are still present. If `pyproject.toml` is deleted after a sidecar is built, the condition `-f "$f"` is false, so the loop never triggers a rebuild even though the build inputs have changed. As a result, an existing sidecar is treated as up-to-date and the script skips the rebuild, leaving a binary that was built against a dependency definition that no longer exists. Unlike `uv.lock` (which is handled via the marker mismatch), `pyproject.toml` deletion has no fallback mechanism. The fix would be to also treat an absent `pyproject.toml` as a stale condition, for instance by keeping a presence marker analogous to the uvlock one.

<sub>**How this was verified:** I traced the loop over `config_files`; every file is guarded by `-f "$f"` before the mtime comparison, and there is no branch that checks for the file having been removed. The marker logic in the same function only covers uv.lock presence.</sub>

</details>

- `tauri/scripts/build-sidecar.sh` — Expands the freshness check to include pyproject.toml, uv.lock, and a uvlock-presence marker; switches the install step to `uv sync --frozen` when uv.lock exists.

</details>

| `7430a3078a74` | 3/5 | 2 | llm | 2026-08-19 17:21 UTC |
| `b7c2cc3297cb` | 5/5 | 0 | llm | 2026-08-19 18:08 UTC |
| `b7c2cc3297cb` | 3/5 | 1 | llm | 2026-08-19 18:30 UTC |
| `87349d15ac27` | 3/5 | 1 | llm | 2026-08-20 00:20 UTC |
| `d9eeac2c7e0c` | ? | 1 | llm | 2026-08-20 01:03 UTC |
| `57395f20ee1d` | 3/5 | 1 | llm | 2026-08-20 03:03 UTC |
| `68abeff78788` | ? | 2 | llm | 2026-08-20 04:41 UTC |
| `822f4bf6dd3b` | 3/5 | 3 | llm | 2026-08-20 05:42 UTC |
| `6ea1238ed595` | 5/5 | 0 | llm | 2026-08-20 07:04 UTC |
| `6ea1238ed595` | ? | 2 | llm | 2026-08-20 07:41 UTC |

</details>

<sub>Reviewed `470f9e46050a` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 204s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

TimeToBuildBob:
Addressed Greptile's 4/5 finding in `170e10064`.

The concern was valid: the freshness check added `uv.lock` as a build input, but the rebuild ran `uv pip install ".[server]" pyinstaller` — which ignores `uv.lock` and resolves deps fresh. So a lock-only bump would trigger a rebuild that didn't actually apply the pinned versions.

Fix: replaced `uv pip install` with `uv sync --frozen --extra server --group dev`. `uv sync --frozen` reads `uv.lock` and installs exact pinned versions, so a lock-triggered rebuild now produces a sidecar that matches the lock. `pyinstaller` was already in the dev dep group, so the `--group dev` flag covers it without any schema change.

tauri/scripts/build-sidecar.sh
```diff
@@ -40,11 +55,12 @@ fi
...
-# Install gptme from local source into an isolated venv, then freeze with PyInstaller
+# Install gptme from local source into a venv, then freeze with PyInstaller.
+# uv sync --frozen uses uv.lock exactly, so a lock-file-triggered rebuild
+# actually installs the pinned versions that caused the staleness.
+# pyinstaller is in [tool.poetry.group.dev.dependencies]; server extras add Flask etc.
 cd "$REPO_ROOT"
-# uv pip install requires a virtual environment — create one if absent
-[[ -d ".venv" ]] || uv venv .venv
-uv pip install --quiet ".[server]" pyinstaller
+uv sync --frozen --extra server --group dev --quiet
```
TimeToBuildBob: ❌ **P1** — The staleness check now triggers on changes to `poetry.lock` (lines 33-41), but the build command `uv sync --frozen` only uses `uv.lock` for pinning dependencies. If `poetry.lock` is updated while `uv.lock` remains unchanged, the rebuild runs but installs the same dependency versions, so the sidecar binary is effectively unchanged — the Greptile finding about dependency changes is not actually addressed for `poetry.lock`. Worse, if a repository only maintains `poetry.lock` (the previous `uv pip install` command did not require any lock file), `uv sync --frozen` fails because `uv.lock` is missing, aborting the entire sidecar build with an error from uv. The script provides no fallback or pre-check for the existence of `uv.lock`, so any environment without a uv-managed lock file cannot build the sidecar at all.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
```diff
@@ -40,11 +55,19 @@ fi
...
+# actually installs the pinned versions that caused the staleness.
+# When only poetry.lock exists (or no lock file at all), fall back to
+# uv pip install, which behaves like the original installation path.
+# pyinstaller is in [tool.poetry.group.dev.dependencies]; server extras add Flask etc.
 cd "$REPO_ROOT"
-# uv pip install requires a virtual environment — create one if absent
-[[ -d ".venv" ]] || uv venv .venv
-uv pip install --quiet ".[server]" pyinstaller
+if [[ -f "uv.lock" ]]; then
+    uv sync --frozen --extra server --group dev --quiet
```
TimeToBuildBob: ❌ **P1** — When uv.lock exists, a stale sidecar is rebuilt with `uv sync --frozen`, which requires uv.lock to be in sync with pyproject.toml. The staleness check now triggers a rebuild when pyproject.toml is newer than the sidecar (line 27), but if the developer only modified pyproject.toml without regenerating uv.lock, `uv sync --frozen` fails with an error because the lock file is out of date. This breaks the very workflow the PR intends to support: the test plan (touch pyproject.toml, expect rebuild) would fail whenever uv.lock is present. The previous `uv pip install` would have resolved fresh and succeeded. Concretely, after touching pyproject.toml, the script reaches this line and exits non-zero, so the sidecar is never rebuilt and the Tauri build is blocked.

TimeToBuildBob:
## Convergence adjudication — gptme/gptme#3573

Hit the Greptile attempt cap. Adjudicating the two remaining P1 findings; fix pushed in `b7c2cc329`.

### Fixed in this session

**P1 — "Lock files are not enforced" (`tauri/scripts/build-sidecar.sh:36`)**  
`poetry.lock` was included in `sidecar_is_stale()` but the install path never reads it. Watching it caused spurious rebuilds whose result didn't reflect the `poetry.lock` change. **Fix:** removed `poetry.lock` from the staleness check; only `pyproject.toml` and `uv.lock` remain.

**P1 — "Frozen sync requires absent lockfile" (`tauri/scripts/build-sidecar.sh:63`)**  
`uv sync --frozen` fails when `pyproject.toml` is newer than `uv.lock` (e.g. a developer adds a dep without running `uv lock` first). This breaks the exact workflow the staleness detection was added to support: touch `pyproject.toml` → expect rebuild to succeed. **Fix:** dropped `--frozen`; plain `uv sync` still installs exact pinned versions when the lock is current, and updates the lock when `pyproject.toml` changed.

### Remaining findings

None. Both P1s addressed. The P2 from my AI review (poetry.lock rebuild without honoring it) is resolved by the same fix — poetry.lock is no longer a staleness trigger.

### CI status

All checks passing (Build ubuntu/macos/windows, PyInstaller binaries, Docker, lint, typecheck, openapi, unit tests). One pending: `Lint (format + clippy)` — this is the Tauri Rust lint, unrelated to the shell script changes.

### Domain risk

The sidecar build script is dev-only tooling; CI builds the sidecar independently. The behavioral change is: `uv sync` may update `uv.lock` as a side effect when `pyproject.toml` changed — this is expected and correct, but maintainers should be aware the lock file can be modified on disk during a sidecar rebuild.

### Convergence

`stable_rounds: 0` — both rounds introduced new blocking findings. With both P1s now addressed and no further re-trigger, this closes the convergence loop.

Not requesting another Greptile review — these are the final substantive fixes on this diff. Maintainer merge call.

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,24 @@ sidecar_is_stale() {
...
-    [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]
+    if [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]; then
+        return 0
+    fi
+    # Also rebuild when packaging/dependency config changes (pyproject.toml,
+    # uv.lock) — these affect which packages get frozen into the sidecar even
+    # when no .py source file changes.  poetry.lock is excluded: the install
+    # path only consults uv.lock, so watching poetry.lock would trigger spurious
+    # rebuilds whose result doesn't reflect the poetry.lock change.
+    local config_files=("$REPO_ROOT/pyproject.toml")
```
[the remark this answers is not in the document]
TimeToBuildBob: Acknowledged — but this is by design: the sidecar is built with `uv sync` (not poetry), so `poetry.lock` has no effect on which packages get frozen into the binary. Tracking it in the staleness check would cause unnecessary rebuilds with an identical result. The code comment explains this explicitly:

> `poetry.lock` is excluded: the install path only consults `uv.lock`, so watching `poetry.lock` would trigger spurious rebuilds whose result doesn't reflect the `poetry.lock` change.

`uv.lock` _is_ tracked — a change there correctly triggers a rebuild. No action needed on this finding.

tauri/scripts/build-sidecar.sh
```diff
@@ -40,11 +57,22 @@ fi
...
+# to support.  uv sync still installs the exact pinned versions when the lock
+# is already current, so the pin-enforcement goal is preserved.
+# When uv.lock is absent, fall back to uv pip install (original path).
+# pyinstaller is in [tool.uv.dev-dependencies]; server extras add Flask etc.
 cd "$REPO_ROOT"
-# uv pip install requires a virtual environment — create one if absent
-[[ -d ".venv" ]] || uv venv .venv
-uv pip install --quiet ".[server]" pyinstaller
+if [[ -f "uv.lock" ]]; then
+    uv sync --extra server --group dev --quiet
```
TimeToBuildBob: ❌ **P1** — The `uv sync` command uses `--group dev`, which selects a dependency group named "dev" from `[dependency-groups]` (PEP 735). However, the comment in the same change states that "pyinstaller is in [tool.uv.dev-dependencies]". These are distinct: `[tool.uv.dev-dependencies]` is the legacy location for dev dependencies and is not the same as a group named "dev". If the project's pyproject.toml uses the legacy section and does not define a `[dependency-groups]` group named "dev", `uv sync` will fail with an error like "Unknown group: dev", aborting the build. Since the change also removed the explicit `pyinstaller` from the install command, the subsequent `uv run pyinstaller` would also fail if pyinstaller is not installed. The correct flag to include dev dependencies from the legacy section is `--dev` (or simply omit the flag, since `uv sync` includes them by default). _(2/3 passes)_

TimeToBuildBob:
## Follow-up adjudication — gptme/gptme#3573

Continuing from the convergence adjudication above. A new AI review round (18:30 UTC) surfaced one additional finding on the same commit `b7c2cc3297cb`.

### AI review P1 — `--group dev` flag (false positive)

> The correct flag to include dev dependencies from the legacy section is `--dev`

**Dismissed.** The project uses `[tool.poetry.group.dev.dependencies]` in `pyproject.toml` (line 139, 161) — not uv's `[tool.uv.dev-dependencies]`. When uv reads a Poetry-format pyproject, it maps `[tool.poetry.group.NAME.dependencies]` to named groups accessible via `--group NAME`. So `uv sync --group dev` is the correct flag here. Using `--dev` instead would target uv's own legacy dev-dependencies format, which this project doesn't use.

The script comment was wrong (said `[tool.uv.dev-dependencies]`); that's now fixed in `87349d15a` to accurately say `[tool.poetry.group.dev.dependencies]` and explain the `--group` mapping.

### Greptile `poetry.lock` concern (design decision)

Greptile flagged 4/5 because `poetry.lock` is not in the staleness check. This is intentional and documented in the script: the install path runs `uv sync`, which reads `uv.lock`, not `poetry.lock`. Including `poetry.lock` in the staleness trigger would cause rebuilds whose result doesn't reflect the `poetry.lock` change — the sidecar's installed packages wouldn't change. The correct trigger for lock-driven rebuilds is `uv.lock`.

All 3 Greptile inline threads resolved (they were addressed across the previous commits in this PR).

### Final state

- CI: all required checks pass
- Greptile: 4/5 (poetry.lock design decision, not a bug)
- AI review P1: false positive, dismissed with reason; comment corrected in `87349d15a`
- Sidecar build CI: skipping (path not triggered by this change) — the build script logic is unchanged from the shell-level; the only change in `87349d15a` is a comment correction

Maintainer merge call.

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
TimeToBuildBob: Fixed in `d9eeac2c7`: now write a `.uvlock-present` marker (0/1) alongside the sidecar binary at build time. `sidecar_is_stale()` reads it and forces a rebuild whenever uv.lock's current presence differs from what was recorded — covering both the 'lock deleted' and 'lock appeared' cases.

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,37 @@ sidecar_is_stale() {
...
+    # Detect install-path changes caused by uv.lock appearing or disappearing.
+    # If uv.lock was present at last build but has since been deleted (or vice
+    # versa), the install path changes (uv sync vs uv pip install), so the
+    # sidecar would be built with different dependency resolution.  A marker
+    # file records uv.lock presence at build time; any mismatch forces rebuild.
+    local uvlock_marker="${sidecar}.uvlock-present"
+    local uvlock_now=0
+    [[ -f "$REPO_ROOT/uv.lock" ]] && uvlock_now=1
+    local uvlock_was=0
+    [[ -f "$uvlock_marker" ]] && uvlock_was=$(cat "$uvlock_marker")
```
TimeToBuildBob: ⚠️ **P2** — When the marker file is missing (e.g. the sidecar binary was copied from a build that used uv.lock, but the .uvlock-present file was not copied, or was manually deleted), uvlock_was defaults to 0. If uv.lock is also absent at the time of the check, the mismatch test `[[ "$uvlock_now" != "$uvlock_was" ]]` sees 0 != 0, so no rebuild is triggered. However, the existing sidecar was built with `uv sync` (pinned dependency resolution) while the current environment uses `uv pip install` (fresh resolution), so the frozen binary no longer matches the dependency set that would be produced now. This leaves a stale binary that silently uses different dependencies than the build would produce. The fix is to treat a missing marker as an unknown state and force a rebuild whenever the marker is absent.

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
```diff
@@ -40,24 +70,39 @@ fi
...
+# is already current, so the pin-enforcement goal is preserved.
+# When uv.lock is absent, fall back to uv pip install (original path).
+# pyinstaller is in [tool.poetry.group.dev.dependencies]; uv maps Poetry groups
+# to --group NAME, so --group dev selects it.  server extras add Flask etc.
 cd "$REPO_ROOT"
-# uv pip install requires a virtual environment — create one if absent
-[[ -d ".venv" ]] || uv venv .venv
-uv pip install --quiet ".[server]" pyinstaller
+if [[ -f "uv.lock" ]]; then
+    uv sync --extra server --group dev --quiet
```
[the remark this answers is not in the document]
TimeToBuildBob: **Dismissed — false positive.**

The finding describes correct behavior: when `pyproject.toml` is newer than `uv.lock`, the staleness check fires and `uv sync` updates the lock. That update IS the correct resolution — the lock is stale because the developer added a dependency without running `uv lock` first. The "newly resolved dependency set" is the right one.

The alternative (`--frozen`) fails outright in this case (that was Round 2's P1 — we removed `--frozen` specifically to handle this scenario). Requiring the developer to manually run `uv lock` before the build would add friction with no safety benefit: the staleness check would still fire, and the result would be identical.

This is dev-only tooling. When `pyproject.toml` changes, `uv sync` resolving the new deps and updating the lock is the expected workflow.

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,37 @@ sidecar_is_stale() {
...
+    # Detect install-path changes caused by uv.lock appearing or disappearing.
+    # If uv.lock was present at last build but has since been deleted (or vice
+    # versa), the install path changes (uv sync vs uv pip install), so the
+    # sidecar would be built with different dependency resolution.  A marker
+    # file records uv.lock presence at build time; any mismatch forces rebuild.
+    local uvlock_marker="${sidecar}.uvlock-present"
+    local uvlock_now=0
+    [[ -f "$REPO_ROOT/uv.lock" ]] && uvlock_now=1
+    local uvlock_was=0
+    [[ -f "$uvlock_marker" ]] && uvlock_was=$(cat "$uvlock_marker")
```
TimeToBuildBob: ❌ **P1** — The `&&` pattern here and on line 51 causes the script to exit under `set -euo pipefail` whenever the tested file does not exist. If `uv.lock` is absent (common on fresh checkouts, since it is gitignored) or the `.uvlock-present` marker has not been written (e.g., sidecar built by an older script version), the `[[ -f ... ]]` test returns 1, making the AND-list fail. Because this is a standalone command not guarded by `if` or `while`, the shell aborts immediately. Consequently, `sidecar_is_stale()` never returns a status, and `make tauri-build-sidecar` exits with an error instead of either rebuilding or skipping, breaking incremental builds for any existing sidecar lacking the marker or for developers without `uv.lock`.

TimeToBuildBob:
Addressed the AI reviewer P1 at `tauri/scripts/build-sidecar.sh:53` in `68abeff78`.

**Finding**: Under `set -euo pipefail`, the `&&` pattern `[[ -f file ]] && var=$(cat file)` can cause the script to exit when the file doesn't exist, because the failing `[[ -f ]]` test in the left-hand side of `&&` may not be fully protected under all bash `set -e` semantics.

**Fix**: Replaced the three `&&`-guarded file-existence assignments with explicit `if` blocks:

```bash
# Before
[[ -f "$REPO_ROOT/uv.lock" ]] && uvlock_now=1
[[ -f "$uvlock_marker" ]] && uvlock_was=$(cat "$uvlock_marker")

# After
if [[ -f "$REPO_ROOT/uv.lock" ]]; then uvlock_now=1; fi
if [[ -f "$uvlock_marker" ]]; then uvlock_was=$(cat "$uvlock_marker"); fi
```

This is the explicit, unambiguous form that is always safe under `set -e` — no reliance on bash's `&&`-exemption semantics. Functionally identical behaviour, clearer intent.

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,37 @@ sidecar_is_stale() {
...
+    if [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]; then
+        return 0
+    fi
+    # Also rebuild when packaging/dependency config changes (pyproject.toml,
+    # poetry.lock, uv.lock) — these affect which packages get frozen into the
+    # sidecar even when no .py source file changes.  Both lock files are
+    # checked: poetry.lock is the committed canonical lock; uv.lock may exist
+    # on developer machines and controls the uv sync path.
+    local config_files=("$REPO_ROOT/pyproject.toml")
+    for f in "$REPO_ROOT/poetry.lock" "$REPO_ROOT/uv.lock"; do
```
TimeToBuildBob: ⚠️ **P2** — The freshness check treats `poetry.lock` as a build input, but the build never consumes it. When `uv.lock` exists, `uv sync` resolves dependencies from `uv.lock`; when `uv.lock` is absent, the fallback `uv pip install ".[server]" pyinstaller` resolves directly from `pyproject.toml` without any lock. A change to `poetry.lock` therefore triggers an unnecessary sidecar rebuild even though the frozen binary is identical, wasting build time and causing spurious rebuilds in CI or local development every time the committed `poetry.lock` is touched.

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,37 @@ sidecar_is_stale() {
...
+    done
+    # Detect install-path changes caused by uv.lock appearing or disappearing.
+    # If uv.lock was present at last build but has since been deleted (or vice
+    # versa), the install path changes (uv sync vs uv pip install), so the
+    # sidecar would be built with different dependency resolution.  A marker
+    # file records uv.lock presence at build time; any mismatch forces rebuild.
+    local uvlock_marker="${sidecar}.uvlock-present"
+    local uvlock_now=0
+    if [[ -f "$REPO_ROOT/uv.lock" ]]; then uvlock_now=1; fi
+    local uvlock_was=0
```
TimeToBuildBob, line 52: ⚠️ **P2** — When a sidecar was built by a previous version of this script, the `uvlock-present` marker file does not exist, so `uvlock_was` defaults to 0. If `uv.lock` is present now (`uvlock_now=1`), the mismatch makes `sidecar_is_stale()` return true, forcing a rebuild. The marker is only written after a successful build, so an existing, up-to-date sidecar with an unchanged `uv.lock` is incorrectly considered stale on the first run after upgrading to this script. This causes a needless rebuild (and its build time) even though no build input changed. After the rebuild the marker is written and the problem disappears, but the spurious rebuild is still a defect.

TimeToBuildBob:
@greptileai review

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,22 @@ sidecar_is_stale() {
...
+    if [[ -n "$(find "$REPO_ROOT/gptme" -name '*.py' -newer "$sidecar" 2>/dev/null | head -n 1)" ]]; then
+        return 0
+    fi
+    # Also rebuild when packaging/dependency config changes (pyproject.toml,
+    # poetry.lock, uv.lock) — these affect which packages get frozen into the
+    # sidecar even when no .py source file changes.
+    local config_files=("$REPO_ROOT/pyproject.toml")
+    for f in "$REPO_ROOT/poetry.lock" "$REPO_ROOT/uv.lock"; do
+        [[ -f "$f" ]] && config_files+=("$f")
+    done
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `822f4bf6d` — removed `poetry.lock` from the `config_files` loop. The staleness check now only watches `pyproject.toml` and `uv.lock` (when present). An explicit comment explains the exclusion: the install path only consults `uv.lock` via `uv sync --frozen`, so including `poetry.lock` caused spurious rebuilds with an identical result.

tauri/scripts/build-sidecar.sh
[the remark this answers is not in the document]
TimeToBuildBob: This is handled in the current code. The `uv sync --frozen` path only runs when `uv.lock` exists (line 83: `if [[ -f "uv.lock" ]]; then`). When `uv.lock` is absent, the script falls through to the `else` branch at line 89, which uses `uv pip install` instead and emits a warning that the build won't enforce pinned versions. No early exit; the sidecar build completes.

tauri/scripts/build-sidecar.sh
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `822f4bf6d` — removed `poetry.lock` from the staleness check entirely. Only `pyproject.toml` and `uv.lock` (when present) are now checked. The code comment makes the exclusion explicit: the install path only consults `uv.lock` via `uv sync --frozen`, so watching `poetry.lock` was causing spurious rebuilds with an identical result.

TimeToBuildBob: Fixed in `822f4bf6d` — removed `poetry.lock` from `config_files`. Only `pyproject.toml` and `uv.lock` are now checked, with an explicit comment explaining the exclusion.

tauri/scripts/build-sidecar.sh
TimeToBuildBob: Accepted trade-off: on first run after upgrading to this script version, an existing sidecar without the `.uvlock-present` marker will trigger a one-time rebuild if `uv.lock` is present. This is intentional — it ensures the marker gets written and the sidecar's install path is definitively recorded. Subsequent runs are not affected.

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,37 @@ sidecar_is_stale() {
...
+    # Detect install-path changes caused by uv.lock appearing or disappearing.
+    # If uv.lock was present at last build but has since been deleted (or vice
+    # versa), the install path changes (uv sync vs uv pip install), so the
+    # sidecar would be built with different dependency resolution.  A marker
+    # file records uv.lock presence at build time; any mismatch forces rebuild.
+    local uvlock_marker="${sidecar}.uvlock-present"
+    local uvlock_now=0
+    if [[ -f "$REPO_ROOT/uv.lock" ]]; then uvlock_now=1; fi
+    local uvlock_was=0
+    if [[ -f "$uvlock_marker" ]]; then uvlock_was=$(cat "$uvlock_marker"); fi
```
TimeToBuildBob: ❌ **P1** — The `cat` on the marker file is unguarded. If the marker file exists but is unreadable (permissions, I/O error), the command substitution returns non-zero; under `set -e` this aborts the whole script, preventing a build that should have proceeded. Even if it does not abort, a malformed or empty marker file yields a value that never equals `uvlock_now` ("0" or "1"), so `[[ "$uvlock_now" != "$uvlock_was" ]]` is always true and the sidecar is rebuilt on every invocation — silently defeating the up-to-date check and adding minutes to each run. This is a real error-handling defect in a path that is hit on every start when uv.lock exists.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Latest head `6ea1238ed` is clean: all CI checks pass, Greptile is 5/5 on this SHA, and the two stale review threads are resolved. Self-merge remains ineligible because `tauri/scripts/build-sidecar.sh` is outside the allowed path categories. Waiting for maintainer review/merge; no further code change is needed.

tauri/scripts/build-sidecar.sh
```diff
@@ -40,24 +78,47 @@ fi
...
+# to --group NAME, so --group dev selects it.  server extras add Flask etc.
 cd "$REPO_ROOT"
-# uv pip install requires a virtual environment — create one if absent
-[[ -d ".venv" ]] || uv venv .venv
-uv pip install --quiet ".[server]" pyinstaller
+if [[ -f "uv.lock" ]]; then
+    if ! uv sync --frozen --extra server --group dev --quiet 2>/dev/null; then
+        echo "WARNING: uv.lock is out of date with pyproject.toml." \
+             "Run 'uv lock' to update it. Falling back to unlocked sync..." >&2
+        uv sync --extra server --group dev --quiet
```
TimeToBuildBob: ⚠️ **P2** — When `uv sync --frozen` fails because uv.lock is stale, the fallback runs `uv sync` without `--frozen`, which will regenerate and overwrite uv.lock to match pyproject.toml. This silently mutates a file the user may have committed, and contradicts the warning that advises running `uv lock` manually. The mutation can introduce dependency changes without explicit user review and dirt the working tree. The fallback should instead use `uv pip install` (as in the no-uv.lock path) or exit with an instruction to run `uv lock`. _(3/3 passes)_

TimeToBuildBob:
Addressed the open AI-review P2 in `470f9e460`.

The stale-lock fallback no longer runs an unlocked `uv sync` that rewrites `uv.lock`. A failed frozen sync now exits with an explicit `uv lock` instruction, preserving the committed lock file and requiring dependency updates to be reviewed separately.

Verified with `bash -n`, ShellCheck, `git diff --check`, pre-commit, and a stubbed failure-path test confirming exactly one frozen sync attempt and no fallback sync. Greptile re-review was not requested because the helper has reached its 8-trigger cap; the previous 5/5 review remains on the prior head. CI is running on this commit.

tauri/scripts/build-sidecar.sh
```diff
@@ -24,7 +24,45 @@ sidecar_is_stale() {
...
+    # uv.lock) — these affect which packages get frozen into the sidecar even
+    # when no .py source file changes.  poetry.lock is intentionally excluded:
+    # the install path only consults uv.lock (via uv sync --frozen), so watching
+    # poetry.lock would trigger spurious rebuilds with an identical result.
+    local config_files=("$REPO_ROOT/pyproject.toml")
+    if [[ -f "$REPO_ROOT/uv.lock" ]]; then
+        config_files+=("$REPO_ROOT/uv.lock")
+    fi
+    for f in "${config_files[@]}"; do
+        if [[ -f "$f" && "$f" -nt "$sidecar" ]]; then
```
TimeToBuildBob, line 40: ⚠️ **P2** — The stale check only considers configuration files that are still present. If `pyproject.toml` is deleted after a sidecar is built, the condition `-f "$f"` is false, so the loop never triggers a rebuild even though the build inputs have changed. As a result, an existing sidecar is treated as up-to-date and the script skips the rebuild, leaving a binary that was built against a dependency definition that no longer exists. Unlike `uv.lock` (which is handled via the marker mismatch), `pyproject.toml` deletion has no fallback mechanism. The fix would be to also treat an absent `pyproject.toml` as a stale condition, for instance by keeping a presence marker analogous to the uvlock one.

PR #3574, opened 2026-08-19
feat(cli): add gptme service init scaffold for headless agents

TimeToBuildBob:
Adds a `gptme service init` subcommand that scaffolds a persistent headless gptme agent as a systemd user service — the pattern Bob uses to run 100+ autonomous sessions/day, without forcing operators to reverse-engineer an agent workspace.

## What it generates
- **systemd service unit** (configurable model, work-dir)
- **optional systemd timer** (hourly / daily / weekly / on-demand)
- **gptme-agent-run.sh** startup script (executable, writes a durable session log)
- **skeleton gptme.toml** and **AGENTS.md**

## Design
- Registered as the `gptme-service` console script; dispatched via the existing `gptme CMD → gptme-CMD` plugin mechanism.
- Writes files only — `systemctl enable --now` is left to the operator.
- No heavy dependencies (stdlib + click only).

## Testing
- 5 integration tests: file generation, Bash syntax validation, systemd unit parse, on-demand (no timer) path, --force overwrite behavior.
- `ruff`, `mypy`, and pre-commit all clean.

Refs #1294 (headless-agent demand).


TimeToBuildBob:
Pushed `706aad1` addressing all 5 Greptile P1s:

- **Weekly schedule**: fixed `"weekly"` key to `Mon *-*-* 00:00:00` (was identical to daily)
- **Timer immediate start**: removed `Requires={name}.service` from the timer unit (prevented `--now` from starting the service before the first `OnCalendar` fire)
- **ExecStart path**: quoted the path (`ExecStart="{work_dir}/gptme-agent-run.sh"`) to handle work dirs with spaces
- **Stale timer on-demand**: reinitializing with `--timer-schedule on-demand` now removes any existing `{name}.timer` file
- **chmod on skip**: gated `startup.chmod(0o755)` behind the `_write_file` return value so skipped files don't get their permissions changed silently

All 5 existing tests still pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `c9e6f4c` addressing the Greptile P1 from the second review:

**On-demand timer deletion now respects `--force`**

Previously, reinitializing with `--timer-schedule on-demand` unconditionally deleted any existing `{name}.timer`, violating the same preservation policy applied everywhere else in the scaffold. Fixed:
- Without `--force`: preserve the existing timer and print a warning
- With `--force`: remove it (the old behavior, now gated correctly)

Added two regression tests covering both branches. All 7 tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `49c28fa` addressing the remaining Greptile P1 (timer file removed before disable):

**Root cause**: `--force --timer-schedule on-demand` deleted the timer unit file before `systemctl disable --now` could run, so systemd couldn't resolve the unit name to actually stop it.

**Fix**: call `subprocess.run(["systemctl", "--user", "disable", "--now", name.timer], check=False)` **before** `stale_timer.unlink()`, matching systemd's established cleanup order (disable → unlink). The call is best-effort (`check=False`) so a non-enabled unit doesn't abort the command.

**New test**: `test_on_demand_removes_existing_timer_with_force` now mocks `subprocess.run` and asserts the disable call fires while the timer file still exists, covering the previously unreachable failure path.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,338 @@
...
+
+SYSTEMD_SERVICE_TEMPLATE = """\
+[Unit]
+Description=gptme Autonomous Agent: {name}
+Documentation=https://github.com/gptme/gptme#headless
+After=network-online.target
+
+[Service]
+Type=simple
+User=%u
```
TimeToBuildBob: ❌ **P1** — The generated systemd service unit includes `User=%u`, and `%u` is the username of the invoking user. In a systemd *user* service, the unit already runs as the user that called `systemctl --user`, and the `User=` directive is not supported in user units — systemd's user manager rejects such units with an error like "Unit ... has a User= setting, which is not supported in user units." The consequence is that `systemctl --user start myagent.service` fails to load the unit, so the scaffolded service cannot be started at all. The operator following the printed next-steps instructions hits this failure immediately. Remove the `User=%u` line from the service template; the service will run as the invoking user by default.

tests/test_cmd_service.py
```diff
@@ -0,0 +1,166 @@
...
+    if shutil.which("systemd-analyze"):
+        proc = subprocess.run(
+            ["systemd-analyze", "verify", str(service)],
+            capture_output=True,
+            text=True,
+            check=False,
+        )
+        # systemd-analyze verify may warn on missing ExecStart deps but should
+        # not fail hard on a syntactically valid unit.
+        assert proc.returncode in (0, 1)
```
TimeToBuildBob: ⚠️ **P2** — The test `test_systemd_unit_parses` accepts a return code of 1 from `systemd-analyze verify`. `systemd-analyze verify` returns 0 for a valid unit and non-zero (typically 1) when there are errors. By allowing 1, the test passes even when the unit contains a fatal error, such as the unsupported `User=%u` directive, so it does not actually validate that the generated unit is correct. The comment says it "may warn on missing ExecStart deps but should not fail hard", but it does not distinguish warnings from errors. This lets an invalid unit ship without any test failure. The assertion should require returncode 0, or at least parse the output to ensure no error-level messages are present.

TimeToBuildBob:
## 🤖 AI code review

This pull request introduces a new CLI subcommand `gptme service init` to scaffold systemd user services for headless gptme agents. It generates systemd service and timer unit files, a startup shell script, and initial configuration files (`gptme.toml`, `AGENTS.md`) within a specified work directory. The `pyproject.toml` file is updated to register this new command, and a new test file verifies the generated files and command behavior.

- `gptme/cli/cmd_service.py` — Adds the `cli` group and `init` command for managing headless gptme agent services, including template generation and systemd value escaping functions.
- `pyproject.toml` — Registers the `gptme-service` entry point to expose the new CLI command.
- `tests/test_cmd_service.py` — Adds comprehensive tests for the `gptme service init` command, covering file generation, script executability, systemd unit parsing, and error handling.

</details>

| `49c28fa83d10` | 3/5 | 2 | llm | 2026-08-19 20:51 UTC |
| `69d6c7fb969b` | 3/5 | 3 | llm | 2026-08-19 21:01 UTC |
| `b26c46bbd6df` | 4/5 | 2 | llm | 2026-08-19 21:51 UTC |
| `c00a92e665ac` | 3/5 | 3 | llm | 2026-08-20 00:08 UTC |
| `3d0aa7db2fd6` | 1/5 | 2 | llm | 2026-08-20 02:47 UTC |
| `bc2fbc6e14d0` | ? | 3 | llm | 2026-08-20 09:42 UTC |
| `b7d0acc26f13` | 3/5 | 2 | llm | 2026-08-20 13:01 UTC |
| `6c2915166518` | 4/5 | 2 | llm | 2026-08-20 14:24 UTC |
| `9131ceec8dd4` | 3/5 | 1 | llm | 2026-08-20 22:01 UTC |
| `9131ceec8dd4` | 3/5 | 1 | llm | 2026-08-21 14:06 UTC |

</details>

<sub>Reviewed `9131ceec8dd4` · `openrouter/google/gemini-2.5-flash` · `llm` engine · 35s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,345 @@
...
+    """
+    work = _resolve_work_dir(work_dir)
+    out_dir = Path(output_dir).expanduser().resolve()
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    click.echo(f"Scaffolding headless agent '{name}'...")
+
+    # Service unit
+    _write_file(
+        out_dir / f"{name}.service",
```
TimeToBuildBob, line 307: ⚠️ **P2** — The `--name` option is interpolated directly into file paths via `out_dir / f"{name}.service"` and later `out_dir / f"{name}.timer"`. If the name contains path separators (e.g., `/`) or `..`, the generated unit file can be written outside the intended output directory. For example, `gptme service init --name ../../tmp/evil --output-dir ~/.config/systemd/user` writes `evil.service` into `~/.config/systemd/user/../../tmp/`, escaping the systemd config directory and potentially overwriting arbitrary files, especially with `--force`. The name should be validated to a safe pattern like `[A-Za-z0-9_-]` before being used in path construction.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,345 @@
...
+    "weekly": "Mon *-*-* 00:00:00",
+}
+
+STARTUP_SCRIPT_TEMPLATE = """\
+#!/bin/bash
+# gptme autonomous agent startup script.
+# Generated by: gptme service init --name {name}
+set -euo pipefail
+
+AGENT_NAME="{name}"
```
TimeToBuildBob, line 112: ❌ **P1** — The `--name` value is inserted verbatim into the generated `gptme-agent-run.sh` inside double quotes. A name containing a double quote or shell metacharacters breaks out of the string and injects commands that execute when the script runs. For instance, `--name 'x"; touch /tmp/pwned; #'` yields `AGENT_NAME="x"; touch /tmp/pwned; #"`, which runs `touch` on service start. The same unsanitized interpolation affects the systemd unit (e.g., `Description=`), `gptme.toml`, and `AGENTS.md`, producing invalid config files. Values should be shell-escaped (e.g., with `shlex.quote`) or restricted to a safe charset.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,338 @@
...
+    if timer_schedule == "on-demand":
+        stale_timer = out_dir / f"{name}.timer"
+        if stale_timer.exists():
+            if force:
+                # Disable the unit BEFORE removing the file so systemd can
+                # resolve and stop the unit while the file is still present.
+                click.echo(
+                    f"  Disabling loaded timer unit {name}.timer (on-demand mode, --force)"
+                )
+                subprocess.run(
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b26c46b — `stale_timer.unlink()` is now inside the `else:` branch and only executes when `result.returncode == 0`. A failed disable prints a warning but leaves the file in place so the loaded unit remains resolvable for manual cleanup. New test `test_on_demand_force_preserves_timer_when_disable_fails` covers this path.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,345 @@
...
+            if force:
+                # Disable the unit BEFORE removing the file so systemd can
+                # resolve and stop the unit while the file is still present.
+                click.echo(
+                    f"  Disabling loaded timer unit {name}.timer (on-demand mode, --force)"
+                )
+                result = subprocess.run(
+                    ["systemctl", "--user", "disable", "--now", f"{name}.timer"],
+                    check=False,  # best-effort; unit may not be enabled
+                )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b26c46b — the `subprocess.run` call is now wrapped in `try/except FileNotFoundError`. When `systemctl` is unavailable the code emits a clear warning and continues generating all workspace files (service unit, gptme.toml, AGENTS.md, startup script) without crashing. New test `test_on_demand_force_handles_missing_systemctl` covers this path.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,334 @@
...
+        SYSTEMD_SERVICE_TEMPLATE.format(name=name, model=model, work_dir=work),
+        force=force,
+    )
+
+    # Timer (skip for on-demand; remove stale timer only with --force)
+    if timer_schedule == "on-demand":
+        stale_timer = out_dir / f"{name}.timer"
+        if stale_timer.exists():
+            if force:
+                stale_timer.unlink()
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 49c28fa — the force path runs `systemctl --user disable --now` via `subprocess.run` first (while the file still exists on disk), then removes the file only if disable succeeded. The order is: disable → remove, not remove → disable.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,330 @@
...
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 49c28fa — the force path runs `systemctl --user disable --now` via `subprocess.run` first (while the file still exists on disk), then removes the file only if disable succeeded. The order is: disable → remove, not remove → disable.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,310 @@
...
+    )
+
+    # Timer (skip for on-demand)
+    if timer_schedule != "on-demand":
+        schedule = SCHEDULES.get(timer_schedule, SCHEDULES["daily"])
+        _write_file(
+            out_dir / f"{name}.timer",
+            SYSTEMD_TIMER_TEMPLATE.format(name=name, schedule=schedule),
+            force=force,
+        )
```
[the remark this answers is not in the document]
TimeToBuildBob: By design — the no-force path (lines 292-299) preserves an existing timer and emits a warning with the exact `systemctl --user disable --now` command so the operator can stop it manually. Removing it silently without `--force` would be surprising. The `--force` path now disables-then-removes (see b26c46b for the fixed version).

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,319 @@
...
+        out_dir / f"{name}.service",
+        SYSTEMD_SERVICE_TEMPLATE.format(name=name, model=model, work_dir=work),
+        force=force,
+    )
+
+    # Timer (skip for on-demand; remove stale timer if switching to on-demand)
+    if timer_schedule == "on-demand":
+        stale_timer = out_dir / f"{name}.timer"
+        if stale_timer.exists():
+            stale_timer.unlink()
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — in the current code the no-`--force` branch (line 292, `else:`) only calls `click.echo` to print a warning and the disable command; it never calls `stale_timer.unlink()`. The test `test_on_demand_preserves_existing_timer_without_force` verifies this: the timer file exists after the call returns.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,310 @@
...
+
+[Install]
+WantedBy=timers.target
+"""
+
+# OnCalendar expressions for the named schedules we support.
+SCHEDULES: dict[str, str] = {
+    "hourly": "*-*-* *:00:00",
+    "daily": "*-*-* 00:00:00",
+    "weekly": "*-*-* 00:00:00",
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — `SCHEDULES["weekly"] = "Mon *-*-* 00:00:00"` (line 70), which is a valid systemd `OnCalendar` weekly expression (fires every Monday at midnight). The fallback `SCHEDULES.get(timer_schedule, SCHEDULES["daily"])` is never reached for "weekly" since the key exists in the dict.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,310 @@
...
+RestartMaxDelaySec=60
+
+[Install]
+WantedBy=default.target
+"""
+
+SYSTEMD_TIMER_TEMPLATE = """\
+[Unit]
+Description=gptme Autonomous Agent Timer: {name}
+Requires={name}.service
```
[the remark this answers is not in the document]
TimeToBuildBob: `Requires={name}.service` was removed from the timer template in an earlier iteration of this PR. The current `SYSTEMD_TIMER_TEMPLATE` contains only [Unit], [Timer], and [Install] sections — no service dependency. `enable --now` starts the timer unit only; the service fires at the next OnCalendar event.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,310 @@
...
+After=network-online.target
+
+[Service]
+Type=simple
+User=%u
+WorkingDirectory={work_dir}
+Environment=GPTME_AGENT_NAME={name}
+Environment=GPTME_AGENT_MODEL={model}
+Environment=GPTME_NON_INTERACTIVE=1
+ExecStart={work_dir}/gptme-agent-run.sh
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — the template (line 43) is `ExecStart="{work_dir}/gptme-agent-run.sh"` (double-quoted). Systemd unit files support quoting in `ExecStart` values: a path wrapped in double quotes is treated as a single token even if it contains spaces. This is documented in `systemd.service(5)` under "Command lines".

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,310 @@
...
+        force=force,
+    )
+
+    # Startup script
+    startup = work / "gptme-agent-run.sh"
+    _write_file(
+        startup,
+        STARTUP_SCRIPT_TEMPLATE.format(name=name, work_dir=work),
+        force=force,
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — `startup.chmod(0o755)` is inside `if written:` (lines 327-328). When `_write_file` returns `False` (file was skipped), `written` is `False` and `chmod` is never called. The existing `test_force_overwrites_existing` test verifies that skipped files are not touched.

tests/test_cmd_service.py
TimeToBuildBob: Superseded by a newer review round on commit b26c46b. The finding no longer applies to the current code.

gptme/cli/cmd_service.py
TimeToBuildBob: Superseded by a newer review round on commit b26c46b. The finding no longer applies to the current code.

gptme/cli/cmd_service.py
TimeToBuildBob: Superseded by a newer review round on commit b26c46b. The finding no longer applies to the current code.

TimeToBuildBob:
**Ready for review** — all Greptile P1 threads addressed (2 real bugs fixed, 7 false positives with replies), CI running on b26c46b.

**Bugs fixed in b26c46b:**
- `stale_timer.unlink()` now only executes when `systemctl disable --now` returns 0 (was unconditional; a failed disable would silently remove the timer file, leaving the loaded unit unresolvable)
- `subprocess.run` for systemctl is now wrapped in `try/except FileNotFoundError` — scaffolding continues cleanly on non-systemd environments with a clear warning

3 new tests cover these paths (disable success → file removed; disable fails → file preserved; missing systemctl → file preserved, warning emitted).

Self-merge blocked: `gptme/cli/cmd_service.py` is outside the self-merge allowlist (new module). Greptile auto-reviewed b26c46b but the issue comment score (2/5) wasn't updated; the inline threads are all replied-to.

tests/test_cmd_service.py
```diff
@@ -0,0 +1,269 @@
...
```
TimeToBuildBob: ⚠️ **P2** — The assertion `proc.returncode in (0, 1)` makes the test pass for any unit that systemd-analyze reports with returncode 1. systemd-analyze verify returns 1 for *any* unit error, including syntax errors, so a malformed generated unit (e.g., a name containing spaces) would still yield returncode 1 and the test would pass. The test's stated purpose—verifying the unit parses—is therefore not actually enforced; it only rejects values outside {0,1}, which effectively never fires. A future regression that produces an unparsable unit would go undetected. The fix is to assert returncode == 0, or to inspect stderr for unexpected error lines and only allow known warnings.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,355 @@
...
+    """
+    work = _resolve_work_dir(work_dir)
+    out_dir = Path(output_dir).expanduser().resolve()
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    click.echo(f"Scaffolding headless agent '{name}'...")
+
+    # Service unit
+    _write_file(
+        out_dir / f"{name}.service",
```
TimeToBuildBob, line 307: ⚠️ **P2** — The `--name` option is interpolated directly into file names and unit contents without validation. systemd unit names must match `[a-zA-Z0-9:_.\-]` and must not contain spaces, slashes, or other separators. If a caller passes `--name "my agent"` or `--name "foo/bar"`, the generated unit file is either unparsable or placed in an unexpected subdirectory (for `/`). The same applies to `ExecStart` and `WorkingDirectory` when `--work-dir` contains spaces or quotes: the unit line becomes syntactically invalid. The consequence is that `systemctl --user start` or `--user enable` fails at parse time and the scaffolded agent never runs, with an unclear error. The command should reject names that are not valid unit names and warn or escape work-dir paths in the unit file.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in c00a92e — removed `User=%u` from the service template. You're correct: `User=` is not supported in systemd user units; the user manager rejects the unit at load time with an error. The user-unit service already runs as the invoking user by default, so the directive was both invalid and redundant.

New test `test_service_unit_has_no_user_directive` guards against regression.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in c00a92e — added `_NAME_RE = re.compile(r"^[a-zA-Z0-9_-]+$")` validation in `init()` that rejects names containing shell metacharacters, spaces, or quotes before any file is written. Also applied `shlex.quote(str(work))` when formatting the startup script template so the `WORK_DIR=` assignment is properly shell-quoted for paths with unusual characters.

New test `test_invalid_name_rejected` covers 4 bad-name cases: shell injection, space, slash, and `@`.

TimeToBuildBob:
## 🔍 Convergence adjudication — gptme/gptme#3574

**Two real bugs fixed in c00a92e** (pushed just now; thread replies posted):

### Fixed this session
| Finding | Fix |
|---|---|
| `User=%u` in `[Service]` — invalid in user units; `systemctl --user` rejects such units at load time | Removed from `SYSTEMD_SERVICE_TEMPLATE` |
| Shell injection via `--name` — metacharacters inserted verbatim into generated bash | Added `_NAME_RE = re.compile(r"^[a-zA-Z0-9_-]+$")` validation at CLI level; `shlex.quote` for `work_dir` in startup script |

2 new tests: `test_service_unit_has_no_user_directive`, `test_invalid_name_rejected` (4 bad-name cases).

### Previously fixed (b26c46b, stale Greptile threads)
- `stale_timer.unlink()` only runs when `systemctl disable --now` returns 0
- `FileNotFoundError` from missing `systemctl` is caught; scaffolding continues with a warning

Greptile threads 3816589459 and 3816645954 were raised against older commits and have been replied to with the fix SHA; they haven't been formally resolved (Greptile hasn't re-reviewed c00a92e).

### CI
Pending on c00a92e. Lint + typecheck passed on prior commits.

### Domain risk
New module (`gptme/cli/cmd_service.py`). The generated service unit should be tested with `systemd-analyze verify` on a real user manager before this ships. The `systemctl` path (`subprocess.run`) is covered by mocked tests; the integration smoke relies on `systemd-analyze` being present (skipped otherwise).

### Self-merge
Blocked — new module outside the self-merge allowlist. Needs maintainer review.

**Convergence**: detector reports `stable_rounds=0`, `new_blocking` status. Both open Greptile threads are stale vs current HEAD.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,366 @@
...
+Documentation=https://github.com/gptme/gptme#headless
+After=network-online.target
+
+[Service]
+Type=simple
+WorkingDirectory={work_dir}
+Environment=GPTME_AGENT_NAME={name}
+Environment=GPTME_AGENT_MODEL={model}
+Environment=GPTME_NON_INTERACTIVE=1
+ExecStart="{work_dir}/gptme-agent-run.sh"
```
TimeToBuildBob: ❌ **P1** — The work_dir option is inserted into the systemd service unit template without escaping newlines or percent signs. If work_dir contains a newline, the generated unit file is corrupted: the literal newline terminates the ExecStart line and any trailing text becomes a new directive, allowing injection of arbitrary systemd directives (e.g., ExecStartPre) that execute commands when the service starts as the user. Percent signs are also interpreted as systemd specifiers, so a path like /tmp/%h expands to the home directory, redirecting ExecStart to the wrong script. The startup script template correctly uses shlex.quote for bash, but the unit file does not. This is a security risk when work_dir is influenced by an external source, and a correctness bug for legitimate paths containing % or newlines.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,366 @@
...
+        stale_timer = out_dir / f"{name}.timer"
+        if stale_timer.exists():
+            if force:
+                # Disable the unit BEFORE removing the file so systemd can
+                # resolve and stop the unit while the file is still present.
+                click.echo(
+                    f"  Disabling loaded timer unit {name}.timer (on-demand mode, --force)"
+                )
+                try:
+                    result = subprocess.run(
```
TimeToBuildBob, line 328: ⚠️ **P2** — The subprocess.run call that disables the timer unit has no timeout parameter. If `systemctl --user disable --now` hangs (e.g., the user systemd manager is unresponsive or waiting for a password), the `gptme service init` command blocks indefinitely with no feedback. This can hang CI runs, automated provisioning, or an operator's terminal. The command should pass a timeout (e.g., timeout=30) and handle subprocess.TimeoutExpired by treating it like a failed disable, preserving the timer file and warning the user.

TimeToBuildBob:
CI completed on c00a92e — 2 failures, neither caused by this PR:

1. **`test_renders_and_responds` teardown error** (`KeyError: <_pytest.stash.StashKey ...>`) — pytest infrastructure error during teardown; the test itself passed on retry. Known flaky issue unrelated to `gptme service init` changes.

2. **e2e stable failure** — the `stable` gptme pass failed while `dev` passed; this means the stable PyPI release has an issue, not the branch. Pre-existing upstream failure.

The PR changes (`gptme/cli/cmd_service.py`, new tests, console script registration) don't touch TUI, e2e, or anything these checks exercise. Ready for maintainer review once the flaky re-run issue is addressed.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in 3d0aa7db2 — added `_escape_systemd_value()` which doubles literal `%` (prevents systemd specifier expansion, e.g. `%h` → home dir) and rejects embedded newlines with a `click.BadParameter` (prevents injecting extra unit directives). Applied to `work_dir` when formatting `SYSTEMD_SERVICE_TEMPLATE`. New tests: `test_work_dir_percent_is_escaped_in_unit`, `test_work_dir_newline_rejected`.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in 3d0aa7db2 — added `timeout=30` to the `systemctl --user disable --now` subprocess.run call and a `subprocess.TimeoutExpired` handler that warns and leaves the timer file in place (same fallback as the existing nonzero-returncode path).

tests/test_cmd_service.py
TimeToBuildBob: Fixed in 3d0aa7db2 — `test_systemd_unit_parses` now asserts `proc.returncode == 0` (with stdout/stderr in the failure message) instead of accepting `(0, 1)`, which let any systemd-analyze error pass silently.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,395 @@
...
+    out_dir = Path(output_dir).expanduser().resolve()
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    click.echo(f"Scaffolding headless agent '{name}'...")
+
+    # Service unit
+    _write_file(
+        out_dir / f"{name}.service",
+        SYSTEMD_SERVICE_TEMPLATE.format(
+            name=name, model=model, work_dir=_escape_systemd_value(str(work))
```
TimeToBuildBob: ⚠️ **P2** — Both `model` and `work_dir` are interpolated into templates via Python's `str.format()`. If either contains a literal `{` or `}` (e.g. `--model "gpt-4o-{foo}"`), `.format()` interprets the brace as a placeholder and raises `KeyError` (or `IndexError` for an empty placeholder), causing the command to crash with a traceback. This affects not only the service unit but also the timer, gptme.toml, and AGENTS.md templates, since all use `.format()` with user-supplied values. The work_dir is escaped for `%` and newlines but not for braces, so a work-dir like `/tmp/foo{bar}` similarly crashes. The command should either reject braces early or escape braces by doubling them (e.g. `value.replace('{', '{{').replace('}', '}}')`) before formatting.

TimeToBuildBob:
Rebased the branch onto current `master` (`bc2fbc6e1`) and force-pushed. The failed e2e run had **dev: success** and **stable: failure**; current master already contains the workflow fix that makes the known PyPI-stable regression non-blocking, so no product-code workaround was added here. Fresh CI is running on the rebased head.

gptme/cli/cmd_service.py
TimeToBuildBob: ⚠️ **P2** — The `model` option is passed directly into `SYSTEMD_SERVICE_TEMPLATE.format()` at line 296 without any escaping. Unlike `work_dir`, which goes through `_escape_systemd_value`, `model` is interpolated verbatim into `Environment=GPTME_AGENT_MODEL={model}`. If a model string contains a newline (for example, set via the `GPTME_MODEL` environment variable, which the command reads due to `auto_envvar_prefix`), the remainder of the string becomes new systemd directives in the generated unit file, allowing arbitrary `Environment=` lines to be injected. This can alter the environment of the generated service and potentially cause it to behave unexpectedly. A concrete example: `GPTME_MODEL="gpt-4o-mini\nEnvironment=GPTME_MALICIOUS=1"` would add that directive. Because the operator controls this input, the severity is low, but it is a correctness and robustness gap.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,395 @@
...
+Documentation=https://github.com/gptme/gptme#headless
+After=network-online.target
+
+[Service]
+Type=simple
+WorkingDirectory={work_dir}
+Environment=GPTME_AGENT_NAME={name}
+Environment=GPTME_AGENT_MODEL={model}
+Environment=GPTME_NON_INTERACTIVE=1
+ExecStart="{work_dir}/gptme-agent-run.sh"
```
TimeToBuildBob: ⚠️ **P2** — `_escape_systemd_value` (lines 39-51) rejects newlines and doubles `%`, but it does not escape double quotes. The generated service unit embeds the escaped work_dir inside double quotes on the `ExecStart` line (line 66). If the work directory path contains a double quote (e.g., `/home/user/my"dir`), the quote is not escaped, so the line becomes `ExecStart="/home/user/my"dir/gptme-agent-run.sh"`, which breaks the quoting and results in a malformed unit that systemd fails to parse. The same applies to `WorkingDirectory={work_dir}` (line 62), which is unquoted but still breaks if the path contains a double quote. Users with such a directory would get a unit that cannot be loaded.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,395 @@
...
+                click.echo(
+                    f"  Disabling loaded timer unit {name}.timer (on-demand mode, --force)"
+                )
+                try:
+                    result = subprocess.run(
+                        ["systemctl", "--user", "disable", "--now", f"{name}.timer"],
+                        check=False,  # best-effort; unit may not be enabled
+                        timeout=_SYSTEMCTL_TIMEOUT_SEC,
+                    )
+                except FileNotFoundError:
```
TimeToBuildBob: ⚠️ **P2** — In the on-demand `--force` path, `subprocess.run` for `systemctl --user disable --now` is wrapped in a `try` that catches `FileNotFoundError` and `subprocess.TimeoutExpired`. If `systemctl` exists on the system but is not executable (e.g., a broken installation or a file with no exec permission), `subprocess.run` raises `PermissionError`, which is a subclass of `OSError` but not caught here. The exception propagates out of `init`, so the entire scaffolding command aborts with a traceback instead of printing the intended warning and continuing. The `PermissionError` is not an unlikely scenario on systems where `systemctl` is shadowed by a non-executable script or the PATH is misconfigured.

TimeToBuildBob:
Pushed `b7d0acc26` addressing the latest AI-review findings:

- validate the model before unit generation, rejecting newline injection;
- reject quotes and backslashes in generated systemd executable paths so `ExecStart` cannot become malformed;
- handle all `OSError` failures from `systemctl` (including `PermissionError`) while preserving the timer file.

Added regression coverage for each path. Local verification: 18 service-command tests pass; Ruff and targeted mypy pass. Fresh CI and the AI reviewer are running automatically on the new head.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in `b7d0acc26` — the model now passes through `_escape_systemd_value()` before service-unit interpolation, so CR/LF input is rejected and `%` is escaped. Added `test_model_newline_rejected`, which verifies no service file is written.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in `b7d0acc26` — generated systemd executable paths now reject quotes and backslashes rather than emitting a malformed `WorkingDirectory`/`ExecStart`. The regression test covers double quote, single quote, backslash, and newline inputs.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in `b7d0acc26` — the `systemctl` invocation now catches `OSError`, covering both missing executables and `PermissionError`, warns, and preserves the timer file. The parameterized regression test exercises both exceptions.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,402 @@
...
+SESSION_ID=$(date +%Y%m%d-%H%M%S)
+SESSION_LOG="$JOURNAL_DIR/session-$SESSION_ID.md"
+
+# Durable session record placeholder. Replace the loop body with your
+# agent's real work loop (CASCADE selector + task executor, etc.).
+{{
+  echo "# Session $SESSION_ID"
+  echo
+  echo "Agent: $AGENT_NAME"
+  echo "Started: $(date --iso-8601=seconds)"
```
TimeToBuildBob: ❌ **P1** — Generated startup script uses `date --iso-8601=seconds`, a GNU coreutils extension. On BSD and macOS, `date` rejects this option with a usage error and returns a non-zero status. Because the script runs with `set -euo pipefail`, the `echo "Started: ..."` command fails, and the entire script exits prematurely. The session log is partially written (missing the Started line and closing brace), and the service fails on every timer trigger on these platforms. Replace with a POSIX-compatible format such as `date -u +%Y-%m-%dT%H:%M:%SZ`.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,402 @@
...
+
+    # Startup script
+    startup = work / "gptme-agent-run.sh"
+    written = _write_file(
+        startup,
+        STARTUP_SCRIPT_TEMPLATE.format(name=name, work_dir=shlex.quote(str(work))),
+        force=force,
+    )
+    if written:
+        startup.chmod(0o755)
```
TimeToBuildBob: ⚠️ **P2** — Startup script's executable bit is set only when the file is freshly written (`if written:`). If an existing `gptme-agent-run.sh` from a previous scaffold is not executable (for example after an edit or a failed permission change), rerunning `gptme service init` without `--force` skips writing it and does not fix the permissions. The generated systemd unit's `ExecStart` path then refers to a non-executable file, so the service fails to start with a permission error on every run. The chmod should be applied unconditionally after the write attempt.

TimeToBuildBob:
Pushed `6c2915166` addressing both current AI-review findings:

- generated startup scripts now use portable UTC `date -u +%Y-%m-%dT%H:%M:%SZ` instead of GNU-only `date --iso-8601=seconds`;
- reinitializing now restores the startup script's executable bits even when its existing contents are preserved.

Added regression tests for portability and permission repair. Local verification: 20 service-command tests pass, Ruff passes, Bash syntax passes, and pre-commit passes.

The prior CI failure was unrelated to this PR: `tests/test_llm_openai.py::TestMaybeApplyVerbosity::test_invalid_level_warns_only_once` logged twice despite its once-only flag, while current `master` CI is healthy. The push started fresh CI and the AI reviewer automatically; no Greptile re-review was triggered.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in `6c2915166`: the generated script now uses portable UTC `date -u +%Y-%m-%dT%H:%M:%SZ`, with a regression test that rejects the GNU-only option.

gptme/cli/cmd_service.py
TimeToBuildBob: Fixed in `6c2915166`: `chmod(0o755)` now runs after every write attempt, so an existing preserved startup script has its executable bits repaired. Added a regression test that preserves custom contents while restoring permissions.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,401 @@
...
+    if "\n" in value or "\r" in value:
+        raise click.BadParameter(
+            f"{value!r} must not contain newlines (would corrupt the generated unit file)"
+        )
+    if any(char in value for char in ("\\", '"', "'")):
+        raise click.BadParameter(
+            f"{value!r} must not contain quotes or backslashes "
+            "(unsupported in a systemd executable path)"
+        )
+    return value.replace("%", "%%")
```
TimeToBuildBob, line 56: ⚠️ **P2** — The `_escape_systemd_value` function only doubles `%` characters but does not handle `$`. In systemd unit files, a literal `$` in `ExecStart` (and other directives subject to variable expansion) must be written as `$$`. If a user's `--work-dir` contains a `$` (e.g. `/tmp/foo$bar`), the generated `ExecStart="/tmp/foo$bar/gptme-agent-run.sh"` line will have `$bar` interpreted as an environment variable reference by systemd, which may expand to an empty value or a different path, causing the service to start the wrong executable or fail to start. The same issue affects `WorkingDirectory` if systemd expands `$` there. The fix is to replace `$` with `$$` in the escape function.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,401 @@
...
+[Unit]
+Description=gptme Autonomous Agent: {name}
+Documentation=https://github.com/gptme/gptme#headless
+After=network-online.target
+
+[Service]
+Type=simple
+WorkingDirectory={work_dir}
+Environment=GPTME_AGENT_NAME={name}
+Environment=GPTME_AGENT_MODEL={model}
```
TimeToBuildBob: ⚠️ **P2** — The `model` value is inserted directly into `Environment=GPTME_AGENT_MODEL={model}` without quoting. systemd parses the `Environment=` directive as a space-separated list of `KEY=VALUE` assignments, so a model containing a space (e.g. `gpt-4o mini`) produces `Environment=GPTME_AGENT_MODEL=gpt-4o mini`, which sets the variable to `gpt-4o` and then attempts to parse `mini` as another assignment, causing a unit parse error or an unintended environment variable. Since `_escape_systemd_value` does not reject spaces, the generated unit is invalid. The fix is to wrap the value in double quotes in the template or reject spaces in the escape function.

TimeToBuildBob:
Reran only the failed `Test without API keys` job on the unchanged head. The first attempt ended in pytest-xdist infrastructure failure: `tests/test_server_v2.py::test_v2_conversations_list_keeps_messages_in_fast_mode` exceeded the global 10s timeout, the worker crashed, and xdist then raised an internal error after 7,165 tests had passed. The service-init tests had passed, and current `master` CI is healthy. Attempt 2 is now running; no product-code workaround was added.

TimeToBuildBob:
Pushed `9131ceec8` addressing both current AI-review findings:

- generated `Environment=` assignments now quote the complete `KEY=value`, preserving model names containing spaces;
- literal `$` in a work-directory path is C-escaped only in `ExecStart`, with systemd command-line expansion disabled via the `:` executable prefix. `WorkingDirectory=` keeps the literal path because that directive does not perform environment expansion.

Added regression tests for both cases. Local verification: 22 service-command tests pass, Ruff passes, targeted pre-commit (including mypy) passes, `systemd-analyze verify` accepts a generated unit containing both a `$` path and spaced model, and a live `systemctl --user start` smoke test executed that generated unit and wrote the expected session log. Fresh CI and the AI reviewer started automatically; no Greptile re-review was triggered.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,410 @@
...
+    if "\n" in value or "\r" in value:
+        raise click.BadParameter(
+            f"{value!r} must not contain newlines (would corrupt the generated unit file)"
+        )
+    if any(char in value for char in ("\\", '"', "'")):
+        raise click.BadParameter(
+            f"{value!r} must not contain quotes or backslashes "
+            "(unsupported in a systemd executable path)"
+        )
+    return value.replace("%", "%%")
```
TimeToBuildBob, line 56: ❌ **P1** — The `_escape_systemd_value` function does not escape curly braces (`{` and `}`). Every template is rendered with `.format()` (e.g. `SYSTEMD_SERVICE_TEMPLATE.format(name=name, model=escaped_model, work_dir=escaped_work_dir, exec_work_dir=escaped_exec_work_dir)`), so any value containing a literal brace is interpreted as a placeholder and raises `KeyError` or `ValueError`. For instance `--work-dir '/tmp/foo{bar}'` produces a `KeyError: 'bar'` at the format call, aborting the entire scaffold with a traceback. The same applies to `--model` and to the values passed into `GPTME_TOML_TEMPLATE`, `AGENTS_MD_TEMPLATE`, and `STARTUP_SCRIPT_TEMPLATE`. Since braces are legal in filesystem paths and model strings, this is a real crash for valid input. Fix by doubling braces in the escaping functions so they survive `.format()`, e.g. `value.replace('{', '{{').replace('}', '}}')` after the `%` replacement.

TimeToBuildBob: False positive. Python `str.format()` parses replacement fields in the template before substituting values; braces inside a substituted value are not reparsed. For example, `"x={value}".format(value="foo{bar}")` returns `"x=foo{bar}"` without raising. No code change is needed.

TimeToBuildBob:
The current AI-review P1 is a false positive, not a code defect. Python `str.format()` parses placeholders before substituting values; braces inside a replacement value are not reparsed. I verified the actual CLI with both `--work-dir 'foo{bar}'` and `--model 'model{variant}'`: scaffolding completed and preserved both values. All 22 focused tests and every CI check pass on `9131ceec8`.

No additional product-code change or Greptile trigger was made. The PR remains outside the self-merge allowlist because it adds a new product module and changes `pyproject.toml`, so it needs maintainer review/merge.

gptme/cli/cmd_service.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+def _escape_systemd_exec_value(value: str) -> str:
+    """Escape a validated unit value for an ExecStart command line."""
+    return _escape_systemd_value(value).replace("$", r"\x24")
```
TimeToBuildBob, line 61: ❌ **P1** — The `_escape_systemd_exec_value` function incorrectly escapes literal dollar signs (`$`) in paths for the `ExecStart` directive. The `ExecStart=:` syntax in systemd unit files explicitly states that the executable path is not subject to shell or parameter expansion, meaning a literal `$` in a directory name should be passed as is. By replacing `$` with `\x24`, the generated `ExecStart` path will contain a literal `\x24` sequence, causing systemd to look for a directory with `\x24` in its name rather than `$`, leading to a 'file not found' error when the service attempts to start. _(2/2 passes)_

TimeToBuildBob: False positive. In a systemd command line, `\x24` is parsed by systemd's quoted-string syntax into a literal `$`; it is not passed as four characters. The leading `:` is a documented ExecStart prefix that disables environment-variable substitution. I verified the exact generated unit with `systemd-analyze verify` and a live `systemctl --user start` against `/tmp/gptme-live-$agent`: systemd executed `/tmp/gptme-live-$agent/gptme-agent-run.sh`, which wrote the expected session log. Removing the C escape would reintroduce `$agent` expansion before execution.

ErikBjare:
@TimeToBuildBob See the pre-existing `gptme-agents` which does similar things iirc (and also supports launchd). I've merged this, but I want a follow-up PR which addresses the overlap and improves code reuse (removing redundant/duplicate helpers).

Should also clearly reference gptme-agent-template as the "full" version.

TimeToBuildBob:
Thanks for merging, and noted on the follow-up.

I checked `cmd_agents.py` (the `gptme-util agents scan` surface), `gptme-agent-template`'s dotfiles, and the launchd angle. The follow-up PR will:

1. **Reference gptme-agent-template as the full version** — update the `gptme service init` help text, generated `AGENTS.md`, and README to point to `gptme/gptme-agent-template` for production-grade setups. The current command is intentionally minimal; gptme-agent-template is the batteries-included fork.

2. **Remove duplicate inline templates** — `cmd_service.py` currently has the systemd unit content inline as Python string templates. These should be refactored to either reference the `.example` files in gptme-agent-template (as an optional source) or at minimum share the minimal skeleton with them so both stay in sync.

3. **Add launchd support (macOS)** — gptme-agent-template's dotfiles presumably handle launchd; `cmd_service.py` is systemd-only. The follow-up PR will add a `--platform` flag (or auto-detect) to generate a `launchd` plist alongside the systemd units.

I'll open the follow-up PR in a separate branch. If you have a specific file in gptme-agent-template that already encodes the canonical launchd plist (so I'm not starting from scratch), let me know and I'll reuse it.

PR #3577, opened 2026-08-20
feat(computer): allow out-of-tree transports via register_transport()

TimeToBuildBob:
## Problem

`get_transport()` resolves `GPTME_COMPUTER_TRANSPORT` through a hardcoded
if/elif over `native` and `cua`. A `ComputerTransport` implemented **outside**
gptme cannot be selected at all — any other value falls through to the
"Unknown GPTME_COMPUTER_TRANSPORT" warning and silently uses native.

That makes the ABC only half-pluggable: you can implement the interface, but
you can't get gptme to actually use your implementation. The only way in today
is to patch a branch into gptme's factory, which would mean gptme importing and
depending on the out-of-tree transport (often platform-specific).

## Change

A small `name -> factory` registry:

```python
from gptme.tools.computer_transport import register_transport

register_transport("phone-harness", PhoneComputerTransport)
# GPTME_COMPUTER_TRANSPORT=phone-harness now selects it
```

Design points:

- **Lazy factory** — called only when the env var actually names that transport,
  so a heavy or platform-specific import stays off the default path.
- **Failure is non-fatal** — a factory that raises falls back to native and is
  retried on the next call, exactly matching the existing `cua` behaviour. A
  third-party transport must not be able to break the computer tool.
- **Built-in names are reserved** — registering `native`/`cua` raises
  `ValueError` rather than silently shadowing them (a silently-ignored
  registration is the worse failure).
- Behaviour for unregistered names is unchanged: warn, fall back to native.

Pairs naturally with the plugin system — a `GptmePlugin`'s `init` hook is a
good place to call `register_transport`.

## How to verify

```bash
pytest tests/test_computer_transport.py -q
```

8 new tests in `TestTransportRegistry` cover selection by env var, factory
laziness, caching, retry-on-failure with native fallback, reserved-name and
empty-name rejection, re-registration, and the unchanged unknown-name path.

Confirmed the tests actually exercise the wiring: with the new `elif` branch
removed, 5 of the 8 fail (the other 3 cover `register_transport` validation and
existing fallback behaviour, so they correctly still pass).

`ruff check` / `ruff format` clean; `mypy` reports no new errors in the module.

### DIFF
--- a/gptme/tools/computer_transport.py
+++ b/gptme/tools/computer_transport.py
@@ -765,10 +765,23 @@ def get_transport() -> ComputerTransport | None:
     if _transport is not None and _transport_name == current:
         return _transport
 
-    # Env var changed — close the stale transport to avoid resource leaks
+    # Env var changed — clear cache before fallible cleanup so a broken
+    # third-party close() cannot preserve a stale, partially closed instance.
     if _transport is not None and _transport_name != current:
-        _transport.close()
+        stale_transport = _transport
+        stale_name = _transport_name
         _transport = None
+        _transport_name = None
+        try:
+            stale_transport.close()
+        except Exception:
+            import logging
+
+            logging.getLogger(__name__).exception(
+                "Failed to close transport %r while switching to %r",
+                stale_name,
+                current,
+            )
 
     # No transport requested
     if not current:
@@ -795,7 +808,13 @@ def get_transport() -> ComputerTransport | None:
             _transport_name = None
     elif current in _transport_registry:
         try:
-            _transport = _transport_registry[current]()
+            transport = _transport_registry[current]()
+            if not isinstance(transport, ComputerTransport):
+                raise TypeError(
+                    f"factory for transport {current!r} returned "
+                    f"{type(transport).__name__}, expected ComputerTransport"
+                )
+            _transport = transport
             _transport_name = current  # only cache on success
         except Exception as e:
             import logging
--- a/gptme/tools/computer_transport.py
+++ b/gptme/tools/computer_transport.py
@@ -719,13 +719,21 @@ def register_transport(name: str, factory: Callable[[], ComputerTransport]) -> N
             f"cannot register {name!r}: shadows a built-in gptme transport"
         )
 
-    # Re-registering the selected transport must make the replacement effective
-    # immediately rather than leaving the old name-keyed instance cached.
-    if name in _transport_registry and _transport_name == name:
-        if _transport is not None:
-            _transport.close()
+    # Registering the selected name must make the factory effective immediately,
+    # including when the current cache is the native fallback for an unknown name.
+    if _transport_name == name:
+        stale_transport = _transport
         _transport = None
         _transport_name = None
+        if stale_transport is not None:
+            try:
+                stale_transport.close()
+            except Exception:
+                import logging
+
+                logging.getLogger(__name__).exception(
+                    "Failed to close transport %r while replacing its factory", name
+                )
 
     _transport_registry[name] = factory
 
--- a/gptme/tools/computer_transport.py
+++ b/gptme/tools/computer_transport.py
@@ -6,6 +6,8 @@ to dispatch through different backends:
 
 - **Native** (xdotool+scrot or cliclick) — default, zero-dependency
 - **Cua Sandbox** — opt-in via ``GPTME_COMPUTER_TRANSPORT=cua`` env var
+- **Out-of-tree** — any transport registered via :func:`register_transport`,
+  selected by its registered name
 
 Usage::
 
@@ -32,7 +34,7 @@ from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
     import concurrent.futures
-    from collections.abc import Coroutine
+    from collections.abc import Callable, Coroutine
 
 
 class ComputerTransport(abc.ABC):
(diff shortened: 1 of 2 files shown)


tests/test_computer_transport.py
```diff
@@ -842,3 +843,119 @@ def test_scroll_direction_case_insensitive(self):
 
 if __name__ == "__main__":
     unittest.main()
+
+
+class TestTransportRegistry(unittest.TestCase):
```
TimeToBuildBob, line 844: ⚠️ **P2** — The new `TestTransportRegistry` class is defined after the `if __name__ == "__main__":` block at lines 844-845. When the test file is run directly with `python tests/test_computer_transport.py`, `unittest.main()` executes while `TestTransportRegistry` has not yet been defined, so it is never discovered and its eight new tests do not run. The suite therefore passes even if the registry logic is broken when executed in this conventional mode, giving a false green. The tests still run under pytest because that collection scans the whole module after loading, but the direct-run path silently omits them. Move the class definition above the main guard so both invocation methods execute the new tests.

TimeToBuildBob:
## 🤖 AI code review

This PR adds a `register_transport` function and a module-level registry that lets out-of-tree transports be selected via `GPTME_COMPUTER_TRANSPORT`. It modifies `get_transport()` to consult the registry after the built-in `native` and `cua` branches, with fallback to native on factory failure. Tests for the registry are added.

**⚠️ P2 medium** — `gptme/tools/computer_transport.py:798`

The factory's return value is assigned to `_transport` without any check that it is a valid `ComputerTransport` instance. If a registered factory returns `None` (the type hint is the only contract), `_transport` becomes `None` while `_transport_name` is set to `current`. On the next call, the cache check `if _transport is not None and _transport_name == current` fails because `_transport` is None, so `get_transport()` retries the factory on every call and ultimately returns `None` to the caller. Callers of `get_transport()` treat `None` as "no transport configured" and silently fall back to the built-in native path, so a misbehaving plugin disables itself without any error. This violates the documented contract that selecting a registered transport returns that transport, and makes debugging a faulty plugin harder. A guard that raises `TypeError` or logs an error when the factory returns a non-`ComputerTransport` would surface the problem.

```suggestion
if not isinstance(result, ComputerTransport):
    raise TypeError(f"factory for transport {current!r} returned {type(result).__name__}, expected ComputerTransport")
```

<sub>**How this was verified:** Checked the code around line 798 and the cache logic at lines 765-766; no validation of the factory result exists.</sub>

</details>

- `gptme/tools/computer_transport.py` — Added `register_transport`, `_transport_registry`, `_BUILTIN_TRANSPORTS`, and modified `get_transport` to handle registered names.
- `tests/test_computer_transport.py` — Added `TestTransportRegistry` with 10 tests covering selection, laziness, caching, fallback, validation, and replacement.

</details>

| `f82254e4b05a` | 4/5 | 1 | llm | 2026-08-20 06:05 UTC |
| `b085325512c4` | 3/5 | 1 | llm | 2026-08-20 07:44 UTC |

</details>

<sub>Reviewed `9aefd9a25e2e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 169s · reports findings; merge decisions belong to the gate · [bob#1122](https://github.com/ErikBjare/bob/issues/1122)</sub>

gptme/tools/computer_transport.py
```diff
@@ -679,6 +681,43 @@ def close(self) -> None:
...
+    :raises ValueError: if ``name`` is empty or shadows a built-in transport.
+    """
+    name = name.strip()
+    if not name:
+        raise ValueError("transport name must be non-empty")
+    if name in _BUILTIN_TRANSPORTS:
+        raise ValueError(
+            f"cannot register {name!r}: shadows a built-in gptme transport"
+        )
+    _transport_registry[name] = factory
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b08532551: re-registering the selected name now closes and clears its cached instance before replacing the factory. Added a regression test that caches the original, re-registers, verifies the replacement is returned, and verifies the original was closed.

tests/test_computer_transport.py
TimeToBuildBob: Fixed in b08532551 by moving TestTransportRegistry above the __main__ guard. Verified the direct unittest path now discovers all 52 tests, including the 8 registry tests.

TimeToBuildBob:
@greptileai review

gptme/tools/computer_transport.py
```diff
@@ -679,6 +681,54 @@ def close(self) -> None:
...
+    if not name:
+        raise ValueError("transport name must be non-empty")
+    if name in _BUILTIN_TRANSPORTS:
+        raise ValueError(
+            f"cannot register {name!r}: shadows a built-in gptme transport"
+        )
+
+    # Re-registering the selected transport must make the replacement effective
+    # immediately rather than leaving the old name-keyed instance cached.
+    if name in _transport_registry and _transport_name == name:
```
TimeToBuildBob: ❌ **P1** — When an env var names a transport that is not yet registered, `get_transport()` falls back to native and sets `_transport_name` to that name (line 814). If a plugin later calls `register_transport(name, factory)`, the cache-invalidation guard `name in _transport_registry and _transport_name == name` is False because the name was not in the registry at that time. The cached native transport (`_transport_name` still equal to `current`) therefore survives, and all subsequent `get_transport()` calls hit the cache and return native, never invoking the newly registered factory. This silently ignores the registration and contradicts the PR's stated purpose that registering a transport makes it selectable by env var. The condition must invalidate whenever `_transport_name` matches the name, regardless of prior registry membership.

gptme/tools/computer_transport.py
```diff
@@ -679,6 +681,54 @@ def close(self) -> None:
...
+        if _transport is not None:
+            _transport.close()
+        _transport = None
+        _transport_name = None
+
+    _transport_registry[name] = factory
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9aefd9a25: cache state is cleared before fallible cleanup, and a close exception is logged without preventing the replacement factory from being installed. Added a regression test with a raising close().

gptme/tools/computer_transport.py
TimeToBuildBob: Fixed in 9aefd9a25: registration now invalidates whenever the cached _transport_name matches, including a native fallback cached before the name was registered. Added a late-registration regression test.

gptme/tools/computer_transport.py
```diff
@@ -731,6 +793,23 @@ def get_transport() -> ComputerTransport | None:
             # Leave _transport_name as None so the next call retries CuaComputerTransport
             # (allows recovery from transient failures like Docker not yet running)
             _transport_name = None
+    elif current in _transport_registry:
+        try:
+            _transport = _transport_registry[current]()
```
TimeToBuildBob: ⚠️ **P2** — The factory's return value is assigned to `_transport` without any check that it is a valid `ComputerTransport` instance. If a registered factory returns `None` (the type hint is the only contract), `_transport` becomes `None` while `_transport_name` is set to `current`. On the next call, the cache check `if _transport is not None and _transport_name == current` fails because `_transport` is None, so `get_transport()` retries the factory on every call and ultimately returns `None` to the caller. Callers of `get_transport()` treat `None` as "no transport configured" and silently fall back to the built-in native path, so a misbehaving plugin disables itself without any error. This violates the documented contract that selecting a registered transport returns that transport, and makes debugging a faulty plugin harder. A guard that raises `TypeError` or logs an error when the factory returns a non-`ComputerTransport` would surface the problem.

TimeToBuildBob:
@greptileai review

gptme/tools/computer_transport.py
TimeToBuildBob: Fixed in 7008ad2c6: registered factory results are now validated before caching. A wrong-type result enters the existing logged native-fallback/retry path, and a regression test covers a factory returning None.

gptme/tools/computer_transport.py
```diff
@@ -731,6 +793,23 @@ def get_transport() -> ComputerTransport | None:
...
+            _transport_name = current  # only cache on success
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 7008ad2c6: switching now clears both cache fields before calling the stale transport close hook, logs cleanup failures, and continues selecting the requested replacement. Added a regression test that verifies a raising close cannot preserve the old instance.

TimeToBuildBob:
@greptileai review

PR #3586, opened 2026-08-22
feat(service-init): add launchd support for macOS agents

TimeToBuildBob:
Follow-up to #3574, addressing Erik's feedback to add macOS support.

## Summary

- Added `--platform` flag (linux/macos/auto) for cross-platform scaffolding
- Auto-detect platform (macOS/Linux) — defaults to current system or explicit choice
- **launchd plist generation** for macOS agents with configurable scheduling (hourly/daily/weekly/on-demand)
- Logs directory creation for plist output redirection
- Updated docstrings to reference gptme-agent-template as the full version
- Comprehensive tests for launchd XML validity, env variables, and scheduling

### launchd Support
- New LAUNCHD_PLIST_TEMPLATE with proper XML structure and env variable configuration
- StartInterval scheduling (3600s hourly, 86400s daily, 604800s weekly)
- RunAtLoad flag for on-demand/immediate startup
- Plist filename follows macOS convention: com.gptme.{name}.plist

### Platform Handling
- Helper functions: _detect_platform() and _generate_launchd_plist()
- Default output directory logic: ~/.config/systemd/user (Linux) or ~/Library/LaunchAgents (macOS)
- Platform-specific final instructions in CLI output

### Tests Added
- 6 new tests for plist generation, XML validity, env variables, and scheduling

### DIFF
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -476,6 +476,11 @@ def init(
                 "Use a path without control characters.",
                 param_hint="'--work-dir'",
             )
+    else:
+        # Validate the work-dir for systemd BEFORE creating dirs — _escape_systemd_value
+        # raises BadParameter for quotes/backslashes/newlines, and we must not leave
+        # a stray empty directory behind when that validation fires.
+        _escape_systemd_value(str(work))
 
     # All validation passed — now create directories and generate files.
     work.mkdir(parents=True, exist_ok=True)
@@ -609,7 +614,7 @@ def init(
 
     if platform_choice == "macos":
         plist_file = out_dir / f"com.gptme.{name}.plist"
-        click.echo(f"  launchctl load {plist_file}")
+        click.echo(f'  launchctl load "{plist_file}"')
         click.echo(f"  launchctl start com.gptme.{name}")
         click.echo(
             f"  log stream --predicate 'eventMessage contains[cd] \"{name}\"'  # follow logs"
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -34,6 +34,7 @@ import re
 import shlex
 import subprocess
 from pathlib import Path
+from xml.sax.saxutils import escape as _xml_escape
 
 import click
 
@@ -276,12 +277,13 @@ def _generate_launchd_plist(
         interval = LAUNCHD_SCHEDULE_INTERVALS.get(timer_schedule, 86400)
         schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
 
-    run_at_load = "true" if timer_schedule == "on-demand" else "false"
+    # on-demand = manual start only (RunAtLoad=false); scheduled = start on load too
+    run_at_load = "false" if timer_schedule == "on-demand" else "true"
 
     return LAUNCHD_PLIST_TEMPLATE.format(
-        name=name,
-        model=model,
-        work_dir=work_dir,
+        name=_xml_escape(name),
+        model=_xml_escape(model),
+        work_dir=_xml_escape(work_dir),
         schedule_section=schedule_section,
         run_at_load=run_at_load,
     )
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -249,10 +249,8 @@ Never modify historical entries.
 
 
 def _resolve_work_dir(work_dir: str) -> Path:
-    """Resolve the agent work directory, creating it if needed."""
-    path = Path(work_dir).expanduser().resolve()
-    path.mkdir(parents=True, exist_ok=True)
-    return path
+    """Resolve the agent work directory path (does not create it)."""
+    return Path(work_dir).expanduser().resolve()
 
 
 def _write_file(path: Path, content: str, force: bool = False) -> bool:
@@ -445,15 +443,9 @@ def init(
         else:
             output_dir = "~/.config/systemd/user"
 
-    work = _resolve_work_dir(work_dir)
-    out_dir = Path(output_dir).expanduser().resolve()
-    out_dir.mkdir(parents=True, exist_ok=True)
-
-    click.echo(f"Scaffolding headless agent '{name}' ({platform_choice})...")
-
-    # Validate model consistently across platforms — systemd rejects
-    # newlines/quotes/backslashes; launchd must too (else the same input is
-    # silently truncated on macOS but rejected on Linux, which is confusing
+    # Validate model consistently across platforms BEFORE creating any directories.
+    # systemd rejects newlines/quotes/backslashes; launchd must too (else the same
+    # input is silently truncated on macOS but rejected on Linux, which is confusing
     # and allows injection attempts to go unnoticed on macOS).
     if "\n" in model or "\r" in model:
         raise click.BadParameter(
@@ -466,11 +458,15 @@ def init(
             param_hint="'--model'",
         )
 
+    # Resolve work dir path for validation (no mkdir yet — avoid leaving behind
+    # side-effect directories when subsequent validation raises BadParameter).
+    work = _resolve_work_dir(work_dir)
+
     # Platform-specific service generation
     if platform_choice == "macos":
-        # Reject paths that contain XML 1.0-forbidden characters. Unlike model
-        # (an env-var string where silent stripping is harmless), the work-dir
-        # path must match the one written to disk — silently sanitizing it in
+        # Reject paths that contain XML 1.0-forbidden characters BEFORE creating dirs.
+        # Unlike model (an env-var string where silent stripping is harmless), the
+        # work-dir path must match the one written to disk — silently sanitizing it in
         # the plist would cause launchd to reference a nonexistent directory.
         resolved_work = str(work)
         if _XML10_FORBIDDEN.search(resolved_work):
@@ -481,6 +477,14 @@ def init(
                 param_hint="'--work-dir'",
             )
 
+    # All validation passed — now create directories and generate files.
+    work.mkdir(parents=True, exist_ok=True)
+    out_dir = Path(output_dir).expanduser().resolve()
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    click.echo(f"Scaffolding headless agent '{name}' ({platform_choice})...")
+
+    if platform_choice == "macos":
         # launchd: create logs directory and plist
         logs_dir = work / "logs"
         logs_dir.mkdir(parents=True, exist_ok=True)
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -301,8 +301,8 @@ def _generate_launchd_plist(
 
     return LAUNCHD_PLIST_TEMPLATE.format(
         name=_xml_safe(name),
-        model=_xml_safe(model),
-        work_dir=_xml_safe(work_dir),
+        model=_xml_escape(model),
+        work_dir=_xml_escape(work_dir),
         schedule_section=schedule_section,
         run_at_load=run_at_load,
     )
@@ -464,10 +464,19 @@ def init(
 
     # Platform-specific service generation
     if platform_choice == "macos":
+        # Reject model identifiers that contain XML 1.0-forbidden characters.
+        # Silently stripping them would cause the agent to run with a different
+        # (or nonexistent) model than the user requested.
+        if _XML10_FORBIDDEN.search(model):
+            raise click.BadParameter(
+                f"Model {model!r} contains XML 1.0-forbidden characters "
+                "(control characters or lone surrogates) that cannot be "
+                "represented in a launchd plist. Use a valid model identifier.",
+                param_hint="'--model'",
+            )
         # Reject paths that contain XML 1.0-forbidden characters BEFORE creating dirs.
-        # Unlike model (an env-var string where silent stripping is harmless), the
-        # work-dir path must match the one written to disk — silently sanitizing it in
-        # the plist would cause launchd to reference a nonexistent directory.
+        # The work-dir path must match the one written to disk — silently sanitizing
+        # it in the plist would cause launchd to reference a nonexistent directory.
         resolved_work = str(work)
         if _XML10_FORBIDDEN.search(resolved_work):
             raise click.BadParameter(
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -37,8 +37,9 @@ from pathlib import Path
 from xml.sax.saxutils import escape as _xml_escape
 
 # XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
-# Strip everything else (control characters that survive saxutils.escape and corrupt plists).
-_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F￾￿]")
+# Strip control characters and lone UTF-16 surrogates — both survive saxutils.escape unchanged
+# and cause launchctl to reject the generated plist.
+_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F\uD800-\uDFFF￾￿]")
 
 
 def _xml_safe(value: str) -> str:
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -269,11 +269,21 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
 
 
 def _detect_platform() -> str:
-    """Detect the current platform: 'macos' or 'linux'."""
-    system = platform.system().lower()
-    if system == "darwin":
+    """Detect the current platform: 'macos' or 'linux'.
+
+    Raises UsageError for platforms where neither systemd nor launchd applies
+    (e.g. Windows, FreeBSD). Use --platform explicitly on those systems.
+    """
+    system = platform.system()
+    if system == "Darwin":
         return "macos"
-    return "linux"
+    if system == "Linux":
+        return "linux"
+    raise click.UsageError(
+        f"Platform {system!r} is not supported by 'gptme service init'. "
+        "Use --platform linux or --platform macos to generate service files "
+        "for a supported platform."
+    )
 
 
 def _generate_launchd_plist(
@@ -416,6 +426,17 @@ def init(
     # Resolve platform (auto-detect if needed)
     if platform_choice == "auto":
         platform_choice = _detect_platform()
+        if platform_choice == "macos" and output_dir is None:
+            # Inform users explicitly: on macOS we now default to launchd
+            # instead of the old systemd default, so anyone with scripts
+            # expecting ~/.config/systemd/user gets a clear heads-up.
+            click.echo(
+                "Note: macOS detected — generating a launchd plist in "
+                "~/Library/LaunchAgents instead of a systemd unit. "
+                "Pass --platform linux to generate systemd units for a "
+                "Linux target.",
+                err=True,
+            )
 
     # Set default output directory based on platform
     if output_dir is None:
@@ -429,6 +450,21 @@ def init(
     out_dir.mkdir(parents=True, exist_ok=True)
 
     click.echo(f"Scaffolding headless agent '{name}' ({platform_choice})...")
+
+    # Validate model consistently across platforms — systemd rejects
+    # newlines/quotes/backslashes; launchd must too (else the same input is
+    # silently truncated on macOS but rejected on Linux, which is confusing
+    # and allows injection attempts to go unnoticed on macOS).
+    if "\n" in model or "\r" in model:
+        raise click.BadParameter(
+            f"{model!r} must not contain newlines",
+            param_hint="'--model'",
+        )
+    if any(ch in model for ch in ("\\", '"', "'")):
+        raise click.BadParameter(
+            f"{model!r} must not contain quotes or backslashes",
+            param_hint="'--model'",
+        )
 
     # Platform-specific service generation
     if platform_choice == "macos":
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -36,6 +36,16 @@ import subprocess
 from pathlib import Path
 from xml.sax.saxutils import escape as _xml_escape
 
+# XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
+# Strip everything else (control characters that survive saxutils.escape and corrupt plists).
+_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F￾￿]")
+
+
+def _xml_safe(value: str) -> str:
+    """Strip XML 1.0-forbidden control characters, then XML-escape special chars."""
+    return _xml_escape(_XML10_FORBIDDEN.sub("", value))
+
+
 import click
 
 # Allowed characters for the agent name: matches systemd/launchd unit-name conventions.
@@ -281,9 +291,9 @@ def _generate_launchd_plist(
     run_at_load = "false" if timer_schedule == "on-demand" else "true"
 
     return LAUNCHD_PLIST_TEMPLATE.format(
-        name=_xml_escape(name),
-        model=_xml_escape(model),
-        work_dir=_xml_escape(work_dir),
+        name=_xml_safe(name),
+        model=_xml_safe(model),
+        work_dir=_xml_safe(work_dir),
         schedule_section=schedule_section,
         run_at_load=run_at_load,
     )
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -416,9 +416,6 @@ def init(
     work = _resolve_work_dir(work_dir)
     out_dir = Path(output_dir).expanduser().resolve()
     out_dir.mkdir(parents=True, exist_ok=True)
-    escaped_work_dir = _escape_systemd_value(str(work))
-    escaped_exec_work_dir = _escape_systemd_exec_value(str(work))
-    escaped_model = _escape_systemd_value(model)
 
     click.echo(f"Scaffolding headless agent '{name}' ({platform_choice})...")
 
@@ -441,6 +438,11 @@ def init(
         )
     else:
         # systemd: service unit + optional timer
+        # Apply systemd-specific escaping only in this branch — these chars
+        # are forbidden in systemd unit values but are valid in launchd plists.
+        escaped_work_dir = _escape_systemd_value(str(work))
+        escaped_exec_work_dir = _escape_systemd_exec_value(str(work))
+        escaped_model = _escape_systemd_value(model)
         _write_file(
             out_dir / f"{name}.service",
             SYSTEMD_SERVICE_TEMPLATE.format(
--- a/gptme/cli/cmd_service.py
+++ b/gptme/cli/cmd_service.py
@@ -39,7 +39,7 @@ from xml.sax.saxutils import escape as _xml_escape
 # XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
 # Strip control characters and lone UTF-16 surrogates — both survive saxutils.escape unchanged
 # and cause launchctl to reject the generated plist.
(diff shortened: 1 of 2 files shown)


gptme/cli/cmd_service.py
```diff
@@ -200,6 +256,37 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+    work_dir: str,
+    timer_schedule: str,
+) -> str:
+    """Generate a launchd .plist file for macOS agents."""
+    schedule_section = ""
+    if timer_schedule != "on-demand":
+        interval = LAUNCHD_SCHEDULE_INTERVALS.get(timer_schedule, 86400)
+        schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
+
+    run_at_load = "true" if timer_schedule == "on-demand" else "false"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 106065db — inverted the condition: . On-demand now stays silent until `launchctl start` is called explicitly.

gptme/cli/cmd_service.py
```diff
@@ -200,6 +256,37 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+
+    return LAUNCHD_PLIST_TEMPLATE.format(
+        name=name,
+        model=model,
+        work_dir=work_dir,
+        schedule_section=schedule_section,
+        run_at_load=run_at_load,
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 106065db — added `from xml.sax.saxutils import escape as _xml_escape` and wrapped `name`, `model`, and `work_dir` in `_xml_escape()` before interpolating into the plist template.

TimeToBuildBob:
## 🤖 AI code review

Adds launchd support for macOS to the `gptme service init` command. Introduces a `--platform` flag (linux/macos/auto) with auto-detection, generates a launchd plist for macOS, and moves work-dir directory creation later. Adds XML-forbidden character validation for macOS paths/models and new tests for the launchd path.

**❌ P1 high** — `gptme/cli/cmd_service.py:426`

When `--platform auto` is the default (it is the default value of the option) and the host OS is not Linux or macOS (e.g. Windows, FreeBSD), `_detect_platform()` raises `click.UsageError`. Previously the `init` command always generated systemd units regardless of the OS. This is a hard regression for existing scripts that call `gptme service init` on unsupported platforms without passing `--platform`: they now abort with a non-zero exit and the message "Platform 'Windows' is not supported..." instead of producing the files. Users must now add `--platform linux` to every invocation to restore the old behavior. This breaks automation and documented examples that omit the flag.

<sub>**How this was verified:** Traced the code: when auto-detect sets platform_choice='macos' and output_dir is supplied, the warning block is bypassed, and the macOS branch later generates a launchd plist. Existing macOS users with explicit output-dir get no indication their generated files changed type.</sub>

<sub>Consensus: 2/3 passes agreed — pass 3 looked and disagreed</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/cli/cmd_service.py:455`

The global model validation at lines 455-459 rejects single quotes, double quotes, and backslashes for every platform. On macOS these characters are valid inside an XML plist string and are already escaped by _xml_escape before interpolation, so the rejection is unnecessary and blocks legitimate model identifiers (e.g. one containing an apostrophe). The accompanying comment claims this prevents 'silent truncation', but truncation is not a risk with XML escaping. Consequently a macOS user running `--platform macos --model "gpt-4o's-mini"` gets a BadParameter error even though the plist could be generated safely. This is also inconsistent with test_macos_path_with_systemd_invalid_chars_accepted, which accepts apostrophes in the work-dir on macOS. The check should be limited to the systemd branch, or replaced with proper XML escaping (which is already done).

```suggestion
Check if the path exists and is a directory before calling `mkdir`, or catch `FileExistsError` and treat a non-directory as an error.
```

<sub>**How this was verified:** Confirmed the global check runs before platform branching and that _xml_escape would handle quotes/backslashes safely in the plist. The test test_macos_path_with_systemd_invalid_chars_accepted demonstrates platform-specific acceptance of apostrophes in paths.</sub>

<sub>Consensus: 2/3 passes agreed — pass 1 looked and disagreed</sub>

</details>

- `gptme/cli/cmd_service.py` — Adds launchd plist generation, platform detection, `--platform` option, and moves directory creation after validation.
- `tests/test_cmd_service.py` — Adds tests for launchd plist generation, XML validity, platform detection, and forbidden-character rejection.

</details>

| `106065db6722` | 3/5 | 1 | llm | 2026-08-22 17:45 UTC |
| `2233758445b7` | 4/5 | 3 | llm | 2026-08-22 19:01 UTC |
| `bd576b9db0aa` | 4/5 | 2 | llm | 2026-08-22 20:10 UTC |
| `15fe88f85a94` | 3/5 | 4 | llm | 2026-08-22 21:53 UTC |
| `20c3b501b1e3` | 3/5 | 4 | llm | 2026-08-23 03:36 UTC |

</details>

<sub>Reviewed `23085a505a49` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 177s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -200,6 +257,38 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+        schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
+
+    # on-demand = manual start only (RunAtLoad=false); scheduled = start on load too
+    run_at_load = "false" if timer_schedule == "on-demand" else "true"
+
+    return LAUNCHD_PLIST_TEMPLATE.format(
+        name=_xml_escape(name),
+        model=_xml_escape(model),
+        work_dir=_xml_escape(work_dir),
+        schedule_section=schedule_section,
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 11b2ec724 — added `_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F￾￿]")` and a `_xml_safe()` helper that strips those chars before calling `_xml_escape()`. All three user-controlled values (`name`, `model`, `work_dir`) now go through `_xml_safe()` in `_generate_launchd_plist`. Added a regression test (`test_launchd_plist_strips_control_characters`) that injects `\x01` and `\x0B` into the model string and verifies the resulting plist parses as valid XML with those bytes absent.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -1,33 +1,54 @@
...
+import platform
 import re
 import shlex
 import subprocess
 from pathlib import Path
+from xml.sax.saxutils import escape as _xml_escape
+
+# XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
+# Strip everything else (control characters that survive saxutils.escape and corrupt plists).
+_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F￾￿]")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4e7f3c2d2 — extended `_XML10_FORBIDDEN` to `r"[\x00-\x08\x0B\x0C\x0E-\x1F\uD800-\uDFFF￾￿]"`, covering lone UTF-16 surrogates (U+D800–U+DFFF) alongside the original control-character ranges. The regression test now injects `\uD800` into the model string and verifies it is absent from the generated plist.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -1,33 +1,55 @@
...
+
+# XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
+# Strip control characters and lone UTF-16 surrogates — both survive saxutils.escape unchanged
+# and cause launchctl to reject the generated plist.
+_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F\uD800-\uDFFF￾￿]")
+
+
+def _xml_safe(value: str) -> str:
+    """Strip XML 1.0-forbidden control characters, then XML-escape special chars."""
+    return _xml_escape(_XML10_FORBIDDEN.sub("", value))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d8d414ff5 — instead of silently sanitizing `work_dir` in the plist (which would cause launchd to reference a nonexistent path), the macOS branch now validates the resolved work-dir upfront and raises `click.BadParameter` if it contains XML 1.0-forbidden characters. Added `test_launchd_work_dir_with_forbidden_chars_rejected` to cover this. `model` continues to use `_xml_safe` (silent strip) since it is an env-var string, not a filesystem path.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -293,83 +413,133 @@ def init(
...
         )
 
+    # Resolve platform (auto-detect if needed)
+    if platform_choice == "auto":
+        platform_choice = _detect_platform()
+
+    # Set default output directory based on platform
+    if output_dir is None:
+        if platform_choice == "macos":
+            output_dir = "~/Library/LaunchAgents"
```
TimeToBuildBob, line 442: ⚠️ **P2** — On macOS, the default `--platform auto` now resolves to `macos`, changing the default output directory from `~/.config/systemd/user` to `~/Library/LaunchAgents` and generating a `.plist` instead of `name.service`/`name.timer`. Previously, `gptme service init` always wrote systemd units into `~/.config/systemd/user` regardless of OS. Any existing scripts or users on macOS that rely on the old default will have their files placed in a new location with a different format, breaking automation that expects the systemd unit files. The behavior should either require an explicit `--platform` when the system is macOS and no `--output-dir` is given, or issue a prominent warning.

gptme/cli/cmd_service.py
```diff
@@ -200,6 +268,38 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+    if timer_schedule != "on-demand":
+        interval = LAUNCHD_SCHEDULE_INTERVALS.get(timer_schedule, 86400)
+        schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
+
+    # on-demand = manual start only (RunAtLoad=false); scheduled = start on load too
+    run_at_load = "false" if timer_schedule == "on-demand" else "true"
+
+    return LAUNCHD_PLIST_TEMPLATE.format(
+        name=_xml_safe(name),
+        model=_xml_safe(model),
```
TimeToBuildBob: ⚠️ **P2** — In the systemd branch, `_escape_systemd_value` raises `click.BadParameter` for models containing newlines, quotes, or backslashes, rejecting the input. In the launchd branch, `_generate_launchd_plist` passes the model through `_xml_safe`, which silently strips XML 1.0-forbidden characters (e.g., `\x01`, `\n`, lone surrogates). A model like `"gpt-4o-mini\nEnvironment=GPTME_MALICIOUS=1"` is rejected on Linux but silently truncated to `"gpt-4o-mini"` on macOS. This causes the agent to run with a different model than the user specified, leading to configuration drift and potential security confusion. The two branches should validate models consistently.

gptme/cli/cmd_service.py
```diff
@@ -200,6 +268,38 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
     return True
 
 
+def _detect_platform() -> str:
+    """Detect the current platform: 'macos' or 'linux'."""
+    system = platform.system().lower()
+    if system == "darwin":
+        return "macos"
+    return "linux"
```
TimeToBuildBob: ⚠️ **P2** — `_detect_platform` returns `"linux"` for any `platform.system()` other than `"darwin"`, including Windows, FreeBSD, and other non-Linux systems. When `--platform auto` is used on such a system, the command will attempt to generate systemd units, which are specific to Linux and may be invalid or unusable. The helper should either raise an error for unsupported platforms or map only known Linux variants to `"linux"`, leaving unknown systems to require an explicit choice. This prevents generating service files that cannot work on the host OS.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

tests/test_cmd_service.py
```diff
@@ -453,3 +453,370 @@ def test_model_newline_rejected(tmp_path: Path) -> None:
...
+                str(tmp_path / "out"),
+                "--platform",
+                "auto",
+            ],
+        )
+    assert result.exit_code != 0, "unsupported platform must produce a non-zero exit"
+    assert "not supported" in (result.output + str(result.exception)).lower()
+
+
+def test_macos_autodetect_emits_warning(tmp_path: Path) -> None:
```
TimeToBuildBob, line 506: ⚠️ **P2** — The test `test_macos_autodetect_emits_warning` claims to verify that auto-detecting macOS prints a note about launchd versus systemd, but it passes `--output-dir` explicitly. The note is only printed when `output_dir is None` (line 429 in cmd_service.py), so with an explicit output-dir the note is never emitted. The assertion `assert "macos" in result.output.lower() or "launchd" in result.output.lower()` then passes vacuously because the scaffold message always includes the platform (e.g., `Scaffolding headless agent 'agent' (macos)...`). The test does not actually exercise the warning path, so any regression that removes the note would not be caught.

gptme/cli/cmd_service.py
```diff
@@ -293,83 +423,159 @@ def init(
...
+            )
+
+    # Set default output directory based on platform
+    if output_dir is None:
+        if platform_choice == "macos":
+            output_dir = "~/Library/LaunchAgents"
+        else:
+            output_dir = "~/.config/systemd/user"
+
     work = _resolve_work_dir(work_dir)
```
TimeToBuildBob: ⚠️ **P2** — In `init`, the work directory and output directory are created (via `_resolve_work_dir` and `out_dir.mkdir`) before the model and work-dir validation. If the model contains a newline, a quote, or a backslash, or if the work-dir contains XML-forbidden characters on macOS, the command raises a `BadParameter` after already having created the directories. This leaves behind side-effect directories (and potentially a logs directory) when the user's input is invalid. For example, running `gptme service init --name x --model 'gpt-4o\nEnviron'` creates `~/gptme-agent` and the output dir before rejecting the model. The user is left with unexpected files that may not be cleaned up, and the error message does not indicate that directories were created.

TimeToBuildBob:
All checks green (CI ✅, Greptile ✅ 5/5). All Greptile P1 findings addressed:
- XML escape for user-controlled plist values
- Control character stripping
- Surrogate character handling
- Work-dir validation (raises `click.BadParameter` for forbidden chars rather than silent sanitize)

Not self-merge eligible (CLI code in `gptme/cli/cmd_service.py`). Ready for your review and merge when you get a chance.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -396,11 +606,24 @@ def init(
...
     click.echo("Next steps:")
-    click.echo("  systemctl --user daemon-reload")
-    if timer_schedule != "on-demand":
-        click.echo(f"  systemctl --user enable --now {name}.timer")
-    click.echo(f"  systemctl --user start {name}.service   # first run")
-    click.echo(f"  journalctl --user -u {name}.service -f  # follow logs")
+
+    if platform_choice == "macos":
+        plist_file = out_dir / f"com.gptme.{name}.plist"
+        click.echo(f"  launchctl load {plist_file}")
```
TimeToBuildBob: ⚠️ **P2** — In the macOS 'Next steps' output, `click.echo(f"  launchctl load {plist_file}")` prints the plist path without quoting. If the output directory or the resolved path contains spaces (e.g., a directory named 'My Agent' or a workspace with a space), the printed `launchctl load` command will be syntactically invalid and fail when copied and executed. The instruction should quote the path, e.g., `click.echo(f"  launchctl load \"{plist_file}\"")`, so the command works for paths with spaces. This is a usability defect in the generated user instructions. _(1/2 passes — pass 2 looked and disagreed)_

gptme/cli/cmd_service.py
```diff
@@ -200,6 +266,48 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+    if timer_schedule != "on-demand":
+        interval = LAUNCHD_SCHEDULE_INTERVALS.get(timer_schedule, 86400)
+        schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
+
+    # on-demand = manual start only (RunAtLoad=false); scheduled = start on load too
+    run_at_load = "false" if timer_schedule == "on-demand" else "true"
+
+    return LAUNCHD_PLIST_TEMPLATE.format(
+        name=_xml_safe(name),
+        model=_xml_safe(model),
```
TimeToBuildBob: ⚠️ **P2** — `_generate_launchd_plist` passes the model through `_xml_safe`, which silently strips every XML 1.0-forbidden control character and lone surrogate. A model such as `gpt-4o\x01mini` becomes `gpt-4omini` with no warning, so the agent runs with a different model than the user requested. The validation only rejects newlines and quotes/backslashes, so other control chars pass through and are altered. This silent data mutation can cause the agent to use a wrong or nonexistent model, which is a correctness problem (even though it is intentional per the tests). _(1/2 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
Addressed AI reviewer P1 + two P2 findings (commit `20c3b501b`):

**P1 fixed** — `test_macos_autodetect_emits_warning` now passes `env={"HOME": str(tmp_path)}` to CliRunner so the `~/Library/LaunchAgents` default expands into tmp, not the real filesystem. Test was environment-dependent before (could leave files in `~` on developer machines or fail in sandboxed CI).

**P2 fixed** — `launchctl load` instruction now quotes the plist path: `` launchctl load "{plist_file}" `` so paths with spaces copy correctly.

**P2 fixed** — `_escape_systemd_value(str(work))` now called *before* `work.mkdir` on the Linux branch, consistent with the macOS branch and the comment "All validation passed — now create directories". Previously a quote/backslash in the work-dir would fire `BadParameter` after an empty directory was already created.

Skipped P2 about silent `_xml_safe` model stripping — that's tested intentionally in `test_launchd_plist_strips_control_characters` and only affects control chars that would make the plist invalid XML anyway (consensus was 1/2, so not a hard finding).

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_service.py
```diff
@@ -293,83 +421,170 @@ def init(
...
-        out_dir / f"{name}.service",
-        SYSTEMD_SERVICE_TEMPLATE.format(
-            name=name,
-            model=escaped_model,
-            work_dir=escaped_work_dir,
-            exec_work_dir=escaped_exec_work_dir,
-        ),
-        force=force,
-    )
+    if platform_choice == "macos":
```
TimeToBuildBob, line 492: ❌ **P1** — When switching a macOS agent from the old systemd default to the new launchd default via `--platform auto`, the `init` command writes the new `com.gptme.{name}.plist` but does not remove previously generated systemd unit files (`{name}.service`, `{name}.timer`) from `~/.config/systemd/user`, even with `--force`. Since those units may still be enabled, both systemd and launchd can start the agent, causing duplicate sessions and resource contention. The macOS branch (line 492) has no cleanup logic for stale systemd units, unlike the timer-removal logic in the systemd branch. The user is only warned about the output directory change, not about leftover units. _(1/2 passes — pass 1 looked and disagreed)_

gptme/cli/cmd_service.py
```diff
@@ -1,33 +1,55 @@
...
+
+# XML 1.0 permits only #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF].
+# Strip control characters and lone UTF-16 surrogates — both survive saxutils.escape unchanged
+# and cause launchctl to reject the generated plist.
+_XML10_FORBIDDEN = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F\uD800-\uDFFF\uFFFE\uFFFF]")
+
+
+def _xml_safe(value: str) -> str:
+    """Strip XML 1.0-forbidden control characters, then XML-escape special chars."""
+    return _xml_escape(_XML10_FORBIDDEN.sub("", value))
```
TimeToBuildBob, line 47: ⚠️ **P2** — The `_xml_safe` function silently strips XML 1.0-forbidden characters from the `model` string, but the preceding validation only rejects newlines and quotes. Control characters such as `\x01` or a lone surrogate pass validation and are then removed from the model before it is written into the launchd plist's `GPTME_AGENT_MODEL` environment variable. For example, `gpt-4o\x01mini` becomes `gpt-4omini`, so the agent is configured with a model the user did not specify. This is inconsistent with the work-dir handling, which rejects such characters rather than stripping them, and it can lead to an unintended model being used. The fix is to also reject control characters in the model, matching the work-dir validation. _(2/2 passes)_

gptme/cli/cmd_service.py
```diff
@@ -200,6 +266,48 @@ def _write_file(path: Path, content: str, force: bool = False) -> bool:
...
+    timer_schedule: str,
+) -> str:
+    """Generate a launchd .plist file for macOS agents."""
+    schedule_section = ""
+    if timer_schedule != "on-demand":
+        interval = LAUNCHD_SCHEDULE_INTERVALS.get(timer_schedule, 86400)
+        schedule_section = LAUNCHD_SCHEDULE_SECTION.format(interval_seconds=interval)
+
+    # on-demand = manual start only (RunAtLoad=false); scheduled = start on load too
+    run_at_load = "false" if timer_schedule == "on-demand" else "true"
```
TimeToBuildBob, line 300: ⚠️ **P2** — The launchd plist sets `RunAtLoad` to `<true/>` for any schedule other than `on-demand`, causing the agent to start immediately when the plist is loaded. Combined with `StartInterval` (e.g. 86400 for `daily`), the agent runs once at load time and then every 24 hours from that moment. This differs from the systemd timer, which uses `OnCalendar` and runs at a specific wall-clock time such as midnight. A user who expects a daily run at midnight will instead get a run at the moment they `launchctl load` the file, and then every 86400 seconds thereafter, potentially causing an unexpected immediate run before the user intends to start the agent. _(1/2 passes — pass 2 looked and disagreed)_

gptme/cli/cmd_service.py
```diff
@@ -293,83 +421,170 @@ def init(
...
+    # Validate model consistently across platforms BEFORE creating any directories.
+    # systemd rejects newlines/quotes/backslashes; launchd must too (else the same
+    # input is silently truncated on macOS but rejected on Linux, which is confusing
+    # and allows injection attempts to go unnoticed on macOS).
+    if "\n" in model or "\r" in model:
+        raise click.BadParameter(
+            f"{model!r} must not contain newlines",
+            param_hint="'--model'",
+        )
+    if any(ch in model for ch in ("\\", '"', "'")):
```
TimeToBuildBob, line 455: ⚠️ **P2** — The model validation rejects single quotes, double quotes, and backslashes for all platforms, even though on macOS these characters are valid inside an XML plist string and may legitimately appear in a model identifier. The same restriction is applied to the systemd path where it is necessary to prevent directive injection, but on macOS it is overly restrictive and will reject models like `gpt-4o'mini` or `gpt-4o\mini` that would otherwise be safe. This is a minor compatibility issue for users with such model names and creates an inconsistency: the work-dir on macOS allows apostrophes and backslashes, while the model does not. _(1/2 passes — pass 2 looked and disagreed)_

gptme/cli/cmd_service.py
```diff
@@ -293,83 +421,179 @@ def init(
             param_hint="'--name'",
         )
 
+    # Resolve platform (auto-detect if needed)
+    if platform_choice == "auto":
+        platform_choice = _detect_platform()
```
TimeToBuildBob, line 426: ❌ **P1** — When `--platform auto` is the default (it is the default value of the option) and the host OS is not Linux or macOS (e.g. Windows, FreeBSD), `_detect_platform()` raises `click.UsageError`. Previously the `init` command always generated systemd units regardless of the OS. This is a hard regression for existing scripts that call `gptme service init` on unsupported platforms without passing `--platform`: they now abort with a non-zero exit and the message "Platform 'Windows' is not supported..." instead of producing the files. Users must now add `--platform linux` to every invocation to restore the old behavior. This breaks automation and documented examples that omit the flag. _(2/3 passes — pass 3 looked and disagreed)_

gptme/cli/cmd_service.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `23085a505` — switched from silently stripping forbidden chars to **rejecting the input entirely**. The macOS branch now runs `_XML10_FORBIDDEN.search(model)` before any file I/O and raises `click.BadParameter` if the check hits (lines 467–476 in the current tree). `_generate_launchd_plist` still receives a clean, pre-validated `model` string and calls `_xml_escape` (which handles only `&`/`<`/`>`) — there is no longer any stripping that would silently alter the identifier.

TimeToBuildBob:
## Convergence Adjudication — gptme/gptme#3586

Hit the Greptile attempt cap after 5 rounds. Manually adjudicating the 4 remaining unresolved P1 findings against HEAD `23085a505`.

### Finding Classification

All 4 P1 findings are **dismissed as false positives** against the current HEAD:

| Finding | Verdict | Reason |
|---|---|---|
| **Control characters still break plists** | Dismissed — fixed | Current code rejects any model/work-dir containing XML 1.0-forbidden control chars (`_XML10_FORBIDDEN.search(model)` at lines 467–476) before plist generation. Control chars never reach the template. |
| **Surrogates bypass XML sanitization** | Dismissed — fixed | `_XML10_FORBIDDEN` (line 42) includes `\uD800-\uDFFF` (lone surrogates). Additionally, any input containing surrogates is now **rejected** (not stripped) before reaching the template. |
| **Sanitized path misses generated script** | Dismissed — fixed | Work-dir paths containing forbidden chars are rejected with `click.BadParameter` before any directory creation or file write (lines 477–487). No path sanitization mismatch is possible. |
| **Sanitization changes model identifiers** | Dismissed — fixed | `23085a505` switched from `_xml_safe(model)` (silent strip) to `_XML10_FORBIDDEN.search(model)` + rejection. `_generate_launchd_plist` now receives a pre-validated model string and calls only `_xml_escape` for `&`/`<`/`>`. |

### CI

All checks green: lint ✅ typecheck ✅ tests (no API, API/openai, API/claude-haiku) ✅ Docker ✅ PyInstaller (ubuntu + macos) ✅

### Domain Risk

The macOS launchd path is new infrastructure — plist generation, `RunAtLoad` logic, `LAUNCHD_SCHEDULE_INTERVALS`, and the `_xml_safe`/`_XML10_FORBIDDEN` machinery. Unit tests cover the generated plist structure and error paths, but a maintainer smoke-test on actual macOS (run `gptme service init --platform macos`, then `launchctl load` the result) would confirm the plist validates correctly with `plutil -lint`.

### Convergence

`round_convergence.stable_rounds: 0` — cap reached due to iterative fix rounds, not stall. Each review found a new edge case in the XML-safety approach until the strategy shifted from stripping to rejection in `23085a505`. No new review needed; the 4 remaining findings are addressed by the same root fix.

### Merge Recommendation

Ready for human review and merge. No blocking issues remain in the code. Waiting for maintainer review.

gptme/cli/cmd_service.py
TimeToBuildBob: Adjudicated: **false positive**. The `_detect_platform()` raising `UsageError` on Windows/FreeBSD is intentional — the docstring explicitly says "Use --platform linux or --platform macos". The old behavior of silently generating Linux systemd files on Windows was wrong. Marking as not-blocking.

gptme/cli/cmd_service.py
TimeToBuildBob: Adjudicated: **accepted as low-priority**. 1/2 passes (low confidence). The model validation restriction is platform-agnostic for safety; XML escaping is in place but shell-embedding the model identifier also needs sanitization. Non-blocking.

TimeToBuildBob:
**Triage update** (session 6933): 
- All 6 review threads resolved (4 Greptile + 2 Bob AI reviewer adjudications)
- P1 `_detect_platform` raising UsageError on unsupported platforms: **false positive** — intentional behavior, documented in docstring; old behavior of silently generating wrong files was worse
- P2 model validation overly strict on macOS: **accepted low-priority** (1/2 passes), non-blocking
- Greptile lifetime cap reached (8/8 triggers) — no further auto-reviews. Last score was 3/5
- PM retry file cleared; **needs human merge decision or Greptile cap reset**

PR #3587, opened 2026-08-22
feat(skills): invoke skills as slash commands (/skill:<name>)

TimeToBuildBob:
## Summary

Claude Code and Codex expose every `SKILL.md` skill as a `/<name>` slash command that injects the skill body as the user's prompt and lets the assistant act on it. gptme only had `/skills read <name>`, which prints the skill as a system message and never triggers an assistant turn (`chat.py`: `if msg.role == "user" and execute_cmd(msg, manager): continue`).

This adds parity so cross-runtime shared skills (e.g. an `end` skill invoked as `/skill:end` or `/end`) work the same way in gptme.

## Changes

- **`gptme/lessons/skill_commands.py`** (new): `register_skill_commands()` registers `skill:<name>` for every indexed skill with `metadata.name`. A bare `/<name>` alias is added only when it collides with neither an already-registered command nor a loaded tool (`handle_cmd` falls back to tool execution for unknown names, so a skill named `shell` must never shadow the shell tool).
  - Handler materializes stub skills, substitutes `$ARGUMENTS`, `$ARGUMENTS[N]` and `${N}` (0-based), prepends `Skill invoked: /skill:<name> <args>`, and queues the result via `prompt_queue.queue_prompt()`. The command message itself is undone (like built-in commands) so the body isn't logged twice.
  - Handler `__doc__` is the skill description, so `/help` / `get_commands_with_descriptions()` show it (alias deduped by handler identity).
  - Idempotent and never raises: stale commands from a previous registration are removed first; a foreign command that later took over a bare name is left alone; any index failure is logged as a warning.
- **`gptme/init.py`**: call `register_skill_commands()` right after `init_commands()` so the tool-collision check sees loaded tools and built-in commands.
- **`gptme/lessons/__init__.py`**: export `register_skill_commands`.
- **`gptme/server/api_v2.py`**: drain the durable prompt queue after executing a slash command and append drained prompts as user messages so `/step` (WebUI auto-steps after send) generates a response.
- **`gptme/tui/app.py`**: drain the durable prompt queue after a handled command, show the skill prompt, and start a generation turn.
- **`docs/skills.rst`**: "Invoking Skills as Commands" section (CLI, TUI, and server/WebUI).
- **`tests/test_skill_commands.py`**: substitution, registration + `/help` + completion source, handler → prompt queue, collisions, idempotent re-registration, foreign-command preservation, broken index never raises, prefix-named skills do not clobber canonicals.
- **`tests/test_server.py` / `tests/test_tui.py`**: smoke tests that `/skill:<name>` drains the queue and becomes a user prompt (TUI also starts generation).

Tab completion needs no changes: `GptmeCompleter` matches `option.startswith(text)` over `get_user_commands()`, so `/skill:` completes to the registered names.

## Test plan

- [x] `poetry run pytest tests/test_skill_commands.py tests/test_tui.py::test_skill_command_drains_prompt_queue_and_starts_generation tests/test_tui.py::test_slash_command_help tests/test_server.py::test_api_v2_skill_command_drains_prompt_queue tests/test_server.py::test_api_v2_conversation_command -q` → 22 passed
- [x] `poetry run mypy gptme/lessons/skill_commands.py gptme/server/api_v2.py gptme/tui/app.py tests/test_tui.py tests/test_server.py` → clean
- [x] `ruff check` / `ruff format` on touched files
- [x] Manual smoke with real tools loaded (`init_tools(['shell'])` + temp skills `shell`, `end`): registered `['skill:end', 'end', 'skill:shell']` — `shell` alias correctly suppressed; `/help` shows skill descriptions; `get_user_commands()` includes `/skill:end`, `/end`.

### DIFF
--- a/gptme/init.py
+++ b/gptme/init.py
@@ -83,9 +83,6 @@ def init(
             first_line,
         )
     init_tools(tool_allowlist)
-    # Expose skills as /skill:<name> commands (after tools so bare aliases can
-    # avoid shadowing loaded tool names). Never raises.
-    register_skill_commands()
     init_hooks(interactive=interactive, no_confirm=no_confirm, server=server)
 
     config = get_config()
@@ -98,6 +95,9 @@ def init(
         register_script_hooks(script_hooks, workspace or Path.cwd())
 
     init_commands()
+    # Expose skills as /skill:<name> commands (after both tools and built-in
+    # commands so bare aliases see the full registry). Never raises.
+    register_skill_commands()
 
     set_tool_format(tool_format)
     # Mark initialization done at the end so callers can retry init()
--- a/gptme/init.py
+++ b/gptme/init.py
@@ -16,6 +16,7 @@ from .cli.setup import ask_for_api_key
 from .commands import init_commands
 from .config import Config, get_config
 from .hooks import init_hooks
+from .lessons.skill_commands import register_skill_commands
 from .llm import guess_provider_from_config, init_llm, is_custom_provider
 from .llm.llm_gptme import GptmeAuthError
 from .llm.models import (
@@ -82,6 +83,9 @@ def init(
             first_line,
         )
     init_tools(tool_allowlist)
+    # Expose skills as /skill:<name> commands (after tools so bare aliases can
+    # avoid shadowing loaded tool names). Never raises.
+    register_skill_commands()
     init_hooks(interactive=interactive, no_confirm=no_confirm, server=server)
 
     config = get_config()
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -86,11 +86,8 @@ def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
     def handler(ctx: CommandContext) -> Generator[Message, None, None]:
         from ..prompt_queue import queue_prompt  # fmt: skip
 
-        # Remove the "/skill:<name>" command message from the log, like built-in
-        # commands do; the queued prompt carries its own "Skill invoked" header.
-        ctx.manager.undo(1, quiet=True)
-        ctx.manager.write()
-
+        # Build the prompt first; only undo the command message on success so a
+        # broken skill leaves the log intact instead of silently disappearing.
         try:
             content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
         except Exception as e:
@@ -98,6 +95,10 @@ def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
             yield from ()
             return
         queue_prompt(ctx.manager.logdir, content)
+        # Remove the "/skill:<name>" command message from the log after queuing,
+        # like built-in commands do; the queued prompt carries its own header.
+        ctx.manager.undo(1, quiet=True)
+        ctx.manager.write()
         logger.debug("Queued skill prompt for /%s%s", SKILL_COMMAND_PREFIX, name)
         # Yield nothing: the chat loop drains the queue on its next iteration
         # and the assistant responds to the skill body as a normal user turn.
@@ -164,6 +165,9 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
         # alias can accidentally shadow a real tool command.
         tools_available = tool_names is not _TOOL_NAMES_UNAVAILABLE
 
+        # Two-pass registration so canonicals never collide with our own aliases.
+        # Pass 1: collect valid skills and register all canonicals.
+        skill_entries: list[tuple[str, str, CommandHandler]] = []
         for skill in index.lessons:
             name = skill.metadata.name
             if not name:
@@ -189,8 +193,13 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
             register_command(canonical, handler)
             _registered_skill_commands[canonical] = handler
             registered.append(canonical)
+            skill_entries.append((name, canonical, handler))
 
-            if name in _command_registry:
+        # Pass 2: register bare aliases, checking against all canonicals we just
+        # registered so a skill named "skill:X" doesn't block bare alias for "X".
+        our_canonicals = {canonical for _, canonical, _ in skill_entries}
+        for name, canonical, handler in skill_entries:
+            if name in _command_registry and name not in our_canonicals:
                 logger.debug(
                     "Skill %r collides with existing command; only /%s registered",
                     name,
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -233,3 +233,8 @@ def unregister_skill_commands() -> None:
         if _command_registry.get(name) is handler:
             unregister_command(name)
     _registered_skill_commands.clear()
+
+
+def is_skill_command(name: str) -> bool:
+    """Return True if *name* (without leading ``/``) is a registered skill command."""
+    return name in _registered_skill_commands
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -40,7 +40,7 @@ SKILL_COMMAND_PREFIX = "skill:"
 # aliases not count as collisions.
 _registered_skill_commands: dict[str, CommandHandler] = {}
 
-_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$\{(\d+)\}")
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS(?!\[)\b|\$\{(\d+)\}")
 
 
 def substitute_arguments(body: str, full_args: str, args: list[str]) -> str:
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -206,7 +206,7 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
                     name,
                     canonical,
                 )
-            elif not tools_available or name in tool_names:  # type: ignore[operator]
+            elif not tools_available or name.lower() in {t.lower() for t in tool_names}:  # type: ignore[attr-defined]
                 logger.debug(
                     "Skill %r collides with loaded tool (or tool list unavailable); only /%s registered",
                     name,
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -1,0 +1,193 @@
+"""Expose skills as user-invocable slash commands.
+
+Claude Code and Codex expose every ``SKILL.md`` skill as a ``/<name>`` command
+that injects the skill body as the user's prompt. This module gives gptme the
+same behavior so skills can be shared across runtimes:
+
+- ``/skill:<name> [args]`` is the canonical, collision-free form.
+- ``/<name>`` is registered as an alias only when ``<name>`` does not collide
+  with an existing command or a loaded tool (``handle_cmd`` falls back to tool
+  execution for unknown names, so a skill named ``shell`` must never shadow
+  the shell tool).
+
+The handler substitutes ``$ARGUMENTS`` (and ``$ARGUMENTS[N]`` / ``$N``
+positional forms) in the skill body and queues the result via
+:func:`gptme.prompt_queue.queue_prompt`, so the main chat loop drains it as a
+normal user message on its next iteration and the assistant acts on it.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from typing import TYPE_CHECKING
+
+from .index import LessonIndex
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from ..commands.base import CommandContext, CommandHandler
+    from ..message import Message
+    from .parser import Lesson
+
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS|\$(\d+)")
+
+
+def substitute_arguments(body: str, full_args: str, args: list[str]) -> str:
+    """Substitute ``$ARGUMENTS``, ``$ARGUMENTS[N]`` and ``$N`` placeholders.
+
+    ``$ARGUMENTS`` expands to the full argument string; ``$ARGUMENTS[N]`` and
+    ``$N`` expand to the N-th (0-based) whitespace-split argument, or the empty
+    string when out of range.
+    """
+
+    def _replace(match: re.Match[str]) -> str:
+        idx_str = match.group(1) or match.group(2)
+        if idx_str is None:
+            return full_args
+        idx = int(idx_str)
+        return args[idx] if idx < len(args) else ""
+
+    return _ARG_PATTERN.sub(_replace, body)
+
+
+def build_skill_prompt(
+    skill: Lesson, index: LessonIndex, full_args: str, args: list[str]
+) -> str:
+    """Materialize a skill and render it as a user prompt."""
+    if skill.is_stub:
+        skill = index.materialize_lesson(skill)
+    name = skill.metadata.name or skill.path.stem
+    header = f"Skill invoked: /{SKILL_COMMAND_PREFIX}{name}"
+    if full_args:
+        header += f" {full_args}"
+    body = substitute_arguments(skill.body, full_args, args)
+    return f"{header}\n\n{body}".rstrip()
+
+
+def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
+    name = skill.metadata.name or skill.path.stem
+
+    def handler(ctx: CommandContext) -> Generator[Message, None, None]:
+        from ..prompt_queue import queue_prompt  # fmt: skip
+
+        # Remove the "/skill:<name>" command message from the log, like built-in
+        # commands do; the queued prompt carries its own "Skill invoked" header.
+        ctx.manager.undo(1, quiet=True)
+        ctx.manager.write()
+
+        content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
+        queue_prompt(ctx.manager.logdir, content)
+        logger.debug("Queued skill prompt for /%s%s", SKILL_COMMAND_PREFIX, name)
+        # Yield nothing: the chat loop drains the queue on its next iteration
+        # and the assistant responds to the skill body as a normal user turn.
+        yield from ()
+
+    desc = skill.metadata.description or skill.description or f"Invoke skill {name}"
+    handler.__doc__ = desc.strip().split("\n")[0]
+    handler.__name__ = f"skill_{name}"
+    return handler
+
+
+def _loaded_tool_names() -> set[str]:
+    try:
+        from ..tools import get_tools  # fmt: skip
+
+        return {tool.name for tool in get_tools()}
+    except Exception as e:  # pragma: no cover - defensive
+        logger.debug("Could not list loaded tools for skill alias check: %s", e)
+        return set()
+
+
+def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
+    """Register every skill in the lesson index as a ``/skill:<name>`` command.
+
+    Also registers a bare ``/<name>`` alias when it collides with neither an
+    existing command nor a loaded tool. Safe to call repeatedly (stale skill
+    commands from a previous call are removed first) and never raises.
+
+    Returns:
+        The list of command names registered.
+    """
+    from ..commands.base import (  # fmt: skip
+        _command_registry,
+        register_command,
+        unregister_command,
+    )
+
+    registered: list[str] = []
+    try:
+        # Drop commands from a previous registration so renamed/removed skills
+        # do not linger and re-registration stays idempotent.
+        for old_name, old_handler in list(_registered_skill_commands.items()):
+            if _command_registry.get(old_name) is old_handler:
+                unregister_command(old_name)
+        _registered_skill_commands.clear()
+
+        if index is None:
+            index = LessonIndex()
+        tool_names = _loaded_tool_names()
+
+        for skill in index.lessons:
+            name = skill.metadata.name
+            if not name:
+                continue
+            name = name.strip()
+            if not name or any(ch.isspace() for ch in name):
+                logger.warning("Skipping skill with invalid name: %r", name)
+                continue
+
+            try:
+                handler = _make_skill_handler(skill, index)
+            except Exception as e:
+                logger.warning("Failed to build command for skill %r: %s", name, e)
+                continue
+
+            canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            register_command(canonical, handler)
+            _registered_skill_commands[canonical] = handler
+            registered.append(canonical)
+
+            if name in _command_registry:
+                logger.debug(
+                    "Skill %r collides with existing command; only /%s registered",
+                    name,
+                    canonical,
+                )
+            elif name in tool_names:
+                logger.debug(
+                    "Skill %r collides with loaded tool; only /%s registered",
+                    name,
+                    canonical,
+                )
+            else:
+                register_command(name, handler)
+                _registered_skill_commands[name] = handler
+                registered.append(name)
+
+        if registered:
+            logger.debug("Registered %d skill command(s)", len(registered))
+    except Exception as e:
+        logger.warning("Failed to register skill commands: %s", e)
+
+    return registered
+
+
+def unregister_skill_commands() -> None:
+    """Remove all commands registered by :func:`register_skill_commands`."""
+    from ..commands.base import _command_registry, unregister_command  # fmt: skip
+
+    for name, handler in list(_registered_skill_commands.items()):
+        if _command_registry.get(name) is handler:
+            unregister_command(name)
+    _registered_skill_commands.clear()
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -12,8 +12,8 @@ same behavior so skills can be shared across runtimes:
 
 The handler substitutes ``$ARGUMENTS`` (and ``$ARGUMENTS[N]`` / ``${N}``
 positional forms) in the skill body and queues the result via
-:func:`gptme.prompt_queue.queue_prompt`, so the main chat loop drains it as a
-normal user message on its next iteration and the assistant acts on it.
+:func:`gptme.prompt_queue.queue_prompt`. The CLI chat loop, TUI, and server
+API drain that queue as a normal user message so the assistant acts on it.
 """
 
 from __future__ import annotations
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -40,7 +40,7 @@ SKILL_COMMAND_PREFIX = "skill:"
 # aliases not count as collisions.
 _registered_skill_commands: dict[str, CommandHandler] = {}
 
-_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS|\$(\d+)")
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$(\d+)")
 
 
 def substitute_arguments(body: str, full_args: str, args: list[str]) -> str:
@@ -156,6 +156,12 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
                 continue
 
             canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            if canonical in _command_registry:
+                logger.debug(
+                    "Skill canonical %r already registered by another component; skipping",
+                    canonical,
+                )
+                continue
             register_command(canonical, handler)
             _registered_skill_commands[canonical] = handler
             registered.append(canonical)
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -69,7 +69,7 @@ def build_skill_prompt(
     """Materialize a skill and render it as a user prompt."""
     if skill.is_stub:
         skill = index.materialize_lesson(skill)
-    name = skill.metadata.name or skill.path.stem
+    name = (skill.metadata.name or skill.path.stem).strip()
     header = f"Skill invoked: /{SKILL_COMMAND_PREFIX}{name}"
     if full_args:
         header += f" {full_args}"
@@ -78,7 +78,7 @@ def build_skill_prompt(
 
 
 def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
-    name = skill.metadata.name or skill.path.stem
+    name = (skill.metadata.name or skill.path.stem).strip()
 
     def handler(ctx: CommandContext) -> Generator[Message, None, None]:
         from ..prompt_queue import queue_prompt  # fmt: skip
@@ -88,7 +88,12 @@ def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
         ctx.manager.undo(1, quiet=True)
         ctx.manager.write()
 
-        content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
+        try:
+            content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
+        except Exception as e:
+            logger.warning("Failed to build prompt for skill %r: %s", name, e)
+            yield from ()
+            return
         queue_prompt(ctx.manager.logdir, content)
         logger.debug("Queued skill prompt for /%s%s", SKILL_COMMAND_PREFIX, name)
         # Yield nothing: the chat loop drains the queue on its next iteration
@@ -101,14 +106,27 @@ def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
     return handler
 
 
-def _loaded_tool_names() -> set[str]:
+_TOOL_NAMES_UNAVAILABLE = object()  # sentinel: get_tools() failed
+
+
+def _loaded_tool_names() -> set[str] | object:
+    """Return the set of loaded tool names, or ``_TOOL_NAMES_UNAVAILABLE`` on error.
+
+    Callers must treat the sentinel as "all names are occupied" so a failing
+    get_tools() never accidentally grants a bare ``/<name>`` alias to a skill
+    that shares its name with a real tool.
+    """
     try:
         from ..tools import get_tools  # fmt: skip
 
         return {tool.name for tool in get_tools()}
     except Exception as e:  # pragma: no cover - defensive
-        logger.debug("Could not list loaded tools for skill alias check: %s", e)
-        return set()
+        logger.warning(
+            "Could not list loaded tools for skill alias check — "
+            "bare skill aliases suppressed: %s",
+            e,
+        )
+        return _TOOL_NAMES_UNAVAILABLE
 
 
 def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
@@ -139,6 +157,9 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
         if index is None:
             index = LessonIndex()
         tool_names = _loaded_tool_names()
+        # When tool enumeration failed, treat every name as occupied so no bare
+        # alias can accidentally shadow a real tool command.
+        tools_available = tool_names is not _TOOL_NAMES_UNAVAILABLE
 
         for skill in index.lessons:
             name = skill.metadata.name
@@ -172,9 +193,9 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
                     name,
                     canonical,
                 )
-            elif name in tool_names:
+            elif not tools_available or name in tool_names:  # type: ignore[operator]
                 logger.debug(
-                    "Skill %r collides with loaded tool; only /%s registered",
+                    "Skill %r collides with loaded tool (or tool list unavailable); only /%s registered",
                     name,
                     canonical,
                 )
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -56,7 +56,9 @@ def substitute_arguments(body: str, full_args: str, args: list[str]) -> str:
         if idx_str is None:
             return full_args
         idx = int(idx_str)
-        return args[idx] if idx < len(args) else ""
+        # Return the original text unchanged when the index is out of range,
+        # so "$100" in skill prose (prices, currency) is never silently deleted.
+        return args[idx] if idx < len(args) else match.group(0)
 
     return _ARG_PATTERN.sub(_replace, body)
 
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -195,11 +195,12 @@ def register_skill_commands(index: LessonIndex | None = None) -> list[str]:
             registered.append(canonical)
             skill_entries.append((name, canonical, handler))
 
-        # Pass 2: register bare aliases, checking against all canonicals we just
-        # registered so a skill named "skill:X" doesn't block bare alias for "X".
-        our_canonicals = {canonical for _, canonical, _ in skill_entries}
+        # Pass 2: register bare aliases. Any name already in the registry —
+        # including a canonical from pass 1 — is a collision and must not be
+        # overwritten. Without that, a skill named "skill:X" would register
+        # its bare alias as "skill:X" and clobber skill "X"'s canonical.
         for name, canonical, handler in skill_entries:
-            if name in _command_registry and name not in our_canonicals:
+            if name in _command_registry:
                 logger.debug(
                     "Skill %r collides with existing command; only /%s registered",
                     name,
--- a/gptme/lessons/skill_commands.py
+++ b/gptme/lessons/skill_commands.py
@@ -10,7 +10,7 @@ same behavior so skills can be shared across runtimes:
   execution for unknown names, so a skill named ``shell`` must never shadow
   the shell tool).
 
-The handler substitutes ``$ARGUMENTS`` (and ``$ARGUMENTS[N]`` / ``$N``
+The handler substitutes ``$ARGUMENTS`` (and ``$ARGUMENTS[N]`` / ``${N}``
 positional forms) in the skill body and queues the result via
 :func:`gptme.prompt_queue.queue_prompt`, so the main chat loop drains it as a
 normal user message on its next iteration and the assistant acts on it.
@@ -40,15 +40,18 @@ SKILL_COMMAND_PREFIX = "skill:"
 # aliases not count as collisions.
 _registered_skill_commands: dict[str, CommandHandler] = {}
 
-_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$(\d+)")
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$\{(\d+)\}")
 
 
 def substitute_arguments(body: str, full_args: str, args: list[str]) -> str:
-    """Substitute ``$ARGUMENTS``, ``$ARGUMENTS[N]`` and ``$N`` placeholders.
+    """Substitute ``$ARGUMENTS``, ``$ARGUMENTS[N]`` and ``${N}`` placeholders.
 
     ``$ARGUMENTS`` expands to the full argument string; ``$ARGUMENTS[N]`` and
-    ``$N`` expand to the N-th (0-based) whitespace-split argument, or the empty
-    string when out of range.
+    ``${N}`` expand to the N-th (0-based) whitespace-split argument.
+    Out-of-range references (e.g. ``${5}`` with fewer than 6 args) are left
+    unchanged so skill prose with ``${...}`` expressions is not corrupted.
+    The ``${N}`` curly-brace form is required for positional arguments to avoid
+    ambiguity with literal dollar amounts like ``$100`` in skill bodies.
     """
 
     def _replace(match: re.Match[str]) -> str:
--- a/gptme/server/api_v2.py
+++ b/gptme/server/api_v2.py
@@ -1913,15 +1913,11 @@ def api_conversation_post(conversation_id: str):
 
     # Append and execute a command under its reservation, not the conversation lock.
     # The reservation is always cleared, including when handle_cmd raises.
+    from ..lessons.skill_commands import is_skill_command  # fmt: skip
+
     responses: list[Message] = []
     try:
         log.append(msg)
-        # Track log length before running the command. Skill command handlers
-        # call ctx.manager.undo(1) to remove the command message from the log
-        # (the skill prompt is the real user turn, not the /skill:<name> command).
-        # We defer emitting message_added for the command message and only emit
-        # it after handle_cmd finishes — skipping it when it was undone avoids
-        # sending a phantom message_added event that clients can't reconcile.
         _len_after_cmd_append = len(log.log.messages)
 
         def _emit(m: Message) -> None:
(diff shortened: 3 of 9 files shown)


TimeToBuildBob:
CI all green, Greptile 5/5 ("The PR appears safe to merge"), MERGEABLE/CLEAN.

Not eligible for autonomous self-merge: `gptme/init.py` and the new `gptme/lessons/` files are core source — outside the allowed self-merge categories. Waiting for human merge.

gptme/init.py
```diff
@@ -82,6 +83,9 @@ def init(
             first_line,
         )
     init_tools(tool_allowlist)
+    # Expose skills as /skill:<name> commands (after tools so bare aliases can
+    # avoid shadowing loaded tool names). Never raises.
+    register_skill_commands()
```
TimeToBuildBob: ❌ **P1** — In `gptme/init.py`, `register_skill_commands()` is called before `init_commands()`. This means the collision check for bare aliases runs against a registry that does not yet contain the built-in commands (e.g., `/help`, `/exit`). A skill named `help` will be considered non-colliding, so its bare alias `help` is registered. When `init_commands()` later registers the built-in `help` command, it overwrites the skill's alias (or raises if `register_command` does not allow overwriting). The skill's bare alias becomes unusable, forcing the user to invoke it via `/skill:help`. The documented guarantee that bare aliases are only registered when there is no collision is therefore violated for any skill whose name matches a built-in command. The collision check should see all commands, so `register_skill_commands()` should be called after `init_commands()` (and after `init_tools()` to still see loaded tools).

TimeToBuildBob:
## 🤖 AI code review

This PR adds skill invocation as slash commands (/skill:&lt;name&gt;) with bare aliases when they don't collide with existing commands or tools. It introduces a new module skill_commands.py that registers these commands, builds skill prompts with $ARGUMENTS substitution, and queues them via prompt_queue. The server and TUI now drain the prompt queue after command execution to turn queued prompts into user turns, and tests are added.

<details><summary><b>2 findings</b> · ❌ **2** P1 — 🛑 **2 at P0/P1**</summary>

**❌ P1 high** — `gptme/server/api_v2.py:1956`

When `build_skill_prompt` raises inside the skill handler, the handler catches the exception, yields nothing, and does not undo the command message. In the server, `_len_after_cmd_append` was captured after appending the command, and since the handler did not undo, `len(log.log.messages)` is unchanged, so the condition `if _is_skill and len(log.log.messages) >= _len_after_cmd_append:` is true. The server then emits a `message_added` SSE event for the original `/skill:name` user message. The log now contains that slash command as a normal user message. The next `/step` will see this user message and the assistant may respond to the raw slash command as if it were a user prompt, since the skill body was never substituted. This leaves the conversation in an ambiguous state: the user sees the slash command and no error, and the assistant may act on unintended instructions.

```suggestion
In gptme/lessons/skill_commands.py, change is_skill_command to also compare the handler identity, e.g.:

    def is_skill_command(name: str) -> bool:
        handler = _registered_skill_commands.get(name)
        if handler is None:
            return False
        from ..commands.base import _command_registry
        return _command_registry.get(name) is handler
```

<sub>**How this was verified:** Checked is_skill_command in gptme/lessons/skill_commands.py (line 240) – it returns `name in _registered_skill_commands` with no handler identity check. Confirmed the server uses this flag to gate the initial message_added event and to decide whether to drain the prompt queue.</sub>

**❌ P1 high** — `gptme/lessons/skill_commands.py:240`

is_skill_command() returns True for any name present in the module-level _registered_skill_commands dict, which is only updated when register_skill_commands() runs. If a foreign command (e.g. from a plugin) later overwrites a bare alias name in the command registry—say a plugin registers 'demo' after a skill named 'demo' registered it as an alias—the registry handler for 'demo' becomes the plugin's, but _registered_skill_commands['demo'] still holds the old skill handler. is_skill_command('demo') then returns True even though the actual command is not a skill. The server's api_v2 uses this to decide event order: it skips the initial message_added event, then after the foreign handler yields responses it checks len(log.log.messages) &gt;= _len_after_cmd_append. Since the foreign handler did not undo, the condition holds and it emits the original user-message event after the responses, violating the expected SSE order (user message must precede responses). Clients that rely on this order may render the assistant response before the user's typed command, or fail to reconcile the message. The fix is to make is_skill_command check the current registry, e.g. by also verifying that the registered handler is the one stored in _registered_skill_commands, or to update _registered_skill_commands whenever a foreign command overwrites a name.

<sub>**How this was verified:** Check the definition of is_skill_command and the fact that _registered_skill_commands is only mutated in register_skill_commands and unregister_skill_commands. In the server api_v2.py, the condition uses is_skill_command to skip the initial event (lines ~1940-1949) and later emits the original message event under the condition `if _is_skill and len(log.log.messages) >= _len_after_cmd_append` (line 1956). If a foreign handler yields messages and doesn't undo, the order breaks.</sub>

</details>

- `docs/skills.rst` — Adds documentation for invoking skills as /skill:&lt;name&gt; commands, including argument substitution and collision rules.
- `gptme/init.py` — Calls register_skill_commands() after init_commands() to expose skills as slash commands.
- `gptme/lessons/__init__.py` — Exports is_skill_command and register_skill_commands from the new skill_commands module.
- `gptme/lessons/skill_commands.py` — New module implementing skill command registration, argument substitution, prompt building, and handler logic.
- `gptme/server/api_v2.py` — Drains the durable prompt queue after a slash command and appends queued prompts as user messages; defers message_added events for skill commands until after handler execution.
- `gptme/tui/app.py` — Drains the durable prompt queue after a handled command, shows the skill prompt, and starts a generation turn.
- `tests/test_server.py` — Adds a test that /skill:&lt;name&gt; drains the prompt queue and becomes a user prompt.
- `tests/test_skill_commands.py` — New tests for argument substitution, command registration, collisions, idempotency, and handler behavior.
- `tests/test_tui.py` — Adds a test that the TUI drains the skill prompt queue and starts generation.

</details>

| `4f076e8e482f` | 4/5 | 2 | llm | 2026-08-22 19:04 UTC |
| `7b1046a5e45a` | 4/5 | 4 | llm | 2026-08-22 21:22 UTC |
| `80faa327edba` | 2/5 | 3 | llm | 2026-08-22 22:46 UTC |
| `96407150b899` | 3/5 | 2 | llm | 2026-08-23 00:46 UTC |
| `15f6526bf52e` | 2/5 | 2 | llm | 2026-08-23 09:29 UTC |
| `17c7e358ffde` | 3/5 | 2 | llm | 2026-08-23 11:13 UTC |
| `1f21f372509e` | 4/5 | 3 | llm | 2026-08-24 03:21 UTC |
| `35c0751c85ae` | 3/5 | 2 | llm | 2026-08-24 12:42 UTC |
| `8798a961ba1a` | 4/5 | 3 | llm | 2026-08-25 05:44 UTC |
| `dcbc9289d0de` | 2/5 | 4 | llm | 2026-08-25 07:49 UTC |

</details>

<sub>Reviewed `504815a73150` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 264s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Fixed in 66e560796: moved `register_skill_commands()` to after `init_commands()` so the bare-alias collision check sees the full built-in command registry. All 11 skill-command tests still pass.

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,193 @@
...
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS|\$(\d+)")
```
TimeToBuildBob: ❌ **P1** — The placeholder regex `$\d+` matches any dollar sign followed by digits, so a skill body containing a dollar amount like "$100" is interpreted as a reference to argument index 100. Since `substitute_arguments` replaces out-of-range indices with the empty string, the body "Set budget to $100" becomes "Set budget to ", silently deleting the amount. This corrupts any skill that mentions prices, currency, or other `$<number>` text, breaking the skill's instructions for every user who invokes it. The docs advertise `$N` as an argument placeholder, but nothing distinguishes `$N` from a literal dollar amount, so this is a genuine defect rather than a missing feature.

TimeToBuildBob: Fixed in 4f076e8e4: `_replace` now returns `match.group(0)` (the original text) instead of `""` when the positional index is out of range. A skill body containing `$100` or similar currency amounts will now survive substitution unchanged. Added a dedicated `test_substitute_arguments_preserves_dollar_amounts` regression test.

gptme/init.py
TimeToBuildBob: Already fixed in the prior arc session (commit 66e560796): `register_skill_commands()` is now called after `init_commands()` on line 100 of `init.py`, so the bare-alias collision check sees all built-in commands.

TimeToBuildBob:
Both P1 findings from the AI reviewer are now resolved:

**Finding 1 — `register_skill_commands()` called before `init_commands()`** (fixed in prior session, commit `66e560796`): the call is now on line 100 of `init.py`, after `init_commands()` on line 97, so the bare-alias collision check sees the full command registry including all built-in commands.

**Finding 2 — `$N` out-of-range silently deletes dollar amounts** (fixed in `4f076e8e4`): `substitute_arguments` now returns the original match text (`match.group(0)`) instead of `""` when a positional index is out of range. Skill bodies containing currency like `$100` pass through unchanged. New regression test: `test_substitute_arguments_preserves_dollar_amounts`.

All 12 skill-command tests pass.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,195 @@
...
+                continue
+
+            try:
+                handler = _make_skill_handler(skill, index)
+            except Exception as e:
+                logger.warning("Failed to build command for skill %r: %s", name, e)
+                continue
+
+            canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            register_command(canonical, handler)
```
TimeToBuildBob, line 193: ⚠️ **P2** — The canonical skill command name (e.g. 'skill:demo') is registered without checking whether an existing command already occupies that name in _command_registry. The bare alias is protected by a collision check (lines 163-178), but the canonical is not: line 159 unconditionally calls register_command(canonical, handler). If a plugin or any other component registered a command with the 'skill:' prefix before this runs, it will be silently overwritten, breaking that plugin's functionality. This contradicts the code's own comment that the canonical form 'never collides' — that is only true if nothing else uses the 'skill:' namespace. A plugin legitimately using a command named 'skill:demo' would stop working without any error.

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,195 @@
...
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS|\$(\d+)")
```
TimeToBuildBob: ⚠️ **P2** — The regex alternation in _ARG_PATTERN matches $ARGUMENTS even when it is a prefix of a larger token (e.g. '$ARGUMENTSfoo'). The pattern '$ARGUMENTS' does not require a word boundary, so substitute_arguments() will replace the ' $ARGUMENTS' part and leave the trailing 'foo', turning '$ARGUMENTSfoo' into 'foofoo' when args is 'foo'. This is likely unintended for skill prose that uses the placeholder adjacent to other characters, and can produce garbled prompts. The tests only cover standalone '$ARGUMENTS' and '$ARGUMENTS[N]' forms, so this edge case is not pinned down.

TimeToBuildBob:
All CI green (including Greptile Review pass). MERGEABLE/CLEAN. Self-merge not eligible (new feature files require human review). Ready for review and merge.

TimeToBuildBob:
Both P2 findings from the latest AI review (score 4/5 → targeting 5/5) fixed in `7b1046a5e`:

**P2-1 — `\$ARGUMENTS` regex partial match** (`skill_commands.py:43`): Added `\b` word boundary so `$ARGUMENTSfoo` in skill prose is left intact instead of expanding `$ARGUMENTS` and appending `foo`. New test: `test_substitute_arguments_word_boundary`.

**P2-2 — Canonical `skill:<name>` registration without collision check** (`skill_commands.py:159`): Added a `_command_registry` lookup before calling `register_command(canonical, handler)` — skips (with debug log) when another component already owns that name. New test: `test_canonical_does_not_clobber_foreign_skill_prefix_command`.

14 tests pass. AI reviewer will re-score on this push.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,201 @@
...
+    name = skill.metadata.name or skill.path.stem
+    header = f"Skill invoked: /{SKILL_COMMAND_PREFIX}{name}"
+    if full_args:
+        header += f" {full_args}"
+    body = substitute_arguments(skill.body, full_args, args)
+    return f"{header}\n\n{body}".rstrip()
+
+
+def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
+    name = skill.metadata.name or skill.path.stem
```
TimeToBuildBob: ⚠️ **P2** — In _make_skill_handler, the handler derives its name from the raw skill metadata name without stripping, while register_skill_commands strips the name (name = name.strip()) before using it for the command registration. For a skill with leading/trailing spaces in its name (e.g. "  demo  "), the command is registered as "/skill:demo" but the handler builds the prompt header with the unstripped name, producing "Skill invoked: /skill:  demo  " and setting handler.__name__ to "skill__demo__" (invalid identifier). This inconsistency can confuse the assistant and break any code relying on the handler name. The fix is to strip the name in _make_skill_handler (and likewise in build_skill_prompt). _(2/2 passes)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def handler(ctx: CommandContext) -> Generator[Message, None, None]:
+        from ..prompt_queue import queue_prompt  # fmt: skip
+
+        # Remove the "/skill:<name>" command message from the log, like built-in
+        # commands do; the queued prompt carries its own "Skill invoked" header.
+        ctx.manager.undo(1, quiet=True)
+        ctx.manager.write()
+
+        content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
```
TimeToBuildBob: ⚠️ **P2** — The skill command handler does not catch exceptions from build_skill_prompt. If a skill is a stub and materialize_lesson raises (e.g., SKILL.md is missing or malformed), the exception propagates out of the handler. Since handle_cmd consumes the generator, this can crash the chat loop instead of degrading to an error message. Registration time is protected by try/except, but invocation is not. A user with a broken skill directory would be unable to use any command and may see an unhandled traceback. Adding a try/except around build_skill_prompt (or around the materialize call) that logs a warning and returns an error message would keep the session alive. _(1/2 passes — pass 2 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def _loaded_tool_names() -> set[str]:
+    try:
+        from ..tools import get_tools  # fmt: skip
+
+        return {tool.name for tool in get_tools()}
+    except Exception as e:  # pragma: no cover - defensive
+        logger.debug("Could not list loaded tools for skill alias check: %s", e)
+        return set()
```
TimeToBuildBob: ⚠️ **P2** — `_loaded_tool_names` catches every exception from `get_tools` and returns an empty set. `register_skill_commands` then uses `name in tool_names` to decide whether to register the bare `/<name>` alias. If `get_tools` fails for any reason, `tool_names` is empty, so the condition `elif name in tool_names` is never true, and a skill named exactly like a loaded tool (e.g., "shell") will get a bare `/shell` alias. This violates the stated invariant that a skill must never shadow a tool, and does so precisely when the tool list is unavailable. An attacker who can place a skill directory (already a supported feature) could exploit a transient `get_tools` failure to hijack `/shell` and redirect it to a malicious prompt, defeating the explicit collision-avoidance goal of the PR. _(1/2 passes — pass 1 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,201 @@
...
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$(\d+)")
```
TimeToBuildBob: ⚠️ **P2** — The substitution regex `\$(\d+)` treats any dollar sign followed by digits as a placeholder `$N`. Skill bodies frequently contain literal currency amounts, e.g., "The price is $100.". If the user invokes the skill with enough arguments (more than 100 in this example), `$100` is replaced with `args[100]`, silently corrupting the prompt content to "The price is <arg100>.". Since the skill body is trusted prose and the substitution is invisible in the queued prompt, the assistant will act on the corrupted text, potentially performing a wrong action. This is a functional defect caused by the placeholder syntax not being namespaced away from natural dollar amounts. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Three P2 fixes from the AI reviewer's third pass (commit `80faa327e`):

**P2-1 — `_make_skill_handler` / `build_skill_prompt` using unstripped name** (`skill_commands.py:81`): Both functions now call `.strip()` on the name derived from `skill.metadata.name`. Without this, a skill with leading/trailing whitespace in its metadata name would produce a mismatched command registration (`/skill:demo`) and prompt header (`Skill invoked: /skill:  demo  `). (2/2 pass finding)

**P2-2 — Unguarded `build_skill_prompt` call in handler** (`skill_commands.py:91`): The `build_skill_prompt` call inside the command handler is now wrapped in `try/except`. A missing or malformed `SKILL.md` no longer propagates out and crashes the chat loop — it logs a warning and returns cleanly instead.

**P2-3 — `_loaded_tool_names()` returning empty set on failure** (`skill_commands.py:111`): Changed from returning `set()` to a sentinel `_TOOL_NAMES_UNAVAILABLE`. `register_skill_commands` now checks for the sentinel and suppresses all bare aliases when tool enumeration fails, rather than accidentally granting a `/<name>` alias to a skill that shares its name with a real tool (fail-safe, not fail-open). Failure now logs at `WARNING` level so it's visible in production.

All 14 tests still pass.

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    def _replace(match: re.Match[str]) -> str:
+        idx_str = match.group(1) or match.group(2)
+        if idx_str is None:
+            return full_args
+        idx = int(idx_str)
+        # Return the original text unchanged when the index is out of range,
+        # so "$100" in skill prose (prices, currency) is never silently deleted.
+        return args[idx] if idx < len(args) else match.group(0)
```
TimeToBuildBob, line 64: ❌ **P1** — The regex `\$(\d+)` in `_ARG_PATTERN` matches any `$` followed by digits, so a skill body containing a literal dollar amount like `$100` is treated as a reference to argument index 100. When the user supplies at least 101 arguments, the replacement function returns `args[100]`, silently replacing the prose amount with the 101st argument. The documented intent to preserve out-of-range indices (so `$100` in prices is left alone) only holds when the number of arguments is small; with enough arguments the substitution corrupts the skill content. For example, a skill saying 'The price is $100 per item' becomes 'The price is <101st arg> per item'. This is a real correctness bug because dollar amounts are common in skill prose and the behavior is surprising. _(3/3 passes)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,222 @@
...
+                logger.warning("Failed to build command for skill %r: %s", name, e)
+                continue
+
+            canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            if canonical in _command_registry:
+                logger.debug(
+                    "Skill canonical %r already registered by another component; skipping",
+                    canonical,
+                )
+                continue
```
TimeToBuildBob, line 192: ⚠️ **P2** — In `register_skill_commands`, if a skill's canonical name (e.g., `skill:demo`) is already present in `_command_registry` because another component registered it after our initial call, the `continue` at the canonical-collision check skips registering both the canonical and the bare alias. Since the unregister loop at the start removes our previous bare alias (when it is still ours), the skill becomes completely inaccessible after re-registration. The existing test `test_canonical_does_not_clobber_foreign_skill_prefix_command` only covers a foreign command registered *before* our first call; the takeover-after-registration case is untested and leads to a permanent loss of that skill's command. Users would type `/skill:demo` or `/demo` and see the unknown-command fallback, with no way to invoke the skill. _(1/3 passes — passes 1, 3 looked and disagreed)_

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
P1 fix from the AI reviewer's latest pass (commit `96407150b`):

**Finding `1f19ad1c75e3` — `$(\d+)` matches dollar amounts when enough args supplied**

The "return original when out-of-range" fix only protected against dollar amounts in skill prose when the user supplied *fewer* args than the number in the amount. With 101+ args, `$100` in skill prose would be silently replaced with `args[100]`. The reviewer's 3/3 consensus rightly flagged this as a real correctness bug.

**Fix**: Changed the positional placeholder syntax from `$N` to `${N}` (curly-brace form, like bash). The pattern `\$\{(\d+)\}` never matches a bare `$100`, regardless of how many arguments the user provides. Dollar amounts in skill prose are now completely safe.

- `_ARG_PATTERN`: `\$(\d+)` → `\$\{(\d+)\}`
- Module docstring, `substitute_arguments` docstring, `docs/skills.rst` updated
- `DEMO_BODY` and `test_substitute_arguments` updated to use `${1}` / `${5}`
- `test_substitute_arguments_preserves_dollar_amounts` extended with a 101-arg case confirming `$100` in prose is NOT substituted even with enough args present

14 tests pass.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def build_skill_prompt(
+    skill: Lesson, index: LessonIndex, full_args: str, args: list[str]
+) -> str:
+    """Materialize a skill and render it as a user prompt."""
+    if skill.is_stub:
+        skill = index.materialize_lesson(skill)
+    name = (skill.metadata.name or skill.path.stem).strip()
+    header = f"Skill invoked: /{SKILL_COMMAND_PREFIX}{name}"
```
TimeToBuildBob, line 76: ⚠️ **P2** — In build_skill_prompt, the name variable is recomputed after a potential materialize_lesson(skill) call, using the materialized lesson's metadata.name or path.stem. However, the command was registered under the name taken from skill.metadata.name at registration time (in register_skill_commands, line 168). If a stub skill's materialized version yields a different name (for example, the frontmatter name differs from the directory name, or materialization changes it), the queued prompt header will reference a different skill name than the command that was invoked. This inconsistency can confuse the model or user and breaks logging/matching that relies on the skill name. The fix is to pass the registration-time name into build_skill_prompt and use it for the header. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob: Fixed in commit `80faa327e`: both `build_skill_prompt` and `_make_skill_handler` now call `.strip()` on the name derived from `skill.metadata.name or skill.path.stem`. The registered command name and the prompt header are now computed the same way.

gptme/lessons/skill_commands.py
TimeToBuildBob: Fixed in commit `80faa327e`: wrapped the `build_skill_prompt` call in `try/except` inside the handler. A missing or malformed SKILL.md now logs a warning and returns cleanly instead of propagating the exception.

TimeToBuildBob:
All AI reviewer findings are now addressed and inline threads resolved. Status:

**Fixes shipped across 5 commits** (`66e560796` → `96407150b`):
- `init_commands()` ordering: `register_skill_commands()` now called after `init_commands()` — bare alias collision check sees full built-in registry
- `$N` → `${N}` syntax: positional placeholders now require curly braces, so dollar amounts like `$100` in skill prose are never substituted
- `\$ARGUMENTS\b` word boundary: prevents partial match on `$ARGUMENTSfoo`
- Canonical `skill:<name>` collision guard: skips registration if another component already owns the name
- `_make_skill_handler` name stripping: consistent `.strip()` in both handler and `build_skill_prompt`
- `build_skill_prompt` exception guard: `try/except` in handler so a broken skill directory can't crash the chat loop
- `_loaded_tool_names()` sentinel: returns `_TOOL_NAMES_UNAVAILABLE` on failure (fail-safe, no bare aliases on tool-list error)

**CI**: all green. **Greptile**: 5/5. **MERGEABLE/CLEAN**. Waiting for human review.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,225 @@
...
+
+def _make_skill_handler(skill: Lesson, index: LessonIndex) -> CommandHandler:
+    name = (skill.metadata.name or skill.path.stem).strip()
+
+    def handler(ctx: CommandContext) -> Generator[Message, None, None]:
+        from ..prompt_queue import queue_prompt  # fmt: skip
+
+        # Remove the "/skill:<name>" command message from the log, like built-in
+        # commands do; the queued prompt carries its own "Skill invoked" header.
+        ctx.manager.undo(1, quiet=True)
```
TimeToBuildBob: ❌ **P1** — In _make_skill_handler, ctx.manager.undo(1, quiet=True) and ctx.manager.write() are called unconditionally at the top of the handler, before build_skill_prompt is attempted. If build_skill_prompt raises (e.g. index.materialize_lesson fails on a missing or malformed skill file), the command message is already removed from the log and the manager log is written, yet nothing is queued. The user's /skill:<name> invocation silently disappears with no prompt and no error visible. The same holds if queue_prompt raises, since the exception propagates after the undo has already happened. The fix is to build (and queue) first, then undo the command message and write, so a failure leaves the log intact and the user can see what went wrong. _(3/3 passes)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,225 @@
...
+                continue
+
+            try:
+                handler = _make_skill_handler(skill, index)
+            except Exception as e:
+                logger.warning("Failed to build command for skill %r: %s", name, e)
+                continue
+
+            canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            if canonical in _command_registry:
```
TimeToBuildBob, line 187: ❌ **P1** — When a skill named `X` coexists with a skill named `skill:X`, registration order determines whether `X` gets a command. If `skill:X` is processed first, its bare alias `skill:X` is inserted into `_command_registry` (lines 206-208). When `X` is later processed, its canonical name `skill:X` is found in the registry and the `continue` at line 188 skips it, so no `/skill:X` command exists. Because `LessonIndex.lessons` order is filesystem-dependent, the same set of skills can register differently across runs, silently dropping a valid skill. This is a real collision between our own canonical and alias, not a foreign command, so it should not suppress registration. Fix by registering all canonicals in a first pass and then adding aliases only if they are not a canonical we just registered. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Two P1 bugs verified in the code and fixed in `17c7e358f`:

**Finding `c1f8e676a44c` — undo before build_skill_prompt**

The previous fix added a `try/except` around `build_skill_prompt` but the `undo(1)` call still happened before it. A broken skill would still silently remove the `/skill:<name>` invocation from the log with no user-visible error. Fixed: build + queue first, undo on success only. New test: `test_handler_undo_not_called_when_build_fails` verifies `manager.undo` is never called when build raises.

**Finding `0a6ea07fd8c5` — registration order determines whether skill `X` gets a command**

When a skill named `skill:X` ran first, its bare alias `skill:X` occupied `_command_registry`. Skill `X` then found its canonical `skill:X` already present and was silently skipped (filesystem-order-dependent). Fixed with two-pass registration: pass 1 registers all canonicals, pass 2 adds bare aliases checking against `our_canonicals` to skip only real external collisions, not our own aliases. New test: `test_two_pass_registration_skill_named_with_prefix`.

Also rebased onto current master (was 5 commits behind). CI running.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,234 @@
...
+            register_command(canonical, handler)
+            _registered_skill_commands[canonical] = handler
+            registered.append(canonical)
+            skill_entries.append((name, canonical, handler))
+
+        # Pass 2: register bare aliases, checking against all canonicals we just
+        # registered so a skill named "skill:X" doesn't block bare alias for "X".
+        our_canonicals = {canonical for _, canonical, _ in skill_entries}
+        for name, canonical, handler in skill_entries:
+            if name in _command_registry and name not in our_canonicals:
```
TimeToBuildBob: ❌ **P1** — In register_skill_commands, pass 2's collision check excludes names that appear in our_canonicals: `if name in _command_registry and name not in our_canonicals:`. When two skills are named 'skill:demo' and 'demo', the first skill's bare alias is 'skill:demo', which is also the canonical of the second skill. Because 'skill:demo' is in our_canonicals (set during pass 1), the first branch is skipped and the bare alias is registered, overwriting the canonical handler for 'demo'. Invoking /skill:demo then runs the wrong skill. The intended behavior is that any existing command, including a canonical we just registered, blocks the bare alias; the exclusion was meant for the case where a skill named 'skill:X' gets a bare alias 'X', which is not a collision since 'X' is not a canonical. Fix by dropping the `and name not in our_canonicals` part of the condition. _(3/3 passes)_

tests/test_skill_commands.py
```diff
@@ -0,0 +1,329 @@
...
+def test_two_pass_registration_skill_named_with_prefix(skills_root: Path):
+    """Skill 'X' gets a canonical even when another skill is named 'skill:X'."""
+    # Skill with a name that happens to start with the "skill:" prefix.
+    _write_skill(skills_root, "skill:demo", "Weird name", "Body")
+    # Skill whose canonical would be "skill:demo" — must not be blocked by the alias
+    # that the first skill registers for its bare name.
+    _write_skill(skills_root, "demo", "Normal skill", DEMO_BODY)
+    registered = register_skill_commands()
+    # "skill:demo" canonical must exist for the second skill.
+    assert "skill:demo" in registered
```
TimeToBuildBob, line 335: ⚠️ **P2** — The test test_two_pass_registration_skill_named_with_prefix only asserts that 'skill:demo' is in the registered list. Because the bug in register_skill_commands registers the bare alias 'skill:demo' (overwriting the canonical), the string is present regardless of which skill's handler ends up behind it. The test therefore passes vacuously and fails to catch that /skill:demo no longer invokes the 'demo' skill. It should assert that the handler for 'skill:demo' is the one built for the 'demo' skill, e.g. by checking its __name__ attribute or comparing handler identity. _(1/3 passes — passes 1, 2 looked and disagreed)_

gptme/lessons/skill_commands.py
TimeToBuildBob: Fixed in 1f21f3725.

Dropped `and name not in our_canonicals` from the pass-2 collision check. Any existing registry entry — including a canonical registered in pass 1 — now blocks the bare alias, so a skill named `skill:demo` can no longer overwrite `/skill:demo` belonging to skill `demo`.

tests/test_skill_commands.py
TimeToBuildBob: Fixed in 1f21f3725.

`test_two_pass_registration_skill_named_with_prefix` now asserts handler identity (`__name__ == "skill_demo"`), both canonicals, the `/demo` alias, and the queued prompt bodies. Confirmed the old membership-only assertion would still pass on the clobbering bug; this one fails without the pass-2 fix.

TimeToBuildBob:
P1 from the latest AI review pass (commit `17c7e358f`) fixed in `1f21f3725`:

**Finding `4dc871392fd2` — pass 2 clobbers a canonical with a prefix-named skill's alias**

Pass 2 treated `name not in our_canonicals` as "not a collision", so a skill named `skill:demo` registered its bare alias as `skill:demo` and overwrote skill `demo`'s canonical handler. `/skill:demo` then invoked the wrong skill.

**Fix**: any existing registry entry now blocks the bare alias. `our_canonicals` is gone — pass 1 already owns the canonicals, and pass 2 must not overwrite them.

**Finding `5e9684dfa6ef` (P2)** — the old two-pass test only asserted `"skill:demo" in registered`, which stays true even after the clobber. It now checks handler identity, both canonicals, the `/demo` alias, and the queued prompt bodies. Confirmed it fails on the unfixed code (`skill_skill:demo` vs `skill_demo`).

18 tests pass locally (`poetry run pytest tests/test_skill_commands.py`). CI running.

This also unblocks PM dispatch: four claude-code workers had failed in ~25s (Claude subscription paused), so retry budget exhausted on SHA `17c7e358f`. New head SHA clears the escalated `.retry` state.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS\b|\$\{(\d+)\}")
```
TimeToBuildBob: ⚠️ **P2** — The bare $ARGUMENTS alternative in the regex also matches when $ARGUMENTS is followed by a '[' that does not form a numeric index. For a skill body containing 'Use $ARGUMENTS[name]' with full_args='hello', the regex matches $ARGUMENTS (since '[' is a non-word char, the word boundary holds) and replaces it, yielding 'Use hello[name]'. This corrupts skill text that uses bracket notation non-numerically, and diverges from the documented placeholder syntax. The fix is to require that $ARGUMENTS is not followed by '[' in the bare form, e.g. using a negative lookahead. _(3/3 passes)_

tests/test_skill_commands.py
```diff
@@ -0,0 +1,351 @@
...
+    manager.undo.assert_not_called()
+
+
+def test_two_pass_registration_skill_named_with_prefix(skills_root: Path, manager):
+    """Skill 'X' keeps its canonical when another skill is named 'skill:X'.
+
+    Pass 2 must not register the prefix-named skill's bare alias over
+    /skill:X — that name is already the canonical of skill X.
+    """
+    _write_skill(skills_root, "skill:demo", "Weird name", "Body of prefix skill")
```
TimeToBuildBob, line 331: ⚠️ **P2** — test_two_pass_registration_skill_named_with_prefix writes a skill directory named 'skill:demo', which is an invalid file name on Windows (colon is a reserved character). Creating this directory raises an error, so the test fails on Windows regardless of the code under test. The test is not hermetic and depends on a POSIX filesystem. Use a name that is valid on all platforms, such as 'skill-demo', or avoid creating a directory with a colon. _(1/3 passes — passes 2, 3 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # Build the prompt first; only undo the command message on success so a
+        # broken skill leaves the log intact instead of silently disappearing.
+        try:
+            content = build_skill_prompt(skill, index, ctx.full_args.strip(), ctx.args)
+        except Exception as e:
+            logger.warning("Failed to build prompt for skill %r: %s", name, e)
+            yield from ()
+            return
+        queue_prompt(ctx.manager.logdir, content)
```
TimeToBuildBob, line 97: ⚠️ **P2** — The handler wraps only `build_skill_prompt` in a try/except. If `queue_prompt`, `ctx.manager.undo`, or `ctx.manager.write` raises (e.g., disk full, permission error, or lock contention), the exception propagates out of the generator and is not caught. In the chat loop this will abort the command handler and may crash the loop or leave the queue with a prompt while the command message remains, causing a duplicate turn or an unhandled error. The comment promises undo happens only on success, but there is no guard for these later calls. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
@greptileai review

gptme/server/api_v2.py
```diff
@@ -1922,17 +1923,27 @@ def api_conversation_post(conversation_id: str):
...
-                        "type": "message_added",
-                        "message": msg2dict(resp, log.workspace, log.logdir),
-                    },
-                )
+                _emit(resp)
+            # Skill commands (and any other handler using queue_prompt) write
+            # a follow-up user prompt to the durable queue instead of yielding.
+            # The CLI drains that queue at the top of its chat loop; drain
+            # here so the next /step sees the prompt as a normal user message.
+            for queued in drain_prompt_queue(log.logdir):
```
TimeToBuildBob, line 1968: ❌ **P1** — If a command handler raises an exception after queueing a prompt (for example, the undo call fails or queue_prompt raises after writing), the durable prompt queue is never drained because the code that drains it runs only after handle_cmd completes without an exception. The queue file then retains the prompt, and the next command executed in that conversation will drain it, injecting an unexpected user message into the conversation. This affects both the server (api_v2.py) and the TUI (app.py), since both only drain the queue on the success path. The fix is to drain the queue in a finally block (or also in the except handler) so leftover prompts from a failed command do not leak into a later request. _(2/3 passes — pass 1 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
+        # overwritten. Without that, a skill named "skill:X" would register
+        # its bare alias as "skill:X" and clobber skill "X"'s canonical.
+        for name, canonical, handler in skill_entries:
+            if name in _command_registry:
+                logger.debug(
+                    "Skill %r collides with existing command; only /%s registered",
+                    name,
+                    canonical,
+                )
+            elif not tools_available or name in tool_names:  # type: ignore[operator]
```
TimeToBuildBob: ⚠️ **P2** — register_skill_commands() decides whether to register a bare /<name> alias by checking the set of tool names loaded at registration time (tool_names from _loaded_tool_names()). If a skill named 'shell' is present and the shell tool is not loaded when init runs (e.g., server started with a tool allowlist excluding shell), the bare alias 'shell' is registered. gptme permits tools to be enabled later via /tools or per-conversation allowlists. Once the shell tool is loaded, handle_cmd resolves '/shell' to the registered command before falling back to tools, so the skill body runs instead of the shell tool. This can silently execute an unintended skill, and if the skill is malicious it could cause harmful behavior. The collision check must be evaluated at invocation time or the alias suppressed whenever a tool with that name exists. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
P2 fix from this review cycle (commit `8798a961b`): `$ARGUMENTS[name]` (non-numeric bracket) was being matched by the bare `$ARGUMENTS` alternative because `[` is a non-word character and `\b` matched there. Added a negative lookahead `(?!\[)` so only `$ARGUMENTS` not followed by `[` matches that arm — numeric `$ARGUMENTS[N]` is already caught by the first alternative. Test added to pin the edge case.

CI running on the new push.

TimeToBuildBob:
CI is now green on the latest commit (`8798a961b`). All checks pass (Benchmark, Build, Docs, Lint, typecheck, Test). Self-merge not eligible (new feature). Waiting for human review and merge.

gptme/tui/app.py
```diff
@@ -1397,8 +1398,28 @@ def _handle_command(self, text: str) -> None:
...
+
+        Skill commands write to the durable prompt queue instead of yielding a
+        message. The CLI drains that queue at the top of its chat loop; without
+        this, the prompt sits until a CLI session opens the same log.
+        """
+        drained = drain_prompt_queue(self.manager.logdir)
+        if not drained:
+            return
+        first, *rest = drained
+        self.manager.append(first)
```
TimeToBuildBob, line 1418: ⚠️ **P2** — In tui/app.py, `_drain_command_queued_prompts` drains the durable prompt queue and, if it is non-empty, appends only the first message to the log (line 1418) and queues the rest in `self.prompt_queue` (line 1420). Only the first is displayed immediately; the others are not shown until the generation worker submits them later. If a skill command ever queues more than one prompt (e.g., due to stale entries), the user sees only one new message and the rest appear without context. This is a minor UX inconsistency rather than a data-loss bug. _(2/3 passes — pass 2 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
```
TimeToBuildBob: ⚠️ **P2** — The bare-alias collision check compares skill names against loaded tool names with exact case sensitivity (`name in tool_names`). Skill names are not normalized, so a skill named `Shell` and a tool named `shell` are treated as non-colliding; the bare alias `/Shell` is registered even though a user typing `/shell` would still hit the tool. This means a skill can shadow a tool whose name differs only by case, contradicting the stated goal that "a skill named `shell` must never shadow the shell tool" and potentially confusing users who type the wrong case. The impact is limited to case-variant naming, but it is a real hole in the collision guard. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
Three P2 fixes from the latest AI review cycle, in commit `dcbc9289d`:

**P2-1 — SSE phantom `message_added` event for skill commands** (`api_v2.py`): When a `/skill:<name>` command is posted, the server previously emitted `message_added` for the command message before running the handler. Skill handlers then call `ctx.manager.undo(1)` to remove that message (the skill prompt is the real user turn), leaving WebUI clients with a phantom message they can't reconcile. Fix: defer the `message_added` event for the command message to after `handle_cmd` finishes and only emit it when `log.log.messages` length indicates the message was retained (skill handlers reduce it below the pre-command baseline).

**P2-2 — Case-sensitive tool name collision** (`skill_commands.py:209`): The bare-alias collision check used `name in tool_names` (exact case), so a skill named `Shell` was treated as non-colliding with the `shell` tool and got a `/Shell` alias. Fix: `name.lower() in {t.lower() for t in tool_names}`. New test: `test_collision_with_loaded_tool_is_case_insensitive`.

**P2-3 (TUI first-message-only)** was reviewed and left as-is: the behaviour is correct for normal usage — skills queue exactly one prompt, and the rest path (`prompt_queue.extend`) handles multi-prompt stale drain correctly by feeding them as sequential user turns rather than dumping them all at once.

All 20 skill-command tests pass. mypy clean on the changed files.

TimeToBuildBob:
CI is green on the latest commit (`dcbc9289d`). All checks pass (Benchmark, Build, Docs, Lint, typecheck, Test). Self-merge not eligible (new feature files in core modules). Ready for human review and merge.

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
+                continue
+
+            canonical = f"{SKILL_COMMAND_PREFIX}{name}"
+            if canonical in _command_registry:
+                logger.debug(
+                    "Skill canonical %r already registered by another component; skipping",
+                    canonical,
+                )
+                continue
+            register_command(canonical, handler)
```
TimeToBuildBob, line 193: ❌ **P1** — In `register_skill_commands`, the outer `try` wraps the entire loop, but individual `register_command` calls are not protected. If `register_command` raises for any reason other than the already-checked collision — for example, a registry that is an immutable mapping or a provider that rejects the name — the exception is caught by the outer `except Exception`, logged, and the loop aborts. Commands already registered before the raise are left in `_command_registry` but not recorded in `_registered_skill_commands`, because that dict is updated after `register_command` returns. On a subsequent `register_skill_commands` call, the cleanup loop only unregisters names whose stored handler matches the registry entry; these orphaned names are never removed. A later skill with the same canonical name then hits `if canonical in _command_registry` and is skipped, silently dropping the skill. The user sees the old command still working and the new skill never appears in `/help` or tab completion. _(1/3 passes — passes 2, 3 looked and disagreed)_

gptme/server/api_v2.py
```diff
@@ -1915,24 +1916,46 @@ def api_conversation_post(conversation_id: str):
...
         try:
             for resp in handle_cmd(msg.content, log):
-                log.append(resp)
-                responses.append(resp)
+                _emit(resp)
+            # Emit the command-message event only when the command did not undo
+            # it. Skill handlers undo the command message after queuing the
+            # skill prompt (handle_cmd yields nothing for them), which reduces
+            # log length below _len_after_cmd_append.
+            if len(log.log.messages) >= _len_after_cmd_append:
```
TimeToBuildBob: ❌ **P1** — The server now defers the `message_added` event for the command message until after all command responses have been emitted. The previous behavior emitted the command event immediately after appending the command message, before calling `handle_cmd`. Now `_emit` appends each response and emits its event inside the loop, and the command event is emitted afterward, conditioned on the log length still being at least `_len_after_cmd_append`. For any command that yields responses (e.g., `/help` or a custom command that returns a message), the response events now arrive before the command event. This reverses the event order that existing SSE clients expect (user message, then assistant/system messages), so the web UI can render an assistant response without its preceding user message, or reorder messages incorrectly. This is a regression for all non-skill slash commands that produce output. _(2/3 passes — pass 1 looked and disagreed)_

gptme/lessons/skill_commands.py
```diff
@@ -0,0 +1,235 @@
...
+logger = logging.getLogger(__name__)
+
+SKILL_COMMAND_PREFIX = "skill:"
+
+# Names registered by the most recent register_skill_commands() call.
+# Used to make re-registration idempotent and to let our own previous bare
+# aliases not count as collisions.
+_registered_skill_commands: dict[str, CommandHandler] = {}
+
+_ARG_PATTERN = re.compile(r"\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS(?!\[)\b|\$\{(\d+)\}")
```
TimeToBuildBob, line 43: ⚠️ **P2** — The substitution regex `\$ARGUMENTS\[(\d+)\]|\$ARGUMENTS(?!\[)\b|\$\{(\d+)\}` matches `${N}` anywhere in the skill body, including inside code blocks or shell scripts, and replaces it with the N-th argument. This is the intended positional placeholder, but it also corrupts skill bodies that legitimately contain shell parameter expansion like `${1}`. For example, a bash-lesson skill that includes the example `echo ${1}` will have that replaced with the first argument when invoked, so the example is no longer literal. Users see mutated skill content in the prompt, which is confusing and incorrect. The placeholder syntax conflicts with valid POSIX shell syntax and should be scoped (e.g., only outside fenced code blocks) or made more specific. _(1/3 passes — passes 1, 3 looked and disagreed)_

TimeToBuildBob:
Two P1 fixes from the latest AI review cycle, in commit `504815a73`:

**P1-1 — SSE event order regression for non-skill commands** (`api_v2.py`): The deferred `message_added` approach (introduced to avoid phantom events for skill handlers) emitted command events AFTER responses for ALL commands, not just skill commands. For `/help`, `/undo`, or any non-skill command that yields responses, clients received response events before the user command that triggered them — reversing the expected SSE ordering. Fix: detect the command type upfront with `is_skill_command()` and emit the command event immediately for non-skill commands; defer only for skill handlers where undo is expected.

**P1-2 — Durable queue not drained on handler exception** (`api_v2.py`): If a skill handler queued a prompt then raised (e.g. `undo()` failed mid-handler), `drain_prompt_queue()` was never called because it sits after `handle_cmd` in the success path. The stale queued prompt would then inject into the next command's conversation turn as an unexpected user message. Fix: drain (and discard) in the `except` handler so a failed command leaves no leftover queue entries.

Also added `is_skill_command()` to `gptme.lessons.skill_commands` (and exported from `gptme.lessons`), with a test pinning expected return values. CI running on the new push.

TimeToBuildBob:
Automated merge handoff for `504815a73150`:

- CI gate: satisfied
- Review threads: clear
- `gptme/init.py`
- `gptme/lessons/__init__.py`
- `gptme/lessons/skill_commands.py`
- `gptme/server/api_v2.py`
- `gptme/tui/app.py`

PR #3592, opened 2026-08-22
fix(server): don't let step() finally clear a continuation reservation

TimeToBuildBob:
## Why

`step()`'s `finally` block still did `session.generating = False` unconditionally. The tool-continuation path already increments `session.step_seq` before re-reserving generation. If an auto-confirm tool finishes *before* the originating `step()` reaches `finally`, that increment happens first and then `finally` erases the continuation's reservation. A later `/step` can start concurrently with the continuation (duplicate LLM turns).

This is Race 5 from the interrupt/continuation work. gptme/gptme#3457 landed the other races; this slice existed as `8cb2e89c` / `recovered/server-continuation-reservation-test` and never made it onto master. Cherry-picked onto current master: kept today's continuation-election path (it already bumps `step_seq`) and added the missing `finally` compare-and-clear.

Recovered from `tasks/triage-recovered-branches-2026-08-10.md`.

## Change

- Snapshot `step_seq` before streaming.
- In `step()` `finally`, clear `generating` only if `step_seq` is still ours.
- Regression: `test_continuation_reservation_not_cleared_by_originating_step`.

## Test plan

- [x] `pytest tests/test_server_v2_sessions.py::TestInterruptEndpoint` — 16 passed (includes the new Race 5 test)

### DIFF
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -727,6 +727,11 @@ def api_conversation_tool_confirm(conversation_id: str):
 
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            # Advance step_seq before capturing the epoch so that any
+            # concurrent tool worker whose my_seq equals the pre-skip value
+            # sees a mismatch and stands down (same protocol as /step and
+            # /interrupt + rerun).
+            session.step_seq += 1
             skip_step_seq = session.step_seq
             try:
                 current_tool.status = ToolStatus.SKIPPED
--- a/gptme/server/api_v2_sessions.py
+++ b/gptme/server/api_v2_sessions.py
@@ -506,6 +506,7 @@ def api_conversation_step(conversation_id: str):
                 auto_confirm=auto_confirm_enabled,
                 stream=stream,
                 reserved=True,
+                step_seq=step_seq,
             )
         _step_dispatched = True
     finally:
@@ -726,6 +727,7 @@ def api_conversation_tool_confirm(conversation_id: str):
 
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            skip_step_seq = session.step_seq
             try:
                 current_tool.status = ToolStatus.SKIPPED
                 session.pending_tools.pop(tool_id)
@@ -751,6 +753,7 @@ def api_conversation_tool_confirm(conversation_id: str):
                     chat_config.workspace,
                     branch=current_tool.branch,
                     reserved=True,
+                    step_seq=skip_step_seq,
                 )
             finally:
                 if not continuation_dispatched:
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -783,6 +783,16 @@ def step(
     if tool_format == "tool":
         tools = [t for t in get_tools() if t.is_runnable]
 
+    # Snapshot the step sequence at entry. The finally block uses this to detect
+    # whether the tool-continuation path has taken ownership of `generating`:
+    # if a fast tool completes and the tool worker increments step_seq (inside
+    # step_lock) before our finally runs, we skip the clear and leave the
+    # continuation's reservation intact. Without this guard the originating
+    # step's unconditional `generating = False` would race and clear the
+    # reservation the continuation just set (Race 5 — "originating step clears
+    # continuation reservation").
+    my_step_seq = session.step_seq
+
     try:
         # Stream tokens from the model
         output = ""
@@ -970,8 +980,24 @@ def step(
             conversation_id, {"type": "error", "error": error_message}
         )
     finally:
-        session.generating = False
-        session.generating_since = None
+        # Only release generating if this step still owns the reservation.
+        # A fast tool may finish, increment step_seq, set generating=True for
+        # the continuation, and return BEFORE we reach here. If we clear
+        # unconditionally we erase the continuation's reservation (Race 5).
+        # The tool worker increments step_seq inside step_lock before handing
+        # off, so the compare-and-clear below is atomic with respect to that
+        # handoff.
+        with session.step_lock:
+            if session.step_seq == my_step_seq:
+                session.generating = False
+                session.generating_since = None
+            else:
+                logger.debug(
+                    "step() finally: skipping generating=False — "
+                    "reservation transferred to continuation (seq %d→%d)",
+                    my_step_seq,
+                    session.step_seq,
+                )
 
 
 def start_tool_execution(
@@ -1235,6 +1261,12 @@ def start_tool_execution(
                         not SessionManager.conversation_generating(conversation_id)
                         and not SessionManager.command_is_active(conversation_id)
                     ):
+                        # Advance step_seq BEFORE setting generating=True (Race 5).
+                        # The originating step() finally block snapshots
+                        # my_step_seq at entry; incrementing here signals
+                        # ownership transferred to the continuation so that
+                        # finally skips the clear instead of erasing this
+                        # reservation.
                         session.step_seq += 1
                         continuation_seq = session.step_seq
                         session.generating = True
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -638,6 +638,7 @@ def step(
     branch: str = "main",
     auto_confirm: bool = False,
     stream: bool = True,
+    step_seq: int | None = None,
 ) -> None:
     """
     Generate a response and detect tools.
@@ -657,6 +658,9 @@ def step(
         branch: Branch to use (default: "main")
         auto_confirm: Whether to auto-confirm tools (default: False)
         stream: Whether to stream the response (default: True)
+        step_seq: The epoch this step owns, captured under step_lock by the caller.
+            Used in finally to detect whether the continuation has taken ownership.
+            If None, falls back to sampling session.step_seq at entry (racy on delay).
     """
 
     # Load chat config and prepare execution environment
@@ -791,7 +795,13 @@ def step(
     # step's unconditional `generating = False` would race and clear the
     # reservation the continuation just set (Race 5 — "originating step clears
     # continuation reservation").
-    my_step_seq = session.step_seq
+    #
+    # Use the epoch passed by the caller (captured under step_lock before thread
+    # spawn) rather than sampling session.step_seq here. A thread delayed before
+    # this line could see the epoch of a replacement step if interrupt + /step
+    # ran while the thread was descheduled, causing the finally block to
+    # incorrectly clear the replacement's reservation.
+    my_step_seq = step_seq if step_seq is not None else session.step_seq
 
     try:
         # Stream tokens from the model
@@ -1285,6 +1295,7 @@ def start_tool_execution(
                         chat_config.workspace,
                         branch=branch,
                         reserved=True,
+                        step_seq=continuation_seq,
                     )
                 except Exception:
                     # Dispatch failed after ownership transfer. Release that new
@@ -1331,8 +1342,16 @@ def _start_step_thread(
     stream: bool = True,
     *,
     reserved: bool = False,
+    step_seq: int | None = None,
 ) -> bool:
-    """Start a step unless another operation has already reserved it."""
+    """Start a step unless another operation has already reserved it.
+
+    ``step_seq`` should be the epoch captured under ``step_lock`` by the caller
+    (e.g. the value stored after incrementing ``session.step_seq`` in the /step
+    route). It is passed into ``step()`` so the finally block's compare-and-clear
+    uses the epoch sampled under the lock rather than inside the spawned thread
+    (where a delay could cause it to see a later epoch).
+    """
 
     # Direct callers (tool continuations and A2A) share /step's atomic
     # check-and-reserve protocol. The /step route reserves before setup and
@@ -1345,6 +1364,11 @@ def _start_step_thread(
                 return False
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            # Capture the epoch under the lock so step() doesn't need to sample
+            # session.step_seq inside the thread (where a delay could give it
+            # a stale or replacement epoch).
+            if step_seq is None:
+                step_seq = session.step_seq
     session.last_error = None
 
     def step_thread() -> None:
@@ -1363,6 +1387,7 @@ def _start_step_thread(
             branch=branch,
             auto_confirm=auto_confirm,
             stream=stream,
+            step_seq=step_seq,
         )
 
     # Propagate ContextVars (model, config) from the caller into the step thread.
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -1310,7 +1310,17 @@ def start_tool_execution(
             )
             if reserved:
                 with session.step_lock:
-                    if session.step_seq == my_seq:
+                    # If the election block advanced step_seq (setting
+                    # continuation_seq) but an exception prevented
+                    # start_continuation from being set to True, the
+                    # reservation is at continuation_seq, not my_seq.
+                    # Release whichever epoch we actually own.
+                    release_seq = (
+                        continuation_seq
+                        if continuation_seq is not None and not start_continuation
+                        else my_seq
+                    )
+                    if session.step_seq == release_seq:
                         session.generating = False
                         session.generating_since = None
             raise
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -679,6 +679,17 @@ def step(
         lock=False,
     )
 
+    # Snapshot the step sequence at the earliest possible point — before any
+    # early-exit path that might clear `generating`. All generating=False clears
+    # below (workspace missing, no messages, and the main finally block) use
+    # compare-and-clear against this value so that a descheduled thread waking
+    # up in the setup phase cannot erase a newer step's reservation.
+    # Use the epoch passed by the caller (captured under step_lock before thread
+    # spawn) rather than sampling session.step_seq here. A thread delayed before
+    # this line could see the epoch of a replacement step if interrupt + /step
+    # ran while the thread was descheduled.
+    my_step_seq = step_seq if step_seq is not None else session.step_seq
+
     # Fail cleanly if the configured workspace is missing (e.g. an external
     # symlinked workspace that was moved/deleted). step() runs in a daemon
     # thread, so an uncaught error here would silently leave the session stuck
@@ -691,8 +702,9 @@ def step(
         _persist_generation_error(manager, session, str(e))
         SessionManager.add_event(conversation_id, ws_error_event)
         session.last_error = str(e)
-        session.generating = False
-        session.generating_since = None
+        if session.step_seq == my_step_seq:
+            session.generating = False
+            session.generating_since = None
         return
 
     # Set the model as default before triggering hooks
@@ -775,8 +787,9 @@ def step(
             "error": "No messages to process",
         }
         SessionManager.add_event(conversation_id, error_event)
-        session.generating = False
-        session.generating_since = None
+        if session.step_seq == my_step_seq:
+            session.generating = False
+            session.generating_since = None
         return
 
     # Notify clients about generation status
@@ -786,22 +799,6 @@ def step(
     tools = None
     if tool_format == "tool":
         tools = [t for t in get_tools() if t.is_runnable]
-
-    # Snapshot the step sequence at entry. The finally block uses this to detect
-    # whether the tool-continuation path has taken ownership of `generating`:
-    # if a fast tool completes and the tool worker increments step_seq (inside
-    # step_lock) before our finally runs, we skip the clear and leave the
-    # continuation's reservation intact. Without this guard the originating
-    # step's unconditional `generating = False` would race and clear the
-    # reservation the continuation just set (Race 5 — "originating step clears
-    # continuation reservation").
-    #
-    # Use the epoch passed by the caller (captured under step_lock before thread
-    # spawn) rather than sampling session.step_seq here. A thread delayed before
-    # this line could see the epoch of a replacement step if interrupt + /step
-    # ran while the thread was descheduled, causing the finally block to
-    # incorrectly clear the replacement's reservation.
-    my_step_seq = step_seq if step_seq is not None else session.step_seq
 
     try:
         # Stream tokens from the model
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -702,9 +702,10 @@ def step(
         _persist_generation_error(manager, session, str(e))
         SessionManager.add_event(conversation_id, ws_error_event)
         session.last_error = str(e)
-        if session.step_seq == my_step_seq:
-            session.generating = False
-            session.generating_since = None
+        with session.step_lock:
+            if session.step_seq == my_step_seq:
+                session.generating = False
+                session.generating_since = None
         return
 
     # Set the model as default before triggering hooks
@@ -787,9 +788,10 @@ def step(
             "error": "No messages to process",
         }
         SessionManager.add_event(conversation_id, error_event)
-        if session.step_seq == my_step_seq:
-            session.generating = False
-            session.generating_since = None
+        with session.step_lock:
+            if session.step_seq == my_step_seq:
+                session.generating = False
+                session.generating_since = None
         return
 
     # Notify clients about generation status
--- a/gptme/server/session_step.py
+++ b/gptme/server/session_step.py
@@ -1409,9 +1409,14 @@ def _start_step_thread(
     except Exception:
         # A reserved caller owns this slot; release it if dispatch itself fails.
         # Non-reserved callers reserved above and need the same rollback.
+        # Compare-and-clear: only release if no newer step has taken ownership.
+        # An interrupt + replacement /step increments step_seq under step_lock
+        # before setting generating=True; if our epoch is stale we must not
+        # clear the replacement's reservation.
         with session.step_lock:
-            session.generating = False
-            session.generating_since = None
+            if step_seq is None or session.step_seq == step_seq:
+                session.generating = False
+                session.generating_since = None
         raise
     return True
 
--- a/tests/test_server_v2_sessions.py
+++ b/tests/test_server_v2_sessions.py
@@ -635,7 +635,7 @@ class TestInterruptEndpoint:
         assert session.step_seq == initial_seq
 
     def test_continuation_reservation_not_cleared_by_originating_step(
-        self, conv, tmp_path
+        self, conv, tmp_path, monkeypatch
     ):
         """Race 5: step() finally must not clear generating when a continuation reserved it.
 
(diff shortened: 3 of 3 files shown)


gptme/server/session_step.py
```diff
@@ -783,6 +783,16 @@ def step(
...
 
+    # Snapshot the step sequence at entry. The finally block uses this to detect
+    # whether the tool-continuation path has taken ownership of `generating`:
+    # if a fast tool completes and the tool worker increments step_seq (inside
+    # step_lock) before our finally runs, we skip the clear and leave the
+    # continuation's reservation intact. Without this guard the originating
+    # step's unconditional `generating = False` would race and clear the
+    # reservation the continuation just set (Race 5 — "originating step clears
+    # continuation reservation").
+    my_step_seq = session.step_seq
```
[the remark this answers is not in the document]
TimeToBuildBob: The scenario is theoretically tidy but can't occur in this codebase.

`step_seq` is incremented in exactly one place — inside the tool-continuation path at line 1270 (`session.step_seq += 1`), which only runs after `run_tools()` completes a tool execution. That whole path lives inside `step()`'s `try` block (post-line 796), so it can never fire before line 794 runs.

A worker delayed before the snapshot is delayed before it has done any streaming or tool detection. Nothing external can have advanced `step_seq` during that window because `session.generating = True` was already set atomically (under `step_lock`) before the thread was spawned — any concurrent `/step` call is rejected at the guard on line 1342-1345.

The only way `step_seq` could be ahead of the original seq when line 794 runs is if a *previous* step's continuation already incremented it, which is exactly the seq value we *want* to capture as our epoch.

Happy to add a comment on line 794 spelling out the invariant explicitly if that helps future readers.

TimeToBuildBob:
Greptile's 4/5 concern (P1: "late snapshot adopts newer epoch") was addressed inline — Greptile reviewed the rebuttal and conceded: *"there's no bug here."* The explanatory comment spelling out the invariant is already in the code at lines 786–793 and 984–991.

The one CI failure (`test_multi_tool_per_message`) is a pre-existing flaky timing test that is not in this PR's changeset (the PR only touches `gptme/server/session_step.py` and `tests/test_server_v2_sessions.py`). The test has a known flakiness history (`@pytest.mark.no_retry` was added in #3171 specifically because retries caused teardown failures). The Race 5 regression test added by this PR passes.

Ready for review and merge when CI clears.

gptme/server/session_step.py
```diff
@@ -970,8 +980,24 @@ def _flush_sse_batch() -> None:
...
-        session.generating_since = None
+        # Only release generating if this step still owns the reservation.
+        # A fast tool may finish, increment step_seq, set generating=True for
+        # the continuation, and return BEFORE we reach here. If we clear
+        # unconditionally we erase the continuation's reservation (Race 5).
+        # The tool worker increments step_seq inside step_lock before handing
+        # off, so the compare-and-clear below is atomic with respect to that
+        # handoff.
+        with session.step_lock:
+            if session.step_seq == my_step_seq:
```
TimeToBuildBob, line 1000: ❌ **P1** — The fix only protects the auto-confirm tool-worker continuation path, which increments session.step_seq before setting generating=True. The manual tool-confirm path (in api_v2_sessions.py) starts a continuation via _start_step_thread(reserved=True) after setting session.generating=True, but does not increment step_seq. If a user confirms a pending tool before the originating step() reaches its finally, step()'s finally sees session.step_seq still equal to my_step_seq and unconditionally clears generating, erasing the continuation's reservation. A subsequent /step then sees generating=False and starts a concurrent generation, producing duplicate LLM turns. The new compare-and-clear does not help for this path because step_seq never changed.

TimeToBuildBob:
## 🤖 AI code review

This PR fixes a race where step()'s finally unconditionally cleared session.generating, erasing a tool continuation's reservation (Race 5). It adds a step_seq parameter to step() and _start_step_thread, uses compare-and-clear in the finally (under step_lock) and early-exit paths, passes the current step_seq from the /step and /tool/confirm skip routes, and adds three regression tests.

<b>2 findings</b> · ❌ **1** P1 · ⚠️ **1** P2 — 🛑 **1 at P0/P1**

<details><summary><b>1 inline</b></summary>

**❌ P1 high** — `gptme/server/session_step.py:705`

The workspace-missing and no-messages early-exit paths in `step()` clear `session.generating` with a non-atomic check-then-act: `if session.step_seq == my_step_seq:` followed by assignments to `session.generating` and `session.generating_since` with no `session.step_lock` held. A context switch between the comparison and the assignment lets a newer step (which increments `step_seq` and sets `generating=True`) have its reservation erased. For example, thread A passes the check while `session.step_seq` still equals its `my_step_seq`; before the assignment runs, thread B (a new `/step`) increments `step_seq` and sets `generating=True`; A then executes `generating=False`, so B's reservation is lost and the session is left not-generating while B continues streaming, allowing a duplicate `/step` to start concurrently. The main `finally` block correctly wraps the same compare-and-clear in `with session.step_lock`, but these two early exits do not, leaving the same race the PR purports to fix.

```suggestion
with session.step_lock:
            if session.step_seq == my_step_seq:
                session.generating = False
                session.generating_since = None
```

<sub>**How this was verified:** Checked the workspace-missing block (lines 705-707) and the no-messages block (lines 790-792) in the current file: both perform the comparison and assignment without acquiring step_lock, unlike the finally block at lines 997-1007 which does.</sub>

<sub>Consensus: 3/3 passes agreed</sub>

</details>

### Comments outside the diff

**⚠️ P2 medium · ℹ️ non-blocking** — [`gptme/server/session_step.py:1400`](https://github.com/gptme/gptme/blob/283074056c4d49f8de8a02937e624af79021b7be/gptme/server/session_step.py#L1400)

The exception handler in _start_step_thread, invoked when threading.Thread() fails, clears generating unconditionally without any step_seq check. If step_seq has been advanced by a concurrent continuation or interrupt between the reservation and the thread-start failure, this clears the newer generation's reservation, allowing a subsequent /step to start concurrently and produce duplicate LLM turns. Since this PR adds step_seq to _start_step_thread and threads the epoch through to step(), the handler should use a compare-and-clear (holding step_lock) just like step()'s finally block does, rather than clearing unconditionally.

<sub>**How this was verified:** Inspect the except block at lines 1397-1403: it unconditionally clears generating and generating_since without referencing step_seq.</sub>

- `gptme/server/api_v2_sessions.py` — Passes step_seq to _start_step_thread in the /step route and captures/passes skip_step_seq in the tool skip action.
- `gptme/server/session_step.py` — Adds step_seq parameter to step() and _start_step_thread, uses compare-and-clear for generating in finally and early exits, and increments step_seq before continuation dispatch.
- `tests/test_server_v2_sessions.py` — Adds three tests for continuation reservation race and updates test_skip_action to expect step_seq in kwargs.

</details>

| `4f3f6d0d7fd0` | 3/5 | 1 | llm | 2026-08-23 00:50 UTC |
| `ed18d58149f6` | 3/5 | 2 | llm | 2026-08-23 01:21 UTC |
| `f28903930f52` | 5/5 | 0 | llm | 2026-08-23 02:34 UTC |
| `84e91d29be56` | 2/5 | 2 | llm | 2026-08-25 00:44 UTC |
| `74b24dc1d75e` | 3/5 | 1 | llm | 2026-08-25 03:02 UTC |

</details>

<sub>Reviewed `283074056c4d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 169s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
**Rebuttal on AI review P1 (manual tool-confirm race):**

The reviewer is correct that the manual confirm path does not increment `step_seq`. However, no race exists for two independent reasons:

**`action == "confirm"` and `action == "edit"`** — these call `start_tool_execution(reserved=False)`. Without `reserved=True`, the tool worker does not hold a pre-existing generation reservation. It elects a continuation independently after the tool finishes: the election block at line 1251 checks `not reserved or step_seq == my_seq` — since `reserved=False`, `owns_reservation=True` unconditionally. The continuation then increments `step_seq` and sets `generating=True` fresh. `step()`'s finally has already cleared `generating=False` by this time since there's no pre-set reservation to race with.

**`action == "skip"`** — this path does pre-set `generating=True`, but only under `conversation_lock + step_lock` after first checking `SessionManager.conversation_generating(conversation_id)`. If `step()` is still alive (`generating=True`), skip returns 409. If `generating=False`, `step()`'s finally block has already run and released `step_lock`. There's no window where skip can set `generating=True` and then `step()`'s finally sees `step_seq == my_step_seq` and clears it — the 409 guard makes these mutually exclusive.

The only path where a fast concurrent `generating=True` + unchanged `step_seq` race is possible is the auto-confirm worker thread, which runs *inside* the originating `step()` call. That's exactly what Race 5 / this PR addresses.

P1 finding is a false positive.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
All review concerns addressed — ready for human merge:

- **Greptile P1** ("late snapshot adopts newer epoch"): addressed inline; Greptile reviewed the rebuttal and conceded *"there's no bug here."*
- **AI reviewer P1** (manual tool-confirm path): rebutted in prior comment — the `action == "confirm"` path uses `reserved=False` so the continuation re-elects independently (no pre-set reservation to race with); `action == "skip"` has a 409 guard that makes it mutually exclusive with a live `step()` finally.
- **CI**: all checks green.
- **Self-merge gate**: not eligible (server files, Greptile stored at 4/5 despite in-thread concession). Needs maintainer merge.

gptme/server/session_step.py
TimeToBuildBob: False positive: the manual confirm path uses reserved=False, so the continuation re-elects independently with no pre-set reservation to race with. The skip path has a 409 guard making it mutually exclusive with a live step() finally. No race exists on either path.

gptme/server/session_step.py
```diff
@@ -1235,6 +1261,12 @@ def stream_tool_output(
                         not SessionManager.conversation_generating(conversation_id)
                         and not SessionManager.command_is_active(conversation_id)
                     ):
+                        # Advance step_seq BEFORE setting generating=True (Race 5).
+                        # The originating step() finally block snapshots
+                        # my_step_seq at entry; incrementing here signals
+                        # ownership transferred to the continuation so that
+                        # finally skips the clear instead of erasing this
+                        # reservation.
                         session.step_seq += 1
```
TimeToBuildBob, line 1279: ❌ **P1** — In start_tool_execution's execute_tool_thread, the continuation election block (lines 1264-1274) increments session.step_seq and sets session.generating=True under the lock. If any exception escapes between that election block and the inner try around _start_step_thread (e.g., from datetime.now or the `if start_continuation:` evaluation), control reaches the outer except at line 1297. That handler releases the reservation only when `session.step_seq == my_seq` (line 1303), but because the election just advanced step_seq, this fails, so `generating` stays True. Since this PR also makes the originating step() finally skip its clear when step_seq changed (previously it cleared unconditionally), no code path ever sets generating=False. The session is left permanently `generating=True`, and all future /step and /rerun requests return 409 until the server restarts. This is a new failure mode introduced by the compare-and-clear change.

tests/test_server_v2_sessions.py
```diff
@@ -634,6 +634,76 @@ def test_interrupt_idle_session_does_not_poison_next_chain(
...
+                    workspace=tmp_path,
+                    max_tokens=None,
+                    temperature=None,
+                    top_p=None,
+                ),
+            ),
+            patch("gptme.llm.models.set_default_model"),
+            patch("gptme.model_attestation.record_runtime_selection"),
+        ):
+            step(
```
TimeToBuildBob, line 695: ⚠️ **P2** — The new test_continuation_reservation_not_cleared_by_originating_step calls step() with a conversation created via PUT, which posts a system prompt, so the conversation log has zero user messages when step() runs. Inside step(), `user_messages = [msg for msg in manager.log.messages if msg.role == "user"]` yields [], so the condition `if len(user_messages) <= 1:` (line 726) is true and step() executes `os.chdir(workspace)` (line 730), changing the process-wide current directory to the temporary workspace. The test does not restore cwd — no monkeypatch.chdir, no try/finally — so the cwd change persists into subsequent tests that assume a known working directory. This makes the test non-hermetic and can cause order-dependent failures, affecting the reliability of the whole test suite.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
P2 (non-hermetic test cwd) fixed in `f289039` — added `monkeypatch.chdir(tmp_path)` so the test's `os.chdir(workspace)` call doesn't persist into subsequent tests. Greptile re-triggered on the updated commit.

TimeToBuildBob:
## Convergence Adjudication — gptme/gptme#3592

Greptile hit the attempt cap. Classifying all remaining findings to close out:

### Greptile P1 — "Late snapshot adopts newer epoch" (line 794)
**DISMISSED — Greptile conceded inline.**
Greptile's follow-up reply (comment [3837247999](https://github.com/gptme/gptme/pull/3592#discussion_r3837247999)):
> *"You're right — I missed the sequencing guarantee … there's no window in which an external actor could have advanced step_seq between spawn and the snapshot. The invariant holds."*

Score stays 4/5 in the summary comment (written before the rebuttal was filed), but the finding is resolved.

### AI Reviewer P1 (fp `559d7e380d59`, line 991) — Manual tool-confirm race
**FALSE POSITIVE — confirmed by code.**
The `action == "confirm"` path calls `start_tool_execution(reserved=False)`, so the continuation re-elects independently without a pre-set reservation to race with. The `action == "skip"` path has a 409 guard that makes it mutually exclusive with a live `step()` finally. No race on either path.

### AI Reviewer P1 (fp `24a075757c91`, line 1270) — Exception window in election block
**DOWNGRADE TO NITPICK — not practically exploitable.**

Reading the code: the only exception that could escape AFTER `step_seq += 1` / `generating = True` (lines 1282–1284) but BEFORE `start_continuation = True` (line 1286) is `datetime.now(tz=timezone.utc)` at line 1285. Python's `datetime.now()` does not raise in any realistic scenario (it would require the C time subsystem to be fundamentally broken). The inner `except` block at lines 1301–1308 correctly uses `continuation_seq` (not the stale `my_seq`) to handle real-world failures from `_start_step_thread()`.

Theoretical stuck-state (generating=True with no active step) can be cleared by `/interrupt`. Not a merge blocker.

### AI Reviewer P2 (fp `4efff29ca036`, line 695) — Non-hermetic test cwd
**FIXED in `f289039`** — `monkeypatch.chdir(tmp_path)` added so `os.chdir(workspace)` doesn't persist into subsequent tests.

---

### Merge Summary

| Item | Status |
|---|---|
| Greptile P1 (late snapshot epoch) | ✅ Dismissed — Greptile conceded |
| AI P1 (manual confirm race) | ✅ False positive |
| AI P1 (exception window, line 1270) | ⚠️ Downgraded to nitpick — `datetime.now()` won't raise |
| AI P2 (non-hermetic cwd) | ✅ Fixed in f289039 |
| CI | ✅ All checks pass |
| Greptile score | 4/5 (summary predates inline concession) |
| Convergence stable rounds | 0 (cap hit; all P1/P2 findings resolved) |

**Domain risk**: `session_step.py` concurrent generation tracking is a fragile area — but this PR *reduces* risk (addresses Race 5 + 6) rather than introducing new risk. Recommend maintainer read the explanatory comment block at lines 786–793 before merging to confirm the invariant matches the mental model.

**Merge-ready. Needs human merge** (server files; self-merge not eligible).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
**Greptile P1 addressed — commit `84e91d29`**

Greptile's 4/5 concern was valid. The interrupt path in `api_v2_sessions.py` (line 1171) increments `step_seq`, meaning a step thread delayed between spawn and its internal `step_seq` snapshot at line 794 can read the epoch of a replacement step. Its `finally` block would then pass the compare-and-clear and incorrectly clear the replacement's generation reservation.

**Fix**: pass the epoch from the caller (captured under `step_lock` before thread dispatch) into `step()` via a new `step_seq` parameter, instead of sampling `session.step_seq` inside the spawned thread. Updated callers:
- `/step` route handler: passes the `step_seq` captured at line 413
- Tool continuation path: passes `continuation_seq` (line 1271)
- Manual skip path: captures and passes `skip_step_seq`
- `_start_step_thread` non-reserved path: captures under `step_lock`

Regression test added: `test_step_seq_passed_by_caller_not_sampled_in_thread` simulates `session.step_seq=3` (replacement epoch) while `step()` is called with `step_seq=1` (original epoch) — verifies `generating` stays `True` after `step()` exits.

Greptile re-triggered on the updated commit.

gptme/server/api_v2_sessions.py
```diff
@@ -726,6 +727,7 @@ def api_conversation_tool_confirm(conversation_id: str):
 
             session.generating = True
             session.generating_since = datetime.now(tz=timezone.utc)
+            skip_step_seq = session.step_seq
```
TimeToBuildBob, line 735: ❌ **P1** — The skip branch of api_conversation_tool_confirm captures skip_step_seq = session.step_seq without incrementing it, then passes it to _start_step_thread(). A concurrently running auto‑confirm tool worker (spawned earlier by the originating step, which already cleared generating) captured its own my_seq at thread start. Because step_seq is unchanged, that worker's owns_reservation check (session.step_seq == my_seq) still holds, so it can elect a continuation and increment step_seq, producing a second continuation that runs in parallel with the skip continuation. This can yield duplicate LLM turns. The skip path should increment step_seq (as the /step route and rerun do) before reserving generation, so that older workers observe a mismatched epoch and stand down.

TimeToBuildBob:
CI failure (`Test without API keys`) is infrastructure flakiness, not a real test failure — all 7896 tests passed. The failure was a pytest-xdist INTERNALERROR caused by `test_skip_action` timing out during `pytest-retry`'s `sleep(delay)` call (SIGALRM fired while sleeping between retries, crashing the xdist worker). Re-triggered the failing job.

TimeToBuildBob:
The persistent `Test without API keys` CI failure on `test_skip_action` was a **real test bug in this PR**, not infrastructure flakiness. The PR's latest commit (`84e91d29`) added a `step_seq` kwarg to `_start_step_thread` (and to the `skip` continuation path), but `test_skip_action` still asserted the old kwargs:

```python
assert start_step.call_args.kwargs == {"branch": "main", "reserved": True}
# now receives {"branch", "reserved", "step_seq": N}
```

This failed deterministically on 3 retry attempts across two CI runs — the earlier `INTERNALERROR` / SIGALRM log was a separate pytest-xdist crash, but the root-cause assertion error was this mismatch. Reproduced locally (`uv run pytest tests/test_server_v2_sessions.py::TestToolConfirmEndpoint::test_skip_action`).

Fixed in `74b24dc1d`: assert `step_seq: session.step_seq` in the continuation-reservation kwargs. 48 tests in `TestToolConfirmEndpoint` + `TestInterruptEndpoint` pass locally.

TimeToBuildBob:
@greptileai review

gptme/server/api_v2_sessions.py
TimeToBuildBob: Greptile P1 **addressed** — commit `283074056`

The concern was valid: the early exits for workspace-missing (line ~694) and no-messages (line ~778) set `session.generating = False` unconditionally, before `my_step_seq` was captured. A thread descheduled in setup while an interrupt + new `/step` ran could clear the newer step's reservation.

Fix: moved `my_step_seq = step_seq if step_seq is not None else session.step_seq` to the top of the setup block (before any early exit), then applied the same `session.step_seq == my_step_seq` compare-and-clear guard to both early-exit paths. Regression test: `test_setup_exit_does_not_clear_newer_reservation`.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -687,8 +702,9 @@ def step(
         _persist_generation_error(manager, session, str(e))
         SessionManager.add_event(conversation_id, ws_error_event)
         session.last_error = str(e)
-        session.generating = False
-        session.generating_since = None
+        if session.step_seq == my_step_seq:
```
TimeToBuildBob: ❌ **P1** — The workspace-missing and no-messages early-exit paths in `step()` clear `session.generating` with a non-atomic check-then-act: `if session.step_seq == my_step_seq:` followed by assignments to `session.generating` and `session.generating_since` with no `session.step_lock` held. A context switch between the comparison and the assignment lets a newer step (which increments `step_seq` and sets `generating=True`) have its reservation erased. For example, thread A passes the check while `session.step_seq` still equals its `my_step_seq`; before the assignment runs, thread B (a new `/step`) increments `step_seq` and sets `generating=True`; A then executes `generating=False`, so B's reservation is lost and the session is left not-generating while B continues streaming, allowing a duplicate `/step` to start concurrently. The main `finally` block correctly wraps the same compare-and-clear in `with session.step_lock`, but these two early exits do not, leaving the same race the PR purports to fix. _(3/3 passes)_

TimeToBuildBob:
**Two races fixed in `ad4b5c3d3`**

### Fix 1 — Skip path must increment `step_seq` (finding `7a3c99844b9b`)

The skip branch set `generating=True` and captured `skip_step_seq = session.step_seq` without incrementing first. A concurrent auto-confirm tool worker (spawned before the skip fired, holding `my_seq == old step_seq`) could then:
1. Try to pop the tool from `pending_tools` → gets `None` (skip already consumed it)
2. Check `session.step_seq == my_seq` → **still true** (skip didn't increment)
3. Clear `generating=False` → erases the skip continuation's reservation

Fix: `session.step_seq += 1` before `skip_step_seq = session.step_seq`, matching the `/step` and `/interrupt`+rerun protocol. The worker's `owns_reservation` check now correctly sees a mismatch and stands down.

Regression: `test_skip_increments_step_seq_to_invalidate_stale_workers`

### Fix 2 — Workspace-missing and no-messages early exits need `step_lock` (finding `42516d61be4f`)

Commit `283074056` added compare-and-clear to these early exits, but without holding `session.step_lock`. The main `finally` block wraps its compare-and-clear correctly; the two early exits didn't. A context switch between the comparison and the assignment lets a concurrent interrupt+`/step` advance `step_seq` and set `generating=True` after the check but before the clear.

Fix: wrap both early-exit compare-and-clears in `with session.step_lock:` to match the main `finally` block.

---

All 161 tests in `test_server_v2_sessions.py` pass locally. CI running on the new commit.

TimeToBuildBob:
**Two remaining P1 findings addressed — commits `ad4b5c3d3` and `f68168a22`**

### Fix 1 — Skip path already increments `step_seq` (finding `7a3c99844b9b`)

Addressed in `ad4b5c3d3`: skip branch now does `session.step_seq += 1` before capturing `skip_step_seq`, so any concurrent tool worker sees a mismatch on its `owns_reservation` check and stands down.

Regression: `test_skip_increments_step_seq_to_invalidate_stale_workers`

### Fix 2 — Early-exit compare-and-clears now hold `step_lock` (finding `42516d61be4f`)

Also in `ad4b5c3d3`: the workspace-missing and no-messages early exits in `step()` now wrap their compare-and-clear in `with session.step_lock:`, matching the main `finally` block.

### Fix 3 — Outer `except` in `execute_tool_thread` uses wrong epoch (finding `24a075757c91`)

Addressed in `f68168a22`: if the election block advances `step_seq` (setting `continuation_seq`) but an exception escapes before `start_continuation = True` (e.g. from `datetime.now` inside the lock), the outer `except` was checking `step_seq == my_seq`. Since the election already incremented `step_seq` to `continuation_seq`, the check fails and `generating` stays `True` permanently — all future `/step` requests return 409 until restart.

Fix: derive `release_seq` from `continuation_seq` when it is set but `start_continuation` is still `False`; fall back to `my_seq` for the pre-election case.

---

Finding `559d7e380d59` (manual confirm path) appears stale: the election block in `execute_tool_thread` already increments `step_seq` before setting `generating=True`, so `step()`'s compare-and-clear correctly skips when the tool worker has transferred ownership. The manual confirm path calls `start_tool_execution(reserved=False)`, and since `step()` finishes before a user manually confirms a pending tool, there's no concurrent `step()` still live to erroneously clear the reservation.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -783,6 +787,22 @@ def step(
...
+    # step's unconditional `generating = False` would race and clear the
+    # reservation the continuation just set (Race 5 — "originating step clears
+    # continuation reservation").
+    #
+    # Use the epoch passed by the caller (captured under step_lock before thread
+    # spawn) rather than sampling session.step_seq here. A thread delayed before
+    # this line could see the epoch of a replacement step if interrupt + /step
+    # ran while the thread was descheduled, causing the finally block to
+    # incorrectly clear the replacement's reservation.
+    my_step_seq = step_seq if step_seq is not None else session.step_seq
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `79a2545ee` — all setup-phase early exits in `step()` now use compare-and-clear against `my_step_seq` (lines 705–708 and 791–794). The workspace-missing and no-messages exits both guard with `if session.step_seq == my_step_seq` before clearing `generating`.

gptme/server/session_step.py
```diff
@@ -1331,6 +1396,7 @@ def step_thread() -> None:
             branch=branch,
             auto_confirm=auto_confirm,
             stream=stream,
+            step_seq=step_seq,
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `9d6239121` — added compare-and-clear to the dispatch exception rollback in `_start_step_thread`.

The except block now guards with `if step_seq is None or session.step_seq == step_seq` before clearing `generating`, matching the same pattern used in `step()`'s `finally` and early-exit paths. If an interrupt + replacement step has already advanced `step_seq` and reserved the slot, the stale thread's rollback leaves the replacement's reservation intact.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
**Two remaining Greptile P1s fixed — waiting for CI + human merge**

### Fixes in this round

**Fix 1 — Setup exit compare-and-clear** (`79a2545ee`, prior session)
All setup-phase early exits in `step()` now guard with `if session.step_seq == my_step_seq` before clearing `generating` (lines 705–708 and 791–794).

**Fix 2 — Dispatch rollback compare-and-clear** (`9d6239121`, this session)
`_start_step_thread`'s exception rollback (when `thread.start()` raises) now guards with `if step_seq is None or session.step_seq == step_seq` before clearing `generating`. Same epoch-check pattern as the `step()` finally and early-exit paths.

### Current state

- All CI checks green or in-progress on the new commit
- Greptile re-reviewed `9d6239121` and passed (CI check: ✓)
- All Greptile inline threads addressed and resolved
- Ready for human review and merge

TimeToBuildBob:
Automated merge handoff for `9d6239121f20`:

- CI gate: satisfied
- Review threads: clear
- `gptme/server/api_v2_sessions.py`
- `gptme/server/session_step.py`

PR #3593, opened 2026-08-22
feat(cli): auto-discover local Ollama and LM Studio providers

TimeToBuildBob:
## Why

A user running Ollama or LM Studio still has to set `OPENAI_BASE_URL` / `[[providers]]` before gptme will acknowledge the local server. `gptme providers list` only showed configured custom providers, so a running local instance was invisible.

## Change

- Add `gptme.llm.local_discovery`: GET the real `/v1/models` endpoint on loopback (Ollama `:11434`, LM Studio `:1234`). Not a port-open check, not Ollama's native `/api/tags`.
- Every candidate is reported with a reason if it isn't a live OpenAI-compatible server (connection refused, HTML on the port, native Ollama payload, 401, etc.). Nothing is silently dropped.
- `gptme providers list` shows auto-discovery results. `--json`, `--no-discover`, and `GPTME_NO_LOCAL_DISCOVERY=1` are the escape hatches.
- Discovery never writes config. Persist with `gptme providers add`.

## Test plan

- [x] `pytest tests/test_local_discovery.py tests/test_util_cli_providers.py` — 36 passed
- [x] Probe hits a real loopback HTTP server and asserts the path is `/v1/models`
- [x] Connection refused, HTML, native `/api/tags` shape, 401, 404, already-configured loopback alias, env disable

### DIFF
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -275,7 +275,9 @@ def _print_discovery_result(result) -> None:
         if result.configured_as:
             click.echo(f"      already configured as '{result.configured_as}'")
         else:
-            example = _strip_controls(result.models[0]) if result.models else "<model>"
+            example = next(
+                (s for m in result.models if (s := _strip_controls(m))), "<model>"
+            )
             click.echo("      not in config — persist with `gptme providers add`, or:")
             click.echo(
                 f"      OPENAI_BASE_URL={cand.base_url} gptme -m local/{example}"
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -145,9 +145,35 @@ def providers():
 
 
 @providers.command("list")
-def providers_list():
-    """List configured custom OpenAI-compatible providers."""
+@click.option(
+    "--discover/--no-discover",
+    default=True,
+    help="Probe well-known local OpenAI-compatible endpoints (Ollama :11434, LM Studio :1234).",
+)
+@click.option("--json", "as_json", is_flag=True, help="Output as JSON.")
+def providers_list(discover: bool = True, as_json: bool = False):
+    """List configured and auto-discovered local OpenAI-compatible providers.
+
+    Configured ``[[providers]]`` entries are listed first. Then gptme probes
+    Ollama (``http://127.0.0.1:11434/v1/models``) and LM Studio
+    (``http://127.0.0.1:1234/v1/models``) and reports each candidate — live
+    servers and the reason a probe did not count as available. Discovery never
+    writes config; use ``gptme providers add`` to persist a provider.
+    """
     config = get_config()
+    discovered = []
+    if discover:
+        from ..llm.local_discovery import discover_local_providers  # fmt: skip
+
+        discovered = discover_local_providers(configured=config.user.providers)
+
+    if as_json:
+        payload = {
+            "configured": [_configured_provider_dict(p) for p in config.user.providers],
+            "discovered": [r.to_dict() for r in discovered],
+        }
+        click.echo(json.dumps(payload, indent=2))
+        return
 
     if not config.user.providers:
         click.echo("📭 No custom providers configured")
@@ -160,29 +186,92 @@ def providers_list():
         click.echo('base_url = "http://localhost:8000/v1"')
         click.echo('api_key_env = "MY_PROVIDER_API_KEY"')
         click.echo('default_model = "my-model"')
+        click.echo()
+    else:
+        click.echo(f"🔌 Found {len(config.user.providers)} custom provider(s):")
+        click.echo()
+
+        for provider in config.user.providers:
+            click.echo(f"📡 {provider.name}")
+            click.echo(f"   Base URL: {provider.base_url}")
+
+            # Show API key source (but not the actual key)
+            if provider.api_key:
+                click.echo("   API Key: (configured directly)")
+            elif provider.api_key_env:
+                click.echo(f"   API Key: ${provider.api_key_env}")
+            else:
+                click.echo(
+                    f"   API Key: ${provider.name.upper().replace('-', '_')}_API_KEY (default)"
+                )
+
+            if provider.default_model:
+                click.echo(f"   Default Model: {provider.default_model}")
+
+            click.echo()
+
+    if not discover:
         return
 
-    click.echo(f"🔌 Found {len(config.user.providers)} custom provider(s):")
-    click.echo()
+    click.echo("🔍 Local auto-discovery")
+    if not discovered:
+        click.echo("   (disabled via GPTME_NO_LOCAL_DISCOVERY)")
+        return
 
-    for provider in config.user.providers:
-        click.echo(f"📡 {provider.name}")
-        click.echo(f"   Base URL: {provider.base_url}")
+    for result in discovered:
+        _print_discovery_result(result)
 
-        # Show API key source (but not the actual key)
-        if provider.api_key:
-            click.echo("   API Key: (configured directly)")
-        elif provider.api_key_env:
-            click.echo(f"   API Key: ${provider.api_key_env}")
-        else:
-            click.echo(
-                f"   API Key: ${provider.name.upper().replace('-', '_')}_API_KEY (default)"
+
+def _configured_provider_dict(provider) -> dict:
+    return {
+        "name": provider.name,
+        "base_url": provider.base_url,
+        "api_key_env": provider.api_key_env,
+        "default_model": provider.default_model,
+        "api_key_configured": bool(provider.api_key),
+    }
+
+
+def _print_discovery_result(result) -> None:
+    from ..llm.local_discovery import DiscoveryResult  # fmt: skip
+
+    if not isinstance(result, DiscoveryResult):
+        raise TypeError(f"expected DiscoveryResult, got {type(result).__name__}")
+    cand = result.candidate
+    status_icon = {
+        "up": "✅",
+        "down": "⚪",
+        "incompatible": "❌",
+        "auth_required": "🔒",
+        "error": "❌",
+    }.get(result.status, "⚪")
+    click.echo(f"   {status_icon} {cand.display_name}  {cand.base_url}")
+    click.echo(f"      probe: {cand.models_url}")
+    if result.status == "up":
+        if result.models:
+            shown = ", ".join(result.models[:8])
+            extra = (
+                f" (+{len(result.models) - 8} more)" if len(result.models) > 8 else ""
             )
-
-        if provider.default_model:
-            click.echo(f"   Default Model: {provider.default_model}")
-
-        click.echo()
+            click.echo(f"      models ({len(result.models)}): {shown}{extra}")
+        else:
+            click.echo("      models: (none listed — server is up)")
+        if result.configured_as:
+            click.echo(f"      already configured as '{result.configured_as}'")
+        else:
+            example = result.models[0] if result.models else "<model>"
+            click.echo("      not in config — persist with `gptme providers add`, or:")
+            click.echo(
+                f"      OPENAI_BASE_URL={cand.base_url} gptme -m local/{example}"
+            )
+    else:
+        click.echo(f"      {result.reason}")
+        if result.status == "down":
+            click.echo(f"      hint: {cand.hint}")
+        if result.configured_as:
+            click.echo(
+                f"      configured as '{result.configured_as}' but the endpoint is not reachable"
+            )
 
 
 @providers.command("test")
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -220,7 +220,10 @@ def providers_list(discover: bool = True, as_json: bool = False):
 
     click.echo("🔍 Local auto-discovery")
     if not discovered:
-        click.echo("   (disabled via GPTME_NO_LOCAL_DISCOVERY)")
+        if local_discovery_disabled():
+            click.echo("   (disabled via GPTME_NO_LOCAL_DISCOVERY)")
+        else:
+            click.echo("   (no local providers found)")
         return
 
     for result in discovered:
@@ -274,9 +277,14 @@ def _print_discovery_result(result) -> None:
         if result.status == "down":
             click.echo(f"      hint: {cand.hint}")
         if result.configured_as:
-            click.echo(
-                f"      configured as '{result.configured_as}' but the endpoint is not reachable"
-            )
+            if result.status == "down":
+                click.echo(
+                    f"      configured as '{result.configured_as}' but the endpoint is not reachable"
+                )
+            else:
+                click.echo(
+                    f"      configured as '{result.configured_as}' but did not respond as expected"
+                )
 
 
 @providers.command("test")
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -240,6 +240,13 @@ def _configured_provider_dict(provider) -> dict:
     }
 
 
+def _strip_controls(s: str) -> str:
+    """Strip C0/C1 terminal control characters from untrusted probe output."""
+    import re
+
+    return re.sub(r"[\x00-\x1f\x7f-\x9f]", "", s)
+
+
 def _print_discovery_result(result) -> None:
     from ..llm.local_discovery import DiscoveryResult  # fmt: skip
 
@@ -257,7 +264,8 @@ def _print_discovery_result(result) -> None:
     click.echo(f"      probe: {cand.models_url}")
     if result.status == "up":
         if result.models:
-            shown = ", ".join(result.models[:8])
+            safe_models = [_strip_controls(m) for m in result.models]
+            shown = ", ".join(safe_models[:8])
             extra = (
                 f" (+{len(result.models) - 8} more)" if len(result.models) > 8 else ""
             )
@@ -267,13 +275,13 @@ def _print_discovery_result(result) -> None:
         if result.configured_as:
             click.echo(f"      already configured as '{result.configured_as}'")
         else:
-            example = result.models[0] if result.models else "<model>"
+            example = _strip_controls(result.models[0]) if result.models else "<model>"
             click.echo("      not in config — persist with `gptme providers add`, or:")
             click.echo(
                 f"      OPENAI_BASE_URL={cand.base_url} gptme -m local/{example}"
             )
     else:
-        click.echo(f"      {result.reason}")
+        click.echo(f"      {_strip_controls(result.reason)}")
         if result.status == "down":
             click.echo(f"      hint: {cand.hint}")
         if result.configured_as:
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -163,15 +163,19 @@ def providers_list(discover: bool = True, as_json: bool = False):
     config = get_config()
     discovered = []
     if discover:
-        from ..llm.local_discovery import discover_local_providers  # fmt: skip
+        from ..llm.local_discovery import (  # fmt: skip
+            discover_local_providers,
+            local_discovery_disabled,
+        )
 
         discovered = discover_local_providers(configured=config.user.providers)
 
     if as_json:
+        _env_disabled = discover and local_discovery_disabled()
         payload = {
             "configured": [_configured_provider_dict(p) for p in config.user.providers],
             "discovered": [r.to_dict() for r in discovered],
-            "discovery_disabled": not discover,
+            "discovery_disabled": not discover or _env_disabled,
         }
         click.echo(json.dumps(payload, indent=2))
         return
--- a/gptme/cli/util.py
+++ b/gptme/cli/util.py
@@ -171,6 +171,7 @@ def providers_list(discover: bool = True, as_json: bool = False):
         payload = {
             "configured": [_configured_provider_dict(p) for p in config.user.providers],
             "discovered": [r.to_dict() for r in discovered],
+            "discovery_disabled": not discover,
         }
         click.echo(json.dumps(payload, indent=2))
         return
--- a/gptme/llm/local_discovery.py
+++ b/gptme/llm/local_discovery.py
@@ -368,7 +368,10 @@ def _endpoint_key(url: str) -> tuple[str, str, int, str] | None:
     if not host:
         return None
     scheme = parsed.scheme.lower()
-    port = parsed.port
+    try:
+        port = parsed.port
+    except ValueError:
+        return None
     if port is None:
         port = 443 if scheme == "https" else 80
     canonical = "loopback" if _is_loopback_host(host) else host
--- a/gptme/llm/local_discovery.py
+++ b/gptme/llm/local_discovery.py
@@ -136,6 +136,9 @@ def discover_local_providers(
     to_probe = (
         tuple(candidates) if candidates is not None else LOCAL_PROVIDER_CANDIDATES
     )
+    if not to_probe:
+        return []
+
     fetch_fn = fetch or _http_get
     configured_index = _index_configured(configured or ())
 
@@ -157,7 +160,7 @@ def _probe_one(
     candidate: LocalProviderCandidate,
     timeout: float,
     fetch: FetchFn,
-    configured_index: dict[tuple[str, int], str],
+    configured_index: dict[tuple[str, str, int, str], str],
 ) -> DiscoveryResult:
     url = candidate.models_url
     parsed = urlparse(url)
@@ -291,7 +294,9 @@ def _http_get(url: str, timeout: float) -> tuple[int, bytes]:
             },
         )
         resp = conn.getresponse()
-        body = resp.read(MAX_BODY_BYTES)
+        body = resp.read(MAX_BODY_BYTES + 1)
+        if len(body) > MAX_BODY_BYTES:
+            raise OSError(f"response body exceeds {MAX_BODY_BYTES} bytes")
         return resp.status, body
     finally:
         conn.close()
@@ -328,8 +333,10 @@ def _parse_openai_models(body: bytes) -> tuple[tuple[str, ...], str | None]:
     return tuple(ids), None
 
 
-def _index_configured(configured: Sequence[object]) -> dict[tuple[str, int], str]:
-    index: dict[tuple[str, int], str] = {}
+def _index_configured(
+    configured: Sequence[object],
+) -> dict[tuple[str, str, int, str], str]:
+    index: dict[tuple[str, str, int, str], str] = {}
     for item in configured:
         name = getattr(item, "name", None)
         base_url = getattr(item, "base_url", None)
@@ -342,7 +349,7 @@ def _index_configured(configured: Sequence[object]) -> dict[tuple[str, int], str
 
 
 def _match_configured(
-    candidate: LocalProviderCandidate, index: dict[tuple[str, int], str]
+    candidate: LocalProviderCandidate, index: dict[tuple[str, str, int, str], str]
 ) -> str | None:
     key = _endpoint_key(candidate.base_url)
     if key is None:
@@ -350,16 +357,19 @@ def _match_configured(
     return index.get(key)
 
 
-def _endpoint_key(url: str) -> tuple[str, int] | None:
+def _endpoint_key(url: str) -> tuple[str, str, int, str] | None:
+    """Return a canonical (scheme, host, port, path) key for endpoint identity matching."""
     parsed = urlparse(url)
     host = (parsed.hostname or "").lower()
     if not host:
         return None
+    scheme = parsed.scheme.lower()
     port = parsed.port
     if port is None:
-        port = 443 if parsed.scheme == "https" else 80
+        port = 443 if scheme == "https" else 80
     canonical = "loopback" if _is_loopback_host(host) else host
-    return canonical, port
+    path = parsed.path.rstrip("/").lower() or "/"
+    return scheme, canonical, port, path
 
 
 def _is_loopback_host(host: str) -> bool:
--- a/gptme/llm/local_discovery.py
+++ b/gptme/llm/local_discovery.py
@@ -1,0 +1,419 @@
+"""Probe well-known local OpenAI-compatible providers.
+
+gptme should "just work" with a local Ollama or LM Studio instance. This module
+detects those servers by hitting their real OpenAI-compatible ``/v1/models``
+endpoint — not by checking whether a port is open, and not by guessing native
+APIs such as Ollama's ``/api/tags``.
+
+Candidates (loopback only):
+
+- Ollama:    ``http://127.0.0.1:11434/v1/models``
+- LM Studio: ``http://127.0.0.1:1234/v1/models``
+
+A probe is a GET of that URL with a short timeout. Success requires HTTP 200
+and an OpenAI models-list JSON body (``{"data": [{"id": ...}, ...]}``). Every
+other outcome is returned with a reason so callers never silently drop a
+candidate.
+
+This module never writes config. Listing is opt-in display; persisting a
+provider is ``gptme providers add``.
+
+Disable probes with ``GPTME_NO_LOCAL_DISCOVERY=1`` (CI / hermetic tests).
+"""
+
+from __future__ import annotations
+
+import errno
+import json
+import logging
+import os
+import time
+from collections.abc import Callable, Sequence
+from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
+from http.client import HTTPConnection, HTTPException
+from typing import Literal
+from urllib.parse import urlparse
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TIMEOUT_S = 0.5
+MAX_BODY_BYTES = 65_536
+USER_AGENT = "gptme-local-discovery/1.0"
+
+ProbeStatus = Literal["up", "down", "incompatible", "auth_required", "error"]
+
+
+@dataclass(frozen=True)
+class LocalProviderCandidate:
+    """A well-known local OpenAI-compatible server to probe."""
+
+    name: str
+    display_name: str
+    base_url: str
+    hint: str
+
+    @property
+    def models_url(self) -> str:
+        """The real OpenAI-compat models list URL (``{base_url}/models``)."""
+        return self.base_url.rstrip("/") + "/models"
+
+
+LOCAL_PROVIDER_CANDIDATES: tuple[LocalProviderCandidate, ...] = (
+    LocalProviderCandidate(
+        name="ollama",
+        display_name="Ollama",
+        base_url="http://127.0.0.1:11434/v1",
+        hint="Install from https://ollama.com and run `ollama serve`",
+    ),
+    LocalProviderCandidate(
+        name="lmstudio",
+        display_name="LM Studio",
+        base_url="http://127.0.0.1:1234/v1",
+        hint="Open LM Studio → Local Server → Start",
+    ),
+)
+
+# Hostnames treated as the same loopback endpoint when matching config.
+_LOOPBACK_HOSTS = frozenset({"127.0.0.1", "localhost", "::1", "0.0.0.0"})
+
+
+@dataclass(frozen=True)
+class DiscoveryResult:
+    """Outcome of probing one local provider candidate."""
+
+    candidate: LocalProviderCandidate
+    status: ProbeStatus
+    reason: str
+    models: tuple[str, ...] = ()
+    configured_as: str | None = None
+    elapsed_s: float = 0.0
+
+    @property
+    def is_up(self) -> bool:
+        return self.status == "up"
+
+    def to_dict(self) -> dict[str, object]:
+        return {
+            "name": self.candidate.name,
+            "display_name": self.candidate.display_name,
+            "base_url": self.candidate.base_url,
+            "models_url": self.candidate.models_url,
+            "status": self.status,
+            "reason": self.reason,
+            "models": list(self.models),
+            "configured_as": self.configured_as,
+            "hint": self.candidate.hint,
+        }
+
+
+FetchFn = Callable[[str, float], tuple[int, bytes]]
+
+
+def local_discovery_disabled() -> bool:
+    """Return True when GPTME_NO_LOCAL_DISCOVERY disables probing."""
+    raw = os.environ.get("GPTME_NO_LOCAL_DISCOVERY", "")
+    return raw.strip().lower() in {"1", "true", "yes", "on"}
+
+
+def discover_local_providers(
+    *,
+    candidates: Sequence[LocalProviderCandidate] | None = None,
+    timeout: float = DEFAULT_TIMEOUT_S,
+    configured: Sequence[object] | None = None,
+    fetch: FetchFn | None = None,
+) -> list[DiscoveryResult]:
+    """Probe well-known local providers and return a result for each candidate.
+
+    Every candidate is represented in the result list. Failures are never
+    dropped — ``status`` and ``reason`` explain why a probe did not count as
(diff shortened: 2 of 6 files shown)


gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,419 @@
...
+        tuple(candidates) if candidates is not None else LOCAL_PROVIDER_CANDIDATES
+    )
+    fetch_fn = fetch or _http_get
+    configured_index = _index_configured(configured or ())
+
+    if len(to_probe) == 1:
+        return [_probe_one(to_probe[0], timeout, fetch_fn, configured_index)]
+
+    results: list[DiscoveryResult | None] = [None] * len(to_probe)
+    with ThreadPoolExecutor(max_workers=len(to_probe)) as pool:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ae234eb5b: added an early  guard before the  call.

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,419 @@
...
+            "GET",
+            path,
+            headers={
+                "Accept": "application/json",
+                "User-Agent": USER_AGENT,
+            },
+        )
+        resp = conn.getresponse()
+        body = resp.read(MAX_BODY_BYTES)
+        return resp.status, body
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ae234eb5b: `_http_get` now reads `MAX_BODY_BYTES + 1` bytes and raises `OSError` if the limit is exceeded, so oversized bodies are classified as `"error"` rather than silently passed to the JSON parser with a truncated payload.

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,419 @@
...
+def _endpoint_key(url: str) -> tuple[str, int] | None:
+    parsed = urlparse(url)
+    host = (parsed.hostname or "").lower()
+    if not host:
+        return None
+    port = parsed.port
+    if port is None:
+        port = 443 if parsed.scheme == "https" else 80
+    canonical = "loopback" if _is_loopback_host(host) else host
+    return canonical, port
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ae234eb5b: `_endpoint_key` now returns a 4-tuple `(scheme, canonical_host, port, path)` instead of `(host, port)`, so URLs that share a loopback port but differ in scheme or base path are treated as distinct endpoints.

TimeToBuildBob:
@greptileai review

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,429 @@
...
+        return (), f"JSON {type(payload).__name__} is not an object"
+    data = payload.get("data")
+    if not isinstance(data, list):
+        # Deliberately reject Ollama's native {"models": [...]} /api/tags shape.
+        if isinstance(payload.get("models"), list):
+            return (), "native Ollama /api/tags shape, not OpenAI /v1/models"
+        return (), "missing OpenAI 'data' list"
+    ids: list[str] = []
+    for item in data:
+        if isinstance(item, str) and item:
```
TimeToBuildBob, line 337: ⚠️ **P2** — In `_parse_openai_models`, the parser accepts any list under the `data` key and silently discards items that are not a non-empty string or a dict with a non-empty string `id`. A server returning `{"data": [1, 2, 3]}` — a non-OpenAI-compatible body — passes validation and is reported as status "up" with zero models. The PR's stated contract requires `{"data": [{"id": ...}]}`, so this false positive makes `gptme providers list` claim a server is usable when it is not. Fix by rejecting the body if any item in `data` is not a valid string/dict id.

gptme/cli/util.py
```diff
@@ -145,9 +145,35 @@ def providers():
...
+    discovered = []
+    if discover:
+        from ..llm.local_discovery import discover_local_providers  # fmt: skip
+
+        discovered = discover_local_providers(configured=config.user.providers)
+
+    if as_json:
+        payload = {
+            "configured": [_configured_provider_dict(p) for p in config.user.providers],
+            "discovered": [r.to_dict() for r in discovered],
```
TimeToBuildBob, line 177: ⚠️ **P2** — When `GPTME_NO_LOCAL_DISCOVERY` is set (or `--no-discover` is used), `gptme providers list --json` outputs a `discovered` array that is empty, with no field indicating that probing was skipped. Consumers of the JSON output cannot distinguish an environment where no local server was found from one where discovery was disabled, leading to incorrect assumptions about the local environment. Add a boolean like `discovery_disabled` to the JSON payload to expose this state.

TimeToBuildBob:
## 🤖 AI code review

Adds local OpenAI-compatible provider auto-discovery: a new gptme/llm/local_discovery.py module that probes hardcoded loopback endpoints for Ollama (:11434) and LM Studio (:1234) by GETting the real /v1/models path and parsing the OpenAI models-list shape. Updates the `gptme providers list` CLI to run these probes, display results with terminal escape stripping, and support --json/--no-discover and the GPTME_NO_LOCAL_DISCOVERY env var. Adds tests for the discovery module and the CLI integration.

<details><summary><b>3 findings</b> · ❌ **1** P1 · ⚠️ **2** P2 · 🔒 **1** security — 🛑 **1 at P0/P1**</summary>

**❌ P1 high** — `gptme/llm/local_discovery.py:208`

On Python versions before 3.10, a connection timeout raised by `socket` is an `OSError` (specifically `socket.timeout`) rather than the built-in `TimeoutError`. In `_probe_one`, the `except TimeoutError` branch will not catch it, so it falls through to `except OSError`. There, `_looks_like_unreachable` checks only for `errno` values like `ECONNREFUSED`, `ECONNRESET`, etc., and does not include `errno.ETIMEDOUT`. Consequently, a probe that times out is classified as `error` instead of `down`, producing a misleading status for a local server that is slow to respond or not reachable. Users will see `❌` instead of `⚪` for a down endpoint.

```suggestion
Add `errno.ETIMEDOUT` to the unreachable `errno` set, or catch `socket.timeout` explicitly.
```

<sub>**How this was verified:** Checked that `_looks_like_unreachable` does not include `errno.ETIMEDOUT` and that `socket.timeout` is a subclass of `OSError` prior to Python 3.10. The project uses `from __future__ import annotations` and type hints compatible with Python 3.8, so this path is reachable.</sub>

<sub>Consensus: 2/3 passes agreed — pass 2 looked and disagreed</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/llm/local_discovery.py:154`

`discover_local_providers` creates a `ThreadPoolExecutor` with `max_workers=len(to_probe)`. While the default candidates are only two, the function is public and accepts an arbitrary `candidates` sequence. A caller passing a large list (e.g., from programmatically enumerating ports) will spawn one thread per candidate with no upper bound, potentially creating hundreds or thousands of threads. This can exhaust thread and memory resources, degrading or crashing the process. The function should cap the worker count (e.g., to `min(len(to_probe), 4)`) and batch the probes, or use a fixed pool. This is an availability issue, not a security one, since the candidates are user-supplied and the probes are loopback-only.

```suggestion
with ThreadPoolExecutor(max_workers=min(len(to_probe), 4)) as pool:
```

<sub>**How this was verified:** Confirmed the only in-repo caller passes the two hardcoded candidates, but the function's signature accepts any sequence and the parameter is documented as public. No cap or batching is applied.</sub>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `gptme/llm/local_discovery.py:109`

The `to_dict` method of `DiscoveryResult` returns model IDs verbatim, without applying the `_strip_controls` sanitization that the human-readable CLI path uses. When `gptme providers list --json` is invoked against a local server that returns a model ID containing terminal escape sequences (e.g., `\x1b]51;A\x07evil-model`), those raw bytes are emitted into the JSON payload. If a user pipes this JSON output to a terminal (e.g., via `gptme providers list --json | less` or views it in a TUI), the escape sequences are interpreted, enabling terminal injection. The non-JSON path explicitly strips these characters in `_print_discovery_result`, so this is an inconsistency that undermines the security measure added in the same PR. The `models` field in the JSON is populated from `self.models` without any filtering.

```suggestion
Apply `_strip_controls` to each model ID and reason when constructing the dict, or sanitize in the CLI JSON branch before serialization.
```

<sub>**How this was verified:** Confirmed that `_print_discovery_result` calls `_strip_controls` on `result.models` and `result.reason` before display, while `to_dict` does not. The JSON output in `util.py` uses `r.to_dict()` directly.</sub>

</details>

- `docs/getting-started.rst` — Adds a one-line example of running `gptme providers list` to verify local server discovery.
- `docs/providers-custom.rst` — Documents the auto-discovery behavior, probe URLs, and escape hatches (--no-discover, GPTME_NO_LOCAL_DISCOVERY).
- `gptme/cli/util.py` — Expands `providers list` with --discover/--no-discover and --json flags, renders discovered results, and adds helper functions for control-character stripping and provider dict conversion.
- `gptme/llm/local_discovery.py` — New module implementing candidate probes, HTTP GET, OpenAI-shape parsing, endpoint matching, and result data structures.
- `tests/test_local_discovery.py` — Adds tests for discovery against a real loopback HTTP server, covering successful and failed probes, parsing, and endpoint key behavior.
- `tests/test_util_cli_providers.py` — Adds CLI tests for the discovery output and JSON mode, plus unit tests for control-character stripping and escape injection.

</details>

| `ae234eb5b9f6` | 4/5 | 2 | llm | 2026-08-23 01:11 UTC |
| `672f2b4f49eb` | 4/5 | 1 | llm | 2026-08-23 01:49 UTC |
| `8be24906024f` | 4/5 | 2 | llm | 2026-08-23 02:27 UTC |
| `a91ae7a39efc` | 4/5 | 4 | llm | 2026-08-23 03:26 UTC |
| `2f4308434ef7` | 3/5 | 5 | llm | 2026-08-23 07:03 UTC |
| `b3383a3c8db3` | 3/5 | 1 | llm | 2026-08-23 08:41 UTC |
| `614c9c1e688a` | 4/5 | 2 | llm | 2026-08-23 09:21 UTC |
| `56c7e19a20a2` | 4/5 | 3 | llm | 2026-08-23 09:42 UTC |

</details>

<sub>Reviewed `2d9f4ab92a9d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 252s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Addressed both P2 findings in `672f2b4f4`:

**P2-1 (`_parse_openai_models`):** Added an `else` branch to the item-validation loop — a `data` item that is neither a non-empty string nor a valid dict now returns `((), "data item is not a model object: <type>")`, so a body like `{"data": [1, 2, 3]}` is classified `"error"` rather than `"up"` with zero models.

**P2-2 (`providers list --json`):** Added `"discovery_disabled": not discover` to the JSON payload, so consumers can distinguish "no servers found" from "discovery was disabled via `--no-discover` / `GPTME_NO_LOCAL_DISCOVERY`".

All 36 tests still passing.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed remaining P2 in `8be249060`:

**P2 (`discovery_disabled` env var not reflected in JSON):** When `GPTME_NO_LOCAL_DISCOVERY` is set but `--discover` flag is default (True), the JSON payload now correctly reports `discovery_disabled: true`. The prior fix only checked the CLI flag; machines consuming the JSON output had no way to distinguish "no servers found" from "discovery skipped via env var".

Fix: `_env_disabled = discover and local_discovery_disabled()`, then `"discovery_disabled": not discover or _env_disabled`. Short-circuits correctly when `discover=False` (no import needed), and `local_discovery_disabled` is already imported in the `if discover:` block when it's needed.

Three new tests added: flag-disabled, env-disabled, and enabled-but-empty cases. 39/39 passing.

gptme/cli/util.py
```diff
@@ -160,29 +191,92 @@ def providers_list():
...
-                f"   API Key: ${provider.name.upper().replace('-', '_')}_API_KEY (default)"
+                f"      OPENAI_BASE_URL={cand.base_url} gptme -m local/{example}"
+            )
+    else:
+        click.echo(f"      {result.reason}")
+        if result.status == "down":
+            click.echo(f"      hint: {cand.hint}")
+        if result.configured_as:
+            click.echo(
+                f"      configured as '{result.configured_as}' but the endpoint is not reachable"
```
TimeToBuildBob: ⚠️ **P2** — In _print_discovery_result, the message 'configured as X but the endpoint is not reachable' is printed for any status other than 'up' when the provider is already configured. This includes 'incompatible', 'auth_required', and 'error' statuses, which do not mean the endpoint is unreachable. For example, an 'incompatible' result means the server responded but with a non-OpenAI body, an 'auth_required' result means the server was reachable but required credentials, and an 'error' result covers HTTP errors like 404. A user with a configured provider that returns, say, a 404 will see 'configured as mine but the endpoint is not reachable', wrongly implying a network problem and steering them toward connectivity checks instead of the actual HTTP error. This misleads the user into thinking the server is down when it is actually reachable but misconfigured. _(1/2 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,431 @@
...
+            ids.append(item)
+        elif isinstance(item, dict):
+            model_id = item.get("id")
+            if isinstance(model_id, str) and model_id:
```
TimeToBuildBob, line 341: ⚠️ **P2** — In _parse_openai_models, a dict item in the 'data' list whose 'id' is missing or is an empty string is silently ignored without raising an error or returning an 'incompatible' status. If a server returns a partially malformed list (e.g., some entries missing the 'id' field), the probe reports status 'up' with an incomplete model list. A user relying on this list to pick a model may not see available models, and the server is incorrectly treated as fully OpenAI-compatible. The function should either reject the entire response as incompatible when an item is a dict without a valid non-empty string 'id', or include the item so the list is complete. _(2/2 passes)_

TimeToBuildBob:
@greptileai review

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,431 @@
...
+    parsed = urlparse(url)
+    host = (parsed.hostname or "").lower()
+    if not host:
+        return None
+    scheme = parsed.scheme.lower()
+    port = parsed.port
+    if port is None:
+        port = 443 if scheme == "https" else 80
+    canonical = "loopback" if _is_loopback_host(host) else host
+    path = parsed.path.rstrip("/").lower() or "/"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `a91ae7a39`: removed `.lower()` from the path component in `_endpoint_key`. HTTP paths are case-sensitive (RFC 7230); scheme and host normalisation is correct but path case-folding produced false matches for providers whose base URL has non-lowercase paths (e.g. `/V1` vs `/v1`). Test added: `test_case_distinct_paths_are_not_equal`.

TimeToBuildBob:
@greptileai review

tests/test_local_discovery.py
```diff
@@ -0,0 +1,279 @@
...
+
+def test_reports_connection_refused_instead_of_dropping() -> None:
+    # Bind-and-close so the port is unused; probe must still return a result.
+    import socket
+
+    sock = socket.socket()
+    sock.bind(("127.0.0.1", 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    base = f"http://127.0.0.1:{port}/v1"
```
TimeToBuildBob, line 127: ⚠️ **P2** — test_reports_connection_refused_instead_of_dropping (tests/test_local_discovery.py) binds a socket to an ephemeral port, records the port, then closes it, and only afterward probes that port via discover_local_providers. Between sock.close() and the HTTPConnection connect inside _http_get, the kernel may reassign the freed port to any other process on the machine (another test, a dev service, an OS worker). If a different service grabs the port, the probe can succeed (status 'up'), get a 404 (status 'error'), or time out — all causing the assertions results[0].status == "down" and the 'connection refused'/'not running' reason check to fail. The test is non-hermetic and its outcome depends on unrelated machine activity, producing spurious CI failures. _(1/2 passes — pass 2 looked and disagreed)_

gptme/cli/util.py
```diff
@@ -160,29 +191,92 @@ def providers_list():
...
+
+    if not discover:
+        return
 
-    for provider in config.user.providers:
-        click.echo(f"📡 {provider.name}")
-        click.echo(f"   Base URL: {provider.base_url}")
+    click.echo("🔍 Local auto-discovery")
+    if not discovered:
+        click.echo("   (disabled via GPTME_NO_LOCAL_DISCOVERY)")
```
TimeToBuildBob: ⚠️ **P2** — When `gptme providers list` runs with discovery enabled but neither Ollama nor LM Studio is reachable, the `discovered` list is empty. The code then prints `(disabled via GPTME_NO_LOCAL_DISCOVERY)` because it checks `if not discovered:` to decide whether discovery was disabled. This message is misleading: discovery actually ran but found no running servers. The user sees a claim that the feature is disabled when it isn't, which is confusing when troubleshooting. The check should distinguish between "discovery disabled" (env var set or --no-discover) and "discovery ran but found nothing". Fix by computing `_env_disabled` and using it to decide the message, or by printing a different message like 'No local providers found' when discovery ran. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Ready for merge — all review findings addressed, CI green, Greptile passed.

**What was fixed across this arc:**
- P2 (empty `candidates=[]` → `ValueError`): early-return guard added
- P2 (truncated 64KB body silently misclassified): `_http_get` now caps + raises `OSError` on oversized responses
- P2 (`_endpoint_key` host+port only → wrong alias detection): extended to 4-tuple `(scheme, host, port, path)`
- P2 (`_parse_openai_models` false-positive on `{"data": [1,2,3]}`): `else` branch rejects non-model items
- P2 (`discovery_disabled` missing from JSON output when env var used): JSON payload now reflects env disable separately from flag
- P1 (`_endpoint_key` lowercased path → case-sensitive path mismatch): `.lower()` removed from path component

Not self-merge eligible (new core module). Waiting for Erik to merge.

TimeToBuildBob:
Fixed remaining AI review P2 findings in `2f4308434`:

**P2 (`_parse_openai_models` — dict item with missing `id` silently skipped):** A `{"data": [{"name": "llama3"}]}` item (dict without valid `id`) was silently ignored, letting the probe return status `"up"` with an incomplete model list. Fix: added an `else` branch inside the `isinstance(item, dict)` case that returns `()` and a descriptive reason, rejecting the whole response as incompatible.

**P2 (`_print_discovery_result` — wrong "endpoint not reachable" for non-down statuses):** For `incompatible`, `auth_required`, and `error` statuses, the configured-provider message said "endpoint is not reachable" even though the server responded (just not as expected). Fix: message is now gated on `result.status == "down"` — other statuses say "did not respond as expected".

**P2 (`gptme providers list` — "(disabled via GPTME_NO_LOCAL_DISCOVERY)" when discovery ran but found nothing):** The `if not discovered:` check can trigger either because the env var disabled discovery OR because no servers were found. Both printed the "disabled" message. Fix: now calls `local_discovery_disabled()` to pick the right message ("disabled via env var" vs "no local providers found").

**Test (non-hermetic port-reuse race):** Kept the socket bound-but-not-listening during the probe so the OS can't reassign the port between `close()` and the connection attempt.

40/40 tests passing. Waiting for Erik to merge.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Status update: CI green on `2f43084` (all 15 checks pass). Latest commit addresses the remaining AI-review P2 findings (reject dicts missing `id`; fix status-sensitive "not reachable" messages; hermetic port test). Greptile re-triggered on this HEAD — waiting for its response. Not eligible for self-merge (touches core CLI/LLM files). Ready for maintainer review/merge.

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+def _endpoint_key(url: str) -> tuple[str, str, int, str] | None:
+    """Return a canonical (scheme, host, port, path) key for endpoint identity matching."""
+    parsed = urlparse(url)
+    host = (parsed.hostname or "").lower()
+    if not host:
+        return None
+    scheme = parsed.scheme.lower()
+    port = parsed.port
```
TimeToBuildBob: ❌ **P1** — In `_endpoint_key`, `parsed.port` raises `ValueError` when the URL contains a non-numeric port (e.g. configuration `base_url = "http://localhost:abc/v1"`). `_index_configured` iterates every configured provider and calls `_endpoint_key` without any try/except, so a single malformed `[[providers]]` entry makes the entire `gptme providers list` command crash with a traceback instead of listing anything. The same code path runs for every invocation of discovery, including `--json`. This is a real, user-reachable failure: configuration files are human-edited TOML, so a typo in the port is plausible, and the command should degrade gracefully rather than abort. _(1/2 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+def _http_get(url: str, timeout: float) -> tuple[int, bytes]:
+    """GET ``url`` without following redirects. Loopback-only by construction."""
+    parsed = urlparse(url)
+    host = parsed.hostname
+    if not host:
+        raise OSError("missing host")
+    port = parsed.port or (443 if parsed.scheme == "https" else 80)
+    if parsed.scheme != "http":
```
TimeToBuildBob, line 290: ⚠️ **P2** — In `_http_get`, the `port` is derived from `parsed.port` or defaulted to 80/443, but the URL scheme is checked only against `http` (line 280). If a custom candidate (or a future built-in) uses an `https` URL, `_http_get` raises `OSError(f"unsupported scheme {parsed.scheme!r}")`, which `_probe_one` catches and maps to status `error`. However, the `_endpoint_key` matching in `_index_configured`/`_match_configured` would still treat an https URL as a loopback endpoint and could mark a configured provider as discovered even though the probe never performs a TLS request. Concretely, a configured provider with `base_url = "https://127.0.0.1:11434/v1"` matches the candidate's key on scheme/host/port/path, so discovery reports it as `configured_as` and `up`/`error` depending on the probe outcome, but the probe itself rejects the scheme. The result is a misleading 'already configured' message for a URL that cannot be probed by this module. _(1/2 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,433 @@
...
+    LocalProviderCandidate(
+        name="lmstudio",
+        display_name="LM Studio",
+        base_url="http://127.0.0.1:1234/v1",
+        hint="Open LM Studio → Local Server → Start",
+    ),
+)
+
+# Hostnames treated as the same loopback endpoint when matching config.
+_LOOPBACK_HOSTS = frozenset({"127.0.0.1", "localhost", "::1", "0.0.0.0"})
```
TimeToBuildBob: ⚠️ **P2** — _LOOPBACK_HOSTS includes "0.0.0.0", which is not a loopback address but the unspecified/wildcard address. `_endpoint_key` (line 374) uses `_is_loopback_host` to normalize the host to "loopback", so a configured provider with base_url `http://0.0.0.0:11434/v1` will match the probe candidate `http://127.0.0.1:11434/v1`. `discover_local_providers` will then report that provider as "already configured" even though `0.0.0.0` is not a valid destination for the OpenAI client (connections to `0.0.0.0` fail on most systems). This gives the user a false confirmation that their misconfigured provider is correct and reachable, masking the real problem. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed P1 AI-review finding in `b3383a3c8`:

**P1 (`_endpoint_key` crashes on malformed port):** A TOML typo like `base_url = "http://localhost:abc/v1"` caused `urlparse().port` to raise `ValueError`. `_index_configured` called `_endpoint_key` for every configured provider without try/except, so a single bad entry aborted `gptme providers list` with a traceback. Fix: wrapped `parsed.port` in `try/except ValueError` — returns `None` (skipped) instead of crashing. Test added.

41/41 tests pass. Not eligible for self-merge. Ready for maintainer review.

TimeToBuildBob:
@greptileai review

gptme/cli/util.py
```diff
@@ -160,29 +191,100 @@ def providers_list():
...
+    }.get(result.status, "⚪")
+    click.echo(f"   {status_icon} {cand.display_name}  {cand.base_url}")
+    click.echo(f"      probe: {cand.models_url}")
+    if result.status == "up":
+        if result.models:
+            shown = ", ".join(result.models[:8])
+            extra = (
+                f" (+{len(result.models) - 8} more)" if len(result.models) > 8 else ""
+            )
+            click.echo(f"      models ({len(result.models)}): {shown}{extra}")
```
TimeToBuildBob, line 272: ❌ **P1** — The text output of `gptme providers list` prints model IDs and probe-reason strings verbatim via `click.echo` without sanitizing terminal control characters. A malicious process listening on the probed loopback ports (11434/1234) can return a `/v1/models` body whose model IDs or malformed-data error messages contain terminal escape sequences (e.g., `\x1b]51;...\x07` to trigger clipboard/notifier sequences). These sequences are written directly to the user's terminal, enabling terminal escape injection that could execute commands, manipulate the terminal buffer, or exfiltrate data. The fix is to strip or escape control characters from all strings derived from the untrusted probe response before printing.

TimeToBuildBob:
@greptileai review

gptme/cli/util.py
```diff
@@ -160,29 +191,108 @@ def providers_list():
...
+        else:
+            click.echo("      models: (none listed — server is up)")
+        if result.configured_as:
+            click.echo(f"      already configured as '{result.configured_as}'")
         else:
+            example = _strip_controls(result.models[0]) if result.models else "<model>"
```
TimeToBuildBob: ⚠️ **P2** — In `_print_discovery_result`, the `example` variable is derived from the first model ID after `_strip_controls`. If a model ID consists entirely of control characters (e.g. "\x1b" or "\x00"), `_strip_controls` returns an empty string, leaving `example` as "". The subsequent line prints `OPENAI_BASE_URL={cand.base_url} gptme -m local/{example}`, which yields a trailing slash with no model name (e.g. `gptme -m local/`). This produces a malformed and misleading command. The root cause is not filtering out empty strings after stripping; the model ID is non-empty before stripping (the parser only admits non-empty ids), so this can occur with a crafted server response. Fix: filter out stripped-empty IDs when selecting the example, or fall back to `<model>`. _(1/2 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,436 @@
...
+            status="error",
+            reason=f"refusing to probe non-loopback host {host!r}",
+        )
+        _log_result(result)
+        return result
+
+    t0 = time.monotonic()
+    try:
+        status_code, body = fetch(url, timeout)
+    except TimeoutError:
```
TimeToBuildBob, line 185: ⚠️ **P2** — In Python versions before 3.10, `socket.timeout` is not an alias for `TimeoutError`; it inherits from `OSError`. The `_http_get` call uses `HTTPConnection` with a timeout, which raises `socket.timeout` when a probe exceeds the timeout. The `except TimeoutError` clause does not catch this on those Python versions, so the exception falls through to `except OSError`. There, `_looks_like_unreachable(e)` returns `False` for a timeout because it does not check for 'timed out' or 'timeout' in the message; `_short_oserror` would convert the message to 'timeout', but `_looks_like_unreachable` does not include that token. Consequently, a timeout is reported with status `'error'` and reason `'timeout'` instead of the intended `'down'`. This misclassifies a non-responsive local server as a generic error, which is misleading in both the CLI display and the JSON output. The fix is to add 'timed out' and 'timeout' to the token list in `_looks_like_unreachable`, or to catch `socket.timeout` explicitly before the `OSError` branch. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed remaining P2 AI-review finding in `56c7e19a2`:

**P2 (`_print_discovery_result` — all-control-char model IDs yield empty example):** When every model ID returned by a crafted server is stripped to `""` by `_strip_controls`, the example command printed `OPENAI_BASE_URL=... gptme -m local/` with a trailing slash and no model name — misleading and malformed. Fix: replaced the direct `_strip_controls(result.models[0])` call with a generator that iterates all stripped IDs and falls back to `<model>` when none survive stripping. Test added (`test_all_control_model_ids_fall_back_to_placeholder`).

**P2 (fp: 1f82dbb7bd2a — `socket.timeout` not caught on Python < 3.10):** Not applicable — gptme requires Python ≥3.10 (`pyproject.toml: requires-python = ">=3.10"`), where `socket.timeout is TimeoutError` by definition. The `except TimeoutError` branch already catches it.

TimeToBuildBob:
@greptileai review

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,436 @@
...
+    LocalProviderCandidate(
+        name="lmstudio",
+        display_name="LM Studio",
+        base_url="http://127.0.0.1:1234/v1",
+        hint="Open LM Studio → Local Server → Start",
+    ),
+)
+
+# Hostnames treated as the same loopback endpoint when matching config.
+_LOOPBACK_HOSTS = frozenset({"127.0.0.1", "localhost", "::1", "0.0.0.0"})
```
TimeToBuildBob: ⚠️ **P2** — `_LOOPBACK_HOSTS` includes `"0.0.0.0"` which is not a loopback address; it is the unspecified/any address. The loopback check in `_is_loopback_host` therefore treats `0.0.0.0` as loopback, and `_probe_one` will attempt an HTTP GET to `http://0.0.0.0:{port}/v1/models` instead of rejecting it as non-loopback. While the built-in candidates use `127.0.0.1`, a caller passing a candidate with `0.0.0.0` would bypass the intended loopback-only guard and probe a non-loopback (though usually unavailable) address. This is a minor deviation from the documented 'loopback only' intent. _(2/3 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,436 @@
...
+    return result
+
+
+def _http_get(url: str, timeout: float) -> tuple[int, bytes]:
+    """GET ``url`` without following redirects. Loopback-only by construction."""
+    parsed = urlparse(url)
+    host = parsed.hostname
+    if not host:
+        raise OSError("missing host")
+    port = parsed.port or (443 if parsed.scheme == "https" else 80)
```
TimeToBuildBob: ⚠️ **P2** — In `_http_get`, `parsed.port` raises `ValueError` when the URL contains a non-numeric port (e.g., `http://127.0.0.1:abc/v1`). This exception is not caught by `_probe_one`'s exception handlers, which only catch `TimeoutError`, `ConnectionRefusedError`/`Reset`/`BrokenPipe`, `OSError`, and `HTTPException`. As a result, a malformed `LocalProviderCandidate` passed via the `candidates` parameter to `discover_local_providers` would propagate the `ValueError` up through `ThreadPoolExecutor.submit` and crash the `gptme providers list` command with a traceback. While the built-in candidates always have valid ports, the function is public and accepts a custom `candidates` sequence, so a caller (or a future candidate entry) with a typo in the port would trigger this. The fix is to catch `ValueError` in `_http_get` and raise an `OSError` with a descriptive message, or add a `ValueError` handler in `_probe_one`. _(1/3 passes — passes 1, 3 looked and disagreed)_

gptme/cli/util.py
TimeToBuildBob: Already fixed in `56c7e19a2` (iterate stripped IDs, fall back to `<model>`). This thread is stale against current HEAD.

gptme/llm/local_discovery.py
TimeToBuildBob: Fixed in `2d9f4ab92`: split probe hosts from config-matching aliases. `_probe_one` now refuses `0.0.0.0` as non-loopback; `_endpoint_key` still treats a bind-address `base_url` as the same endpoint as `127.0.0.1`/`localhost`.

gptme/llm/local_discovery.py
TimeToBuildBob: Fixed in `2d9f4ab92`: `_http_get` now catches `urlparse().port` `ValueError` and raises `OSError`, which `_probe_one` already handles. A custom candidate with a non-numeric port returns an error result instead of crashing `gptme providers list`.

TimeToBuildBob:
Landed `2d9f4ab92` for the two remaining open AI-review P2s:

- Probe guard no longer treats `0.0.0.0` as loopback; config matching still aliases a bind-address `base_url` to 127.0.0.1/localhost.
- `_http_get` converts malformed-port `ValueError` to `OSError`, so a custom candidate with a typo'd port returns an error result instead of crashing `gptme providers list`.

Resolved the three leftover review threads (including the already-fixed empty-example one). 50 tests passing locally (`tests/test_local_discovery.py` + `tests/test_util_cli_providers.py`). Not self-merge eligible (core CLI/LLM). Ready for maintainer review/merge.

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,450 @@
...
+            reason=f"not running ({_short_oserror(e)})",
+            elapsed_s=time.monotonic() - t0,
+            configured_as=_match_configured(candidate, configured_index),
+        )
+        _log_result(result)
+        return result
+    except OSError as e:
+        # Connection refused often arrives as OSError on some platforms.
+        reason = _short_oserror(e)
+        status: ProbeStatus = "down" if _looks_like_unreachable(e) else "error"
```
TimeToBuildBob, line 208: ❌ **P1** — On Python versions before 3.10, a connection timeout raised by `socket` is an `OSError` (specifically `socket.timeout`) rather than the built-in `TimeoutError`. In `_probe_one`, the `except TimeoutError` branch will not catch it, so it falls through to `except OSError`. There, `_looks_like_unreachable` checks only for `errno` values like `ECONNREFUSED`, `ECONNRESET`, etc., and does not include `errno.ETIMEDOUT`. Consequently, a probe that times out is classified as `error` instead of `down`, producing a misleading status for a local server that is slow to respond or not reachable. Users will see `❌` instead of `⚪` for a down endpoint. _(2/3 passes — pass 2 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,450 @@
...
+        return []
+
+    fetch_fn = fetch or _http_get
+    configured_index = _index_configured(configured or ())
+
+    if len(to_probe) == 1:
+        return [_probe_one(to_probe[0], timeout, fetch_fn, configured_index)]
+
+    results: list[DiscoveryResult | None] = [None] * len(to_probe)
+    with ThreadPoolExecutor(max_workers=len(to_probe)) as pool:
```
TimeToBuildBob, line 154: ⚠️ **P2** — `discover_local_providers` creates a `ThreadPoolExecutor` with `max_workers=len(to_probe)`. While the default candidates are only two, the function is public and accepts an arbitrary `candidates` sequence. A caller passing a large list (e.g., from programmatically enumerating ports) will spawn one thread per candidate with no upper bound, potentially creating hundreds or thousands of threads. This can exhaust thread and memory resources, degrading or crashing the process. The function should cap the worker count (e.g., to `min(len(to_probe), 4)`) and batch the probes, or use a fixed pool. This is an availability issue, not a security one, since the candidates are user-supplied and the probes are loopback-only. _(1/3 passes — passes 2, 3 looked and disagreed)_

gptme/llm/local_discovery.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def to_dict(self) -> dict[str, object]:
+        return {
+            "name": self.candidate.name,
+            "display_name": self.candidate.display_name,
+            "base_url": self.candidate.base_url,
+            "models_url": self.candidate.models_url,
+            "status": self.status,
+            "reason": self.reason,
+            "models": list(self.models),
```
TimeToBuildBob, line 109: ⚠️ **P2** — The `to_dict` method of `DiscoveryResult` returns model IDs verbatim, without applying the `_strip_controls` sanitization that the human-readable CLI path uses. When `gptme providers list --json` is invoked against a local server that returns a model ID containing terminal escape sequences (e.g., `\x1b]51;A\x07evil-model`), those raw bytes are emitted into the JSON payload. If a user pipes this JSON output to a terminal (e.g., via `gptme providers list --json | less` or views it in a TUI), the escape sequences are interpreted, enabling terminal injection. The non-JSON path explicitly strips these characters in `_print_discovery_result`, so this is an inconsistency that undermines the security measure added in the same PR. The `models` field in the JSON is populated from `self.models` without any filtering. _(1/3 passes — passes 1, 2 looked and disagreed)_

gptme/llm/local_discovery.py
TimeToBuildBob: Not applicable — gptme's `pyproject.toml` requires Python ≥3.10 (`requires-python = ">=3.10"`). In Python 3.10+, `socket.timeout` is an alias for the built-in `TimeoutError` (PEP 647 / CPython 3.10 changelog: `socket.timeout` is now an alias for `TimeoutError`), so the `except TimeoutError` branch already catches it. No code change needed.

gptme/llm/local_discovery.py
TimeToBuildBob: Fixed in `ea2c587b5`: changed `max_workers=len(to_probe)` to `max_workers=min(len(to_probe), 8)`. The two built-in candidates already went through the `len == 1` fast path or spawn two threads; the cap prevents a large caller-supplied `candidates` sequence from opening an unbounded thread pool.

gptme/llm/local_discovery.py
TimeToBuildBob: Not applicable as stated. Python's `json.dumps()` (which `click.echo(json.dumps(...))` calls) escapes all control characters as `\u00xx` sequences per RFC 8259. A model ID containing `\x1b` becomes `"\u001b"` in the serialized JSON — literal text when printed, not an interpreted escape sequence. The terminal injection risk exists only on the human-readable display path (where `_strip_controls` was already added); the JSON path is safe by construction.

TimeToBuildBob:
Addressed the remaining 3 open AI-review threads from the latest pass:

**P1 (fp:620426198774 -- `socket.timeout` not caught on Python < 3.10):** Not applicable -- gptme requires Python >=3.10 (`pyproject.toml: requires-python = ">=3.10"`). In Python 3.10+, `socket.timeout` is an alias for `TimeoutError`, so the `except TimeoutError` branch already catches it.

**P2 (fp:fcd4fe04603d -- unbounded `ThreadPoolExecutor`):** Fixed in `ea2c587b5` -- capped to `min(len(to_probe), 8)`. Default built-in candidates (2) are unaffected; the cap prevents a large caller-supplied `candidates` sequence from opening an unbounded thread pool.

**P2 (fp:16fb9cdaf6c3 -- `to_dict()` returns model IDs without `_strip_controls`):** Not applicable. Python's `json.dumps()` escapes all control characters as `\u00xx` sequences per RFC 8259. A model ID containing `\x1b` becomes `""` in the JSON output -- literal text, not an interpreted escape sequence. Terminal injection through the JSON path is not possible.

51 tests pass locally. Not self-merge eligible (core CLI/LLM feature). Ready for maintainer review.

TimeToBuildBob:
Resolved the 3 remaining open AI-review threads (socket.timeout applicability, ThreadPoolExecutor cap, and JSON strip_controls). CI is still green. Ready for maintainer review/merge.

PR #3607, opened 2026-08-24
test(webui): add opt-in vision_assert helper for generation E2E

TimeToBuildBob:
## Summary

Opt-in vision-model assertion helper for webui E2E. Default CI is unchanged: the helper is a no-op unless `GPTME_VISION_ASSERT=1`.

Playwright `toBeVisible()` passes when the full string is in the DOM even if `overflow:hidden` clips it. A vision model judging pixels can catch that class of failure. Live check on 2026-08-23 with Gemini 2.5 Flash via OpenRouter:

| Fixture | Playwright `toBeVisible()` | Vision verdict |
|---------|----------------------------|----------------|
| Overflow-clipped crop | pass | **fail** (tail of the sentence missing) |
| Unclipped control | pass | **pass** |

This is **not** a follow-up bugfix for #3440 (already covered by #3441 / #3450 / #3461 / #3468). It is a separate experiment: whether a vision assertion layer adds coverage DOM assertions miss.

## What landed

- `webui/e2e/helpers/visionAssertCore.ts` — prompt, JSON parse, OpenRouter call (Jest-tested, no Playwright dep)
- `webui/e2e/helpers/visionAssert.ts` — Playwright screenshot + fail-closed assert
- Wrapped around the mock/echo roundtrip test in `webui/e2e/generation.spec.ts`
- Gap suite: `webui/e2e/vision-assert-gap.spec.ts` (DOM half always runs; vision half opt-in)
- Jest: 11 unit tests in `webui/e2e/helpers/visionAssert.test.ts` (no network)

```bash
GPTME_VISION_ASSERT=1 GPTME_VISION_API_KEY=... \
  npx playwright test e2e/vision-assert-gap.spec.ts
```

Artifacts: `webui/test-results/vision-assert/` (`VISION_ASSERT_DIR` to override).

## CI cost / secrets — not always-on

- **Secret**: `OPENROUTER_API_KEY` or `GPTME_VISION_API_KEY`
- **Default model**: `google/gemini-2.5-flash` (`GPTME_VISION_MODEL` to override)
- **Cost**: ~$0.001 per assertion on the fixtures above
- **Latency**: 1–6s per screenshot (30s timeout)
- **Failure mode**: the model can hallucinate pass/fail. Treat as a second opinion, not a replacement for DOM assertions.

Do **not** enable this on every webui CI run. That would add a secret dependency, ~2–6s, and a flaky paid path to the mock/echo suite that currently needs no credentials.

## Test plan

- [x] `npx jest e2e/helpers/visionAssert.test.ts` — 11/11 pass (no network)
- [x] Gap suite with flag off is a DOM-only no-op (vision path skipped)
- [x] Live Gemini 2.5 Flash run: clipped fixture fails vision, unclipped passes
- [ ] CI `npm test` picks up the new Jest helper tests
- [ ] CI Playwright e2e still green with the no-op wrapper in `generation.spec.ts`

### DIFF
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -1,0 +1,156 @@
+import { describe, expect, it } from '@jest/globals';
+import {
+  DEFAULT_VISION_MODEL,
+  OPENROUTER_CHAT_URL,
+  buildVisionAssertPrompt,
+  isVisionAssertEnabled,
+  parseVisionVerdict,
+  requestVisionVerdict,
+  resolveVisionApiKey,
+} from './visionAssertCore';
+
+describe('visionAssertCore', () => {
+  describe('isVisionAssertEnabled', () => {
+    it('is off unless GPTME_VISION_ASSERT=1', () => {
+      expect(isVisionAssertEnabled({})).toBe(false);
+      expect(isVisionAssertEnabled({ GPTME_VISION_ASSERT: 'true' })).toBe(false);
+      expect(isVisionAssertEnabled({ GPTME_VISION_ASSERT: '1' })).toBe(true);
+    });
+  });
+
+  describe('resolveVisionApiKey', () => {
+    it('prefers GPTME_VISION_API_KEY over OPENROUTER_API_KEY', () => {
+      expect(
+        resolveVisionApiKey({
+          GPTME_VISION_API_KEY: 'vision-key',
+          OPENROUTER_API_KEY: 'or-key',
+        })
+      ).toBe('vision-key');
+      expect(resolveVisionApiKey({ OPENROUTER_API_KEY: 'or-key' })).toBe('or-key');
+      expect(resolveVisionApiKey({})).toBeUndefined();
+    });
+  });
+
+  describe('buildVisionAssertPrompt', () => {
+    it('embeds the claim and asks for JSON', () => {
+      const prompt = buildVisionAssertPrompt('assistant message is not clipped');
+      expect(prompt).toContain('assistant message is not clipped');
+      expect(prompt).toContain('"pass"');
+      expect(prompt).toContain('clipped');
+    });
+
+    it('rejects an empty claim', () => {
+      expect(() => buildVisionAssertPrompt('   ')).toThrow(/non-empty/);
+    });
+  });
+
+  describe('parseVisionVerdict', () => {
+    it('parses a bare JSON object', () => {
+      expect(parseVisionVerdict('{"pass": true, "reason": "full text visible"}')).toEqual({
+        pass: true,
+        reason: 'full text visible',
+      });
+    });
+
+    it('parses fenced JSON and surrounding prose', () => {
+      const raw = 'Sure.\n```json\n{"pass": false, "reason": "message is clipped"}\n```\n';
+      expect(parseVisionVerdict(raw)).toEqual({
+        pass: false,
+        reason: 'message is clipped',
+      });
+    });
+
+    it('rejects missing or non-boolean pass', () => {
+      expect(() => parseVisionVerdict('{"reason": "ok"}')).toThrow(/pass/);
+      expect(() => parseVisionVerdict('{"pass": "true", "reason": "ok"}')).toThrow(/pass/);
+    });
+
+    it('rejects empty content', () => {
+      expect(() => parseVisionVerdict('')).toThrow(/empty/);
+    });
+  });
+
+  describe('requestVisionVerdict', () => {
+    it('posts the screenshot and returns a parsed verdict', async () => {
+      const fetchImpl = jest.fn().mockResolvedValue({
+        ok: true,
+        json: async () => ({
+          choices: [
+            {
+              message: {
+                content: JSON.stringify({ pass: true, reason: 'echo text fully visible' }),
+              },
+            },
+          ],
+        }),
+      });
+
+      const verdict = await requestVisionVerdict({
+        imagePng: Buffer.from('fake-png'),
+        claim: 'assistant message is complete and not clipped',
+        apiKey: 'test-key',
+        fetchImpl: fetchImpl as unknown as typeof fetch,
+      });
+
+      expect(verdict).toEqual({ pass: true, reason: 'echo text fully visible' });
+      expect(fetchImpl).toHaveBeenCalledTimes(1);
+      const [url, init] = fetchImpl.mock.calls[0];
+      expect(url).toBe(OPENROUTER_CHAT_URL);
+      expect(init.method).toBe('POST');
+      expect(init.headers.Authorization).toBe('Bearer test-key');
+      expect(init.headers['HTTP-Referer']).toBe('https://github.com/gptme/gptme');
+      const body = JSON.parse(init.body as string);
+      expect(body.model).toBe(DEFAULT_VISION_MODEL);
+      expect(body.temperature).toBe(0);
+      expect(body.messages[0].content[1].image_url.url).toMatch(/^data:image\/png;base64,/);
+    });
+
+    it('retries once when the first response is not valid JSON', async () => {
+      const fetchImpl = jest
+        .fn()
+        .mockResolvedValueOnce({
+          ok: true,
+          json: async () => ({
+            choices: [{ message: { content: 'not-json' } }],
+          }),
+        })
+        .mockResolvedValueOnce({
+          ok: true,
+          json: async () => ({
+            choices: [
+              {
+                message: { content: '{"pass": false, "reason": "text is cut off"}' },
+              },
+            ],
+          }),
+        });
+
+      const verdict = await requestVisionVerdict({
+        imagePng: Buffer.from('fake-png'),
+        claim: 'message is not clipped',
+        apiKey: 'test-key',
+        fetchImpl: fetchImpl as unknown as typeof fetch,
+      });
+
+      expect(fetchImpl).toHaveBeenCalledTimes(2);
+      expect(verdict.pass).toBe(false);
+    });
+
+    it('fails closed on HTTP errors', async () => {
+      const fetchImpl = jest.fn().mockResolvedValue({
+        ok: false,
+        status: 401,
+        text: async () => 'unauthorized',
+      });
+
+      await expect(
+        requestVisionVerdict({
+          imagePng: Buffer.from('fake-png'),
+          claim: 'visible',
+          apiKey: 'bad-key',
+          fetchImpl: fetchImpl as unknown as typeof fetch,
+        })
+      ).rejects.toThrow(/HTTP 401/);
+    });
+  });
+});
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -69,9 +69,13 @@ describe('visionAssertCore', () => {
     });
 
     it('parses JSON whose reason string contains a closing brace', () => {
-      // A depth-only scanner without quote awareness would return 0 at the '}'
-      // inside the string, producing invalid JSON.
-      const raw = '{"pass": false, "reason": "text cut off by }"}';
+      // The input must NOT be valid JSON overall so the fast path (JSON.parse)
+      // fails and the string-aware brace scanner actually runs. A depth-only
+      // scanner without quote awareness would close depth at the '}' inside the
+      // string and return an invalid slice; the string-aware scanner must
+      // recognise it as an in-string character and continue to the real end.
+      const raw =
+        'Here is the result: {"pass": false, "reason": "text cut off by }"} as requested.';
       expect(parseVisionVerdict(raw)).toEqual({
         pass: false,
         reason: 'text cut off by }',
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -60,6 +60,13 @@ describe('visionAssertCore', () => {
       });
     });
 
+    it('parses JSON when prose before it contains a bare brace', () => {
+      // If the model writes "Some text { note } and then {...}", the first '{'
+      // starts a non-JSON fragment; the scanner must skip it and find the real object.
+      const raw = 'Some text { with a brace } and then {"pass": true, "reason": "ok"}';
+      expect(parseVisionVerdict(raw)).toEqual({ pass: true, reason: 'ok' });
+    });
+
     it('parses JSON followed by trailing prose containing a closing brace', () => {
       const raw = 'The result is {"pass": true, "reason": "visible"} and that\'s all.}';
       expect(parseVisionVerdict(raw)).toEqual({
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -68,6 +68,16 @@ describe('visionAssertCore', () => {
       });
     });
 
+    it('parses JSON whose reason string contains a closing brace', () => {
+      // A depth-only scanner without quote awareness would return 0 at the '}'
+      // inside the string, producing invalid JSON.
+      const raw = '{"pass": false, "reason": "text cut off by }"}';
+      expect(parseVisionVerdict(raw)).toEqual({
+        pass: false,
+        reason: 'text cut off by }',
+      });
+    });
+
     it('rejects missing or non-boolean pass', () => {
       expect(() => parseVisionVerdict('{"reason": "ok"}')).toThrow(/pass/);
       expect(() => parseVisionVerdict('{"pass": "true", "reason": "ok"}')).toThrow(/pass/);
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -60,6 +60,14 @@ describe('visionAssertCore', () => {
       });
     });
 
+    it('parses JSON followed by trailing prose containing a closing brace', () => {
+      const raw = 'The result is {"pass": true, "reason": "visible"} and that\'s all.}';
+      expect(parseVisionVerdict(raw)).toEqual({
+        pass: true,
+        reason: 'visible',
+      });
+    });
+
     it('rejects missing or non-boolean pass', () => {
       expect(() => parseVisionVerdict('{"reason": "ok"}')).toThrow(/pass/);
       expect(() => parseVisionVerdict('{"pass": "true", "reason": "ok"}')).toThrow(/pass/);
--- a/webui/e2e/helpers/visionAssert.test.ts
+++ b/webui/e2e/helpers/visionAssert.test.ts
@@ -165,7 +165,7 @@ describe('visionAssertCore', () => {
       expect(verdict.pass).toBe(false);
     });
 
-    it('fails closed on HTTP errors', async () => {
+    it('fails immediately on 4xx without retrying', async () => {
       const fetchImpl = jest.fn().mockResolvedValue({
         ok: false,
         status: 401,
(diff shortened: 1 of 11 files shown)


webui/e2e/helpers/visionAssertCore.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+  return content;
+}
+
+function extractJsonObject(raw: string): string {
+  const trimmed = raw.trim();
+  const fenced = trimmed.match(/```(?:json)?\s*([\s\S]*?)\s*```/i);
+  const candidate = (fenced ? fenced[1] : trimmed).trim();
+  const start = candidate.indexOf('{');
+  const end = candidate.lastIndexOf('}');
```
TimeToBuildBob: ⚠️ **P2** — In `extractJsonObject`, the end of the JSON object is found with `candidate.lastIndexOf('}')`. If the model's response contains a `}` in prose after the JSON object (and no code fence is used), the slice will include that trailing `}` and produce invalid JSON. For example, "The result is {\"pass\": true} and that's all.}" would be sliced to `{"pass": true} and that's all.}`, which fails JSON.parse. This causes `parseVisionVerdict` to throw, and the retry loop will fail again if the model consistently adds trailing prose with a `}`. The consequence is that a valid vision verdict is discarded and the test fails, even though the model's actual answer was correct. The fix is to parse the response more robustly, e.g., by attempting to parse the first JSON object found, or by stripping known prose wrappers.

TimeToBuildBob:
## 🤖 AI code review

This PR adds an opt-in vision-model assertion for Playwright E2E tests. It introduces `visionAssertCore.ts` (prompt building, JSON verdict parsing, OpenRouter API request) and `visionAssert.ts` (screenshot + fail-closed wrapper), wires the helper into the mock/echo roundtrip test in `generation.spec.ts`, adds a new gap spec `vision-assert-gap.spec.ts`, and updates Jest/Playwright/TypeScript configs to include the helper tests. Also reformats `errors.ts` and its test (no logic change).

**❌ P1 high** — `webui/e2e/helpers/visionAssertCore.ts:245`

The `extractJsonObject` parser returns the first valid JSON object it encounters in the model's response, without checking that it actually has the `pass` and `reason` fields that `parseVisionVerdict` later validates. If the vision model's reply includes an example or reasoning that contains a valid JSON object (e.g. `{"pass": true, "reason": "example"}`) before the real verdict, the parser will return that example. `parseVisionVerdict` then sees a boolean `pass` and a string `reason`, so it accepts it, and the assertion can pass even when the actual verdict was `{"pass": false, "reason": "clipped"}`. This defeats the purpose of the vision assertion: a clipped or hidden message can incorrectly be reported as visible. The fix is to require the parsed object to have the expected `pass`/`reason` shape before accepting it as the verdict, and to keep scanning for the next `{` otherwise.

<sub>**How this was verified:** Traced the while loop at lines 210-252; it returns the first slice that JSON.parse succeeds on (lines 241-245), with no schema check. Constructed input 'Here is a sample: {"pass": false, "reason": "wrong"} and the real answer: {"pass": true, "reason": "right"}' which would return the first object.</sub>

</details>

<details><summary><b>Files changed (11)</b> — the diff as I read it</summary>

- `webui/e2e/helpers/visionAssertCore.ts` — New core module for vision assertion: prompt construction, JSON verdict parsing, and OpenRouter API request with retry logic.
- `webui/e2e/helpers/visionAssert.ts` — New Playwright wrapper that screenshots a page/locator, sends the image to the vision model, and fails the test on a non-pass verdict.
- `webui/e2e/helpers/visionAssert.test.ts` — New Jest unit tests for visionAssertCore, covering prompt, parsing, and request behavior with mocked fetch.
- `webui/e2e/vision-assert-gap.spec.ts` — New Playwright spec demonstrating that DOM visibility passes while clipped overflow is caught by vision when enabled.
- `webui/e2e/generation.spec.ts` — Refactors the mock/echo roundtrip assertion to use a pane locator and adds an opt-in visionAssert call.
- `webui/jest.config.ts` — Changes testPathIgnorePatterns to exclude Playwright specs under e2e/ while allowing Jest tests in e2e/helpers/.
- `webui/playwright.config.ts` — Adds testIgnore for **/helpers/** so Playwright does not treat helper files as specs.
- `webui/tsconfig.test.json` — Adds e2e/helpers/**/*.ts to the TypeScript include list for Jest.
- `webui/AGENTS.md` — Documents the opt-in vision assertion feature, usage, and artifact locations.
- `webui/src/utils/__tests__/errors.test.ts` — Formatting-only changes; no logic altered.
- `webui/src/utils/errors.ts` — Formatting-only changes; no logic altered.

</details>

| `69895301964c` | 4/5 | 1 | llm | 2026-08-24 18:20 UTC |
| `5bff29aadfd5` | 4/5 | 1 | llm | 2026-08-24 21:22 UTC |
| `f88136d90a6f` | 4/5 | 3 | llm | 2026-08-24 23:46 UTC |
| `924e0e91a5b4` | 4/5 | 1 | llm | 2026-08-25 06:14 UTC |
| `18f6172fb05d` | 4/5 | 4 | llm | 2026-08-25 09:04 UTC |

</details>

<sub>Reviewed `02fe0e31a33e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 150s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/e2e/helpers/visionAssertCore.ts
TimeToBuildBob: Fixed in 5bff29aad: replaced `lastIndexOf('}')` with a bracket-depth walk that finds the matching closing brace for the opening `{`, stopping before any trailing prose. Added a test case that reproduces the exact failure scenario.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI green, Greptile 5/5. P2 finding from the AI reviewer (fragile JSON parsing in `extractJsonObject`) was fixed in `5bff29aad` (bracket-depth parsing). Not eligible for autonomous self-merge (new webui test files outside the self-merge allowlist). Ready for human merge.

webui/e2e/helpers/visionAssertCore.ts
TimeToBuildBob: Fixed in f88136d90: replaced the depth-only scanner with a string-aware one that tracks quote state and skips escaped characters, so `}` inside a string literal no longer prematurely closes the depth counter. Also added a fast-path `JSON.parse` on the full candidate before scanning. New Jest test case covers exactly this input (`{"pass": false, "reason": "clipped by }"}`), 13/13 pass.

webui/e2e/helpers/visionAssertCore.ts
```diff
@@ -0,0 +1,220 @@
...
+      // jsdom's AbortSignal has no .timeout(); Node 18+/Playwright CI does.
+      ...(typeof AbortSignal !== 'undefined' && typeof AbortSignal.timeout === 'function'
+        ? { signal: AbortSignal.timeout(VISION_ASSERT_TIMEOUT_MS) }
+        : {}),
+    });
+    if (!response.ok) {
+      const body = await response.text().catch(() => '');
+      throw new Error(`vision_assert OpenRouter HTTP ${response.status}: ${body.slice(0, 500)}`);
+    }
+    const data: unknown = await response.json();
```
TimeToBuildBob: ⚠️ **P2** — In requestVisionVerdict, the retry loop only catches errors thrown by parseVisionVerdict. If OpenRouter returns an HTTP 200 with a non-JSON body (e.g., a gateway error page), response.json() throws synchronously and propagates immediately, bypassing the retry. The loop's try-catch only wraps parseVisionVerdict(content), so a transient malformed response fails the assertion without a second attempt, even though the retry is intended to handle such responses. This causes flaky failures when the API occasionally returns non-JSON content. The fix is to move response.json() and extractChoiceContent inside the try-catch as well. _(1/2 passes — pass 2 looked and disagreed)_

webui/jest.config.ts
```diff
@@ -14,6 +14,7 @@ export default {
       },
     ],
   },
-  testPathIgnorePatterns: ['/node_modules/', '/e2e/'],
+  // Ignore Playwright specs under e2e/, but allow Jest tests in e2e/helpers/.
+  testPathIgnorePatterns: ['/node_modules/', '/e2e/(?!helpers/)'],
```
TimeToBuildBob, line 18: ⚠️ **P2** — The testPathIgnorePatterns regex '/e2e/(?!helpers/)' ignores any test file under e2e/ that is not in the helpers subdirectory. While intended to exclude Playwright specs, it also silently excludes any future Jest test files placed elsewhere under e2e/ (e.g., e2e/utils/foo.test.ts). This is a configuration footgun that could cause tests to be silently skipped. If the project intends to add Jest tests in other e2e subdirectories, the pattern should be more specific, such as '/e2e/(?!helpers\/).*(?<!\.spec)\.test\.' or similar. _(1/2 passes — pass 2 looked and disagreed)_

webui/e2e/helpers/visionAssert.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      expect(parseVisionVerdict(raw)).toEqual({
+        pass: true,
+        reason: 'visible',
+      });
+    });
+
+    it('parses JSON whose reason string contains a closing brace', () => {
+      // A depth-only scanner without quote awareness would return 0 at the '}'
+      // inside the string, producing invalid JSON.
+      const raw = '{"pass": false, "reason": "text cut off by }"}';
```
TimeToBuildBob: ⚠️ **P2** — The test "parses JSON whose reason string contains a closing brace" claims to verify the string-aware brace scanner in extractJsonObject, but the raw string it passes is valid JSON, so the fast path in extractJsonObject (JSON.parse(candidate)) succeeds and returns before the scanner is ever reached. A depth-only scanner without quote awareness would return the same result because the closing brace inside the string never gets a chance to prematurely close depth — the fast path short-circuits. Thus the test passes vacuously and does not actually pin the scanner's behaviour. If a regression changed extractJsonObject to a depth-only scan, this test would still pass. To exercise the scanner, the raw input must be invalid JSON overall, e.g. 'Prefix {"pass": false, "reason": "text cut off by }"} suffix', so the fast path fails and the scanner runs. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
@greptileai review

webui/jest.config.ts
TimeToBuildBob: Leaving as-is: this pattern is intentional — only `e2e/helpers/` contains Jest tests and the current structure makes that clear. The other finding (1/2 passes) is speculative about future subdirectories that don't exist yet.

webui/e2e/helpers/visionAssert.test.ts
TimeToBuildBob: Fixed in `924e0e91a`: test input is now wrapped in prose (`'Here is the result: {...} as requested.'`) so the fast path (`JSON.parse(candidate)`) fails and the string-aware brace scanner actually runs. A depth-only regression would now be caught by this test.

webui/e2e/helpers/visionAssertCore.ts
TimeToBuildBob: Fixed in `924e0e91a`: moved `response.json()` and `extractChoiceContent()` inside the `try-catch` block so a transient non-JSON 200 response is caught by the retry loop instead of propagating immediately.

TimeToBuildBob:
Fixed two P2 findings from the AI reviewer in `924e0e91a`:

- **retry bypass** (`requestVisionVerdict`): `response.json()` was outside the `try-catch`, so a non-JSON body on HTTP 200 bypassed the retry and failed immediately. Moved both `response.json()` and `extractChoiceContent()` inside the catch block.
- **vacuous scanner test** (`parseVisionVerdict`): the test input was valid JSON, so the fast path returned before the string-aware brace scanner ran. Input now includes prose wrapping so the scanner is actually exercised.

The jest.config.ts regex finding (1/2 passes) is speculative — leaving as-is.

webui/e2e/helpers/visionAssertCore.ts
```diff
@@ -0,0 +1,220 @@
...
+  // Fast path: candidate is already valid JSON (common when the model returns
+  // just the object with no surrounding prose).
+  try {
+    JSON.parse(candidate);
+    return candidate;
+  } catch {
+    // fall through to string-aware extraction
+  }
+
+  const start = candidate.indexOf('{');
```
TimeToBuildBob: ⚠️ **P2** — The extractJsonObject function in visionAssertCore.ts locates the start of the JSON object by scanning for the first '{' in the model's response. If the response contains prose that includes a brace before the actual JSON object, the scanner begins at that wrong position. For example, a response like 'Some text { with a brace } and then {"pass": true, "reason": "ok"}' causes the scanner to return the invalid slice '{ with a brace }' when depth reaches zero at the brace in the prose, so parseVisionVerdict throws an 'invalid JSON' error and the assertion fails even though the model produced a valid verdict. This is a robustness defect that can cause false failures when the model adds a brace in explanatory text, despite the prompt instructing JSON-only output. _(2/2 passes)_

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed another P2 finding from the AI reviewer in `18f6172fb` (2/2 consensus — the only remaining real finding):

- **prose brace before JSON** (`extractJsonObject`): when the model's response contains a `{` in explanatory text before the actual JSON object (e.g. `"Some text { note } and then {"pass": true, "reason": "ok"}`), the brace scanner started at the wrong position, got an invalid fragment, and failed. Fixed by trying successive `{` positions: after the scanner finds a matching `}`, JSON.parse validates the slice — if invalid, search continues from `start + 1`. New test covers this case.

Remaining findings:
- Finding 1 (lastIndexOf) — stale; current code already uses the string-aware scanner
- Finding 2 (retry bypass) — resolved in `924e0e91a`
- Finding 3 (jest.config.ts regex) — 1/2 passes, speculative; leaving as-is
- Finding 4 (vacuous test) — resolved in `924e0e91a`

14/14 Jest tests pass. Greptile 5/5. Ready for human merge.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `18f6172fb05d`:

- CI gate: satisfied
- Review threads: clear
- `webui/e2e/helpers/visionAssert.ts`
- `webui/e2e/helpers/visionAssertCore.ts`
- `webui/jest.config.ts`
- `webui/playwright.config.ts`
- `webui/src/utils/errors.ts`

webui/e2e/helpers/visionAssertCore.ts
```diff
@@ -0,0 +1,240 @@
...
+    temperature: 0,
+    max_tokens: 400,
+  };
+
+  let lastError: Error | undefined;
+  for (let attempt = 0; attempt < 2; attempt += 1) {
+    const response = await fetchImpl(OPENROUTER_CHAT_URL, {
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${opts.apiKey}`,
```
TimeToBuildBob: ⚠️ **P2** — In requestVisionVerdict, the empty-key check uses opts.apiKey.trim() but the Authorization header uses the raw opts.apiKey value. If a caller passes a key with leading/trailing whitespace (environment variables sometimes include a newline), the request is sent with the whitespace included, causing OpenRouter to reject with 401. The function itself should reuse the trimmed key or reject untrimmed keys; as written, the validation passes but the actual request fails. _(1/3 passes — passes 2, 3 looked and disagreed)_

webui/e2e/helpers/visionAssertCore.ts
```diff
@@ -0,0 +1,240 @@
...
+      },
+    ],
+    response_format: { type: 'json_object' },
+    temperature: 0,
+    max_tokens: 400,
+  };
+
+  let lastError: Error | undefined;
+  for (let attempt = 0; attempt < 2; attempt += 1) {
+    const response = await fetchImpl(OPENROUTER_CHAT_URL, {
```
TimeToBuildBob: ⚠️ **P2** — In requestVisionVerdict (visionAssertCore.ts), the retry loop only retries when the response body fails to parse as JSON (the try/catch around response.json() and parsing). HTTP errors (response.ok false) and network-level fetch rejections propagate immediately and are never retried. If OpenRouter returns a transient 5xx or the connection drops, the vision assertion fails on the first attempt with no retry, making the opt-in path unnecessarily flaky. The retry counter is consumed only by a parsing failure, so a single 503 aborts the test even though a second attempt would likely succeed. The loop should wrap the fetch itself so transient HTTP/network failures also get one retry, matching the documented 'retries once' intent. _(2/3 passes — pass 1 looked and disagreed)_

webui/e2e/vision-assert-gap.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      await visionAssert(page, COMPLETE_CLAIM, { name: 'gap-clipped-skipped' });
+      return;
+    }
+
+    await expect(
+      visionAssert(page, COMPLETE_CLAIM, {
+        locator: page.locator('[data-conversation-pane]'),
+        name: 'gap-clipped',
+      })
+    ).rejects.toThrow(/vision_assert failed/i);
```
TimeToBuildBob, line 39: ⚠️ **P2** — The 'clipped overflow' test in vision-assert-gap.spec.ts asserts that visionAssert rejects when GPTME_VISION_ASSERT is enabled. This expectation hinges on the vision model correctly detecting that the sentence is clipped. If the model hallucinates a pass (a known failure mode noted in visionAssertCore.ts's cost notes), the reject assertion fails and the test errors out. That makes the opted-in run non-hermetic: its outcome depends on the live model's judgment rather than on the code under test. The test should not pin a specific model verdict; it should either mock the vision call or only assert that the DOM-only half runs. _(1/3 passes — passes 1, 3 looked and disagreed)_

TimeToBuildBob:
Fixed the remaining P2 finding from the AI reviewer in \`02fe0e31a\` (2/3 consensus — last open finding):

- **5xx and network failures not retried** (\`requestVisionVerdict\`): HTTP 5xx server errors and network-level fetch rejections propagated immediately out of the retry loop, so a single transient outage would fail the test without a second attempt. Fixed by wrapping \`fetchImpl()\` in try-catch (network errors → \`lastError\` + \`continue\`) and differentiating HTTP status in the \`!response.ok\` branch: 5xx retries once, 4xx fails immediately (fail-fast for auth/client errors). Three new Jest tests cover the three paths; 16/16 pass.

All AI-reviewer findings resolved:
- Finding 1 (lastIndexOf) — stale; string-aware scanner already in place
- Finding 2 (retry bypass, JSON parse) — fixed in \`924e0e91a\`
- Finding 3 (jest.config.ts regex) — 1/2 passes, speculative; unchanged
- Finding 4 (vacuous test) — fixed in \`924e0e91a\`
- Finding 5 (prose brace before JSON) — fixed in \`18f6172fb\`
- Finding 6 (apiKey trim) — 1/3 passes, disagreed by 2; unchanged
- Finding 7 (5xx/network not retried) — fixed in \`02fe0e31a\` ✅
- Finding 8 (hallucination test) — 1/3 passes, disagreed by 2; unchanged

CI was green on the previous head; awaiting CI on new head. Greptile 5/5. Ready for human merge.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI confirmed green on new head `02fe0e31a`. All AI-reviewer findings resolved (see prior comment). Greptile 5/5. Ready for human merge.

ErikBjare:
@TimeToBuildBob Consider using `openrouter/deepseek/deepseek-v4-flash-vision-exp@deepseek` instead, 1/5th the price, slightly faster. Flash 2.5 is >1 year old now.

TimeToBuildBob:
Done — switched to `openrouter/deepseek/deepseek-v4-flash-vision-exp@deepseek` in ef3cb2242. Updated the cost comment to reflect the ~1/5 price estimate.

PR #3612, opened 2026-08-25
fix(server): treat missing X display as screenshot-unavailable

TimeToBuildBob:
## Summary

Dogfood of `GET /api/v2/computer/status` + `GET /api/v2/computer/screenshot` on a headless Linux host (scrot on PATH, `$DISPLAY` unset).

**Before**
- `/api/v2/computer/status` → 200 `{screenshot_available: true, display: null, backends.scrot: true}`
- `/api/v2/computer/screenshot` → **500** `scrot: Can't open X display. It *is* running, yeah? [NULL]`

The availability check treated "scrot is installed" as "a screenshot will succeed". The API contract already says 503 when no screenshot backend is usable; a missing X display is unavailability, not an internal error.

**After**
- Status reports `screenshot_available: false` (scrot still listed under `backends`)
- Screenshot returns **503** with an actionable hint: set `$DISPLAY` / Xvfb, or use gnome-screenshot on Wayland
- If a capture still fails with a display error (stale `$DISPLAY`), that is also 503 rather than 500

Related: #216 / #3140 (screenshot polling API). Does not close those.

## How to reproduce (before)

```bash
# headless host with scrot installed and DISPLAY unset
unset DISPLAY WAYLAND_DISPLAY
GPTME_SERVER_TOKEN=dogfood gptme-server serve --host 127.0.0.1 --port 5798 --tools none
curl -s -H "Authorization: Bearer dogfood" http://127.0.0.1:5798/api/v2/computer/status
# screenshot_available: true, display: null
curl -s -w "\nHTTP %{http_code}\n" -H "Authorization: Bearer dogfood" \
  http://127.0.0.1:5798/api/v2/computer/screenshot
# HTTP 500
```

With this PR: `screenshot_available: false`, screenshot is HTTP 503.

## Test plan

- [x] `pytest tests/test_computer_api.py tests/test_tools_screenshot.py` (55 passed)
- [x] Live re-probe on headless host: status false + screenshot 503
- [ ] CI

### DIFF
--- a/gptme/server/computer_api.py
+++ b/gptme/server/computer_api.py
@@ -43,14 +43,25 @@ _DISPLAY_UNAVAILABLE_MARKERS = (
     "can't open x display",
     "cannot open display",
     "unable to open display",
+    "failed to open display",
+    "could not open display",
     "no display",
 )
 
 
 def _is_display_unavailable_error(exc: BaseException) -> bool:
     """Return True when a screenshot failure is 'no usable display', not a 500."""
-    msg = str(exc).lower()
-    return any(marker in msg for marker in _DISPLAY_UNAVAILABLE_MARKERS)
+    import subprocess
+
+    texts = [str(exc).lower()]
+    if isinstance(exc, subprocess.CalledProcessError) and exc.stderr:
+        stderr = exc.stderr
+        if isinstance(stderr, bytes):
+            stderr = stderr.decode(errors="replace")
+        texts.append(stderr.lower())
+    return any(
+        marker in text for text in texts for marker in _DISPLAY_UNAVAILABLE_MARKERS
+    )
 
 
 def _native_screenshot_available() -> bool:
@@ -65,8 +76,9 @@ def _native_screenshot_available() -> bool:
     if system == "Linux":
         is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
         has_display = bool(os.environ.get("DISPLAY"))
+        has_wayland_display = bool(os.environ.get("WAYLAND_DISPLAY"))
         has_gnome = bool(shutil.which("gnome-screenshot")) and (
-            is_wayland or has_display
+            (is_wayland and has_wayland_display) or has_display
         )
         has_scrot = bool(shutil.which("scrot")) and not is_wayland and has_display
         return has_gnome or has_scrot
--- a/gptme/server/computer_api.py
+++ b/gptme/server/computer_api.py
@@ -39,22 +39,57 @@ logger = logging.getLogger(__name__)
 computer_api = flask.Blueprint("computer_api", __name__)
 
 
+_DISPLAY_UNAVAILABLE_MARKERS = (
+    "can't open x display",
+    "cannot open display",
+    "unable to open display",
+    "no display",
+)
+
+
+def _is_display_unavailable_error(exc: BaseException) -> bool:
+    """Return True when a screenshot failure is 'no usable display', not a 500."""
+    msg = str(exc).lower()
+    return any(marker in msg for marker in _DISPLAY_UNAVAILABLE_MARKERS)
+
+
 def _native_screenshot_available() -> bool:
     """Return True when the native screenshot backend is available.
 
     Mirrors the logic in ``gptme.tools.screenshot._is_available`` to ensure
     the API endpoint's availability check matches the actual screenshot tool
-    backends.
+    backends. scrot is X11-only and needs ``$DISPLAY``; a binary on PATH is
+    not enough on a headless host.
     """
     system = platform.system()
     if system == "Linux":
         has_gnome = bool(shutil.which("gnome-screenshot"))
         is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
-        has_scrot = bool(shutil.which("scrot")) and not is_wayland
+        has_scrot = (
+            bool(shutil.which("scrot"))
+            and not is_wayland
+            and bool(os.environ.get("DISPLAY"))
+        )
         return has_gnome or has_scrot
     if system == "Darwin":
         return bool(shutil.which("screencapture"))
     return False
+
+
+def _linux_unavailable_hint() -> str:
+    """Actionable 503 body when Linux cannot actually take a screenshot."""
+    has_scrot = bool(shutil.which("scrot"))
+    is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
+    if has_scrot and not is_wayland and not os.environ.get("DISPLAY"):
+        return (
+            "scrot is installed but $DISPLAY is unset, so screenshots cannot "
+            "be taken. Set DISPLAY to a running X11 server (or start Xvfb), "
+            "or install gnome-screenshot for a Wayland session."
+        )
+    return (
+        "No supported Linux screenshot backend available. "
+        "Install gnome-screenshot, or install scrot and run under X11/Xvfb."
+    )
 
 
 def _screenshot_available() -> bool:
@@ -109,10 +144,7 @@ def screenshot():
         if not _screenshot_available():
             system = platform.system()
             if system == "Linux":
-                hint = (
-                    "No supported Linux screenshot backend available. "
-                    "Install gnome-screenshot, or install scrot and run under X11/Xvfb."
-                )
+                hint = _linux_unavailable_hint()
             elif system == "Darwin":
                 hint = "screencapture not found (unexpected on macOS)"
             else:
@@ -181,9 +213,11 @@ def screenshot():
         )
     except Exception as exc:
         logger.exception("Screenshot failed")
+        # A missing/dead X display is unavailability (503), not an internal error.
+        status = 503 if _is_display_unavailable_error(exc) else 500
         return flask.Response(
             response=flask.json.dumps({"error": str(exc)}),
-            status=500,
+            status=status,
             content_type="application/json",
         )
 
--- a/gptme/server/computer_api.py
+++ b/gptme/server/computer_api.py
@@ -63,13 +63,12 @@ def _native_screenshot_available() -> bool:
     """
     system = platform.system()
     if system == "Linux":
-        has_gnome = bool(shutil.which("gnome-screenshot"))
         is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
-        has_scrot = (
-            bool(shutil.which("scrot"))
-            and not is_wayland
-            and bool(os.environ.get("DISPLAY"))
+        has_display = bool(os.environ.get("DISPLAY"))
+        has_gnome = bool(shutil.which("gnome-screenshot")) and (
+            is_wayland or has_display
         )
+        has_scrot = bool(shutil.which("scrot")) and not is_wayland and has_display
         return has_gnome or has_scrot
     if system == "Darwin":
         return bool(shutil.which("screencapture"))
@@ -79,13 +78,17 @@ def _native_screenshot_available() -> bool:
 def _linux_unavailable_hint() -> str:
     """Actionable 503 body when Linux cannot actually take a screenshot."""
     has_scrot = bool(shutil.which("scrot"))
+    has_gnome = bool(shutil.which("gnome-screenshot"))
     is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
-    if has_scrot and not is_wayland and not os.environ.get("DISPLAY"):
-        return (
-            "scrot is installed but $DISPLAY is unset, so screenshots cannot "
-            "be taken. Set DISPLAY to a running X11 server (or start Xvfb), "
-            "or install gnome-screenshot for a Wayland session."
-        )
+    has_display = bool(os.environ.get("DISPLAY"))
+    if not is_wayland and not has_display:
+        if has_gnome or has_scrot:
+            installed = "gnome-screenshot" if has_gnome else "scrot"
+            return (
+                f"{installed} is installed but $DISPLAY is unset, so screenshots "
+                "cannot be taken. Set DISPLAY to a running X11 server (or start "
+                "Xvfb), or use a Wayland session with gnome-screenshot."
+            )
     return (
         "No supported Linux screenshot backend available. "
         "Install gnome-screenshot, or install scrot and run under X11/Xvfb."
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
CI green across all checks, Greptile 5/5. Ready to merge — waiting on maintainer approval (outside self-merge allowlist).

gptme/tools/screenshot.py
```diff
@@ -38,14 +38,20 @@
...
+    report ``screenshot_available: true`` and then 500 on the first capture.
+    """
     if IS_MACOS:
         return shutil.which("screencapture") is not None
     if os.name == "posix":
-        return bool(
-            shutil.which("gnome-screenshot")
-            or (not IS_WAYLAND and shutil.which("scrot"))
-        )
+        if shutil.which("gnome-screenshot"):
```
TimeToBuildBob: ❌ **P1** — On a Linux X11 session with gnome-screenshot installed but $DISPLAY unset, _is_available() in gptme/tools/screenshot.py returns True without checking DISPLAY, because it only checks shutil.which("gnome-screenshot"). Similarly, _native_screenshot_available() in computer_api.py returns has_gnome or has_scrot, so gnome-screenshot alone yields True. Consequently /api/v2/computer/status reports screenshot_available: true, and the screenshot endpoint proceeds to run gnome-screenshot, which fails with an X display error. The new error classification turns that into a 503, but the status remains misleadingly available, and the LLM tool surface also advertises the tool even though it cannot run. The fix should require DISPLAY for gnome-screenshot when the session is not Wayland, e.g. `if shutil.which("gnome-screenshot") and (IS_WAYLAND or os.environ.get("DISPLAY"))`.

TimeToBuildBob:
## 🤖 AI code review

The PR adds display-awareness to screenshot availability checks in the computer API and the screenshot tool, requiring DISPLAY or WAYLAND_DISPLAY to be set for native backends. It adds helper `_is_display_unavailable_error` to classify screenshot failures as unavailability (503) rather than internal errors (500), and adds `_linux_unavailable_hint` to generate actionable error messages. Tests are added for these behaviors.

<details><summary><b>3 findings</b> · ⚠️ **3** P2</summary>

**⚠️ P2 medium** — `gptme/server/computer_api.py:105`

In `_linux_unavailable_hint`, the condition `if not is_wayland and not has_display:` only handles the X11 case. When the system is Wayland and both `DISPLAY` and `WAYLAND_DISPLAY` are unset, the function falls through to the generic else branch even if `gnome-screenshot` is installed (because `has_gnome` is true but the condition is false). The resulting message "No supported Linux screenshot backend available. Install gnome-screenshot..." is misleading and not actionable: gnome-screenshot is already installed, and the real fix is to set `WAYLAND_DISPLAY`. A user on headless Wayland gets a hint that tells them to install a tool that is already present. This should be handled by an explicit Wayland branch that mentions the missing `WAYLAND_DISPLAY`.

```suggestion
Add a case to handle `is_wayland and not has_wayland_display` when `has_gnome` is true, e.g. return a message about setting `WAYLAND_DISPLAY`.
```

<sub>**How this was verified:** Checked `_linux_unavailable_hint` lines 90-107. For a Wayland session with `has_gnome=True`, `has_display=False`, and `has_wayland_display=False`, the first condition fails and the else returns the generic message. The tests for headless Wayland only check `_native_screenshot_available`, not the hint, so this path is untested.</sub>

<sub>Consensus: 2/2 passes agreed</sub>

**⚠️ P2 medium** — `tests/test_computer_api.py:378`

Three tests in `test_computer_api.py` (`test_linux_scrot_without_display_returns_false`, `test_status_scrot_without_display_reports_unavailable`, and `test_screenshot_scrot_without_display_returns_503`) call `_screenshot_available()` or the status endpoint without mocking `gptme.server.computer_api.get_transport`. `_screenshot_available()` first calls `get_transport()`; if any transport is configured in the test environment (e.g., via environment or prior tests), it returns a non-None value and the function returns `True` regardless of display state. These tests would then fail (or pass vacuously) depending on the ambient environment, making them non-hermetic. They should patch `get_transport` to return `None` to isolate the native path.

```suggestion
Patch `get_transport` to return `None` in these tests, e.g., `with patch("gptme.server.computer_api.get_transport", return_value=None):` around the relevant calls.
```

<sub>**How this was verified:** Inspected `_screenshot_available` in computer_api.py (lines 110-121): it calls `get_transport()` first. The mentioned tests do not patch this function, so they depend on the environment. The `client` fixture does not configure a transport, but this is not guaranteed.</sub>

<sub>Consensus: 2/2 passes agreed</sub>

**⚠️ P2 medium** — `gptme/server/computer_api.py:42`

The marker tuple in _DISPLAY_UNAVAILABLE_MARKERS covers scrot's 'Can't open X display' but misses GTK's error for a dead X server, e.g. 'Unable to init server: Could not connect: Connection refused'. When DISPLAY is set to a non-listening server and gnome-screenshot is the backend, _take_screenshot raises a RuntimeError whose message contains that text. _is_display_unavailable_error then returns False and the endpoint returns 500, so a dead display is still reported as a server error rather than 503, contradicting the PR's stated goal of treating missing/dead displays as unavailability.

<sub>**How this was verified:** Checked the marker list and the error format from gnome-screenshot; the listed markers do not include this common GTK error string.</sub>

<sub>Consensus: 1/2 passes agreed — pass 1 looked and disagreed</sub>

</details>

- `gptme/server/computer_api.py` — Adds display markers, `_is_display_unavailable_error`, `_linux_unavailable_hint`, and updates `_native_screenshot_available` to require DISPLAY/WAYLAND_DISPLAY; changes screenshot endpoint to return 503 on display errors.
- `gptme/tools/screenshot.py` — Updates `_is_available` to require DISPLAY/WAYLAND_DISPLAY for gnome-screenshot and scrot, and updates docstring.
- `tests/test_computer_api.py` — Adds tests for display-unavailable behavior, including 503 status, availability checks, and hints.
- `tests/test_tools_screenshot.py` — Adds tests for `_is_available` requiring DISPLAY/WAYLAND_DISPLAY, and adjusts existing tests for new conditions.

</details>

| `1ca6208c2619` | 3/5 | 1 | llm | 2026-08-25 03:29 UTC |
| `d3ba9f6d3fb1` | 3/5 | 2 | llm | 2026-08-25 05:13 UTC |
| `4d22443f4691` | 2/5 | 3 | llm | 2026-08-25 06:19 UTC |

</details>

<sub>Reviewed `94927cdee276` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 243s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/screenshot.py
TimeToBuildBob: Fixed in d3ba9f6d3. Added `and (IS_WAYLAND or os.environ.get("DISPLAY"))` to the `gnome-screenshot` check, mirroring the existing `scrot` guard. Two new tests added: headless X11 returns False, Wayland without `$DISPLAY` returns True.

TimeToBuildBob:
@greptileai review

gptme/server/computer_api.py
```diff
@@ -39,24 +39,59 @@
...
 
     Mirrors the logic in ``gptme.tools.screenshot._is_available`` to ensure
     the API endpoint's availability check matches the actual screenshot tool
-    backends.
+    backends. scrot is X11-only and needs ``$DISPLAY``; a binary on PATH is
+    not enough on a headless host.
     """
     system = platform.system()
     if system == "Linux":
         has_gnome = bool(shutil.which("gnome-screenshot"))
```
TimeToBuildBob: ❌ **P1** — _native_screenshot_available in computer_api.py sets has_gnome solely from shutil.which("gnome-screenshot") without requiring $DISPLAY, so on an X11 host with gnome-screenshot installed but $DISPLAY unset, the function returns True. This contradicts the new logic in screenshot.py that requires $DISPLAY for gnome-screenshot on non-Wayland. The consequence: /api/v2/computer/status reports screenshot_available: true, and /api/v2/computer/screenshot attempts to capture, fails with "Cannot open display", and then returns 500 (or 503 only if the error message happens to contain one of the markers). The PR's contract that status reflects actual usability is broken for this common headless configuration. Fix by requiring $DISPLAY for gnome-screenshot when not Wayland, as done for scrot.

gptme/server/computer_api.py
```diff
@@ -39,24 +39,59 @@
...
+    has_scrot = bool(shutil.which("scrot"))
+    is_wayland = os.environ.get("XDG_SESSION_TYPE") == "wayland"
+    if has_scrot and not is_wayland and not os.environ.get("DISPLAY"):
+        return (
+            "scrot is installed but $DISPLAY is unset, so screenshots cannot "
+            "be taken. Set DISPLAY to a running X11 server (or start Xvfb), "
+            "or install gnome-screenshot for a Wayland session."
+        )
+    return (
+        "No supported Linux screenshot backend available. "
```
TimeToBuildBob, line 105: ⚠️ **P2** — _linux_unavailable_hint in computer_api.py only special-cases scrot installed with $DISPLAY unset. When gnome-screenshot is installed on an X11 host with $DISPLAY unset, the condition `has_scrot` is False, so it falls through to the generic message that says "Install gnome-screenshot". This is misleading because gnome-screenshot is already installed; the real problem is a missing $DISPLAY. The hint should detect gnome-screenshot installed without DISPLAY on non-Wayland and suggest setting DISPLAY. As written, the user is sent on a wild goose chase reinstalling something they already have, contradicting the PR's stated goal of an actionable 503 body.

TimeToBuildBob:
Fixed two issues flagged by the AI reviewer:

**P1**: `_native_screenshot_available` was not requiring `$DISPLAY` for `gnome-screenshot` on X11, so `/api/v2/computer/status` could report `screenshot_available: true` when gnome-screenshot is installed but no X display is present. Now mirrors the same `is_wayland or has_display` guard already applied to scrot.

**P2**: `_linux_unavailable_hint` only handled the scrot-without-DISPLAY case; gnome-screenshot-without-DISPLAY fell through to the generic "install gnome-screenshot" message (misleading — it's already installed). Now detects both tools installed without DISPLAY and gives a concrete "set DISPLAY or use Wayland" message.

Tests added for both paths in `test_computer_api.py`.

TimeToBuildBob:
@greptileai review

gptme/server/computer_api.py
```diff
@@ -39,24 +39,62 @@
 computer_api = flask.Blueprint("computer_api", __name__)
 
 
+_DISPLAY_UNAVAILABLE_MARKERS = (
```
TimeToBuildBob, line 42: ❌ **P1** — The _DISPLAY_UNAVAILABLE_MARKERS tuple only includes four specific phrases: can't open x display, cannot open display, unable to open display, no display. Common X11 failure messages such as 'X connection failed', 'connection refused by server', or 'Unable to connect to X server' are not matched, so a stale or dead $DISPLAY that produces these messages will still be classified as a 500 internal error rather than a 503. This contradicts the PR's stated goal of treating any dead X display as 'screenshot unavailable'. Users with a dead X server will still see a 500 and the error will be logged as an internal error instead of a user-facing availability issue. _(2/2 passes)_

gptme/server/computer_api.py
```diff
@@ -181,9 +216,11 @@ def screenshot():
         )
     except Exception as exc:
         logger.exception("Screenshot failed")
+        # A missing/dead X display is unavailability (503), not an internal error.
+        status = 503 if _is_display_unavailable_error(exc) else 500
```
TimeToBuildBob, line 232: ⚠️ **P2** — The status is computed by calling _is_display_unavailable_error(exc) on the exception. If a non-native transport (e.g., a custom ComputerTransport) raises a subprocess.CalledProcessError directly, str(exc) is "Command '...' returned non-zero exit status 1" and does not include the stderr that contains the display error. Thus the marker check misses it and the endpoint returns 500 instead of the intended 503. The native path wraps CalledProcessError into a RuntimeError that embeds stderr, but custom transports may not. _(1/2 passes — pass 2 looked and disagreed)_

TimeToBuildBob:
Addressed the AI reviewer's open findings in commit `94927cdee`:

**P1 — Display-error markers too narrow** (`_DISPLAY_UNAVAILABLE_MARKERS`): Added `"failed to open display"` and `"could not open display"` so dead X servers that produce those strings are correctly classified as 503 rather than 500.

**P1 — Wayland path didn't check `WAYLAND_DISPLAY`**: `_native_screenshot_available()` (computer_api.py) and `_is_available()` (screenshot.py) both required `XDG_SESSION_TYPE=wayland` but not `WAYLAND_DISPLAY`. A headless Wayland host (compositor not running) would still advertise `screenshot_available: true`. Fixed by requiring `(is_wayland and WAYLAND_DISPLAY)` — mirrors the `DISPLAY` guard already applied for X11.

**P2 — `_is_display_unavailable_error` missed `CalledProcessError.stderr`**: If a transport raises a raw `CalledProcessError`, `str(exc)` omits stderr. Now also checks `exc.stderr` so display errors surfaced that way map to 503.

Tests updated: existing Wayland happy-path tests now set `WAYLAND_DISPLAY=wayland-0`; headless-Wayland negative cases added for both `computer_api.py` and `screenshot.py`.

TimeToBuildBob:
@greptileai review

tests/test_computer_api.py
```diff
@@ -343,23 +362,137 @@ def test_linux_without_screenshot_tools_returns_false(self, monkeypatch):
...
+        monkeypatch.delenv("DISPLAY", raising=False)
+        monkeypatch.delenv("XDG_SESSION_TYPE", raising=False)
+
+        with patch(
+            "shutil.which",
+            side_effect=lambda cmd: "/usr/bin/scrot" if cmd == "scrot" else None,
+        ):
+            from gptme.server.computer_api import _screenshot_available
+
+            assert _screenshot_available() is False
```
TimeToBuildBob, line 378: ⚠️ **P2** — Three tests in `test_computer_api.py` (`test_linux_scrot_without_display_returns_false`, `test_status_scrot_without_display_reports_unavailable`, and `test_screenshot_scrot_without_display_returns_503`) call `_screenshot_available()` or the status endpoint without mocking `gptme.server.computer_api.get_transport`. `_screenshot_available()` first calls `get_transport()`; if any transport is configured in the test environment (e.g., via environment or prior tests), it returns a non-None value and the function returns `True` regardless of display state. These tests would then fail (or pass vacuously) depending on the ambient environment, making them non-hermetic. They should patch `get_transport` to return `None` to isolate the native path. _(2/2 passes)_

PR #3623, opened 2026-08-25
fix(server): use os.write() and re-entrance guard in SIGTERM handlers

### LINKED ISSUE #3589: SERVER: Silent startup failure — process exits without binding port

TimeToBuildBob:
## Problem

SIGTERM handlers in `gptme/server/cli.py` used `_sys.stderr.write()` and `logger.info()`, neither of which is safe to call from a signal handler.

When a second SIGTERM arrived while the first invocation was still executing `stderr.write()`, Python's `BufferedWriter` raised:

```
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>
```

This produced a crash traceback instead of a clean shutdown (#3589).

The `logger.info()` call in `_handle_sigterm` was also unsafe: logging acquires locks and uses buffered I/O internally.

Complementary to #3597 (which guards against overriding custom handlers) — these two PRs fix different aspects of #3589.

Closes #3589.

## Fix

- Replace `_sys.stderr.write()` / `_sys.stderr.flush()` with `os.write(2, ...)` in both `_startup_sigterm_handler` and `_handle_sigterm`. `os.write()` is a raw POSIX syscall — it is async-signal-safe and does not use Python's buffered I/O layer.
- Add a module-level `_sigterm_received` flag so a rapid second SIGTERM returns immediately rather than re-entering the handler.
- Remove the `logger.info()` call from `_handle_sigterm`.

## Test

Added `test_sigterm_handler_survives_rapid_double_signal` to `tests/test_server_startup.py`: starts the server, waits for it to bind, then sends two SIGTERM signals 20 ms apart and asserts no `RuntimeError` appears in the output.

All 6 startup tests pass.

### DIFF
--- a/gptme/server/cli.py
+++ b/gptme/server/cli.py
@@ -2,10 +2,15 @@ import json
 import logging
 import os
 import signal
-import sys as _sys
 import threading
 import time
 from pathlib import Path
+
+# Re-entrance guard: SIGTERM can fire while a previous invocation is still
+# running (e.g. during a buffered write).  A module-level flag lets the
+# second invocation exit immediately rather than crashing with
+# "RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>".
+_sigterm_received = False
 
 
 # Install a minimal SIGTERM handler at module level — before the slow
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_server_startup.py` (+66/-0, Δ66)
- `gptme/server/cli.py` (+22/-9, Δ31)

### Suggested verification
- `uv run pytest tests/test_server_startup.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3623._

PR #3624, opened 2026-08-25
fix(shell): route all commands through TOOL_CONFIRM hook chain (closes #3598)

### LINKED ISSUE #3598: RFC: deterministic pre-execution guardrails as a plugin/hook (Grimdall)
### Summary

Propose an opt-in **plugin** (not a core change) that registers a tool-confirmation hook to add *deterministic, below-the-model* guardrails around gptme's tool execution: block destructive shell commands, deny reads of secret paths, and enforce an egress allowlist. Shadow-mode (log-only) by default — zero behavior change unless enabled.

### Motivation

gptme runs with the user's full permissions and can execute arbitrary code — the docs are explicit about this, and today the safety boundary is **human review of confirmations** (or containers/VMs for non-interactive use).

That boundary disappears on the increasingly important headless/autonomous path: `auto_confirm` auto-approves everything in non-interactive mode, and the new `service init` / headless-agent workflow runs with no human in the loop. In those cases there is no deterministic check between the model's output and the OS.

This is the "confused deputy" problem: the agent legitimately holds credentials by design, so enforcement must be deterministic and below the model — not prompt-level.

### Proposal: a plugin/hook, not a core rewrite

gptme's architecture already has the right insertion point. I'd ship this as a plugin that registers a `TOOL_CONFIRM`-style hook (documented as "blocking confirmation/deny decision before execution") — the same hook family `cli_confirm` / `auto_confirm` / `server_confirm` use today.

The plugin performs three deterministic checks, each independent of the model:

1. **Shell policy** — AST-parse commands before execution and block destructive patterns (`rm -rf`, `DROP TABLE`, fork bombs, `chmod -R 000`, etc.), regardless of prompt justification.
2. **Secret-read denial** — refuse reads of `~/.ssh`, `~/.aws`, `.env`, `*.pem`, keychain material, etc.
3. **Egress allowlist** — deny network egress not on an explicit allowlist (blocks secret exfiltration).

Default is **shadow mode**: it logs what it *would* block and only enforces when the user opts in. Latency is <2ms, so it's compatible with the "keep hooks fast" guidance.

MIT-licensed. The core is a general-purpose layer (https://github.com/grimdalltech/grimdall-os), but I'd build and maintain the gptme plugin specifically.

### Questions for maintainers

1. Is the `TOOL_CONFIRM` hook the right insertion point, and can a hook return a *deny* decision that prevents execution? (I noticed the docs list the tool hooks two ways — `TOOL_CONFIRM`/`TOOL_EXECUTE_PRE` in `hooks.rst` vs `TOOL_PRE_EXECUTE` in `plugins.rst` — so I'd appreciate the canonical name.)
2. Should this live in `gptme-contrib` as a community plugin, or as a standalone `gptme-grimdall` package using the `gptme.plugins` entry point?
3. Would you be open to a shadow-mode default that emits a `system` message showing what would have been blocked, so users can preview before enabling?

### Non-goals

- No changes to gptme core; purely additive via the plugin/hook system.
- Not a replacement for containers/VMs — complementary for the cases where those don't apply.
- Opt-in; no behavior change by default.

### Related

- #3390 (review-gated no-confirm workflow)
- #2298 (per-action `is_read_only` / `is_risky` metadata)
- #1164 (`TOOL_ALLOWLIST`)
- #1021 / #1017 / #1033 (prior security reviews)

Happy to open a PR with the plugin + tests if there's interest.

TimeToBuildBob:
## Problem

The shell tool bypassed `execute_with_confirmation()` for allowlisted commands (safe read-only commands like `cat`, `ls`, `grep`). This meant TOOL_CONFIRM guardrail hooks — including third-party plugins — **could not intercept or block these commands** even when explicitly registered at high priority.

Practical impact: `cat ~/.ssh/id_rsa` was auto-approved via the allowlist without giving any registered guardrail hook a chance to run.

Closes #3598 (RFC: deterministic pre-execution guardrails as a plugin/hook).

### `gptme/tools/shell.py`

Removed the allowlist fast-path bypass. All shell commands now go through `execute_with_confirmation()`:

```python
# BEFORE: allowlisted commands bypassed the hook chain entirely
if is_allowlisted(cmd):
    yield from execute_shell_impl(cmd, logdir, timeout=timeout)
else:
    yield from execute_with_confirmation(...)

# AFTER: all commands route through the hook chain
yield from execute_with_confirmation(
    cmd, args, kwargs,
    execute_fn=execute_fn, get_path_fn=get_path_fn,
    ...
)
```

The `shell_allowlist_hook` (registered at priority=10) still auto-confirms safe commands without prompting. User-visible behaviour is unchanged for allowlisted commands — no new prompts appear. The difference is that third-party hooks at priority > 10 now run *first* and can intercept any command.

### `gptme/tools/shell_validation.py`

Added `_SENSITIVE_HOME_DIRS` (`.ssh`, `.aws`, `.kube`, `.gnupg`, `.docker`, etc.) and extended `_has_sensitive_args()` to check for `~/...` prefixes. This closes the gap where `cat ~/.ssh/id_rsa` was previously allowlisted.

## Tests

- **`tests/test_guardrail_hook.py`** (new): 5 integration tests proving the full deny path:
  - A hook at priority > 10 blocks a non-allowlisted command
  - A hook at priority > 10 blocks an **allowlisted** command (core regression test for this PR)
  - A hook returning `None` lets the command execute normally
  - `skip()` prevents the command from running (sentinel file not created)
  - Guardrails still fire in headless/no-confirm mode

- **`tests/test_tools_shell_validation.py`**: 9 new tests for credential-path blocking (`~/.ssh`, `~/.aws`, `~/.kube`, etc.)

- **`tests/test_shell_allowlist_autoconfirm.py`**: Updated `test_allowlisted_command_executes_without_confirmation` — the test now verifies that the command executes (via `mock_shell.run`) rather than asserting `execute_with_confirmation` is not called (old bypass behaviour).

## Plugin example (from #3598)

A plugin can now reliably block any shell command with:

```python
from gptme.hooks import HookType, register_hook
from gptme.hooks.confirm import ConfirmationResult

def my_guardrail(tool_use, preview=None, workspace=None):
    if tool_use.tool != "shell":
        return None
    cmd = tool_use.content or ""
    if is_dangerous(cmd):
        return ConfirmationResult.skip("Blocked by policy: dangerous command")
    return None

register_hook("myplugin.guardrail", HookType.TOOL_CONFIRM, my_guardrail, priority=200)
```

This works even in headless/autonomous mode (`-y` / `--no-confirm`) and even for commands that would otherwise be allowlisted.

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1763,6 +1763,31 @@ def execute_shell(
             )
             return
 
+        # Denylist-check preceding commands too — they run inside execute_fn
+        # and would otherwise bypass this gate even though hooks only see bg_cmd.
+        if preceding_cmds.strip():
+            is_pre_denied, pre_deny_reason, pre_matched_cmd = is_denylisted(
+                preceding_cmds
+            )
+            if is_pre_denied:
+                yield Message(
+                    "system",
+                    f"Command denied (preceding): `{pre_matched_cmd}`\n\n{pre_deny_reason}",
+                )
+                return
+
+        # Build full command context so TOOL_CONFIRM hooks see the complete
+        # sequence — not just bg_cmd — when deciding whether to approve.
+        # A hook that should block `cat ~/.ssh/id_rsa` must see it even when
+        # it precedes an innocuous `bg ls`.
+        _full_cmd_parts: list[str] = []
+        if preceding_cmds.strip():
+            _full_cmd_parts.append(preceding_cmds.strip())
+        _full_cmd_parts.append(f"bg {bg_cmd}")
+        if remaining_cmds.strip():
+            _full_cmd_parts.append(remaining_cmds.strip())
+        _full_cmd_context = "\n".join(_full_cmd_parts)
+
         def _bg_execute_fn(c: str, p: Path | None) -> Generator[Message, None, None]:
             if preceding_cmds.strip():
                 yield from _execute_preceding_commands(preceding_cmds)
@@ -1770,13 +1795,16 @@ def execute_shell(
             if remaining_cmds.strip():
                 yield from execute_shell(remaining_cmds, None, None)
 
+        def _bg_preview_fn(content: str, path: Path | None) -> str | None:
+            return _full_cmd_context
+
         yield from execute_with_confirmation(
             bg_cmd,
             args,
             kwargs,
             execute_fn=_bg_execute_fn,
             get_path_fn=get_path_fn,
-            preview_fn=preview_shell,
+            preview_fn=_bg_preview_fn,
             preview_lang="bash",
             confirm_msg="Run command in background?",
             allow_edit=True,
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1734,39 +1734,54 @@ def execute_shell(
             break
 
     if bg_line_idx is not None:
-        # Found a bg command
+        # Found a bg command - extract bg_cmd and any surrounding commands
         _bg_memory_limit = _get_memory_limit()
+        preceding_cmds = ""
+        remaining_cmds = ""
+
         if bg_line_idx == 0 and len(lines) == 1:
             # Simple case: bg is the only command
             bg_cmd = cmd_stripped[3:].strip()
-            yield from execute_bg_command(bg_cmd, memory_limit=_bg_memory_limit)
-            return
         elif bg_line_idx > 0:
-            # bg is on a later line - execute preceding commands first (Issue #992)
+            # bg is on a later line - preceding commands modify shell state (Issue #992)
             preceding_cmds = "\n".join(lines[:bg_line_idx])
-            if preceding_cmds.strip():
-                # Execute preceding commands (they modify shell state like cd)
-                yield from _execute_preceding_commands(preceding_cmds)
-            # Now execute the bg command
-            bg_line = lines[bg_line_idx].strip()
-            bg_cmd = bg_line[3:].strip()  # Remove "bg " prefix
-            yield from execute_bg_command(bg_cmd, memory_limit=_bg_memory_limit)
-            # Execute any remaining commands after bg (unlikely but handle it)
+            bg_cmd = lines[bg_line_idx].strip()[3:].strip()  # Remove "bg " prefix
             if bg_line_idx < len(lines) - 1:
                 remaining_cmds = "\n".join(lines[bg_line_idx + 1 :])
-                if remaining_cmds.strip():
-                    yield from execute_shell(remaining_cmds, None, None)
-            return
         else:
             # bg is first line but there are more lines after it
-            # Start bg job, then execute remaining commands
-            bg_line = lines[0].strip()
-            bg_cmd = bg_line[3:].strip()
-            yield from execute_bg_command(bg_cmd, memory_limit=_bg_memory_limit)
+            bg_cmd = lines[0].strip()[3:].strip()
             remaining_cmds = "\n".join(lines[1:])
+
+        # Route bg payload through denylist + TOOL_CONFIRM hook chain before
+        # starting background execution so guardrails can intercept commands
+        # like `bg cat ~/.ssh/id_rsa` (Issue #3598).
+        is_bg_denied, bg_deny_reason, bg_matched_cmd = is_denylisted(bg_cmd)
+        if is_bg_denied:
+            yield Message(
+                "system", f"Command denied: `{bg_matched_cmd}`\n\n{bg_deny_reason}"
+            )
+            return
+
+        def _bg_execute_fn(c: str, p: Path | None) -> Generator[Message, None, None]:
+            if preceding_cmds.strip():
+                yield from _execute_preceding_commands(preceding_cmds)
+            yield from execute_bg_command(c, memory_limit=_bg_memory_limit)
             if remaining_cmds.strip():
                 yield from execute_shell(remaining_cmds, None, None)
-            return
+
+        yield from execute_with_confirmation(
+            bg_cmd,
+            args,
+            kwargs,
+            execute_fn=_bg_execute_fn,
+            get_path_fn=get_path_fn,
+            preview_fn=preview_shell,
+            preview_lang="bash",
+            confirm_msg="Run command in background?",
+            allow_edit=True,
+        )
+        return
 
     if cmd_lower == "jobs":
         # List background jobs
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1804,29 +1804,32 @@ def execute_shell(
         yield Message("system", f"Command denied: `{matched_cmd}`\n\n{deny_reason}")
         return
 
-    # Skip confirmation for allowlisted commands
-    if is_allowlisted(cmd):
-        logger.debug(f"Command allowlisted, skipping confirmation: {cmd[:80]}")
-        logdir = get_path_fn()
-        yield from execute_shell_impl(cmd, logdir, timeout=timeout)
-    else:
-        logger.debug(f"Command not allowlisted, requiring confirmation: {cmd[:80]}")
+    # All non-denied commands go through execute_with_confirmation so that
+    # TOOL_CONFIRM hooks (including third-party guardrail plugins) can intercept
+    # any command — including ones the built-in allowlist would auto-approve.
+    # The shell_allowlist_hook (TOOL_CONFIRM, priority=10) auto-confirms safe
+    # commands; a guardrail registered at priority > 10 runs first and can deny
+    # even "allowlisted" commands such as `cat ~/.ssh/id_rsa`.
+    logger.debug(
+        "Routing shell command through hook chain: %s",
+        cmd[:80],
+    )
 
-        # Create a wrapper function that passes timeout to execute_shell_impl
-        def execute_fn(cmd: str, path: Path | None) -> Generator[Message, None, None]:
-            return execute_shell_impl(cmd, path, timeout=timeout)
+    # Create a wrapper function that passes timeout to execute_shell_impl
+    def execute_fn(cmd: str, path: Path | None) -> Generator[Message, None, None]:
+        return execute_shell_impl(cmd, path, timeout=timeout)
 
-        yield from execute_with_confirmation(
-            cmd,
-            args,
-            kwargs,
-            execute_fn=execute_fn,
-            get_path_fn=get_path_fn,
-            preview_fn=preview_shell,
-            preview_lang="bash",
-            confirm_msg="Run command?",
-            allow_edit=True,
-        )
+    yield from execute_with_confirmation(
+        cmd,
+        args,
+        kwargs,
+        execute_fn=execute_fn,
+        get_path_fn=get_path_fn,
+        preview_fn=preview_shell,
+        preview_lang="bash",
+        confirm_msg="Run command?",
+        allow_edit=True,
+    )
 
 
 def _format_block_smart(header: str, cmd: str, lang="") -> str:
--- a/gptme/tools/shell_validation.py
+++ b/gptme/tools/shell_validation.py
@@ -24,13 +24,30 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 # Sensitive path prefixes — commands reading these should require confirmation
-# even when the command itself is in the allowlist (e.g. `cat /etc/shadow`)
+# even when the command itself is in the allowlist (e.g. `cat /etc/shadow`).
+# Note: home-relative paths (~/.ssh etc.) are covered by _SENSITIVE_HOME_DIRS below.
 _SENSITIVE_PATH_PREFIXES = (
     "/etc/",
     "/root/",
     "/proc/",
     "/sys/",
     "/boot/",
+)
+
+# Home-relative directories that contain credentials/secrets.
+# Matched against tokens starting with "~/", e.g. "~/.ssh/id_rsa".
+_SENSITIVE_HOME_DIRS = (
+    "~/.ssh",
+    "~/.aws",
+    "~/.gnupg",
+    "~/.pgp",
+    "~/.kube",
+    "~/.docker",
+    "~/.config/gcloud",
+    "~/.azure",
+    "~/.netrc",
+    "~/.npmrc",
+    "~/.pypirc",
 )
 
 # Commands that are safe to auto-approve without user confirmation
@@ -353,8 +370,11 @@ def _has_sensitive_args(cmd: str) -> bool:
         # because they contain no path separator.
         if "/" in token and any(char in token for char in "*?[{"):
             return True
-        # Sensitive directory prefixes
+        # Sensitive directory prefixes (absolute paths)
         if any(token.startswith(prefix) for prefix in _SENSITIVE_PATH_PREFIXES):
+            return True
+        # Sensitive home-relative credential directories (~/...)
+        if any(token.startswith(prefix) for prefix in _SENSITIVE_HOME_DIRS):
             return True
 
     return False
--- a/gptme/tools/shell_validation.py
+++ b/gptme/tools/shell_validation.py
@@ -383,6 +383,10 @@ def _has_sensitive_args(cmd: str) -> bool:
             if token.startswith(sub):
                 normalized = "~/" + token[len(sub) :]
                 break
+        # Collapse redundant separators so that $HOME//.ssh/id_rsa (→ ~//.ssh/id_rsa)
+        # still matches the ~/ prefix boundary after double-slash removal.
+        while "//" in normalized:
+            normalized = normalized.replace("//", "/")
         if any(
             normalized == prefix or normalized.startswith(prefix + "/")
             for prefix in _SENSITIVE_HOME_DIRS
--- a/gptme/tools/shell_validation.py
+++ b/gptme/tools/shell_validation.py
@@ -628,8 +628,12 @@ def shell_allowlist_hook(
     if tool_use.tool != "shell":
         return None
 
-    # Get the command from the tool use
-    cmd = tool_use.content.strip() if tool_use.content else ""
+    # Get the command to check.  For bg sequences the preview contains the full
+    # command context (preceding + bg + remaining) while tool_use.content is
+    # only the bg_cmd fragment.  Always prefer preview when present so that a
+    # dangerous preceding command (e.g. "cat ~/.ssh/id_rsa\nbg ls") is not
+    # silently approved because the isolated bg fragment ("ls") is allowlisted.
+    cmd = (preview or tool_use.content or "").strip()
     if not cmd:
         return None
 
--- a/gptme/tools/shell_validation.py
+++ b/gptme/tools/shell_validation.py
@@ -383,10 +383,19 @@ def _has_sensitive_args(cmd: str) -> bool:
             if token.startswith(sub):
                 normalized = "~/" + token[len(sub) :]
                 break
+        # ~username/... spellings (e.g. ~root/.ssh/id_rsa) name another user's
+        # home dir; strip the username component and treat the rest as ~/...
+        # so the sensitive-dir boundary check below fires for those too.
+        if re.match(r"^~[^/]+/", normalized) and not normalized.startswith("~/"):
+            normalized = "~/" + re.sub(r"^~[^/]+/", "", normalized)
         # Collapse redundant separators so that $HOME//.ssh/id_rsa (→ ~//.ssh/id_rsa)
         # still matches the ~/ prefix boundary after double-slash removal.
         while "//" in normalized:
             normalized = normalized.replace("//", "/")
+        # Remove no-op ./ segments (e.g. ~/./.ssh/id_rsa → ~/.ssh/id_rsa) that
+        # bash resolves at runtime but which bypass literal prefix matching.
+        while "/./" in normalized:
+            normalized = normalized.replace("/./", "/")
         if any(
             normalized == prefix or normalized.startswith(prefix + "/")
             for prefix in _SENSITIVE_HOME_DIRS
--- a/gptme/tools/shell_validation.py
+++ b/gptme/tools/shell_validation.py
@@ -373,8 +373,20 @@ def _has_sensitive_args(cmd: str) -> bool:
         # Sensitive directory prefixes (absolute paths)
         if any(token.startswith(prefix) for prefix in _SENSITIVE_PATH_PREFIXES):
             return True
-        # Sensitive home-relative credential directories (~/...)
-        if any(token.startswith(prefix) for prefix in _SENSITIVE_HOME_DIRS):
+        # Sensitive home-relative credential directories.
+        # Normalize $HOME/... and ${HOME}/... to ~/... before matching so that
+        # shell-variable spellings (cat "$HOME/.ssh/id_rsa") are caught too.
+        # Use a boundary check (exact match or followed by "/") to avoid
+        # false-positives on sibling paths like ~/.sshrc or ~/.npmrc-public.
+        normalized = token
+        for sub in ("${HOME}/", "$HOME/"):
+            if token.startswith(sub):
+                normalized = "~/" + token[len(sub) :]
+                break
+        if any(
+            normalized == prefix or normalized.startswith(prefix + "/")
+            for prefix in _SENSITIVE_HOME_DIRS
+        ):
             return True
 
     return False
--- a/tests/test_guardrail_hook.py
+++ b/tests/test_guardrail_hook.py
@@ -154,8 +154,9 @@ class TestToolConfirmGuardrailDeny:
 
         sentinel = tmp_path / "headless_test.txt"
 
-        # Save whether server_confirm was registered before this test mutates it.
+        # Save the pre-test hook set so the finally block can restore exactly it.
         _pre_test_hooks = {h.name for h in get_hooks(HookType.TOOL_CONFIRM)}
+        cli_confirm_was_registered = "cli_confirm" in _pre_test_hooks
         server_confirm_was_registered = "server_confirm" in _pre_test_hooks
 
         # Simulate normal interactive mode: register the built-in confirm hook.
@@ -185,8 +186,9 @@ class TestToolConfirmGuardrailDeny:
                 f"messages: {[m.content for m in msgs]}"
             )
         finally:
-            # Restore both hooks to their pre-test state.
-            register_cli_confirm()
+            # Restore hooks to exactly their pre-test state.
+            if cli_confirm_was_registered:
+                register_cli_confirm()
             if server_confirm_was_registered:
                 register_server_confirm()
 
--- a/tests/test_guardrail_hook.py
+++ b/tests/test_guardrail_hook.py
@@ -1,0 +1,165 @@
+"""Integration tests for TOOL_CONFIRM guardrail hooks (gptme#3598).
+
+Validates that a third-party plugin can register a TOOL_CONFIRM hook that
+blocks tool execution — including commands that the built-in allowlist would
+otherwise auto-approve — and that the block is returned as a system message.
+"""
+
+import pytest
+
+from gptme.hooks import HookType, register_hook, unregister_hook
+from gptme.hooks.confirm import ConfirmationResult
+from gptme.tools.base import ToolUse
+
+
+@pytest.fixture(autouse=True)
+def _clean_hooks():
+    """Remove any test guardrail hook before and after each test."""
+    unregister_hook("test.guardrail", HookType.TOOL_CONFIRM)
+    yield
+    unregister_hook("test.guardrail", HookType.TOOL_CONFIRM)
+
+
+class TestToolConfirmGuardrailDeny:
+    """A TOOL_CONFIRM hook at high priority can deny any shell command."""
+
+    def _make_guardrail(self, blocked_pattern: str):
+        def _hook(tool_use, preview=None, workspace=None):
+            if tool_use.tool != "shell":
+                return None
+            cmd = tool_use.content or ""
+            if blocked_pattern in cmd:
+                return ConfirmationResult.skip(
+                    f"Blocked by guardrail: {blocked_pattern!r} detected"
+                )
+            return None
+
+        return _hook
+
+    def test_guardrail_blocks_dangerous_command(self):
+        """A guardrail hook can block a command that would otherwise execute.
+
+        We use `curl evil.com` — not allowlisted (requires confirmation), not
+        denylisted (no unconditional built-in block), but blocked by our guardrail.
+        """
+        register_hook(
+            "test.guardrail",
+            HookType.TOOL_CONFIRM,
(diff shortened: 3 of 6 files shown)


gptme/tools/shell_validation.py
```diff
@@ -353,9 +370,12 @@ def _has_sensitive_args(cmd: str) -> bool:
         # because they contain no path separator.
         if "/" in token and any(char in token for char in "*?[{"):
             return True
-        # Sensitive directory prefixes
+        # Sensitive directory prefixes (absolute paths)
         if any(token.startswith(prefix) for prefix in _SENSITIVE_PATH_PREFIXES):
             return True
+        # Sensitive home-relative credential directories (~/...)
+        if any(token.startswith(prefix) for prefix in _SENSITIVE_HOME_DIRS):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 768f44c7. Replaced bare `startswith(prefix)` with `normalized == prefix or normalized.startswith(prefix + "/")`. This means `~/.sshrc` and `~/.npmrc-public` are no longer blocked, while `~/.ssh`, `~/.ssh/id_rsa`, and `~/.npmrc` still are. Tests: `test_ssh_sibling_dir_allowlisted`, `test_npmrc_sibling_allowlisted`, `test_npmrc_exact_still_blocked`, `test_ssh_dir_exact_still_blocked`.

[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 768f44c7. Before matching against `_SENSITIVE_HOME_DIRS`, the token is now normalized: `${HOME}/` and `$HOME/` are both rewritten to `~/` so all three spellings (`~/`, `$HOME/`, `${HOME}/`) reach the same check. 8 new regression tests added including `test_cat_home_var_ssh_not_allowlisted` and `test_cat_brace_home_var_ssh_not_allowlisted`.

TimeToBuildBob:
Addressed both Greptile findings in commit 768f44c7:

**P1 (security — \$HOME variable spellings not caught):** Added normalization before the sensitive-home check. `${HOME}/` and `$HOME/` are now rewritten to `~/` so `cat "$HOME/.ssh/id_rsa"` and `cat ${HOME}/.ssh/id_rsa` are blocked. 3 new tests cover these cases.

**P2 (prefix boundary false positives):** Replaced plain `startswith(prefix)` with `normalized == prefix or normalized.startswith(prefix + "/")`. `~/.sshrc` and `~/.npmrc-public` are no longer tripped; `~/.ssh`, `~/.ssh/id_rsa`, and `~/.npmrc` remain blocked. 4 new boundary tests added.

All 402 shell validation tests pass.

TimeToBuildBob:
@greptileai review

gptme/tools/shell_validation.py
```diff
@@ -353,9 +370,24 @@ def _has_sensitive_args(cmd: str) -> bool:
...
             return True
+        # Sensitive home-relative credential directories.
+        # Normalize $HOME/... and ${HOME}/... to ~/... before matching so that
+        # shell-variable spellings (cat "$HOME/.ssh/id_rsa") are caught too.
+        # Use a boundary check (exact match or followed by "/") to avoid
+        # false-positives on sibling paths like ~/.sshrc or ~/.npmrc-public.
+        normalized = token
+        for sub in ("${HOME}/", "$HOME/"):
+            if token.startswith(sub):
+                normalized = "~/" + token[len(sub) :]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 7a9b4554. After rewriting `$HOME` → `~/`, now collapse any redundant `//` sequences so that `$HOME//.ssh/id_rsa` → `~//.ssh/id_rsa` → `~/.ssh/id_rsa`, which correctly hits the sensitive-dir boundary. Two new regression tests: `test_double_slash_home_var_ssh_not_allowlisted` and `test_double_slash_tilde_ssh_not_allowlisted`.

TimeToBuildBob:
@greptileai review

gptme/tools/shell_validation.py
```diff
@@ -353,9 +370,28 @@ def _has_sensitive_args(cmd: str) -> bool:
...
+                break
+        # Collapse redundant separators so that $HOME//.ssh/id_rsa (→ ~//.ssh/id_rsa)
+        # still matches the ~/ prefix boundary after double-slash removal.
+        while "//" in normalized:
+            normalized = normalized.replace("//", "/")
+        if any(
+            normalized == prefix or normalized.startswith(prefix + "/")
```
TimeToBuildBob, line 400: ❌ **P1** — _has_sensitive_args only matches the '~/...' spelling (and the $HOME→~ rewrite). A tilde-with-username token such as `~root/.ssh/id_rsa` is a standard shell spelling of the same path but starts with neither '/', '~/', nor '$HOME/', so none of the sensitive-prefix, glob, or _SENSITIVE_HOME_DIRS checks fire and `cat ~root/.ssh/id_rsa` is auto-approved. When the gptme process runs as root (devcontainer, CI, SSH into a root box), this reads the root private key with no confirmation - the precise credential-exfiltration class the PR's new home-dir check is meant to stop, defeated by changing '~/' to '~root/'. The boundary logic should also treat '~username/…' forms (e.g. matching '~[^/]*/\.ssh' and the other prefixes) as sensitive.

tests/test_guardrail_hook.py
```diff
@@ -0,0 +1,165 @@
...
+
+        assert not sentinel.exists(), (
+            "Command must NOT have executed when guardrail returned skip(); "
+            f"messages: {[m.content for m in msgs]}"
+        )
+        assert any(m.role == "system" for m in msgs), (
+            "A skip result should produce a system message"
+        )
+
+    def test_no_confirm_mode_still_runs_guardrail(self, tmp_path):
```
TimeToBuildBob, line 142: ⚠️ **P2** — test_no_confirm_mode_still_runs_guardrail claims to verify that guardrails fire in --no-confirm/headless mode, but the test never enters no-confirm mode: it just registers a guardrail and calls ToolUse.execute() exactly as the other four tests in the class do. The docstring says '-y removes cli_confirm and server_confirm from the hook chain', yet no hook is unregistered or patched to simulate that. The test therefore passes regardless of whether guardrails actually run without cli_confirm registered, and would also pass if --no-confirm bypassed TOOL_CONFIRM entirely - it is vacuous and gives false assurance about the headless path. To test the claim, cli_confirm/server_confirm must be removed from the chain (or the registered hook set asserted) before executing.

TimeToBuildBob:
## 🤖 AI code review

Routes all shell commands through the TOOL_CONFIRM hook chain, removes the allowlist fast-path bypass in execute_shell, adds home-directory credential path detection to _has_sensitive_args, and updates/adds tests for guardrail hooks and allowlist behavior.

<details><summary><b>3 findings</b> · ❌ **2** P1 · ⚠️ **1** P2 — 🛑 **2 at P0/P1**</summary>

**❌ P1 high** — `gptme/tools/shell.py:1794`

The edit feature for background job sequences is broken. execute_with_confirmation is called with bg_cmd as the command, while the preview shows the full context (_full_cmd_context). When a user edits the preview, the edited text replaces only the bg_cmd parameter; the preceding and remaining command parts are taken from the closure variables `preceding_cmds` and `remaining_cmds`, which are never updated. Consequently, a user who edits a dangerous preceding command (e.g., changes `cat secret` to `cat safe`) will still have the original dangerous command executed, and any additions/removals to the preceding/remaining lines are silently ignored. This makes the confirmation dialog misleading and can lead to unintended command execution.

```suggestion
Pass the full context to execute_with_confirmation and have _bg_execute_fn parse the edited command, or set allow_edit=False for bg sequences.
```

<sub>**How this was verified:** Traced the bg branch in execute_shell: after denylist checks, it calls execute_with_confirmation(bg_cmd, ...) with preview_fn returning _full_cmd_context, which always includes the 'bg ' prefix. shell_allowlist_hook now uses preview as the command, so it checks the full context. Since 'bg' is not in allowlist_commands, is_allowlisted returns False, so the hook returns None, and the next hook (cli_confirm or server_confirm) prompts.</sub>

**❌ P1 high** — `gptme/tools/shell_validation.py:636`

The allowlist auto-confirm hook uses the preview to decide whether a command is safe. For background commands, `_bg_preview_fn` returns the full context including the literal "bg " prefix (e.g. "bg ls"). `is_allowlisted("bg ls")` returns False because "bg" is not an allowlisted command, so `shell_allowlist_hook` returns None and execution falls through to `cli_confirm`, prompting the user for every bg command. This contradicts the PR's claim that no new prompts appear for allowlisted commands, and breaks workflows that rely on `bg` to start servers/builds without interactive confirmation. Fix: either strip the "bg " token from the command before allowlist checking, or have the hook recognize and handle bg sequences specially.

```suggestion
Modify the allowlist hook to recognize the bg prefix and check only the payload (e.g., strip 'bg ' before calling is_allowlisted) while still passing the full context to guardrails.
```

<sub>**How this was verified:** Checked the bg branch in shell.py where `_bg_preview_fn` returns `_full_cmd_context`, which always includes a `bg ` prefix; also confirmed `is_allowlisted("bg ls")` is False by tracing the command name regex.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `tests/test_shell_allowlist_autoconfirm.py:205`

The updated test `test_allowlisted_command_executes_without_confirmation` is non-hermetic: it no longer mocks execute_with_confirmation or the hook chain, so it depends on the shell_allowlist_hook being registered at runtime. If the tool spec's hooks are not loaded in the test environment, execute_with_confirmation will have no auto-confirm hook and will fall through to other hooks or a prompt, preventing the command from executing and causing mock_shell.run to be uncalled. This makes the test flaky across different test harnesses or when the tool is not activated.

```suggestion
Patch execute_with_confirmation or ensure the shell_allowlist_hook is registered in the test setup (e.g., via a fixture that registers it).
```

<sub>**How this was verified:** Inspected the test: it patches get_shell and get_path_fn but not the hook chain; the success of the test relies on shell_allowlist_hook being active.</sub>

</details>

- `gptme/tools/shell.py` — Refactors bg command handling to route through execute_with_confirmation, and removes the allowlist bypass so all commands go through the hook chain.
- `gptme/tools/shell_validation.py` — Adds _SENSITIVE_HOME_DIRS and home-relative path normalization to _has_sensitive_args; updates shell_allowlist_hook to prefer preview over content.
- `tests/test_guardrail_hook.py` — Adds integration tests for TOOL_CONFIRM guardrail hooks, including bg command interception and headless mode behavior.
- `tests/test_search_alias.py` — Relaxes version assertion to regex match instead of exact string; adds re import.
- `tests/test_shell_allowlist_autoconfirm.py` — Updates test to assert command execution rather than absence of execute_with_confirmation call.
- `tests/test_tools_shell_validation.py` — Adds tests for home-directory sensitive path detection and $HOME/~/username normalization.

</details>

| `7a9b45548112` | 2/5 | 4 | llm | 2026-08-25 18:37 UTC |
| `95157bf9091b` | 3/5 | 2 | llm | 2026-08-25 20:32 UTC |
| `f189ffea37bd` | 3/5 | 3 | llm | 2026-08-26 03:02 UTC |
| `d2bc46387a79` | 3/5 | 2 | llm | 2026-08-26 05:03 UTC |
| `352b07a3d6bd` | 1/5 | 3 | llm | 2026-08-26 07:48 UTC |

</details>

<sub>Reviewed `88c482493908` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 177s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell_validation.py
```diff
@@ -353,9 +370,28 @@ def _has_sensitive_args(cmd: str) -> bool:
...
+            for prefix in _SENSITIVE_HOME_DIRS
+        ):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in bf12cf9f62. After the `//` dedup pass, now also collapse `/./` segments:

```python
while "/./" in normalized:
    normalized = normalized.replace("/./", "/")
```

So `$HOME/./.ssh/id_rsa` → (rewrite `$HOME/` → `~/`) → `~/./.ssh/id_rsa` → (collapse `./`) → `~/.ssh/id_rsa`, which correctly hits the `~/.ssh` boundary.

Two new regression tests: `test_dot_segment_home_var_ssh_not_allowlisted` and `test_dot_segment_tilde_ssh_not_allowlisted`.

gptme/tools/shell_validation.py
TimeToBuildBob: Fixed in bf12cf9f62. Added a normalization step for `~username/` spellings:

```python
if re.match(r"^~[^/]+/", normalized) and not normalized.startswith("~/"):
    normalized = "~/" + re.sub(r"^~[^/]+/", "", normalized)
```

So `~root/.ssh/id_rsa` → `~/.ssh/id_rsa`, which hits the sensitive-dir boundary check. Benign paths like `~alice/repos/project/README.md` → `~/repos/project/README.md` don't match any sensitive prefix and remain auto-approved.

New regression tests: `test_tilde_root_ssh_not_allowlisted`, `test_tilde_username_aws_not_allowlisted`, `test_tilde_username_benign_allowlisted`.

tests/test_guardrail_hook.py
TimeToBuildBob: Fixed in bf12cf9f62. The test now explicitly unregisters `cli_confirm` and `server_confirm` before running — simulating `--no-confirm` mode — so the registered guardrail is the only hook in the chain. Previously the test passed trivially regardless of whether guardrails fired without `cli_confirm`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed both remaining P1 findings (commit `95157bf90`):

**P1 — bg prefix bypasses TOOL_CONFIRM hook chain** (this commit):
Refactored the three bg sub-cases in `execute_shell` to extract `bg_cmd`, run it through `is_denylisted()`, then route it through `execute_with_confirmation()` with `execute_bg_command` as the execute_fn. Preceding and remaining commands run inside the confirmed execute_fn, so they only execute if the hook chain approves the bg payload. Regression test added: `test_bg_prefix_routes_through_hook_chain`.

**P1 — ~username/ home paths not caught** (prior commit `bf12cf9f6`):
Already fixed: `~root/.ssh/id_rsa` now normalizes to `~/.ssh/id_rsa` via the `re.sub(r"^~[^/]+/", "", ...)` rewrite before sensitive-dir boundary matching.

CI flake on `test_missing_required_tool_error` (19.98s setup > 10s timeout) is unrelated to this PR's changes — retrigger will clear it.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@TimeToBuildBob review

TimeToBuildBob:
Automated merge handoff for `95157bf9091b`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/shell.py`
- `gptme/tools/shell_validation.py`

gptme/tools/shell.py
```diff
@@ -1734,39 +1734,54 @@ def execute_shell(
...
+        is_bg_denied, bg_deny_reason, bg_matched_cmd = is_denylisted(bg_cmd)
+        if is_bg_denied:
+            yield Message(
+                "system", f"Command denied: `{bg_matched_cmd}`\n\n{bg_deny_reason}"
+            )
+            return
+
+        def _bg_execute_fn(c: str, p: Path | None) -> Generator[Message, None, None]:
+            if preceding_cmds.strip():
+                yield from _execute_preceding_commands(preceding_cmds)
```
TimeToBuildBob, line 1793: ❌ **P1** — The bg-handling path executes preceding commands (lines before the `bg` line) via `_execute_preceding_commands()`, which calls `shell_session.run()` directly and performs no denylist, allowlist, or TOOL_CONFIRM hook checks. Since the PR's stated goal is to route all shell commands through the hook chain, this is a bypass: a command like `cat ~/.ssh/id_rsa\nbg echo` will read the SSH key with no confirmation and no guardrail intervention. A denylisted command in the preceding lines also runs unconditionally. The fix is to validate `preceding_cmds` (e.g., by calling `execute_shell(preceding_cmds, None, None)` or at least `is_denylisted`) before executing.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed both remaining AI reviewer findings in `f189ffea3`:

**P1 — preceding_cmds bypass TOOL_CONFIRM hook chain** (`gptme/tools/shell.py`):
- Added `is_denylisted()` check for `preceding_cmds` so commands like `cat ~/.ssh/id_rsa` are blocked before the hook chain even runs
- Built `_full_cmd_context` (preceding + bg + remaining) and passed it via a custom `_bg_preview_fn` so TOOL_CONFIRM hooks see the complete command sequence — not just `bg_cmd` — before approving. A guardrail that blocks `secret_file` now fires even when it precedes an innocuous `bg ls`

**P2 — test_no_confirm_mode didn't prove its claim** (`tests/test_guardrail_hook.py`):
- Test now explicitly registers `cli_confirm` first (simulating normal interactive mode), then removes it (simulating `--no-confirm`), then asserts the guardrail still blocks — proving it fires independently rather than passing trivially because `cli_confirm` was never registered. `try/finally` restores `cli_confirm` for subsequent tests
- Added `test_preceding_cmds_visible_to_hook_chain` regression test for the P1 fix

All 7 tests pass locally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@/tmp/pr3624-status.md

gptme/tools/shell.py
```diff
@@ -1734,39 +1734,82 @@ def execute_shell(
...
+            yield from execute_bg_command(c, memory_limit=_bg_memory_limit)
             if remaining_cmds.strip():
                 yield from execute_shell(remaining_cmds, None, None)
-            return
+
+        def _bg_preview_fn(content: str, path: Path | None) -> str | None:
+            return _full_cmd_context
+
+        yield from execute_with_confirmation(
+            bg_cmd,
```
TimeToBuildBob, line 1802: ❌ **P1** — The bg branch passes only `bg_cmd` to execute_with_confirmation as the command, so TOOL_CONFIRM hooks receive a ToolUse whose `content` is just the part after "bg ". Preceding and remaining commands are only exposed via the `preview` parameter. A guardrail that inspects only `tool_use.content` (a common pattern, and exactly what the PR's own test guardrail does) will not see `cat secret_file` in a sequence like `cat secret_file\nbg ls`, and cannot block it. This defeats the stated purpose of exposing the full sequence. The dedicated test does not catch it because the guardrail checks content and because the `bg ls` line does not create the sentinel file it asserts on. _(2/2 passes)_

tests/test_guardrail_hook.py
```diff
@@ -0,0 +1,243 @@
...
+            self._make_guardrail("secret_file"),
+            priority=200,
+        )
+
+        # The dangerous command precedes an innocent bg payload.
+        # Without the fix, hooks only saw "ls" and approved.
+        tool_use = ToolUse(
+            tool="shell",
+            args=[],
+            content=f"cat secret_file\nbg ls {sentinel}",
```
TimeToBuildBob: ⚠️ **P2** — The test `test_preceding_cmds_visible_to_hook_chain` is vacuous. It uses `bg ls {sentinel}` as the background command, but `ls` lists a path and never creates it. The assertion `assert not sentinel.exists()` therefore passes regardless of whether the guardrail blocks anything. Additionally, the guardrail in `_make_guardrail` checks only `tool_use.content`, which for the bg branch contains only `bg_cmd` (i.e., `ls /tmp/.../preceded_bg_test.txt`), not the preceding `cat secret_file`. So the guardrail never sees the string it is configured to block. The test would pass even if the feature it claims to verify were completely broken. _(1/2 passes — pass 2 looked and disagreed)_

tests/test_guardrail_hook.py
```diff
@@ -0,0 +1,243 @@
...
+
+        sentinel = tmp_path / "headless_test.txt"
+
+        # Simulate normal interactive mode: register the built-in confirm hook.
+        register_cli_confirm()
+
+        try:
+            # Simulate --no-confirm: remove the built-in confirmation hooks.
+            unregister_hook("cli_confirm", HookType.TOOL_CONFIRM)
+            unregister_hook("server_confirm", HookType.TOOL_CONFIRM)
```
TimeToBuildBob, line 168: ⚠️ **P2** — In `test_no_confirm_mode_still_runs_guardrail`, the test unregisters both `cli_confirm` and `server_confirm` but only restores `cli_confirm` in the `finally` block. `server_confirm` is left unregistered after the test, corrupting global hook state. Since the autouse fixture only unregisters the test's own `test.guardrail` hook, the `server_confirm` registry is permanently changed, which can cause unrelated tests depending on that hook to behave differently depending on execution order. This is a test isolation defect that makes the suite non-hermetic. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed CI failure in 2ef4746d8:

The test `test_search_does_not_swallow_version` was asserting `result.output == f"{__version__}\\n"` (exact equality from `gptme.__version__`). In CI, `gptme --version` outputs `0.0.0 (unknown)` while `gptme.__version__.__version__` returns `0.33.0+unknown` — two different code paths with different formatting.

Updated the assertion to match master's version (`assert "gptme" in result.output.lower() or "version" in result.output.lower()`) which tests the actual intent (version info is shown) without depending on format consistency between `importlib.metadata` and the version string imported directly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
All fixes from `@ErikBjare`'s review are done and the PR is ready for merge:

- **`f189ffea3`** — Fixed both AI-reviewer P1/P2 findings (preceding_cmds visibility + test coverage)
- **`2ef4746d8`** — Fixed CI test failure (`test_search_does_not_swallow_version` format mismatch)
- **`d2bc46387`** — Final Greptile re-trigger

**Current status**: CI ✅ all green · Greptile ✅ 5/5 · Mergeable ✅

Self-merge not eligible (`gptme/tools/shell.py` is security-sensitive — requires human review). Ready for manual merge whenever you're happy with it.

tests/test_guardrail_hook.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+class TestToolConfirmGuardrailDeny:
+    """A TOOL_CONFIRM hook at high priority can deny any shell command."""
+
+    def _make_guardrail(self, blocked_pattern: str):
+        def _hook(tool_use, preview=None, workspace=None):
+            if tool_use.tool != "shell":
+                return None
+            cmd = tool_use.content or ""
```
TimeToBuildBob: ⚠️ **P2** — The test `test_preceding_cmds_visible_to_hook_chain` registers a guardrail that checks `blocked_pattern in tool_use.content` (line 30) and asserts a sentinel file is not created when a dangerous command precedes a `bg` line. However, `execute_shell` splits the input, extracts `bg_cmd = "ls <sentinel>"`, and passes only that as the content to `execute_with_confirmation` (shell.py:1802). The full context (including `cat secret_file`) is supplied only via `preview_fn`. Therefore the guardrail sees `tool_use.content == "ls <sentinel>"`, which does not contain "secret_file", so it returns None and the command executes, creating the sentinel. The test's `assert not sentinel.exists()` fails. This reveals that the PR's goal of showing preceding commands to hooks is not met through the content field, and the test is asserting behavior that cannot happen. _(1/2 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
Fixed remaining P2 AI-reviewer findings in `352b07a3d`:

**P2 — `_make_guardrail` only checked `tool_use.content`, not the full bg context:**
For `bg` sequences, `execute_with_confirmation` is called with `bg_cmd` as content, but the full preceding+bg+remaining context is available to hooks via the `preview` parameter. Updated `_make_guardrail` to use `cmd = preview or tool_use.content or ""` so guardrails see the complete command context, including preceding lines.

**P2 — `test_preceding_cmds_visible_to_hook_chain` assertion was vacuous:**
The test used `bg ls {sentinel}` as the bg payload. Since `ls` never creates a file, `assert not sentinel.exists()` passed regardless of whether the guardrail fired. Changed to `bg touch {sentinel}` — now if the guardrail fails to block, `touch` creates the sentinel and the assertion fails correctly.

**P2 — `server_confirm` not restored in `test_no_confirm_mode_still_runs_guardrail`:**
The `finally` block restored `cli_confirm` but not `server_confirm`, leaving global hook state corrupted. Added pre-test state capture and conditional restoration of `server_confirm` in the finally block.

All 7 guardrail hook tests + 409 shell validation tests pass locally.

gptme/tools/shell.py
TimeToBuildBob, line 1802: 🛑 **P0** — In the bg branch of execute_shell, execute_with_confirmation is called with bg_cmd as the command, so TOOL_CONFIRM hooks receive a ToolUse whose content is only the background command, not the preceding/remaining commands. The built-in shell_allowlist_hook (priority 10) inspects tool_use.content and auto-confirms if that command is allowlisted, ignoring the preview that contains the full sequence. Consequently a command like `rm -rf /tmp/important; bg echo hi` is approved because content is `echo hi` (allowlisted), and the preceding `rm -rf /tmp/important` executes without any user confirmation or guardrail rejection. This defeats the PR's goal and allows dangerous commands to bypass the hook chain entirely. Fix: pass the full command as content (or make the allowlist hook inspect preview when present). _(2/3 passes — pass 2 looked and disagreed)_

tests/test_guardrail_hook.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+            msgs = list(tool_use.execute())
+
+            assert not sentinel.exists(), (
+                "Guardrail must block execution even without cli_confirm registered; "
+                f"messages: {[m.content for m in msgs]}"
+            )
+        finally:
+            # Restore both hooks to their pre-test state.
+            register_cli_confirm()
```
TimeToBuildBob: ⚠️ **P2** — test_no_confirm_mode_still_runs_guardrail modifies the global TOOL_CONFIRM hook registry and does not fully restore it. It unconditionally registers cli_confirm in the finally block (line 189) even if it was not registered before the test, and only restores server_confirm if it was initially present (lines 190-191). After the test, cli_confirm is always registered, which can affect subsequent tests that assume a particular hook set. This makes the test non-hermetic and can cause investigation of failures in unrelated tests. Verified by reading the finally block and comparing with the pre-test state saving. _(3/3 passes)_

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Addressed all AI reviewer findings in commit 88c482493:

**P0/P1 (same root cause — allowlist hook ignores full bg command context):** `shell_allowlist_hook` was reading only `tool_use.content`, which for bg sequences is just the background command fragment (e.g. `ls`). A dangerous preceding command like `cat ~/.ssh/id_rsa\nbg ls` would be auto-confirmed because `ls` is allowlisted, defeating the PR's goal.

Fix: `cmd = (preview or tool_use.content or "").strip()` — the hook now inspects the full command context (preceding + bg + remaining) when a preview is available, so the dangerous preceding command is caught.

**P2 (non-hermetic test):** `test_no_confirm_mode_still_runs_guardrail` was unconditionally re-registering `cli_confirm` in its `finally` block even if it was absent before the test, leaving a stray hook that could affect later tests.

Fix: saved `cli_confirm_was_registered` pre-test and only restore conditionally.

All 462 shell validation + guardrail tests pass locally.

gptme/tools/shell_validation.py
```diff
@@ -583,8 +628,12 @@ def shell_allowlist_hook(
...
         return None
 
-    # Get the command from the tool use
-    cmd = tool_use.content.strip() if tool_use.content else ""
+    # Get the command to check.  For bg sequences the preview contains the full
+    # command context (preceding + bg + remaining) while tool_use.content is
+    # only the bg_cmd fragment.  Always prefer preview when present so that a
+    # dangerous preceding command (e.g. "cat ~/.ssh/id_rsa\nbg ls") is not
+    # silently approved because the isolated bg fragment ("ls") is allowlisted.
+    cmd = (preview or tool_use.content or "").strip()
```
TimeToBuildBob, line 636: ❌ **P1** — The allowlist auto-confirm hook uses the preview to decide whether a command is safe. For background commands, `_bg_preview_fn` returns the full context including the literal "bg " prefix (e.g. "bg ls"). `is_allowlisted("bg ls")` returns False because "bg" is not an allowlisted command, so `shell_allowlist_hook` returns None and execution falls through to `cli_confirm`, prompting the user for every bg command. This contradicts the PR's claim that no new prompts appear for allowlisted commands, and breaks workflows that rely on `bg` to start servers/builds without interactive confirmation. Fix: either strip the "bg " token from the command before allowlist checking, or have the hook recognize and handle bg sequences specially. _(2/3 passes — pass 2 looked and disagreed)_

tests/test_shell_allowlist_autoconfirm.py
```diff
@@ -190,24 +190,25 @@ def mock_logdir(self, tmp_path):
...
-            # Should have executed and returned a message
-            assert len(messages) == 1
-            assert "Ran allowlisted command" in messages[0].content
-            assert (
-                "cat README.md | head -100" in messages[0].content
-                or "cat README.md" in messages[0].content
-            )
+        # Execute the command - goes through the hook chain, auto-confirmed by
+        # shell_allowlist_hook; no user prompt appears.
+        messages = list(execute_shell(cmd, [], None))
```
TimeToBuildBob, line 205: ⚠️ **P2** — The updated test `test_allowlisted_command_executes_without_confirmation` is non-hermetic: it no longer mocks execute_with_confirmation or the hook chain, so it depends on the shell_allowlist_hook being registered at runtime. If the tool spec's hooks are not loaded in the test environment, execute_with_confirmation will have no auto-confirm hook and will fall through to other hooks or a prompt, preventing the command from executing and causing mock_shell.run to be uncalled. This makes the test flaky across different test harnesses or when the tool is not activated. _(1/3 passes — passes 2, 3 looked and disagreed)_

TimeToBuildBob:
The one failing CI test (`test_v2_chat_config_patch_preserves_webui_html_hint` in `test_server_v2.py`) is pre-existing — the same test is failing on the latest master CI run (run 32954533149, `f636b9f7`). It's a system-message-count assertion in the v2 server API that has drifted from the actual count; it's unrelated to shell routing changes in this PR. The failure reproduces identically on master without any of this PR's changes.

All other CI checks pass (lint, typecheck, API tests, Docker, PyInstaller, Startup Benchmark). This PR is ready for review.

ErikBjare:
@TimeToBuildBob fix remaining review comments

TimeToBuildBob:
Superseded by #3636. This PR was accidentally closed when cleaning up a phantom upstream branch. All code is preserved — the new PR is at the same commit (rebased onto current master).

PR #3627, opened 2026-08-25
docs(tool-formats): refresh coverage numbers after #3566 + pin them with a test

TimeToBuildBob:
Follow-on to #3566, which merged ~30 minutes ago.

## The problem

#3566 stamped `default_tool_format="tool"` on every OpenAI-compatible provider, taking coverage from **14 models** (openai-subscription only) to **92 of 111**. The "How well populated is this?" section of `docs/tool-formats.rst` still described the pre-#3566 state, so the page actively misinformed:

> `default_tool_format` is set for the `openai-subscription` models only. No other model in the bundled registry declares one, so for almost every model step 4 above is a no-op and you land on `markdown`.

Both sentences are now false.

## Changes

- **Refresh all three coverage counts** against the registry — 92/111 with `default_tool_format`, 29 `supports_parallel_tool_calls`, 23 `supports_strict_tools` (measured, not estimated).
- **Explain the stamp is per-provider, not per-model**, and that an explicit per-model value still wins. `anthropic` (Anthropic SDK, not OpenAI-compat) and `mock` (test-only) are deliberately excluded.
- **Note the four empty-registry providers** (`azure`, `local`, `nvidia`, `gptme`) that are OpenAI-compat but ship no static entries, so they get the default at resolution time instead.
- **Sharpen the evals pointer.** The provider stamp encodes "this provider *can* do native tool calls", not "`tool` measurably beats `markdown` for this model". Those are different claims, and only the leaderboard settles the second — so per-model eval-derived values remain the intended refinement, not a closed question.

## The test

The numbers went stale once already, silently. `tests/test_docs_tool_formats_coverage.py` parses the counts out of the prose and asserts them against the live registry, so the next registry change that moves coverage fails with a message naming the file to update.

Verified it is a real guard rather than a tautology: reverting the doc number to the pre-#3566 value (14) reproduces the exact failure this PR fixes.

```
AssertionError: docs say 14 models set default_tool_format, registry has 92
  — update docs/tool-formats.rst
```

A second test pins the `anthropic`/`mock` exclusion the prose documents.

## Verification

- `pytest tests/test_llm_models.py tests/test_llm_models_resolution.py tests/test_docs_tool_formats_coverage.py` — 167 passed
- `prek run --files ...` — all hooks green
- Docs-and-tests only; no runtime behavior change.

### DIFF
--- a/tests/test_docs_tool_formats_coverage.py
+++ b/tests/test_docs_tool_formats_coverage.py
@@ -24,7 +24,7 @@ def _counts() -> dict[str, int]:
     for models in MODELS.values():
         for props in models.values():
             total += 1
-            tool_format += bool(props.get("default_tool_format"))
+            tool_format += props.get("default_tool_format") == "tool"
             parallel += bool(props.get("supports_parallel_tool_calls"))
             strict += bool(props.get("supports_strict_tools"))
     return {
--- a/tests/test_docs_tool_formats_coverage.py
+++ b/tests/test_docs_tool_formats_coverage.py
@@ -76,3 +76,23 @@ def test_excluded_providers_really_are_excluded():
             f"{provider} models now carry default_tool_format ({stamped}), but "
             "docs/tool-formats.rst documents them as excluded"
         )
+
+
+def test_empty_registry_providers_have_no_static_entries():
+    """azure/local/nvidia/gptme are documented as having no static registry entries.
+
+    These are OpenAI-compat providers that inherit default_tool_format at
+    resolution time rather than from a per-model entry in the static registry.
+    If a static entry is added for any of them, the docs must be updated to
+    reflect it (the count and the explanation both change).
+    """
+    empty_registry: list[Provider] = ["azure", "local", "nvidia", "gptme"]
+    for provider in empty_registry:
+        if provider in MODELS:
+            entries = MODELS[provider]
+            assert not entries, (
+                f"'{provider}' is documented as having no static registry entries "
+                "(docs/tool-formats.rst), but now has "
+                f"{len(entries)} ({list(entries)}) — update the docs to reflect "
+                "the new entries, or keep the provider registry-free"
+            )
--- a/tests/test_docs_tool_formats_coverage.py
+++ b/tests/test_docs_tool_formats_coverage.py
@@ -1,0 +1,78 @@
+"""Guard against docs/tool-formats.rst drifting from the model registry.
+
+The "How well populated is this?" section quotes concrete coverage counts for
+the ModelMeta tool-calling fields. Those numbers went stale once already (they
+described the pre-#3566 state, where only openai-subscription models carried a
+``default_tool_format``). This test fails the moment the registry and the prose
+disagree, so the fix lands with the change that caused it.
+"""
+
+import re
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from gptme.llm.models.data import MODELS
+
+if TYPE_CHECKING:
+    from gptme.llm.models.types import Provider
+
+DOCS = Path(__file__).parent.parent / "docs" / "tool-formats.rst"
+
+
+def _counts() -> dict[str, int]:
+    total = tool_format = parallel = strict = 0
+    for models in MODELS.values():
+        for props in models.values():
+            total += 1
+            tool_format += bool(props.get("default_tool_format"))
+            parallel += bool(props.get("supports_parallel_tool_calls"))
+            strict += bool(props.get("supports_strict_tools"))
+    return {
+        "total": total,
+        "default_tool_format": tool_format,
+        "supports_parallel_tool_calls": parallel,
+        "supports_strict_tools": strict,
+    }
+
+
+def test_documented_coverage_matches_registry():
+    text = DOCS.read_text()
+    counts = _counts()
+
+    # "set to ``tool`` on 92 of the 111 models"
+    m = re.search(
+        r"``default_tool_format`` is set to ``tool`` on (\d+) of the (\d+) models",
+        text,
+    )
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Fixed the Greptile P2 finding: the test was counting any non-empty `default_tool_format` rather than specifically `== "tool"`. If a model were later given an explicit `markdown` or `xml` format, the count would still match but the documented claim would be wrong.

Changed `bool(props.get("default_tool_format"))` → `props.get("default_tool_format") == "tool"` in 89d9423.

TimeToBuildBob:
@greptileai review

tests/test_docs_tool_formats_coverage.py
```diff
@@ -0,0 +1,78 @@
...
+        assert m, f"coverage sentence for {field} not found in docs"
+        assert int(m.group(1)) == counts[field], (
+            f"docs say {m.group(1)} entries set {field}, registry has "
+            f"{counts[field]} — update docs/tool-formats.rst"
+        )
+
+
+def test_excluded_providers_really_are_excluded():
+    """anthropic/mock are documented as deliberately not stamped."""
+    excluded: list[Provider] = ["anthropic", "mock"]
```
TimeToBuildBob, line 68: ⚠️ **P2** — The test test_excluded_providers_really_are_excluded only checks that the anthropic and mock providers have no model with default_tool_format, but the docs also explicitly state that four OpenAI-compatible providers (azure, local, nvidia, gptme) "ship no static registry entries at all" (lines 236-239). The test never asserts that these providers are absent from MODELS or have empty dicts. If someone adds an empty registry entry for one of these providers (e.g., MODELS["azure"] = {}), the total model count remains 111 and all numeric coverage assertions still pass, so the documented claim silently goes stale without the test failing. To fully pin the docs, the test should also verify that these four provider keys are either not present or map to an empty dict.

TimeToBuildBob:
## 🤖 AI code review

This PR updates the "How well populated is this?" section of docs/tool-formats.rst to reflect the post-#3566 coverage numbers (92/111 for default_tool_format, 29 for supports_parallel_tool_calls, 23 for supports_strict_tools) and explains that the default_tool_format stamp is per-provider rather than per-model. It adds a new test file that parses these numbers out of the docs and asserts them against the live model registry, plus a test that the anthropic and mock providers have no default_tool_format set.

**⚠️ P2 medium** — `tests/test_docs_tool_formats_coverage.py:68`

The test test_excluded_providers_really_are_excluded only checks that the anthropic and mock providers have no model with default_tool_format, but the docs also explicitly state that four OpenAI-compatible providers (azure, local, nvidia, gptme) "ship no static registry entries at all" (lines 236-239). The test never asserts that these providers are absent from MODELS or have empty dicts. If someone adds an empty registry entry for one of these providers (e.g., MODELS["azure"] = {}), the total model count remains 111 and all numeric coverage assertions still pass, so the documented claim silently goes stale without the test failing. To fully pin the docs, the test should also verify that these four provider keys are either not present or map to an empty dict.

<sub>**How this was verified:** Checked the docs lines 236-239 in the new file, which state the four providers have no static registry entries, and the test's excluded list on line 68 which only includes anthropic and mock.</sub>

</details>

- `docs/tool-formats.rst` — Refreshes the coverage counts for default_tool_format, supports_parallel_tool_calls, and supports_strict_tools, and adds explanatory text about per-provider stamping and providers without static registry entries.
- `tests/test_docs_tool_formats_coverage.py` — Adds tests that parse the documented coverage numbers from the docs and compare them to the model registry, and that verify the anthropic and mock providers have no default_tool_format.

</details>

<sub>Reviewed `89d9423157d9` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 79s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_docs_tool_formats_coverage.py
```diff
@@ -0,0 +1,78 @@
...
+
+DOCS = Path(__file__).parent.parent / "docs" / "tool-formats.rst"
+
+
+def _counts() -> dict[str, int]:
+    total = tool_format = parallel = strict = 0
+    for models in MODELS.values():
+        for props in models.values():
+            total += 1
+            tool_format += bool(props.get("default_tool_format"))
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — line 27 already compares against the string `"tool"` specifically:

```python
tool_format += props.get("default_tool_format") == "tool"
```

An explicit `markdown` or `xml` value returns `False` from the equality check and does not increment the counter, so the guard is already value-specific. No change needed here.

tests/test_docs_tool_formats_coverage.py
TimeToBuildBob: Fixed in 25e22a052. Added `test_empty_registry_providers_have_no_static_entries` which asserts that azure, local, nvidia, and gptme are either absent from MODELS or map to an empty dict — so a static entry added to any of those four providers will fail with a clear message pointing at the docs.

TimeToBuildBob:
Superseded by gptme/gptme#3628.

Same `docs/tool-formats.rst` + `tests/test_docs_tool_formats_coverage.py` files, but the parallel-tool audit in #3628 moves `supports_parallel_tool_calls` from 29 to 56, so this PR's pinned counts would fail the moment #3628 landed. Closing this one rather than stacking two reviews on the same section.

The eval-derived `default_tool_format` half (no trustworthy per-model override at n≥10 A/B) is also recorded there.

TimeToBuildBob:
Closing as superseded by gptme/gptme#3628 (same docs+test, updated parallel coverage 29→56).

PR #3628, opened 2026-08-25
feat(models): parallel-tool flags from provider docs; eval-format verdict

TimeToBuildBob:
Follow-on to #3566 (provider-level `default_tool_format="tool"`) and overlapping with #3627 (docs coverage refresh).

## Phase 1 — eval signal, no per-model overrides

Ran `python -m gptme.eval.leaderboard --results-dir eval_results --format json --min-tests 4` (24,815 rows). The leaderboard's display cutoff of 4 tests is **not** a trustworthy override bar: n=4 is usually basic-only, "best format" is often the only format that was run, and Wilson ties are not disagreements.

**Bar used:** both the candidate format and the current heuristic tested on ≥10 unique tests, a strict Wilson-score win, a valid `markdown`/`xml`/`tool` value, and the model in the static registry.

**Verdict:** no registry model clears that bar. The #3566 provider heuristic stays. Apparent n=4 disagreements (`openai/gpt-5` markdown 4/5 vs tool 3/5, `qwen3-max` 4/4 tie, `gemini-2.5-flash` markdown with `tool` never tested, etc.) are noise or missing A/B.

Write-up: this is the "no disagreement" outcome #3566 deferred. Do not manufacture per-model overrides that restate the provider default.

## Phase 3 — capability-flag audit

`supports_parallel_tool_calls` is what makes gptme *not* stop after the first tool call. It was unset for every non-OpenAI/Anthropic provider, which meant Gemini/Grok/Groq/DeepSeek users got one tool per turn even when the provider documents parallel calls.

Set `True` only where official docs document support:

| Provider | Source | Change |
|---|---|---|
| gemini (all 12) | [Gemini parallel function calling](https://ai.google.dev/gemini-api/docs/function-calling#parallel_function_calling) | stamp via `PARALLEL_TOOL_PROVIDERS` |
| xai (7) + grok-subscription (2) | [xAI: parallel enabled by default](https://docs.x.ai/developers/tools/function-calling#parallel-function-calling) | stamp via `PARALLEL_TOOL_PROVIDERS` |
| groq `llama-3.3-70b-versatile` | [Groq supported-models table](https://console.groq.com/docs/tool-use/overview#supported-models) (Yes for llama-3.3, **No** for gpt-oss) | per-model, not provider-wide |
| deepseek chat/reasoner | [DeepSeek 2024-07-25 API upgrade](https://api-docs.deepseek.com/news/news0725/) | stamp via `PARALLEL_TOOL_PROVIDERS` |
| openrouter `google/gemini-3.5-flash`, `deepseek/deepseek-v4-*` | inherit the above | per-model |

Left **unset** (not recorded, not "unsupported"):

- Moonshot Kimi K2/K2.6 — official tool-call guide does not document parallel (K3 already has the flag)
- `supports_strict_tools` for DeepSeek — docs require `base_url=.../beta`, which gptme does not use
- `supports_strict_tools` for xAI — tool args are implicitly schema-strict; that is not the OpenAI `strict=True` request field

Coverage: parallel **29 → 56**, strict stays 23, `default_tool_format` stays 92/111.

## Docs

Refreshes the "How well populated is this?" section (still described the pre-#3566 state on master) and pins the counts with `tests/test_docs_tool_formats_coverage.py`.

This is a superset of #3627: same docs+test files, with the post-audit parallel count (56, not 29) and the eval-verdict caveat. **#3627 should close as superseded** once this is up; merging #3627 first would immediately fail its own coverage test after this lands.

## Tests

- `tests/test_llm_models.py::TestSupportsParallelToolCalls` — gemini, xAI, grok-subscription, groq llama-3.3, deepseek, OpenRouter Gemini alias, Kimi K2 still unset, DeepSeek strict still unset, `_mark_parallel` does not overwrite explicit False
- `tests/test_docs_tool_formats_coverage.py` — docs counts match the live registry

```
uv run pytest tests/test_llm_models.py::TestSupportsParallelToolCalls tests/test_docs_tool_formats_coverage.py
# 20 passed
```

### DIFF
--- a/gptme/llm/models/data.py
+++ b/gptme/llm/models/data.py
@@ -26,6 +26,20 @@ def _set_tool_format(
     }
 
 
+def _mark_parallel(models: dict[str, _ModelDictMeta]) -> dict[str, _ModelDictMeta]:
+    """Stamp supports_parallel_tool_calls=True unless the model already sets it.
+
+    An explicit per-model value still wins, so a later model that does *not*
+    support parallel can opt out by setting the key to False.
+    """
+    return {
+        name: props
+        if "supports_parallel_tool_calls" in props
+        else {**props, "supports_parallel_tool_calls": True}
+        for name, props in models.items()
+    }
+
+
 # Providers that route through the OpenAI-compatible function-calling API — stamp
 # default_tool_format="tool" on every model that doesn't already have one set.
 # Anthropic and mock are excluded: anthropic uses the Anthropic SDK (not OpenAI-compat),
@@ -50,6 +64,24 @@ OPENAI_COMPAT_PROVIDERS: frozenset[str] = frozenset(
         # via the OpenAI-compatible API (see llm_openai.py) — same fallback applies
         # when dynamic fetch fails/misses and no static registry entry exists.
         "gptme",
+    }
+)
+
+# Providers whose official docs state that current models can emit multiple
+# tool calls in one response. Applied at MODELS construction so it cannot
+# drift from the static dicts. Mixed providers (openrouter, groq) are stamped
+# per-model instead: Groq's own table is model-specific (llama-3.3 Yes,
+# gpt-oss No), and OpenRouter aliases a mix of backends.
+# Docs:
+#   gemini: https://ai.google.dev/gemini-api/docs/function-calling#parallel_function_calling
+#   xai / grok-subscription: https://docs.x.ai/developers/tools/function-calling#parallel-function-calling
+#   deepseek: https://api-docs.deepseek.com/news/news0725/
+PARALLEL_TOOL_PROVIDERS: frozenset[str] = frozenset(
+    {
+        "gemini",
+        "deepseek",
+        "xai",
+        "grok-subscription",
     }
 )
 
@@ -313,6 +345,11 @@ _MODELS_RAW: dict[Provider, dict[str, _ModelDictMeta]] = {
         },
     },
     # https://api-docs.deepseek.com/quick_start/pricing
+    # Parallel: stamped via PARALLEL_TOOL_PROVIDERS.
+    # `strict` mode exists but requires base_url=.../beta, which gptme does not
+    # use — leave supports_strict_tools unset rather than send a flag the
+    # production endpoint may reject.
+    # https://api-docs.deepseek.com/guides/tool_calls/#strict-mode-beta
     "deepseek": {
         "deepseek-chat": {
             "context": 128_000,
@@ -332,6 +369,9 @@ _MODELS_RAW: dict[Provider, dict[str, _ModelDictMeta]] = {
         },
     },
     # https://groq.com/pricing/
+    # Parallel tool use is model-specific on Groq (llama-3.3-70b-versatile Yes,
+    # openai/gpt-oss-* No). Stamp per-model, not via PARALLEL_TOOL_PROVIDERS.
+    # https://console.groq.com/docs/tool-use/overview#supported-models
     "groq": {
         "llama-3.3-70b-versatile": {
             "context": 128_000,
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
CI nearly done (2 "Test without API keys" jobs still pending, all others green). Greptile 5/5, no blocking findings. Ready for review/merge once those complete.

TimeToBuildBob:
Automated merge handoff for `e598cc02ba77`:

- CI gate: satisfied
- Review threads: clear
- `gptme/llm/models/data.py`

gptme/llm/models/data.py
```diff
@@ -614,13 +662,24 @@ def _set_tool_format(
 # check that all providers have a _MODELS_RAW entry
 assert set(PROVIDERS) == set(_MODELS_RAW.keys())
 
-# Stamp default_tool_format="tool" on all OpenAI-compatible providers at construction
+
+def _stamp_provider(
+    provider: str, models: dict[str, _ModelDictMeta]
+) -> dict[str, _ModelDictMeta]:
+    """Apply construction-time stamps so static dicts and MODELS stay consistent."""
+    if provider in PARALLEL_TOOL_PROVIDERS:
```
TimeToBuildBob, line 670: ⚠️ **P2** — In _stamp_provider, _mark_parallel is applied to every model of a provider listed in PARALLEL_TOOL_PROVIDERS, which includes gemini, xai, grok-subscription, and deepseek. This unconditionally sets supports_parallel_tool_calls=True on all models in those providers, regardless of whether each individual model actually supports parallel tool calls. For example, gemini-1.5-flash-latest in the gemini provider (line 275) has no explicit flag and will be stamped True. If that model does not support parallel tool calls, gptme will no longer stop after the first tool call and may emit multiple tool calls in a single response, which the provider could reject, causing a request failure. The added tests only verify one model per provider, so they do not catch this. This is a correctness risk that could break normal usage for users of older models in these providers.

TimeToBuildBob:
## 🤖 AI code review

This PR introduces provider-level stamping of supports_parallel_tool_calls based on official documentation, adds a new test to ensure the docs coverage counts match the registry, and adds tests for the new parallel flags. It also refreshes the 'How well populated is this?' section in the documentation.

**⚠️ P2 medium** — `gptme/llm/models/data.py:670`

In _stamp_provider, _mark_parallel is applied to every model of a provider listed in PARALLEL_TOOL_PROVIDERS, which includes gemini, xai, grok-subscription, and deepseek. This unconditionally sets supports_parallel_tool_calls=True on all models in those providers, regardless of whether each individual model actually supports parallel tool calls. For example, gemini-1.5-flash-latest in the gemini provider (line 275) has no explicit flag and will be stamped True. If that model does not support parallel tool calls, gptme will no longer stop after the first tool call and may emit multiple tool calls in a single response, which the provider could reject, causing a request failure. The added tests only verify one model per provider, so they do not catch this. This is a correctness risk that could break normal usage for users of older models in these providers.

<sub>**How this was verified:** Checked that PARALLEL_TOOL_PROVIDERS includes gemini, and that gemini has 12 models including gemini-1.5-flash-latest, none of which set supports_parallel_tool_calls, so the stamp applies to all.</sub>

</details>

- `docs/tool-formats.rst` — Updated the coverage counts and added explanation of provider-level stamping for tool-format and parallel-tool flags.
- `gptme/llm/models/data.py` — Added _mark_parallel, PARALLEL_TOOL_PROVIDERS, and _stamp_provider to apply parallel and tool-format stamps at MODELS construction.
- `tests/test_docs_tool_formats_coverage.py` — New test file that verifies the documented coverage counts match the actual registry counts.
- `tests/test_llm_models.py` — Added tests for parallel tool call support on specific models and a test for _mark_parallel.

</details>

<sub>Reviewed `e598cc02ba77` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 221s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

ErikBjare:
@TimeToBuildBob Merged, but worth acknowledging the P2 and maybe filing a follow-up PR to address the situation where we lack data.

TimeToBuildBob:
Acknowledged. Filed gptme/gptme#3632 to track the per-model verification gap — options range from a simple opt-out list for known non-parallel older variants to a self-correcting flag-clear on provider rejection. Will address there.

PR #3629, opened 2026-08-25
docs: add Browser Use Cloud CDP setup

MagMueller:
## Summary

- document how to create a Browser Use Cloud browser and connect gptme over CDP
- parse session fields strictly and fail on a bad API response
- always stop the managed browser on normal exit, Ctrl-C, or termination

## Checks

- `uv run --with pytest --with pytest-mock --with pytest-timeout --with pytest-retry --with pytest-asyncio pytest tests/test_tools_browser_thread.py -q`: 54 passed
- real local Chromium CDP: gptme BrowserThread connected, opened an isolated context, and read `Browser Use price: USD 49` from a live local page
- the shell block parses with `bash -n`; Markdown fences and create → ID → CDP → trap → gptme → cleanup order were checked
- `git diff --check`
- one file: `docs/browser.md`
- no credentials in the diff

## Test boundary

The real CDP path was tested with local Chromium. The live Browser Use Cloud create/connect/stop path was not run because no authorized `BROWSER_USE_API_KEY` is available locally. No production or customer credential was used.

### DIFF
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -144,8 +144,8 @@ cleanup_browser
 trap - EXIT INT TERM
 ```
 
-Create an API key in the [Browser Use Cloud
-dashboard](https://cloud.browser-use.com) and export it as
+Create an API key on the [Browser Use Cloud API key
+page](https://cloud.browser-use.com/settings?tab=api-keys&new=1) and export it as
 `BROWSER_USE_API_KEY` first. The managed browser includes a hardened Chromium
 build and residential proxy; gptme keeps using its existing Playwright tools.
 
--- a/docs/browser.md
+++ b/docs/browser.md
@@ -112,6 +112,34 @@ export GPTME_BROWSER_CDP_URL=http://127.0.0.1:9222
 gptme "read https://example.com"
 ```
 
+### Use Browser Use Cloud
+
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Thanks, @MagMueller — this turns the existing CDP support into a concrete managed-browser workflow without adding provider-specific code to gptme.

I rechecked head `225f289a5` locally:

```text
uv run --with requests --with playwright --with pytest --with pytest-mock \
  --with pytest-timeout --with pytest-retry --with pytest-asyncio \
  pytest tests/test_tools_browser_thread.py -q
54 passed in 6.07s

# extracted the documented shell block verbatim
bash -n /tmp/pr3629-browser-use.sh
# exit 0
```

The example uses `curl -f`, `jq -e`, and an `EXIT INT TERM` trap, so malformed session responses fail early and an allocated cloud browser is stopped on normal exit or interruption. I also verified that the trap is cleared after explicit cleanup. The remaining boundary is exactly the one documented in the PR: the live Browser Use API path needs an authorized key; the gptme CDP path and shell lifecycle are covered. Good docs contribution — and thanks for tightening cleanup and linking directly to API-key creation in follow-up commits.

PR #3630, opened 2026-08-26
fix(cli): forward nested help to utility shortcuts

TimeToBuildBob:
## Problem

The documented gptme utility shortcut did not forward nested help. `gptme chats list --help` exited successfully but rendered top-level gptme help instead of the `chats list` options because Click consumed the eager help flag before shortcut dispatch.

## Fix

Stop interspersed top-level option parsing when the first token is a known mirrored gptme-util command. Arguments after that command remain opaque and reach gptme-util unchanged. Normal chat parsing and leading gptme flags are unaffected.

## Verification

- `pytest tests/test_search_alias.py` — 20 passed
- targeted regression test — passed
- live `gptme chats list --help` now renders `Usage: gptme-util chats list [OPTIONS]`
- ruff check and format — clean
- mypy on changed files — clean

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -120,6 +120,10 @@ class _DynamicHelpCommand(click.Command):
             if skip_next:
                 skip_next = False
                 continue
+            if a == "--":
+                # End-of-options delimiter: don't intercept positionals after it.
+                # `gptme -- chats list` must NOT dispatch to gptme-util.
+                break
             if a.startswith("-"):
                 # --opt=val form carries its value inline; bare --opt consumes next token.
                 opt_name = a.split("=")[0] if "=" in a else a
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -100,10 +100,12 @@ class _DynamicHelpCommand(click.Command):
 
     def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
         """Keep arguments after a mirrored utility command opaque to Click."""
-        if args and not args[0].startswith("-"):
+        # Scan past any leading top-level flags to find the first positional.
+        first_positional = next((a for a in args if not a.startswith("-")), None)
+        if first_positional is not None:
             from .util import UTIL_SUBCOMMANDS
 
-            if args[0] in UTIL_SUBCOMMANDS:
+            if first_positional in UTIL_SUBCOMMANDS:
                 # Otherwise eager or recognized top-level options such as --help
                 # are consumed before main() can forward them to gptme-util.
                 ctx.allow_interspersed_args = False
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -98,6 +98,17 @@ class _DynamicHelpCommand(click.Command):
 
     _help_expanded = False
 
+    def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
+        """Keep arguments after a mirrored utility command opaque to Click."""
+        if args and not args[0].startswith("-"):
+            from .util import UTIL_SUBCOMMANDS
+
+            if args[0] in UTIL_SUBCOMMANDS:
+                # Otherwise eager or recognized top-level options such as --help
+                # are consumed before main() can forward them to gptme-util.
+                ctx.allow_interspersed_args = False
+        return super().parse_args(ctx, args)
+
     def _expand_dynamic_help(self) -> None:
         if self._help_expanded:
             return
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -121,8 +121,11 @@ class _DynamicHelpCommand(click.Command):
                 skip_next = False
                 continue
             if a == "--":
-                # End-of-options delimiter: don't intercept positionals after it.
-                # `gptme -- chats list` must NOT dispatch to gptme-util.
+                # End-of-options delimiter: stop scanning for a subcommand name.
+                # With allow_interspersed_args still True, Click treats everything
+                # after '--' as positionals.  main() will still see the first
+                # positional as prompts[0] and forward to gptme-util, so dispatch
+                # does occur for `gptme -- chats list --help`.
                 break
             if a.startswith("-"):
                 # --opt=val form carries its value inline; bare --opt consumes next token.
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -100,8 +100,35 @@ class _DynamicHelpCommand(click.Command):
 
     def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
         """Keep arguments after a mirrored utility command opaque to Click."""
-        # Scan past any leading top-level flags to find the first positional.
-        first_positional = next((a for a in args if not a.startswith("-")), None)
+        # Build the set of option names that consume a following value token
+        # (non-flag options).  These must be skipped over when scanning for
+        # the first true positional so that option *values* are not mistaken
+        # for subcommands (e.g. `--model gpt-4` must not yield 'gpt-4').
+        value_opts: set[str] = set()
+        for param in self.params:
+            if (
+                isinstance(param, click.Option)
+                and not param.is_flag
+                and param.nargs != 0
+            ):
+                value_opts.update(param.opts)
+
+        # Scan past leading option/value pairs to find the first positional.
+        skip_next = False
+        first_positional: str | None = None
+        for a in args:
+            if skip_next:
+                skip_next = False
+                continue
+            if a.startswith("-"):
+                # --opt=val form carries its value inline; bare --opt consumes next token.
+                opt_name = a.split("=")[0] if "=" in a else a
+                if "=" not in a and opt_name in value_opts:
+                    skip_next = True
+                continue
+            first_positional = a
+            break
+
         if first_positional is not None:
             from .util import UTIL_SUBCOMMANDS
 
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -880,7 +880,12 @@ def main(
 
     # Plugin dispatch: `gptme CMD [args...]` → `gptme-CMD [args...]` if installed
     # Enables extensibility: `gptme sessions` works if gptme-sessions is in PATH.
-    if prompts and not show_version:
+    # Suppressed after '--' for the same literal-prompt semantics as util dispatch.
+    if (
+        prompts
+        and not show_version
+        and not _ctx.meta.get("util_dispatch_suppressed", False)
+    ):
         plugin = f"gptme-{prompts[0]}"
         if plugin_path := shutil.which(plugin):
             sys.exit(subprocess.call([plugin_path, *prompts[1:]]))
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -145,19 +145,28 @@ class _DynamicHelpCommand(click.Command):
                         skip_next = True
                     continue
                 # Short option, possibly grouped (e.g. -vm where -m takes a value).
-                # Two forms have the value embedded in the token and do NOT consume
-                # the next token:
-                #   -t=read-only  → '=' separates the inline value
-                #   -mgpt-4       → value is embedded directly after the option char
-                # Only skip the next token when the value-taker is the last char
-                # in the cluster (e.g. -vm where -m needs the next token).
-                if "=" not in a:
-                    chars = a[1:]
-                    for idx, ch in enumerate(chars):
-                        if f"-{ch}" in value_opts:
-                            if idx == len(chars) - 1:
-                                skip_next = True
-                            break
+                # Inspect flags from left to right, as Click does.  The remainder
+                # of the token becomes an attached value as soon as a value-taking
+                # option is found; otherwise every character must be a known flag.
+                chars = a[1:]
+                for idx, ch in enumerate(chars):
+                    opt_name = f"-{ch}"
+                    if opt_name not in known_opts:
+                        # With ignore_unknown_options=True, Click preserves an
+                        # unknown short option as a positional argument.  Stop
+                        # here rather than treating a later utility name as the
+                        # first positional (e.g. `-x chats list --help`).
+                        first_positional = a
+                        break
+                    if opt_name in value_opts:
+                        # No characters after the option means its value is the
+                        # next token.  Otherwise the remainder (including an '='
+                        # prefix) is the attached value.
+                        if idx == len(chars) - 1:
+                            skip_next = True
+                        break
+                if first_positional is not None:
+                    break
                 continue
             first_positional = a
             break
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -143,21 +143,21 @@ class _DynamicHelpCommand(click.Command):
                         break
                     if "=" not in a and opt_name in value_opts:
                         skip_next = True
-                else:
-                    # Short option, possibly grouped (e.g. -vm where -m takes a value).
-                    # Two forms have the value embedded in the token and do NOT consume
-                    # the next token:
-                    #   -t=read-only  → '=' separates the inline value
-                    #   -mgpt-4       → value is embedded directly after the option char
-                    # Only skip the next token when the value-taker is the last char
-                    # in the cluster (e.g. -vm where -m needs the next token).
-                    if "=" not in a:
-                        chars = a[1:]
-                        for idx, ch in enumerate(chars):
-                            if f"-{ch}" in value_opts:
-                                if idx == len(chars) - 1:
-                                    skip_next = True
-                                break
+                    continue
+                # Short option, possibly grouped (e.g. -vm where -m takes a value).
+                # Two forms have the value embedded in the token and do NOT consume
+                # the next token:
+                #   -t=read-only  → '=' separates the inline value
+                #   -mgpt-4       → value is embedded directly after the option char
+                # Only skip the next token when the value-taker is the last char
+                # in the cluster (e.g. -vm where -m needs the next token).
+                if "=" not in a:
+                    chars = a[1:]
+                    for idx, ch in enumerate(chars):
+                        if f"-{ch}" in value_opts:
+                            if idx == len(chars) - 1:
+                                skip_next = True
+                            break
                 continue
             first_positional = a
             break
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -141,11 +141,19 @@ class _DynamicHelpCommand(click.Command):
                         skip_next = True
                 else:
                     # Short option, possibly grouped (e.g. -vm where -m takes a value).
-                    # If any char in the group is a value consumer, skip the next token.
-                    for ch in a[1:]:
-                        if f"-{ch}" in value_opts:
-                            skip_next = True
-                            break
+                    # Two forms have the value embedded in the token and do NOT consume
+                    # the next token:
+                    #   -t=read-only  → '=' separates the inline value
+                    #   -mgpt-4       → value is embedded directly after the option char
+                    # Only skip the next token when the value-taker is the last char
+                    # in the cluster (e.g. -vm where -m needs the next token).
+                    if "=" not in a:
+                        chars = a[1:]
+                        for idx, ch in enumerate(chars):
+                            if f"-{ch}" in value_opts:
+                                if idx == len(chars) - 1:
+                                    skip_next = True
+                                break
                 continue
             first_positional = a
             break
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -105,13 +105,12 @@ class _DynamicHelpCommand(click.Command):
         # the first true positional so that option *values* are not mistaken
         # for subcommands (e.g. `--model gpt-4` must not yield 'gpt-4').
         value_opts: set[str] = set()
+        known_opts: set[str] = set()  # all known option names (flags + value-takers)
         for param in self.params:
-            if (
-                isinstance(param, click.Option)
-                and not param.is_flag
-                and param.nargs != 0
-            ):
-                value_opts.update(param.opts)
+            if isinstance(param, click.Option):
+                known_opts.update(param.opts)
+                if not param.is_flag and param.nargs != 0:
+                    value_opts.update(param.opts)
 
         # Scan past leading option/value pairs to find the first positional.
         skip_next = False
@@ -121,11 +120,10 @@ class _DynamicHelpCommand(click.Command):
                 skip_next = False
                 continue
             if a == "--":
-                # End-of-options delimiter: stop scanning for a subcommand name.
-                # With allow_interspersed_args still True, Click treats everything
-                # after '--' as positionals.  main() will still see the first
-                # positional as prompts[0] and forward to gptme-util, so dispatch
-                # does occur for `gptme -- chats list --help`.
+                # '--' is the POSIX end-of-options sentinel: the user wants
+                # everything after it treated as literal prompts, not as a
+                # utility shortcut.  Signal main() to skip util dispatch.
+                ctx.meta["util_dispatch_suppressed"] = True
                 break
             if a == "-":
                 # '-' is gptme's MULTIPROMPT_SEPARATOR, not an option prefix.
@@ -137,6 +135,12 @@ class _DynamicHelpCommand(click.Command):
                 if a.startswith("--"):
                     # Long option: --opt=val (inline) or --opt val (next token).
                     opt_name = a.split("=")[0]
+                    if opt_name not in known_opts:
+                        # Unknown long option: with ignore_unknown_options=True,
+                        # Click treats it as a positional.  Mirror that here so
+                        # the scan doesn't skip past the real first positional.
+                        first_positional = a
+                        break
                     if "=" not in a and opt_name in value_opts:
                         skip_next = True
                 else:
@@ -847,10 +851,14 @@ def main(
 
     # gptme-util subcommand mirroring: `gptme chats [...]` → `gptme-util chats [...]`
     # Any top-level gptme-util subcommand can be invoked without typing 'gptme-util'.
+    # Suppressed when '--' was used (parse_args sets util_dispatch_suppressed).
     if prompts and not show_version:
         from .util import UTIL_SUBCOMMANDS  # cheap: just a sorted list constant
 
-        if prompts[0] in UTIL_SUBCOMMANDS:
+        _ctx = click.get_current_context()
+        if prompts[0] in UTIL_SUBCOMMANDS and not _ctx.meta.get(
+            "util_dispatch_suppressed", False
+        ):
             if util_exec := shutil.which("gptme-util"):
                 sys.exit(subprocess.call([util_exec, *prompts]))
             else:
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -127,11 +127,25 @@ class _DynamicHelpCommand(click.Command):
                 # positional as prompts[0] and forward to gptme-util, so dispatch
                 # does occur for `gptme -- chats list --help`.
                 break
+            if a == "-":
+                # '-' is gptme's MULTIPROMPT_SEPARATOR, not an option prefix.
+                # Treat as a positional; it is never a UTIL_SUBCOMMAND so the
+                # scan stops here without enabling the gptme-util forward path.
+                first_positional = a
+                break
             if a.startswith("-"):
-                # --opt=val form carries its value inline; bare --opt consumes next token.
-                opt_name = a.split("=")[0] if "=" in a else a
-                if "=" not in a and opt_name in value_opts:
-                    skip_next = True
+                if a.startswith("--"):
+                    # Long option: --opt=val (inline) or --opt val (next token).
+                    opt_name = a.split("=")[0]
+                    if "=" not in a and opt_name in value_opts:
+                        skip_next = True
+                else:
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Automated merge handoff for `a4c047d4f8ec`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

gptme/cli/main.py
```diff
@@ -98,6 +98,17 @@ class _DynamicHelpCommand(click.Command):
 
     _help_expanded = False
 
+    def parse_args(self, ctx: click.Context, args: list[str]) -> list[str]:
+        """Keep arguments after a mirrored utility command opaque to Click."""
+        if args and not args[0].startswith("-"):
```
TimeToBuildBob: ⚠️ **P2** — The `parse_args` guard only triggers when `args[0]` is a known UTIL_SUBCOMMAND and does not start with a dash. If a leading top-level option precedes the subcommand (e.g. `gptme --verbose chats list --help`), `args[0]` is `--verbose`, the condition short-circuits, and `allow_interspersed_args` is never set to False. With interspersed args still enabled, Click parses the trailing `--help` as an eager option and renders the top-level gptme help, exiting before `main()` runs. The PR's goal of forwarding nested help to the utility command thus only works when the subcommand is the very first token; any leading flag silently reverts to the broken behavior. The comment in the code overstates the guarantee, claiming arguments after a mirrored utility command are kept opaque to Click, which is not the case when leading flags are present.

TimeToBuildBob:
## 🤖 AI code review

Adds a parse_args override in _DynamicHelpCommand that scans for the first positional token and, when it matches a known gptme-util subcommand, sets allow_interspersed_args=False so that options after the subcommand are treated as positionals. It also sets util_dispatch_suppressed in ctx.meta when '--' is encountered, and main() skips utility and plugin dispatch when that flag is set. New tests cover nested help forwarding and suppression after '--'.

<b>4 findings</b> · ❌ **1** P1 · ⚠️ **3** P2 · 🔒 **1** security — 🛑 **1 at P0/P1**

<details><summary><b>3 inline</b></summary>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `gptme/cli/main.py:119`

The parse_args scanner skips tokens when skip_next is True without checking for '--'. If a value-taking option is immediately before a '--' token, the '--' is consumed as the option's value and util_dispatch_suppressed is never set. For example, `gptme --model -- chats list` treats '--' as the model value, so main() sees prompts[0]=='chats' and dispatches to gptme-util, even though the user used '--' to force literal prompts. This bypasses the intended suppression mechanism. The fix should check for '--' before applying skip_next, or at least record that '--' was seen even when skipped.

<sub>**How this was verified:** Traced the loop: a value option sets skip_next=True, then the next token (which could be '--') is consumed by the skip_next branch without checking if it equals '--', so the flag is not set. Confirmed by reading lines 119-121 and the earlier skip_next assignment at line 144.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `tests/test_search_alias.py:305`

Tests `test_util_subcmd_after_unknown_option_does_not_dispatch` and `test_util_subcmd_after_unknown_short_option_does_not_dispatch` do not patch `shutil.which`. If a binary named `gptme--bogus` or `gptme-x` happens to exist in PATH during the test run, the plugin dispatch branch will call it, making `mock_call.assert_not_called()` fail. This makes the tests dependent on the local environment. They should patch `shutil.which` to return None for the plugin name, as other tests do.

```suggestion
with patch("gptme.cli.main.shutil.which", return_value=None):
```

<sub>**How this was verified:** Checked the test bodies; they only patch subprocess.call, and the plugin dispatch code calls shutil.which('gptme--bogus') and shutil.which('gptme-x') before attempting the call. Without a return value of None, an existing binary would be executed.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/cli/main.py:180`

The new `parse_args` sets `ctx.allow_interspersed_args = False` whenever the first positional is a mirrored utility subcommand (e.g. `chats`). This changes parsing of any tokens after that subcommand: options that were previously parsed as top-level gptme options are now treated as positional arguments and forwarded to gptme-util. For example, `gptme chats --verbose` previously started a chat with the prompt 'chats' and verbose output enabled; now it invokes `gptme-util chats --verbose`. Users who intended the word 'chats' as a literal prompt can no longer combine it with top-level gptme options placed after it; placing options before it (e.g. `gptme --verbose chats`) still works, but the positional order changes meaning. This is a behavioral regression for existing invocations that rely on interspersed options after a prompt that happens to be a known subcommand name.

<sub>**How this was verified:** Traced the parse_args loop: when `first_positional` is 'chats', it sets allow_interspersed_args=False. Click then treats `--verbose` as a positional argument, so prompts becomes ['chats', '--verbose'], and main's dispatch sends it to gptme-util.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`gptme/cli/main.py:848`](https://github.com/gptme/gptme/blob/d495d4ea3f64c5e65f97a5297f3551e0b5620d2f/gptme/cli/main.py#L848)

The `--` sentinel sets `util_dispatch_suppressed` in `ctx.meta` to suppress util and plugin dispatch, but the `search` alias does not consult that flag. As a result, `gptme -- search` still runs the `search` command (calls `search_chats`) instead of treating `search` as a literal prompt, violating the documented end-of-options semantics where everything after `--` is a prompt. This is inconsistent with the util and plugin dispatch blocks, which both check `_ctx.meta.get("util_dispatch_suppressed", False)`. The user intending to start a chat with the word "search" would unexpectedly execute a search operation instead.

```suggestion
if prompts and prompts[0] == "search" and not show_version and not _ctx.meta.get("util_dispatch_suppressed", False):
```

<sub>**How this was verified:** Verified that the search alias block (lines 848-859) does not reference `_ctx.meta` or `util_dispatch_suppressed`, while the util dispatch block at line 868 and plugin dispatch block at line 884 both do.</sub>

- `gptme/cli/main.py` — Adds parse_args override to scan for first positional and set allow_interspersed_args or util_dispatch_suppressed; main() checks util_dispatch_suppressed before dispatching to gptme-util or plugins.
- `tests/test_search_alias.py` — Adds tests for nested help forwarding with leading flags/value options, suppression after '--', and non-dispatch for unknown options.

</details>

| `4f66d8d150a2` | 4/5 | 1 | llm | 2026-08-26 04:51 UTC |
| `5800df1d8b4f` | 3/5 | 1 | llm | 2026-08-26 06:23 UTC |
| `4bc6302dc6df` | 4/5 | 1 | llm | 2026-08-26 07:43 UTC |
| `170feda5feaa` | 3/5 | 2 | llm | 2026-08-26 10:32 UTC |
| `a258d7dab3b7` | 3/5 | 2 | llm | 2026-08-26 11:04 UTC |
| `834d859a8345` | 3/5 | 1 | llm | 2026-08-26 12:26 UTC |
| `42569146a6f8` | 3/5 | 1 | llm | 2026-08-26 14:22 UTC |
| `14e23b15d4ed` | 4/5 | 1 | llm | 2026-08-26 15:41 UTC |
| `bdd4c8ff2d37` | 4/5 | 1 | llm | 2026-08-27 00:58 UTC |
| `574d14c46cf7` | 3/5 | 2 | llm | 2026-08-27 15:31 UTC |

</details>

<sub>Reviewed `d495d4ea3f64` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 929s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/main.py
TimeToBuildBob: Fixed in `4f66d8d`. Changed `args[0]` to scan past leading flags with `next((a for a in args if not a.startswith("-")), None)` so `gptme --verbose chats list --help` correctly forwards to gptme-util. Added a regression test covering this case.

TimeToBuildBob:
@TimeToBuildBob review

TimeToBuildBob:
Automated merge handoff for `4f66d8d150a2`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
TimeToBuildBob: Fixed in 5800df1. The scan now builds a set of all non-flag options from `self.params` and skips their following value token when walking `args`. For `--model gpt-4 chats list --help`, 'gpt-4' is now correctly skipped as an option value, so 'chats' is found as the first positional and allow_interspersed_args is set to False. Added a regression test covering this case (`test_util_subcmd_forwards_nested_help_with_leading_value_option`).

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -98,6 +98,50 @@ class _DynamicHelpCommand(click.Command):
...
+        # Scan past leading option/value pairs to find the first positional.
+        skip_next = False
+        first_positional: str | None = None
+        for a in args:
+            if skip_next:
+                skip_next = False
+                continue
+            if a == "--":
+                # End-of-options delimiter: don't intercept positionals after it.
+                # `gptme -- chats list` must NOT dispatch to gptme-util.
```
TimeToBuildBob: ⚠️ **P2** — The comment on line 125 claims that `gptme -- chats list` must NOT dispatch to gptme-util, but the actual behaviour (and the test `test_util_subcmd_after_double_dash_forwards_help`) is that it DOES dispatch. The scanner breaks on `--` and leaves `allow_interspersed_args` as True, so Click parses `chats` as the first positional, and main() sees prompts[0]=='chats' and forwards to gptme-util. The misleading comment could lead a maintainer to 'fix' the behaviour to not dispatch, which would break the documented shortcut and the passing test. The comment should say that dispatch does occur. _(2/2 passes)_

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
TimeToBuildBob: Fixed in 1f20362ac: updated the comment to accurately describe the dispatch behaviour. The comment now says dispatch *does* occur for `gptme -- chats list --help` since `allow_interspersed_args` remains True after `--`, so Click treats everything after as positionals and `main()` sees `chats` as the first prompt and forwards to gptme-util.

ErikBjare: This seems backwards, feels like you should indeed NOT dispatch/intercept positionals in the `--` case. (see treatment of `-` separators for prompt chaining)

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
TimeToBuildBob: You're right — applying the fix now. `--` should suppress the shortcut dispatch entirely, treating the trailing positionals as literal prompts (same intuition as `-` for prompt chaining).

Fix pushed in `170feda5`:
- `parse_args` records `ctx.meta["double_dash_seen"] = True` when `--` is encountered
- `main()` skips the gptme-util dispatch when that flag is set
- Test updated: `test_util_subcmd_after_double_dash_does_not_dispatch` now asserts the flag is set (tested at parse_args level to avoid spinning up a full session in the test runner)

gptme/cli/main.py
```diff
@@ -98,6 +98,53 @@ class _DynamicHelpCommand(click.Command):
...
+                skip_next = False
+                continue
+            if a == "--":
+                # End-of-options delimiter: stop scanning and record the flag.
+                # main() checks double_dash_seen to suppress gptme-util dispatch,
+                # treating `gptme -- chats list` as literal prompts (consistent
+                # with how '-' separates chained prompts in gptme syntax).
+                ctx.meta["double_dash_seen"] = True
+                break
+            if a.startswith("-"):
```
TimeToBuildBob, line 134: ⚠️ **P2** — The scan in `parse_args` treats a bare `-` as an option because it checks `a.startswith("-")`. In gptme, `-` is the multiprompt separator defined by MULTIPROMPT_SEPARATOR. When a user runs `gptme - chats --help`, the scan skips `-` and identifies `chats` as the first positional. Since `chats` is a UTIL_SUBCOMMAND, it sets `allow_interspersed_args=False`, causing `--help` to be treated as a positional argument rather than a gptme option. This misparses the command, grouping prompts incorrectly and potentially forwarding options meant for gptme into the user prompt. The scan should treat a lone `-` as a positional, not as an option flag. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -98,6 +98,53 @@ class _DynamicHelpCommand(click.Command):
...
+            first_positional = a
+            break
+
+        if first_positional is not None:
+            from .util import UTIL_SUBCOMMANDS
+
+            if first_positional in UTIL_SUBCOMMANDS:
+                # Otherwise eager or recognized top-level options such as --help
+                # are consumed before main() can forward them to gptme-util.
+                ctx.allow_interspersed_args = False
```
TimeToBuildBob, line 180: ⚠️ **P2** — The scan for `first_positional` treats any token not starting with `-` as the first positional, but with the CLI's `ignore_unknown_options=True` setting, tokens that start with `-` but are not recognized (e.g. `--foo`) are treated as positionals by Click. For a command like `gptme --foo --verbose chats`, the scan sees `chats` as the first positional and sets `ctx.allow_interspersed_args=False`. Click then, with `allow_interspersed_args=False`, treats `--foo` as the first positional and stops option processing, so `--verbose` is forwarded as a prompt instead of setting the verbose flag, and `chats` is no longer first, so gptme-util dispatch is skipped. The user's intended verbose flag is silently ignored and an extra prompt is added. _(1/3 passes — passes 1, 2 looked and disagreed)_

TimeToBuildBob:
Rebased onto current master (picking up #3635 which fixed ). That test failure was unrelated to this PR — the test was asserting by index (expecting 2 system message parts) but master now has more parts; #3635 fixed the assertion to check by content instead. CI should be clean now.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `834d859a8345`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

gptme/cli/main.py
```diff
@@ -98,6 +98,53 @@ class _DynamicHelpCommand(click.Command):
...
+                # End-of-options delimiter: stop scanning for a subcommand name.
+                # With allow_interspersed_args still True, Click treats everything
+                # after '--' as positionals.  main() will still see the first
+                # positional as prompts[0] and forward to gptme-util, so dispatch
+                # does occur for `gptme -- chats list --help`.
+                break
+            if a.startswith("-"):
+                # --opt=val form carries its value inline; bare --opt consumes next token.
+                opt_name = a.split("=")[0] if "=" in a else a
+                if "=" not in a and opt_name in value_opts:
```
TimeToBuildBob: ❌ **P1** — The scanning in `parse_args` does not handle grouped short options that include a value-taking option. For a token like `-vm`, Click parses it as `-v` (flag) and `-m` (value option), so the next token is the value for `-m`. However, the scanner treats the entire token `-vm` as a single option name, checks whether `'-vm' in value_opts` (which it is not), and thus does not skip the next token. That next token (e.g. `gpt-4`) is then misidentified as the first positional. Since it is not a known utility subcommand, `allow_interspersed_args` is left True, and Click later processes a trailing `--help` as an eager option, printing top-level help instead of forwarding to gptme-util. Concretely, `gptme -vm gpt-4 chats list --help` fails to forward. _(3/3 passes)_

TimeToBuildBob:
Fixed two edge cases in the `parse_args` scanner flagged by the AI reviewer:

**P1 (grouped short options)**: `-vm gpt-4 chats list --help` where `-m` takes a value — the scanner was checking the whole token `-vm` against `value_opts` (which only stores `-m`, not `-vm`), so `gpt-4` was not skipped and was misidentified as the first positional. Fixed by splitting long vs short option handling and iterating over each char in a grouped short option.

**P2 (bare `-`)**: The multiprompt separator `-` starts with `-` so the old scanner skipped it as an option prefix, then found the next token (`chats`) as the first positional and incorrectly set `allow_interspersed_args=False`, enabling gptme-util dispatch for what is actually a multi-prompt invocation. Fixed by treating bare `-` as a positional that stops the scan; since `-` is not a `UTIL_SUBCOMMAND`, dispatch is not enabled.

Added regression test for the grouped short option case (`-vm gpt-4 chats list --help`). All 24 tests pass locally.

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -98,6 +98,67 @@ class _DynamicHelpCommand(click.Command):
...
+            if a.startswith("-"):
+                if a.startswith("--"):
+                    # Long option: --opt=val (inline) or --opt val (next token).
+                    opt_name = a.split("=")[0]
+                    if "=" not in a and opt_name in value_opts:
+                        skip_next = True
+                else:
+                    # Short option, possibly grouped (e.g. -vm where -m takes a value).
+                    # If any char in the group is a value consumer, skip the next token.
+                    for ch in a[1:]:
```
TimeToBuildBob: ❌ **P1** — The scanner misclassifies short options with attached values (e.g. '-mgpt-4' or '-t=read-only') as option clusters that consume the next token. In the else branch at line 145, it iterates over every character after the leading '-', and if any char (like 'm' in '-m') is a value-taking option, it sets skip_next = True on lines 147-148. This causes the scanner to skip the next argument even though the value is already contained within the token. For 'gptme -mgpt-4 chats list --help', the scanner skips 'chats', finds 'list' as the first positional, and because 'list' is not in UTIL_SUBCOMMANDS it never sets allow_interspersed_args = False. Click then parses '--help' as a top-level option, so the nested help is never forwarded to gptme-util and the user sees gptme's help instead. The same happens with '-t=read-only'. The fix is to detect when a short option token has an attached value (either via '=' or because the option takes a value and the rest of the token is that value) and skip only the attached value, not the next token. _(3/3 passes)_

TimeToBuildBob:
Fixed the remaining P1 from the AI reviewer: short options with inline attached values.

**The bug**: for `-mgpt-4 chats list --help`, the scanner iterated over every char in the `-m…` cluster and set `skip_next=True` when it found `-m` in `value_opts` — even though the value (`gpt-4`) was embedded inline in the same token. This caused the scanner to skip `chats` (the real first positional), find `list` instead, and since `list` is not a `UTIL_SUBCOMMAND`, `allow_interspersed_args` was never set to False. Click then consumed `--help` eagerly, showing top-level gptme help instead of forwarding to gptme-util.

**The fix**: only set `skip_next=True` when the value-taking char is the **last** char in the cluster (meaning the value must come from the next token). If more chars follow in the token (i.e. the value is embedded inline), skip_next stays False. Also added an early-exit guard for the `-t=read-only` (= inline) form.

Regression test added: `gptme -mgpt-4 chats list --help` now correctly calls `gptme-util chats list --help`. All 25 tests pass.

TimeToBuildBob:
Automated merge handoff for `14e23b15d4ed`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

gptme/cli/main.py
```diff
@@ -98,6 +98,75 @@ class _DynamicHelpCommand(click.Command):
...
+                # positional as prompts[0] and forward to gptme-util, so dispatch
+                # does occur for `gptme -- chats list --help`.
+                break
+            if a == "-":
+                # '-' is gptme's MULTIPROMPT_SEPARATOR, not an option prefix.
+                # Treat as a positional; it is never a UTIL_SUBCOMMAND so the
+                # scan stops here without enabling the gptme-util forward path.
+                first_positional = a
+                break
+            if a.startswith("-"):
```
TimeToBuildBob, line 134: ⚠️ **P2** — The scanner assumes any token starting with '-' is an option, but with ignore_unknown_options=True, unknown options are treated as positional arguments by Click. This can cause the scanner to miss the actual first positional. For example, running `gptme --bogus chats list --help` makes the scanner skip '--bogus' as an option, then see 'chats' as the first positional, setting allow_interspersed_args=False. Click then parses '--bogus' as a positional, so the first positional is '--bogus' (not 'chats'), and because allow_interspersed_args is False, it stops parsing options, turning '--help' into a positional. The result is no dispatch to gptme-util and no help output; instead the CLI enters chat mode with the tokens as prompts. This is a regression from previous behavior where `gptme --bogus chats list --help` would show top-level help. The fix should account for unknown options being positionals, perhaps by scanning for the first token that is neither a known option nor a value. _(3/3 passes)_

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
TimeToBuildBob: Fixed in `42569146a`: bare `-` is now handled explicitly at lines 130–135. The scan treats it as a positional that stops further scanning. Since `-` is never a `UTIL_SUBCOMMAND`, `allow_interspersed_args` is not set to False and dispatch is not enabled for `gptme - chats --help`.

gptme/cli/main.py
TimeToBuildBob: Fixed in `42569146a`: the scanner now iterates over each character in a grouped cluster (e.g. `-vm`). It only sets `skip_next=True` when the value-taking char is the **last** char in the cluster — meaning the value must come from the next token. For `-vm gpt-4 chats list --help`, `m` is the last char, `skip_next=True` is set, `gpt-4` is skipped, and `chats` is correctly found as the first positional.

gptme/cli/main.py
TimeToBuildBob: Fixed in `14e23b15d`: for `-mgpt-4`, the scanner sees `m` at index 0 and `len(chars)-1 == 5`, so `idx != len(chars)-1` → `skip_next` stays False. `chats` is then correctly found as the first positional. For `-t=read-only`, the `"=" not in a` guard short-circuits the whole cluster scan, so no token is skipped.

gptme/cli/main.py
TimeToBuildBob: Fixed in `d8df08c8a`: the scanner now builds a `known_opts` set from all params. When it sees `--bogus` (not in `known_opts`), it sets `first_positional = '--bogus'` and breaks — matching what Click does with `ignore_unknown_options=True`. Since `--bogus` is not a `UTIL_SUBCOMMAND`, `allow_interspersed_args` stays True and `--help` fires as an eager option. Regression test added.

gptme/cli/main.py
TimeToBuildBob: The 3/3 sibling finding (`4ffda7bede53`) on the same topic was addressed in `d8df08c8a` — unknown long options now stop the scan and leave `allow_interspersed_args` unchanged. This 1/3 finding describes a similar shape; the fix covers the well-supported case.

gptme/cli/main.py
TimeToBuildBob: The actual behavior fix is now in `d8df08c8a`: `parse_args` sets `ctx.meta["util_dispatch_suppressed"] = True` when it encounters `--`, and `main()` checks that flag before the `UTIL_SUBCOMMANDS` dispatch. Test updated from "forwards" to "suppresses" (using `make_context` to verify the meta flag without triggering LLM init).

TimeToBuildBob:
Fixed the remaining open AI reviewer findings in `d8df08c8a`:

**`--` dispatch suppression** (Erik's request):
- `parse_args` now sets `ctx.meta["util_dispatch_suppressed"] = True` when `--` appears
- `main()` checks the flag before the `UTIL_SUBCOMMANDS` dispatch
- `gptme -- chats list --help` is no longer forwarded to gptme-util; the tokens after `--` are treated as literal prompts
- Test renamed and rewritten to assert suppression (uses `make_context` to test `parse_args` directly without triggering LLM init)

**Unknown long options** (P2, 3/3 passes):
- Scanner now builds `known_opts` from all `self.params`
- Unknown `--foo` (not in `known_opts`) is treated as the first positional rather than skipped as an option — matching what Click does with `ignore_unknown_options=True`
- `allow_interspersed_args` stays True, so `--help` fires as an eager option
- Regression test added for `gptme --bogus chats list --help`

All previously fixed issues (P1 grouped short options, P1 inline attached values, P2 bare `-`, P2 leading flags/values) remain in place. 26 tests pass.

TimeToBuildBob:
CI failure is two unrelated flaky timeout tests (`test_create_template_mode_e2e` and `test_install_nonexistent_path`) — both hit their timeout limits, not caused by this PR's changes. Triggered a re-run of the failed jobs.

TimeToBuildBob:
Automated merge handoff for `bdd4c8ff2d37`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

TimeToBuildBob:
CI is now fully green (the two flaky timeout tests from yesterday re-ran cleanly). All review findings from the AI reviewer and from Erik's inline comment (-- separator behaviour) have been addressed. PR is mergeable — ready for maintainer merge.

TimeToBuildBob:
Fixed the remaining AI-review finding in `574d14c46`.

Unknown short options now mirror unknown long options under Click's `ignore_unknown_options=True`: the scanner stops at the unknown token instead of scanning ahead to `chats` and incorrectly enabling utility dispatch. The short-option scan now also validates every member of a known grouped cluster while preserving attached values (`-mgpt-4`) and next-token values (`-vm gpt-4`).

Added `gptme -x chats list --help` regression coverage. Verified the full utility-mirroring test class (15 passed), the six option-scanning regressions, Ruff check, and Ruff format. The in-band wrapper pushed the commit; its local reviewer was skipped because `bobutils` was unavailable in that worktree, so the normal post-push AI review will provide the final verdict.

TimeToBuildBob:
Automated merge handoff for `574d14c46cf7`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

TimeToBuildBob:
Fixed the latest P1 in `d495d4ea3`: the `--` literal-prompt sentinel now suppresses both mirrored `gptme-util` dispatch and generic `gptme-*` plugin dispatch. Previously `gptme -- sessions` could still execute `gptme-sessions`, despite the suppression semantics Erik requested.

Added a regression test proving `gptme -- sessions` reaches normal prompt handling and never invokes the plugin subprocess. Verification: all 28 tests in `tests/test_search_alias.py` pass; scoped Ruff, Ruff format, and pre-commit mypy pass. The required in-band reviewer was attempted but timed out, so its documented fail-open path pushed the verified fix; normal post-push AI review and CI are now pending.

TimeToBuildBob:
Automated merge handoff for `d495d4ea3f64`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

gptme/cli/main.py
```diff
@@ -98,6 +98,88 @@ class _DynamicHelpCommand(click.Command):
...
+            if isinstance(param, click.Option):
+                known_opts.update(param.opts)
+                if not param.is_flag and param.nargs != 0:
+                    value_opts.update(param.opts)
+
+        # Scan past leading option/value pairs to find the first positional.
+        skip_next = False
+        first_positional: str | None = None
+        for a in args:
+            if skip_next:
```
TimeToBuildBob, line 119: ⚠️ **P2** — The parse_args scanner skips tokens when skip_next is True without checking for '--'. If a value-taking option is immediately before a '--' token, the '--' is consumed as the option's value and util_dispatch_suppressed is never set. For example, `gptme --model -- chats list` treats '--' as the model value, so main() sees prompts[0]=='chats' and dispatches to gptme-util, even though the user used '--' to force literal prompts. This bypasses the intended suppression mechanism. The fix should check for '--' before applying skip_next, or at least record that '--' was seen even when skipped.

tests/test_search_alias.py
```diff
@@ -164,6 +178,145 @@ def test_util_subcmd_passes_all_args(self, runner: CliRunner):
...
+    ):
+        """gptme --bogus chats list --help shows gptme help, not gptme-util.
+
+        With ignore_unknown_options=True, Click treats '--bogus' as a positional.
+        The scanner mirrors this: unknown long options stop the scan without
+        enabling util dispatch, so allow_interspersed_args stays True and
+        --help fires as an eager option showing gptme's top-level help.
+        """
+        with patch("gptme.cli.main.subprocess.call") as mock_call:
+            result = runner.invoke(main, ["--bogus", "chats", "list", "--help"])
```
TimeToBuildBob, line 305: ⚠️ **P2** — Tests `test_util_subcmd_after_unknown_option_does_not_dispatch` and `test_util_subcmd_after_unknown_short_option_does_not_dispatch` do not patch `shutil.which`. If a binary named `gptme--bogus` or `gptme-x` happens to exist in PATH during the test run, the plugin dispatch branch will call it, making `mock_call.assert_not_called()` fail. This makes the tests dependent on the local environment. They should patch `shutil.which` to return None for the plugin name, as other tests do.

ErikBjare:
@TimeToBuildBob Decide whether to fix remaining issues in a follow-up, create one if so

TimeToBuildBob:
Opened gptme/gptme#3670 for the two real leftovers:

- `gptme -- search` still bypassed the literal-prompt semantics and invoked the built-in search alias.
- The unknown-option tests depended on whether `gptme--bogus` / `gptme-x` happened to exist in `PATH`.

I did **not** implement the reviewer's `--model -- chats list` suggestion: Click consumes that `--` as the value of `--model`, so it is not an end-of-options sentinel in that command. The follow-up is focused on the actual dispatch inconsistency and test isolation; 29 targeted tests pass.

PR #3631, opened 2026-08-26
docs(getting-started): add free cloud providers section

TimeToBuildBob:
## Summary

- Adds a **Free Cloud Providers (No Credit Card Required)** section to `docs/getting-started.rst`
- Documents OpenRouter (recommended, browser OAuth via `/account setup`), Google Gemini free tier, and Groq free tier
- Adds `.. _local-models:` anchor so the new tip cross-references the existing Ollama section

## Motivation

The current getting-started doc covers local Ollama models as a no-API-key path, but omits free cloud providers entirely. New users drop off when they see "API key required" without knowing that free tiers with no credit card exist. OpenRouter in particular has a browser-based OAuth flow already wired into gptme (`/account setup openrouter`), making it the lowest-friction zero-cost starting point and removing the most common onboarding objection before it forms.

## What changed

`docs/getting-started.rst`: 48 lines added after the existing Ollama section:
- OpenRouter: `/account setup openrouter` + `:free`-suffix model examples  
- Gemini: free API key at aistudio.google.com + `gemini/gemini-2.5-flash`
- Groq: free API key at console.groq.com + `groq/llama-3.3-70b-versatile`
- Tip contrasting free-cloud vs local-private tradeoffs

## Test plan

- [ ] `make docs` builds without warnings
- [ ] `:ref:\`local models <local-models>\`` cross-reference resolves correctly

### DIFF
--- a/docs/getting-started.rst
+++ b/docs/getting-started.rst
@@ -84,6 +84,8 @@ Here are some compelling examples to get you started:
     # Resume conversations
     gptme -r
 
+.. _local-models:
+
 Local Models (No API Key Required)
 -----------------------------------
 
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Self-merge attempted but blocked by the internal consensus gate (score 2/5). The gate's concern is a false positive:

> "The `.. _local-models:` anchor is placed before the new free-cloud section instead of the existing Ollama section"

The anchor is actually placed correctly — the diff shows it added immediately before the existing **"Local Models (No API Key Required)"** heading (line ~84), which is the Ollama section. The new "Free Cloud Providers" section comes 40+ lines later. The cross-reference `:ref:\`local models <local-models>\`` in the tip resolves correctly to the Ollama section.

**Status**: CI ✅ green, Greptile ✅ passed, `mergeable: CLEAN`. The only blocker is the internal consensus gate false positive. Needs manual merge or override.

PR #3638, opened 2026-08-26
feat(memory): add memory write tool for cross-runtime memory sharing

TimeToBuildBob:
## Summary

- Adds a `memory` tool (`gptme/tools/memory.py`) that saves persistent memories to the shared Claude Code memory path (`~/.claude/projects/<workspace-hash>/memory/`)
- Memories saved via gptme are immediately readable by future sessions on any runtime (CC, gptme, Codex) — CC reads them natively, gptme loads them via `prompt_workspace()` (already shipped in #3626)
- Each memory is stored as an individual `.md` file with YAML frontmatter (CC-compatible format) plus an entry in the `MEMORY.md` index
- 27 new tests covering: slug normalization, index create/append/update, save/overwrite, frontmatter format, execute error cases, and tool spec

## Background

PR #3626 (already merged) implemented the **read** side of #3625: gptme now auto-loads `~/.claude/projects/<workspace-hash>/memory/MEMORY.md` in workspace context. But the **write** side was missing — gptme had no way to save memories that CC (or other runtimes) could later read.

This PR completes the feature by adding the write side.

## Usage

```
memory prefer-short-answers
User prefers concise, direct answers without preamble.
More context: applies to all responses, especially code explanations.
```

The first content line becomes the MEMORY.md index description; remaining lines are the body. Names are slugified to safe filenames.

## Test plan

- [x] 27 unit tests in `tests/test_memory_tool.py` — all passing
- [x] 12 existing CC memory tests (`tests/test_cc_memory.py`) still passing
- [x] Pre-commit hooks (ruff, mypy) pass

Completes #3625 (the write side — #3626 shipped the read side)

### DIFF
--- a/gptme/tools/memory.py
+++ b/gptme/tools/memory.py
@@ -1,0 +1,165 @@
+"""
+Persistent memory tool for gptme.
+
+Saves memories to the shared Claude Code memory path so they are readable
+by future sessions on any runtime (CC, gptme, Codex).
+
+Memory files are written to:
+  ~/.claude/projects/<workspace-hash>/memory/<name>.md
+
+And an entry is added to MEMORY.md (the index file), which is auto-loaded
+by gptme (via prompt_workspace) and by Claude Code in every session.
+
+Usage:
+  memory save <name>
+  <description line>
+  <body content>
+"""
+
+import logging
+import re
+from collections.abc import Generator
+from pathlib import Path
+
+from ..dirs import get_cc_memory_dir, get_workspace
+from ..message import Message
+from .base import ToolSpec, ToolUse
+
+logger = logging.getLogger(__name__)
+
+instructions = """
+Save a persistent memory that will be automatically loaded in future sessions
+across all runtimes (gptme, Claude Code, Codex).
+
+Memories are stored in the shared Claude Code memory directory and loaded
+automatically when a session starts in this workspace.
+
+Use memories to remember:
+- Important facts about the user or project
+- Patterns and preferences learned from the conversation
+- Decisions and their rationale
+
+The first line of the content is treated as the one-line description
+(shown in the MEMORY.md index); the rest is the memory body.
+""".strip()
+
+instructions_format = {
+    "markdown": "Use a code block with the language tag `memory <name>` to save a memory. The first line is the description; the rest is the body.",
+}
+
+
+def examples(tool_format):
+    return f"""
+> User: remember that I prefer short answers
+> Assistant:
+{ToolUse("memory", ["prefer-short-answers"], "User prefers short, direct answers without preamble.").to_output(tool_format)}
+> System: Memory 'prefer-short-answers' saved.
+""".strip()
+
+
+def _slugify(name: str) -> str:
+    """Convert a name to a safe filename slug."""
+    slug = name.lower().strip()
+    slug = re.sub(r"[^a-z0-9]+", "-", slug)
+    slug = slug.strip("-")
+    return slug or "memory"
+
+
+def _update_memory_index(
+    memory_dir: Path, name: str, filename: str, description: str
+) -> None:
+    """Add or update an entry in MEMORY.md."""
+    index_path = memory_dir / "MEMORY.md"
+    entry = f"- [{name}]({filename}) — {description}\n"
+
+    if not index_path.exists():
+        index_path.write_text(f"# Persistent Memory\n\n{entry}", encoding="utf-8")
+        return
+
+    content = index_path.read_text(encoding="utf-8")
+    # Update existing entry for this name/file if present
+    pattern = rf"^- \[{re.escape(name)}\]\({re.escape(filename)}\).*$"
+    if re.search(pattern, content, re.MULTILINE):
+        content = re.sub(pattern, entry.rstrip(), content, flags=re.MULTILINE)
+        index_path.write_text(content, encoding="utf-8")
+    else:
+        # Append new entry
+        if not content.endswith("\n"):
+            content += "\n"
+        index_path.write_text(content + entry, encoding="utf-8")
+
+
+def save_memory(name: str, content: str, workspace: Path | None = None) -> str:
+    """Save a memory to the shared CC memory directory.
+
+    Args:
+        name: Memory name (used as filename slug and index key).
+        content: Memory content. First line is the description; rest is the body.
+        workspace: Workspace root (defaults to get_workspace()).
+
+    Returns:
+        Path to the saved memory file as a string.
+    """
+    if workspace is None:
+        workspace = get_workspace()
+
+    memory_dir = get_cc_memory_dir(workspace)
+    memory_dir.mkdir(parents=True, exist_ok=True)
+
+    slug = _slugify(name)
+    filename = f"{slug}.md"
+    file_path = memory_dir / filename
+
+    lines = content.strip().splitlines()
+    description = lines[0].strip() if lines else name
+    body = "\n".join(lines[1:]).strip() if len(lines) > 1 else content.strip()
+
+    frontmatter = f"---\nname: {slug}\ndescription: {description}\nmetadata:\n  type: general\n---\n\n"
+    file_path.write_text(frontmatter + body + "\n", encoding="utf-8")
+
+    _update_memory_index(memory_dir, name, filename, description)
+
+    logger.debug(f"Saved memory '{name}' to {file_path}")
+    return str(file_path)
+
+
+def execute_memory(
+    code: str | None,
+    args: list[str] | None,
+    kwargs: dict[str, str] | None = None,
+) -> Generator[Message, None, None]:
+    """Execute the memory tool — save a memory entry."""
+    if not args:
+        yield Message(
+            "system",
+            "Error: memory tool requires a name argument, e.g. `memory my-memory-name`",
+        )
+        return
+
+    name = args[0]
+
+    content = (code or "").strip()
+    if not content:
+        yield Message(
+            "system",
+            "Error: memory content is empty. Provide the memory text in the code block body.",
+        )
+        return
+
+    try:
+        file_path = save_memory(name, content)
+        yield Message("system", f"Memory '{name}' saved to `{file_path}`.")
+    except OSError as e:
+        yield Message("system", f"Error saving memory '{name}': {e}")
+
+
+tool = ToolSpec(
+    name="memory",
+    desc="Save a persistent memory to the shared cross-runtime memory store (readable by CC, gptme, Codex).",
+    instructions=instructions,
+    instructions_format=instructions_format,
+    examples=examples,
+    execute=execute_memory,
+    block_types=["memory"],
+    parameters=[],
+)
--- a/gptme/tools/memory.py
+++ b/gptme/tools/memory.py
@@ -109,8 +109,9 @@ def _update_memory_index(
             else:
                 pattern = rf"^- \[{re.escape(slug)}\]\({re.escape(filename)}\).*$"
                 if re.search(pattern, content, re.MULTILINE):
+                    replacement = entry.rstrip()
                     new_content = re.sub(
-                        pattern, entry.rstrip(), content, flags=re.MULTILINE
+                        pattern, lambda _: replacement, content, flags=re.MULTILINE
                     )
                 else:
                     if not content.endswith("\n"):
@@ -148,7 +149,8 @@ def save_memory(name: str, content: str, workspace: Path | None = None) -> str:
     description = lines[0].strip() if lines else name
     body = "\n".join(lines[1:]).strip() if len(lines) > 1 else content.strip()
 
-    frontmatter = f"---\nname: {slug}\ndescription: {description}\nmetadata:\n  type: general\n---\n\n"
+    safe_desc = description.replace("\\", "\\\\").replace('"', '\\"')
+    frontmatter = f'---\nname: {slug}\ndescription: "{safe_desc}"\nmetadata:\n  type: general\n---\n\n'
     file_path.write_text(frontmatter + body + "\n", encoding="utf-8")
 
     # Use slug as the index key so colliding names update the same entry.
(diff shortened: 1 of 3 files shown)


gptme/tools/memory.py
```diff
@@ -0,0 +1,165 @@
...
+        Path to the saved memory file as a string.
+    """
+    if workspace is None:
+        workspace = get_workspace()
+
+    memory_dir = get_cc_memory_dir(workspace)
+    memory_dir.mkdir(parents=True, exist_ok=True)
+
+    slug = _slugify(name)
+    filename = f"{slug}.md"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f9a08aded4f0.

`_update_memory_index` now takes `slug` as the key (not the original name), and `save_memory` passes `slug` rather than `name` to the index. Distinct names that normalise to the same slug now update a single index entry instead of creating duplicates. Added `test_slug_collision_produces_one_index_entry` and `test_index_uses_slug_not_raw_name` to guard this.

gptme/tools/memory.py
```diff
@@ -0,0 +1,165 @@
...
+    # Update existing entry for this name/file if present
+    pattern = rf"^- \[{re.escape(name)}\]\({re.escape(filename)}\).*$"
+    if re.search(pattern, content, re.MULTILINE):
+        content = re.sub(pattern, entry.rstrip(), content, flags=re.MULTILINE)
+        index_path.write_text(content, encoding="utf-8")
+    else:
+        # Append new entry
+        if not content.endswith("\n"):
+            content += "\n"
+        index_path.write_text(content + entry, encoding="utf-8")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f9a08aded4f0.

`_update_memory_index` now uses `fcntl.flock(LOCK_EX)` around the full read-modify-write. The file is opened in `a+` mode (creates when absent), locked exclusively, read, modified, and written back — all within the same open handle so no other process can interleave. Falls back to a no-op on Windows (best-effort, since the CC memory path doesn't exist there anyway).

gptme/tools/memory.py
```diff
@@ -0,0 +1,165 @@
...
+        return
+
+    name = args[0]
+
+    content = (code or "").strip()
+    if not content:
+        yield Message(
+            "system",
+            "Error: memory content is empty. Provide the memory text in the code block body.",
+        )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f9a08aded4f0.

`execute_memory` now routes through `execute_with_confirmation`, which previews the memory content and asks for user review before persisting. In autonomous/non-interactive mode (no confirmation hook registered) it auto-confirms with `default_confirm=True`, so existing tests and autonomous runs are unaffected.

gptme/tools/memory.py
```diff
@@ -0,0 +1,165 @@
...
+Save a persistent memory that will be automatically loaded in future sessions
+across all runtimes (gptme, Claude Code, Codex).
+
+Memories are stored in the shared Claude Code memory directory and loaded
+automatically when a session starts in this workspace.
+
+Use memories to remember:
+- Important facts about the user or project
+- Patterns and preferences learned from the conversation
+- Decisions and their rationale
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f9a08aded4f0.

The instructions are now framed around concrete agent decisions ("save this when you learn something worth remembering for future work") with examples of the three main use cases, rather than documenting storage and loading mechanics.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Re: slug collision — this is an intentional design choice, not a bug.

The slug is the **canonical identifier** for a memory slot. Two names that normalise to the same slug (`"prefer short answers"` and `"prefer-short answers"`) are considered the same memory — writing either one updates that slot. This is consistent with how CC's own memory system works: each memory lives at a stable slug-based path, and saving again under the same slug is an update.

The alternative (collision error or index disambiguation) would make the UX worse with no benefit: if you want a *different* memory, choose a name that produces a different slug. The test `test_slug_collision_produces_one_index_entry` documents this intentional behavior.

The three P1 findings (index duplicates, concurrent race, confirmation bypass) are addressed in f9a08aded4f0. Leaving the score at 4/5 for human/maintainer review.

gptme/tools/memory.py
```diff
@@ -0,0 +1,225 @@
...
+        try:
+            f.seek(0)
+            content = f.read()
+            if not content:
+                new_content = f"# Persistent Memory\n\n{entry}"
+            else:
+                pattern = rf"^- \[{re.escape(slug)}\]\({re.escape(filename)}\).*$"
+                if re.search(pattern, content, re.MULTILINE):
+                    new_content = re.sub(
+                        pattern, entry.rstrip(), content, flags=re.MULTILINE
```
TimeToBuildBob: ❌ **P1** — The index update uses `re.sub(pattern, entry.rstrip(), ...)` where `entry` includes the user-controlled `description` (first line of memory content). If `description` contains a backslash followed by a digit (e.g., `\1`) or a `\g<...>` reference, `re.sub` interprets it as a group reference. Since the pattern has no capturing groups, Python raises `re.error: invalid group reference`. This exception is not caught in `_update_memory_index`, nor is it caught by `_do_save` (which only catches `OSError`), so the tool crashes with an unhandled error and no memory is saved. A user can trigger this by saving a memory whose first line contains such a sequence, causing a denial of service of the memory tool.

gptme/tools/memory.py
```diff
@@ -0,0 +1,225 @@
...
+
+    slug = _slugify(name)
+    filename = f"{slug}.md"
+    file_path = memory_dir / filename
+
+    lines = content.strip().splitlines()
+    description = lines[0].strip() if lines else name
+    body = "\n".join(lines[1:]).strip() if len(lines) > 1 else content.strip()
+
+    frontmatter = f"---\nname: {slug}\ndescription: {description}\nmetadata:\n  type: general\n---\n\n"
```
TimeToBuildBob: ⚠️ **P2** — The frontmatter written to each memory file inserts the user-controlled `description` directly into a YAML value: `description: {description}`. If the description contains a colon followed by a space (e.g., "Remember: use tabs"), the resulting line is not valid YAML because the second colon is interpreted as a mapping separator. When Claude Code (or any YAML parser) loads the memory file, it will fail to parse the frontmatter, making the memory unreadable and effectively losing the saved content. This is triggered by a normal user input that happens to include a colon and space, leading to silent data loss.

TimeToBuildBob:
## 🤖 AI code review

Adds a new `memory` tool that saves persistent memories to the shared Claude Code memory directory, writing markdown files with YAML frontmatter and updating a MEMORY.md index. Adds a new test file covering the tool and adjusts the prompt token ceiling in test_prompts.py.

**❌ P1 high** — `gptme/tools/memory.py:154`

The memory file is written via `file_path.write_text()` at line 154 without any file lock, while the index update in `_update_memory_index()` is guarded by an exclusive `flock`. If two gptme sessions or gptme and Claude Code save memories with the same slug concurrently, both processes open the same `.md` file with truncation and write. The writes are not atomic, so the final file can contain interleaved content or be truncated by the other's open, and the index entry (updated under the lock) may describe a different version than the file on disk. The next session that reads MEMORY.md will display a description that does not match the actual memory body, and the memory content may be corrupted. Because the lock is only on the index, the race is unsolved.

```suggestion
Wrap the file write and the index update in the same lock, e.g., lock the memory file itself before writing, or perform the write and index update under a single lock for that workspace.
```

<sub>**How this was verified:** Verified that _update_memory_index uses _lock_exclusive around the index write, but the memory file write in save_memory is not protected by any lock.</sub>

</details>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/tools/memory.py:1`

This appears to overlap with `gptme/gptme-contrib:packages/gptme-cc-memory` which already provides Claude Code memory extraction and retrieval. Matched signals: claude, memories, memory. Informational only — not a blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement.

</details>

- `gptme/tools/memory.py` — New file implementing the memory tool with slugification, file writing, index updates, and the execute entry point.
- `tests/test_memory_tool.py` — New test file covering slugify, index updates, saving memories, execute error cases, and tool spec.
- `tests/test_prompts.py` — Raises the upper token bound in test_get_prompt_full from 8000 to 8100.

</details>

| `f9a08aded4f0` | 3/5 | 2 | llm | 2026-08-26 16:23 UTC |

</details>

<sub>Reviewed `7bb89e6c544c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 129s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/memory.py
```diff
@@ -0,0 +1,225 @@
...
+    file_path.write_text(frontmatter + body + "\n", encoding="utf-8")
```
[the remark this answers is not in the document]
TimeToBuildBob: The collision-case update addresses the INDEX (so two names normalising to the same slug produce one index entry rather than two). You are right that the *backing file* was still overwritten unconditionally — that part was unaddressed.

Slug collision is an intentional design choice here: the slug IS the memory identity, so two names that normalise the same way are the same slot. That said, the `execute_with_confirmation` path (added in the previous commit) already previews the target path before writing, so a user in interactive mode sees exactly which file will be overwritten and can cancel. In autonomous mode the agent chose that name, so the overwrite is deliberate.

If Greptile's concern is silent data loss in a truly blind case: the confirmation step is the guard. I'll leave the design as-is and note this in the PR description rather than adding extra file-existence checks that would complicate the flow.

gptme/tools/memory.py
TimeToBuildBob: Fixed in 7bb89e6c54. The replacement string in `re.sub` is now wrapped in a lambda — `lambda _: replacement` — so Python never interprets the content as a replacement template. User-controlled descriptions with `\1`, `\g<name>`, or any other backslash sequence no longer trigger `re.error`.

gptme/tools/memory.py
TimeToBuildBob: Fixed in 7bb89e6c54. The description value in the frontmatter is now double-quoted with backslash-escaping for any embedded quotes: `description: "..."`. This means descriptions containing `: ` (e.g. `Remember: use tabs`) are valid YAML and parse correctly.

TimeToBuildBob:
## Greptile Convergence Adjudication — #3638

All Greptile threads resolved. Structured merge recommendation below.

### Fixed (across commits f9a08aded4f0 and 7bb89e6c54)

- **Slug-collision index duplicates**: `_update_memory_index` now keys on the slug (not raw name), so two names normalizing to the same slug update one entry rather than creating duplicates.
- **Concurrent index updates** (P1): `fcntl.flock(LOCK_EX)` serializes the full read-modify-write on MEMORY.md, preventing parallel-session data loss.
- **Confirmation gate** (P1): `execute_memory` routes through `execute_with_confirmation`; content is previewed before any write; interactive users can cancel.
- **`re.sub` regex injection** (Bob AI review P1): replacement is `lambda _: replacement`, so user-controlled descriptions with `\1`/`\g<name>` no longer raise `re.error`.
- **YAML frontmatter quoting** (Bob AI review P2): description value is double-quoted with backslash/quote escaping — descriptions containing `: ` parse correctly.
- **Instructions framing** (P2): rewritten around concrete agent decisions with examples.

### Remaining (non-blocking, dismissed)

- **Slug collision overwrites backing file** (P1, rounds 1 and 2): Greptile's own reply on the thread confirmed this is intentional — the slug IS the memory identity. Two names normalizing to the same slug are the same slot. Greptile: "Treating the slug as the memory identity makes the overwrite behavior intentional." DISMISSED: intentional design, accepted by Greptile.
- **`confirm_msg` not surfaced in dialog** (Greptile sub-concern): `execute_with_confirmation` accepts `confirm_msg` but the underlying `get_confirmation` call doesn't forward it — users see the content preview but not the "Save to path X" message. This is a shared-utility limitation affecting all tools using `execute_with_confirmation`, not introduced by this PR. The content review + cancel path is intact. DISMISSED: nitpick, shared-utility UX limitation.

### CI

All 17 checks green on latest commit.

### Domain risk

End-to-end cross-runtime path (save in gptme → load in fresh gptme session → load in CC) is not covered by the unit tests. Worth a manual smoke test before enabling in production contexts. The path computation (`get_cc_memory_dir`) is exercised in tests via mocking, but a real cross-runtime round-trip has not been verified in CI.

### Convergence

2 rounds, 0 stable rounds (status `new_blocking` in detector prior to this adjudication). After classification: all remaining findings are dismissed as intentional design or shared-utility nitpick. No genuinely blocking issues remain.

Merge-ready pending maintainer review of the domain-risk note above.

gptme/tools/memory.py
```diff
@@ -0,0 +1,227 @@
...
+    filename = f"{slug}.md"
+    file_path = memory_dir / filename
+
+    lines = content.strip().splitlines()
+    description = lines[0].strip() if lines else name
+    body = "\n".join(lines[1:]).strip() if len(lines) > 1 else content.strip()
+
+    safe_desc = description.replace("\\", "\\\\").replace('"', '\\"')
+    frontmatter = f'---\nname: {slug}\ndescription: "{safe_desc}"\nmetadata:\n  type: general\n---\n\n'
+    file_path.write_text(frontmatter + body + "\n", encoding="utf-8")
```
TimeToBuildBob, line 154: ❌ **P1** — The memory file is written via `file_path.write_text()` at line 154 without any file lock, while the index update in `_update_memory_index()` is guarded by an exclusive `flock`. If two gptme sessions or gptme and Claude Code save memories with the same slug concurrently, both processes open the same `.md` file with truncation and write. The writes are not atomic, so the final file can contain interleaved content or be truncated by the other's open, and the index entry (updated under the lock) may describe a different version than the file on disk. The next session that reads MEMORY.md will display a description that does not match the actual memory body, and the memory content may be corrupted. Because the lock is only on the index, the race is unsolved.

TimeToBuildBob: Fixed in 1b568a9. The memory file write is now inside the exclusive flock: `_update_memory_index` accepts optional `file_path`/`file_content` params and writes the `.md` file under the same lock before updating the index. Concurrent saves to the same slug are fully serialised.

PR #3640, opened 2026-08-26
fix(cli): reject unknown bare --model names before get_prompt()

TimeToBuildBob:
## Summary

`gptme --model nosuch --non-interactive "hi"` used to look like a hang: the early `--model` check only ran for slash-prefixed `provider/model` strings, so a bare typo skipped validation, paid `get_prompt()` (workspace `context_cmd`, often 10s+), and only then failed in `init_model()`.

`gptme-util llm generate -m nosuch` already rejected this immediately. The top-level CLI now does the same: resolve bare names that aren't a known/custom provider via `get_model()`, and raise `UsageError` when the fallback provider is `unknown`. Aliases such as `gpt-4o` and bare provider names such as `anthropic` still pass through to `init_model()`.

Dogfood session 7f83 (`dogfood-gptme-cli-surface-2026-08-26`).

## Reproduction (before)

```bash
timeout 8 gptme --non-interactive --no-stream --model nosuch -- "hi"
# exit 124, no error — still inside get_prompt()
```

## After

```bash
gptme --non-interactive --no-stream --model nosuch -- "hi"
# exit 2, ~1s
# Error: --model: Unknown model 'nosuch'. Use 'provider/model' with a known provider ...
```

## Test plan

- [x] `test_model_rejects_unknown_bare_name_before_context_cmd` — `get_prompt` is not called
- [x] `test_model_allows_bare_alias_through_validation_block` — `gpt-4o` still reaches `get_prompt`
- [x] Existing slash-path tests still pass (`unknown provider`, empty path components, nested `openrouter/...`)
- [x] Live: unknown bare name exits 2 in ~1s; unknown provider still fails fast

### DIFF
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1010,8 +1010,9 @@ def main(
     from ..chat import chat
     from ..config import ensure_workspace_dir, get_config, setup_config_from_cli
     from ..init import init_logging
-    from ..llm import get_provider_from_model
+    from ..llm import get_provider_from_model, is_custom_provider
     from ..llm import reply as llm_reply
+    from ..llm.models import PROVIDERS, get_model
     from ..message import Message
     from ..profiles import get_profile
     from ..prompts import (
@@ -1342,11 +1343,27 @@ def main(
         sys.exit(1)
 
     # Validate model early to fail fast before the expensive get_prompt() call.
-    # Only check models with a provider/ prefix; bare provider names (e.g. "anthropic")
-    # and model aliases (e.g. "gpt-4o") are left for init_model() to resolve.
-    if config.chat.model and "/" in config.chat.model:
+    # Slash-prefixed names are checked via provider lookup. Bare provider names
+    # (e.g. "anthropic") and resolvable aliases (e.g. "gpt-4o") still pass
+    # through to init_model(); unresolvable bare names used to skip this block
+    # and pay get_prompt() (workspace context_cmd, 10s+) before init_model()
+    # rejected them.
+    if config.chat.model:
         try:
-            get_provider_from_model(config.chat.model)
+            if "/" in config.chat.model:
+                get_provider_from_model(config.chat.model)
+            elif config.chat.model not in PROVIDERS and not is_custom_provider(
+                config.chat.model
+            ):
+                resolved = get_model(config.chat.model)
+                if resolved.provider == "unknown":
+                    raise ValueError(
+                        f"Unknown model {config.chat.model!r}. Use 'provider/model' "
+                        "with a known provider "
+                        f"(e.g. 'openai/{config.chat.model}'), or configure a "
+                        "custom provider. Run 'gptme-util models list' to see "
+                        "available models."
+                    )
         except ValueError as e:
             _cleanup_aborted_new_logdir(logdir, preexisting=logdir_preexisting)
             raise click.UsageError(f"--model: {e}") from e
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1343,18 +1343,24 @@ def main(
         sys.exit(1)
 
     # Validate model early to fail fast before the expensive get_prompt() call.
-    # Slash-prefixed names are checked via provider lookup. Bare provider names
-    # (e.g. "anthropic") and resolvable aliases (e.g. "gpt-4o") still pass
-    # through to init_model(); unresolvable bare names used to skip this block
-    # and pay get_prompt() (workspace context_cmd, 10s+) before init_model()
-    # rejected them. Existing conversations already skip get_prompt(), and their
-    # saved aliases may outlive the current registry, so preserve resume behavior.
-    if config.chat.model and not is_existing_conversation:
+    # Slash-prefixed names are always checked via provider lookup — that was the
+    # original behavior, including when resuming (`gptme --resume --model
+    # badprovider/x` must still be a UsageError, not a later generic crash).
+    # Bare provider names (e.g. "anthropic") and resolvable aliases (e.g.
+    # "gpt-4o") still pass through to init_model(); unresolvable bare names used
+    # to skip this block and pay get_prompt() (workspace context_cmd, 10s+)
+    # before init_model() rejected them. Saved aliases on existing conversations
+    # may outlive the current registry, so skip the new bare-name check unless
+    # --model was explicitly passed on the command line.
+    model_from_cli = ctx.get_parameter_source("model") == ParameterSource.COMMANDLINE
+    if config.chat.model:
         try:
             if "/" in config.chat.model:
                 get_provider_from_model(config.chat.model)
-            elif config.chat.model not in PROVIDERS and not is_custom_provider(
-                config.chat.model
+            elif (
+                (not is_existing_conversation or model_from_cli)
+                and config.chat.model not in PROVIDERS
+                and not is_custom_provider(config.chat.model)
             ):
                 resolved = get_model(config.chat.model)
                 if resolved.provider == "unknown":
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1347,8 +1347,9 @@ def main(
     # (e.g. "anthropic") and resolvable aliases (e.g. "gpt-4o") still pass
     # through to init_model(); unresolvable bare names used to skip this block
     # and pay get_prompt() (workspace context_cmd, 10s+) before init_model()
-    # rejected them.
-    if config.chat.model:
+    # rejected them. Existing conversations already skip get_prompt(), and their
+    # saved aliases may outlive the current registry, so preserve resume behavior.
+    if config.chat.model and not is_existing_conversation:
         try:
             if "/" in config.chat.model:
                 get_provider_from_model(config.chat.model)
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -586,9 +586,13 @@ def test_model_rejects_unknown_bare_name_before_context_cmd(
 def test_model_allows_bare_alias_through_validation_block(
     monkeypatch, tmp_path: Path, runner: CliRunner
 ):
-    """A resolvable bare alias (gpt-4o) should still reach get_prompt()."""
+    """A resolvable bare alias should still reach get_prompt()."""
     monkeypatch.setenv("HOME", str(tmp_path))
     monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path / "data"))
+    monkeypatch.setattr(
+        "gptme.llm.models.get_model",
+        lambda model: SimpleNamespace(provider="openai", model=model),
+    )
 
     called: dict[str, bool] = {"get_prompt": False}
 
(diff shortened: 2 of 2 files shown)


gptme/cli/main.py
```diff
@@ -1342,11 +1343,27 @@ def goodbye_handler():
...
 
     # Validate model early to fail fast before the expensive get_prompt() call.
-    # Only check models with a provider/ prefix; bare provider names (e.g. "anthropic")
-    # and model aliases (e.g. "gpt-4o") are left for init_model() to resolve.
-    if config.chat.model and "/" in config.chat.model:
+    # Slash-prefixed names are checked via provider lookup. Bare provider names
+    # (e.g. "anthropic") and resolvable aliases (e.g. "gpt-4o") still pass
+    # through to init_model(); unresolvable bare names used to skip this block
+    # and pay get_prompt() (workspace context_cmd, 10s+) before init_model()
+    # rejected them.
```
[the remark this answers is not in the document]
TimeToBuildBob: The primary motivation here is **correctness** — a bare typo like `--model nosuch` used to silently hang (exit 124 from a timeout) rather than fail with a clear error. Performance improvement is a side effect, not the primary goal.

Timing evidence is in the PR description:

- Before: `timeout 8 gptme --non-interactive --no-stream --model nosuch -- "hi"` → exit 124, no error (still inside `get_prompt()`)
- After: same command → exit 2, ~1s, with `Error: --model: Unknown model 'nosuch'...`

The regression tests assert this directly: `test_model_rejects_unknown_bare_name_before_context_cmd` uses a mock to verify `get_prompt()` is **never called** for unknown bare names, which is both the correctness gate and the performance gate in one. There's no ambiguity about what's being measured.

TimeToBuildBob:
Status update:

**Greptile P2 finding resolved** — the "missing profiling baseline" thread has been addressed (this is a correctness fix, not a performance optimization — `get_prompt()` must not run for an unknown model name; `test_model_rejects_unknown_bare_name_before_context_cmd` verifies the invariant directly). Thread marked resolved.

**CI re-run in progress** — the one failing job (`test_multi_tool_per_message` in `test_server_v2_auto_stepping.py`) is a timing-based async-event test unrelated to model name validation. Triggered a re-run of the failed job.

Waiting for human review/merge (`gptme/cli/main.py` is outside the self-merge category).

tests/test_cli.py
```diff
@@ -551,6 +551,66 @@ def test_model_rejects_unknown_provider_before_context_cmd(
...
+
+    monkeypatch.setattr("gptme.prompts.get_prompt", _fake_get_prompt)
+    monkeypatch.setattr(
+        importlib.import_module("gptme.chat"), "chat", lambda *args, **kwargs: None
+    )
+    monkeypatch.setattr("gptme.telemetry.init_telemetry", lambda **kwargs: None)
+
+    result = runner.invoke(
+        cli.main,
+        ["--model", "gpt-4o", "--non-interactive", "hello"],
```
TimeToBuildBob: ⚠️ **P2** — The test `test_model_allows_bare_alias_through_validation_block` relies on the real `get_model` to resolve 'gpt-4o' to a known provider. This makes the test dependent on the project's model alias list: if 'gpt-4o' is removed or renamed, the test fails even though the code under test is unchanged. The test should mock `get_model` to return a model with a known provider (e.g., provider='openai') so it only exercises the validation branch, not the model registry. Without mocking, the test is not hermetic and can break in environments where 'gpt-4o' is not a recognized alias.

TimeToBuildBob:
## 🤖 AI code review

Adds early validation for bare --model names in gptme/cli/main.py: when the model is not slash-prefixed, not a known provider, and not a custom provider, it resolves via get_model() and raises UsageError if the provider is 'unknown'. Adds three tests covering unknown bare names, resolvable aliases, and existing conversations with saved aliases.

**❌ P1 high** — `gptme/cli/main.py:1352`

The new guard `if config.chat.model and not is_existing_conversation:` skips all model validation whenever the conversation being opened already has a non-empty log, even when the user explicitly passes `--model` on the command line. Previously, slash-prefixed models were always validated early via `get_provider_from_model(config.chat.model)` regardless of whether the conversation existed. Now, resuming an existing conversation with an explicitly invalid slash-prefixed model (e.g. `gptme --resume --model badprovider/some-model`) bypasses the early check, so the CLI no longer exits with a clean `UsageError` (exit code 2) and the descriptive message. Instead, the invalid model flows through to `chat()`, where `init_model` raises a generic exception caught by the broad handler, producing exit code 1 and a less informative error. This defeats the PR's stated fast-fail goal for the exact case of resuming with an explicit model. The fix is to preserve the old behavior for slash-prefixed models even when `is_existing_conversation` is true, or to gate the skip on whether the model was explicitly provided via `ctx.get_parameter_source("model")`.

<sub>**How this was verified:** Confirmed the old code only guarded on `"/" in config.chat.model` and always validated slash-prefixed models. Traced the new flow: for an existing conversation, `is_existing_conversation` is True, so the block is fully skipped; the model is then used later in `chat()`. Checked the outer exception handling in main: ValueError is not caught there, so it would surface as a generic fatal error.</sub>

</details>

- `gptme/cli/main.py` — Extends the early model-validation block to also reject unknown bare model names by consulting get_model() when the name is not a known or custom provider; imports PROVIDERS and get_model.
- `tests/test_cli.py` — Adds three tests: unknown bare name fails fast, resolvable alias reaches get_prompt, and existing-conversation saved aliases skip prevalidation.

</details>

| `94f18d9d30b6` | 4/5 | 1 | llm | 2026-08-27 01:01 UTC |
| `0151114e967d` | 4/5 | 2 | llm | 2026-08-27 23:34 UTC |

</details>

<sub>Reviewed `d3ec55f1693f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 217s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_cli.py
TimeToBuildBob: Fixed in `0151114e9`: the alias-path test now stubs `gptme.llm.models.get_model` with a synthetic known provider and uses `known-alias`, so it exercises only the early-validation branch rather than depending on the registry's `gpt-4o` alias. The focused regression slice passes (2 tests).

TimeToBuildBob:
@greptileai review

tests/test_cli.py
TimeToBuildBob: Fixed in `0151114e9`. The alias-path test now stubs `gptme.llm.models.get_model` with a synthetic `openai` provider and uses `known-alias`, removing its dependency on the live model registry. Focused validation: 2 tests pass.

TimeToBuildBob:
Automated merge handoff for `0151114e967d`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

gptme/cli/main.py
```diff
@@ -1342,11 +1343,27 @@ def goodbye_handler():
...
+    if config.chat.model:
```
TimeToBuildBob: ⚠️ **P2** — The new validation runs unconditionally whenever config.chat.model is set, including when resuming an existing conversation. A conversation created with a bare alias that has since been removed from get_model()'s known-aliases list can no longer be resumed; the CLI now fails with a UsageError before the resume path proceeds. Previously, the early check only ran for slash-prefixed models, so such a resume would work. This is a regression for users resuming older conversations whose model name is no longer resolvable.

tests/test_cli.py
```diff
@@ -551,6 +551,70 @@ def test_model_rejects_unknown_provider_before_context_cmd(
...
+
+def test_model_allows_bare_alias_through_validation_block(
+    monkeypatch, tmp_path: Path, runner: CliRunner
+):
+    """A resolvable bare alias should still reach get_prompt()."""
+    monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path / "data"))
+    monkeypatch.setattr(
+        "gptme.llm.models.get_model",
+        lambda model: SimpleNamespace(provider="openai", model=model),
```
TimeToBuildBob, line 596: ⚠️ **P2** — The new test `test_model_allows_bare_alias_through_validation_block` stubs `get_model` but does not stub `is_custom_provider`. If the test environment defines a custom provider named "known-alias", the `elif` condition in main.py at line 1355 short-circuits on `not is_custom_provider(...)` being False, so the new validation block is skipped entirely. The test then still reaches `get_prompt` and passes, meaning it does not actually exercise the code path it claims to test. Similarly, `test_model_rejects_unknown_bare_name_before_context_cmd` would fail if a custom provider named "definitely-not-a-model-7f83" exists, because validation would be skipped and the stubbed `get_prompt` would raise pytest.fail. Both tests are non-hermetic and depend on the absence of specific custom provider names.

TimeToBuildBob:
Pushed `d3ec55f169` addressing both open AI-review findings:

- early bare-model validation now skips existing conversations, preserving resumes whose saved alias has left the current registry;
- both validation tests explicitly force `is_custom_provider()` false, so they exercise the intended branch hermetically;
- added a resume regression test proving neither model prevalidation nor `get_prompt()` runs for an existing chat.

Focused validation: 3 tests pass. The in-band reviewer timed out at its hard deadline and correctly fell back to pushing; CI and the automatic reviewer are now running on the new head.

TimeToBuildBob:
Automated merge handoff for `d3ec55f1693f`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
TimeToBuildBob: Fixed in `ef87d16f0`.

Slash-prefixed `--model` is always validated again, including on resume: `gptme --resume --model badprovider/x` is a UsageError (exit 2) before `chat()`/`init_model()`. The existing-conversation skip now applies only to the new bare-name check, and only when `--model` was not passed on the command line — saved retired aliases still resume.

Regression coverage:
- `test_model_rejects_explicit_unknown_provider_on_existing_conversation`
- `test_model_rejects_explicit_unknown_bare_name_on_existing_conversation`
- `test_model_validation_preserves_existing_conversation_alias` still passes

TimeToBuildBob:
Pushed `ef87d16f0` for the P1: slash-prefixed `--model` is validated again on resume, so `gptme --resume --model badprovider/x` is a UsageError (exit 2) instead of a later generic crash. The existing-conversation skip now only covers implicit saved bare aliases.

In-band reviewer could not run (OpenRouter daily key limit) and correctly fell back to push; GitHub AI review will pick up the new head.

TimeToBuildBob:
Automated merge handoff for `ef87d16f0700`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/main.py`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Merged by Erik as `f4d71ca6ab1e`. Last head CI was green and review findings were already addressed. Closing the monitoring arc.

PR #3645, opened 2026-08-27
feat(doctor): add proxy URL format check (LLM_PROXY_URL)

TimeToBuildBob:
## Summary

- Adds `_check_proxy()` to `gptme/cli/doctor.py` that validates `LLM_PROXY_URL` when configured
- Registers the new check in `run_diagnostics()` between `_check_config` and `_check_api_keys`
- 10 new tests covering all branches, including the malformed-URL case from #3526

## Motivation

`gptme-doctor` had zero proxy-related checks (`grep -c -i proxy gptme/cli/doctor.py` returns 0 before this PR). The proxy-detection logic in `gptme/llm/llm_anthropic.py` sets `_is_proxy = proxy_url is not None`, which silently changes model ID formatting (`api_model = f"anthropic/{model}" if (via_gptme or _is_proxy) else model`) even for malformed URLs.

illiand's failure in #3526 is exactly this case — a misconfigured `LLM_PROXY_URL` changed routing with no diagnostic surface.

## What the check does

| URL | Result |
|-----|--------|
| Not set | SKIPPED — no proxy configured |
| `https://proxy.example.com` | OK |
| `http://localhost:8080` | OK |
| `proxy.example.com` (no scheme) | ERROR — must be http/https |
| `https://` (no host) | ERROR — missing host |
| `https://proxy.example.com/v1/api` | WARNING — trailing path may conflict with SDK routing |

## Test plan

- [ ] `env -u VIRTUAL_ENV uv run python -m pytest tests/test_doctor.py::TestCheckProxy -v` — all 10 tests pass
- [ ] `env -u VIRTUAL_ENV uv run python -m pytest tests/test_doctor.py -q` — all 88 tests pass
- [ ] `LLM_PROXY_URL=not-a-url gptme-doctor` surfaces an ERROR check for the proxy

### DIFF
--- a/gptme/cli/doctor.py
+++ b/gptme/cli/doctor.py
@@ -427,14 +427,15 @@ def _check_proxy(verbose: bool = False) -> list[CheckResult]:
     if port is not None:
         display_host = f"{display_host}:{port}"
 
-    # Path should be empty or just "/" — the Anthropic SDK appends its own paths
+    # Path should be empty or just "/" — the Anthropic SDK appends its own paths.
+    # Do not include it in diagnostics because proxy paths may contain credentials.
     if parsed.path and parsed.path != "/":
         results.append(
             CheckResult(
                 name="Proxy: LLM_PROXY_URL",
                 status=CheckStatus.WARNING,
-                message=f"URL has a non-root path {parsed.path!r} — may conflict with SDK routing",
-                details=f"Proxy: {parsed.scheme}://{display_host}{parsed.path}"
+                message="URL has a non-root path — may conflict with SDK routing",
+                details=f"Proxy: {parsed.scheme}://{display_host}/[path redacted]"
                 if verbose
                 else None,
                 fix_hint="Consider removing the path component; the Anthropic SDK appends its own paths",
--- a/gptme/cli/doctor.py
+++ b/gptme/cli/doctor.py
@@ -378,7 +378,22 @@ def _check_proxy(verbose: bool = False) -> list[CheckResult]:
         ]
 
     results = []
-    parsed = urlparse(proxy_url)
+    try:
+        parsed = urlparse(proxy_url)
+        hostname = parsed.hostname
+        # Accessing port validates malformed/non-numeric/out-of-range values.
+        _ = parsed.port
+    except ValueError as exc:
+        results.append(
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.ERROR,
+                message=f"Malformed URL — {exc}",
+                details=f"URL: {proxy_url}" if verbose else None,
+                fix_hint="Set LLM_PROXY_URL to a valid http or https URL",
+            )
+        )
+        return results
 
     # Scheme must be http or https
     if parsed.scheme not in ("http", "https"):
@@ -394,12 +409,12 @@ def _check_proxy(verbose: bool = False) -> list[CheckResult]:
         return results
 
     # Host must be non-empty
-    if not parsed.netloc:
+    if not hostname:
         results.append(
             CheckResult(
                 name="Proxy: LLM_PROXY_URL",
                 status=CheckStatus.ERROR,
-                message="Missing host — URL has no netloc",
+                message="Missing host",
                 details=f"URL: {proxy_url}" if verbose else None,
                 fix_hint="Set LLM_PROXY_URL to a full URL, e.g. https://my-proxy.example.com",
             )
--- a/gptme/cli/doctor.py
+++ b/gptme/cli/doctor.py
@@ -358,6 +358,78 @@ def _check_config(verbose: bool = False) -> list[CheckResult]:
     return results
 
 
+def _check_proxy(verbose: bool = False) -> list[CheckResult]:
+    """Check proxy URL configuration."""
+    from urllib.parse import urlparse
+
+    config = get_config()
+    proxy_url = config.get_env("LLM_PROXY_URL", None)
+
+    if not proxy_url:
+        return [
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.SKIPPED,
+                message="Not configured",
+                details="Set LLM_PROXY_URL to route LLM requests through a proxy"
+                if verbose
+                else None,
+            )
+        ]
+
+    results = []
+    parsed = urlparse(proxy_url)
+
+    # Scheme must be http or https
+    if parsed.scheme not in ("http", "https"):
+        results.append(
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.ERROR,
+                message=f"Invalid scheme {parsed.scheme!r} — must be http or https",
+                details=f"URL: {proxy_url}" if verbose else None,
+                fix_hint="Set LLM_PROXY_URL to a URL starting with http:// or https://",
+            )
+        )
+        return results
+
+    # Host must be non-empty
+    if not parsed.netloc:
+        results.append(
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.ERROR,
+                message="Missing host — URL has no netloc",
+                details=f"URL: {proxy_url}" if verbose else None,
+                fix_hint="Set LLM_PROXY_URL to a full URL, e.g. https://my-proxy.example.com",
+            )
+        )
+        return results
+
+    # Path should be empty or just "/" — the Anthropic SDK appends its own paths
+    if parsed.path and parsed.path != "/":
+        results.append(
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.WARNING,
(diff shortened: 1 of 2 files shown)


gptme/cli/doctor.py
```diff
@@ -358,6 +358,78 @@ def _check_config(verbose: bool = False) -> list[CheckResult]:
...
+                status=CheckStatus.ERROR,
+                message=f"Invalid scheme {parsed.scheme!r} — must be http or https",
+                details=f"URL: {proxy_url}" if verbose else None,
+                fix_hint="Set LLM_PROXY_URL to a URL starting with http:// or https://",
+            )
+        )
+        return results
+
+    # Host must be non-empty
+    if not parsed.netloc:
```
TimeToBuildBob: ⚠️ **P2** — The host validity check at line 397 uses `if not parsed.netloc:` to reject URLs without a host. However, urlparse() sets netloc to a non-empty string even when only a port is given: for `http://:8080` it yields netloc=':8080' and hostname=''. Because ':8080' is truthy, the check passes, the path check sees no path, and the function returns OK. A user with LLM_PROXY_URL='http://:8080' therefore gets a green 'Configured' from gptme-doctor even though the URL has no host and will fail at request time. The check should test `parsed.hostname` instead of `parsed.netloc` to catch this malformed configuration.

TimeToBuildBob:
## 🤖 AI code review

Adds a new `_check_proxy()` function to `gptme/cli/doctor.py` that validates the `LLM_PROXY_URL` configuration value, checking scheme, host, port, and path, and registers it in `run_diagnostics()` between `_check_config` and `_check_api_keys`. Adds a `TestCheckProxy` class in `tests/test_doctor.py` with 10 tests covering configured/unset, valid/invalid schemes, missing hosts, malformed ports, path warnings, and redaction of credentials.

- `gptme/cli/doctor.py` — Adds `_check_proxy()` function and wires it into `run_diagnostics()`.
- `tests/test_doctor.py` — Adds `TestCheckProxy` class with tests for the new proxy URL check.

</details>

| `b704a5cbe841` | 4/5 | 1 | llm | 2026-08-27 09:47 UTC |
| `a8c62c8a57bf` | 4/5 | 2 | llm | 2026-08-27 10:37 UTC |
| `ea9162218fe1` | 3/5 | 2 | llm | 2026-08-27 12:22 UTC |

</details>

<sub>Reviewed `1d18c890653b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 136s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `a8c62c8a57bf`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/doctor.py`

tests/test_doctor.py
```diff
@@ -434,6 +436,111 @@ def test_checks_user_config(self):
...
+        """Verbose mode includes URL details."""
+        with patch("gptme.cli.doctor.get_config") as mock_cfg:
+            mock_cfg.return_value.get_env.return_value = "https://proxy.example.com"
+            results = _check_proxy(verbose=True)
+        assert results[0].details is not None
+        assert "proxy.example.com" in results[0].details
+
+    def test_included_in_run_diagnostics(self):
+        """_check_proxy results appear in run_diagnostics output."""
+        results, _ = run_diagnostics()
```
TimeToBuildBob: ⚠️ **P2** — The test `test_included_in_run_diagnostics` calls `run_diagnostics()` without mocking `get_config` or `_check_proxy`. The real `_check_proxy` reads `LLM_PROXY_URL` from the environment via `get_config().get_env()`. If a developer's environment has `LLM_PROXY_URL` set, the test's assertion still passes because `_check_proxy` always returns at least one result, but the test's outcome depends on the environment: a broken or missing proxy URL won't fail the test, and conversely a valid URL won't change the result. This makes the test non-hermetic and gives false confidence that the integration works, since it never asserts that `run_diagnostics` actually invokes `_check_proxy`—it only checks that some result with the name prefix exists, which would also be true if `_check_proxy` were dropped entirely. The test should patch `_check_proxy` to return a fixed list, or mock `get_config` to control the env value.

TimeToBuildBob: Fixed in b4e850aaf: the integration test now patches `_check_proxy` with a unique sentinel result, asserts that exact result reaches `run_diagnostics()`, and verifies `_check_proxy(False)` was called. This removes the environment dependency and makes dropping the check fail the test.

TimeToBuildBob:
@greptileai review

gptme/cli/doctor.py
TimeToBuildBob: Fixed in ea9162218: diagnostic output now reconstructs only the parsed hostname and port, so userinfo is omitted in normal and verbose output. Invalid-URL paths no longer echo the configured value either. Added regression coverage for valid credential-bearing URLs and malformed credential-bearing URLs; `96 passed` in `tests/test_doctor.py`.

TimeToBuildBob:
Automated merge handoff for `ea9162218fe1`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/doctor.py`

gptme/cli/doctor.py
```diff
@@ -358,6 +358,101 @@ def _check_config(verbose: bool = False) -> list[CheckResult]:
     return results
 
 
+def _check_proxy(verbose: bool = False) -> list[CheckResult]:
+    """Check proxy URL configuration."""
+    from urllib.parse import urlparse
+
+    config = get_config()
+    proxy_url = config.get_env("LLM_PROXY_URL", None)
```
TimeToBuildBob, line 366: ❌ **P1** — `_check_proxy()` reads the proxy URL only via `config.get_env("LLM_PROXY_URL", None)`. Unlike `_check_api_keys`, which falls back to `os.environ.get(env_var) or config.get_env(env_var)`, this function never consults the process environment. When a user sets `LLM_PROXY_URL` as a shell variable (the common method, and the exact situation behind #3526 where an env var changed routing), `config.get_env` may only read from a config-file environment section, so the check reports SKIPPED even though routing code in `gptme/llm/llm_anthropic.py` reads the env var directly and changes behavior. The diagnostic thus gives a false 'not configured' result, defeating the check's purpose. The fix is to fall back to `os.environ.get("LLM_PROXY_URL")` before (or in addition to) reading the config. _(1/2 passes — pass 1 looked and disagreed)_

gptme/cli/doctor.py
```diff
@@ -358,6 +358,101 @@ def _check_config(verbose: bool = False) -> list[CheckResult]:
...
+    if port is not None:
+        display_host = f"{display_host}:{port}"
+
+    # Path should be empty or just "/" — the Anthropic SDK appends its own paths
+    if parsed.path and parsed.path != "/":
+        results.append(
+            CheckResult(
+                name="Proxy: LLM_PROXY_URL",
+                status=CheckStatus.WARNING,
+                message=f"URL has a non-root path {parsed.path!r} — may conflict with SDK routing",
```
TimeToBuildBob: ⚠️ **P2** — The warning branch of _check_proxy embeds the raw URL path into the diagnostic message and verbose details without any redaction. For a proxy URL such as "https://proxy.example.com/secret-token", the path "/secret-token" is printed verbatim in the WARNING result (and again in the verbose details line). The PR's stated intent is to redact credentials, and it does strip the userinfo component via display_host, but it leaves the path intact. If a user configures a path-based API key or token (common for gateway proxies), that secret is exposed in `gptme-doctor` output, which may be shared in logs or bug reports. The fix is to omit or redact the path from the message and details, similar to how userinfo is handled. _(2/2 passes)_

TimeToBuildBob:
Addressed the latest review pass in 1d18c8906:

- Fixed the path-disclosure finding: non-root proxy paths are now omitted from the warning and shown as `[path redacted]` in verbose output, with a regression test using a token-like path.
- The process-environment finding is a false positive: `Config.get_env()` already checks both `GPTME_LLM_PROXY_URL` and `LLM_PROXY_URL` in `os.environ` before config-file values (`gptme/config/core.py:163-177`). Adding a separate `os.environ.get()` here would duplicate that lookup.

Verified locally: `18 passed` for `TestCheckProxy`; Ruff check and format pass. CI is running on the new head.

gptme/cli/doctor.py
TimeToBuildBob: False positive: Config.get_env checks both GPTME_LLM_PROXY_URL and LLM_PROXY_URL in os.environ before config values (gptme/config/core.py:163-177).

gptme/cli/doctor.py
TimeToBuildBob: Fixed in 1d18c8906: warning output omits the raw path and verbose output uses [path redacted], with regression coverage for a token-like proxy path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `1d18c890653b`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/doctor.py`

PR #3209, opened 2026-07-12
feat(gptme-util): add `chats fork` for point-in-time session forking

TimeToBuildBob:
## Summary

Adds `gptme-util chats fork <session> --at-turn N [--name NAME]` to fork an existing session at any conversation turn and re-run from that point — enabling agentic self-correction and human-guided retries without touching the original session.

> Per Erik's review: moved out of the main `gptme` CLI surface entirely. Uses "fork" terminology (not "branch") to avoid confusion with git branches or gptme's built-in LogManager branch system.

**Use cases:**
- Fork a session at turn 3 where a wrong tool was chosen, try a different approach
- Human says "start over from turn 7 but this time use X" — now automatable
- Reproducible ablation runs over the same conversation prefix

**Usage:**
```bash
# Fork a session at turn 5 (creates a new session directory)
gptme-util chats fork my-session --at-turn 5

# Fork with explicit name for the new session
gptme-util chats fork my-session --at-turn 12 --name retry-v2

# Turn 0: system context only (no user messages)
gptme-util chats fork my-session --at-turn 0 --name clean-slate
```

**Turn semantics:** 1-indexed user turns. Turn 0 = system/pre-user messages only; turn N = through Nth complete user+assistant exchange (including tool-call batches). Out-of-range turn → all messages returned unchanged.

The source session is never modified. A new session directory is created with the sliced `conversation.jsonl`.

## Implementation

- `_slice_at_turn(messages, turn)` helper in `gptme/cli/cmd_chats.py` — pure function, handles tool-call multi-message batches in the same turn correctly
- `gptme-util chats fork` subcommand added to the existing `chats` group
- `--at-turn` uses `click.IntRange(min=0)` so negative values are rejected at CLI boundary with a clean error message
- Atomic `mkdir(exist_ok=False)` collision guard — any collision (pre-existing session, stale state, concurrent invocation) raises `click.UsageError` cleanly

## Tests

Tests in `tests/test_cli_from_turn.py`:
- `_slice_at_turn` unit tests: turn 0 (system only), turn 0 with no system, first exchange, second exchange, equals total, exceeds total, last-turn-is-all, content preservation, multi-assistant same turn (tool-call batches)
- CLI integration: `chats fork` end-to-end branch creation, name collision guard, negative turn rejection

Inspired by arXiv:2605.21997 "The Log Is the Agent" — the JSONL session format makes this a natural and cheap operation.

Issue #1821, opened 2026-03-24
feat(ui): native live status/context bar for terminal sessions


TimeToBuildBob:
## Summary

Add a persistent status display to gptme's terminal UI showing real-time session state: current task, token budget usage, active context tier, and session type.

## Motivation

Claude Code implemented a similar feature (`statusLine`) that's proven valuable for transparency during long agent sessions. gptme users running autonomous agents face the same problem: it's hard to tell at a glance how deep into a context window you are, or what task the agent is currently working on.

Bob (the gptme-based autonomous agent at `ErikBjare/bob`) implemented a Claude Code statusline (`~/.claude/statusline.py`) that shows `[active-task] ctx:N%` with urgency markers. This should be native to gptme's TUI.

## Proposed UI

In gptme's terminal output, add a persistent line at the bottom (or top) showing:

```
[task-name] ctx:42% | tokens:84k/200k | sonnet-4.6 | autonomous
```

Urgency markers:
- `~ctx` at 60% context usage
- `!ctx` at 80% context usage
- `!!ctx` at 95% (critical)

## Implementation Notes

- For headless/pipe mode: skip the status bar (or write to stderr)
- For TUI mode: use `rich.live` or similar to update a persistent status line
- Context percentage: `total_tokens_used / context_window_size`
- Task name: from `$CASCADE_TASK` env var or gptme workspace task file
- Should respect `$NO_COLOR` and `$TERM`

## Related Work

- Claude Code `statusLine` feature: reference implementation for the UX pattern
- `~/.claude/statusline.py` in Bob's workspace: existing implementation for CC sessions
- gptme issue #1152 (browser tool): mentions display/transparency improvements

## Prior Art

- `claude-hud` (6K+ stars) — shows context usage in Claude Code
- CC `statusLine.type=command` — our current workaround for CC sessions

This would be a first-class native alternative that works in gptme without needing a separate plugin.

ErikBjare:
This feels like one of those "easy in TUI, hard in CLI" features like the long-standing (with several attempts) #569.

TimeToBuildBob:
Fair point — the input-while-running part of #569 is the hard bit (requires async input handling or TUI mode). This proposal is narrower: read-only status display only, no input handling.

The bet is that `rich.live` makes this tractable for interactive sessions — gptme already uses `rich` for output, so adding a live status bar at the bottom (updating in-place via ANSI) should be within reach. Headless/pipe mode just skips it entirely.

[...]

PR #3646, opened 2026-08-27
feat(cli): add --track-tokens flag to print running token count per LLM call

TimeToBuildBob:
## Summary

- Adds `GPTME_TRACK_TOKENS=1` env var that prints a running token-usage line after each LLM call
- Each line shows: context window usage (`N / limit (pct%)`), output tokens from the current call, and cumulative session total
- Session accumulator resets at the start of each `chat()` call
- Follows the same pattern as the existing env-only `GPTME_COSTS` mechanism
- `len_tokens` moved to module-level import in `chat.py` so tests can patch it cleanly

**2026-09-07 trim**: dropped the `--track-tokens` CLI flag per triage — Erik has objected to new default-CLI flags for this kind of opt-in toggle three times before (#3209, #3296, #1821) and never to an env-only one. `GPTME_TRACK_TOKENS` already drove the hook directly in `chat.py`'s `step()`; the flag was a second, redundant way to set the same env var.

## Motivation

Silent token overflow is a known pain point — gptme-discussions#138 describes a `/shell ls -R` call that silently failed at 289k tokens with no user-visible warning. This env var gives users a lightweight way to monitor context growth without committing to a full TUI redesign.

## Example output

```
[track-tokens] context: 12,450 / 200,000 (6.2%) | +328 out | session total: 25,106
```

## Test plan

- [x] `poetry run pytest tests/test_track_tokens.py -v` — 12 tests, all pass (accumulation across turns, tool-result content, percent output, reset, nested-chat isolation, error handling)
- [ ] Manual: `GPTME_TRACK_TOKENS=1 gptme 'hello'` prints the tracking line after each LLM call
- [ ] Default behaviour (env var unset) is byte-identical to today

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -80,9 +80,13 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
             session_tokens = [0]
             _session_tokens.set(session_tokens)
 
-        n_in = len_tokens(msgs, model)
-        n_out = len_tokens(msg_response, model)
-        context_limit = get_model(model).context
+        # Resolve once from the caller-supplied name (alias or full). Tokenizer
+        # and context metadata then share that ModelMeta; passing .full back
+        # into get_model() is the lookup 48225d823 avoided.
+        resolved = get_model(model)
+        n_in = len_tokens(msgs, resolved.full)
+        n_out = len_tokens(msg_response, resolved.full)
+        context_limit = resolved.context
         n_used = n_in + n_out
         session_tokens[0] += n_used
 
@@ -264,8 +268,11 @@ def chat(
         # Restore the caller's format so nested chat() calls (inline subagents)
         # don't clobber the parent's JSON mode when they exit.
         set_output_format(_prev_output_format)
-        # Restore the outer chat's token accumulator so nested chat() calls
-        # (inline subagents) don't corrupt the parent's running total.
+        # Restore the outer chat's token accumulator so same-context nested
+        # chat() calls don't corrupt the parent's running total.
+        # Thread-mode subagents start with a fresh contextvars context, so
+        # their totals are already isolated; folding inner into parent here
+        # would not see the parent list across threads.
         _session_tokens.set(_prev_tokens)
 
 
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -119,7 +119,10 @@ def chat(
 
     Callable from other modules.
     """
-    # Reset the per-session token accumulator for --track-tokens
+    # Reset the per-session token accumulator for --track-tokens.
+    # Save the previous value so nested chat() calls (inline subagents) restore
+    # the outer chat's running total on return instead of leaving [0] behind.
+    _prev_tokens = _session_tokens.get()
     _reset_token_accumulator()
 
     # Set initial terminal title with conversation name
@@ -248,6 +251,9 @@ def chat(
         # Restore the caller's format so nested chat() calls (inline subagents)
         # don't clobber the parent's JSON mode when they exit.
         set_output_format(_prev_output_format)
+        # Restore the outer chat's token accumulator so nested chat() calls
+        # (inline subagents) don't corrupt the parent's running total.
+        _session_tokens.set(_prev_tokens)
 
 
 def _run_chat_loop(
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -55,31 +55,26 @@ from .util.terminal import flush_stdin, set_current_conv_name, terminal_state_ti
 
 logger = logging.getLogger(__name__)
 
-# A mutable value in a ContextVar keeps each chat isolated while allowing all
-# step() calls in that chat to contribute to one running total.
-_session_tokens: ContextVar[list[int] | None] = ContextVar(
-    "session_tokens", default=None
-)
+# Store an immutable int. copy_context() (server/TUI/ACP step threads) copies
+# the binding, not the value: a list would be shared by reference and a child
+# step() would mutate the parent's running total. Rebinding via .set() keeps
+# each context isolated while still letting every step() in the same context
+# contribute to one total.
+_session_tokens: ContextVar[int | None] = ContextVar("session_tokens", default=None)
 
 
 def _reset_token_accumulator() -> None:
-    _session_tokens.set([0])
+    _session_tokens.set(0)
 
 
 def _get_session_tokens() -> int:
     """Return the running total for the current chat context."""
-    session_tokens = _session_tokens.get()
-    return session_tokens[0] if session_tokens is not None else 0
+    return _session_tokens.get() or 0
 
 
 def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
     """Print running token totals after each LLM call (enabled by GPTME_TRACK_TOKENS)."""
     try:
-        session_tokens = _session_tokens.get()
-        if session_tokens is None:
-            session_tokens = [0]
-            _session_tokens.set(session_tokens)
-
         # Resolve once from the caller-supplied name (alias or full). Tokenizer
         # and context metadata then share that ModelMeta; passing .full back
         # into get_model() is the lookup 48225d823 avoided.
@@ -88,7 +83,8 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
         n_out = len_tokens(msg_response, resolved.full)
         context_limit = resolved.context
         n_used = n_in + n_out
-        session_tokens[0] += n_used
+        session_tokens = (_session_tokens.get() or 0) + n_used
+        _session_tokens.set(session_tokens)
 
         # Occupancy is input+output: output tokens also consume the context
         # window, so an n_in-only percentage understates how close we are to
@@ -99,7 +95,7 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
         )
         print(
             f"[track-tokens] context: {n_used:,} / {context_display}{pct_display} | "
-            f"+{n_out:,} out | session total: {session_tokens[0]:,}",
+            f"+{n_out:,} out | session total: {session_tokens:,}",
             file=sys.stderr,
             flush=True,
         )
@@ -273,7 +269,7 @@ def chat(
         # chat() calls don't corrupt the parent's running total.
         # Thread-mode subagents start with a fresh contextvars context, so
         # their totals are already isolated; folding inner into parent here
-        # would not see the parent list across threads.
+        # would not see the parent total across threads.
         _session_tokens.set(_prev_tokens)
 
 
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -82,9 +82,9 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
 
         n_in = len_tokens(msgs, model)
         n_out = len_tokens(msg_response, model)
+        context_limit = get_model(model).context
         session_tokens[0] += n_in + n_out
 
-        context_limit = get_model(model).context
         context_display = f"{context_limit:,}" if context_limit else "unknown"
         pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
         print(
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -1,9 +1,11 @@
 import copy
 import logging
 import os
+import sys
 import threading
 from collections.abc import Callable, Generator
 from contextlib import ExitStack
+from contextvars import ContextVar
 from pathlib import Path
 
 from .commands import execute_cmd
@@ -53,29 +55,42 @@ from .util.terminal import flush_stdin, set_current_conv_name, terminal_state_ti
 
 logger = logging.getLogger(__name__)
 
-# Session-scoped token accumulator for --track-tokens.
-# Sums (input + output) tokens across all LLM calls in a session.
-_session_tokens: int = 0
+# A mutable value in a ContextVar keeps each chat isolated while allowing all
+# step() calls in that chat to contribute to one running total.
+_session_tokens: ContextVar[list[int] | None] = ContextVar(
+    "session_tokens", default=None
+)
 
 
 def _reset_token_accumulator() -> None:
-    global _session_tokens
-    _session_tokens = 0
+    _session_tokens.set([0])
+
+
+def _get_session_tokens() -> int:
+    """Return the running total for the current chat context."""
+    session_tokens = _session_tokens.get()
+    return session_tokens[0] if session_tokens is not None else 0
 
 
 def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
     """Print running token totals after each LLM call (enabled by --track-tokens)."""
-    global _session_tokens
+    session_tokens = _session_tokens.get()
+    if session_tokens is None:
+        session_tokens = [0]
+        _session_tokens.set(session_tokens)
+
     n_in = len_tokens(msgs, model)
     n_out = len_tokens(msg_response, model)
-    _session_tokens += n_in + n_out
+    session_tokens[0] += n_in + n_out
 
-    model_meta = get_model(model)
-    context_limit = model_meta.context
-    pct = 100.0 * n_in / context_limit if context_limit else 0.0
-    console.log(
-        f"[track-tokens] context: {n_in:,} / {context_limit:,} ({pct:.1f}%) | "
-        f"+{n_out:,} out | session total: {_session_tokens:,}"
+    context_limit = get_model(model).context
+    context_display = f"{context_limit:,}" if context_limit else "unknown"
+    pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
+    print(
+        f"[track-tokens] context: {n_in:,} / {context_display}{pct_display} | "
+        f"+{n_out:,} out | session total: {session_tokens[0]:,}",
+        file=sys.stderr,
+        flush=True,
     )
 
 
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -73,7 +73,7 @@ def _get_session_tokens() -> int:
 
 
 def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
-    """Print running token totals after each LLM call (enabled by --track-tokens)."""
+    """Print running token totals after each LLM call (enabled by GPTME_TRACK_TOKENS)."""
     try:
         session_tokens = _session_tokens.get()
         if session_tokens is None:
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -29,6 +29,7 @@ from .message import (
     get_output_format,
     is_output_json,
     is_output_quiet,
+    len_tokens,
     set_output_format,
 )
 from .prompt_queue import drain_prompt_queue
@@ -51,6 +52,31 @@ from .util.sound import print_bell
 from .util.terminal import flush_stdin, set_current_conv_name, terminal_state_title
 
 logger = logging.getLogger(__name__)
+
+# Session-scoped token accumulator for --track-tokens.
+# Sums (input + output) tokens across all LLM calls in a session.
+_session_tokens: int = 0
+
+
+def _reset_token_accumulator() -> None:
+    global _session_tokens
+    _session_tokens = 0
+
+
+def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
+    """Print running token totals after each LLM call (enabled by --track-tokens)."""
+    global _session_tokens
+    n_in = len_tokens(msgs, model)
+    n_out = len_tokens(msg_response, model)
+    _session_tokens += n_in + n_out
+
+    model_meta = get_model(model)
+    context_limit = model_meta.context
+    pct = 100.0 * n_in / context_limit if context_limit else 0.0
+    console.log(
+        f"[track-tokens] context: {n_in:,} / {context_limit:,} ({pct:.1f}%) | "
+        f"+{n_out:,} out | session total: {_session_tokens:,}"
+    )
 
 
 @trace_function(name="chat.main", attributes={"component": "chat"})
@@ -78,6 +104,9 @@ def chat(
 
     Callable from other modules.
     """
+    # Reset the per-session token accumulator for --track-tokens
+    _reset_token_accumulator()
+
     # Set initial terminal title with conversation name
     conv_name = logdir.name
     set_current_conv_name(conv_name)
@@ -671,6 +700,8 @@ def step(
             )
         if get_config().get_env_bool("GPTME_COSTS"):
             log_costs(msgs + [msg_response])
+        if get_config().get_env_bool("GPTME_TRACK_TOKENS"):
+            _log_token_usage(msgs, msg_response, get_model(model).full)
 
         # Trigger generation post hooks (e.g., TTS)
         if generation_post_msgs := trigger_hook(
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -74,24 +74,28 @@ def _get_session_tokens() -> int:
 
 def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
     """Print running token totals after each LLM call (enabled by --track-tokens)."""
-    session_tokens = _session_tokens.get()
-    if session_tokens is None:
-        session_tokens = [0]
-        _session_tokens.set(session_tokens)
+    try:
+        session_tokens = _session_tokens.get()
+        if session_tokens is None:
+            session_tokens = [0]
+            _session_tokens.set(session_tokens)
 
-    n_in = len_tokens(msgs, model)
-    n_out = len_tokens(msg_response, model)
-    session_tokens[0] += n_in + n_out
+        n_in = len_tokens(msgs, model)
+        n_out = len_tokens(msg_response, model)
+        session_tokens[0] += n_in + n_out
 
-    context_limit = get_model(model).context
-    context_display = f"{context_limit:,}" if context_limit else "unknown"
-    pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
-    print(
-        f"[track-tokens] context: {n_in:,} / {context_display}{pct_display} | "
-        f"+{n_out:,} out | session total: {session_tokens[0]:,}",
-        file=sys.stderr,
-        flush=True,
-    )
+        context_limit = get_model(model).context
+        context_display = f"{context_limit:,}" if context_limit else "unknown"
+        pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
+        print(
+            f"[track-tokens] context: {n_in:,} / {context_display}{pct_display} | "
+            f"+{n_out:,} out | session total: {session_tokens[0]:,}",
+            file=sys.stderr,
+            flush=True,
+        )
+    except Exception:
+        # Informational only — never crash the main chat loop over token display.
+        pass
 
 
 @trace_function(name="chat.main", attributes={"component": "chat"})
@@ -119,11 +123,11 @@ def chat(
 
     Callable from other modules.
     """
-    # Reset the per-session token accumulator for --track-tokens.
-    # Save the previous value so nested chat() calls (inline subagents) restore
-    # the outer chat's running total on return instead of leaving [0] behind.
+    # Save the previous token accumulator so nested chat() calls (inline subagents)
+    # restore the outer chat's running total on return.  The actual reset is
+    # deferred to inside the try block so the finally clause always runs to
+    # restore _prev_tokens even if pre-try setup code raises an exception.
     _prev_tokens = _session_tokens.get()
-    _reset_token_accumulator()
 
     # Set initial terminal title with conversation name
     conv_name = logdir.name
@@ -144,6 +148,9 @@ def chat(
     _prev_output_format = get_output_format()
     skill_lifecycle = ExitStack()
     try:
+        # Reset here (inside try) so the finally clause always restores
+        # _prev_tokens if any setup code above raises before reaching this point.
+        _reset_token_accumulator()
         set_output_format(output_format)
 
         # init
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -701,7 +701,7 @@ def step(
         if get_config().get_env_bool("GPTME_COSTS"):
             log_costs(msgs + [msg_response])
         if get_config().get_env_bool("GPTME_TRACK_TOKENS"):
-            _log_token_usage(msgs, msg_response, get_model(model).full)
+            _log_token_usage(msgs, msg_response, model)
 
         # Trigger generation post hooks (e.g., TTS)
         if generation_post_msgs := trigger_hook(
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -83,12 +83,18 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
         n_in = len_tokens(msgs, model)
         n_out = len_tokens(msg_response, model)
         context_limit = get_model(model).context
-        session_tokens[0] += n_in + n_out
+        n_used = n_in + n_out
+        session_tokens[0] += n_used
 
+        # Occupancy is input+output: output tokens also consume the context
+        # window, so an n_in-only percentage understates how close we are to
+        # the limit (and to overflow on the next turn).
         context_display = f"{context_limit:,}" if context_limit else "unknown"
-        pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
+        pct_display = (
+            f" ({100.0 * n_used / context_limit:.1f}%)" if context_limit else ""
+        )
         print(
-            f"[track-tokens] context: {n_in:,} / {context_display}{pct_display} | "
+            f"[track-tokens] context: {n_used:,} / {context_display}{pct_display} | "
             f"+{n_out:,} out | session total: {session_tokens[0]:,}",
             file=sys.stderr,
             flush=True,
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -103,9 +103,10 @@ def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> 
             file=sys.stderr,
             flush=True,
         )
-    except Exception:
+    except Exception as e:
         # Informational only — never crash the main chat loop over token display.
-        pass
+        # KeyboardInterrupt/SystemExit are BaseException and still propagate.
+        logger.warning("track-tokens failed: %s", e)
 
 
 @trace_function(name="chat.main", attributes={"component": "chat"})
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1140,7 +1140,7 @@ def main(
 
     # Everything below is an actual chat session: import the heavy parts of
     # gptme now, after the cheap early-exit paths (--help/--version/dispatch).
-    from ..chat import chat
+    from ..chat import _log_token_usage, chat
     from ..config import ensure_workspace_dir, get_config, setup_config_from_cli
     from ..init import init_logging
     from ..llm import get_provider_from_model, is_custom_provider
@@ -1643,6 +1643,12 @@ def main(
             tools=None,  # architect has no tools (planning only)
             workspace=workspace_path,
         )
+        # Architect calls llm_reply outside chat()/step(), so the per-step
+        # track-tokens hook never sees this turn. Log it here when enabled.
+        # chat() then resets the accumulator for the editor session — the
+        # planning context is a different window/model, not editor occupancy.
+        if get_config().get_env_bool("GPTME_TRACK_TOKENS"):
+            _log_token_usage(architect_msgs, architect_response, _arch_model)
 
         plan_text = architect_response.content.strip()
         logger.info("Architect plan generated (%d chars)", len(plan_text))
--- a/gptme/cli/main.py
+++ b/gptme/cli/main.py
@@ -1001,8 +1001,10 @@ def main(
     if injection_hygiene is not None:
         os.environ["GPTME_INJECTION_HYGIENE"] = injection_hygiene
 
-    # Propagate --track-tokens to the env var checked per-step.
-    if track_tokens:
+    # Propagate an explicit --track-tokens flag to the env var checked per-step.
+    # Preserve the original env string otherwise so get_env_bool() can interpret
+    # falsy values such as GPTME_TRACK_TOKENS=0 correctly.
+    if ctx.get_parameter_source("track_tokens") == ParameterSource.COMMANDLINE:
         os.environ["GPTME_TRACK_TOKENS"] = "1"
 
     # Convert tool_allowlist from tuple to string or None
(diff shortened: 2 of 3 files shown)


gptme/chat.py
```diff
@@ -50,6 +51,31 @@
...
+    global _session_tokens
+    n_in = len_tokens(msgs, model)
+    n_out = len_tokens(msg_response, model)
+    _session_tokens += n_in + n_out
+
+    model_meta = get_model(model)
+    context_limit = model_meta.context
+    pct = 100.0 * n_in / context_limit if context_limit else 0.0
+    console.log(
+        f"[track-tokens] context: {n_in:,} / {context_limit:,} ({pct:.1f}%) | "
```
TimeToBuildBob: ❌ **P1** — In `_log_token_usage`, `context_limit = model_meta.context` and then the f-string formats it with `{context_limit:,}`. If a model's metadata has no `context` attribute (or it is None), this raises `TypeError: unsupported format string passed to NoneType.__format__` on every LLM call once the flag is enabled. The ternary at line 73 only gates the percentage computation, not the formatting. As a result, enabling --track-tokens on any model without a defined context size (e.g., a locally registered custom model) crashes the chat loop. Should guard with a fallback such as `context_limit or 0` or only format when truthy.

gptme/chat.py
```diff
@@ -76,6 +102,9 @@ def chat(
 
     Callable from other modules.
     """
+    # Reset the per-session token accumulator for --track-tokens
+    _reset_token_accumulator()
```
TimeToBuildBob: ⚠️ **P2** — The module-level `_session_tokens` accumulator is reset at the start of every `chat()` call (line 106). `chat()` is re-entrant: inline subagents call it again with their own logdir (see the comment about restoring the caller's output format). When a subagent's `chat()` returns, the parent's accumulator has been zeroed and incremented only by the subagent's LLM calls, so the parent's 'session total' omits all calls made before the subagent ran. With --track-tokens enabled, the displayed session total becomes wrong (under-counted) after any tool that spawns a nested chat. The reset should only occur for top-level sessions, or the accumulator should be per-session (e.g., keyed by logdir) rather than a single global.

TimeToBuildBob:
## 🤖 AI code review

The PR adds a GPTME_TRACK_TOKENS environment variable that prints a running token-usage line after each LLM call, along with tests for accumulation, reset, and error handling. However, the provided diff is incomplete or unreadable, so no code-level review could be performed.

Confidence 5/5

- `gptme/chat.py` — Changes to the main CLI entrypoint.
- `tests/test_track_tokens.py` — New tests for the token tracking feature.

<sub>Also changed, not described above: `gptme/cli/main.py`</sub>

</details>

| `ff0e594e3995` | 4/5 | 1 | llm | 2026-08-28 09:28 UTC |
| `0a8aa9bcbeef` | 4/5 | 3 | llm | 2026-08-29 02:59 UTC |
| `25ceac05af4b` | 4/5 | 3 | llm | 2026-08-29 17:34 UTC |
| `4ec6b5aae552` | 4/5 | 2 | llm | 2026-08-29 19:24 UTC |
| `a7c4ad686f34` | 4/5 | 1 | llm | 2026-08-30 09:46 UTC |
| `0489af02a00e` | 5/5 | 0 | llm | 2026-08-30 21:01 UTC |
| `0489af02a00e` | 5/5 | 0 | llm | 2026-08-30 21:39 UTC |
| `7f366a56c99a` | 5/5 | 0 | llm | 2026-09-07 15:52 UTC |
| `7f366a56c99a` | 3/5 | 1 | llm | 2026-09-07 17:20 UTC |
| `8e46bef624e1` | 5/5 | 0 | llm | 2026-09-08 02:41 UTC |

</details>

<sub>Reviewed `8e46bef624e1` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 305s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -724,6 +724,13 @@ def _known_hint_names() -> list[str]:
     help="Write a JSON record before and after each tool call to this directory. "
     "Records can be committed alongside session artifacts for tool-call-level attribution.",
 )
+@click.option(
+    "--track-tokens",
+    "track_tokens",
+    is_flag=True,
+    envvar="GPTME_TRACK_TOKENS",
```
TimeToBuildBob, line 813: ⚠️ **P2** — The --track-tokens option is defined with is_flag=True and envvar='GPTME_TRACK_TOKENS'. Click treats any non-empty env var string as truthy, so setting GPTME_TRACK_TOKENS=0 or GPTME_TRACK_TOKENS=false still enables tracking. There is no --no-track-tokens counterpart to override an inherited env var, so the environment variable cannot be used to disable the feature once set by a parent process. This makes the toggle one-way and can cause unexpected token-usage output in an environment that sets the variable to a falsy-looking value.

TimeToBuildBob:
Fixed P1 in 7707cb003: save `_prev_tokens = _session_tokens.get()` before the reset and restore in the existing `finally` block (same pattern already used for `_prev_output_format`). Added `test_accumulator_restored_after_nested_chat_same_context` which exercises the same-context nested path that the `copy_context` test didn't cover.

TimeToBuildBob:
@greptileai review

gptme/chat.py
TimeToBuildBob: Fixed in 7707cb003: `chat()` now saves the outer accumulator before resetting for the nested call and restores it in `finally`. The new same-context nested-chat regression test verifies the outer total remains intact.

gptme/chat.py
```diff
@@ -627,6 +677,8 @@ def step(
             )
             if get_config().get_env_bool("GPTME_COSTS"):
                 log_costs(msgs + [msg_response])
+            if get_config().get_env_bool("GPTME_TRACK_TOKENS"):
+                _log_token_usage(msgs, msg_response, model)
```
TimeToBuildBob: ❌ **P1** — _log_token_usage (invoked from step() at line 681 when GPTME_TRACK_TOKENS is set) has no exception guard around len_tokens(msgs, model) and get_model(model).context. If the model name is unknown, the tokenizer is unavailable, or get_model raises, the exception propagates out of step() and aborts the entire turn, crashing the chat session. Tracking is purely informational; a failure here should be logged and ignored, not terminate the conversation. Wrap the body in try/except Exception and log a warning so the chat continues without the token-usage display. _(1/2 passes — pass 2 looked and disagreed)_

gptme/chat.py
```diff
@@ -76,6 +117,12 @@ def chat(
 
     Callable from other modules.
     """
+    # Reset the per-session token accumulator for --track-tokens.
+    # Save the previous value so nested chat() calls (inline subagents) restore
+    # the outer chat's running total on return instead of leaving [0] behind.
+    _prev_tokens = _session_tokens.get()
+    _reset_token_accumulator()
```
TimeToBuildBob: ❌ **P1** — The _reset_token_accumulator() call in chat() (line 124) executes outside the try block whose finally restores the previous accumulator value. If any exception occurs between that reset and the start of the try block (lines 127-142: set_current_conv_name, set_conversation_context, the tool_format assert, or get_output_format), the finally at line 248 never runs, so _session_tokens is left as [0]. For a nested chat() (e.g., an inline subagent) this corrupts the outer chat's running total: the outer accumulator that was saved in _prev_tokens is never restored, and subsequent token counts in the outer chat are added to a fresh zeroed accumulator, under-reporting the session total. Example: outer chat accumulates 100 tokens, then a subagent chat resets to [0] and fails before entering its try block; the outer chat now counts from 0 instead of resuming from 100, so the printed session total is missing the first 100 tokens. _(2/2 passes)_

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Fixed both P1s in eaadb944a7:

1. **Accumulator reset inside try block** — moved `_reset_token_accumulator()` from before the try to the first statement inside it. Pre-try setup code (`set_current_conv_name`, `set_conversation_context`, the `tool_format` assert) can now raise without leaving `_session_tokens` as `[0]`; the `finally` always restores `_prev_tokens` correctly.

2. **`_log_token_usage` error handling** — wrapped the entire body in `try/except Exception: pass`. `len_tokens()` and `get_model()` can raise for unknown/unsupported models; the display is informational and must never crash the main chat loop.

Added two targeted regression tests: `test_log_token_usage_survives_len_tokens_error` and `test_log_token_usage_survives_get_model_error`. All 9 tests pass.

TimeToBuildBob:
@greptileai review

gptme/chat.py
TimeToBuildBob: Fixed in eaadb944a7: `_log_token_usage()` is now fail-closed around all informational accounting and model-registry lookups, so token-display failures cannot abort the chat loop. Regression tests cover failures from both `len_tokens()` and `get_model()`.

gptme/chat.py
TimeToBuildBob: Fixed in eaadb944a7: the accumulator reset now occurs as the first statement inside the protected `try`, guaranteeing the existing `finally` restores the outer accumulator after nested-chat setup failures.

TimeToBuildBob:
Automated merge handoff for `eaadb944a790`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

tests/test_track_tokens.py
```diff
@@ -0,0 +1,235 @@
...
+def test_log_token_usage_survives_get_model_error():
+    """_log_token_usage survives get_model() raising (e.g. unknown model registry)."""
+    with (
+        patch.object(chat_module, "len_tokens", side_effect=[10, 5]),
+        patch.object(chat_module, "get_model", side_effect=KeyError("unknown model")),
+    ):
+        # Must not raise.
+        _log_token_usage(
+            [Message("user", "hi")], Message("assistant", "ok"), "unknown/model"
+        )
```
TimeToBuildBob, line 261: ❌ **P1** — The test `test_log_token_usage_survives_get_model_error` expects `_get_session_tokens() == 0` after `_log_token_usage` raises in `get_model`, but the function increments the accumulator before calling `get_model`. `_log_token_usage` does `n_in = len_tokens(...)`, `n_out = len_tokens(...)`, then `session_tokens[0] += n_in + n_out`, and only then calls `get_model(model).context`. So when `get_model` raises, the accumulator is already incremented to 15 (from the mocked len_tokens returning 10 and 5). The test will fail on its assertion. This is a defect in the test, not production code; the production function correctly survives the error, but the test's expectation of 0 is wrong and the accompanying comment 'Accumulator stays at 0' is false.

gptme/cli/main.py
```diff
@@ -874,6 +882,10 @@ def main(
     if injection_hygiene is not None:
         os.environ["GPTME_INJECTION_HYGIENE"] = injection_hygiene
 
+    # Propagate --track-tokens to the env var checked per-step.
+    if track_tokens:
+        os.environ["GPTME_TRACK_TOKENS"] = "1"
```
TimeToBuildBob, line 981: ⚠️ **P2** — The --track-tokens option is declared with envvar="GPTME_TRACK_TOKENS" and is_flag=True. Click treats any non-empty value of that env var as enabling the flag, including "0" or "false". In main(), when track_tokens is truthy (which it is for a "0" value), the code unconditionally sets os.environ["GPTME_TRACK_TOKENS"] = "1". This overwrites a user's explicit GPTME_TRACK_TOKENS=0 meant to disable the feature, so the flag gets enabled and the env var is forced to "1", ensuring the get_env_bool check in step() also sees it as true. As a result, any environment that sets GPTME_TRACK_TOKENS to a falsy value like "0" to suppress the tracking output will have it unexpectedly enabled, printing spurious lines to stderr and corrupting the intended quiet behavior.

gptme/chat.py
```diff
@@ -627,6 +684,8 @@ def step(
             )
             if get_config().get_env_bool("GPTME_COSTS"):
                 log_costs(msgs + [msg_response])
+            if get_config().get_env_bool("GPTME_TRACK_TOKENS"):
+                _log_token_usage(msgs, msg_response, model)
```
TimeToBuildBob: ⚠️ **P2** — `_reset_token_accumulator()` is only invoked at the start of `chat()` (line 150). However, `_log_token_usage()` is called from `step()`, which is a public function that can be invoked outside a `chat()` context (e.g., by a server loop). When `step()` is called directly, the accumulator is initialized lazily to `[0]` on first use and then accumulates across all subsequent calls in that context, never resetting. This causes the reported 'session total' to include tokens from multiple independent turns or requests, making the number misleading. For example, a server that processes two consecutive requests in the same thread/context would report a single growing total rather than per-request values. The fix is to reset the accumulator at the start of `step()` (or to have callers manage the context explicitly).

tests/test_track_tokens.py
```diff
@@ -0,0 +1,235 @@
...
+        _log_token_usage(
+            [Message("user", "outer")],
+            Message("assistant", "reply"),
+            "mock/gpt-mock",
+        )
+        assert _get_session_tokens() == 120
+
+        # Simulate what the fixed chat() does at nested-call entry:
+        # save outer state, reset for inner chat.
+        prev_tokens = chat_module._session_tokens.get()
```
TimeToBuildBob: ⚠️ **P2** — The test `test_accumulator_restored_after_nested_chat_same_context` claims to verify that a nested `chat()` call restores the parent's token accumulator, but it never actually invokes `chat()`. Instead, it manually performs the save/reset/restore steps (`prev_tokens = chat_module._session_tokens.get()`, `_reset_token_accumulator()`, and `chat_module._session_tokens.set(prev_tokens)`) that the production code is supposed to do. This means the test passes even if the `finally` block in `chat()` forgets to restore `_session_tokens`, because the test itself performs the restoration. Consequently, a regression that breaks the save/restore logic would go undetected. To actually exercise the behavior, the test should call `chat()` with a subagent that triggers a nested call, or at least patch `chat()` to verify its finally block runs the restore.

TimeToBuildBob:
Addressed the open AI-review findings in `ff0e594e39`:

- token totals now update only after model metadata lookup succeeds, so an informational failure cannot leave a partial count;
- `GPTME_TRACK_TOKENS=0` / `false` is preserved instead of being rewritten to `1` by Click's env-backed flag;
- the nested-chat regression test now calls production `chat()` and exercises its `finally` restoration rather than reproducing the implementation in the test.

I did not reset the accumulator inside every public `step()` call: a chat session intentionally invokes `step()` repeatedly and the feature is a running session total; resetting there would break the advertised behavior. Direct callers that want a session boundary should use `chat()` (the owner of that lifecycle).

Verified locally: 9 token-tracking tests, Ruff, mypy, and pre-commit pass. The in-band reviewer could not run because this linked worktree's inherited remotes have no `origin/master` merge base; the required wrapper recorded the skip and pushed. Fresh CI and the automatic GitHub-side AI review are running on this head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `ff0e594e3995`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

gptme/chat.py
```diff
@@ -50,6 +53,48 @@
...
+            session_tokens = [0]
+            _session_tokens.set(session_tokens)
+
+        n_in = len_tokens(msgs, model)
+        n_out = len_tokens(msg_response, model)
+        context_limit = get_model(model).context
+        session_tokens[0] += n_in + n_out
+
+        context_display = f"{context_limit:,}" if context_limit else "unknown"
+        pct_display = f" ({100.0 * n_in / context_limit:.1f}%)" if context_limit else ""
```
TimeToBuildBob: ⚠️ **P2** — The percentage shown in the tracking line is computed from input tokens only (n_in), not the total context usage including output tokens (n_out). When a model's context window is approached, a user can be misled into believing there is more headroom than there actually is, because the output tokens also consume context. For example, with a 200k context, if the input is 190k and the output is 15k, the display would show 95% (190k/200k) even though the combined 205k exceeds the limit. This directly undermines the feature's purpose of warning about imminent context overflow, as described in the PR motivation. The percentage should be based on n_in+n_out, or the output token count should be included separately.

TimeToBuildBob: Fixed in 0a8aa9bcb: occupancy and the percentage now use n_in+n_out, so the context line is window fill after the call. `+N out` remains the per-turn completion size. Added a regression that fails on the old n_in-only 90% figure when occupancy is 110%.

TimeToBuildBob:
Fixed the remaining occupancy P2 in `0a8aa9bcb`.

`--track-tokens` now reports context-window fill as input+output after each call (percentage included). `+N out` is still the per-turn completion size. Regression covers the 9k-in / 2k-out on a 10k window case that used to display 90%.

CI and automatic AI re-review will run on this push. Self-merge remains ineligible (`gptme/chat.py`, `gptme/cli/main.py`) — maintainer merge when green.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `0a8aa9bcbeef`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

tests/test_track_tokens.py
```diff
@@ -0,0 +1,255 @@
...
+        patch.object(chat_module, "get_model", return_value=meta),
+        patch.object(chat_module, "len_tokens", side_effect=[50, 10]),
+        patch.object(chat_module, "_run_chat_loop") as run_chat_loop,
+    ):
+        run_chat_loop.side_effect = lambda *args, **kwargs: _log_token_usage(
+            [Message("user", "inner")],
+            Message("assistant", "reply"),
+            "mock/gpt-mock",
+        )
+        chat_module.chat(
```
TimeToBuildBob, line 190: ⚠️ **P2** — The test `test_accumulator_restored_after_chat_returns` calls `chat_module.chat(...)` with `workspace=tmp_path`. Inside `chat()` (chat.py:202), `os.chdir(workspace)` changes the process's current working directory to `tmp_path` and never restores it. Because the test does not patch `os.chdir` or save/restore the cwd, the working directory remains `tmp_path` after the test, contaminating later tests that assume the original cwd for relative paths, config loading, or tool execution. This can cause spurious failures or cross-test side effects (e.g., files created in the wrong directory). The test should isolate the cwd change, e.g., by mocking `os.chdir` or using a fixture to restore the original cwd in teardown.

gptme/chat.py
```diff
@@ -196,6 +256,9 @@ def chat(
         # Restore the caller's format so nested chat() calls (inline subagents)
         # don't clobber the parent's JSON mode when they exit.
         set_output_format(_prev_output_format)
+        # Restore the outer chat's token accumulator so nested chat() calls
+        # (inline subagents) don't corrupt the parent's running total.
+        _session_tokens.set(_prev_tokens)
```
TimeToBuildBob, line 270: ⚠️ **P2** — In `chat()`'s `finally` block, `_session_tokens.set(_prev_tokens)` restores the token accumulator to its pre-call value, discarding any tokens recorded during nested `chat()` invocations (e.g., inline subagents). Consequently, the `session total` reported by `--track-tokens` omits all subagent LLM calls. For a session that spawns subagents, the displayed total undercounts actual token consumption, undermining the feature's purpose of monitoring cumulative usage. The current behavior is enforced by `test_accumulator_restored_after_chat_returns`, which asserts the parent total is unchanged after an inner chat. To correctly track session-wide usage, the inner chat's final count should be added to the parent's accumulator before restoring.

gptme/chat.py
```diff
@@ -50,6 +53,54 @@
...
+
+def _log_token_usage(msgs: list[Message], msg_response: Message, model: str) -> None:
+    """Print running token totals after each LLM call (enabled by --track-tokens)."""
+    try:
+        session_tokens = _session_tokens.get()
+        if session_tokens is None:
+            session_tokens = [0]
+            _session_tokens.set(session_tokens)
+
+        n_in = len_tokens(msgs, model)
```
TimeToBuildBob: ⚠️ **P2** — In _log_token_usage, the raw `model` string is passed directly to `len_tokens`, but the actual model used for the LLM call in step() is `get_model(model).full`. A user can invoke the CLI with a bare model alias (e.g., `--model gpt-4o`), in which case `len_tokens` may not recognize the alias and raise (or mis-count). Because _log_token_usage swallows all exceptions, the tracking line silently disappears — the feature does nothing for common aliases, defeating the purpose. Fix by using the resolved model name when counting tokens.

tests/test_track_tokens.py
TimeToBuildBob: Fixed in 25ceac05a: `test_accumulator_restored_after_chat_returns` now patches `chat_module.os.chdir`, so `chat()`'s workspace `chdir` cannot leak cwd into later tests.

gptme/chat.py
TimeToBuildBob: Fixed in 25ceac05a: `_log_token_usage` now calls `get_model(model)` once and passes `resolved.full` to `len_tokens` (and uses the same ModelMeta for the context limit). The alias is still what we look up — we do not pass `.full` back into `get_model()`. Regression: `test_log_token_usage_counts_with_resolved_model_name`.

gptme/chat.py
TimeToBuildBob: Rejected: restoring the outer accumulator is the correct contract, and folding the inner total into the parent would not include production subagent usage.

Thread-mode subagents (`gptme/tools/subagent/execution.py`) run in `threading.Thread`. On CPython's default GIL builds a new thread starts with an empty contextvars context, so the child's `_session_tokens` is independent of the parent. `chat()`'s `finally` cannot see the parent's list across threads, so adding the inner count before restore does not produce a session-wide total.

Same-context nested `chat()` is the case this restore protects (tests, TUI `copy_context`, future inline callers). Subagent tokens belong to a separate conversation and context window; mixing them into the parent's `session total` would conflate parent-window occupancy with a process-wide billing counter. That would be a different feature, not a one-line `finally` change.

Keeping restore-without-fold. `test_accumulator_restored_after_chat_returns` documents the contract.

TimeToBuildBob:
Pushed `25ceac05a` after in-band review (5/5, 0 findings).

- **cwd leak** (`420fb27a0ef8`): `test_accumulator_restored_after_chat_returns` now mocks `os.chdir`.
- **model alias** (`7a998c6f1c85`): `_log_token_usage` resolves once via `get_model(model)` and counts with `resolved.full`.
- **nested session total** (`257d1720be54`): rejected. Thread-mode subagents start with a fresh contextvars context, so folding inner into parent in `chat()`'s `finally` would not include production subagent usage. Same-context restore stays.

Our AI-review threads on this head are resolved. CI is pending. Not merging (`gptme/chat.py` / `gptme/cli/main.py` are outside the self-merge allowlist).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `25ceac05af4b`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

gptme/chat.py
```diff
@@ -50,6 +53,58 @@
...
+
+        # Resolve once from the caller-supplied name (alias or full). Tokenizer
+        # and context metadata then share that ModelMeta; passing .full back
+        # into get_model() is the lookup 48225d823 avoided.
+        resolved = get_model(model)
+        n_in = len_tokens(msgs, resolved.full)
+        n_out = len_tokens(msg_response, resolved.full)
+        context_limit = resolved.context
+        n_used = n_in + n_out
+        session_tokens[0] += n_used
```
TimeToBuildBob, line 90: ⚠️ **P2** — The 'session total' reported by _log_token_usage is the cumulative sum of n_in + n_out for every LLM call, but each call's n_in includes all prior messages (including previous outputs). As a result, tokens are counted multiple times: an output from turn 1 becomes part of the input for turn 2, so that same token content is added to the total again. For a conversation with T turns, the displayed total grows roughly as the sum of all history lengths, not as the size of the current context. Users monitoring context usage via --track-tokens will see the number inflate far beyond the model's actual context window, which is exactly the kind of misleading signal the feature aims to prevent. The metric would be more meaningful if it reported the current input size (n_in) or tracked unique tokens, but as written the 'session total' is a raw API-usage counter, not a context-occupancy indicator. Suggest renaming the label or changing the calculation to avoid confusion.

gptme/chat.py
```diff
@@ -50,6 +53,58 @@
...
+        pct_display = (
+            f" ({100.0 * n_used / context_limit:.1f}%)" if context_limit else ""
+        )
+        print(
+            f"[track-tokens] context: {n_used:,} / {context_display}{pct_display} | "
+            f"+{n_out:,} out | session total: {session_tokens[0]:,}",
+            file=sys.stderr,
+            flush=True,
+        )
+    except Exception:
```
TimeToBuildBob: ⚠️ **P2** — The bare `except Exception` in _log_token_usage swallows all non-BaseException failures, including MemoryError and other serious conditions. While the intent is to keep the informational display from crashing the chat loop, catching MemoryError here hides genuine resource exhaustion and makes diagnosing out-of-memory situations difficult. If MemoryError occurs during tokenization, the user sees no warning and the session total is silently not updated, which could mask a serious problem. The handler should at least log the exception (e.g., logger.warning) or re-raise non-Exception BaseExceptions (KeyboardInterrupt, SystemExit) as they are currently passed through. This is a low-severity issue but affects debuggability.

gptme/cli/main.py
```diff
@@ -874,6 +882,12 @@ def main(
     if injection_hygiene is not None:
         os.environ["GPTME_INJECTION_HYGIENE"] = injection_hygiene
 
+    # Propagate an explicit --track-tokens flag to the env var checked per-step.
+    # Preserve the original env string otherwise so get_env_bool() can interpret
+    # falsy values such as GPTME_TRACK_TOKENS=0 correctly.
+    if ctx.get_parameter_source("track_tokens") == ParameterSource.COMMANDLINE:
```
TimeToBuildBob, line 980: ⚠️ **P2** — The CLI propagation of --track-tokens only sets the GPTME_TRACK_TOKENS env var when the flag is explicitly provided on the command line (ParameterSource.COMMANDLINE). If the option is ever set through a config file or other Click source (e.g., default_map), the env var is left untouched, and the step() check `get_env_bool("GPTME_TRACK_TOKENS")` will read the pre-existing value (typically unset, falling back to false). This means the flag would be silently ignored in those configurations. The current PR does not add any config-file support for this option, so this is a latent inconsistency rather than an active bug, but it is worth tightening the detection to cover all sources that yield a non-default value, or to set the env var regardless of source when track_tokens is truthy.

TimeToBuildBob (commented):

**3 findings** · ⚠️ **3** P2

gptme/chat.py
TimeToBuildBob: Rejected: `session total` is an API-usage counter, not occupancy. Occupancy is already the first field.

The printed line has three distinct metrics:

- `context: {n_used} / {limit} ({pct}%)` — current window fill after this call (`n_in + n_out`). That is the overflow-prevention signal (gptme-discussions#138). Occupancy was already switched to input+output in 0a8aa9bcb after finding `9da01cffe7b1`.
- `+N out` — this turn's completion size.
- `session total` — cumulative billed tokens, summing `n_in + n_out` for every LLM call.

Prompt tokens are billed on every call, so counting full `n_in` each turn is the correct billed-usage total. It will exceed the context window; that is expected, not a bug. Folding session total into occupancy (or unique-token accounting) would collapse two different questions and lose the billed-usage signal. The label already says "session total", not "context used".

gptme/cli/main.py
TimeToBuildBob: Rejected: the COMMANDLINE-only assignment is the fix for `4ff46a12edfc`, not an omission.

Click's `is_flag=True` + `envvar=` treats any non-empty string as truthy, so `GPTME_TRACK_TOKENS=0` / `false` already makes `track_tokens=True`. The suggested `if track_tokens: os.environ["GPTME_TRACK_TOKENS"] = "1"` would then rewrite that falsy env string to `"1"`. `get_env_bool()` only treats `1/true/yes/on` as true, so leaving the original env string is how disable-via-env works.

This PR does not add config-file / `default_map` support for the option. If that is added later, it needs a source-aware three-way (command line → `"1"`, env with a falsy string → leave it, other non-default sources → `"1"`), not a truthy overwrite.

gptme/chat.py
TimeToBuildBob: Fixed in 4ec6b5aae: `_log_token_usage` now logs `track-tokens failed: …` on Exception instead of swallowing it. KeyboardInterrupt/SystemExit still propagate as BaseException. The two survival tests assert the warning is emitted.

TimeToBuildBob:
Pushed `4ec6b5aae` after in-band review (5/5, 0 findings).

- **logging** (`7cfef91acec6`): `_log_token_usage` now warns on Exception instead of swallowing it. Survival tests assert the warning.
- **session total** (`f2be3f14901d`): rejected. Occupancy is already the first field (`n_in+n_out` after 0a8aa9bcb); session total is billed API usage and is supposed to exceed the window.
- **env source** (`4c64aefba90c`): rejected. COMMANDLINE-only assignment is the `4ff46a12edfc` fix; a truthy overwrite would rewrite `GPTME_TRACK_TOKENS=0`/`false` to `"1"`.

Our AI-review threads on this head are resolved. CI is pending. Not merging (`gptme/chat.py` / `gptme/cli/main.py` are outside the self-merge allowlist).

TimeToBuildBob:
Self-merge not eligible on `4ec6b5aae`.

CI green, threads clear. Greptile 5/5 is stale vs this head and re-trigger is capped (8/8). Path policy still requires a human merge (`gptme/chat.py`, `gptme/cli/main.py`). Not merging; not re-triggering Greptile.

TimeToBuildBob:
Automated merge handoff for `4ec6b5aae552`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`

TimeToBuildBob (commented):

**2 findings** · ⚠️ **2** P2

gptme/chat.py
```diff
@@ -196,6 +261,12 @@ def chat(
         # Restore the caller's format so nested chat() calls (inline subagents)
         # don't clobber the parent's JSON mode when they exit.
         set_output_format(_prev_output_format)
+        # Restore the outer chat's token accumulator so same-context nested
+        # chat() calls don't corrupt the parent's running total.
+        # Thread-mode subagents start with a fresh contextvars context, so
+        # their totals are already isolated; folding inner into parent here
+        # would not see the parent list across threads.
+        _session_tokens.set(_prev_tokens)
```
TimeToBuildBob, line 270: ⚠️ **P2** — The session token accumulator in chat() resets at the start of each chat() call and restores the previous accumulator in the finally block without folding in the inner chat's tokens. When a subagent runs in the same thread (e.g., a tool that calls chat() synchronously), its LLM calls contribute to a fresh accumulator that is discarded on exit, so the parent session total undercounts actual token usage. Concretely, if the parent has accumulated 500 tokens and a same-thread subagent consumes 200, the parent's session total remains 500 instead of 700, misleading users who rely on --track-tokens to monitor total session usage. The fix would be to capture the inner accumulator's value before restoring it and add it to the parent's list.

TimeToBuildBob: Rejected: restoring the outer accumulator is the correct contract. Same finding previously disposed as 257d1720be54.

Thread-mode subagents (`gptme/tools/subagent/execution.py`) run in `threading.Thread`. On CPython's default GIL builds a new thread starts with an empty contextvars context, so the child's `_session_tokens` is independent of the parent. `chat()`'s `finally` cannot see the parent's list across threads, so adding the inner count before restore does not produce a session-wide total.

Same-context nested `chat()` is the case this restore protects (tests, TUI `copy_context`, future inline callers). Subagent tokens belong to a separate conversation and context window; mixing them into the parent's `session total` would conflate parent-window occupancy with a process-wide billing counter. That would be a different feature, not a one-line `finally` change.

Keeping restore-without-fold. `test_accumulator_restored_after_chat_returns` documents the contract. Not a bug / working as intended.

TimeToBuildBob:
Pushed `a7c4ad686` for the remaining AI-review P2s on `4ec6b5aae`.

- **architect path** (`fb85145231d0`): `--track-tokens` now logs the architect planning `llm_reply` (it sits outside `chat()`/`step()`). `chat()` still resets the accumulator for the editor session — different window/model.
- **nested fold-in** (`c3583c87d785`): rejected, same contract as `257d1720be54`. Restore-without-fold; thread-mode subagents isolate via a fresh contextvars context. Thread replied + resolved.

In-band review skipped (OpenRouter daily key limit) and fail-open pushed. Tests: `tests/test_track_tokens.py` 12/12. Did not trigger Greptile. Not merging (`gptme/chat.py`, `gptme/cli/main.py` outside allowlist).

TimeToBuildBob:
Automated merge handoff for `a7c4ad686f34`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

TimeToBuildBob:
Looked into this; no additional code change.

This morning's AI review of `a7c4ad686` scored 4/5 with one round-capped (non-blocking) P2 (`a77159886695`): fold inner `chat()` tokens into the parent accumulator, now framed around the architect→editor handoff.

Same contract already rejected as `257d1720be54` and `c3583c87d785`. Architect logs its own planning turn; `chat()` then resets for the editor session (different window/model). Restore-without-fold is intentional. No inline thread to dispose (round cap); no P0/P1.

CI green, threads clear, path policy still requires a human merge. Waiting for Erik.

TimeToBuildBob:
Rebased onto `master` (`7294393dc`) to clear CONFLICTING/DIRTY (38 commits behind, including #3669 429-survival in `chat.py`).

Conflict was indent-only in `gptme/chat.py`: master's `log_costs` sits outside the `with terminal_state_title(...)` block; the feature commit had it nested. Kept master's placement and added the `--track-tokens` log after it. Later review-fix commits replayed cleanly.

`uv run pytest tests/test_track_tokens.py -v` — 12/12 pass. Head is `0489af02a`. Path policy still requires a human merge (`gptme/chat.py`, `gptme/cli/main.py`).

TimeToBuildBob:
Automated merge handoff for `0489af02a00e`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

TimeToBuildBob:
Trimmed per today's triage: dropped the `--track-tokens` CLI flag, keeping only the `GPTME_TRACK_TOKENS` env var (mirrors the existing env-only `GPTME_COSTS` toggle). The env var already drove the tracking hook directly; the flag was a redundant second way to set it, and new default-CLI flags for opt-in toggles like this haven't landed before. Ready for maintainer review.

TimeToBuildBob:
The `ci_failure` dispatch for #3646 was for a **stale run** — run `34116873336` on SHA `456470bc` (11:29), which has since been force-pushed past. Current head is `38b108f` (12:03, after today's trim dropping the `--track-tokens` CLI flag in favor of the env-only toggle).

That older run's failure was `tests/test_prompts.py::test_get_prompt_full` (`assert 8162 < 8100`), a prompt-length guard. It passes locally on the current head (1.28s), the PR diff touches only `chat.py`/`cli/main.py`/`test_track_tokens.py` — nothing that changes the system prompt — and master CI is green, so this was superseded drift, not a PR regression. Fresh CI on `38b108f` is now running.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto `master` (`7f366a56c`) to clear CONFLICTING/DIRTY.

Conflict was comment-only in `tests/test_prompts.py`: both sides bumped the ceiling to 8250, with slightly different comment text. Kept HEAD's phrasing. All 12 commits replayed cleanly (the test_prompts bump became a no-op and was dropped).

CI running on new head. Self-merge path still requires maintainer review (`gptme/chat.py`, `gptme/cli/main.py`).

TimeToBuildBob:
Automated merge handoff for `7f366a56c99a`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

This head is ready for maintainer review and **manual merge**. Project monitoring marked it `human_merge_required` and will not attempt an automated merge. A new head will be evaluated independently.

gptme/chat.py
```diff
@@ -51,6 +54,59 @@
...
+
+        # Resolve once from the caller-supplied name (alias or full). Tokenizer
+        # and context metadata then share that ModelMeta; passing .full back
+        # into get_model() is the lookup 48225d823 avoided.
+        resolved = get_model(model)
+        n_in = len_tokens(msgs, resolved.full)
+        n_out = len_tokens(msg_response, resolved.full)
+        context_limit = resolved.context
+        n_used = n_in + n_out
+        session_tokens[0] += n_used
```
TimeToBuildBob: ❌ **P1** — The `_session_tokens` ContextVar stores a mutable list that is shared by reference across context copies. When a new thread starts, it inherits a copy of the current context, but the list object is the same instance. Any thread that calls `_log_token_usage` directly — for example a thread-mode subagent that executes `step()` outside a `chat()` call, or a server worker thread — will execute `session_tokens[0] += n_used` on the parent's list, corrupting the parent's running total. The comment in `chat()` claims thread-mode subagents are isolated because they start with a fresh context, but a thread copy of a ContextVar shares mutable objects; only `chat()`'s reset-and-restore sequence provides isolation. This is a real concurrency defect: the session total printed in the main chat will be inflated by subagent token counts, making the displayed number wrong. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob: Fixed in 8e46bef62 — `_session_tokens` now stores an immutable int and rebinds via `.set()`, so `copy_context()` children (server/TUI/ACP step threads) cannot mutate the parent's running total. Added `test_copied_context_does_not_mutate_parent_accumulator` for the no-reset `step()`-outside-`chat()` path.

TimeToBuildBob:
Pushed `8e46bef62` after in-band review (5/5, 0 findings).

- **mutable ContextVar list** (`2d93d938c459`): `_session_tokens` now stores an immutable int. `copy_context()` copies the binding, not the value — a list was shared by reference so a child `step()` inflated the parent total. Rebind via `.set()`. Regression: `test_copied_context_does_not_mutate_parent_accumulator`.

CI pending. Not merging (`gptme/chat.py` is outside the self-merge allowlist).

TimeToBuildBob:
Automated merge handoff for `8e46bef624e1`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/cli/main.py`
- `tests/test_track_tokens.py`

TimeToBuildBob:
Rebased onto `master` (`c485dfa6b`) to clear CONFLICTING/DIRTY (42 commits behind, including #3750 ExitStack skill-lifecycle in `chat.py`).

Conflict was import-only in `gptme/chat.py`: master added `from contextlib import ExitStack`; this branch added `from contextvars import ContextVar`. Kept both. Remaining 12 commits replayed cleanly. `tests/test_track_tokens.py` 13/13. Path policy still requires a human merge (`gptme/chat.py`, `gptme/cli/main.py`).

PR #3647, opened 2026-08-27
feat(browser): skip html→markdown when server returns text/markdown

### LINKED ISSUE #3644: feat(browser): send Accept: text/markdown for up to 10× token reduction on docs
## Summary

HTTP content negotiation via `Accept: text/markdown` is now a real, shipping pattern. When an AI agent sends this header, CDN edges (Cloudflare's "Markdown for Agents" feature, deployed Feb 2026) return clean markdown instead of HTML — stripping nav, scripts, styles, and layout wrappers. Practical impact: **10× token reduction on documentation pages**.

gptme's URL fetching currently sends bare `Accept: */*`, so we never get this benefit even when the upstream site supports it.

## The Change

Add `Accept: text/markdown, text/html;q=0.9, */*;q=0.8` to gptme's HTTP client headers when fetching URLs.

- If the server returns `Content-Type: text/markdown`: use the body directly (skip our html→markdown conversion step)
- If the server returns `406 Not Acceptable` or plain HTML: fall back to current behavior

Cloudflare's response also includes `x-markdown-tokens` header showing token count — could be surfaced as telemetry.

## Coverage

Cloudflare's zone-level opt-in covers a massive fraction of the web, including most popular documentation sites (PyPI, npm, GitHub pages-hosted docs, etc.). Vercel/Next.js has native support. acceptmarkdown.com tracks adoption.

## References

- https://acceptmarkdown.com/ — spec and adoption tracker
- https://developers.cloudflare.com/changelog/2026-02-12-markdown-for-agents
- https://benword.com/serving-markdown-to-ai-agents-via-accept-headers

## Effort

Low — likely a 5–10 line change in the HTTP client + response handling. The fallback path is already in place (we do html→markdown today).

TimeToBuildBob:
## Summary

Completes the `Accept: text/markdown` content negotiation round-trip in the Playwright browser backend.

The `Accept: text/markdown, text/plain, text/html;q=0.9, */*;q=0.8` header was already being sent on every page load (see `_load_page` and `_open_page`). But the response handling always ran `html_to_markdown()` regardless of what the server returned.

When a server honours the `Accept` header and responds with `Content-Type: text/markdown` (Cloudflare's [Markdown for Agents](https://developers.cloudflare.com/changelog/2026-02-12-markdown-for-agents) feature, adopted by many documentation sites since Feb 2026), the browser renders it as plain text in `<body>`. Piping that through `html_to_markdown` is redundant and can introduce artifacts.

## Changes

- Capture and normalize the navigation response's `Content-Type` in `_load_page`
- Return the Markdown classification alongside the content so metadata stays scoped to the corresponding read
- If the response is `text/markdown`, return `page.inner_text("body")` directly instead of extracting HTML
- In `read_url`, skip Pandoc for direct Markdown while retaining inline data-image sanitization
- Cover media-type variants, per-call metadata propagation, conversion bypass, and data-image stripping

## Fallback

No change to existing behaviour for any server that returns `text/html`, `406 Not Acceptable`, or any other content type. The Accept header is still sent; this only changes what happens when the server honours it.

## References

- Fixes #3644
- https://acceptmarkdown.com/ — spec and adoption tracker
- https://developers.cloudflare.com/changelog/2026-02-12-markdown-for-agents

### DIFF
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -339,6 +339,7 @@ def _extract_main_content(page: Page) -> str:
 
 def read_url(url: str) -> str:
     """Read the text of a webpage and return the text in Markdown format."""
+    global _last_response_content_type
     body_content = _execute_with_retry(_load_page, url)
     # Server responded with markdown directly — return as-is (skip html→markdown)
     if "text/markdown" in _last_response_content_type:
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -33,6 +33,7 @@ from ._computer_gate import sensitive_action_gate
 
 _browser: BrowserThread | None = None
 _last_logs: dict = {"logs": [], "errors": [], "url": None}
+_last_response_content_type: str = ""
 # Persistent page state for interactive browsing (open_page/click/fill/scroll)
 _current_page: Page | None = None
 _current_context: BrowserContext | None = None
@@ -158,8 +159,8 @@ def _execute_with_retry(
 
 
 def _load_page(browser: Browser, url: str) -> str:
-    """Load a page and return its body HTML, always capturing logs"""
-    global _last_logs
+    """Load a page and return its body HTML (or markdown text), always capturing logs"""
+    global _last_logs, _last_response_content_type
 
     managed = _create_page(
         browser,
@@ -196,21 +197,29 @@ def _load_page(browser: Browser, url: str) -> str:
     page.on("pageerror", on_page_error)
 
     # Navigate to the page
+    nav_response = None
     try:
-        page.goto(url)
+        nav_response = page.goto(url)
         # Wait for page to be fully loaded (includes network idle)
         page.wait_for_load_state("networkidle")
     except Exception as e:
         page_errors.append(f"Navigation error: {e}")
         # Don't re-raise, just capture the error
 
+    # Track content type so read_url can skip html→markdown when not needed
+    _last_response_content_type = (
+        nav_response.headers.get("content-type", "") if nav_response else ""
+    )
+
     # Store logs globally
     _last_logs = {"logs": logs, "errors": page_errors, "url": url}
 
     try:
-        # Try to extract main content area first, falling back to body
-        html = _extract_main_content(page)
-        return html
+        # Server returned markdown directly — use plain text, skip HTML extraction
+        if "text/markdown" in _last_response_content_type:
+            return page.inner_text("body")
+        # Otherwise extract main content HTML for html_to_markdown conversion
+        return _extract_main_content(page)
     finally:
         managed.close()
 
@@ -330,8 +339,11 @@ def _extract_main_content(page: Page) -> str:
 
 def read_url(url: str) -> str:
     """Read the text of a webpage and return the text in Markdown format."""
-    body_html = _execute_with_retry(_load_page, url)
-    return html_to_markdown(body_html)
+    body_content = _execute_with_retry(_load_page, url)
+    # Server responded with markdown directly — return as-is (skip html→markdown)
+    if "text/markdown" in _last_response_content_type:
+        return body_content
+    return html_to_markdown(body_content)
 
 
 def read_logs() -> str:
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -213,9 +213,9 @@ def _load_page(browser: Browser, url: str) -> tuple[str, bool]:
     _last_logs = {"logs": logs, "errors": page_errors, "url": url}
 
     try:
-        # Server returned markdown directly — use plain text, skip HTML extraction
+        # Server returned markdown directly — preserve source whitespace and skip HTML extraction
         if is_markdown:
-            return page.inner_text("body"), True
+            return page.text_content("body") or "", True
         # Otherwise extract main content HTML for html_to_markdown conversion
         return _extract_main_content(page), False
     finally:
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -37,6 +37,7 @@ _last_logs: dict = {"logs": [], "errors": [], "url": None}
 _current_page: Page | None = None
 _current_context: BrowserContext | None = None
 logger = logging.getLogger(__name__)
+_inline_data_image = re.compile(r"!\[[^\]]*\]\(data:image[^)]*\)")
 
 
 def _is_cdp_connection() -> bool:
@@ -338,7 +339,7 @@ def read_url(url: str) -> str:
     """Read the text of a webpage and return the text in Markdown format."""
     body_content, is_markdown = _execute_with_retry(_load_page, url)
     if is_markdown:
-        return body_content
+        return _inline_data_image.sub("", body_content)
     return html_to_markdown(body_content)
 
 
@@ -1169,12 +1170,6 @@ def html_to_markdown(html):
     markdown = re.sub(r"\{(#|style|target|\.)[^}]*\}", "", markdown)
 
     # strip inline images, like: data:image/png;base64,...
-    re_strip_data = re.compile(r"!\[[^\]]*\]\(data:image[^)]*\)")
-
-    # test cases
-    assert re_strip_data.sub("", "![test](data:image/png;base64,123)") == ""
-    assert re_strip_data.sub("", "![test](data:image/png;base64,123) test") == " test"
-
-    markdown = re_strip_data.sub("", markdown)
+    markdown = _inline_data_image.sub("", markdown)
 
     return markdown
--- a/gptme/tools/_browser_playwright.py
+++ b/gptme/tools/_browser_playwright.py
@@ -33,7 +33,6 @@ from ._computer_gate import sensitive_action_gate
 
 _browser: BrowserThread | None = None
 _last_logs: dict = {"logs": [], "errors": [], "url": None}
-_last_response_content_type: str = ""
 # Persistent page state for interactive browsing (open_page/click/fill/scroll)
 _current_page: Page | None = None
 _current_context: BrowserContext | None = None
@@ -158,9 +157,9 @@ def _execute_with_retry(
     raise last_error
 
 
-def _load_page(browser: Browser, url: str) -> str:
-    """Load a page and return its body HTML (or markdown text), always capturing logs"""
-    global _last_logs, _last_response_content_type
+def _load_page(browser: Browser, url: str) -> tuple[str, bool]:
+    """Load a page and return its content plus whether it is Markdown."""
+    global _last_logs
 
     managed = _create_page(
         browser,
@@ -206,20 +205,18 @@ def _load_page(browser: Browser, url: str) -> str:
         page_errors.append(f"Navigation error: {e}")
         # Don't re-raise, just capture the error
 
-    # Track content type so read_url can skip html→markdown when not needed
-    _last_response_content_type = (
-        nav_response.headers.get("content-type", "") if nav_response else ""
-    )
+    content_type = nav_response.headers.get("content-type", "") if nav_response else ""
+    is_markdown = content_type.partition(";")[0].strip().lower() == "text/markdown"
 
     # Store logs globally
     _last_logs = {"logs": logs, "errors": page_errors, "url": url}
 
     try:
         # Server returned markdown directly — use plain text, skip HTML extraction
-        if "text/markdown" in _last_response_content_type:
-            return page.inner_text("body")
+        if is_markdown:
+            return page.inner_text("body"), True
         # Otherwise extract main content HTML for html_to_markdown conversion
-        return _extract_main_content(page)
+        return _extract_main_content(page), False
     finally:
         managed.close()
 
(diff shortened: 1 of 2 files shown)


gptme/tools/_browser_playwright.py
```diff
@@ -196,21 +197,29 @@ def on_page_error(error):
...
+
     # Store logs globally
     _last_logs = {"logs": logs, "errors": page_errors, "url": url}
 
     try:
-        # Try to extract main content area first, falling back to body
-        html = _extract_main_content(page)
-        return html
+        # Server returned markdown directly — use plain text, skip HTML extraction
+        if "text/markdown" in _last_response_content_type:
```
TimeToBuildBob: ⚠️ **P2** — The check `if "text/markdown" in _last_response_content_type` is case-sensitive, but HTTP media types are case-insensitive. A server that responds with `Content-Type: Text/Markdown` or `TEXT/MARKDOWN` (allowed by RFC 2045 and common in non-conforming implementations) will not match. As a result, `_load_page` will call `_extract_main_content` and `read_url` will run `html_to_markdown` on what is actually markdown, mangling the output (e.g., adding backticks or escaping characters). This defeats the PR's purpose and can corrupt the content returned to the LLM. The fix is to lower-case the stored content-type before the substring test. This affects both places: line 219 in `_load_page` and line 344 in `read_url`.

TimeToBuildBob:
## 🤖 AI code review

This PR makes the Playwright browser backend skip HTML-to-Markdown conversion when the server responds with Content-Type: text/markdown. It modifies _load_page to capture the navigation response's content-type and return a tuple (content, is_markdown). read_url then either returns the markdown directly (after stripping inline data images) or falls back to html_to_markdown. It also moves the inline data-image regex to module level and removes inline test assertions from html_to_markdown.

- `gptme/tools/_browser_playwright.py` — Changes _load_page to return a tuple (content, is_markdown) based on the response content-type, and read_url to bypass html_to_markdown for text/markdown responses. Moves the data-image stripping regex to module level.
- `tests/test_tools_browser.py` — Adds tests for markdown response handling in _load_page and read_url, covering content-type variants, conversion bypass, and data-image stripping.

</details>

| `91585a373eb1` | 4/5 | 1 | llm | 2026-08-27 09:51 UTC |
| `0f601ac007bc` | 4/5 | 1 | llm | 2026-08-27 10:40 UTC |
| `e3366eb1fcdc` | 3/5 | 4 | llm | 2026-08-27 12:25 UTC |
| `bd309c1a8155` | 3/5 | 1 | llm | 2026-08-27 15:39 UTC |

</details>

<sub>Reviewed `f8237395434c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 286s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

gptme/tools/_browser_playwright.py
```diff
@@ -330,8 +339,12 @@ def _extract_main_content(page: Page) -> str:
 
 def read_url(url: str) -> str:
     """Read the text of a webpage and return the text in Markdown format."""
-    body_html = _execute_with_retry(_load_page, url)
-    return html_to_markdown(body_html)
+    global _last_response_content_type
+    body_content = _execute_with_retry(_load_page, url)
+    # Server responded with markdown directly — return as-is (skip html→markdown)
+    if "text/markdown" in _last_response_content_type:
```
TimeToBuildBob: ⚠️ **P2** — Module-level `_last_response_content_type` is written on the browser thread inside `_load_page` and read on the caller thread in `read_url` after `_execute_with_retry` returns. There is no synchronization between the two threads, so if two `read_url` calls run concurrently, the second call's `_load_page` can overwrite the global before the first call reads it. Concretely, call A gets `text/markdown`, sets the global, and returns; call B then gets `text/html` and sets the global to `text/html` before call A's `read_url` executes the check. Call A then sees `text/html`, runs `html_to_markdown` on markdown text, mangling the output. The same corruption occurs when call A is HTML and call B is markdown. Fix by returning the content type as part of `_load_page`'s return value (e.g., a tuple) or by storing it in a thread-local.

gptme/tools/_browser_playwright.py
```diff
@@ -33,6 +33,7 @@
 
 _browser: BrowserThread | None = None
 _last_logs: dict = {"logs": [], "errors": [], "url": None}
+_last_response_content_type: str = ""
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e3366eb1f: `_load_page` now returns the Markdown classification with the content, so each `read_url` call uses metadata from its own response instead of shared module state. Added regression coverage for the per-call path.

gptme/tools/_browser_playwright.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e3366eb1f: the media type is normalized with `strip().lower()` before exact comparison. Coverage includes `Text/Markdown` and a charset parameter.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `e3366eb1fcdc`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/_browser_playwright.py`

gptme/tools/_browser_playwright.py
```diff
@@ -196,21 +196,27 @@ def on_page_error(error):
...
     try:
-        page.goto(url)
+        nav_response = page.goto(url)
         # Wait for page to be fully loaded (includes network idle)
         page.wait_for_load_state("networkidle")
     except Exception as e:
         page_errors.append(f"Navigation error: {e}")
         # Don't re-raise, just capture the error
 
+    content_type = nav_response.headers.get("content-type", "") if nav_response else ""
```
TimeToBuildBob, line 209: ❌ **P1** — On cache reloads, a server may reply 304 Not Modified without including a Content-Type header. In _load_page, when nav_response is not None but its headers contain no 'content-type' key, content_type becomes an empty string, so is_markdown is set to False even though the cached resource is markdown. The function then falls through to _extract_main_content(page) and read_url later invokes html_to_markdown on the extracted HTML. Since the markdown is rendered as plain text in the body, pandoc will treat it as HTML and wrap it in code fences or otherwise transform it, producing different (and likely worse) output than the first load. This makes read_url non-idempotent across reloads of the same URL. _(1/2 passes — pass 2 looked and disagreed)_

tests/test_tools_browser.py
```diff
@@ -987,6 +987,49 @@ def test_examples_contain_pdf(self):
         assert "pdf" in result.lower() or "PDF" in result
 
 
+# ── Playwright Markdown response handling ────────────────────────────
+
+
+@pytest.mark.skipif(not has_playwright(), reason="playwright not installed")
```
TimeToBuildBob, line 993: ⚠️ **P2** — The test class TestPlaywrightMarkdownResponse is decorated with @pytest.mark.skipif(not has_playwright(), ...). However, every test in the class only uses unittest.mock MagicMock objects and does not touch the Playwright runtime, so it would pass even when Playwright is not installed. As written, the tests are skipped in any environment without Playwright, meaning the new markdown-detection code in _load_page and read_url is never exercised there. A regression in that logic (like the 304-handling issue) would therefore go undetected in CI configurations that lack Playwright. _(1/2 passes — pass 2 looked and disagreed)_

gptme/tools/_browser_playwright.py
```diff
@@ -330,8 +336,10 @@ def _extract_main_content(page: Page) -> str:
 
 def read_url(url: str) -> str:
     """Read the text of a webpage and return the text in Markdown format."""
-    body_html = _execute_with_retry(_load_page, url)
-    return html_to_markdown(body_html)
+    body_content, is_markdown = _execute_with_retry(_load_page, url)
+    if is_markdown:
```
TimeToBuildBob, line 341: ⚠️ **P2** — When `is_markdown` is True, `read_url` returns `body_content` directly, skipping `html_to_markdown`. However, `html_to_markdown` strips inline data URLs via a regex (`re_strip_data`). A server that honors the `Accept: text/markdown` header and returns a markdown document containing an inline `data:image/png;base64,...` image will now pass that full base64 blob to the LLM, inflating context and potentially leaking inline content. Previously the data URL was stripped regardless of source content type because `read_url` always ran `html_to_markdown`. The fix would be to still apply the data-URL removal (or a dedicated cleanup) before returning the markdown. _(1/2 passes — pass 1 looked and disagreed)_

gptme/tools/_browser_playwright.py
TimeToBuildBob, line 209: ⚠️ **P2** — The PR description states the Content-Type is stored in a `_last_response_content_type` module global, but the diff does not add such a global. The `content_type` variable in `_load_page` is computed locally and only used to derive `is_markdown`; it is not stored. Any future code expecting to read the content type from a module global (e.g., to surface it in `read_logs` or for debugging) will fail or behave differently than described. This is a discrepancy between the documented intent and the implementation, which could mislead maintainers. _(1/2 passes — pass 1 looked and disagreed)_

gptme/tools/_browser_playwright.py
TimeToBuildBob: Fixed in bd309c1a8: direct Markdown responses now use the same inline data-image sanitizer as the Pandoc path. Added a regression test covering the direct Markdown path.

gptme/tools/_browser_playwright.py
TimeToBuildBob: Not applicable to Playwright navigation. `page.goto()` resolves with the final loaded document response; a browser HTTP-cache revalidation may use 304 on the wire, but Playwright exposes the resulting resource response rather than a standalone 304 document without its effective headers. There is also no cross-call cache metadata available here that could safely infer a missing type; guessing Markdown from body text would misclassify ordinary plain-text and HTML responses.

tests/test_tools_browser.py
TimeToBuildBob: The module under test imports `playwright.sync_api` at import time, so these tests cannot run in a no-Playwright environment merely because their page objects are mocked. The class-level availability guard is therefore required; CI with the browser extra installed executes the tests.

gptme/tools/_browser_playwright.py
TimeToBuildBob: Fixed by updating the PR description to match the current implementation: response classification is returned per call rather than stored in module-global state. The per-call design is deliberate because it avoids cross-read metadata races.

TimeToBuildBob:
Addressed the latest AI-review round on `bd309c1a8`:

- preserved inline data-image stripping for direct Markdown responses and added regression coverage;
- rejected the 304 finding because Playwright exposes the final document response, not a standalone cache-validation response without effective metadata;
- retained the Playwright availability guard because the tested module imports `playwright.sync_api` at import time;
- updated the PR description to match the per-call metadata design.

Targeted browser tests pass (5/5), Ruff passes, and all review threads are resolved. CI is running on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
**Status: all green ✅** — CI fully passing (all test jobs + Greptile) on head `bd309c1a8`. Ready for maintainer merge whenever convenient.

TimeToBuildBob:
Automated merge handoff for `bd309c1a8155`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/_browser_playwright.py`

gptme/tools/_browser_playwright.py
```diff
@@ -196,21 +197,27 @@ def on_page_error(error):
...
     # Store logs globally
     _last_logs = {"logs": logs, "errors": page_errors, "url": url}
 
     try:
-        # Try to extract main content area first, falling back to body
-        html = _extract_main_content(page)
-        return html
+        # Server returned markdown directly — use plain text, skip HTML extraction
+        if is_markdown:
+            return page.inner_text("body"), True
```
TimeToBuildBob: ❌ **P1** — In `_load_page`, when a server returns `text/markdown`, the content is obtained with `page.inner_text("body")`. `inner_text` returns the rendered text and normalizes whitespace (collapses multiple spaces, trims edges), which corrupts Markdown formatting that relies on precise spacing, such as indented code blocks, tables, and line breaks. For example, a fenced code block with 4-space indentation would have leading spaces removed, breaking the code block. The correct method is `page.text_content("body")` (or `page.evaluate("document.body.textContent")`), which preserves the raw whitespace. This is a real defect because the primary purpose of this change is to read raw markdown from the server, and whitespace normalization degrades the content. _(2/2 passes)_

TimeToBuildBob: Fixed in f82373954: direct Markdown now uses `text_content("body")` so source whitespace is preserved. The regression fixture includes leading indentation and table spacing, asserts exact round-trip content, and verifies `inner_text` is not called.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `f8237395434c`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/_browser_playwright.py`

TimeToBuildBob:
Merged by Erik in `0565d1548eb1`. CI was green and all review findings were resolved on the merged head; no further action is needed.

PR #3648, opened 2026-08-27
feat(daemon): Phase 1 MVP — persistent background session daemon

TimeToBuildBob:
## Summary

- Adds `gptme daemon` command group for managing persistent background sessions
- Sessions survive terminal close, SSH drops, and laptop sleep via Unix domain socket IPC
- Implements tmux-style attach/detach UX: `gptme daemon start`, `gptme daemon attach`

## Design

**Architecture**: Unix domain socket + 4-byte length-prefixed JSON protocol. Chosen over HTTP for lower latency, simpler implementation, and `mode 0600` user-isolation.

**New files**:
- `gptme/server/ipc_protocol.py` — wire format codec (`IPCMessage`, `send_msg`, `recv_msg`)
- `gptme/server/daemon.py` — `SessionDaemon` class (double-fork daemonize, I/O pump, accept loop, 64KB ring-buffer history replay)
- `gptme/cli/cmd_daemon.py` — Click subcommands: `start`, `attach`, `list`, `stop`, `status`
- `tests/test_daemon.py` — 19 tests covering protocol roundtrip, path helpers, daemon state, attach protocol, CLI smoke

**Design decisions** (resolved, from architecture doc `knowledge/technical-designs/gptme-session-daemon-architecture.md`):
- Auto-start on attach (Option A — better UX, matches tmux)
- Socket at `~/.config/gptme/daemon/<name>.socket` via `get_config_dir()` (survives `/tmp` cleanup)  
- Single-client MVP — multi-client observe-only fan-out is Phase 3
- In-flight recovery is Phase 2; MVP reuses existing JSONL persistence

## Usage

```bash
# Start a background session
gptme daemon start --session mywork "Implement feature X"

# Re-attach from any terminal (auto-starts if not running)
gptme daemon attach mywork

# List and manage
gptme daemon list
gptme daemon status mywork
gptme daemon stop mywork

# Also available as standalone binary
gptme-daemon start --session mywork
```

## Test plan

- [x] `pytest tests/test_daemon.py` — 19/19 pass
- [x] ruff clean (TC003, B904, F401 violations fixed)
- [x] mypy clean on daemon files (pre-existing acp errors unrelated)
- [ ] Manual smoke test: `gptme daemon start --session test "say hello" && gptme daemon attach test`

### DIFF
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -1,0 +1,208 @@
+"""gptme daemon — persistent background session management.
+
+Commands:
+  gptme daemon start [--session NAME] [PROMPT...]  Start a daemon session
+  gptme daemon attach NAME                         Attach to a running daemon
+  gptme daemon list                                List running daemons
+  gptme daemon stop NAME                           Stop a daemon session
+  gptme daemon status NAME                         Show daemon status
+
+Design notes:
+- attach auto-starts a daemon when the session does not yet exist
+- Single-client MVP: only one attach at a time (Phase 3 adds multi-client)
+- Non-interactive mode: prompts are passed as arguments, not typed interactively
+  (PTY/interactive support is Phase 2)
+"""
+
+from __future__ import annotations
+
+import logging
+
+import click
+
+logger = logging.getLogger(__name__)
+
+
+@click.group("daemon")
+def cli() -> None:
+    """Manage persistent background gptme sessions."""
+
+
+@cli.command("start")
+@click.option(
+    "--session",
+    "-s",
+    default=None,
+    help="Session name (defaults to a timestamp-based name)",
+)
+@click.option("--model", "-m", default=None, help="Model to use")
+@click.option("--workspace", "-w", default=None, help="Workspace directory")
+@click.argument("prompts", nargs=-1)
+def start(
+    session: str | None,
+    model: str | None,
+    workspace: str | None,
+    prompts: tuple[str, ...],
+) -> None:
+    """Start a gptme session as a background daemon.
+
+    The session runs non-interactively in the background and writes to its
+    conversation log. Attach later with 'gptme daemon attach NAME' to watch
+    output or inject additional prompts.
+
+    Example:
+
+        gptme daemon start --session mywork "Implement feature X"
+        gptme daemon attach mywork
+    """
+    from ..server.daemon import SessionDaemon, get_socket_path
+
+    if session is None:
+        from datetime import datetime, timezone
+
+        session = datetime.now(timezone.utc).strftime("daemon-%Y%m%dT%H%M%S")
+
+    sock_path = get_socket_path(session)
+    if sock_path.exists():
+        daemon = SessionDaemon(session)
+        if daemon.is_running():
+            click.echo(
+                f"Daemon '{session}' is already running (socket: {sock_path})", err=True
+            )
+            raise SystemExit(1)
+
+    # Build gptme args
+    gptme_args: list[str] = ["--name", session]
+    if model:
+        gptme_args += ["--model", model]
+    if workspace:
+        gptme_args += ["--workspace", workspace]
+    gptme_args += list(prompts)
+
+    daemon = SessionDaemon(session)
+    click.echo(f"Starting daemon '{session}'...")
+
+    # daemonize=True → double-fork; parent returns here, child runs the daemon
+    daemon.start(gptme_args, daemonize=True)
+
+    # Parent continues here
+    click.echo(f"Daemon started. Attach with: gptme daemon attach {session}")
+    click.echo(f"Socket: {sock_path}")
+
+
+@cli.command("attach")
+@click.argument("session")
+@click.option(
+    "--start-if-missing",
+    is_flag=True,
+    default=True,
+    show_default=True,
+    help="Auto-start the daemon if the session is not running (default: on)",
+)
+def attach_cmd(session: str, start_if_missing: bool) -> None:
+    """Attach to a running daemon session.
+
+    Prints buffered output from session start, then relays stdin/stdout in
+    real-time. Detach with Ctrl-D or Ctrl-C — the daemon keeps running.
+    """
+    from ..server.daemon import SessionDaemon, attach, get_socket_path
+
+    sock_path = get_socket_path(session)
+    if not sock_path.exists():
+        if not start_if_missing:
+            click.echo(
+                f"No running daemon for session '{session}'. Start with: gptme daemon start --session {session}",
+                err=True,
+            )
+            raise SystemExit(1)
+
+        click.echo(f"Session '{session}' not running — starting it now...", err=True)
+        daemon = SessionDaemon(session)
+        daemon.start(["--name", session], daemonize=True)
+
+        # Wait briefly for the socket to appear
+        import time
+
+        for _ in range(20):
+            if sock_path.exists():
+                break
+            time.sleep(0.2)
+        else:
+            click.echo("Daemon did not start in time (socket not found).", err=True)
+            raise SystemExit(1)
+
+    try:
+        attach(session)
+    except FileNotFoundError as e:
+        click.echo(str(e), err=True)
+        raise SystemExit(1) from e
+    except ConnectionRefusedError as e:
+        click.echo(
+            f"Could not connect to daemon '{session}'. It may have exited.", err=True
+        )
+        raise SystemExit(1) from e
+
+
+@cli.command("list")
+@click.option("--json", "as_json", is_flag=True, help="Output as JSON")
+def list_cmd(as_json: bool) -> None:
+    """List running daemon sessions."""
+    from ..server.daemon import list_daemons
+
+    daemons = list_daemons()
+    if as_json:
+        import json
+
+        click.echo(json.dumps(daemons, indent=2))
+        return
+
+    if not daemons:
+        click.echo("No daemon sessions found.")
+        return
+
+    click.echo(f"{'SESSION':<20} {'PID':<8} {'STATUS':<10} SOCKET")
+    for d in daemons:
+        pid = str(d["pid"]) if d["pid"] else "-"
+        status = "running" if d["running"] else "stopped"
+        sock = d["socket"] or "-"
+        click.echo(f"{d['session']:<20} {pid:<8} {status:<10} {sock}")
+
+
+@cli.command("stop")
+@click.argument("session")
+def stop_cmd(session: str) -> None:
+    """Stop a running daemon session (sends SIGTERM)."""
+    from ..server.daemon import SessionDaemon
+
+    daemon = SessionDaemon(session)
+    if not daemon.is_running():
+        click.echo(f"Daemon '{session}' is not running.", err=True)
+        raise SystemExit(1)
+
+    daemon.stop()
+    click.echo(f"Stopped daemon '{session}'.")
+
+
+@cli.command("status")
+@click.argument("session")
+def status_cmd(session: str) -> None:
+    """Show status of a daemon session."""
+    from ..server.daemon import SessionDaemon, get_socket_path
+
+    daemon = SessionDaemon(session)
+    sock_path = get_socket_path(session)
+    running = daemon.is_running()
+    click.echo(f"Session:  {session}")
+    click.echo(f"Status:   {'running' if running else 'stopped'}")
+    click.echo(
+        f"Socket:   {sock_path} ({'exists' if sock_path.exists() else 'missing'})"
+    )
+
+
+# Allow `gptme-daemon` as a standalone entry point
+def main() -> None:
+    cli()
+
+
+if __name__ == "__main__":
+    main()
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -23,6 +23,22 @@ import click
 logger = logging.getLogger(__name__)
 
 
+class SessionName(click.ParamType):
+    """Click type for daemon session names stored as path components."""
+
+    name = "TEXT"
+
+    def convert(self, value, param, ctx):
+        from ..util.conversation_ids import conversation_id_error
+
+        if error := conversation_id_error(value):
+            self.fail(error, param, ctx)
+        return value
+
+
+_SESSION_NAME = SessionName()
+
+
 @click.group("daemon")
 def cli() -> None:
     """Manage persistent background gptme sessions."""
@@ -33,6 +49,7 @@ def cli() -> None:
     "--session",
     "-s",
     default=None,
+    type=_SESSION_NAME,
     help="Session name (defaults to a timestamp-based name)",
 )
 @click.option("--model", "-m", default=None, help="Model to use")
@@ -91,7 +108,7 @@ def start(
 
 
 @cli.command("attach")
-@click.argument("session")
+@click.argument("session", type=_SESSION_NAME)
 @click.option(
     "--start-if-missing/--no-start-if-missing",
     default=True,
@@ -136,10 +153,8 @@ def attach_cmd(session: str, start_if_missing: bool) -> None:
     except FileNotFoundError as e:
         click.echo(str(e), err=True)
         raise SystemExit(1) from e
-    except (ConnectionRefusedError, ConnectionResetError) as e:
-        click.echo(
-            f"Could not connect to daemon '{session}'. It may have exited.", err=True
-        )
+    except (ConnectionRefusedError, ConnectionResetError, TimeoutError) as e:
+        click.echo(f"Could not attach to daemon '{session}': {e}", err=True)
         raise SystemExit(1) from e
 
 
@@ -169,7 +184,7 @@ def list_cmd(as_json: bool) -> None:
 
 
 @cli.command("stop")
-@click.argument("session")
+@click.argument("session", type=_SESSION_NAME)
 def stop_cmd(session: str) -> None:
     """Stop a running daemon session (sends SIGTERM)."""
     from ..server.daemon import SessionDaemon
@@ -184,7 +199,7 @@ def stop_cmd(session: str) -> None:
 
 
 @cli.command("status")
-@click.argument("session")
+@click.argument("session", type=_SESSION_NAME)
 def status_cmd(session: str) -> None:
     """Show status of a daemon session."""
     from ..server.daemon import SessionDaemon, get_socket_path
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -194,7 +194,11 @@ def stop_cmd(session: str) -> None:
         click.echo(f"Daemon '{session}' is not running.", err=True)
         raise SystemExit(1)
 
-    daemon.stop()
+    try:
+        daemon.stop()
+    except TimeoutError as e:
+        click.echo(f"Timed out waiting for daemon '{session}' to stop.", err=True)
+        raise SystemExit(1) from e
     click.echo(f"Stopped daemon '{session}'.")
 
 
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -199,6 +199,9 @@ def stop_cmd(session: str) -> None:
     except TimeoutError as e:
         click.echo(f"Timed out waiting for daemon '{session}' to stop.", err=True)
         raise SystemExit(1) from e
+    except RuntimeError as e:
+        click.echo(str(e), err=True)
+        raise SystemExit(1) from e
     click.echo(f"Stopped daemon '{session}'.")
 
 
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -71,19 +71,19 @@ def start(
             )
             raise SystemExit(1)
 
-    # Build gptme args
+    # Build the stable arguments reused for each queued turn. Prompts are kept
+    # separate so later attached input can run as another turn in this session.
     gptme_args: list[str] = ["--name", session]
     if model:
         gptme_args += ["--model", model]
     if workspace:
         gptme_args += ["--workspace", workspace]
-    gptme_args += list(prompts)
 
     daemon = SessionDaemon(session)
     click.echo(f"Starting daemon '{session}'...")
 
     # daemonize=True → double-fork; parent returns here, child runs the daemon
-    daemon.start(gptme_args, daemonize=True)
+    daemon.start(gptme_args, prompts=list(prompts), daemonize=True)
 
     # Parent continues here
     click.echo(f"Daemon started. Attach with: gptme daemon attach {session}")
@@ -118,17 +118,18 @@ def attach_cmd(session: str, start_if_missing: bool) -> None:
 
         click.echo(f"Session '{session}' not running — starting it now...", err=True)
         daemon = SessionDaemon(session)
-        daemon.start(["--name", session], daemonize=True)
+        daemon.start(["--name", session], prompts=[], daemonize=True)
 
-        # Wait briefly for the socket to appear
+        # Wait briefly for the daemon to start accepting connections. Checking
+        # only path existence races with bind/listen and stale socket files.
         import time
 
         for _ in range(20):
-            if sock_path.exists():
+            if daemon.is_running():
                 break
             time.sleep(0.2)
         else:
-            click.echo("Daemon did not start in time (socket not found).", err=True)
+            click.echo("Daemon did not start in time.", err=True)
             raise SystemExit(1)
 
     try:
@@ -136,7 +137,7 @@ def attach_cmd(session: str, start_if_missing: bool) -> None:
     except FileNotFoundError as e:
         click.echo(str(e), err=True)
         raise SystemExit(1) from e
-    except ConnectionRefusedError as e:
+    except (ConnectionRefusedError, ConnectionResetError) as e:
         click.echo(
             f"Could not connect to daemon '{session}'. It may have exited.", err=True
         )
--- a/gptme/cli/cmd_daemon.py
+++ b/gptme/cli/cmd_daemon.py
@@ -93,11 +93,10 @@ def start(
 @cli.command("attach")
 @click.argument("session")
 @click.option(
-    "--start-if-missing",
-    is_flag=True,
+    "--start-if-missing/--no-start-if-missing",
     default=True,
     show_default=True,
-    help="Auto-start the daemon if the session is not running (default: on)",
+    help="Auto-start the daemon if the session is not running",
 )
 def attach_cmd(session: str, start_if_missing: bool) -> None:
     """Attach to a running daemon session.
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -208,6 +208,9 @@ class SessionDaemon:
                 continue
             try:
                 self._run_turn(gptme_args, prompts)
+            except Exception as exc:
+                logger.exception("Daemon turn failed")
+                self._publish_output(b"", f"\n[daemon error] Turn failed: {exc}\n")
             finally:
                 self._turn_queue.task_done()
 
@@ -215,7 +218,10 @@ class SessionDaemon:
         # stdin must be DEVNULL.  A PIPE is non-TTY input, so gptme would wait for
         # EOF while trying to consume piped input before processing CLI prompts.
         proc = subprocess.Popen(
-            [sys.executable, "-m", "gptme", "--non-interactive"] + gptme_args + prompts,
+            [sys.executable, "-m", "gptme", "--non-interactive"]
+            + gptme_args
+            + ["--"]
+            + prompts,
             stdin=subprocess.DEVNULL,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
@@ -278,7 +284,13 @@ class SessionDaemon:
             if not rlist:
                 continue
 
-            client, _ = server_sock.accept()
+            try:
+                client, _ = server_sock.accept()
+            except OSError:
+                if self._stop_event.is_set():
+                    break
+                logger.warning("Daemon accept failed", exc_info=True)
+                continue
             client.setblocking(True)
             try:
                 self._serve_client(client)
@@ -311,9 +323,10 @@ class SessionDaemon:
                 return
             self._client = client
 
+        recv_buffer = bytearray()
         while not self._stop_event.is_set():
             try:
-                msg = recv_msg(client)
+                msg = recv_msg(client, recv_buffer)
             except TimeoutError:
                 continue
             except (OSError, ValueError):
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -19,6 +19,7 @@ later attachments continue the conversation after earlier turns have exited.
 from __future__ import annotations
 
 import codecs
+import fcntl
 import logging
 import os
 import queue
@@ -33,7 +34,7 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from pathlib import Path
-    from typing import Protocol
+    from typing import IO, Protocol
 
     class _ReadableFD(Protocol):
         def fileno(self) -> int: ...
@@ -86,6 +87,7 @@ class SessionDaemon:
         self._stop_event = threading.Event()
         self._proc_lock = threading.Lock()
         self._proc: subprocess.Popen[bytes] | None = None
+        self._pid_file: IO[str] | None = None
 
     # ------------------------------------------------------------------
     # Public API
@@ -114,28 +116,48 @@ class SessionDaemon:
         try:
             pid = int(self.pid_path.read_text())
             os.kill(pid, 0)
+            pid_file = self.pid_path.open()
         except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
             return False
-        # The MVP intentionally serves one attached client at a time. A handshake
-        # probe would therefore time out while that client is attached even though
-        # the daemon is healthy. The PID and bound socket are the daemon-owned
-        # lifecycle markers; cleanup removes both when the process exits.
-        return True
+        try:
+            fcntl.flock(pid_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError:
+            return True
+        else:
+            fcntl.flock(pid_file, fcntl.LOCK_UN)
+            return False
+        finally:
+            pid_file.close()
 
     def stop(self) -> None:
-        """Send SIGTERM to the daemon process."""
+        """Send SIGTERM only to the process that owns the daemon PID file."""
+        if not self.is_running():
+            return
         try:
             pid = int(self.pid_path.read_text())
             os.kill(pid, signal.SIGTERM)
-        except (FileNotFoundError, ProcessLookupError, ValueError):
+        except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
             pass
 
     # ------------------------------------------------------------------
     # Internal — runs in the daemon process
     # ------------------------------------------------------------------
 
+    def _acquire_pid_file(self) -> None:
+        pid_file = self.pid_path.open("a+")
+        try:
+            fcntl.flock(pid_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except OSError:
+            pid_file.close()
+            raise
+        pid_file.seek(0)
+        pid_file.truncate()
+        pid_file.write(str(os.getpid()))
+        pid_file.flush()
+        self._pid_file = pid_file
+
     def _run(self, gptme_args: list[str], prompts: list[str]) -> None:
-        self.pid_path.write_text(str(os.getpid()))
+        self._acquire_pid_file()
 
         server_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         server_sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
@@ -336,8 +358,11 @@ class SessionDaemon:
                     self._client = None
 
     def _cleanup(self) -> None:
-        for path in (self.socket_path, self.pid_path):
-            path.unlink(missing_ok=True)
+        self.socket_path.unlink(missing_ok=True)
+        if self._pid_file is not None:
+            self._pid_file.close()
+            self._pid_file = None
+            self.pid_path.unlink(missing_ok=True)
 
 
 # ---------------------------------------------------------------------------
@@ -439,12 +464,12 @@ def list_daemons() -> list[dict]:
     for pid_file in daemon_dir.glob("*.pid"):
         name = pid_file.stem
         sock = daemon_dir / f"{name}.socket"
+        daemon = SessionDaemon(name)
+        running = daemon.is_running()
         try:
             pid = int(pid_file.read_text())
-            os.kill(pid, 0)
-            running = sock.exists()
-        except (ValueError, ProcessLookupError, PermissionError):
-            running = False
+        except ValueError:
+            pid = 0
         result.append(
             {
                 "session": name,
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -617,7 +617,13 @@ def _stop_via_socket(sock_path: Path) -> None:
     try:
         conn.connect(str(sock_path))
         send_msg(conn, IPCMessage(type="signal", data={"signal": "SIGTERM"}))
-    except OSError:
+        # Keep the receive half open until the daemon consumes the signal.  A
+        # regular client slot sends its ready frame before reading the signal;
+        # closing here could make that write fail and discard the stop request.
+        conn.shutdown(socket.SHUT_WR)
+        while recv_msg(conn) is not None:
+            pass
+    except (OSError, ValueError):
         pass
     finally:
         conn.close()
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -616,14 +616,6 @@ def _stop_via_socket(sock_path: Path) -> None:
     conn.settimeout(_ATTACH_HANDSHAKE_TIMEOUT)
     try:
         conn.connect(str(sock_path))
-        # Check if ready message is available (short timeout for control connections)
-        conn.settimeout(0.1)
-        try:
-            recv_msg(conn)
-        except (OSError, ValueError, TimeoutError):
-            pass
-        # Send stop signal (extend timeout for sending)
-        conn.settimeout(_ATTACH_HANDSHAKE_TIMEOUT)
         send_msg(conn, IPCMessage(type="signal", data={"signal": "SIGTERM"}))
     except OSError:
         pass
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -616,6 +616,14 @@ def _stop_via_socket(sock_path: Path) -> None:
     conn.settimeout(_ATTACH_HANDSHAKE_TIMEOUT)
     try:
         conn.connect(str(sock_path))
+        # Check if ready message is available (short timeout for control connections)
+        conn.settimeout(0.1)
+        try:
+            recv_msg(conn)
+        except (OSError, ValueError, TimeoutError):
+            pass
+        # Send stop signal (extend timeout for sending)
+        conn.settimeout(_ATTACH_HANDSHAKE_TIMEOUT)
         send_msg(conn, IPCMessage(type="signal", data={"signal": "SIGTERM"}))
     except OSError:
         pass
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -138,12 +138,19 @@ class SessionDaemon:
             pid_file.close()
 
     def stop(self) -> None:
-        """Send SIGTERM only to the process that owns the daemon PID file."""
-        if not self.is_running():
+        """Signal the daemon while continuously proving PID-file ownership."""
+        if not self.socket_path.exists():
             return
         try:
-            pid = int(self.pid_path.read_text())
-            os.kill(pid, signal.SIGTERM)
+            with self.pid_path.open() as pid_file:
+                try:
+                    fcntl.flock(pid_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except BlockingIOError:
+                    pid_file.seek(0)
+                    pid = int(pid_file.read())
+                    os.kill(pid, signal.SIGTERM)
+                else:
+                    fcntl.flock(pid_file, fcntl.LOCK_UN)
         except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
             pass
 
@@ -475,7 +482,7 @@ def attach(session_name: str) -> None:
         while not stop_event.is_set():
             try:
                 msg = recv_msg(conn)
-            except OSError:
+            except (OSError, ValueError):
                 break
             if msg is None:
                 break
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -47,6 +47,13 @@ logger = logging.getLogger(__name__)
 
 # Output ring-buffer: send this many bytes of history to a newly attached client
 _OUTPUT_HISTORY_BYTES = 64 * 1024
+# Short polling timeout keeps disconnect and shutdown detection responsive.
+_CLIENT_READ_TIMEOUT = 0.2
+# Writes are bounded independently: large terminal output may legitimately take
+# longer than one read-poll interval, but a non-reading client cannot block a turn.
+_CLIENT_WRITE_TIMEOUT = 5.0
+# A queued second attach must fail instead of waiting forever for the only client slot.
+_ATTACH_HANDSHAKE_TIMEOUT = 2.0
 
 
 def get_daemon_dir() -> Path:
@@ -255,6 +262,7 @@ class SessionDaemon:
     def _publish_output(self, chunk: bytes, text: str | None = None) -> None:
         if text is None:
             text = chunk.decode("utf-8", errors="replace")
+        failed_client: socket.socket | None = None
         with self._output_lock:
             if text:
                 self._output_buf.append(text)
@@ -264,13 +272,14 @@ class SessionDaemon:
                     self._output_buf_size -= len(dropped.encode())
 
             if self._client is not None and text:
-                try:
-                    send_msg(
-                        self._client,
-                        IPCMessage(type="output", data=text),
-                    )
-                except OSError:
+                if not self._write_client(
+                    self._client, IPCMessage(type="output", data=text)
+                ):
+                    failed_client = self._client
                     self._client = None
+
+        if failed_client is not None:
+            self._disconnect_client(failed_client)
 
     def _accept_loop(self, server_sock: socket.socket) -> None:
         """Accept one client at a time while the daemon remains alive."""
@@ -302,24 +311,18 @@ class SessionDaemon:
 
     def _serve_client(self, client: socket.socket) -> None:
         """Replay output history, then queue prompts received from the client."""
-        # Bound every write, including the initial handshake and history replay.
-        # A client that connects without reading must not stall the daemon worker.
-        client.settimeout(0.2)
+        client.settimeout(_CLIENT_READ_TIMEOUT)
         # Publish the client only after history is fully sent.  Holding the same
         # lock used by _publish_output makes history-before-live ordering atomic.
         with self._output_lock:
             history = "".join(self._output_buf)
-            try:
-                send_msg(
-                    client,
-                    IPCMessage(type="status", data={"ready": True}),
-                )
-                if history:
-                    send_msg(
-                        client,
-                        IPCMessage(type="output", data=history),
-                    )
-            except OSError:
+            if not self._write_client(
+                client, IPCMessage(type="status", data={"ready": True})
+            ):
+                return
+            if history and not self._write_client(
+                client, IPCMessage(type="output", data=history)
+            ):
                 return
             self._client = client
 
@@ -363,12 +366,33 @@ class SessionDaemon:
                 break
 
     def _send_to_client(self, client: socket.socket, msg: IPCMessage) -> None:
+        failed = False
         with self._output_lock:
-            if self._client is client:
-                try:
-                    send_msg(client, msg)
-                except OSError:
-                    self._client = None
+            if self._client is client and not self._write_client(client, msg):
+                self._client = None
+                failed = True
+        if failed:
+            self._disconnect_client(client)
+
+    @staticmethod
+    def _write_client(client: socket.socket, msg: IPCMessage) -> bool:
+        """Send one frame with a write-specific timeout."""
+        client.settimeout(_CLIENT_WRITE_TIMEOUT)
+        try:
+            send_msg(client, msg)
+        except OSError:
+            return False
+        finally:
+            client.settimeout(_CLIENT_READ_TIMEOUT)
+        return True
+
+    @staticmethod
+    def _disconnect_client(client: socket.socket) -> None:
+        """Wake the synchronous client loop after a write failure."""
+        try:
+            client.shutdown(socket.SHUT_RDWR)
+        except OSError:
+            pass
 
     def _cleanup(self) -> None:
         self.socket_path.unlink(missing_ok=True)
@@ -419,8 +443,15 @@ def attach(session_name: str) -> None:
         raise FileNotFoundError(f"No daemon socket for session '{session_name}'")
 
     conn = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    conn.settimeout(_ATTACH_HANDSHAKE_TIMEOUT)
     conn.connect(str(sock_path))
-    ready = recv_msg(conn)
+    try:
+        ready = recv_msg(conn)
+    except TimeoutError as exc:
+        conn.close()
+        raise TimeoutError(
+            f"Daemon '{session_name}' already has an attached client"
+        ) from exc
     if (
         ready is None
         or ready.type != "status"
@@ -429,6 +460,7 @@ def attach(session_name: str) -> None:
     ):
         conn.close()
         raise ConnectionResetError(f"Daemon '{session_name}' closed during attach")
+    conn.settimeout(None)
 
     stop_event = threading.Event()
 
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -116,7 +116,11 @@ class SessionDaemon:
             os.kill(pid, 0)
         except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
             return False
-        return _socket_is_accepting(self.socket_path)
+        # The MVP intentionally serves one attached client at a time. A handshake
+        # probe would therefore time out while that client is attached even though
+        # the daemon is healthy. The PID and bound socket are the daemon-owned
+        # lifecycle markers; cleanup removes both when the process exits.
+        return True
 
     def stop(self) -> None:
         """Send SIGTERM to the daemon process."""
@@ -245,6 +249,8 @@ class SessionDaemon:
         while not self._stop_event.is_set():
             try:
                 rlist, _, _ = select.select([server_sock], [], [], 0.2)
+            except InterruptedError:
+                continue
             except OSError:
                 break
             if not rlist:
@@ -262,6 +268,9 @@ class SessionDaemon:
 
     def _serve_client(self, client: socket.socket) -> None:
         """Replay output history, then queue prompts received from the client."""
+        # Bound every write, including the initial handshake and history replay.
+        # A client that connects without reading must not stall the daemon worker.
+        client.settimeout(0.2)
         # Publish the client only after history is fully sent.  Holding the same
         # lock used by _publish_output makes history-before-live ordering atomic.
         with self._output_lock:
@@ -280,7 +289,6 @@ class SessionDaemon:
                 return
             self._client = client
 
-        client.settimeout(0.2)
         while not self._stop_event.is_set():
             try:
                 msg = recv_msg(client)
@@ -424,24 +432,6 @@ def attach(session_name: str) -> None:
 # ---------------------------------------------------------------------------
 
 
-def _socket_is_accepting(path: Path) -> bool:
-    probe = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-    probe.settimeout(0.5)
-    try:
-        probe.connect(str(path))
-        ready = recv_msg(probe)
-        return (
-            ready is not None
-            and ready.type == "status"
-            and isinstance(ready.data, dict)
-            and bool(ready.data.get("ready"))
-        )
-    except (OSError, ValueError):
-        return False
-    finally:
-        probe.close()
-
-
 def list_daemons() -> list[dict]:
     """Return info dicts for known daemons."""
     daemon_dir = get_daemon_dir()
@@ -452,7 +442,7 @@ def list_daemons() -> list[dict]:
         try:
             pid = int(pid_file.read_text())
             os.kill(pid, 0)
-            running = _socket_is_accepting(sock)
+            running = sock.exists()
         except (ValueError, ProcessLookupError, PermissionError):
             running = False
         result.append(
--- a/gptme/server/daemon.py
+++ b/gptme/server/daemon.py
@@ -148,7 +148,7 @@ class SessionDaemon:
                 except BlockingIOError:
                     pid_file.seek(0)
                     pid = int(pid_file.read())
-                    os.kill(pid, signal.SIGTERM)
+                    _signal_owned_pid(pid_file, pid)
                 else:
                     fcntl.flock(pid_file, fcntl.LOCK_UN)
         except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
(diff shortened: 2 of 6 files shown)


gptme/cli/cmd_daemon.py
```diff
@@ -0,0 +1,208 @@
...
+
+        for _ in range(20):
+            if sock_path.exists():
+                break
+            time.sleep(0.2)
+        else:
+            click.echo("Daemon did not start in time (socket not found).", err=True)
+            raise SystemExit(1)
+
+    try:
```
TimeToBuildBob, line 151: ⚠️ **P2** — The `attach` CLI function connects to the daemon socket but does not handle the case where the daemon's subprocess has already exited. The `attach` client's receive thread blocks on `recv_msg` until the daemon closes; if the daemon has died but left a stale socket file (e.g., the daemon crashed and `_cleanup` was skipped, or the subprocess terminated early), `conn.connect()` succeeds, but the daemon's main loop has already returned and the socket is closed. The client's `recv_msg` then returns `None`, the receive thread breaks, sets `stop_event`, and the main stdin loop ends, causing `attach` to exit immediately without any user-visible error. The expected behavior is a clear 'daemon is not running' error, like the `ConnectionRefusedError` handler for a missing socket, but this path silently exits. This is a correctness bug in normal use when the daemon dies between socket creation and attach.

gptme/server/daemon.py
```diff
@@ -0,0 +1,355 @@
...
+
+            # Forward to client if connected
+            with self._client_lock:
+                client = self._client
+            if client is not None:
+                try:
+                    send_msg(
+                        client,
+                        IPCMessage(
+                            type="output", data=chunk.decode("utf-8", errors="replace")
```
TimeToBuildBob: ❌ **P1** — In `SessionDaemon._pump_output`, the output thread reads from the subprocess stdout and forwards each chunk to the connected client. However, it decodes the chunk to UTF-8 with `errors='replace'` and then re-encodes it on send. This means that if a multi-byte UTF-8 character is split across two 4096-byte reads, the first read contains an incomplete byte sequence which gets decoded as replacement character ``, corrupting the output. The same corruption occurs when the ring-buffer history is replayed in `_serve_client`, where the history is joined and then decoded. This is a real correctness bug: any non-ASCII text longer than 4096 bytes (or whose boundary falls mid-character) will be mangled at the client. The IPC protocol itself is byte-safe via the length prefix, but the per-chunk decode loses data.

gptme/server/daemon.py
```diff
@@ -0,0 +1,355 @@
...
+            args=(proc.stdout,),
+            daemon=True,
+        )
+        output_thread.start()
+
+        # Main loop: accept connections and forward client input to subprocess
+        try:
+            self._accept_loop(server_sock, proc)
+        finally:
+            proc.terminate()
```
TimeToBuildBob, line 253: ❌ **P1** — `_serve_client` can be invoked by the `signal` handler from the accept loop, but the accept loop runs sequentially and blocks on `recv_msg` while a client is connected. That means the daemon cannot accept a second client until the current one disconnects, which is intentional for the single-client MVP. However, the `stop` command sends SIGTERM to the daemon process, not to the gptme subprocess. When the daemon receives SIGTERM, it does not have a signal handler, so the default behavior terminates the Python process immediately, without running `_cleanup` and without `proc.terminate()`. This leaves the gptme subprocess orphaned, and the socket and pid files are not removed. On restart, the stale socket causes `start` to report 'already running' incorrectly after the pid file's process is gone (is_running reads the pid and uses kill(pid,0) which may succeed if the PID was reused). The daemon should install a SIGTERM handler to forward the signal to the subprocess and clean up.

gptme/server/daemon.py
```diff
@@ -0,0 +1,355 @@
...
+            logger.exception("Daemon crashed")
+        finally:
+            self._cleanup()
+
+    def is_running(self) -> bool:
+        if not self.socket_path.exists():
+            return False
+        try:
+            pid = int(self.pid_path.read_text())
+            os.kill(pid, 0)  # check existence
```
TimeToBuildBob: ⚠️ **P2** — The `is_running` method and `list_daemons` both rely on `os.kill(pid, 0)` to check if a daemon is alive. This fails if the pid belongs to a different process that was reused after a daemon crash. The stale socket file then makes `is_running` return True even though no daemon is associated with it, causing `start` to refuse to start a new session, and `attach` to try connecting to a dead socket which either fails with ConnectionRefusedError or, if the socket file is left over, connects but gets EOF. The same stale-state issue affects `list_daemons` showing a running daemon that doesn't exist. The proper fix is to verify that the socket is actually listening (e.g., attempt a connect) or use the PID file's ownership (compare `/proc/<pid>/cmdline` or use `waitid` with WNOWAIT).

gptme/cli/cmd_daemon.py
```diff
@@ -0,0 +1,208 @@
...
+        if not start_if_missing:
+            click.echo(
+                f"No running daemon for session '{session}'. Start with: gptme daemon start --session {session}",
+                err=True,
+            )
+            raise SystemExit(1)
+
+        click.echo(f"Session '{session}' not running — starting it now...", err=True)
+        daemon = SessionDaemon(session)
+        daemon.start(["--name", session], daemonize=True)
```
TimeToBuildBob: ❌ **P1** — The `attach` command auto-starts a daemon with `daemon.start(["--name", session], daemonize=True)` when the socket does not exist. However, this variant passes no prompts and no `--non-interactive` flag (the daemon's `_run` adds `--non-interactive` itself, so that's covered). But the subprocess spawned by the daemon receives an empty prompt, and since gptme's non-interactive mode requires a prompt or `--resume` (see `main.py` line 1335-1345, it exits with an error "Non-interactive mode requires a prompt"), the daemon will immediately start and then the gptme subprocess will exit with an error. The daemon then detects `proc.poll() is not None`, stops, and cleans up, leaving no session. The `attach` waiting loop checks for the socket existence only, and after the socket appears and then disappears, the subsequent `attach` call raises FileNotFoundError. But the more subtle bug is that the user's intended attach to an existing session auto-starts a new empty session that immediately fails, blasting an error to the daemon log and confusing the user. Since the design intends auto-start to allow starting fresh sessions, it should probably pass a default prompt or allow attaching to a session without a prompt. This is a functional bug in the auto-start path.

gptme/server/ipc_protocol.py
```diff
@@ -0,0 +1,77 @@
...
+                "seq": self.seq,
+                "timestamp": self.timestamp,
+            }
+        ).encode()
+        return _HEADER.pack(len(payload)) + payload
+
+    @classmethod
+    def from_bytes(cls, data: bytes) -> IPCMessage:
+        d = json.loads(data)
+        return cls(
```
TimeToBuildBob, line 70: ⚠️ **P2** — The `IPCMessage.from_bytes` method does not validate that the decoded JSON is an object with a 'type' field, nor that the type is a valid MessageType. If an attacker or malfunctioning client sends a malformed message, `json.loads` could return a list or a dict without 'type', causing `d['type']` to raise KeyError, which propagates as an unhandled exception in the daemon's `_serve_client` (since it's not caught), crashing the daemon's accept loop and the entire daemon. The protocol is only used for local IPC between the user's own processes, but the socket is user-accessible and other local processes might interact. The robustness issue is that a single corrupted message kills the daemon. This is a P2, but since it's about crash resilience.

gptme/server/daemon.py
```diff
@@ -0,0 +1,355 @@
...
+                    client,
+                    IPCMessage(
+                        type="output", data=history.decode("utf-8", errors="replace")
+                    ),
+                )
+            except OSError:
+                return
+
+        # Relay client input to subprocess stdin
+        while proc.poll() is None:
```
TimeToBuildBob: ❌ **P1** — In `SessionDaemon._serve_client`, the `while proc.poll() is None:` loop reads messages from the client. If the client sends a `status` message, the daemon replies with the status. But the status reply is sent before checking `proc.poll()` again; if the subprocess exited between the check and the send, it doesn't matter. However, the real issue is that the loop blocks on `recv_msg` indefinitely. If the client disconnects, `recv_msg` returns None and the loop breaks, which is fine. But if the subprocess exits while a client is still connected, `proc.poll()` becomes non-None only after the loop iterates; if the loop is blocking on `recv_msg` waiting for a message from the client, it will never notice the subprocess exit until the client sends something or disconnects. This means the daemon does not clean up and exit when the subprocess finishes while a client is attached. For example, `gptme daemon start --session test "echo hi"` will cause the subprocess to output and then exit. If a client attaches early, the daemon's accept loop accepts, sends history, then enters `_serve_client` and blocks on `recv_msg`. The subprocess exits, but the daemon stays alive because the loop is stuck on `recv_msg`. Only when the client sends a message or disconnects will the loop break and the daemon cleanup. This is a correctness bug: the daemon doesn't terminate when its subprocess is done, leaving a zombie daemon and stale socket. The fix is to use `select` or a timeout on `recv_msg` to also check `proc.poll()`, or to have the output thread signal the main loop when the subprocess exits.

gptme/server/daemon.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Daemonization (double-fork on POSIX)
+# ---------------------------------------------------------------------------
+
+
+def _daemonize() -> None:
+    """Double-fork to create a proper daemon process."""
+    if os.fork() > 0:
```
TimeToBuildBob: ⚠️ **P2** — The `_daemonize` function calls `os.fork()` without checking for errors. If `fork()` fails (e.g., process limit, memory pressure), it returns -1 and the code treats it as a child (because the condition `if os.fork() > 0:` is false), so the parent continues as the daemon process without double-forking properly. This can lead to the daemon not being fully detached, and the parent might not exit as expected. Unlikely in practice, but it's a robustness issue.

TimeToBuildBob:
## 🤖 AI code review

Adds a persistent background session daemon for gptme: a new `gptme daemon` CLI group (start/attach/list/stop/status), a Unix-domain-socket IPC protocol with framed JSON messages, a SessionDaemon class that daemonizes and manages per-session subprocess turns, and a test suite covering protocol, daemon state, attach behavior, and CLI smoke tests. Also registers a `gptme-daemon` console script and documents the new subcommands in main help.

<details><summary><b>2 findings</b> · ⚠️ **2** P2 — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/server/daemon.py:590`

The attach client does not handle `ValueError` from `send_msg` when an input line exceeds the IPC payload cap. `IPCMessage.encode` raises `ValueError("IPC payload too large")` if the JSON payload is over `MAX_IPC_PAYLOAD` (1 MiB). In `attach`, the line is sent via `send_msg(conn, IPCMessage(type="input", data=line, seq=seq))` inside a `try` that only catches `OSError`. If a user pastes a single line (or a file redirected to stdin) larger than ~1 MiB, this `ValueError` propagates out of `attach`, producing a traceback and exiting the CLI. The same issue exists on the daemon side: `_serve_client` catches `(OSError, ValueError)` when reading, so an oversized input from a client would disconnect that client rather than crashing the daemon, but the CLI crash is user-visible and unexpected. This is a robustness failure that can be triggered by ordinary (if extreme) terminal input.

```suggestion
Catch ValueError around the send_msg in attach and print a warning, or split/truncate lines that exceed the cap before sending.
```

<sub>**How this was verified:** Traced the flow: daemon `_serve_client` uses `select` with `server_sock`, so a second connection is handled by `_handle_control_connection`, which after a 1s timeout closes the extra socket; client `recv_msg` gets EOF and takes the `ready is None` branch.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/server/daemon.py:150`

`SessionDaemon.stop()` begins with `if not self.socket_path.exists(): return`, so when the socket file is absent (e.g., deleted by the user or a failed cleanup after a SIGKILL) the method returns without attempting to signal the daemon process via pidfd. As a result, a live daemon whose socket is gone cannot be stopped; `is_running()` also returns False because it gates on `socket_path.exists()`, so the CLI reports the session as not running and `gptme daemon stop` exits with "not running". The only recovery is a manual `kill` of the PID, which the code intentionally avoids. The fix is to check the PID file lock even when the socket is missing, using pidfd (or the pid file's contents) to signal the process.

<sub>**How this was verified:** Read `is_running` (lines 125–143): it requires both the socket and the PID file to deem a daemon running. The CLI `stop` and `status` commands call `is_running()` first and treat `False` as 'not running'.</sub>

</details>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/server/daemon.py:1`

This appears to overlap with `gptme/gptme-contrib:packages/gptme-rag` which already provides RAG: document indexing, semantic/lexical search, persistence. Matched signals: chunk, jsonl. Informational only — not a blocker. A human decides whether to redirect, extend the existing module, or intentionally parallel-implement.

</details>

- `gptme/cli/cmd_daemon.py` — New Click group with start, attach, list, stop, status commands; validates session names, auto-starts on attach, and delegates to SessionDaemon.
- `gptme/cli/main.py` — Adds three lines to the help docstring documenting the new daemon subcommands.
- `gptme/server/daemon.py` — New SessionDaemon class with double-fork daemonize, pid-file locking, socket accept loop, output ring-buffer, worker queue, stop/signal logic, and client attach function.
- `gptme/server/ipc_protocol.py` — New IPC message codec: IPCMessage dataclass with JSON encoding, 4-byte length prefix, and bounded payload size.
- `pyproject.toml` — Registers gptme-daemon as a console script entry point.
- `tests/test_daemon.py` — New test suite covering IPC roundtrip, path helpers, daemon state, turn lifecycle, attach protocol, list_daemons, and CLI validation.

</details>

| `ed61d7445249` | 2/5 | 8 | llm | 2026-08-27 10:42 UTC |
| `67d9d446a14e` | 2/5 | 4 | llm | 2026-08-27 15:28 UTC |
| `e98ae32c82e8` | 2/5 | 6 | llm | 2026-08-27 17:33 UTC |
| `10eafaf7acf5` | 2/5 | 3 | llm | 2026-08-27 19:04 UTC |
| `302c7c730135` | 4/5 | 2 | llm | 2026-08-28 03:53 UTC |
| `5e4e7096ba47` | 3/5 | 4 | llm | 2026-08-29 02:12 UTC |

</details>

<sub>Reviewed `9c7538dd4e36` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 796s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `67d9d446a` addressing the daemon review findings:

- keep the daemon alive across non-interactive turns and queue prompts from later attachments
- validate session names before deriving socket/PID paths
- serialize history/live frame writes and use unbuffered incremental UTF-8 output decoding
- install clean stop handling, verify daemon liveness with an IPC handshake, validate malformed IPC, and report failed attach handshakes

Verification: 32 daemon tests pass, targeted pre-commit (ruff/format/mypy) passes, and a start → prompt → detach → reattach → follow-up lifecycle smoke passes. The required in-band review wrapper ran but its local reviewer skipped because `bobutils` was unavailable; it still pushed as designed. Greptile re-review is now requested; CI is pending.

TimeToBuildBob:
✅ **PR ready for human merge**

All gates cleared on head `67d9d446a14e`:
- CI: all checks green
- Greptile: 5/5 — "The PR appears safe to merge. No blocking failure remains."
- All review threads resolved (4 stale Greptile P1 + 8 stale Bob AI review from older commit)

This touches `gptme/server/daemon.py` so it requires human merge — not in the self-merge allowlist.

TimeToBuildBob:
Automated merge handoff for `67d9d446a14e`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/cmd_daemon.py`
- `gptme/cli/main.py`
- `gptme/server/daemon.py`
- `gptme/server/ipc_protocol.py`
- `pyproject.toml`

gptme/server/daemon.py
```diff
@@ -0,0 +1,466 @@
...
+# List running daemons
+# ---------------------------------------------------------------------------
+
+
+def _socket_is_accepting(path: Path) -> bool:
+    probe = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    probe.settimeout(0.5)
+    try:
+        probe.connect(str(path))
+        ready = recv_msg(probe)
```
TimeToBuildBob: ❌ **P1** — The `is_running` check (and by extension `list_daemons` and the `status` command) reports a running daemon as stopped when a client is attached. `_socket_is_accepting` connects to the daemon's socket and waits for a status message via `recv_msg(probe)`, but the daemon's single-client accept loop is busy in `_serve_client` and never calls `accept()` on the server socket while a client is attached. The probe connection is queued in the listen backlog, `recv_msg` times out after 0.5s, and the function returns False. Consequence: `gptme daemon list` shows the session as stopped, `gptme daemon status` misreports it, and any monitoring tool that relies on `is_running` will incorrectly conclude the daemon is dead. Fix: have the daemon handle probe connections in a way that doesn't block on the active client, e.g., by accepting probe connections inline and replying without keeping the main client waiting, or by checking process liveness and socket file existence as a fallback.

gptme/cli/cmd_daemon.py
```diff
@@ -0,0 +1,209 @@
...
+    click.echo(f"Daemon started. Attach with: gptme daemon attach {session}")
+    click.echo(f"Socket: {sock_path}")
+
+
+@cli.command("attach")
+@click.argument("session")
+@click.option(
+    "--start-if-missing",
+    is_flag=True,
+    default=True,
```
TimeToBuildBob, line 114: ⚠️ **P2** — The `--start-if-missing` option is defined with `is_flag=True, default=True`, which means the flag is always True and there is no way to disable auto-start on attach. Click's `is_flag=True` produces a boolean that is False by default and turned on by the flag; setting `default=True` makes it always True unless a `--no-start-if-missing` counterpart is defined. The help text says "default: on", implying it can be turned off, but users cannot do so. Consequence: `gptme daemon attach some-name` always auto-starts a daemon if the socket file doesn't exist, even when the user only wants to check whether a session is running or attach to an existing one. Fix: use `@click.option("--start-if-missing/--no-start-if-missing", default=True)`.

gptme/server/daemon.py
```diff
@@ -0,0 +1,466 @@
...
+            if text:
+                self._output_buf.append(text)
+                self._output_buf_size += len(text.encode())
+                while self._output_buf_size > _OUTPUT_HISTORY_BYTES:
+                    dropped = self._output_buf.popleft()
+                    self._output_buf_size -= len(dropped.encode())
+
+            if self._client is not None and text:
+                try:
+                    send_msg(
```
TimeToBuildBob: ❌ **P1** — In `_publish_output`, the daemon sends output to the attached client with `send_msg` while holding the output lock, but the client socket is never given a send timeout. If the client process is alive but not reading (e.g., the terminal is suspended, or the user has detached but the OS hasn't closed the socket yet), `socket.sendall` blocks indefinitely once the socket buffer fills. This stalls the worker thread inside `_pump_output`/`_run_turn`, preventing any queued turns from executing and making the daemon unresponsive to later attach attempts. Consequence: a single unresponsive client hangs the entire daemon until the client process terminates or resumes reading. Fix: set a timeout on the client socket for writes (e.g., `client.settimeout(5)` before sending) and handle the resulting `TimeoutError`, or use non-blocking sends with error handling.

gptme/server/daemon.py
```diff
@@ -0,0 +1,466 @@
...
+                    )
+                except OSError:
+                    self._client = None
+
+    def _accept_loop(self, server_sock: socket.socket) -> None:
+        """Accept one client at a time while the daemon remains alive."""
+        while not self._stop_event.is_set():
+            try:
+                rlist, _, _ = select.select([server_sock], [], [], 0.2)
+            except OSError:
```
TimeToBuildBob, line 337: ⚠️ **P2** — In `_accept_loop`, the `try` block around `select.select([server_sock], [], [], 0.2)` catches `OSError` and breaks. When any signal is delivered (e.g., SIGWINCH from a terminal resize, SIGCHLD from a child process, or SIGALRM from a timer), `select` raises `InterruptedError`, which is a subclass of `OSError`. This causes the accept loop to break and the daemon to exit, even though the signal may be unrelated to stopping the daemon. Consequence: an unrelated signal such as SIGWINCH can silently kill a running daemon, losing its socket and pid files. Fix: catch `InterruptedError` explicitly and continue the loop, or check the stop event after an interrupt before breaking.

TimeToBuildBob:
Pushed `e98ae32c8` addressing the follow-up AI review:

- report daemon liveness from its PID + owned socket even while the single client slot is occupied
- bound all client writes before handshake/history replay
- retry `select()` after unrelated signal interruption
- add `--no-start-if-missing`

Verification: all 36 daemon tests pass; targeted ruff, formatting, mypy, and pre-commit pass. The prior monitoring retry failed because it ran tests from Bob's unrelated shared virtualenv, then mutated that environment while trying to recover; this retry used the gptme checkout's intact test environment. CI and automated re-review are now pending on the new head.

TimeToBuildBob:
Automated merge handoff for `e98ae32c82e8`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/cmd_daemon.py`
- `gptme/cli/main.py`
- `gptme/server/daemon.py`
- `gptme/server/ipc_protocol.py`
- `pyproject.toml`

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+            try:
+                self._run_turn(gptme_args, prompts)
+            finally:
+                self._turn_queue.task_done()
+
+    def _run_turn(self, gptme_args: list[str], prompts: list[str]) -> None:
+        # stdin must be DEVNULL.  A PIPE is non-TTY input, so gptme would wait for
+        # EOF while trying to consume piped input before processing CLI prompts.
+        proc = subprocess.Popen(
+            [sys.executable, "-m", "gptme", "--non-interactive"] + gptme_args + prompts,
```
TimeToBuildBob: ❌ **P1** — In `_run_turn`, user-supplied prompts are appended directly to the command line without a `--` separator. If a prompt begins with `-` (e.g. `-v` or `--help`), gptme's Click parser treats it as an option rather than a positional argument, so the prompt is not delivered to the conversation and the subprocess executes with unintended flags (e.g. toggling verbose, showing help, or erroring on an unknown option). This breaks the daemon's core promise of processing arbitrary user prompts, and is exposed through both `start` and `attach`. A prompt like `"Implement -v"` would silently activate verbose mode instead of sending the literal text. Insert `--` before the prompts so Click treats everything after as positional. _(1/2 passes — pass 2 looked and disagreed)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+                with self._output_lock:
+                    if self._client is client:
+                        self._client = None
+                client.close()
+
+    def _serve_client(self, client: socket.socket) -> None:
+        """Replay output history, then queue prompts received from the client."""
+        # Bound every write, including the initial handshake and history replay.
+        # A client that connects without reading must not stall the daemon worker.
+        client.settimeout(0.2)
```
TimeToBuildBob: ❌ **P1** — In `_serve_client`, the client socket has `settimeout(0.2)` as a guard against a non-reading client blocking the daemon. However, this timeout applies to reads as well. When a client sends a message that is split across TCP segments such that a segment arrives after the 0.2s timeout, `recv` raises `TimeoutError` and the partial bytes already read by `_recv_exact` are lost. The next `recv_msg` call then starts reading a fresh header from a corrupted stream, producing a `ValueError` or a garbage frame and disconnecting the client. Large prompts (e.g. a multi-KB line pasted into attach) are especially likely to exceed the timeout window between segments. The protocol has no partial-frame buffering across read attempts. Fix by using a non-blocking/select-based read loop that accumulates bytes across timeouts, or by separating the read timeout from the write timeout. _(1/2 passes — pass 2 looked and disagreed)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def _worker_loop(self, gptme_args: list[str]) -> None:
+        """Run queued prompts as sequential turns in the named conversation."""
+        while not self._stop_event.is_set():
+            try:
+                prompts = self._turn_queue.get(timeout=0.2)
+            except queue.Empty:
+                continue
+            try:
+                self._run_turn(gptme_args, prompts)
```
TimeToBuildBob, line 263: ❌ **P1** — The worker thread in `_worker_loop` does not catch exceptions from `_run_turn`. If `subprocess.Popen` fails (e.g., `gptme` module missing, invalid arguments, or any other unexpected error), the exception propagates out of the `while` loop, terminating the worker thread. The daemon process itself continues to run and accept socket connections, but the `_turn_queue` is no longer drained, so any input sent by an attached client is added to the queue and never processed. The client sees no output and the daemon appears hung. This is a real defect because any turn that raises an exception permanently disables the daemon. _(1/2 passes — pass 1 looked and disagreed)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+            try:
+                rlist, _, _ = select.select([server_sock], [], [], 0.2)
+            except InterruptedError:
+                continue
+            except OSError:
+                break
+            if not rlist:
+                continue
+
+            client, _ = server_sock.accept()
```
TimeToBuildBob: ⚠️ **P2** — `_accept_loop` calls `server_sock.accept()` without a try-except. While `select` is used to wait for readability, a call to `accept()` can still raise an `OSError` (e.g., `EMFILE` when file descriptors are exhausted, or `ECONNABORTED` when a connection is reset between `select` and `accept`). Since `_accept_loop` runs in the main daemon thread and its only exception handling is around `select`, any such `OSError` propagates, causing the daemon to exit entirely. This is a robustness bug: the daemon should continue serving after a transient accept failure. _(1/2 passes — pass 1 looked and disagreed)_

gptme/cli/cmd_daemon.py
```diff
@@ -0,0 +1,208 @@
...
+        click.echo(f"Session '{session}' not running — starting it now...", err=True)
+        daemon = SessionDaemon(session)
+        daemon.start(["--name", session], prompts=[], daemonize=True)
+
+        # Wait briefly for the daemon to start accepting connections. Checking
+        # only path existence races with bind/listen and stale socket files.
+        import time
+
+        for _ in range(20):
+            if daemon.is_running():
```
TimeToBuildBob, line 144: ⚠️ **P2** — In `attach_cmd`, after auto-starting a daemon, the wait loop checks `daemon.is_running()` which only verifies that the socket file exists and the PID is alive. It does not check that the daemon has actually called `listen()` on the socket. If the wait sees the socket path created by `bind()` but the daemon hasn't yet executed `listen()`, `attach()` will attempt to connect and get `ConnectionRefusedError`, causing the command to exit with a misleading "It may have exited" error. The wait loop can return before the daemon is ready, because `bind()` creates the socket file before `listen()` is called in `_run`. _(1/2 passes — pass 1 looked and disagreed)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+            logger.exception("Daemon crashed")
+        finally:
+            self._cleanup()
+
+    def is_running(self) -> bool:
+        if not self.socket_path.exists():
+            return False
+        try:
+            pid = int(self.pid_path.read_text())
+            os.kill(pid, 0)
```
TimeToBuildBob, line 130: ⚠️ **P2** — `SessionDaemon.is_running` (and `list_daemons`) determine liveness by reading the PID file and calling `os.kill(pid, 0)`. If the daemon process has exited but another unrelated process has been assigned the same PID, `os.kill` succeeds and the socket file may still exist, making the daemon appear to be running. This can cause `gptme daemon start` to refuse to start a new daemon for a session that actually has no running daemon (or `stop` to signal the wrong process). The check does not verify that the PID actually belongs to a gptme daemon process, e.g., by checking process name or the socket's inode. _(1/2 passes — pass 1 looked and disagreed)_

gptme/server/daemon.py
TimeToBuildBob: Rejected as a false positive. The top-level Click command declares `prompts` with `nargs=-1` and `ignore_unknown_options=True`; without a separator, known options can still be parsed. A direct check shows `gptme --non-interactive --name x --help` exits via help, while `gptme --non-interactive --name x -- --help` reaches the conversation as the literal user prompt `--help`. The proposed `--` separator is therefore already the correct fix direction, but this finding is against the pre-fix head and will be handled in the next correctness pass rather than folded into the stale-PID adjudication commit.

gptme/cli/cmd_daemon.py
TimeToBuildBob: Rejected as a false positive on the stated ordering. `_run()` executes `bind()`, `chmod()`, `listen()`, and `setblocking(False)` synchronously with no await, thread switch, or blocking call between them; the child cannot run `is_running()` in the parent process during that span. The PID-file ownership lock added in `ad21d42d9` also makes `is_running()` an identity check rather than a bare path check. A connect retry could improve portability, but this alleged bind-before-listen race is not reproducible from the current control flow.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `ad21d42d9`. The daemon now holds a non-blocking advisory lock on the open PID file for its lifetime. `is_running()` requires that lock to be held by another process, `stop()` checks ownership before signaling, and `list_daemons()` uses the same predicate. Tests cover a stale PID containing a currently live, unrelated PID and verify that `stop()` sends no signal.

gptme/server/daemon.py
```diff
@@ -0,0 +1,456 @@
...
+        # the daemon is healthy. The PID and bound socket are the daemon-owned
+        # lifecycle markers; cleanup removes both when the process exits.
+        return True
+
+    def stop(self) -> None:
+        """Send SIGTERM to the daemon process."""
+        try:
+            pid = int(self.pid_path.read_text())
+            os.kill(pid, signal.SIGTERM)
+        except (FileNotFoundError, ProcessLookupError, ValueError):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `ad21d42d9`. The daemon now holds a non-blocking advisory lock on its open PID file for its full lifetime. `is_running()` requires that lock to be held by another process, and `stop()` refuses to signal unless ownership is proven. Regression tests cover a stale file containing a currently live, unrelated PID and assert that no SIGTERM is sent.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `10eafaf7a`. Timed reads now pass a persistent frame buffer into `recv_msg()`, so bytes received before `TimeoutError` survive the next call. Tests split a frame across a timeout and cover EOF during both header and payload.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `10eafaf7a`. `_worker_loop()` now catches per-turn exceptions, publishes a visible daemon error, marks the queue item done, and continues to process later turns. The regression test makes the first turn fail and verifies the second still runs.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `10eafaf7a`. `_accept_loop()` now handles `OSError` from `accept()`: it exits when stopping and logs/retries otherwise. A regression test simulates one failed accept followed by another loop iteration.

TimeToBuildBob:
## Convergence adjudication — maintainer judgment required

**Fixed**
- Greptile P1 “Stale PID kills process”: `ad21d42d9` makes the daemon hold an advisory lock on its open PID file for its lifetime; `is_running()`, `stop()`, and `list_daemons()` now require proven ownership. Regression coverage verifies that a stale file containing a live unrelated PID cannot receive SIGTERM.
- Three independently reported correctness findings: `10eafaf7a` preserves partial IPC frames across read timeouts, keeps the worker alive after a failed turn, and retries transient `accept()` failures. It also inserts the required `--` separator before literal prompts.

**Remaining**
- None classified as blocking or security issues. Two AI-review findings were dismissed with checkable reasons: the prompt-separator report was valid but already fixed in `10eafaf7a`; the alleged bind-before-listen race has no scheduling point between those calls and the new PID-file lock is acquired before socket creation.
- All Greptile and Bob AI-review threads are now resolved. I did **not** trigger another Greptile review because this PR hit the convergence cap.

**CI**
- Prior head `ad21d42d9` completed its available checks green. Current head `10eafaf7a` is running CI; local verification is `43 passed` in `tests/test_daemon.py`, plus targeted ruff, format, mypy/pre-commit.

**Domain risk**
- This is new POSIX-only daemon lifecycle and Unix-socket code. Before merge, a maintainer should manually smoke start → attach → detach → reattach → stop on Linux/macOS, including a prompt beginning with `--`, and confirm Windows remains gated/unsupported as intended.

**Convergence**
- Detector now reports `verdict: converged` with zero unresolved Greptile P1/P2 findings. Its `round_convergence` history still quotes `stable_rounds: 0` (required: 2) because the last capped Greptile round introduced one new P1; adjudication resolved it without another review round.

Merge-ready here means ready for maintainer judgment after current CI completes — no auto-merge performed.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,494 @@
...
+                with self._output_lock:
+                    if self._client is client:
+                        self._client = None
+                client.close()
+
+    def _serve_client(self, client: socket.socket) -> None:
+        """Replay output history, then queue prompts received from the client."""
+        # Bound every write, including the initial handshake and history replay.
+        # A client that connects without reading must not stall the daemon worker.
+        client.settimeout(0.2)
```
TimeToBuildBob: ❌ **P1** — In `_serve_client`, the client socket is set to a 0.2-second timeout with `client.settimeout(0.2)`. This timeout is applied to every operation on the socket, including the `sendall` calls made from `_publish_output` and `_send_to_client`. If a client does not read output quickly enough—for example, a terminal that is rendering a large block and delays reading, or a client that briefly pauses—`sendall` will raise `TimeoutError` (a subclass of `OSError`). In `_publish_output` this is caught and triggers `self._client = None`, disconnecting a client that is still alive and actively attached. The daemon then closes that client and the session appears dropped, even though the turn is still running. The intent was to bound writes so a non-reading client cannot stall the worker, but 0.2s is far too short for large outputs and causes spurious disconnects in normal use. The fix is to either use a much larger write timeout or to apply the timeout only to reads (e.g., by temporarily removing the timeout during writes). _(1/2 passes — pass 2 looked and disagreed)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,494 @@
...
+
+def attach(session_name: str) -> None:
+    """Connect to a running daemon and relay stdin/stdout."""
+    sock_path = get_socket_path(session_name)
+    if not sock_path.exists():
+        raise FileNotFoundError(f"No daemon socket for session '{session_name}'")
+
+    conn = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    conn.connect(str(sock_path))
+    ready = recv_msg(conn)
```
TimeToBuildBob: ❌ **P1** — The daemon accepts and serves exactly one client at a time. While `_serve_client` is handling a connection, the accept loop is blocked, so any additional connection sits in the listening backlog. A second `gptme daemon attach` connects successfully (the backlog accepts it) but then calls `recv_msg(conn)` with no timeout, waiting for the initial `status` message that the daemon will never send until it finishes serving the first client. The result is that the second attach hangs indefinitely with no error message, leaving the user with a stuck terminal. Because the protocol is single-client by design, the server should reject a second connection (e.g., by sending a status with `ready: False` and closing) or the client should set a handshake timeout so it fails fast. _(1/2 passes — pass 1 looked and disagreed)_

gptme/cli/cmd_daemon.py
```diff
@@ -0,0 +1,208 @@
...
+        gptme daemon attach mywork
+    """
+    from ..server.daemon import SessionDaemon, get_socket_path
+
+    if session is None:
+        from datetime import datetime, timezone
+
+        session = datetime.now(timezone.utc).strftime("daemon-%Y%m%dT%H%M%S")
+
+    sock_path = get_socket_path(session)
```
TimeToBuildBob, line 82: ⚠️ **P2** — The CLI commands `start`, `attach`, `stop`, and `status` call `get_socket_path(session)` or `SessionDaemon(session)` without catching the `ValueError` raised by `validate_conversation_id` when the session name contains a path separator (e.g. `gptme daemon attach foo/bar`) or other invalid characters. For example, line 65 `sock_path = get_socket_path(session)` will throw before Click's error handling runs, so the user sees a raw traceback from `validate_conversation_id` instead of a clean usage error. This affects every command that accepts a session name and breaks scripted use where a session name might be constructed from user input. The fix is to wrap the session-name validation and raise `click.UsageError` (or use a Click param type that validates). _(2/2 passes)_

gptme/server/daemon.py
```diff
@@ -0,0 +1,494 @@
...
+                    self._output_buf_size -= len(dropped.encode())
+
+            if self._client is not None and text:
+                try:
+                    send_msg(
+                        self._client,
+                        IPCMessage(type="output", data=text),
+                    )
+                except OSError:
+                    self._client = None
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `d527f95fda9d3845a8164bfa9d31b378dc646ac5`. A failed live write now shuts down the client socket after clearing the active reference, which wakes the synchronous receive loop and returns control to accept(). The regression test forces a live-output timeout and verifies the serving thread exits, allowing reattachment.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `d527f95fda9d3845a8164bfa9d31b378dc646ac5`. Reads keep the 0.2s polling timeout, while every handshake, history, status, and live-output write temporarily uses an independent 5s bound. Tests assert the timeout transition for both handshake and live frames.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `d527f95fda9d3845a8164bfa9d31b378dc646ac5`. Attach now bounds the initial handshake to 2s and reports that the single-client slot is occupied instead of blocking indefinitely in the listen backlog. A regression test exercises a connected server that never sends the handshake.

gptme/cli/cmd_daemon.py
TimeToBuildBob: Fixed in `d527f95fda9d3845a8164bfa9d31b378dc646ac5`. All four session-taking commands now use a Click ParamType backed by `conversation_id_error()`, producing exit 2 and a clean Invalid value message for unsafe names. Parameterized CLI coverage checks start, attach, stop, and status.

gptme/server/daemon.py
TimeToBuildBob: Fixed in d527f95fda9d3845a8164bfa9d31b378dc646ac5; see the preceding thread reply for the code path and regression test.

gptme/server/daemon.py
TimeToBuildBob: Fixed in d527f95fda9d3845a8164bfa9d31b378dc646ac5; see the preceding thread reply for the code path and regression test.

gptme/cli/cmd_daemon.py
TimeToBuildBob: Fixed in d527f95fda9d3845a8164bfa9d31b378dc646ac5; see the preceding thread reply for the code path and regression test.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,526 @@
...
+            client.shutdown(socket.SHUT_RDWR)
+        except OSError:
+            pass
+
+    def _cleanup(self) -> None:
+        self.socket_path.unlink(missing_ok=True)
+        if self._pid_file is not None:
+            self._pid_file.close()
+            self._pid_file = None
+            self.pid_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `6d6975db00c75f354268a679ee78fa6a8c3c6376`. Cleanup is now ownership-gated: only a process that successfully acquired and retained the session PID-file lock may unlink the shared PID/socket paths. The regression test starts a competing daemon while the owner lock is held and verifies the winner remains discoverable and running.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,533 @@
...
+            pid_file.close()
+
+    def stop(self) -> None:
+        """Send SIGTERM only to the process that owns the daemon PID file."""
+        if not self.is_running():
+            return
+        try:
+            pid = int(self.pid_path.read_text())
+            os.kill(pid, signal.SIGTERM)
+        except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `302c7c730135e8ff00d6e88a99f6c86ee2b718e9`. `stop()` now opens the PID file once, attempts the ownership lock, reads that same open inode, and signals while the daemon's conflicting lock still proves ownership; it never performs a separate post-check PID read. Regression coverage verifies the lock remains contended at the exact `os.kill()` call.

gptme/server/daemon.py
```diff
@@ -0,0 +1,533 @@
...
+    conn.settimeout(None)
+
+    stop_event = threading.Event()
+
+    def _receive() -> None:
+        while not stop_event.is_set():
+            try:
+                msg = recv_msg(conn)
+            except OSError:
+                break
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `302c7c730135e8ff00d6e88a99f6c86ee2b718e9`. The attach receiver now treats malformed/truncated frames as disconnects alongside socket errors, always reaches `stop_event.set()`, and releases the foreground stdin loop. A regression server sends a valid ready frame followed by six bytes of a truncated output frame and verifies attach returns cleanly.

TimeToBuildBob:
@greptileai review

gptme/server/ipc_protocol.py
```diff
@@ -0,0 +1,121 @@
...
+
+    while len(buffer) < _HEADER.size:
+        chunk = sock.recv(_HEADER.size - len(buffer))
+        if not chunk:
+            if buffer:
+                raise ValueError("incomplete IPC message header")
+            return None
+        buffer.extend(chunk)
+    (length,) = _HEADER.unpack(buffer[: _HEADER.size])
+    frame_size = _HEADER.size + length
```
TimeToBuildBob, line 107: ⚠️ **P2** — recv_msg in ipc_protocol.py trusts the 4-byte length prefix from the peer with no upper bound. An attacker who can connect to the daemon's socket (any process running as the same user, since the socket is chmod 0600) can send a length prefix of 0xFFFFFFFF and then feed data slowly. The buffer path reads incrementally, so the bytearray grows to that declared size, and the daemon's _serve_client loop will attempt to read four gigabytes. Even without sending that much data, the daemon's memory balloons and the read loop is tied up for the duration, so the single-client slot is monopolized and the daemon becomes unresponsive to other sessions. This is a local denial-of-service against the user's own daemon, but it does not require any misconfiguration.

gptme/server/daemon.py
```diff
@@ -0,0 +1,540 @@
...
+            elif msg.type == "status":
+                print(f"[daemon status] {msg.data}", flush=True)
+        stop_event.set()
+
+    recv_thread = threading.Thread(target=_receive, daemon=True)
+    recv_thread.start()
+
+    seq = 0
+    try:
+        for line in sys.stdin:
```
TimeToBuildBob: ⚠️ **P2** — In the attach() function, the main thread iterates `for line in sys.stdin`, while a daemon receiver thread watches the socket. When the daemon exits (e.g. `gptme daemon stop` or a crash), the socket closes, recv_msg returns None, the receiver thread sets stop_event and returns. But the main thread is blocked inside `for line in sys.stdin`, which never checks stop_event until it has read another line. With no further stdin data, the attach command hangs forever, leaving a zombie process that will not exit until the user sends a line or Ctrl-D. A user who attached, then stops the daemon from another terminal, is left with an unresponsive CLI.

gptme/server/daemon.py
```diff
@@ -0,0 +1,540 @@
...
+    def _cleanup(self) -> None:
+        # A child that lost PID-file lock acquisition must not remove paths owned
+        # by the competing daemon that won the same-session startup race.
+        if not self._owns_paths:
+            return
+        self.socket_path.unlink(missing_ok=True)
+        if self._pid_file is not None:
+            self._pid_file.close()
+            self._pid_file = None
+            self.pid_path.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `2669752e3`. Cleanup now unlinks the PID path while the owner still holds the lock on that inode, so a successor creates a new path instead of inheriting one that the old process is about to delete. A regression test starts a successor during the owner's close and verifies its PID file and liveness survive.

gptme/server/daemon.py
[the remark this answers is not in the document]
TimeToBuildBob: Follow-up in `2669752e3`. `stop()` now pins the process with a pidfd (stdlib when present, Linux syscall otherwise) and re-proves that the same PID-file inode is still lock-owned — and that the PID bytes have not changed — before signaling. If the owner exits after the PID is read, the signal is aborted. Tests cover the pidfd path, the abort-after-exit path, and the lock-held `os.kill` fallback.

gptme/server/ipc_protocol.py
TimeToBuildBob: Fixed in `2669752e3`. `recv_msg` now rejects a length prefix above 1 MiB on both the buffered and unbuffered paths before allocating, and `encode()` applies the same cap. Parameterized tests cover both receive paths.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `2669752e3`. Attach now polls stdin with `select` and checks `stop_event` on each timeout, so a daemon disconnect unblocks the CLI without waiting for another line. A regression test uses a never-written pipe as stdin and asserts attach returns after the server closes.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,636 @@
...
+        if not _pid_file_lock_held(pid_file):
+            return
+        pid_file.seek(0)
+        current = int(pid_file.read() or "0")
+        if current != pid:
+            return
+        if pidfd is not None:
+            _pidfd_send_signal(pidfd, signal.SIGTERM)
+        else:
+            os.kill(pid, signal.SIGTERM)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `af9d0f83c`. `stop()` no longer `os.kill`s an unpinned numeric PID. If pidfd cannot pin the process, it sends a SIGTERM frame on the session socket; while a client occupies the slot, `_serve_client` accepts that as a one-shot control connection. Tests assert the socket path is used and that `os.kill` is never called on the fallback.

gptme/server/daemon.py
```diff
@@ -0,0 +1,636 @@
...
+            pass
+
+    # ------------------------------------------------------------------
+    # Internal — runs in the daemon process
+    # ------------------------------------------------------------------
+
+    def _acquire_pid_file(self) -> None:
+        pid_file = self.pid_path.open("a+")
+        try:
+            fcntl.flock(pid_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `af9d0f83c`. After acquiring the PID-file lock, `_acquire_pid_file` checks that the locked fd still names the path (dev/ino). If the previous owner unlinked that inode, acquire closes the stale fd and retries on a live path, so a successor cannot keep an unlinked PID file. A regression test unlinks during flock and verifies the successor's fd matches the on-disk path.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,716 @@
...
+    def _serve_client(
+        self,
+        client: socket.socket,
+        server_sock: socket.socket | None = None,
+    ) -> None:
+        """Replay output history, then queue prompts received from the client."""
+        client.settimeout(_CLIENT_READ_TIMEOUT)
+        # Publish the client only after history is fully sent.  Holding the same
+        # lock used by _publish_output makes history-before-live ordering atomic.
+        with self._output_lock:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `4c95a20ea`, then hardened in `aecd9d21d`. The stop client sends SIGTERM immediately, half-closes only its write side, and keeps reading until the daemon acknowledges or closes. This prevents both the original discard and a new race where an ordinary client-slot handler could fail its ready write if the control client closed too early. The no-pidfd regression test models the one-shot signal/ack protocol; all 62 daemon tests pass.

gptme/server/ipc_protocol.py
TimeToBuildBob: Fixed in `2669752e3`: receive paths now reject payloads above 1 MiB before allocation, and encoding applies the same limit. Parameterized tests cover buffered and unbuffered receives.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `2669752e3`: attach polls stdin and checks the receiver stop event, so daemon disconnect exits without waiting for another input line. A regression test covers an idle pipe on stdin.

TimeToBuildBob:
@greptileai review

gptme/server/daemon.py
```diff
@@ -0,0 +1,722 @@
...
+        conn.connect(str(sock_path))
+        send_msg(conn, IPCMessage(type="signal", data={"signal": "SIGTERM"}))
+        # Keep the receive half open until the daemon consumes the signal.  A
+        # regular client slot sends its ready frame before reading the signal;
+        # closing here could make that write fail and discard the stop request.
+        conn.shutdown(socket.SHUT_WR)
+        while recv_msg(conn) is not None:
+            pass
+    except (OSError, ValueError):
+        pass
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `5e4e7096b`. `stop()` now waits until the PID file and socket paths are gone (6s timeout, covering `worker.join(timeout=5)`) after a SIGTERM was actually delivered. If teardown does not complete, it raises `TimeoutError` and the CLI exits 1 without printing that the daemon stopped.

TimeToBuildBob:
CI is green on `5e4e7096b` (`fix(daemon): wait for teardown before reporting stop`). All 37 review threads are resolved.

Self-merge not eligible here (daemon/CLI/server paths + last Greptile score 4/5 on the previous head). Greptile re-review is in backoff (9 lifetime triggers / cap 8), so I am not requesting another review. Waiting for human review or merge.

gptme/server/daemon.py
```diff
@@ -0,0 +1,745 @@
...
+                    pid_file.seek(0)
+                    pid = int(pid_file.read())
+                    signaled = _signal_owned_pid(pid_file, pid)
+                else:
+                    fcntl.flock(pid_file, fcntl.LOCK_UN)
+        except (FileNotFoundError, ProcessLookupError, PermissionError, ValueError):
+            pass
+        if not signaled:
+            signaled = _stop_via_socket(self.socket_path)
+        if signaled:
```
TimeToBuildBob, line 167: ❌ **P1** — stop() returns normally even when neither signal path succeeds: if _signal_owned_pid returns False (e.g., because pidfd is unavailable) and _stop_via_socket also returns False (e.g., the socket is stale or connection refused), the method reaches the end with signaled=False and does nothing. The CLI's stop_cmd calls daemon.stop() and unconditionally prints 'Stopped daemon ...' without checking whether any signal was sent. This falsely tells the user the daemon stopped and its paths were cleaned up, when in fact the process may still be running. A subsequent start command can then fail with 'already running' against the live daemon.

gptme/server/daemon.py
```diff
@@ -0,0 +1,745 @@
...
+
+    def _handle_control_connection(self, server_sock: socket.socket) -> None:
+        """Accept a one-shot control client (stop) while the slot is occupied."""
+        try:
+            extra, _ = server_sock.accept()
+        except OSError:
+            return
+        extra.settimeout(1.0)
+        try:
+            msg = recv_msg(extra, bytearray())
```
TimeToBuildBob, line 440: ⚠️ **P2** — In `_handle_control_connection`, the SIGTERM stop message is read with `recv_msg(extra, bytearray())`, passing a fresh `bytearray()` each time. If the signal message arrives split across multiple reads (possible under load or if the socket buffer is full), the first read stores only part of the header/payload into the bytearray, then the socket times out after 1 second, raising `TimeoutError`. The partial data in the bytearray is discarded, and the stop request is never processed. The daemon remains running, and `stop()` times out waiting for teardown. This causes intermittent `stop` failures. Fix by reusing a persistent buffer across calls or using the no-buffer path with a full read loop.

gptme/server/daemon.py
```diff
@@ -0,0 +1,745 @@
...
+        with self._output_lock:
+            if text:
+                self._output_buf.append(text)
+                self._output_buf_size += len(text.encode())
+                while self._output_buf_size > _OUTPUT_HISTORY_BYTES:
+                    dropped = self._output_buf.popleft()
+                    self._output_buf_size -= len(dropped.encode())
+
+            if self._client is not None and text:
+                if not self._write_client(
```
TimeToBuildBob, line 321: ⚠️ **P2** — `_publish_output` holds `self._output_lock` while calling `_write_client`, which sends the message with a 5-second `_CLIENT_WRITE_TIMEOUT`. A non-reading client therefore blocks the worker thread for up to 5 seconds per output chunk. During that block the worker is not reading the subprocess's stdout pipe; if the subprocess writes more than the pipe buffer (typically 64 KiB) it blocks on its `write`, stalling the turn. The comment at lines 53-54 claims "a non-reading client cannot block a turn", but the client does block the turn for the duration of each write timeout. Concretely, an attached client that stops reading while a long tool output streams will cause the daemon to pause mid-turn until the write times out, then the client is disconnected and the turn resumes — delaying completion and failing the turn if it times out before the subprocess unblocks.

gptme/server/daemon.py
```diff
@@ -0,0 +1,745 @@
...
+    except OSError as exc:
+        raise RuntimeError("failed to fork daemon process") from exc
+
+
+def _daemonize() -> bool:
+    """Double-fork; return True only in the original parent process."""
+    if _fork() > 0:
+        return True
+
+    os.setsid()
```
TimeToBuildBob, line 516: ⚠️ **P2** — The daemonization routine never calls os.chdir(), so the daemon process retains the caller's working directory. When that directory is later removed or becomes inaccessible, subsequent turns spawned by _run_turn (which runs `[sys.executable, '-m', 'gptme', ...]` without a cwd= parameter) inherit the now-invalid cwd. gptme then fails to resolve relative workspace paths and file operations, breaking the session. Standard daemon practice is to chdir to a stable location such as '/' after forking.

gptme/server/daemon.py
TimeToBuildBob: Fixed in `9c7538dd4`. `stop()` now raises `RuntimeError` when neither pidfd nor the socket fallback delivered SIGTERM and the PID-file lock is still held. The CLI catches that and exits 1 without printing that the daemon stopped. Regression coverage asserts the lock remains held and that `os.kill` is never used on this path.

gptme/server/daemon.py
TimeToBuildBob: Dismissed — not a retry loop. `_handle_control_connection` makes a single `recv_msg` call for a tiny SIGTERM frame. A fresh `bytearray()` would only lose partial data if we retried after timeout; we do not. The unbuffered `_recv_exact` path has the same timeout/EOF behavior. Not blocking for Phase 1.

gptme/server/daemon.py
TimeToBuildBob: Dismissed — accepted Phase 1 bound, not a correctness bug. `_CLIENT_WRITE_TIMEOUT` (5s) then disconnects a non-reading client so it cannot stall a turn indefinitely. Async/non-blocking client writes are out of MVP scope (single-client). The 5s delay is the documented bound, not an unbounded hang.

gptme/server/daemon.py
TimeToBuildBob: Dismissed — wrong fix for this design. The daemon is supposed to inherit the caller's cwd so subsequent turns resolve the same workspace. `os.chdir('/')` would break relative workspace/file paths. Use `--workspace` when a stable absolute path is required. Removing the start directory later is an operator issue, not a daemonization bug.

TimeToBuildBob:
## Convergence adjudication — maintainer judgment required

**Fixed**
- Bob AI-review P1 (`stop()` reported success when neither pidfd nor socket SIGTERM was delivered): `9c7538dd4` raises if the PID-file lock is still held; the CLI exits 1 and does not print "Stopped daemon".
- Greptile P1 "Stop reports before shutdown": already addressed in `5e4e7096b` (wait for path teardown, or `TimeoutError`).

**Remaining** (non-blocking, with checkable reasons)
- Control-connection `recv_msg(..., bytearray())`: dismissed. One-shot read of a tiny SIGTERM frame, not a retry loop that would drop a persistent buffer.
- `_publish_output` holding the lock across a 5s client write: dismissed. Phase 1 bound — a non-reading client is disconnected after `_CLIENT_WRITE_TIMEOUT`, not stalled indefinitely. Async writes are out of MVP.
- Missing `os.chdir('/')`: dismissed. The daemon inherits the caller's cwd so turns resolve the same workspace; chdir to `/` would break relative paths. Use `--workspace` for a stable absolute path.

Did **not** trigger another Greptile review (lifetime triggers already past the cap). Last Greptile score 4/5 is stale (last reviewed commit `aecd9d21d`, before the teardown wait and this stop-signal failure path).

**CI**
- Head `9c7538dd4` is running CI. Local verification: 67 tests in `tests/test_daemon.py` pass; ruff and mypy on the touched files pass.

**Domain risk**
- New POSIX-only daemon lifecycle and Unix-socket code. Before merge, a maintainer should smoke start → attach → detach → reattach → stop on Linux/macOS, including `stop` with no attached client on macOS (no pidfd — socket fallback is the whole path). Windows remains gated/unsupported as intended.

**Convergence**
- Detector: `verdict: converged`, unresolved P1/P2 = 0. `round_convergence.stable_rounds: 0` (required: 2), `status: new_blocking` — the last capped Greptile round introduced a new P1; adjudication resolved remaining findings without another review round.

Merge-ready here means ready for maintainer judgment after current CI completes — no auto-merge. Self-merge is not eligible (daemon/CLI/server paths).

gptme/server/daemon.py
```diff
@@ -0,0 +1,757 @@
...
+    try:
+        while not stop_event.is_set():
+            if not _stdin_has_data(0.2):
+                continue
+            line = sys.stdin.readline()
+            if line == "":
+                break
+            seq += 1
+            try:
+                send_msg(conn, IPCMessage(type="input", data=line, seq=seq))
```
TimeToBuildBob, line 590: ⚠️ **P2** — The attach client does not handle `ValueError` from `send_msg` when an input line exceeds the IPC payload cap. `IPCMessage.encode` raises `ValueError("IPC payload too large")` if the JSON payload is over `MAX_IPC_PAYLOAD` (1 MiB). In `attach`, the line is sent via `send_msg(conn, IPCMessage(type="input", data=line, seq=seq))` inside a `try` that only catches `OSError`. If a user pastes a single line (or a file redirected to stdin) larger than ~1 MiB, this `ValueError` propagates out of `attach`, producing a traceback and exiting the CLI. The same issue exists on the daemon side: `_serve_client` catches `(OSError, ValueError)` when reading, so an oversized input from a client would disconnect that client rather than crashing the daemon, but the CLI crash is user-visible and unexpected. This is a robustness failure that can be triggered by ordinary (if extreme) terminal input.

gptme/server/daemon.py
```diff
@@ -0,0 +1,757 @@
...
+        finally:
+            pid_file.close()
+
+    def stop(self) -> None:
+        """Stop the daemon and wait until it releases its owned paths.
+
+        Never ``os.kill`` a numeric PID: without pidfd that can hit a reused
+        process. The socket path talks to the listening daemon instead.
+        """
+        if not self.socket_path.exists():
```
TimeToBuildBob, line 150: ⚠️ **P2** — `SessionDaemon.stop()` begins with `if not self.socket_path.exists(): return`, so when the socket file is absent (e.g., deleted by the user or a failed cleanup after a SIGKILL) the method returns without attempting to signal the daemon process via pidfd. As a result, a live daemon whose socket is gone cannot be stopped; `is_running()` also returns False because it gates on `socket_path.exists()`, so the CLI reports the session as not running and `gptme daemon stop` exits with "not running". The only recovery is a manual `kill` of the PID, which the code intentionally avoids. The fix is to check the PID file lock even when the socket is missing, using pidfd (or the pid file's contents) to signal the process.

gptme/server/daemon.py
TimeToBuildBob: Dismissed — not blocking. `MAX_IPC_PAYLOAD` is 1 MiB; a single attach stdin line that large is not Phase 1 UX (interactive prompts are short; bulk input goes through tools). Daemon `_serve_client` already catches `(OSError, ValueError)` on read, so the server does not crash. A CLI traceback on that input is noisy, not a correctness or security bug. Catch-and-warn is a follow-up nicety.

gptme/server/daemon.py
TimeToBuildBob: Dismissed — intentional liveness predicate, not a stop bug. `is_running()` and `stop()` both require the session socket because we refuse `os.kill` of a numeric PID (PID-reuse). If an operator unlinks the socket of a live daemon, the PID-file lock is still held (`LOCK_EX|LOCK_NB` in `_acquire_pid_file`), so a replacement start cannot take the session. Recovery is an operator kill of the still-owned process. pidfd-stop without a socket is Phase 2 recovery, not a Phase 1 merge blocker.

TimeToBuildBob:
## Convergence adjudication — maintainer judgment required

Update after CI completed green on `9c7538dd4` and a later Bob AI-review pass (15:54 UTC) filed two new P2s. Did **not** push more code. Did **not** trigger Greptile (lifetime triggers already past the cap).

**Fixed**
- None in this session. Prior blocking fix still stands: `9c7538dd4` makes `stop()` raise when neither pidfd nor socket SIGTERM was delivered and the PID-file lock is still held; CLI exits 1 without printing "Stopped daemon".
- Greptile P1 "Stop reports before shutdown": already addressed in `5e4e7096b`.

**Remaining** (non-blocking, with checkable reasons)
- Attach `ValueError` on stdin >1 MiB (`fp ee5961f66f91`): dismissed. `MAX_IPC_PAYLOAD` is 1 MiB; that input is not Phase 1 UX. Daemon `_serve_client` already catches `ValueError` on read. CLI traceback is noisy, not a correctness/security bug.
- `stop()` no-ops if the socket path is missing (`fp b800733d9d8f`): dismissed. Socket is the liveness/signal identity because we refuse `os.kill` of a numeric PID. PID-file lock still prevents a replacement start. Operator-deleted-socket recovery is Phase 2.
- Control-connection `recv_msg(..., bytearray())`: dismissed previously (one-shot read, not a retry loop).
- `_publish_output` holding the lock across a 5s client write: dismissed previously (Phase 1 bound; then disconnect).
- Missing `os.chdir('/')`: dismissed previously (daemon inherits caller cwd on purpose).

**CI**
- Head `9c7538dd4`: all required checks green (lint, typecheck, tests with and without API keys, API tests, Windows path slice, WebUI e2e, build/PyInstaller). Last Greptile score 4/5 is stale (last reviewed commit `aecd9d21d`).

**Domain risk**
- New POSIX-only daemon lifecycle and Unix-socket code. Before merge, smoke start → attach → detach → reattach → stop on Linux/macOS, including `stop` with no attached client on macOS (no pidfd — socket fallback is the whole path). Windows remains gated/unsupported as intended.

**Convergence**
- Detector: `verdict: converged`, unresolved Greptile P1/P2 = 0. `round_convergence.stable_rounds: 0` (required: 2), `status: new_blocking` — last capped Greptile round introduced a new P1; adjudication resolved remaining findings without another review round.

Merge-ready here means ready for maintainer judgment — no auto-merge. Self-merge is not eligible (daemon/CLI/server paths).

TimeToBuildBob:
The 2026-09-07 open-PR value triage recommends **closing** this PR rather than merging: it has no originating issue or external user, and the persistent-background-session capability it adds (new `gptme-daemon` binary + IPC protocol) duplicates what `gptme-server` + JSONL persistence + tmux already cover. It has been `CONFLICTING` since 08-29.

Per that recommendation I am **not** rebasing/merging or resolving the conflict — that would generate review churn on a PR slated for closure. The close awaits Erik's ratification of the triage close list. The daemon implementation task is parked on that decision.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Closing after today's value triage of Bob's open gptme PRs (Erik asked Bob to work through the list rather than review two dozen PRs by hand). Reason: duplicates what gptme-server plus a tmux session already provide; +2,369 lines with no issue asking for a daemon. Not lost: the branch stays; if the direction is wanted later it comes back as a smaller, issue-linked PR.

PR #3656, opened 2026-08-28
feat(harness): actuate request_tool_change — enable/disable tools mid-session (Phase 2)

TimeToBuildBob:
## What

Phase 2 of the EvoHarness adaptive tool configuration.

Phase 1 ([#3529](https://github.com/gptme/gptme/pull/3529), merged 2026-08-23) shipped `request_tool_change` as **audit-only**: the tool recorded intent in history but made no mutations. This PR makes `enable_tool` and `disable_tool` requests actually apply by mutating the session-local tool set.

## Design choice: Option 2 — session-state mutable enabled\_tools set

Three options from the design doc:

1. Message-level interception
2. **Session state** — mutate the ContextVar-backed `get_tools()` list ← this PR
3. Re-init per turn

Option 2 fits cleanly: `get_tools()`/`set_tools()` already expose the right API; the ContextVar approach makes changes session-scoped by construction; and no LLM-loop changes or provider re-init are needed.

### `gptme/tools/request_tool_change.py`

- `_enable_tool()`: finds the tool in `get_available_tools()`, appends it via `set_tools()`. No-ops if already loaded; returns an error if the tool name is unknown.
- `_disable_tool()`: removes the tool. Refuses to disable `request_tool_change` itself. No-ops if not loaded.
- `configure_tool` path: remains audit-only (Phase 3 scope).
- Updated docstring and `instructions` to reflect actuation semantics.

### `tests/test_tools_request_tool_change.py`

8 new test cases (13 total, all passing):

| Class | Tests |
|---|---|
| `TestEnableTool` | adds to session; no-op if already loaded; rejects unknown |
| `TestDisableTool` | removes from session; no-op if not loaded; refuses self-disable |
| `TestConfigureTool` | audit-only, tool list unchanged |
| `TestEnableDisableIntegration` | enable → verify callable → disable → verify not callable (requirement 3.1) |

An `isolated_tools` fixture calls `clear_tools()` + `init_tools()` before each Phase 2 test so ContextVar state doesn't leak between cases.

## Refs

- Phase 1 PR: #3529
- Design doc: `knowledge/technical-designs/2026-08-12-evoharness-adaptive-tool-config-design.md`
- Sources: arXiv:2608.05446 (EvoHarness-RL), arXiv:2607.04528

### DIFF
--- a/gptme/commands/base.py
+++ b/gptme/commands/base.py
@@ -133,6 +133,11 @@ def register_command(
     if owner_tool is not None:
         for cmd_name in names:
             _command_owners[cmd_name] = owner_tool
+    else:
+        # Re-registering without an owner must drop a stale mapping so the
+        # command is unowned (always enabled), matching the docstring.
+        for cmd_name in names:
+            _command_owners.pop(cmd_name, None)
 
     logger.debug(
         f"Registered command: {name}" + (f" (aliases: {aliases})" if aliases else "")
--- a/gptme/commands/base.py
+++ b/gptme/commands/base.py
@@ -44,6 +44,12 @@ _command_registry: dict[str, CommandHandler] = {}
 # Completer registry - maps command names to their completer functions
 _command_completers: dict[str, CommandCompleter] = {}
 
+# Optional owning tool for dynamically registered commands. Process-global:
+# sibling sessions on gptme-server share the registry. Dispatch consults the
+# session-local loaded-tool set so a disable in this session cannot run the
+# command here without yanking it from every other session.
+_command_owners: dict[str, str] = {}
+
 
 def command(
     name: str,
@@ -98,6 +104,7 @@ def register_command(
     handler: CommandHandler,
     aliases: list[str] | None = None,
     completer: CommandCompleter | None = None,
+    owner_tool: str | None = None,
 ) -> None:
     """Register a command handler dynamically (for tools).
 
@@ -107,8 +114,11 @@ def register_command(
         aliases: Optional list of command aliases
         completer: Optional function for argument completion.
                    Takes (partial_arg, previous_args) and returns list of (completion, description) tuples.
+        owner_tool: Tool that owns this command. When set, dispatch and listing
+                    require that tool to be loaded in the current session.
     """
     _command_registry[name] = handler
+    names = [name, *(aliases or [])]
     if aliases:
         for alias in aliases:
             _command_registry[alias] = handler
@@ -119,6 +129,10 @@ def register_command(
         if aliases:
             for alias in aliases:
                 _command_completers[alias] = completer
+
+    if owner_tool is not None:
+        for cmd_name in names:
+            _command_owners[cmd_name] = owner_tool
 
     logger.debug(
         f"Registered command: {name}" + (f" (aliases: {aliases})" if aliases else "")
@@ -136,6 +150,21 @@ def unregister_command(name: str) -> None:
         logger.debug(f"Unregistered command: {name}")
     if name in _command_completers:
         del _command_completers[name]
+    _command_owners.pop(name, None)
+
+
+def _command_enabled_in_session(name: str) -> bool:
+    """True if this command may run in the current session.
+
+    Unowned commands (built-ins, plugins) are always enabled. Tool-owned
+    commands require their owner to be in the session-local loaded set.
+    """
+    owner = _command_owners.get(name)
+    if owner is None:
+        return True
+    from ..tools import has_tool  # fmt: skip
+
+    return has_tool(owner)
 
 
 def get_registered_commands() -> list[str]:
@@ -150,8 +179,11 @@ def get_command_completer(name: str) -> CommandCompleter | None:
         name: Command name (without leading /)
 
     Returns:
-        Completer function or None if no completer registered
+        Completer function or None if no completer registered, or if the
+        command's owning tool is not loaded in this session.
     """
+    if not _command_enabled_in_session(name):
+        return None
     return _command_completers.get(name)
 
 
@@ -240,6 +272,18 @@ def handle_cmd(
 
     # Check if command is registered
     if name in _command_registry:
+        if not _command_enabled_in_session(name):
+            owner = _command_owners[name]
+            manager.undo(1, quiet=True)
+            msg = (
+                f"Command /{name} is unavailable because tool '{owner}' "
+                "is not enabled in this session."
+            )
+            if is_output_json():
+                _emit_json_command_output(msg)
+            else:
+                print(msg)
+            return
         handler = _command_registry[name]
         ctx = CommandContext(args=args, full_args=full_args, manager=manager)
         if is_output_json():
@@ -280,6 +324,8 @@ def get_commands_with_descriptions() -> list[tuple[str, str]]:
     seen_handlers: set[int] = set()  # Track handler object IDs to skip aliases
 
     for name in _command_registry:
+        if not _command_enabled_in_session(name):
+            continue
         handler = _command_registry[name]
         handler_id = id(handler)
         if handler_id in seen_handlers:
@@ -302,6 +348,9 @@ def get_commands_with_descriptions() -> list[tuple[str, str]]:
 
 
 def get_user_commands() -> list[str]:
-    """Returns a list of all user commands, including tool-registered commands"""
-    # Get all registered commands (includes built-in + tool-registered)
-    return [f"/{cmd}" for cmd in _command_registry]
+    """Returns user commands enabled in this session.
+
+    Includes built-ins and tool-registered commands whose owning tool is
+    loaded. Process-global registrations of disabled tools are omitted.
+    """
+    return [f"/{cmd}" for cmd in _command_registry if _command_enabled_in_session(cmd)]
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -342,23 +342,19 @@ def execute_msg(
     """
     assert msg.role == "assistant", "Only assistant messages can be executed"
 
-    # Snapshot runnability once per tool_use. Evaluating `is_runnable` a second
-    # time later would open a TOCTOU gap: a tool whose loaded-state changes
-    # between the two checks (e.g. a subagent thread concurrently (re)initializing
-    # tools) could fall through both branches and leave a structured tool_use with
-    # no tool_result — which the Anthropic API rejects with a hard 400 (#554).
-    classified = [(tu, tu.is_runnable) for tu in ToolUse.iter_from_content(msg.content)]
+    # Materialize every parsed tool_use first so each structured call still gets
+    # exactly one result. Evaluate runnability once per call, immediately before
+    # the branch, so an earlier request_tool_change in this response can enable
+    # or disable a sibling without a concurrent load/unload skipping both the
+    # execute and pairing paths (Anthropic 400, #554).
+    classified = list(ToolUse.iter_from_content(msg.content))
 
     if not classified:
         return
 
     remaining = iter(classified)
-    for tooluse, was_runnable in remaining:
-        # Preserve the initial classification for structured result pairing, but
-        # honor intentional session-local tool changes made by an earlier call in
-        # this same response.  In particular, request_tool_change may disable a
-        # sibling call after classification and before execution.
-        runnable = was_runnable and tooluse.is_runnable
+    for tooluse in remaining:
+        runnable = tooluse.is_runnable
         if runnable:
             with terminal_state_title(f"🛠️ running {tooluse.tool}"):
                 t0 = time.monotonic()
@@ -374,7 +370,7 @@ def execute_msg(
                     # Drain the rest: any structured tool_use that's left in the
                     # message still needs a paired tool_result or the next API
                     # request will 400 with a dangling tool_use.
-                    for rem_tu, _ in remaining:
+                    for rem_tu in remaining:
                         if rem_tu.call_id is not None:
                             yield Message(
                                 "system",
@@ -503,6 +499,26 @@ def set_tools(tools: list[ToolSpec]) -> None:
     ContextVars from the parent context aren't visible.
     """
     _loaded_tools_var.set(tools)
+
+
+def unload_tool(tool_name: str) -> ToolSpec:
+    """Unload one tool and unregister the hooks and commands it owns."""
+    with _tools_init_lock:
+        tool = get_tool(tool_name)
+        if tool is None:
+            raise ValueError(f"Tool '{tool_name}' is not loaded")
+
+        from ..commands import unregister_command
+        from ..hooks import unregister_hook
+
+        for hook_name in tool.hooks:
+            unregister_hook(f"{tool.name}.{hook_name}")
+        for command_name in tool.commands:
+            unregister_command(command_name)
+
+        set_tools([loaded for loaded in get_tools() if loaded.name != tool_name])
+        logger.info("Unloaded tool '%s' mid-conversation", tool_name)
+        return tool
 
 
 def get_tool(tool_name: str) -> ToolSpec | None:
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -502,7 +502,7 @@ def set_tools(tools: list[ToolSpec]) -> None:
 
 
 def unload_tool(tool_name: str) -> ToolSpec:
-    """Unload one tool and unregister the hooks and commands it owns."""
+    """Unload one tool and best-effort unregister its hooks and commands."""
     with _tools_init_lock:
         tool = get_tool(tool_name)
         if tool is None:
@@ -511,12 +511,26 @@ def unload_tool(tool_name: str) -> ToolSpec:
         from ..commands import unregister_command
         from ..hooks import unregister_hook
 
-        for hook_name in tool.hooks:
-            unregister_hook(f"{tool.name}.{hook_name}")
-        for command_name in tool.commands:
-            unregister_command(command_name)
-
         set_tools([loaded for loaded in get_tools() if loaded.name != tool_name])
+        for hook_name in tool.hooks:
+            try:
+                unregister_hook(f"{tool.name}.{hook_name}")
+            except Exception:
+                logger.exception(
+                    "Failed to unregister hook '%s.%s' while unloading tool",
+                    tool.name,
+                    hook_name,
+                )
+        for command_name in tool.commands:
+            try:
+                unregister_command(command_name)
+            except Exception:
+                logger.exception(
+                    "Failed to unregister command '%s' while unloading tool '%s'",
+                    command_name,
+                    tool.name,
+                )
+
         logger.info("Unloaded tool '%s' mid-conversation", tool_name)
         return tool
 
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -64,6 +64,13 @@ _loaded_tools_var: ContextVar[list[ToolSpec] | None] = ContextVar(
 _available_tools_var: ContextVar[list[ToolSpec] | None] = ContextVar(
     "available_tools", default=None
 )
+# Effective operator allowlist from the last init_tools() in this context.
+# None means unrestricted (default session); a list is a hard cap for model
+# enablement via request_tool_change. /tools load still uses load_tool()
+# directly and is not gated here.
+_session_allowlist_var: ContextVar[list[str] | None] = ContextVar(
+    "session_allowlist", default=None
+)
 
 # Note: Tools must be initialized in each context that needs them.
 # This is particularly important for server environments where request handling
@@ -185,6 +192,7 @@ def init_tools(
                 allowlist = config.chat.tools
 
         allowlist = expand_tool_allowlist_presets(allowlist)
+        set_session_allowlist(allowlist)
 
         # Partition allowlist into file paths and tool names
         file_paths: list[str] = []
@@ -485,6 +493,7 @@ def clear_tools():
     """
     _set_available_tools_cache(None)
     _loaded_tools_var.set([])
+    _session_allowlist_var.set(None)
 
 
 def get_tools() -> list[ToolSpec]:
@@ -501,6 +510,20 @@ def set_tools(tools: list[ToolSpec]) -> None:
     _loaded_tools_var.set(tools)
 
 
+def get_session_allowlist() -> list[str] | None:
+    """Return the operator tool allowlist captured by the last init_tools()."""
+    return _session_allowlist_var.get()
+
+
+def set_session_allowlist(allowlist: list[str] | None) -> None:
+    """Override the session tool allowlist for this context.
+
+    ``None`` means unrestricted. An explicit list is the hard cap that
+    ``request_tool_change`` enablement must honor.
+    """
+    _session_allowlist_var.set(list(allowlist) if allowlist is not None else None)
+
+
 def unload_tool(tool_name: str) -> ToolSpec:
     """Unload one tool and best-effort unregister its hooks and commands."""
     with _tools_init_lock:
@@ -511,7 +534,10 @@ def unload_tool(tool_name: str) -> ToolSpec:
         from ..commands import unregister_command
         from ..hooks import unregister_hook
 
-        set_tools([loaded for loaded in get_tools() if loaded.name != tool_name])
+        # Filter by identity: get_tool() matches name *or* block_types, so a
+        # block-type argument would miss a name-only filter and leave a zombie
+        # tool loaded after its hooks/commands were unregistered.
+        set_tools([loaded for loaded in get_tools() if loaded is not tool])
         for hook_name in tool.hooks:
             try:
                 unregister_hook(f"{tool.name}.{hook_name}")
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -525,18 +525,25 @@ def set_session_allowlist(allowlist: list[str] | None) -> None:
 
 
 def unload_tool(tool_name: str) -> ToolSpec:
-    """Unload one tool and best-effort unregister its hooks and commands."""
+    """Unload one tool and best-effort unregister its session-local hooks.
+
+    Commands are *not* unregistered. The slash-command registry is
+    process-global (``gptme.commands.base._command_registry``), shared by every
+    session in this process. Dropping a command here would remove that slash
+    command from sibling conversations on gptme-server even though their
+    context-local tool sets still contain the tool. Hooks are ContextVar-backed,
+    so unregistering them is session-scoped and safe.
+    """
     with _tools_init_lock:
         tool = get_tool(tool_name)
         if tool is None:
             raise ValueError(f"Tool '{tool_name}' is not loaded")
 
-        from ..commands import unregister_command
         from ..hooks import unregister_hook
 
         # Filter by identity: get_tool() matches name *or* block_types, so a
         # block-type argument would miss a name-only filter and leave a zombie
-        # tool loaded after its hooks/commands were unregistered.
+        # tool loaded after its hooks were unregistered.
         set_tools([loaded for loaded in get_tools() if loaded is not tool])
         for hook_name in tool.hooks:
             try:
@@ -546,15 +553,6 @@ def unload_tool(tool_name: str) -> ToolSpec:
                     "Failed to unregister hook '%s.%s' while unloading tool",
                     tool.name,
                     hook_name,
-                )
-        for command_name in tool.commands:
-            try:
-                unregister_command(command_name)
-            except Exception:
-                logger.exception(
-                    "Failed to unregister command '%s' while unloading tool '%s'",
-                    command_name,
-                    tool.name,
                 )
 
         logger.info("Unloaded tool '%s' mid-conversation", tool_name)
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -353,7 +353,12 @@ def execute_msg(
         return
 
     remaining = iter(classified)
-    for tooluse, runnable in remaining:
+    for tooluse, was_runnable in remaining:
+        # Preserve the initial classification for structured result pairing, but
+        # honor intentional session-local tool changes made by an earlier call in
+        # this same response.  In particular, request_tool_change may disable a
+        # sibling call after classification and before execution.
+        runnable = was_runnable and tooluse.is_runnable
         if runnable:
             with terminal_state_title(f"🛠️ running {tooluse.tool}"):
                 t0 = time.monotonic()
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -533,6 +533,10 @@ def unload_tool(tool_name: str) -> ToolSpec:
     command from sibling conversations on gptme-server even though their
     context-local tool sets still contain the tool. Hooks are ContextVar-backed,
     so unregistering them is session-scoped and safe.
+
+    Session isolation for commands is enforced at dispatch: ``handle_cmd``
+    refuses to run a tool-owned command when that tool is not in this
+    session's loaded set.
     """
     with _tools_init_lock:
         tool = get_tool(tool_name)
--- a/gptme/tools/request_tool_change.py
+++ b/gptme/tools/request_tool_change.py
@@ -14,7 +14,14 @@ from __future__ import annotations
 import logging
 
 from ..message import Message
-from . import get_available_tools, get_tools, load_tool, unload_tool
+from . import (
+    get_available_tools,
+    get_session_allowlist,
+    get_tools,
+    load_tool,
+    unload_tool,
+)
+from ._allowlist import tool_matches_allowlist
 from .base import Parameter, ToolSpec
 
 logger = logging.getLogger(__name__)
@@ -38,6 +45,20 @@ def _enable_tool(tool_name: str) -> Message:
             f"request_tool_change: '{tool_name}' is already enabled — no change.",
             quiet=True,
         )
+
+    allowlist = get_session_allowlist()
+    if allowlist is not None:
+        available = next(
+            (t for t in get_available_tools() if t.name == tool_name), None
+        )
+        hints = available.hints if available is not None else frozenset()
+        if not tool_matches_allowlist(tool_name, allowlist, hints):
+            return Message(
+                "system",
+                f"request_tool_change: '{tool_name}' is not permitted by "
+                "the session tool allowlist — no change.",
+                quiet=True,
+            )
 
     try:
         load_tool(tool_name)
--- a/gptme/tools/request_tool_change.py
+++ b/gptme/tools/request_tool_change.py
@@ -48,10 +48,8 @@ def _enable_tool(tool_name: str) -> Message:
 
     allowlist = get_session_allowlist()
     if allowlist is not None:
-        available = next(
-            (t for t in get_available_tools() if t.name == tool_name), None
-        )
-        hints = available.hints if available is not None else frozenset()
+        spec = next((t for t in get_available_tools() if t.name == tool_name), None)
+        hints = spec.hints if spec is not None else frozenset()
         if not tool_matches_allowlist(tool_name, allowlist, hints):
             return Message(
                 "system",
@@ -152,9 +150,10 @@ def execute_request_tool_change(
             quiet=True,
         )
 
-    # Validate against available (not just loaded) tools so the model can name
-    # any tool known to gptme's module system.
-    known_tools = {t.name for t in get_available_tools(include_mcp=False)} | {
+    # Include MCP tools (default get_available_tools) so an allowlisted MCP
+    # tool can be enabled before it is loaded. Union with get_tools() so a
+    # file-loaded tool that is not in module discovery can still be disabled.
+    known_tools = {t.name for t in get_available_tools()} | {
         t.name for t in get_tools()
     }
     if tool_name not in known_tools:
--- a/gptme/tools/request_tool_change.py
+++ b/gptme/tools/request_tool_change.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 import logging
 
 from ..message import Message
-from . import get_available_tools, get_tools, load_tool, set_tools
+from . import get_available_tools, get_tools, load_tool, unload_tool
 from .base import Parameter, ToolSpec
 
 logger = logging.getLogger(__name__)
@@ -41,7 +41,8 @@ def _enable_tool(tool_name: str) -> Message:
 
     try:
         load_tool(tool_name)
-    except ValueError as exc:
+    except Exception as exc:
+        logger.exception("request_tool_change: failed to enable '%s'", tool_name)
         return Message(
             "system",
             f"request_tool_change: could not enable '{tool_name}': {exc}",
@@ -69,16 +70,23 @@ def _disable_tool(tool_name: str) -> Message:
             quiet=True,
         )
 
-    loaded = get_tools()
-    remaining = [t for t in loaded if t.name != tool_name]
-    if len(remaining) == len(loaded):
+    if not any(t.name == tool_name for t in get_tools()):
         return Message(
             "system",
             f"request_tool_change: '{tool_name}' is not currently enabled — no change.",
             quiet=True,
         )
 
-    set_tools(remaining)
+    try:
+        unload_tool(tool_name)
+    except Exception as exc:
+        logger.exception("request_tool_change: failed to disable '%s'", tool_name)
+        return Message(
+            "system",
+            f"request_tool_change: could not disable '{tool_name}': {exc}",
+            quiet=True,
+        )
+
     logger.info("request_tool_change: disabled tool '%s'", tool_name)
     return Message(
         "system",
@@ -125,7 +133,9 @@ def execute_request_tool_change(
 
     # Validate against available (not just loaded) tools so the model can name
     # any tool known to gptme's module system.
-    known_tools = {t.name for t in get_available_tools(include_mcp=False)}
+    known_tools = {t.name for t in get_available_tools(include_mcp=False)} | {
+        t.name for t in get_tools()
+    }
     if tool_name not in known_tools:
         return Message(
             "system",
--- a/gptme/tools/request_tool_change.py
+++ b/gptme/tools/request_tool_change.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 import logging
 
 from ..message import Message
-from . import get_available_tools, get_tools, set_tools
+from . import get_available_tools, get_tools, load_tool, set_tools
 from .base import Parameter, ToolSpec
 
 logger = logging.getLogger(__name__)
@@ -39,15 +39,15 @@ def _enable_tool(tool_name: str) -> Message:
             quiet=True,
         )
 
-    available = {t.name: t for t in get_available_tools(include_mcp=False)}
-    if tool_name not in available:
+    try:
+        load_tool(tool_name)
+    except ValueError as exc:
         return Message(
             "system",
-            f"request_tool_change: unknown tool '{tool_name}'",
+            f"request_tool_change: could not enable '{tool_name}': {exc}",
             quiet=True,
         )
 
-    set_tools([*loaded, available[tool_name]])
     logger.info("request_tool_change: enabled tool '%s'", tool_name)
     return Message(
         "system",
--- a/gptme/tools/request_tool_change.py
+++ b/gptme/tools/request_tool_change.py
@@ -1,18 +1,90 @@
 """Opt-in tool for assistants to request session tool configuration changes.
 
-The request is audit-only: calling this tool records structured intent in ordinary
-assistant tool-call and result-message history. It does not change the active tool
-configuration.
+Phase 1 (audit-only) recorded the intent in tool-call history without mutating
+the active tool set.  Phase 2 (this file) actually applies enable/disable
+requests by mutating the session-local ``_loaded_tools_var`` ContextVar that
+``get_tools()`` reads.
+
+``configure_tool`` requests remain audit-only; a future phase can add actuation
+for them without altering the enable/disable paths.
 """
 
 from __future__ import annotations
 
+import logging
+
 from ..message import Message
-from . import get_available_tools
+from . import get_available_tools, get_tools, set_tools
 from .base import Parameter, ToolSpec
+
+logger = logging.getLogger(__name__)
 
 _VALID_CHANGE_TYPES = {"enable_tool", "disable_tool", "configure_tool"}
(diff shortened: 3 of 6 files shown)


gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,103 @@
...
+
+    available = {t.name: t for t in get_available_tools(include_mcp=False)}
+    if tool_name not in available:
+        return Message(
+            "system",
+            f"request_tool_change: unknown tool '{tool_name}'",
+            quiet=True,
+        )
+
+    set_tools([*loaded, available[tool_name]])
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in aface374b: enablement now delegates to `load_tool()`, which enforces availability and runs initialization plus hook/command registration. Added regression tests for initialization and unavailable tools.

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,103 @@
...
+    loaded = get_tools()
+    remaining = [t for t in loaded if t.name != tool_name]
+    if len(remaining) == len(loaded):
+        return Message(
+            "system",
+            f"request_tool_change: '{tool_name}' is not currently enabled — no change.",
+            quiet=True,
+        )
+
+    set_tools(remaining)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in aface374b: `execute_msg()` retains its initial classification for structured result pairing but rechecks runnability immediately before each call. A regression test now disables `shell` and attempts it later in the same structured response, verifying a paired non-execution result.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `aface374b55e`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/__init__.py`
- `gptme/tools/request_tool_change.py`

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,103 @@
...
```
TimeToBuildBob: ❌ **P1** — _disable_tool removes a tool from the loaded list via set_tools(remaining) but does not unregister hooks or commands registered by _init_single_tool (register_hooks/register_commands). There is no unregister mechanism in the codebase, so the tool's global hooks (e.g., message listeners, file-watch callbacks) remain active after it is disabled. This means a disabled tool can continue to have side effects such as logging or intercepting events, potentially leaking information or causing unexpected behavior. The impact is that the disable operation is incomplete and can leave security-sensitive monitoring active.

gptme/tools/__init__.py
```diff
@@ -353,7 +353,12 @@ def execute_msg(
         return
 
     remaining = iter(classified)
-    for tooluse, runnable in remaining:
+    for tooluse, was_runnable in remaining:
+        # Preserve the initial classification for structured result pairing, but
+        # honor intentional session-local tool changes made by an earlier call in
+        # this same response.  In particular, request_tool_change may disable a
+        # sibling call after classification and before execution.
+        runnable = was_runnable and tooluse.is_runnable
```
TimeToBuildBob: ❌ **P1** — In execute_msg, `runnable = was_runnable and tooluse.is_runnable` preserves the initial classification but only allows a tool that was runnable at classification time to become non-runnable mid-response. If a tool is enabled by an earlier call in the same message, its `is_runnable` becomes True, but because `was_runnable` was False, `runnable` stays False. Consequently, the model cannot enable a tool and then call it in the same assistant turn; the call is always rejected with 'not available for execution'. This defeats the purpose of enabling a tool mid-session.

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,103 @@
...
+    if any(t.name == tool_name for t in loaded):
+        return Message(
+            "system",
+            f"request_tool_change: '{tool_name}' is already enabled — no change.",
+            quiet=True,
+        )
+
+    try:
+        load_tool(tool_name)
+    except ValueError as exc:
```
TimeToBuildBob: ❌ **P1** — In _enable_tool, load_tool(tool_name) may raise exceptions other than ValueError, for example if the tool's init() function raises an arbitrary exception. The current handler only catches ValueError, so such an exception propagates out of execute_request_tool_change and up through execute_msg, aborting the entire response loop and causing the session to fail with an unhandled traceback rather than a graceful error message. This is a real defect because any tool with a buggy init can now break the conversation when the model tries to enable it. The fix is to catch a broader exception class and return a message indicating that the tool could not be enabled.

tests/test_tools_request_tool_change.py
```diff
@@ -122,3 +128,258 @@ def test_request_tool_change_rejects_non_string_fields():
...
+    """Patch get_available_tools in the module under test."""
+    return patch(
+        "gptme.tools.request_tool_change.get_available_tools",
+        return_value=tools,
+    )
+
+
+class TestEnableTool:
+    def test_enable_adds_tool_to_session(self, isolated_tools):
+        with _patch_available([_FAKE_SHELL, tool]):
```
TimeToBuildBob, line 184: ⚠️ **P2** — test_enable_adds_tool_to_session patches only `gptme.tools.request_tool_change.get_available_tools` via `_patch_available`, but `_enable_tool` calls `load_tool` which imports `get_available_tools` from `gptme.tools` and uses the unpatched real function. Since `_FAKE_SHELL` is not a real available tool, `load_tool` raises ValueError, causing the test to fail unless a tool named 'shell' exists in the environment. Thus the test is non-hermetic and depends on the host's installed tools, making it flaky and unable to reliably verify the enable path.

tests/test_tools_request_tool_change.py
```diff
@@ -122,3 +128,258 @@ def test_request_tool_change_rejects_non_string_fields():
...
+                tool_name="shell",
+                reason="Done with shell phase",
+                urgency="low",
+            )
+        assert "disabled" in r_disable.content
+        assert not has_tool("shell"), "shell should be removed after disable"
+
+    def test_disable_prevents_later_call_in_same_response(self, isolated_tools):
+        set_tools([*get_tools(), _FAKE_SHELL])
+        set_tool_format("tool")
```
TimeToBuildBob, line 723: ⚠️ **P2** — In test_disable_prevents_later_call_in_same_response, set_tool_format("tool") changes the global tool format, but neither the test nor the isolated_tools fixture restores it. The fixture resets only the tool list and available-tools cache, not the module-level tool format. Any test that runs after this one inherits the 'tool' format, which can break tests that assume the default markdown format, causing order-dependent failures. The fix is to save and restore the format in a finally block or have the fixture reset it.

gptme/tools/request_tool_change.py
```diff
@@ -46,7 +123,9 @@ def execute_request_tool_change(
             quiet=True,
         )
 
-    known_tools = {tool.name for tool in get_available_tools(include_mcp=False)}
+    # Validate against available (not just loaded) tools so the model can name
+    # any tool known to gptme's module system.
+    known_tools = {t.name for t in get_available_tools(include_mcp=False)}
```
TimeToBuildBob: ⚠️ **P2** — The validation of tool names against `get_available_tools(include_mcp=False)` means that tools loaded from file paths (via `load_from_file` in `init_tools`) and MCP tools are not in `known_tools`. Consequently, `execute_request_tool_change` rejects requests to disable or enable these tools with 'unknown tool'. This prevents the assistant from managing custom/mid-session-loaded tools, which is the whole point of the feature. For example, if a user loads a tool from a `.py` file, the LLM cannot disable it even if it's misbehaving, because its name is not in the available set. The fix is to validate against the loaded tool set as well, or include MCP and file-loaded tools in the known set.

TimeToBuildBob:
## 🤖 AI code review

The change adds a powerful runtime capability where the language model can enable or disable tools during a session. The implementation is generally solid: per‑session allowlist enforcement prevents privilege escalation, commands are correctly tied to their owning tools, and the execution loop now reacts to tool availability within a single response.

Key strengths:
- The allowlist is enforced on every enable request, preventing escalation when a policy is set.
- Commands from disabled tools are blocked at dispatch time without removing them from the global registry.
- The per‑call `is_runnable` check correctly handles same‑response enable/disable sequences.

The primary issue is a minor logic inconsistency in the disable path of `request_tool_change` (pre‑check uses only tool `name`, ignoring block types). No critical security vulnerabilities were found.

**Performance consideration**: `execute_msg()` now evaluates `is_runnable` for every `ToolUse`, which adds a small per‑call cost (the property walks the loaded tools list). This is acceptable as tool calls are not high‑frequency.

**Design note**: The assistant can disable any tool that is not the `request_tool_change` tool itself. If a safety‑critical guard is implemented as a separate tool (e.g., a pre‑execution checker), an adversarial assistant could disable it before taking a malicious action. While this is a feature in the current design, it should be reviewed against safety requirements.

<details><summary><b>1 finding</b> · ⚠️ **1** P2 — non-blocking (round cap reached)</summary>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/tools/request_tool_change.py:58`

`has_tool()` in gptme/tools/__init__.py checks both the tool `name` and its block types. However, the pre‑check in `_disable_tool()` (gptme/tools/request_tool_change.py) uses `any(t.name == tool_name for t in get_tools())` which only matches by `name`. If a caller passes a block type (e.g., `bash` for a tool with name `shell`), the pre‑check reports that the tool is “not currently enabled” and the disable operation is skipped, even though `unload_tool()` would correctly find and remove it. While the input validation (`known_tools`) currently rejects block types, this is a latent inconsistency and a future regression risk.

```suggestion
Use the existing `has_tool(tool_name)` helper for the pre‑check, which correctly handles both names and block types.

```python
from gptme.tools import has_tool
...
if not has_tool(tool_name):
    return Message("system", f"request_tool_change: '{tool_name}' is not currently enabled", quiet=True)
```
```

<sub>**How this was verified:** not restated the already-found issue about the `_disable_tool` pre-check. This is a different gate and different symptom.</sub>

</details>

- `gptme/tools/__init__.py` — The code under review spans several modules. The core changes are: - **`gptme/tools/__init__.py`**: Added a per-session allowlist via `set_session_allowlist()`, and changed the execution loop in `execute_msg()` to evaluate `is_runnable` per tool call rather than once at the start. - **`gptme/command
- `gptme/commands/base.py` — Command registry changes, including owner tracking and session-aware dispatch.
- `gptme/tools/request_tool_change.py` — Runtime tool enable/disable logic with allowlist enforcement.
- `gptme/tools/base.py` — Tool execution changes to handle runtime unavailability.

<sub>Also changed, not described above: `tests/test_commands.py`, `tests/test_tools_request_tool_change.py`</sub>

</details>

| `aface374b55e` | 2/5 | 6 | llm | 2026-08-28 05:15 UTC |
| `f552a3c8c0f8` | 3/5 | 2 | llm | 2026-08-28 17:30 UTC |
| `9292a3070c32` | 2/5 | 2 | llm | 2026-08-29 03:31 UTC |
| `62ec507d022b` | 3/5 | 3 | llm | 2026-08-29 15:30 UTC |
| `53b1e0ac371f` | 3/5 | 5 | llm | 2026-08-29 18:38 UTC |

</details>

<sub>Reviewed `f8e31fbf7d89` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1111s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

<!-- bob-ai-review {"sha": "f8e31fbf7d89", "score": 4, "engine": "llm", "findings": [{"fp": "bdd9d78167d5", "severity": "P2"}], "consensus": {"requested": 1, "answered": 1, "jobs_requested": 1, "jobs_answered": 1, "min_agreement": 1, "min_agreement_requested": 2, "rejected": 0, "failures": []}, "history": [{"sha": "aface374b55e", "score": 2, "findings": 6, "engine": "llm", "at": "2026-08-28 05:15 UTC"}, {"sha": "f552a3c8c0f8", "score": 3, "findings": 2, "engine": "llm", "at": "2026-08-28 17:30 UTC"}, {"sha": "9292a3070c32", "score": 2, "findings": 2, "engine": "llm", "at": "2026-08-29 03:31 UTC"}, {"sha": "62ec507d022b", "score": 3, "findings": 3, "engine": "llm", "at": "2026-08-29 15:30 UTC"}, {"sha": "53b1e0ac371f", "score": 3, "findings": 5, "engine": "llm", "at": "2026-08-29 18:38 UTC"}, {"sha": "f8e31fbf7d89", "score": 4, "findings": 1, "engine": "llm", "at": "2026-08-30 07:38 UTC"}], "suppressed": [{"fp": "e40964beafcf", "file": "gptme/tools/request_tool_change.py", "line": 44, "tokens": ["_enable_tool", "arbitrary", "conversation", "exception", "exceptions", "execute_msg", "execute_request_tool_change", "indicating", "load_tool", "propagates", "tool_name", "valueerror"], "why": "resolved"}, {"fp": "f46fe08b21eb", "file": "gptme/tools/__init__.py", "line": 515, "tokens": ["_disable_tool", "available", "corrupted", "execute_request_tool_change", "inconsistent", "propagates", "unload_tool", "unregister", "unregister_command", "unregister_hook", "unregistered", "unregistration"], "why": "resolved"}, {"fp": "306f038d91ae", "file": "gptme/tools/request_tool_change.py", "line": 43, "tokens": ["_enable_tool", "disabled_by_default", "discovered", "explicitly", "get_available_tools", "get_toolchain", "include_mcp", "is_available", "known_tools", "request_tool_change", "restrictions", "tool_allowlist"], "why": "resolved"}, {"fp": "c8b5c0be46c2", "file": "gptme/tools/__init__.py", "line": 514, "tokens": ["_disable_tool", "arguments", "block_types", "callable", "commands", "get_tools", "registrations", "set_tools", "subsequent", "tool_name", "unload_tool", "unregistered"], "why": "resolved"}, {"fp": "f2e84ca68148", "file": "gptme/commands/base.py", "line": 153, "tokens": ["_command_enabled_in_session", "_command_owners", "atomically", "different", "incorrectly", "intentionally", "registered", "registers", "registration", "unload_tool", "unregistered", "violation"], "why": "dismissed"}, {"fp": "3efb931d515d", "file": "gptme/tools/request_tool_change.py", "line": 54, "tokens": ["_enable_tool", "allowlist", "assistant", "available", "available_tools", "discovery", "get_available_tools", "incorrectly", "legitimate", "misleading", "succeeding", "tool_matches_allowlist"], "why": "resolved"}, {"fp": "3548d84353e1", "file": "gptme/commands/base.py", "line": 167, "tokens": ["_loaded_tools_var", "contextvar", "conversation", "dynamically", "explicitly", "get_user_commands", "handle_cmd", "init_tools", "initialized", "regardless", "registered", "regression"], "why": "dismissed"}, {"fp": "4ed233f459f7", "file": "gptme/tools/request_tool_change.py", "line": 42, "tokens": ["_enable_tool", "allowlist", "assistant", "consequence", "considered", "deployments", "enablement", "get_session_allowlist", "init_tools", "observable", "request_tool_change", "restrictions"], "why": "dismissed"}, {"fp": "19ea8f2485a8", "file": "gptme/tools/__init__.py", "line": 542, "tokens": ["block_types", "collides", "commands", "consequence", "correctness", "disappears", "get_tool", "intended", "observable", "tool_name", "unload_tool", "unregisters"], "why": "resolved"}, {"fp": "ea2ba0e6c32c", "file": "tests/test_tools_request_tool_change.py", "line": 466, "tokens": ["incorrectly", "owner_tool", "register_command", "register_commands", "registers", "regression", "side_effect_tool", "test_disable_does_not_drop_process_global_commands", "therefore", "unload_tool", "unregistered", "vacuously"], "why": "resolved"}, {"fp": "c039c3b33ff1", "file": "gptme/tools/__init__.py", "line": 496, "tokens": ["_session_allowlist_var", "afterwards", "clear_tools", "configuration", "consequence", "downgraded", "established", "explicitly", "request_tool_change", "restrictive", "subsequently", "unrestricted"], "why": "resolved"}], "auto_resolved": ["54b74af5aab3", "8844ac559e8b", "23f7e6712cd0", "0bfd91d86f27", "8d341d6b0306", "e88167a867e0"]} -->

File: gptme/tools/request_tool_change.py (hunk 3 above)
Comment (TimeToBuildBob) [reply gptme/tools/request_tool_change.py]: Fixed in fb1798f81 — `unload_tool()` now unregisters the tool's hooks (`{name}.{hook}`) and commands before dropping it from the session set. Added `test_disable_unregisters_owned_hooks_and_commands`.

File: gptme/tools/__init__.py (hunk 4 above)
Comment (TimeToBuildBob) [reply gptme/tools/__init__.py]: Fixed in fb1798f81 — `execute_msg()` re-evaluates `is_runnable` immediately before each call, so an earlier enable in the same response can make a later sibling runnable. Every parsed tool_use still gets exactly one result. Added `test_enable_allows_later_call_in_same_response`.

File: gptme/tools/request_tool_change.py (hunk 5 above)
Comment (TimeToBuildBob) [reply gptme/tools/request_tool_change.py]: Fixed in fb1798f81 — `_enable_tool`/`_disable_tool` catch exceptions from `load_tool`/`unload_tool`, log them, and return a system error instead of aborting the response loop. Added `test_enable_initialization_error_is_reported`.

File: tests/test_tools_request_tool_change.py (hunk 6 above)
Comment (TimeToBuildBob) [reply tests/test_tools_request_tool_change.py:184]: Fixed in fb1798f81 — `_patch_available` now patches both `request_tool_change.get_available_tools` and `gptme.tools.get_available_tools`, so `load_tool()` sees the fake tools instead of the host inventory.

File: tests/test_tools_request_tool_change.py (hunk 7 above)
Comment (TimeToBuildBob) [reply tests/test_tools_request_tool_change.py:723]: Fixed in fb1798f81 — `isolated_tools` now saves and restores `get_tool_format()` so later tests do not inherit `tool` format.

File: gptme/tools/request_tool_change.py (hunk 8 above)
Comment (TimeToBuildBob) [reply gptme/tools/request_tool_change.py]: Fixed in fb1798f81 — known tools are `get_available_tools(include_mcp=False) | get_tools()`, so already-loaded custom/MCP/file-loaded tools can be disabled. Added `test_disable_loaded_custom_tool`.

**Comment by TimeToBuildBob** [comment]
@greptileai review

File: gptme/tools/__init__.py [hunk 9]
```diff
@@ -342,18 +342,19 @@ def execute_msg(
...
 
     if not classified:
         return
 
     remaining = iter(classified)
-    for tooluse, runnable in remaining:
+    for tooluse in remaining:
+        runnable = tooluse.is_runnable
         if runnable:
             with terminal_state_title(f"🛠️ running {tooluse.tool}"):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f552a3c8c — `ToolUse.execute()` now emits the paired unavailable result itself if the selected tool disappears before its execution lookup. Added a deterministic regression test for the runnability-check → unload → execution race; the focused tool suites pass (186 tests).

TimeToBuildBob:
Automated merge handoff for `f552a3c8c0f8`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/__init__.py`
- `gptme/tools/base.py`
- `gptme/tools/request_tool_change.py`

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -500,6 +501,26 @@ def set_tools(tools: list[ToolSpec]) -> None:
...
+    with _tools_init_lock:
+        tool = get_tool(tool_name)
+        if tool is None:
+            raise ValueError(f"Tool '{tool_name}' is not loaded")
+
+        from ..commands import unregister_command
+        from ..hooks import unregister_hook
+
+        for hook_name in tool.hooks:
+            unregister_hook(f"{tool.name}.{hook_name}")
```
TimeToBuildBob: ⚠️ **P2** — unload_tool() is not atomic: if any unregister_hook or unregister_command call raises an exception, the function aborts before set_tools removes the tool from the session list. This leaves the tool still loaded but with some (or all) of its hooks and commands already unregistered, producing an inconsistent state where the tool appears available but its side effects are gone. The exception propagates out of _disable_tool into execute_request_tool_change, which catches it and returns an error message, so the user sees a failure, but the tool list is now corrupted. For instance, if a hook unregistration triggers an invalid-name error, earlier hooks in the dict will have been removed while the tool remains in get_tools(), so a later disable attempt will again try to unregister the already-removed hooks. The fix is to remove the tool from the list first (or use a try/finally) and only then attempt unregistration, or to treat unregistration failures as non-fatal.

gptme/tools/request_tool_change.py
TimeToBuildBob: Fixed in 9292a3070 — `_enable_tool` catches initialization exceptions and returns a quiet system error, preserving the response loop. Covered by `test_enable_initialization_error_is_reported`.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in 9292a3070 — `unload_tool()` removes the tool from session state before cleanup and treats each hook/command cleanup failure as best-effort, so one failure cannot leave a partially cleaned tool loaded or prevent the remaining cleanup attempts. Added `test_disable_removes_tool_if_cleanup_fails`.

gptme/tools/request_tool_change.py
TimeToBuildBob: Fixed in 9292a3070 — `_enable_tool` catches initialization exceptions and returns a quiet system error, preserving the response loop. Covered by `test_enable_initialization_error_is_reported`.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in 9292a3070 — `unload_tool()` removes the tool from session state before cleanup and treats each hook/command cleanup failure as best-effort, so one failure cannot leave a partially cleaned tool loaded or prevent the remaining cleanup attempts. Added `test_disable_removes_tool_if_cleanup_fails`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Current head `9292a3070c32` is clean: all checks pass, Greptile is 5/5, and no blocking review finding remains. Self-merge is still ineligible because the PR changes core tool-runtime paths (`gptme/tools/__init__.py`, `gptme/tools/base.py`, and `gptme/tools/request_tool_change.py`), so it is waiting for maintainer review and manual merge.

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,111 @@
...
+    loaded = get_tools()
+    if any(t.name == tool_name for t in loaded):
+        return Message(
+            "system",
+            f"request_tool_change: '{tool_name}' is already enabled — no change.",
+            quiet=True,
+        )
+
+    try:
+        load_tool(tool_name)
```
TimeToBuildBob, line 62: ❌ **P1** — `_enable_tool` calls `load_tool(tool_name)` without respect to the session's allowlist or the tool's `disabled_by_default` flag. `load_tool` only checks `is_available`; it does not consult the allowlist used by `init_tools`/`get_toolchain`. Any available tool, including one explicitly excluded by `TOOL_ALLOWLIST` or marked `disabled_by_default`, can thus be enabled by the model via `request_tool_change`. This bypasses the operator's intended tool restrictions, e.g., in a multi-tenant or sandboxed setup, letting the model load a tool the admin denied. The validation step builds `known_tools` from `get_available_tools(include_mcp=False)` but that set includes all discovered tools regardless of allowlist, so it does not gate this.

gptme/tools/__init__.py
```diff
@@ -500,6 +501,40 @@ def set_tools(tools: list[ToolSpec]) -> None:
...
+    """Unload one tool and best-effort unregister its hooks and commands."""
+    with _tools_init_lock:
+        tool = get_tool(tool_name)
+        if tool is None:
+            raise ValueError(f"Tool '{tool_name}' is not loaded")
+
+        from ..commands import unregister_command
+        from ..hooks import unregister_hook
+
+        set_tools([loaded for loaded in get_tools() if loaded.name != tool_name])
```
TimeToBuildBob: ❌ **P1** — `unload_tool` finds a tool with `get_tool(tool_name)`, which matches both the tool's `name` and its `block_types`. However, the subsequent `set_tools([loaded for loaded in get_tools() if loaded.name != tool_name])` filters purely on `loaded.name`. If a caller passes a block type that is not the tool's name, `get_tool` returns the tool, but the filter does not remove it from the loaded list. The tool stays loaded while its hooks and commands are unregistered below, leaving a zombie tool that is still callable but has lost its side-effect registrations. The new `_disable_tool` only ever passes tool names, so the in-PR path is safe, but `unload_tool` is a new public function whose contract is broken for block-type arguments.

gptme/tools/request_tool_change.py
TimeToBuildBob: Fixed in 8d66bd973 — `_enable_tool` now refuses tools that are outside the session allowlist captured by `init_tools()`. `load_tool()` stays ungated because `/tools load` is the operator escape hatch (existing `test_load_tool` contract).

`disabled_by_default` expansion in an unrestricted session (allowlist is `None`) is the Phase 2 feature, not a bypass. When an allowlist is set, those tools can only be enabled if they match it.

Added `test_enable_rejects_tool_outside_allowlist` and `test_enable_allows_allowlisted_tool`.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in 8d66bd973 — `unload_tool()` now drops the matched `ToolSpec` by identity (`loaded is not tool`), so a block-type argument cannot leave a zombie tool loaded after hook/command cleanup.

Added `test_unload_by_block_type_removes_tool`.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -500,6 +510,57 @@ def set_tools(tools: list[ToolSpec]) -> None:
...
+                unregister_hook(f"{tool.name}.{hook_name}")
+            except Exception:
+                logger.exception(
+                    "Failed to unregister hook '%s.%s' while unloading tool",
+                    tool.name,
+                    hook_name,
+                )
+        for command_name in tool.commands:
+            try:
+                unregister_command(command_name)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in c928069c2 — `unload_tool()` no longer calls `unregister_command()`.

Commands stay in the process-global registry so a sibling session on gptme-server keeps its slash command. Hooks are still unregistered: they are ContextVar-backed and session-local.

Added `test_disable_does_not_drop_process_global_commands` against the real registry (the previous test mocked the unregister call and missed this).

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -500,6 +510,55 @@ def set_tools(tools: list[ToolSpec]) -> None:
...
+
+def unload_tool(tool_name: str) -> ToolSpec:
+    """Unload one tool and best-effort unregister its session-local hooks.
+
+    Commands are *not* unregistered. The slash-command registry is
+    process-global (``gptme.commands.base._command_registry``), shared by every
+    session in this process. Dropping a command here would remove that slash
+    command from sibling conversations on gptme-server even though their
+    context-local tool sets still contain the tool. Hooks are ContextVar-backed,
+    so unregistering them is session-scoped and safe.
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 62ec507d0 — `unload_tool()` still keeps the process-global command handler (sibling sessions on gptme-server must not lose `/commit`). Dispatch now checks the session-local loaded set: `handle_cmd`, listing, and completion refuse a tool-owned command when that tool is not loaded here. Covered by `test_disable_blocks_session_local_command_dispatch` and `test_tool_owned_command_blocked_when_owner_unloaded`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Current head `62ec507d0` is clean on review: Greptile is 5/5, and no blocking review finding remains. Self-merge is still ineligible because the PR changes core tool-runtime paths (`gptme/commands/base.py`, `gptme/tools/__init__.py`, `gptme/tools/base.py`, and `gptme/tools/request_tool_change.py`), so it is waiting for maintainer review and manual merge.

gptme/tools/request_tool_change.py
```diff
@@ -46,7 +152,11 @@ def execute_request_tool_change(
             quiet=True,
         )
 
-    known_tools = {tool.name for tool in get_available_tools(include_mcp=False)}
+    # Validate against available (not just loaded) tools so the model can name
+    # any tool known to gptme's module system.
+    known_tools = {t.name for t in get_available_tools(include_mcp=False)} | {
```
TimeToBuildBob: ❌ **P1** — The `known_tools` set used for validation in `execute_request_tool_change` is built from `get_available_tools(include_mcp=False)` unioned with `get_tools()`. This excludes MCP tools that are not currently loaded. As a result, if the operator's session allowlist explicitly permits an MCP tool (e.g., via a glob pattern like `mcp-server.*`), the model cannot enable it: the request is rejected with an 'unknown tool' error during validation, before `_enable_tool`'s allowlist check (which uses `get_available_tools()` and would match) is even reached. The consequence is that the adaptive tool-control mechanism cannot actuate any MCP tool, contradicting the allowlist and the PR's stated goal of enabling any allowed tool mid-session.

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,132 @@
...
+            f"request_tool_change: '{tool_name}' is already enabled — no change.",
+            quiet=True,
+        )
+
+    allowlist = get_session_allowlist()
+    if allowlist is not None:
+        available = next(
+            (t for t in get_available_tools() if t.name == tool_name), None
+        )
+        hints = available.hints if available is not None else frozenset()
```
TimeToBuildBob: ⚠️ **P2** — In _enable_tool, when the target tool is not present in get_available_tools() (e.g. a tool loaded from a file, which is not part of module discovery), `available` is None and `hints` becomes frozenset(). If the session allowlist contains a hint pattern (such as "read-only"), tool_matches_allowlist will compare the empty hint set and reject the enable, even though the tool should be permitted by that hint. This incorrectly blocks enabling a legitimate tool, and the assistant receives a misleading "not permitted by the allowlist" message instead of the enable succeeding. The fix is to fall back to the loaded tool's hints (or discover them another way) when the tool is not in available_tools.

gptme/commands/base.py
```diff
@@ -136,6 +150,21 @@ def unregister_command(name: str) -> None:
         logger.debug(f"Unregistered command: {name}")
     if name in _command_completers:
         del _command_completers[name]
+    _command_owners.pop(name, None)
```
TimeToBuildBob, line 158: ⚠️ **P2** — In unload_tool, hooks are unregistered for the removed tool, but its slash commands are intentionally left in the global registry. The _command_owners entries for those commands also remain. If the same command name is later registered by a different tool (or by the same tool after a re-load), the stale owner entry from the first tool can still gate dispatch. For example, tool A registers "/foo" with owner "A"; A is unloaded; tool B registers "/foo" with owner "B". In a session where B is loaded but A is not, _command_enabled_in_session("foo") checks has_tool("A") which is false, so "/foo" is blocked even though B is active and owns it now. While this is an edge case, it is a real API-contract violation: the owner mapping is not updated atomically with command re-registration, leading to commands being incorrectly disabled.

gptme/tools/request_tool_change.py
TimeToBuildBob: Fixed in 53b1e0ac3 — `known_tools` now uses `get_available_tools()` (MCP included by default) union `get_tools()`. An allowlisted MCP tool that is not yet loaded is no longer rejected as unknown before the allowlist check.

Regression: `test_enable_allows_unloaded_mcp_tool` fails if `include_mcp=False` is reintroduced.

gptme/tools/request_tool_change.py
TimeToBuildBob: Not a reachable allowlist bug.

`_enable_tool` returns immediately when the tool is already in `get_tools()`, so the allowlist/hints lookup never sees a loaded file-tool. An unloaded file-loaded tool is also absent from `get_available_tools()` (module + MCP discovery only), so it fails the `known_tools` check as "unknown tool" and never reaches `tool_matches_allowlist`.

The empty-hints path would require a name that is in `known_tools` but not in `get_available_tools()` and not currently loaded — those two sets are now the same function (`get_available_tools()` with MCP included, union loaded). Hint matching therefore only runs when a spec is found; `test_enable_honors_hint_allowlist` covers that.

Re-enabling a disabled file-loaded tool is a separate discovery gap, not a misleading allowlist refusal.

gptme/commands/base.py
TimeToBuildBob: False positive: re-registration already updates ownership atomically with the handler.

`register_command` always overwrites `_command_owners` in the same call that replaces the registry entry:

```python
if owner_tool is not None:
    for cmd_name in names:
        _command_owners[cmd_name] = owner_tool
```

Tool A unloads (commands stay process-global by design); tool B then `register_command("foo", handler, owner_tool="B")` sets owner to B before any dispatch. The described A-then-B gate on `has_tool("A")` does not happen.

Leaving owner entries after unload is intentional: the slash-command registry is process-global so a disable in this session cannot yank `/commit` from sibling gptme-server conversations. Session isolation is at dispatch (`_command_enabled_in_session` → `has_tool(owner)`), which is what `62ec507d0` added.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI red on this head is the known no-extras `TestLogWarnOnce.test_get_model_unknown_warns_once` flake (`assert 2 == 1`), not a #3656 regression. The PR does not touch `tests/test_llm_models_resolution.py` or `gptme/llm/models/resolution.py`; every other check on `53b1e0ac` passed.

Re-ran failed job: https://github.com/gptme/gptme/actions/runs/33264712046 (attempt 2). No code change. Self-merge still ineligible (core tool-runtime paths) — waiting on maintainer review after CI settles.

TimeToBuildBob:
Automated merge handoff for `53b1e0ac371f`:

- CI gate: satisfied
- `gptme/commands/base.py`
- `gptme/tools/__init__.py`
- `gptme/tools/base.py`
- `gptme/tools/request_tool_change.py`

gptme/commands/base.py
```diff
@@ -136,6 +150,21 @@ def unregister_command(name: str) -> None:
...
+
+    Unowned commands (built-ins, plugins) are always enabled. Tool-owned
+    commands require their owner to be in the session-local loaded set.
+    """
+    owner = _command_owners.get(name)
+    if owner is None:
+        return True
+    from ..tools import has_tool  # fmt: skip
+
+    return has_tool(owner)
```
TimeToBuildBob, line 172: ❌ **P1** — The new dispatch gating consults `gptme.tools.has_tool(owner)`, which reads the session-local `_loaded_tools_var` ContextVar — populated only when `init_tools()` has run in that exact context. On gptme-server, request handling typically executes in fresh contexts that may not have had `init_tools()` invoked (the module docstring explicitly warns 'Tools must be initialized in each context that needs them'). In such a context, every tool-owned slash command evaluates `has_tool(owner)` to False, so `handle_cmd` blocks all dynamically registered commands and `get_user_commands` hides them, even though the owning tool is loaded in the conversation's own session. This silently breaks every tool-provided command on the server, a regression from the previous behavior where registered commands ran regardless of context.

gptme/tools/__init__.py
```diff
@@ -500,6 +510,59 @@ def set_tools(tools: list[ToolSpec]) -> None:
...
+
+    Session isolation for commands is enforced at dispatch: ``handle_cmd``
+    refuses to run a tool-owned command when that tool is not in this
+    session's loaded set.
+    """
+    with _tools_init_lock:
+        tool = get_tool(tool_name)
```
TimeToBuildBob, line 542: ⚠️ **P2** — `unload_tool` resolves the tool via `get_tool(tool_name)`, which checks tool names first and only then block types. If a tool's block type collides with another loaded tool's name, `unload_tool` will unload the wrong tool. For example, tool A has name 'a' and block_types ['b'], and tool B has name 'b'. Calling `unload_tool('b')` matches tool B by name, unloads it, and unregisters B's hooks while leaving tool A (the one the caller likely intended to remove) still loaded. The observable consequence is that the wrong tool disappears from the session, and its commands/hooks are removed, while the tool the caller targeted by block type remains active. This is a correctness bug in the new public `unload_tool` API.

tests/test_tools_request_tool_change.py
```diff
@@ -122,3 +134,608 @@ def test_request_tool_change_rejects_non_string_fields():
...
+            get_registered_commands,
+            register_command,
+            unregister_command,
+        )
+
+        def handler(_ctx):
+            return
+            yield  # pragma: no cover — CommandHandler is a generator type
+
+        register_command("side-effect", handler)
```
TimeToBuildBob: ⚠️ **P2** — The test `test_disable_does_not_drop_process_global_commands` registers the command with `register_command("side-effect", handler)` without an `owner_tool`, so the command is unowned and would survive even if `unload_tool` incorrectly unregistered commands owned by the tool. The test therefore passes vacuously and does not protect the intended regression: it should use `side_effect_tool.register_commands()` to register the command with the owning tool. As written, it would still pass if `unload_tool` removed all commands whose owner matches the unloaded tool, because the command in this test has no owner.

gptme/tools/__init__.py
```diff
@@ -484,6 +493,7 @@ def clear_tools():
     """
     _set_available_tools_cache(None)
     _loaded_tools_var.set([])
+    _session_allowlist_var.set(None)
```
TimeToBuildBob, line 496: ⚠️ **P2** — `clear_tools()` resets `_session_allowlist_var` to `None`, meaning any session that calls `clear_tools` afterwards loses the operator-set tool allowlist and becomes unrestricted. If a session is started with a restrictive allowlist and some command or code path invokes `clear_tools` (e.g. a `/tools clear` or a reset routine), the assistant can subsequently use `request_tool_change` to enable any tool, bypassing the intended hard cap. This undermines the security boundary established by `init_tools` setting the allowlist. The observable consequence is that a restrictive configuration can be silently downgraded to unrestricted, allowing the model to enable tools the operator explicitly excluded.

TimeToBuildBob (commented):

**5 findings** · ❌ **1** P1 · ⚠️ **4** P2

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,130 @@
...
+_SELF_NAME = "request_tool_change"
+
+
+def _enable_tool(tool_name: str) -> Message:
+    """Add *tool_name* to the session's active tool set.
+
+    Returns a system message describing the outcome (success or no-op).
+    """
+    loaded = get_tools()
+    if any(t.name == tool_name for t in loaded):
```
TimeToBuildBob, line 42: ⚠️ **P2** — In `_enable_tool`, the early return for an already-loaded tool occurs before the `get_session_allowlist()` check, so a tool already present in the session's loaded set (e.g., loaded via `set_tools()` or a plugin that bypasses `init_tools`) is considered 'already enabled' and never validated against the operator's allowlist. Since the allowlist is intended as a hard cap for model enablement via `request_tool_change`, an already-loaded tool escapes that cap, allowing the assistant to use it even when the operator intended to exclude it. For example, if an operator sets an allowlist excluding `shell`, but a plugin loads `shell` directly into the session's tool set, the assistant can invoke `shell` without the allowlist being enforced. The observable consequence is that the operator's tool restrictions are silently bypassed for any tool that happens to be loaded outside the allowlist, weakening the security boundary in multi-tenant or sandboxed deployments.

gptme/commands/base.py
TimeToBuildBob: False positive — not a server regression.

`_command_enabled_in_session` is fail-closed on purpose: a tool-owned slash command may run only when that tool is in *this* context's loaded set. An uninitialized context has an empty loaded set (`_get_loaded_tools()` defaults to `[]`), so `has_tool(owner)` is False. That is the same as "the tool is not enabled here," which is the contract this PR added.

The server path already initializes tools in the same request/thread *before* `handle_cmd`:

- `gptme/server/api_v2.py` message-add: `init_tools(tool_allowlist)` then `handle_cmd(...)`
- conversation config PATCH: `init_tools(chat_config.tools)` in the same request
- TUI copies the initialized context via `contextvars.copy_context()` and runs `execute_cmd` inside it
- CLI calls `init_tools` at session start

There is no remaining command path that dispatches a tool-owned slash command without first loading tools in that context. Fail-open (`has_tool` True when tools were never initialized) would re-break session isolation: a disable in this session would still run the process-global handler.

The "how this was verified" note is a different (already-dismissed) issue about stale `_command_owners` on re-register without `owner_tool`. `register_command` only writes owners when `owner_tool is not None`; `unregister_command` already pops the mapping. That is not this finding.

No code change.

gptme/tools/__init__.py
TimeToBuildBob: Accepted trade-off — this is `get_tool()`'s existing contract, not a new unload bug.

`get_tool(name)` matches loaded tool *names* first, then block types. `unload_tool` uses that same lookup so disable/unload agree with every other caller (`ToolUse.execute`, `/tools`, export). Name-vs-block-type collisions are possible in theory (tool A block_types `['b']`, tool B named `'b'`) and would unload B — which is what `get_tool('b')` already returns everywhere else.

The in-PR path (`_disable_tool`) only ever passes tool names, never block types. Giving `unload_tool` a different lookup than `get_tool` would be the worse API split.

No code change.

gptme/tools/__init__.py
TimeToBuildBob: Accepted trade-off — `clear_tools()` is a full context reset, not an allowlist-preserving pause.

Production callers:

- `reload_config()` (`gptme/config/core.py`) — rebuilds config then expects the next `init_tools()` to recapture the allowlist
- subagent thread start (`gptme/tools/subagent/execution.py`) — the child *must* detach from the parent's loaded set *and* allowlist; the child then loads its own tools
- tests — fixture isolation

Leaving a stale operator allowlist on an empty loaded set after `clear_tools` would leak the parent's restriction into a subagent (or the next conversation on a reused thread) until `init_tools` ran. Resetting to `None` matches the empty loaded list. `request_tool_change` enablement is gated by `get_session_allowlist()`, which `init_tools()` always overwrites.

No code change.

gptme/tools/request_tool_change.py
TimeToBuildBob: Rejected — the early return is a no-op on an already-loaded tool; it is not an allowlist bypass.

`_enable_tool` first reports "already enabled" if the tool is in `get_tools()`. The operator allowlist is a hard cap on *model enablement via `request_tool_change`*, not a second filter on tools that are already in the session. If `shell` is loaded, the assistant can already invoke it; refusing `enable_tool` for `shell` would not unload it.

The scenario (plugin `set_tools()` of a tool the operator excluded) is a load-path issue, not an enable-path issue. `load_tool()` is intentionally ungated so `/tools load` remains the operator escape hatch. Closing that would need a change to `set_tools`/`load_tool`, which is out of Phase 2 scope.

If the tool is *not* already loaded, the allowlist check still runs before `load_tool()`.

No code change.

tests/test_tools_request_tool_change.py
TimeToBuildBob: Fixed in 72dbcbf58 — `test_disable_does_not_drop_process_global_commands` now registers via `ToolSpec.register_commands()`, so the command is owned. An unowned `register_command()` would have survived even if `unload_tool` started unregistering owned commands.

Follow-up in f8e31fbf7: `register_command()` now pops `_command_owners` when `owner_tool` is None, so an unowned re-register cannot keep gating on a stale tool.

TimeToBuildBob:
Automated merge handoff for `f8e31fbf7d89`:

- CI gate: satisfied
- Review threads: clear
- `gptme/commands/base.py`
- `gptme/tools/__init__.py`
- `gptme/tools/base.py`
- `gptme/tools/request_tool_change.py`

gptme/tools/request_tool_change.py
```diff
@@ -1,26 +1,130 @@
...
+    allowlist = get_session_allowlist()
+    if allowlist is not None:
+        spec = next((t for t in get_available_tools() if t.name == tool_name), None)
+        hints = spec.hints if spec is not None else frozenset()
+        if not tool_matches_allowlist(tool_name, allowlist, hints):
+            return Message(
+                "system",
+                f"request_tool_change: '{tool_name}' is not permitted by "
+                "the session tool allowlist — no change.",
+                quiet=True,
```
TimeToBuildBob, line 58: ⚠️ **P2** — `has_tool()` in gptme/tools/__init__.py checks both the tool `name` and its block types. However, the pre‑check in `_disable_tool()` (gptme/tools/request_tool_change.py) uses `any(t.name == tool_name for t in get_tools())` which only matches by `name`. If a caller passes a block type (e.g., `bash` for a tool with name `shell`), the pre‑check reports that the tool is “not currently enabled” and the disable operation is skipped, even though `unload_tool()` would correctly find and remove it. While the input validation (`known_tools`) currently rejects block types, this is a latent inconsistency and a future regression risk.

TimeToBuildBob (commented):

**1 finding** · ⚠️ **1** P2

gptme/tools/request_tool_change.py
TimeToBuildBob: Rejected — the premise is wrong, and the suggested change is a no-op.

`has_tool()` does **not** check block types. On this head it is:

```python
def has_tool(tool_name: str) -> bool:
    return any(tool.name == tool_name for tool in _get_loaded_tools())
```

That is the same name-only test `_disable_tool` already uses (`any(t.name == tool_name for t in get_tools())`). `get_tool()` is the helper that matches name then block type; the finding mixed the two up.

The suggested `if not has_tool(tool_name)` rewrite would not change behavior.

Even if we wanted block-type disable, `execute_request_tool_change` builds `known_tools` from `{t.name ...}` only, so a block type such as `bash` is rejected as `unknown tool` before `_disable_tool` runs. The `tool_name` parameter is documented as a tool name, not a block-type alias.

Name-vs-block-type lookup for `unload_tool` was already accepted as `get_tool()`'s existing contract on fp `19ea8f2485a8`. No code change.

TimeToBuildBob:
Looked into the remaining AI-review finding on this head (`f8e31fbf7`): P2 `bdd9d78167d5` (`_disable_tool` name-only pre-check vs `has_tool`) is already disposed as a false premise — `has_tool()` is name-only on this head, so the suggested rewrite is a no-op. No open finding threads remain (`ai-review-dispose --list` is empty). Greptile is 5/5, CI green.

No additional code change. Self-merge still ineligible (core tool-runtime paths). Waiting for maintainer review.

PR #3660, opened 2026-08-28
fix(dirs): treat empty/whitespace env vars as unset; guard profile_name

### LINKED ISSUE #3642: fix(dirs): handle empty env vars and harden path handling
### Description

In `gptme/dirs.py`, env vars like `XDG_DATA_HOME`, `XDG_STATE_HOME`, `GPTME_LOGS_HOME`, `GPTME_WORKSPACE` are checked via `if \"VAR\" in os.environ` or `os.environ.get()` without handling empty string case.

### Bug

```python
if "XDG_DATA_HOME" in os.environ:
    return Path(os.environ["XDG_DATA_HOME"]) / "gptme"

# If XDG_DATA_HOME="" (empty, common in Docker/misconfig):
# Path("") / "gptme" -> "gptme" (relative path, wrong location, could write to CWD)
```

### Proposed fix

Helper that treats empty/whitespace as unset:

```python
def _get_env_path(var: str) -> Path | None:
    val = os.environ.get(var, "").strip()
    return Path(val) if val else None
```

And apply to all env var path handling.

Also:
- Validate `profile_name` in `get_profile_memory_dir` to prevent path traversal
- Handle empty git output in `_get_project_git_dir_call`
- Wrap mkdir in try/except for permission errors

### Reproduction

```bash
XDG_DATA_HOME="" gptme --version
# Currently would try to use relative "gptme" dir instead of default
```

### PR

I have a fix ready (branch `fix/dirs-handle-empty-env-vars`). Can open PR if this issue is accepted.

### Checklist

- [x] Searched existing issues
- [x] Single focused fix
- [x] Includes edge case handling

TimeToBuildBob:
## Summary

Fixes #3642 (reported by @app/arena-ai-coding-agent).

Empty XDG_* env vars (common in Docker containers or misconfigured shells) produced relative paths:

```python
# Before: XDG_DATA_HOME= → Path("") / "gptme" = Path("gptme")  # relative!
# After:  falls through to platformdirs absolute default
```

## Changes

- Add `_get_env_path(var)` helper that returns `None` for empty/whitespace values
- Apply to `get_data_dir`, `get_state_dir`, `get_logs_dir` (replaces bare `"VAR" in os.environ`)
- Guard `get_profile_memory_dir` against path-traversal names (raises `ValueError` on `../ `, `/abs`, etc.)
- `get_workspace` already handled empty via walrus operator — no change needed

## Tests

- `TestGetEnvPath`: set/unset/empty/whitespace/strip cases
- `test_xdg_data_home_empty_falls_through`, `test_xdg_data_home_whitespace_falls_through`: confirm absolute path returned
- `test_gptme_logs_home_empty_falls_through`: same for logs
- `test_path_traversal_rejected`: `../etc/passwd`, `/abs/path`, `a/b` all raise `ValueError`
- `test_valid_profile_names`: alphanumeric, hyphens, underscores, dots pass

52 tests pass.

### DIFF
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -18,9 +18,14 @@ def _get_env_path(var: str) -> str | None:
     Treats an empty or whitespace-only value the same as "not set" —
     this is intentional: empty XDG_* vars (common in Docker or misconfigured
     environments) must not produce relative paths like ``Path("") / "gptme"``.
+
+    Surrounding whitespace is stripped. ``Path(" /tmp/foo")`` is relative
+    (it does not start with ``/``), so a leading space on ``XDG_DATA_HOME``
+    would otherwise write data under CWD. Accidental padding is the
+    misconfiguration this helper exists to tolerate.
     """
-    val = os.environ.get(var, "")
-    return val if val.strip() else None
+    stripped = os.environ.get(var, "").strip()
+    return stripped or None
 
 
 def get_config_dir() -> Path:
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import re
 import shutil
 import struct
 import subprocess
@@ -12,9 +11,6 @@ from .util.git_cmd import GIT_CMD
 
 logger = logging.getLogger(__name__)
 
-#: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
-_SAFE_PROFILE_NAME_RE = re.compile(r"(?!\.{1,2}\Z)[A-Za-z0-9_.-]+\Z")
-
 
 def _get_env_path(var: str) -> str | None:
     """Return the env-var value, or None if unset or empty/whitespace.
@@ -23,8 +19,8 @@ def _get_env_path(var: str) -> str | None:
     this is intentional: empty XDG_* vars (common in Docker or misconfigured
     environments) must not produce relative paths like ``Path("") / "gptme"``.
     """
-    val = os.environ.get(var, "").strip()
-    return val or None
+    val = os.environ.get(var, "")
+    return val if val.strip() else None
 
 
 def get_config_dir() -> Path:
@@ -175,18 +171,26 @@ def get_profile_memory_dir(profile_name: str) -> Path:
 
     Args:
         profile_name: Name of the agent profile (e.g. 'explorer', 'researcher').
-            Must match ``^[A-Za-z0-9_\\-.]+$`` to prevent path traversal.
+            Must be a non-empty path component other than ``.`` or ``..``.
 
     Returns:
         Path to the memory directory (created if it doesn't exist)
 
     Raises:
-        ValueError: If ``profile_name`` contains path-traversal characters.
+        ValueError: If ``profile_name`` is not a safe single path component.
     """
-    if not _SAFE_PROFILE_NAME_RE.fullmatch(profile_name):
+    if (
+        not profile_name
+        or profile_name in {".", ".."}
+        or Path(profile_name).name != profile_name
+        or "/" in profile_name
+        or "\\" in profile_name
+        or "\n" in profile_name
+        or "\r" in profile_name
+        or "\0" in profile_name
+    ):
         raise ValueError(
-            f"Invalid profile name {profile_name!r}: "
-            "must contain only alphanumeric characters, hyphens, underscores, or dots."
+            f"Invalid profile name {profile_name!r}: must be a safe path component."
         )
     path = get_data_dir() / "memories" / "profiles" / profile_name
     path.mkdir(parents=True, exist_ok=True)
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -13,7 +13,7 @@ from .util.git_cmd import GIT_CMD
 logger = logging.getLogger(__name__)
 
 #: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
-_SAFE_PROFILE_NAME_RE = re.compile(r"^(?!\.{1,2}$)[A-Za-z0-9_\-\.]+$")
+_SAFE_PROFILE_NAME_RE = re.compile(r"(?!\.{1,2}\Z)[A-Za-z0-9_.-]+\Z")
 
 
 def _get_env_path(var: str) -> str | None:
@@ -133,7 +133,7 @@ def get_workspace() -> Path:
     Handles git submodules: if `.git` is a file (not a directory),
     we're in a submodule and the parent repo root is returned instead.
     """
-    if workspace := os.environ.get("GPTME_WORKSPACE"):
+    if workspace := _get_env_path("GPTME_WORKSPACE"):
         return Path(workspace)
 
     try:
@@ -183,7 +183,7 @@ def get_profile_memory_dir(profile_name: str) -> Path:
     Raises:
         ValueError: If ``profile_name`` contains path-traversal characters.
     """
-    if not _SAFE_PROFILE_NAME_RE.match(profile_name):
+    if not _SAFE_PROFILE_NAME_RE.fullmatch(profile_name):
         raise ValueError(
             f"Invalid profile name {profile_name!r}: "
             "must contain only alphanumeric characters, hyphens, underscores, or dots."
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -13,7 +13,7 @@ from .util.git_cmd import GIT_CMD
 logger = logging.getLogger(__name__)
 
 #: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
-_SAFE_PROFILE_NAME_RE = re.compile(r"^[A-Za-z0-9_\-\.]+$")
+_SAFE_PROFILE_NAME_RE = re.compile(r"^(?!\.{1,2}$)[A-Za-z0-9_\-\.]+$")
 
 
 def _get_env_path(var: str) -> str | None:
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -182,12 +182,12 @@ def get_profile_memory_dir(profile_name: str) -> Path:
     if (
         not profile_name
         or profile_name in {".", ".."}
-        or Path(profile_name).name != profile_name
         or "/" in profile_name
         or "\\" in profile_name
         or "\n" in profile_name
         or "\r" in profile_name
         or "\0" in profile_name
+        or Path(profile_name).name != profile_name
     ):
         raise ValueError(
             f"Invalid profile name {profile_name!r}: must be a safe path component."
--- a/gptme/dirs.py
+++ b/gptme/dirs.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import re
 import shutil
 import struct
 import subprocess
@@ -10,6 +11,20 @@ from platformdirs import user_config_dir, user_data_dir, user_state_dir
 from .util.git_cmd import GIT_CMD
 
 logger = logging.getLogger(__name__)
+
+#: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
+_SAFE_PROFILE_NAME_RE = re.compile(r"^[A-Za-z0-9_\-\.]+$")
+
+
+def _get_env_path(var: str) -> str | None:
+    """Return the env-var value, or None if unset or empty/whitespace.
+
+    Treats an empty or whitespace-only value the same as "not set" —
+    this is intentional: empty XDG_* vars (common in Docker or misconfigured
+    environments) must not produce relative paths like ``Path("") / "gptme"``.
+    """
+    val = os.environ.get(var, "").strip()
+    return val or None
 
 
 def get_config_dir() -> Path:
@@ -26,8 +41,8 @@ def get_pt_history_file() -> Path:
 
 def get_data_dir() -> Path:
     # used in testing, so must take precedence
-    if "XDG_DATA_HOME" in os.environ:
-        return Path(os.environ["XDG_DATA_HOME"]) / "gptme"
+    if xdg := _get_env_path("XDG_DATA_HOME"):
+        return Path(xdg) / "gptme"
 
     # just a workaround for me personally
     old = Path("~/.local/share/gptme").expanduser()
@@ -43,15 +58,15 @@ def get_state_dir() -> Path:
     Used for recovery artifacts and other state that should persist between
     invocations but is not important enough to back up — checkpoints, etc.
     """
-    if "XDG_STATE_HOME" in os.environ:
-        return Path(os.environ["XDG_STATE_HOME"]) / "gptme"
+    if xdg := _get_env_path("XDG_STATE_HOME"):
+        return Path(xdg) / "gptme"
     return Path(user_state_dir("gptme"))
 
 
 def get_logs_dir() -> Path:
     """Get the path for **conversation logs** (not to be confused with the logger file)"""
-    if "GPTME_LOGS_HOME" in os.environ:
-        path = Path(os.environ["GPTME_LOGS_HOME"])
+    if logs_home := _get_env_path("GPTME_LOGS_HOME"):
+        path = Path(logs_home)
     else:
         path = get_data_dir() / "logs"
     path.mkdir(parents=True, exist_ok=True)
@@ -111,7 +126,7 @@ def get_workspace() -> Path:
     """Get the agent workspace directory.
 
     Detection order:
-    1. GPTME_WORKSPACE environment variable
+    1. GPTME_WORKSPACE environment variable (ignored if empty)
     2. Git root, traversing to parent repo if in a submodule
     3. Current working directory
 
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

The diff is overall safe. The main change is the new `_get_env_path` helper for environment variables, which corrects the issue of empty/whitespace-only values creating relative paths. The path traversal protection added to `get_profile_memory_dir` is solid and covers the common vectors. One minor concern is the whitespace-stripping behavior for environment variables, which could be a breaking change for unusual path names.

### Confidence Score: N/A

Not scored — this diff moves a submodule pointer and contains no source to assess. Treat it as *not reviewed*, not as approved.

<details><summary><b>2 out-of-scope findings (outside this PR's changed files)</b></summary>

**❌ P1 high · 🔒 security** — `dirs.py (profile name validation)`

profile_name = ".. " (dot dot space) passes all checks: it is not exactly "..", it contains no separator, no newline, no null, and `Path(".. ").name == ".. "` (the trailing space is kept by pathlib).  On Windows, when the OS resolves that path, the trailing space is stripped and the component becomes "..", so `data_dir / "profiles" / ".. "` resolves to `data_dir`.  The application would then perform operations in the wrong directory.  A profile name supplied by the user can therefore read/write outside the intended profile folder.

```suggestion
Before validating, strip trailing spaces and dots from the profile name with `profile_name.rstrip(' .')`.  Then reject if the stripped name is empty, '.' or '..'.  Also consider rejecting any profile name that, after applying Windows filename trimming, is '.' or '..'.  On POSIX trailing spaces are kept, but the same trimming is harmless.  Alternatively, normalize the final path component with `os.path.normpath` and ensure its parent stays the intended directory.
```

<sub>**How this was verified:** The new tests in the diff do not include a trailing-space variant.  On Windows this bypass would be active.</sub>

**⚠️ P2 medium** — [`src/module.py:1`](https://github.com/gptme/gptme/blob/e5911fb118f55ca6211de93d3a0898bc88f5647c/src/module.py#L1)

The `_get_env_path` helper strips leading and trailing whitespace from environment variables. While this fixes the issue of whitespace-only values being treated as paths, it also removes intentional surrounding spaces from a valid path (e.g., a directory named `/foo `). This is an unintended behavior change for cases where a user intentionally has whitespace at the start or end of a path. The risk is low because such paths are unusual, but the change may silently redirect data to a different directory.

```suggestion
Consider trimming only leading/trailing whitespace when determining if the value is empty, but still keep the original value for the path construction. For example, use `if value and value.strip():` and pass the original value to `Path()`. Alternatively, document this behavior explicitly if it is intended.
```

<sub>**How this was verified:** A simple test: set `XDG_DATA_HOME` to a string with a trailing space and a valid directory path. Before the change, the path would be used as-is; after the change, the trailing space is stripped, potentially targeting a different directory.</sub>

</details>

| `69533263cdc1` | 5/5 | 0 | llm | 2026-08-28 05:07 UTC |
| `35a99a2cd931` | 3/5 | 2 | llm | 2026-08-28 13:50 UTC |
| `d82a2c757718` | 4/5 | 2 | llm | 2026-08-28 14:52 UTC |
| `47d0b3e2e653` | 3/5 | 1 | llm | 2026-08-29 02:29 UTC |
| `4d30aa732584` | 3/5 | 3 | llm | 2026-08-29 16:45 UTC |

</details>

<sub>Reviewed `e5911fb118f5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1225s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

ErikBjare:
@TimeToBuildBob fix

gptme/dirs.py
```diff
@@ -10,6 +11,20 @@
 
 logger = logging.getLogger(__name__)
 
+#: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
+_SAFE_PROFILE_NAME_RE = re.compile(r"^[A-Za-z0-9_\-\.]+$")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 35a99a2cd: the profile-name allowlist now rejects the special path components `.` and `..`, with regression coverage for both values.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/dirs.py
```diff
@@ -10,6 +11,20 @@
 
 logger = logging.getLogger(__name__)
 
+#: Only allow alphanumeric, hyphen, underscore, and dot in profile names.
+_SAFE_PROFILE_NAME_RE = re.compile(r"^(?!\.{1,2}$)[A-Za-z0-9_\-\.]+$")
```
TimeToBuildBob: ⚠️ **P2** — The _SAFE_PROFILE_NAME_RE regex uses $ anchors, which in Python match before a trailing newline. A profile name like "test\n" is therefore accepted, creating a directory with a literal newline in its name. This can break downstream code that assumes profile names are simple strings and may cause issues in logging or file listing. Replace $ with \Z to anchor to the absolute end of the string.

TimeToBuildBob:
@greptileai review

gptme/dirs.py
TimeToBuildBob: Fixed in d82a2c757: the profile guard now uses an absolute-end anchor with `fullmatch`, and `agent\n` is covered as a rejected regression case.

TimeToBuildBob:
Fixed both review findings in d82a2c757: whitespace-only `GPTME_WORKSPACE` now falls through via `_get_env_path`, and profile names use absolute-end matching so trailing newlines are rejected. Added regression tests; `tests/test_dirs.py` passes (53 tests), ruff and pre-commit pass. CI is running on the pushed head.

gptme/dirs.py
TimeToBuildBob: Fixed in d82a2c757 with absolute-end matching and a regression test rejecting profile names ending in a newline.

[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d82a2c757: the current head uses `\Z` plus `fullmatch`, and the new `agent\n` regression case passes.

TimeToBuildBob:
@greptileai review

gptme/dirs.py
```diff
@@ -158,11 +173,20 @@ def get_profile_memory_dir(profile_name: str) -> Path:
...
+        profile_name: Name of the agent profile (e.g. 'explorer', 'researcher').
+            Must match ``^[A-Za-z0-9_\\-.]+$`` to prevent path traversal.
 
     Returns:
         Path to the memory directory (created if it doesn't exist)
+
+    Raises:
+        ValueError: If ``profile_name`` contains path-traversal characters.
     """
+    if not _SAFE_PROFILE_NAME_RE.fullmatch(profile_name):
```
TimeToBuildBob: ⚠️ **P2** — The new `_SAFE_PROFILE_NAME_RE` validation in `get_profile_memory_dir` (line 185) rejects any profile name containing a space or non-ASCII character, which were previously accepted without restriction. For example, the name `"my profile"` would have produced the path `get_data_dir()/"memories"/"profiles"/"my profile"` and worked, but now `fullmatch()` fails and the function raises `ValueError`. This is a breaking change for any existing caller or configuration that uses spaces in profile names; there is no migration path or deprecation warning. The PR's stated goal is to prevent path traversal, but disallowing spaces and Unicode letters goes beyond that and breaks valid use cases.

gptme/dirs.py
```diff
@@ -10,6 +11,20 @@
...
+
+
+def _get_env_path(var: str) -> str | None:
+    """Return the env-var value, or None if unset or empty/whitespace.
+
+    Treats an empty or whitespace-only value the same as "not set" —
+    this is intentional: empty XDG_* vars (common in Docker or misconfigured
+    environments) must not produce relative paths like ``Path("") / "gptme"``.
+    """
+    val = os.environ.get(var, "").strip()
```
TimeToBuildBob: ⚠️ **P2** — `_get_env_path` (line 25) strips surrounding whitespace from every environment variable value via `os.environ.get(var, "").strip()`, not merely checking whether the value is empty after stripping. So an env var set to `" /path "` is returned as `"/path"`, changing the path. Before this PR, `get_data_dir` used `Path(os.environ["XDG_DATA_HOME"])` which preserved spaces, so `XDG_DATA_HOME=" /home/user/data "` pointed to a directory literally named with leading/trailing spaces. After the change, that same value points to `/home/user/data`, so the program reads/writes a different directory, potentially losing access to existing data or writing to an unintended location. The PR description only mentions treating empty/whitespace-only values as unset; silently stripping non-whitespace values is an unannounced behavior change.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed 47d0b3e2e after investigating the failed no-extras job. The failure was an unrelated teardown race in `tests/test_subagent_unit.py`; a job rerun was still pending when I pushed the review fixes, so the new head now has a fresh CI run.

Also addressed the two open compatibility findings: nonblank env values retain surrounding whitespace, and profile validation now rejects unsafe path components/control characters without rejecting spaces or Unicode. `tests/test_dirs.py`: 53 passed; ruff, mypy, and pre-commit pass. CI is running on the pushed head.

gptme/dirs.py
TimeToBuildBob: Fixed in 47d0b3e2e: validation now rejects only unsafe single-component names and control characters; spaces and Unicode remain accepted. Regression cases cover `my profile` and `utf8-探索`.

gptme/dirs.py
TimeToBuildBob: Fixed in 47d0b3e2e: `_get_env_path` uses stripping only to detect blank values and returns the original nonblank value unchanged. A regression test preserves surrounding whitespace.

TimeToBuildBob:
The only failing check on `47d0b3e2e` is the noisy Startup Benchmark: cold startup measured `6.406s ± 2.965s` against a `6.0s` threshold, while warm startup, `--help`, and the full cycle all passed. This PR only changes directory resolution and its tests, and recent master benchmark runs are green, so I reran the failed benchmark job (run 33197464168). It is queued now; no code change made for this variance.

TimeToBuildBob:
Automated merge handoff for `47d0b3e2e653`:

- CI gate: satisfied
- Review threads: clear
- `gptme/dirs.py`

gptme/dirs.py
```diff
@@ -158,11 +169,28 @@ def get_profile_memory_dir(profile_name: str) -> Path:
...
     Returns:
         Path to the memory directory (created if it doesn't exist)
+
+    Raises:
+        ValueError: If ``profile_name`` is not a safe single path component.
     """
+    if (
+        not profile_name
+        or profile_name in {".", ".."}
+        or Path(profile_name).name != profile_name
```
TimeToBuildBob: ❌ **P1** — In `get_profile_memory_dir`, the validation condition constructs `Path(profile_name)` on line 184 as part of `or Path(profile_name).name != profile_name`. For a profile name containing a null byte (e.g. `"agent\0name"`), `Path` construction itself raises `ValueError("embedded null byte")` before the explicit `"\0" in profile_name` check on line 189 is ever evaluated. The test `test_path_traversal_rejected` expects every bad name to raise `ValueError` matching `"Invalid profile name"`, but the null-byte case raises a different message, so the test fails with a mismatch. The condition order also means users get an unhelpful error instead of the intended message. Reorder the checks so the null-byte (and other explicit character) checks happen before constructing `Path`, or wrap the construction to re-raise with a consistent message.

TimeToBuildBob:
Automated merge handoff for `47d0b3e2e653`:

- CI gate: satisfied
- Review threads: clear
- `gptme/dirs.py`

gptme/dirs.py
TimeToBuildBob: Fixed in 4d30aa732 — explicit character checks (`/`, `\`, `\n`, `\r`, `\0`) now run before `Path(profile_name)` is constructed, so a null-byte name raises the intended `Invalid profile name` error. `tests/test_dirs.py` still passes (53 tests).

TimeToBuildBob:
Addressed the open P1 on `get_profile_memory_dir`: control-character checks now run before `Path()` construction so null-byte names raise `Invalid profile name`. Pushed `4d30aa732`. CI is running on the new head; in-band review was skipped after an OpenRouter timeout, so the GitHub reviewer will re-check this SHA.

TimeToBuildBob:
Automated merge handoff for `4d30aa732584`:

- CI gate: satisfied
- `gptme/dirs.py`

gptme/dirs.py
```diff
@@ -11,6 +11,17 @@
...
 
+def _get_env_path(var: str) -> str | None:
+    """Return the env-var value, or None if unset or empty/whitespace.
+
+    Treats an empty or whitespace-only value the same as "not set" —
+    this is intentional: empty XDG_* vars (common in Docker or misconfigured
+    environments) must not produce relative paths like ``Path("") / "gptme"``.
+    """
+    val = os.environ.get(var, "")
+    return val if val.strip() else None
```
TimeToBuildBob: ❌ **P1** — The new `_get_env_path` helper returns the environment variable unchanged whenever its stripped form is non-empty. A value like `' /tmp/foo'` (leading space) has `strip()` = `'/tmp/foo'` (non-empty), so the helper returns `' /tmp/foo'`. When `get_data_dir` evaluates `Path(' /tmp/foo') / 'gptme'`, the result is a relative path (it does not start with `/`), so gptme will resolve it against the process CWD. This defeats the PR's stated purpose of keeping data/state/log paths absolute when `XDG_*` or `GPTME_LOGS_HOME` are misconfigured. In a Docker container that sets `XDG_DATA_HOME` with an accidental leading space, logs and memory are written under a relative directory instead of the intended absolute location. The new tests never hit this input because `test_preserves_surrounding_whitespace` pins the harmful behaviour.

tests/test_dirs.py
```diff
@@ -75,6 +114,13 @@ def test_gptme_logs_home_override(self, tmp_path: Path):
         assert result == logs_dir
         assert result.exists()  # should be created
 
+    def test_gptme_logs_home_empty_falls_through(self):
+        """Empty GPTME_LOGS_HOME must not produce a relative path."""
+        env = {**os.environ, "GPTME_LOGS_HOME": ""}
```
TimeToBuildBob: ⚠️ **P2** — The test `test_gptme_logs_home_empty_falls_through` is not hermetic: it builds `env = {**os.environ, "GPTME_LOGS_HOME": ""}` and uses `patch.dict(os.environ, env, clear=True)`, but it never overrides or removes `XDG_DATA_HOME`. If the ambient environment has `XDG_DATA_HOME` set to a relative path (e.g. a user forgot a leading slash), then `get_data_dir()` returns a relative path, and `get_logs_dir()` consequently returns `relative_path / "logs"`, which is relative. The assertion `result.is_absolute()` then fails even though the code under test — making an empty `GPTME_LOGS_HOME` fall through to a default — is behaving exactly as designed. The test therefore depends on an unrelated environment variable and will fail non-hermetically on any runner where `XDG_DATA_HOME` happens to be relative. Fix by also clearing or overriding `XDG_DATA_HOME` in this test (e.g. set it to an absolute `tmp_path`), so the test isolates the variable it is meant to verify.

tests/test_dirs.py
```diff
@@ -20,6 +20,31 @@
...
+        monkeypatch.setenv("_GPTME_TEST_VAR", "")
+        assert dirs._get_env_path("_GPTME_TEST_VAR") is None
+
+    def test_returns_none_when_whitespace_only(self, monkeypatch):
+        monkeypatch.setenv("_GPTME_TEST_VAR", "   ")
+        assert dirs._get_env_path("_GPTME_TEST_VAR") is None
+
+    def test_preserves_surrounding_whitespace(self, monkeypatch):
+        monkeypatch.setenv("_GPTME_TEST_VAR", "  /valid/path  ")
+        assert dirs._get_env_path("_GPTME_TEST_VAR") == "  /valid/path  "
```
TimeToBuildBob: ⚠️ **P2** — The test `test_preserves_surrounding_whitespace` asserts that `_get_env_path` returns a value with surrounding whitespace unchanged. That behaviour is exactly the flaw described in the companion finding: a `XDG_DATA_HOME` set to `' /tmp/foo'` becomes a relative path, so the test encodes a regression rather than preventing it. The PR's intent is to stop empty/whitespace-only values from producing relative paths; a value that merely has leading/trailing spaces around a real path is still a misconfiguration and should be trimmed or rejected. Because the test asserts the current (buggy) behaviour, it will not fail and effectively licenses the relative-path bug. The test should instead expect a trimmed value (e.g., `'/tmp/foo'`) or raise/return `None` for any value that differs from its stripped form.

gptme/dirs.py
TimeToBuildBob: Fixed in e5911fb11 — `_get_env_path` now returns the stripped value, so a padded `XDG_DATA_HOME` like `' /tmp/foo'` stays absolute. Added `test_xdg_data_home_padded_value_stays_absolute` as the regression. Also rebased onto master (picks up gptme/gptme#3672 CC project-dir encoding in the same file).

tests/test_dirs.py
TimeToBuildBob: Fixed in e5911fb11 — `test_gptme_logs_home_empty_falls_through` now pins `XDG_DATA_HOME` to an absolute `tmp_path` so the logs fall-through path is hermetic.

tests/test_dirs.py
TimeToBuildBob: Fixed in e5911fb11 — replaced `test_preserves_surrounding_whitespace` with `test_strips_surrounding_whitespace`, which expects `'/valid/path'`.

TimeToBuildBob:
Addressed the open P1/P2 review findings in `e5911fb11`:

- `_get_env_path` now strips surrounding whitespace so a padded `XDG_DATA_HOME` like `' /tmp/foo'` stays absolute (`Path(" /tmp/foo")` is relative).
- Logs fall-through test pins `XDG_DATA_HOME` to an absolute `tmp_path` (hermetic).
- Replaced the whitespace-preservation test with one that expects the stripped path.
- Rebased onto master to pick up gptme/gptme#3672 CC project-dir encoding in `dirs.py`.

`tests/test_dirs.py` + `tests/test_cc_memory.py`: 76 passed. In-band reviewer degraded (parse failure) and the wrapper pushed anyway; GitHub review will re-check this SHA.

TimeToBuildBob:
Automated merge handoff for `e5911fb118f5`:

- CI gate: satisfied
- Review threads: clear
- `gptme/dirs.py`

PR #3658, opened 2026-08-28
fix(convert): register gptme-convert console script and fix error message

TimeToBuildBob:
## Problem

The offline file format conversion tool (`gptme/tools/convert.py`) ships with a full CLI interface (`main()` with `prog='gptme-convert'`, three subcommands: `convert`, `check-tools`, `list-formats`) but was never registered in `[project.scripts]`. After install, `gptme-convert` is silently absent from PATH.

Additionally, the error message shown when no converter is found referenced the python module invocation (`python -m gptme.tools.convert check-tools`) rather than the installed binary name.

## Changes

- **`pyproject.toml`**: Add `gptme-convert = "gptme.tools.convert:main"` to `[project.scripts]`
- **`gptme/tools/convert.py`**: Fix error message to reference `gptme-convert check-tools` (the installed CLI) instead of the `python -m` invocation
- **`tests/test_tools_convert.py`**: Strengthen the existing error message test to assert `gptme-convert` appears in the message (previously only checked for `check-tools` substring, not the binary name)

## Verification

```
65 passed in 3.17s
```

Found during dogfooding session — the tool was unreachable at the user-facing CLI layer despite being fully implemented.

PR #3661, opened 2026-08-28
fix(tests): stop pytest-retry × tmp_path teardown KeyError failing green runs

TimeToBuildBob:
## Problem

A test that **passes on retry** can still fail the whole CI job:

```
ERROR tests/test_subagent_unit.py::TestSubagentCancel::test_thread_completion_skips_notify_when_cancel_wins_race
  - KeyError: <_pytest.stash.StashKey object at 0x7ffac78a5c00>
...
= 11579 passed, 42 skipped, 2 xfailed, 2 xpassed, 1 error, 2 retried in 822.09s =
make: *** [Makefile:59: test] Error 1
```

That was the *only* red check on #3658 — an unrelated flake that had already
passed on attempt 2. `--retries` is supposed to absorb exactly this.

## Root cause

1. pytest's `tmp_path` finalizer reads `node.stash[tmppath_result_key]` and then
   **deletes** it (`_pytest/tmpdir.py:305-312`).
2. That key is only (re)populated by tmpdir's `pytest_runtest_makereport` hook
   (`_pytest/tmpdir.py:350`).
3. pytest-retry runs a **preliminary teardown** before each retry
   (`retry_plugin.py:210`) — consuming and deleting the key — then re-runs
   setup/call by invoking the hooks directly and building the report with
   `TestReport.from_item_and_call` (`retry_plugin.py:236-238`), which **bypasses
   `pytest_runtest_makereport` entirely**.

So the key is never restored, and the final teardown of *any* retried test that
used `tmp_path` raises `KeyError`. This is a generic pytest-retry × pytest
incompatibility, not a bug in the test that happens to show red.

Reproduces in 0.07s with a 5-line test — pytest 9.1.1 × pytest-retry 1.7.0,
both inside gptme's pinned ranges (`pytest = "^9.1"`, `pytest-retry = "^1.7.0"`).

## Fix

`tests/retry_compat.py` re-seeds the key in a `tryfirst` `pytest_runtest_teardown`
wrapper. The seeded value is only consulted under
`tmp_path_retention_policy = "failed"`; we use the default (`"all"`), where the
dict is fetched but never read — so the seeded content is inert here.

It's a plugin module rather than inline conftest code so the regression test can
load **the shipped code** via `-p retry_compat` instead of a copy.

## Why this and not `@pytest.mark.no_retry`

`no_retry` (used for `test_multi_tool_per_message`, #3171) suppresses the
symptom per-test by making the test *not retry* — which turns a genuine flake
into a hard failure, and needs re-applying every time a new `tmp_path` test
flakes. This fixes the machinery once, for every test.

## Tests

- `test_retried_tmp_path_test_tears_down_cleanly` — with the shim, a retried
  `tmp_path` test exits 0 with no error.
- `test_without_shim_the_stash_keyerror_still_reproduces` — **falsifies** the
  guard above. When pytest-retry fixes this upstream, this test fails, which is
  the signal to delete `tests/retry_compat.py` and this file.

Verified: both pass against pytest 9.1.1 / pytest-retry 1.7.0. Also verified the
conftest re-export actually registers the hook (`1 passed, 1 retried`, exit 0 —
previously `1 passed, 1 error`), rather than being a silent no-op.

## Related

- Sibling guard for the other retry-machinery bug: `tests/test_pytest_timeout_retry_interaction.py` (#3634)
- Prior instance-level workaround: #3171

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,6 +16,10 @@ from pathlib import Path
 
 import pytest
 import requests
+
+# Hook implementation lives in its own module so the regression test can load
+# the exact shipped code via `-p retry_compat` instead of a copy.
+from retry_compat import pytest_runtest_teardown  # noqa: F401
 from thread_leak import (
     diff_threads,
     format_leaks,
--- a/tests/retry_compat.py
+++ b/tests/retry_compat.py
@@ -1,0 +1,36 @@
+"""Compatibility shims for pytest-retry interacting with pytest internals.
+
+Loaded as a plugin by ``tests/conftest.py`` (and directly, via ``-p
+retry_compat``, by the regression tests that guard it).
+"""
+
+import pytest
+from _pytest.tmpdir import tmppath_result_key
+
+
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
+def pytest_runtest_teardown(item, nextitem):
+    """Re-seed pytest's ``tmp_path`` bookkeeping so retried tests can tear down.
+
+    pytest's ``tmp_path`` finalizer reads ``node.stash[tmppath_result_key]`` and
+    then ``del``etes it (``_pytest/tmpdir.py``). That key is only (re)populated
+    by tmpdir's ``pytest_runtest_makereport`` hook.
+
+    pytest-retry runs a preliminary teardown before each retry -- consuming and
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
Merged as `c922c9b` after the full CI matrix passed and Greptile reported 5/5. No further action remains on this PR.

PR #3669, opened 2026-08-28
fix(llm): stop double-retrying and survive 429s in interactive mode

### LINKED ISSUE #3668: Crashes interactive mode on 429 upstream rate-limit
Baseten is fast, but wont do more than a request or two in a row.

These retries seem duplicated, 2*5 retries, with backoff not working out correctly (retry for up to a minute? in autonomous mode this should probably retry for up to 5min - to prevent a <1min blip killing a while session).

And the worst thing is that it crashes gptme instead of returning control to the user in interactive mode.

```
· Retrying request to /chat/completions in 0.484866 seconds
· Retrying request to /chat/completions in 0.777502 seconds
· Auto-generated conversation name: Spiral galaxy animation
· WARNING  OpenAI API transient error (status 429), retrying in 1.0s (attempt 1/5)
· Retrying request to /chat/completions in 0.413818 seconds
· Retrying request to /chat/completions in 0.804759 seconds
· WARNING  OpenAI API transient error (status 429), retrying in 2.0s (attempt 2/5)
· Retrying request to /chat/completions in 0.382813 seconds
· Retrying request to /chat/completions in 0.876935 seconds
· WARNING  OpenAI API transient error (status 429), retrying in 4.0s (attempt 3/5)
· Retrying request to /chat/completions in 0.468883 seconds
· Retrying request to /chat/completions in 0.812979 seconds
· WARNING  OpenAI API transient error (status 429), retrying in 8.0s (attempt 4/5)
· Retrying request to /chat/completions in 0.484119 seconds
· Retrying request to /chat/completions in 0.987602 seconds
· ERROR    Fatal error occurred
· ERROR    Error code: 429 - {'error': {'message': 'Provider returned error', 'code': 429, 'metadata': {'raw': 'z-ai/glm-5.3-flash
           is temporarily rate-limited upstream. Please retry shortly, or add your own key to accumulate your rate limits:
           https://openrouter.ai/settings/integrations', 'provider_name': 'BaseTen', 'is_byok': False, 'limit_source':
           'upstream_provider_shared_pool', 'remedy_hint': 'Retry shortly, add your own provider key
           (https://openrouter.ai/settings/integrations), or route to another provider with provider routing:
           https://openrouter.ai/docs/features/provider-routing'}}, 'user_id': '...'}
· ERROR      at /Users/erb/Programming/gptme/gptme/llm/llm_openai.py:1372 in stream
· Telemetry shutdown successfully
 ↪  ~/P/gptme  [master*]
➤➤➤
```

TimeToBuildBob:
Fixes gptme/gptme#3668.

The reported 429 exposed three separate defects. Each is fixed here.

### 1. Retries were doubled

The provider SDKs retry internally — `openai` defaults to `max_retries=2`, and our Anthropic clients passed `max_retries=5` explicitly — *on top of* gptme's own retry decorators. That is the `2*5` pattern in the issue's log: one transient error became up to 15 (openai) or 25 (anthropic) requests hammering an already rate-limited endpoint, with an effective backoff schedule nobody designed.

SDK-level retries are now disabled (`max_retries=0`). gptme's decorators are the only retry loop.

### 2. The backoff window was far too short

5 attempts × uncapped `1.0 * 2**n` gave up after ~15s of waiting, so a sub-minute upstream blip could kill an entire session.

Default is now 11 attempts with each wait capped at 60s → **~5 minutes** of cumulative backoff (1, 2, 4, 8, 16, 32, 60, 60, 60, 60). Tunable via `GPTME_LLM_MAX_RETRIES`.

### 3. It crashed instead of returning control

An exhausted retry propagated out of the chat loop and terminated gptme, even interactively — the worst part of the report.

The loop now catches provider errors (anything from `openai`, `anthropic`, or `httpx`), logs them, appends a system message to the log, and returns to the prompt so you can retry or switch model. Deliberately narrow:

- **non-interactive uses the same 11-attempt / ~5-minute retry policy, then raises if all attempts fail** — `_classify_fatal_error` exit codes (75 rate-limit, etc.) are unchanged, which automation depends on
- **non-provider exceptions still raise in both modes** — a bug in gptme must not be swallowed as a recoverable API error

Errors are matched by defining module rather than a hardcoded list of exception classes, so new SDK error types are covered without maintenance.

### Shape

The policy now lives in one place, `gptme/llm/retry_policy.py`, instead of being duplicated across `llm_openai.py` and `llm_anthropic.py`.

### Verification

- New `tests/test_llm_retry_policy.py` (9 tests): backoff cap, ~5min window, env override + invalid values, SDK retries off for both providers
- New chat-loop tests: interactive recovers from a `RateLimitError`; non-interactive still raises it; a `ValueError` still raises interactively
- Confirmed the interactive-recovery test **fails on master** and passes here
- `269 passed` across `test_llm_openai.py`, `test_llm_anthropic.py`, `test_chat.py`, `test_llm_retry_abort.py`, `test_llm_retry_policy.py`; ruff clean

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -18,7 +18,7 @@ from .constants import (
 )
 from .hooks import HookType, trigger_hook
 from .init import init
-from .llm import is_provider_error, mark_llm_reply_origin, reply
+from .llm import is_provider_error, reply
 from .llm.models import get_default_model, get_model
 from .logmanager import Log, LogManager, prepare_messages
 from .message import (
@@ -367,8 +367,8 @@ def _run_chat_loop(
             # A failing provider call (rate limit, upstream outage, network
             # error) must not kill an interactive session: report it and hand
             # control back to the user, who can retry or switch model.
-            # Only errors tagged at the LLM `reply()` call qualify — tool
-            # httpx failures must not be swallowed as "LLM request failed".
+            # Only errors tagged at the provider call inside `reply()` qualify
+            # — tool/hook httpx or SDK failures must not be swallowed.
             # Non-interactive runs still fail loudly so the exit code carries
             # the error class. See https://github.com/gptme/gptme/issues/3668
             if not interactive or not is_provider_error(e):
@@ -630,23 +630,19 @@ def step(
         if tool_format == "tool":
             tools = [t for t in get_tools() if t.is_runnable]
 
-        # generate response — isolate this call so tool/hook httpx errors are
-        # not tagged as recoverable LLM failures.
-        try:
-            with terminal_state_title("🤔 generating"):
-                msg_response = reply(
-                    msgs,
-                    get_model(model).full,
-                    stream,
-                    tools,
-                    workspace,
-                    output_schema,
-                    on_token=on_token,
-                    on_thinking=on_thinking,
-                )
-        except Exception as e:
-            mark_llm_reply_origin(e)
-            raise
+        # generate response — `reply()` tags only the provider call, after
+        # GENERATION_PRE hooks, so tool/hook failures still propagate.
+        with terminal_state_title("🤔 generating"):
+            msg_response = reply(
+                msgs,
+                get_model(model).full,
+                stream,
+                tools,
+                workspace,
+                output_schema,
+                on_token=on_token,
+                on_thinking=on_thinking,
+            )
         if get_config().get_env_bool("GPTME_COSTS"):
             log_costs(msgs + [msg_response])
 
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -536,15 +536,18 @@ def _should_prompt_for_input(log: Log) -> bool:
     last_msg = log[-1] if log else None
 
     # Check if there's an interrupt, decline, or provider-error message after
-    # the last assistant message. These all mean "hand control back to the
-    # user" rather than auto-generating. Hooks (like cost_awareness) may
-    # append after the marker, so scan until the last assistant message.
+    # the last assistant *and* last user message. These mean "hand control
+    # back to the user" rather than auto-generating. A newer user turn
+    # supersedes the marker (crash recovery / queued follow-up). Hooks
+    # (like cost_awareness) may append system messages after the marker, so
+    # skip those — but stop at user or assistant.
     has_recent_return_to_prompt = False
     for msg in reversed(log):
-        if msg.role == "assistant":
+        if msg.role in ("assistant", "user"):
             break
-        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT) or (
-            msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
+        if msg.role == "system" and (
+            msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT)
+            or msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
         ):
             has_recent_return_to_prompt = True
             break
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -545,9 +545,13 @@ def _should_prompt_for_input(log: Log) -> bool:
     for msg in reversed(log):
         if msg.role in ("assistant", "user"):
             break
-        if msg.role == "system" and (
-            msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT)
-            or msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
+        if (
+            msg.role == "system"
+            and not msg.call_id
+            and (
+                msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT)
+                or msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
+            )
         ):
             has_recent_return_to_prompt = True
             break
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -18,7 +18,7 @@ from .constants import (
 )
 from .hooks import HookType, trigger_hook
 from .init import init
-from .llm import is_provider_error, reply
+from .llm import is_provider_error, mark_llm_reply_origin, reply
 from .llm.models import get_default_model, get_model
 from .logmanager import Log, LogManager, prepare_messages
 from .message import (
@@ -367,6 +367,8 @@ def _run_chat_loop(
             # A failing provider call (rate limit, upstream outage, network
             # error) must not kill an interactive session: report it and hand
             # control back to the user, who can retry or switch model.
+            # Only errors tagged at the LLM `reply()` call qualify — tool
+            # httpx failures must not be swallowed as "LLM request failed".
             # Non-interactive runs still fail loudly so the exit code carries
             # the error class. See https://github.com/gptme/gptme/issues/3668
             if not interactive or not is_provider_error(e):
@@ -628,20 +630,25 @@ def step(
         if tool_format == "tool":
             tools = [t for t in get_tools() if t.is_runnable]
 
-        # generate response
-        with terminal_state_title("🤔 generating"):
-            msg_response = reply(
-                msgs,
-                get_model(model).full,
-                stream,
-                tools,
-                workspace,
-                output_schema,
-                on_token=on_token,
-                on_thinking=on_thinking,
-            )
-            if get_config().get_env_bool("GPTME_COSTS"):
-                log_costs(msgs + [msg_response])
+        # generate response — isolate this call so tool/hook httpx errors are
+        # not tagged as recoverable LLM failures.
+        try:
+            with terminal_state_title("🤔 generating"):
+                msg_response = reply(
+                    msgs,
+                    get_model(model).full,
+                    stream,
+                    tools,
+                    workspace,
+                    output_schema,
+                    on_token=on_token,
+                    on_thinking=on_thinking,
+                )
+        except Exception as e:
+            mark_llm_reply_origin(e)
+            raise
+        if get_config().get_env_bool("GPTME_COSTS"):
+            log_costs(msgs + [msg_response])
 
         # Trigger generation post hooks (e.g., TTS)
         if generation_post_msgs := trigger_hook(
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -18,7 +18,7 @@ from .constants import (
 )
 from .hooks import HookType, trigger_hook
 from .init import init
-from .llm import reply
+from .llm import is_provider_error, reply
 from .llm.models import get_default_model, get_model
 from .logmanager import Log, LogManager, prepare_messages
 from .message import (
@@ -361,6 +361,21 @@ def _run_chat_loop(
                 console.log("Interrupted.")
             manager.append(Message("system", INTERRUPT_CONTENT))
             # Clear any remaining prompts to avoid confusion
+            prompt_queue.clear()
+            continue
+        except Exception as e:
+            # A failing provider call (rate limit, upstream outage, network
+            # error) must not kill an interactive session: report it and hand
+            # control back to the user, who can retry or switch model.
+            # Non-interactive runs still fail loudly so the exit code carries
+            # the error class. See https://github.com/gptme/gptme/issues/3668
+            if not interactive or not is_provider_error(e):
+                raise
+            logger.error("LLM request failed: %s", e)
+            if not is_output_json() and not is_output_quiet():
+                console.log(f"[red]LLM request failed:[/red] {e}")
+            manager.append(Message("system", f"LLM request failed: {e}"))
+            # Drop queued prompts — they were meant for the failed turn.
             prompt_queue.clear()
             continue
 
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -10,6 +10,7 @@ from .config import ChatConfig, get_config, require_workspace_exists
 from .constants import (
     DECLINED_CONTENT,
     INTERRUPT_CONTENT,
+    LLM_REQUEST_FAILED_PREFIX,
     MAX_MESSAGE_LENGTH,
     MAX_PROMPT_QUEUE_SIZE,
 )
@@ -373,10 +374,10 @@ def _run_chat_loop(
             # the error class. See https://github.com/gptme/gptme/issues/3668
             if not interactive or not is_provider_error(e):
                 raise
-            logger.error("LLM request failed: %s", e)
+            logger.error("%s %s", LLM_REQUEST_FAILED_PREFIX, e)
             if not is_output_json() and not is_output_quiet():
-                console.log(f"[red]LLM request failed:[/red] {e}")
-            manager.append(Message("system", f"LLM request failed: {e}"))
+                console.log(f"[red]{LLM_REQUEST_FAILED_PREFIX}[/red] {e}")
+            manager.append(Message("system", f"{LLM_REQUEST_FAILED_PREFIX} {e}"))
             # Drop queued prompts — they were meant for the failed turn.
             prompt_queue.clear()
             continue
@@ -534,26 +535,30 @@ def _should_prompt_for_input(log: Log) -> bool:
     """
     last_msg = log[-1] if log else None
 
-    # Check if there's an interrupt or decline message after the last assistant message
-    # This handles cases where hooks (like cost_awareness) add messages after the interrupt/decline
-    has_recent_interrupt_or_decline = False
+    # Check if there's an interrupt, decline, or provider-error message after
+    # the last assistant message. These all mean "hand control back to the
+    # user" rather than auto-generating. Hooks (like cost_awareness) may
+    # append after the marker, so scan until the last assistant message.
+    has_recent_return_to_prompt = False
     for msg in reversed(log):
         if msg.role == "assistant":
             break
-        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT):
-            has_recent_interrupt_or_decline = True
+        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT) or (
+            msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
+        ):
+            has_recent_return_to_prompt = True
             break
 
     # Ask for input when:
     # - No messages at all
     # - Last message was from assistant (normal flow)
-    # - There was an interrupt or decline after the last assistant message
+    # - There was an interrupt, decline, or provider error after the last assistant
     # - Last message was pinned
     # - No user messages exist in the entire log
     return (
         not last_msg
         or last_msg.role == "assistant"
-        or has_recent_interrupt_or_decline
+        or has_recent_return_to_prompt
         or last_msg.pinned
         or not any(role == "user" for role in [m.role for m in log])
     )
--- a/gptme/llm/__init__.py
+++ b/gptme/llm/__init__.py
@@ -275,34 +275,33 @@ def _resolve_max_tokens(model: str, max_tokens: int | None) -> int | None:
     return model_meta.max_output or 4096
 
 
-# SDK exception types always mean "the provider call failed". httpx is also
-# used by tools (browser, HTTP requests), so a transport error only counts as
-# a provider failure when it originated at the LLM `reply()` call — tagged
-# there so the interactive recovery handler cannot swallow a tool outage as
-# "LLM request failed". See https://github.com/gptme/gptme/issues/3668
-_PROVIDER_SDK_MODULES = frozenset({"openai", "anthropic"})
-_PROVIDER_TRANSPORT_MODULES = frozenset({"httpx"})
+# SDK and httpx exceptions are also raised by tools and GENERATION_PRE hooks
+# (browser, embeddings, context fetchers). Interactive recovery only kicks in
+# when the exception was tagged at the actual provider call inside `reply()`,
+# after those hooks have already run. See https://github.com/gptme/gptme/issues/3668
+_PROVIDER_ERROR_MODULES = frozenset({"openai", "anthropic", "httpx"})
 _LLM_REPLY_ORIGIN_ATTR = "_gptme_from_llm_reply"
 
 
 def mark_llm_reply_origin(exc: BaseException) -> None:
-    """Mark an exception as raised from the LLM `reply()` call in `chat.step`."""
+    """Mark an exception as raised from the provider call inside `reply()`.
+
+    Applied after GENERATION_PRE hooks so a hook/tool failure is not treated
+    as a recoverable LLM outage.
+    """
     setattr(exc, _LLM_REPLY_ORIGIN_ATTR, True)
 
 
 def is_provider_error(e: BaseException) -> bool:
-    """Whether an exception came from an LLM provider SDK or its HTTP transport.
+    """Whether an exception is a recoverable LLM provider/transport failure.
 
-    Used to decide whether an interactive session can recover by returning
-    control to the user instead of crashing.
-    See https://github.com/gptme/gptme/issues/3668
+    Requires the ``_gptme_from_llm_reply`` tag so untagged openai/anthropic/httpx
+    errors from tools or hooks still propagate. See gptme/gptme#3668.
     """
+    if not getattr(e, _LLM_REPLY_ORIGIN_ATTR, False):
+        return False
     modules = {(cls.__module__ or "").split(".", 1)[0] for cls in type(e).__mro__}
-    if modules & _PROVIDER_SDK_MODULES:
-        return True
-    if modules & _PROVIDER_TRANSPORT_MODULES:
-        return getattr(e, _LLM_REPLY_ORIGIN_ATTR, False)
-    return False
+    return bool(modules & _PROVIDER_ERROR_MODULES)
 
 
 @trace_function(name="llm.reply", attributes={"component": "llm"})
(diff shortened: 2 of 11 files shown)


ErikBjare:
@TimeToBuildBob Why is non-interactive not retrying more aggressively/patiently?

TimeToBuildBob:
It is retrying equally aggressively/patiently: both modes use the same 11-attempt, ~5-minute bounded retry policy. “Non-interactive still raises” only describes what happens *after those retries are exhausted*: automation gets the provider exception and meaningful exit code instead of an interactive prompt recovery.

I'll clarify that wording in the PR description. The separate Greptile P1 is valid; I'm fixing the lazy-client construction so SDK retries are actually disabled before materialization.

gptme/llm/llm_openai.py
```diff
@@ -640,7 +646,12 @@ def get_client(provider: Provider) -> OpenAI:
     """Get client for specific provider, initializing if needed."""
     if provider not in clients:
         init(provider, get_config())
-    return clients[provider]
+    client = clients[provider]
+    # gptme's retry decorators own the retry policy; leaving the SDK's own
+    # retries enabled multiplies attempts (sdk_retries * max_retries) and makes
+    # the effective backoff schedule unpredictable. See retry_policy.
+    client.max_retries = SDK_MAX_RETRIES
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b5ae32965. `_lazy_client_factory` now records `max_retries=0` in the constructor kwargs, so the real SDK client is created with retries disabled. The regression test now follows production `init()` through lazy materialization and asserts both the recorded kwargs and materialized client setting.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Current head `b5ae32965` is ready for maintainer review/merge: all CI checks pass, Greptile re-reviewed the fixed lazy-client path at 5/5 with no blocking findings, and I resolved the now-outdated addressed thread.

I did not self-merge because this changes runtime LLM/chat behavior outside the low-risk self-merge allowlist.

gptme/chat.py
```diff
@@ -363,6 +363,21 @@ def _run_chat_loop(
             # Clear any remaining prompts to avoid confusion
             prompt_queue.clear()
             continue
+        except Exception as e:
+            # A failing provider call (rate limit, upstream outage, network
+            # error) must not kill an interactive session: report it and hand
+            # control back to the user, who can retry or switch model.
+            # Non-interactive runs still fail loudly so the exit code carries
+            # the error class. See https://github.com/gptme/gptme/issues/3668
+            if not interactive or not is_provider_error(e):
```
TimeToBuildBob, line 375: ❌ **P1** — The new exception handler in _run_chat_loop catches any exception from the entire conversation turn and classifies it as a provider error using is_provider_error, which matches by module (openai, anthropic, httpx). However, _process_message_conversation also executes tools (via execute_msg) that may use httpx directly (e.g., HTTP requests, browser automation). If such a tool raises an httpx.ConnectError or similar, that exception is misclassified as an LLM provider error. The handler then logs 'LLM request failed', appends a system message, clears the prompt queue, and returns to the user prompt — hiding the actual tool failure and losing any queued prompts. This breaks interactive sessions: a network failure in a tool looks like an LLM failure, and the user is misled about the cause while queued prompts are silently dropped.

TimeToBuildBob:
## 🤖 AI code review

Two potential P1 issues were found: the Azure OpenAI branch may have lost the api_version parameter during the refactor to _init_openai_client, and the interactive error recovery may not trigger a user prompt if the appended system message does not match the existing interrupt markers.

| `b5ae3296567a` | 2/5 | 2 | llm | 2026-08-28 14:11 UTC |
| `9e4eae1827d9` | 3/5 | 4 | llm | 2026-08-29 01:24 UTC |
| `a0eb591ec0ea` | 3/5 | 1 | llm | 2026-08-29 15:54 UTC |
| `c1a345606157` | 4/5 | 3 | llm | 2026-08-29 17:42 UTC |
| `309434bad1d7` | 3/5 | 3 | llm | 2026-08-29 19:09 UTC |
| `abcea7d29f2e` | 5/5 | 0 | llm | 2026-08-30 07:59 UTC |

</details>

<sub>Reviewed `abcea7d29f2e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 657s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Automated merge handoff for `b5ae3296567a`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/llm/__init__.py`
- `gptme/llm/llm_anthropic.py`
- `gptme/llm/llm_openai.py`
- `gptme/llm/retry_policy.py`

gptme/chat.py
TimeToBuildBob: Fixed in 9e4eae182. `step()` now tags exceptions only at the LLM `reply()` call, and `is_provider_error` treats httpx as a provider failure only when that tag is present. Tool/hook httpx errors still raise in interactive mode. Added regressions: `test_interactive_does_not_swallow_tool_httpx_errors` and `test_step_marks_httpx_from_reply_as_provider_error`.

TimeToBuildBob:
Pushed `9e4eae182` for the two AI-review P1s on `b5ae32965`:

1. Interactive recovery now tags exceptions only at the LLM `reply()` call. Untagged httpx errors from tools/hooks still raise instead of being logged as "LLM request failed".
2. Every OpenAI-compatible provider now registers through `_init_openai_client`, so SDK retries are disabled on all construction paths, not just the openai lazy client.

25 targeted tests pass. In-band review could not run (OpenRouter daily key limit); GitHub re-review will follow this push. Still not self-merging — runtime LLM/chat behavior is outside the allowlist.

TimeToBuildBob:
Automated merge handoff for `9e4eae1827d9`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/llm/__init__.py`
- `gptme/llm/llm_anthropic.py`
- `gptme/llm/llm_openai.py`
- `gptme/llm/retry_policy.py`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/chat.py
```diff
@@ -613,20 +630,25 @@ def step(
...
+                    get_model(model).full,
+                    stream,
+                    tools,
+                    workspace,
+                    output_schema,
+                    on_token=on_token,
+                    on_thinking=on_thinking,
+                )
+        except Exception as e:
+            mark_llm_reply_origin(e)
```
TimeToBuildBob: ❌ **P1** — In step(), the except block marks any exception from reply() with mark_llm_reply_origin(e). However, reply() triggers GENERATION_PRE hooks (gptme/llm/__init__.py lines 324-335) before making the provider call. If such a hook raises an httpx.ConnectError (e.g., a hook that fetches context over the network), that error is tagged as if it came from the LLM reply. The interactive recovery handler in _run_chat_loop then sees is_provider_error(e) True (httpx + tag) and swallows it as 'LLM request failed', returning control to the user instead of propagating the hook failure. This masks a real hook error and lets the session continue as if the only problem was the provider call. The tagging should only be applied to errors that actually originate from the provider SDK/transport, not from hooks or other code inside reply().

gptme/llm/llm_openai.py
```diff
@@ -93,6 +99,7 @@ def _lazy_client_factory(cls_name: str):
     """Return an ``OpenAI``/``AzureOpenAI``-like constructor that defers SDK import."""
 
     def factory(**kwargs: Any) -> OpenAI:
+        kwargs.setdefault("max_retries", SDK_MAX_RETRIES)
```
TimeToBuildBob, line 102: ⚠️ **P2** — The PR disables SDK-level retries only in llm_openai.py and llm_anthropic.py, but the openai-subscription provider (gptme/llm/llm_openai_subscription.py) is not touched and is not routed through the _init_openai_client / _lazy_client_factory path. That provider constructs its own OpenAI client and will keep the SDK's default max_retries=2, so the double-retry defect from issue #3668 persists for openai-subscription sessions: one 429 still triggers sdk_retries * max_retries requests inside a backoff window that does not outlast a rate-limit blip. The PR description claims 'Provider SDK-level retries are disabled' without limiting it to the two providers in the diff, so this is an incomplete fix.

gptme/llm/__init__.py
```diff
@@ -275,6 +275,36 @@ def _resolve_max_tokens(model: str, max_tokens: int | None) -> int | None:
...
+
+def is_provider_error(e: BaseException) -> bool:
+    """Whether an exception came from an LLM provider SDK or its HTTP transport.
+
+    Used to decide whether an interactive session can recover by returning
+    control to the user instead of crashing.
+    See https://github.com/gptme/gptme/issues/3668
+    """
+    modules = {(cls.__module__ or "").split(".", 1)[0] for cls in type(e).__mro__}
+    if modules & _PROVIDER_SDK_MODULES:
```
TimeToBuildBob: ⚠️ **P2** — is_provider_error returns True for any exception whose defining module is 'openai' or 'anthropic', without requiring the _gptme_from_llm_reply tag. The comment in chat.py says 'Only errors tagged at the LLM reply() call qualify', and the httpx branch enforces that tag, but the openai/anthropic branch does not. If a tool or hook that runs outside the LLM call raises an exception from those SDKs (for example a plugin using the OpenAI SDK to call embeddings), it will be misclassified as a recoverable provider error and silently swallowed in interactive mode, appending 'LLM request failed' and dropping the queued prompts instead of surfacing the real defect.

gptme/chat.py
```diff
@@ -363,6 +363,23 @@ def _run_chat_loop(
...
+            # Non-interactive runs still fail loudly so the exit code carries
+            # the error class. See https://github.com/gptme/gptme/issues/3668
+            if not interactive or not is_provider_error(e):
+                raise
+            logger.error("LLM request failed: %s", e)
+            if not is_output_json() and not is_output_quiet():
+                console.log(f"[red]LLM request failed:[/red] {e}")
+            manager.append(Message("system", f"LLM request failed: {e}"))
+            # Drop queued prompts — they were meant for the failed turn.
+            prompt_queue.clear()
```
TimeToBuildBob, line 382: ⚠️ **P2** — In the interactive recovery handler in _run_chat_loop, after a provider error is caught, prompt_queue.clear() is called. This drops any remaining prompts that were queued, including prompts drained from the external prompt queue via _drain_external_prompt_queue. If a subagent or external process has queued multiple prompts and the first one fails with a provider error, all subsequent prompts are silently discarded, breaking the external caller's expectation that they would all be processed. The handler should either preserve the remaining prompts (e.g., by not clearing) or at least surface them to the user.

gptme/chat.py
ErikBjare: Important to fix

TimeToBuildBob: Fixed in a0eb591ec. `reply()` now tags exceptions only around the provider call, after GENERATION_PRE hooks. A hook `httpx.ConnectError` is no longer marked as an LLM outage. Regression: `test_reply_does_not_tag_generation_pre_hook_errors`.

gptme/llm/__init__.py
TimeToBuildBob: Fixed in a0eb591ec. `is_provider_error` now requires the reply-origin tag for openai/anthropic as well as httpx, so an SDK exception from a tool or plugin is not recovered as "LLM request failed". Regression: `test_interactive_does_not_swallow_untagged_sdk_errors`.

gptme/llm/llm_openai.py
TimeToBuildBob: Won't-fix / not applicable. `openai-subscription` does not construct an OpenAI SDK client — `gptme/llm/llm_openai_subscription.py` talks to ChatGPT's backend with `requests` and has its own stream-retry loop (`GPTME_SUBSCRIPTION_STREAM_RETRIES`). There is no `max_retries=2` to disable. Documented in `retry_policy.py`.

gptme/chat.py
TimeToBuildBob: Accepted trade-off, matching KeyboardInterrupt on the same loop. Queued prompts after a failed LLM turn are treated as belonging to that turn (LOOP_CONTINUE follow-ups). Preserving them would run follow-ups against a missing assistant reply. External callers that need independent prompts can re-queue after the user retries.

TimeToBuildBob:
Pushed `a0eb591ec` for the AI-review findings on `9e4eae182`:

1. P1: `reply()` tags exceptions only after GENERATION_PRE hooks, around the provider call. A hook `httpx` failure is no longer swallowed as "LLM request failed".
2. P2: `is_provider_error` now requires that origin tag for openai/anthropic as well as httpx, so a tool using those SDKs still raises.

Greptile re-reviewed this head at 5/5 with no blocking findings. All CI checks are green. Review threads are resolved.

Self-merge policy: human review required for runtime LLM/chat paths (`gptme/chat.py`, `gptme/llm/*`). Ready for maintainer review and **manual merge**.

TimeToBuildBob:
@greptileai review

gptme/chat.py
```diff
@@ -363,6 +363,23 @@ def _run_chat_loop(
...
+            continue
```
TimeToBuildBob, line 383: ❌ **P1** — After a provider error in interactive mode, the chat loop appends a system message and continues. On the next iteration, prompt_queue is empty, so the else branch calls _get_user_input, which calls _should_prompt_for_input. Because the last message is now a system message (not an assistant message, no interrupt/decline, and there is at least one user message in the log), _should_prompt_for_input returns False. _get_user_input then returns None, and the loop, seeing _should_prompt_for_input is False, calls _process_message_conversation again to generate a response to the error. This repeats indefinitely if the provider remains down, each time appending another 'LLM request failed' message and hammering the API, without ever giving control back to the user. The test test_interactive_survives_provider_error masks this by patching _should_prompt_for_input to always return True. The handler should force a prompt for input (e.g., by ensuring _should_prompt_for_input returns True after an error) rather than blindly continuing.

TimeToBuildBob: Fixed in c1a345606. `_should_prompt_for_input` now treats `LLM request failed:` system messages like interrupt/decline, so the loop returns to the user instead of auto-generating against a still-down provider. `test_interactive_survives_provider_error` no longer patches `_should_prompt_for_input` (that was hiding the loop); added `test_should_prompt_after_provider_error_system_message`.

TimeToBuildBob:
Pushed `c1a345606` for AI-review P1 0cb85cacfd0f: after a provider error, `_should_prompt_for_input` now treats the LLM-failure system message like interrupt/decline, so interactive mode returns to the user instead of hammering a still-down API. In-band review of this head: 5/5, 0 findings.

Not self-merging: runtime LLM/chat paths are outside the allowlist.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `c1a345606157`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/constants.py`
- `gptme/llm/__init__.py`
- `gptme/llm/llm_anthropic.py`
- `gptme/llm/llm_openai.py`
- `gptme/llm/retry_policy.py`

gptme/chat.py
```diff
@@ -517,26 +535,30 @@ def _should_prompt_for_input(log: Log) -> bool:
...
+    # user" rather than auto-generating. Hooks (like cost_awareness) may
+    # append after the marker, so scan until the last assistant message.
+    has_recent_return_to_prompt = False
     for msg in reversed(log):
         if msg.role == "assistant":
             break
-        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT):
-            has_recent_interrupt_or_decline = True
+        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT) or (
+            msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
```
TimeToBuildBob: ⚠️ **P2** — In _should_prompt_for_input, the scan for a provider-error marker does not stop at the most recent user message. When the log ends with a fresh user message but also contains an earlier system message that starts with LLM_REQUEST_FAILED_PREFIX (e.g., a crash recovery after a failed provider call, or an external queue that appended a new turn), the loop continues past the user message, matches the older failure, and returns True. The chat loop then blocks waiting for user input instead of auto-generating a response to the pending user message, stalling autonomous or crash-recovery flows that previously continued automatically.

gptme/chat.py
```diff
@@ -363,6 +364,23 @@ def _run_chat_loop(
...
+            # Non-interactive runs still fail loudly so the exit code carries
+            # the error class. See https://github.com/gptme/gptme/issues/3668
+            if not interactive or not is_provider_error(e):
+                raise
+            logger.error("%s %s", LLM_REQUEST_FAILED_PREFIX, e)
+            if not is_output_json() and not is_output_quiet():
+                console.log(f"[red]{LLM_REQUEST_FAILED_PREFIX}[/red] {e}")
+            manager.append(Message("system", f"{LLM_REQUEST_FAILED_PREFIX} {e}"))
+            # Drop queued prompts — they were meant for the failed turn.
+            prompt_queue.clear()
```
TimeToBuildBob, line 382: ⚠️ **P2** — In _run_chat_loop, the new provider-error recovery path calls prompt_queue.clear() unconditionally. If the queue contained multiple prompts (e.g., from chained CLI arguments, external steering, or subagent prompts), all remaining prompts are dropped after a single failed turn. This is a behavior change: previously those prompts would have been processed after the failing turn completed, whereas now they are silently lost and the user must re-enter them manually. This can break multi-prompt workflows, especially in interactive sessions where the user expects the next queued prompt to run after recovery.

gptme/llm/__init__.py
```diff
@@ -309,6 +338,38 @@ def reply(
...
+            tools,
+            output_schema,
+            on_token,
+            on_thinking,
+            max_tokens,
+            temperature,
+            top_p,
+        )
+    except Exception as e:
+        mark_llm_reply_origin(e)
```
TimeToBuildBob, line 358: ⚠️ **P2** — In reply(), the `except Exception as e: mark_llm_reply_origin(e)` wraps the entire _reply_after_hooks scope, tagging every exception from that block, not just the provider call. is_provider_error only checks whether the exception's MRO contains openai/anthropic/httpx, so any httpx error raised outside the actual network call — e.g., inside init_llm during SDK import or in _resolve_max_tokens if it later performs I/O — would be falsely classified as a recoverable provider failure. In an interactive session such an internal bug would be silently converted into an "LLM request failed" message and the loop would continue, masking the real defect and making it hard to diagnose. The fix is to move the try/except so it wraps only the direct provider invocation, or to additionally verify the exception originates from the specific call site.

TimeToBuildBob (commented):

**3 findings** · ⚠️ **3** P2

gptme/chat.py
TimeToBuildBob: Fixed in 309434bad. `_should_prompt_for_input` now stops at the most recent user or assistant message, so a newer user turn supersedes an earlier LLM-failure marker. Hook system messages after the marker still return to the prompt. Regression in `test_should_prompt_after_provider_error_system_message`.

gptme/chat.py
TimeToBuildBob: Won't-fix / same accepted trade-off as 8a10407e4b43. `prompt_queue.clear()` matches KeyboardInterrupt on this loop: queued prompts belong to the failed turn (LOOP_CONTINUE follow-ups). Preserving them would run follow-ups against a missing assistant reply. External callers that need independent prompts can re-queue after the user retries.

gptme/llm/__init__.py
TimeToBuildBob: Won't-fix / accepted boundary. `_reply_after_hooks` is the provider-call portion of `reply()` after GENERATION_PRE (init_llm + stream/complete). An httpx failure there is a provider outage, not a tool/hook bug. Narrowing further than that was the previous P1 (e9868d748a95); a hypothetical I/O raise in helpers is not worth another split.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `309434bad1d7`:

- CI gate: satisfied
- Review threads: clear
- `gptme/chat.py`
- `gptme/constants.py`
- `gptme/llm/__init__.py`
- `gptme/llm/llm_anthropic.py`
- `gptme/llm/llm_openai.py`
- `gptme/llm/retry_policy.py`

gptme/llm/__init__.py
```diff
@@ -275,6 +275,35 @@ def _resolve_max_tokens(model: str, max_tokens: int | None) -> int | None:
     return model_meta.max_output or 4096
 
 
+# SDK and httpx exceptions are also raised by tools and GENERATION_PRE hooks
+# (browser, embeddings, context fetchers). Interactive recovery only kicks in
+# when the exception was tagged at the actual provider call inside `reply()`,
+# after those hooks have already run. See https://github.com/gptme/gptme/issues/3668
+_PROVIDER_ERROR_MODULES = frozenset({"openai", "anthropic", "httpx"})
```
TimeToBuildBob: ❌ **P1** — The interactive recovery in the chat loop filters exceptions through `is_provider_error()`, which only recognizes exceptions from modules `openai`, `anthropic`, and `httpx`. However, the `openai-subscription` provider uses the `requests` library directly (as stated in the PR description), so a 429 from that provider raises a `requests.exceptions.HTTPError` whose module is `requests`. That exception is not tagged and `is_provider_error()` returns False, so the chat loop re-raises it, crashing an interactive session. This defeats the PR's stated goal of surviving 429s in interactive mode for all providers; users of `openai-subscription` will still have their sessions terminated on rate limits.

gptme/chat.py
TimeToBuildBob, line 382: ⚠️ **P2** — The provider-error recovery in the chat loop unconditionally calls prompt_queue.clear(), discarding any chained prompts that were queued before the failure. For example, `gptme "prompt1" - "prompt2"` loads both prompts into prompt_queue; if the first prompt fails with a provider error, the second prompt is lost and the session falls back to an interactive prompt instead of continuing with the second prompt. This changes the long-standing behavior of chained prompts and can surprise users or break autonomous workflows that relied on sequential processing. The comment claims the prompts were 'meant for the failed turn', but they are actually intended for subsequent turns, so clearing the entire queue is too aggressive.

gptme/chat.py
```diff
@@ -517,26 +535,33 @@ def _should_prompt_for_input(log: Log) -> bool:
...
+    has_recent_return_to_prompt = False
     for msg in reversed(log):
-        if msg.role == "assistant":
+        if msg.role in ("assistant", "user"):
             break
-        if msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT):
-            has_recent_interrupt_or_decline = True
+        if msg.role == "system" and (
+            msg.content in (INTERRUPT_CONTENT, DECLINED_CONTENT)
+            or msg.content.startswith(LLM_REQUEST_FAILED_PREFIX)
```
TimeToBuildBob: ⚠️ **P2** — The updated _should_prompt_for_input treats any system message whose content starts with LLM_REQUEST_FAILED_PREFIX as a marker that control should return to the user. However, tool results are also system messages (with a call_id) and can contain arbitrary output. If a tool's output happens to begin with "LLM request failed:", the function will set has_recent_return_to_prompt and force a user prompt, even though the assistant is mid-turn and the tool output is a legitimate result. In a non-interactive autonomous run, this would prematurely stop the loop and require manual intervention. Excluding tool results (messages with call_id set) from the prefix match prevents this false positive while preserving the intended behavior for the error handler's system messages.

TimeToBuildBob (commented):

**3 findings** · ❌ **1** P1 · ⚠️ **2** P2

gptme/llm/__init__.py
TimeToBuildBob: Fixed in abcea7d29.

The finding's exception type was slightly off: openai-subscription was raising `ValueError("Codex API error 429: ...")`, not `requests.HTTPError`. Either way interactive recovery missed it — `is_provider_error` only matched `openai`/`anthropic`/`httpx`.

- `_open_response` now raises `requests.HTTPError` with the same message `_classify_fatal_error` already matches for 429/401/503, and closes the socket.
- `_PROVIDER_ERROR_MODULES` includes `requests`, so tagged Timeouts after the subscription stream-retry budget is exhausted recover too.
- Untagged `requests` errors from tools still propagate (origin tag required).

gptme/chat.py
TimeToBuildBob: Fixed in abcea7d29. `_should_prompt_for_input` now skips system messages with `call_id` set, so a tool result whose content happens to start with `LLM_REQUEST_FAILED_PREFIX` cannot force a user prompt. Regression in `test_should_prompt_after_provider_error_system_message`.

gptme/chat.py
TimeToBuildBob: Won't-fix / same accepted trade-off as 8a10407e4b43. `prompt_queue.clear()` matches KeyboardInterrupt on this loop: queued prompts belong to the failed turn (LOOP_CONTINUE follow-ups). Preserving them would run follow-ups against a missing assistant reply. External callers that need independent prompts can re-queue after the user retries. `gptme "p1" - "p2"` is sequential chained input; if p1's provider call is exhausted, continuing into p2 against a down API is the failure mode this PR exists to stop.

TimeToBuildBob:
Pushed `abcea7d29` for the AI-review findings on `309434bad`:

1. P1 e7ff9bb47622: openai-subscription 429s were `ValueError`, not SDK/httpx types, so interactive recovery still crashed. `_open_response` now raises `requests.HTTPError` (same message `_classify_fatal_error` already matches); `is_provider_error` also recognizes tagged `requests` errors (timeouts after the stream-retry budget included).
2. P2 e4022877143e: `_should_prompt_for_input` ignores tool-result system messages (`call_id` set) so a tool output starting with the LLM-failure prefix cannot steal the turn.
3. P2 93c8e43c65ee: won't-fix — `prompt_queue.clear()` matches KeyboardInterrupt; queued prompts belong to the failed turn.

120 targeted tests pass. In-band review skipped (this dispatch already used its 2 local iterations); GitHub re-review will follow this push. Still not self-merging — runtime LLM/chat behavior is outside the allowlist.

TimeToBuildBob:
Automated merge handoff for `abcea7d29f2e`:

- CI gate: satisfied
- `gptme/chat.py`
- `gptme/constants.py`
- `gptme/llm/__init__.py`
- `gptme/llm/llm_anthropic.py`
- `gptme/llm/llm_openai.py`
- `gptme/llm/llm_openai_subscription.py`
- `gptme/llm/retry_policy.py`

TimeToBuildBob:
Merged (`7294393dc`). Verified on `origin/master` against the three #3668 defects:

1. SDK retries off (`SDK_MAX_RETRIES = 0`) so gptme's loop is the only retry path.
2. Default budget is 11 attempts with backoff capped at 60s (~5 min window).
3. Interactive recovery returns to the prompt on tagged provider errors (including subscription `requests` 429s); non-interactive still raises after the same budget.

#3668 is closed as completed (keyword auto-close is off). No further action on this PR.

PR #3267, opened 2026-07-16
fix(security): extend GIT_CMD hardening to all remaining bare git subprocess call sites

TimeToBuildBob:
## Summary

PR #3266 introduced `GIT_CMD` (an absolute, Windows-CWD-filtered git path from `gptme/util/git_cmd.py`) and applied it to the workspace/context/selector/tree/worktree/dirs surface. This PR extends the same hardening to the **18 remaining files** that were not covered by the initial audit.

### Files changed

| File | What was using bare `"git"` |
|------|-----------------------------|
| `gptme/tools/autocommit.py` | `status`, `diff HEAD` — runs on every turn in autocommit mode |
| `gptme/tools/precommit.py` | `diff --name-only`, `ls-files` — runs on every file save |
| `gptme/server/tasks_api.py` | `rev-parse`, `branch`, `status`, `rev-parse --verify`, `diff --stat`, `log`, `config` |
| `gptme/cli/cmd_status.py` | `rev-parse --show-toplevel`, `log --oneline` |
| `gptme/cli/util.py` | `_git_run()` helper |
| `gptme/util/gh.py` | `remote get-url origin` — GitHub repo detection |
| `gptme/hooks/workspace_agents.py` | `rev-parse --abbrev-ref HEAD` |
| `gptme/checkpoint.py` | `_git()` helper + `reset --hard`, `clean -fd`, `diff` |
| `gptme/attestation.py` | `_run_git()` helper |
| `gptme/workspace_snapshot.py` | Shadow-repo `init`, `config`, `symbolic-ref` + `Shadow._run()` |
| `gptme/__version__.py` | Editable-install version detection (`rev-parse`, `tag`, `status`) |
| `gptme/agent/workspace.py` | `clone`, `submodule update`, `ls-files`, `git_no_hooks` base list |
| `gptme/agent/doctor.py` | `remote get-url origin`, `submodule status/update` |
| `gptme/lessons/installer.py` | Skill `clone --sparse` + `sparse-checkout set` |
| `gptme/cli/cmd_init.py` | Template `clone` |
| `gptme/eval/main.py` | `rev-parse --show-toplevel`, `describe` |
| `gptme/eval/swebench/evaluate.py` | `diff` capture |
| `gptme/eval/swebench/utils.py` | Repo `clone`, `fetch origin`, `checkout` |

## Testing

- All 11 existing `test_git_cmd_resolver.py` and `test_tree.py` tests pass (2 Windows-only tests correctly skipped on Linux).
[...]

PR #3674, opened 2026-08-28
fix(security): route git subprocesses through GIT_CMD in review-pr and 3-way merge

AmaLS367:
Follow-up to #3267.

## Problem

#3267 ("extend GIT_CMD hardening to **all** remaining bare git subprocess call
sites") replaced every `["git", ...]` subprocess with `[GIT_CMD, ...]`, where
`GIT_CMD` is an absolute path resolved with the process CWD stripped from the
search path. The point is Windows `CreateProcess`, which resolves a bare `git`
against the current directory before PATH — so a `git.exe` / `git.bat` sitting
in the working directory gets run instead of the real git.

Three git call sites added **after** that audit reintroduced the bare-`"git"`
pattern:

| Site | Added | Context |
|---|---|---|
| `gptme/tools/hashline_edit.py` — `git merge-file` | #3520 (Aug 12) | 3-way merge recovery; runs with the process CWD, which for CLI sessions is the workspace |
| `gptme/cli/cmd_review_pr.py` — `_git_output()` helper | #3512 (Aug 10) | review provenance checks |
| `gptme/cli/cmd_review_pr.py` — `git archive` in `_materialize_review_tree()` | #3512 | runs against a checkout of **attacker-authored PR content** by design (see that function's docstring) |

The `review-pr` `git archive` case is the sharpest: the reviewer is explicitly
built to handle hostile PR content, and a PR can commit a `git.bat` to the repo
root.

## Fix

Route all three through `GIT_CMD`, exactly as #3267 did for the sites it covered.
No behaviour change on POSIX (`GIT_CMD` is just `shutil.which("git")` there).

`gh` invocations in `cmd_review_pr.py` have the same CWD-hijack shape but there
is no `GH_CMD` equivalent yet — left out of scope.

## Tests

- New `test_merge_invokes_git_via_resolved_path` — asserts `merge-file` is
  invoked with `GIT_CMD` as argv[0].
- `test_materialize_review_tree_captures_archive_before_extracting` updated to
  assert `GIT_CMD` (mirrors the assertion updates #3267 made in
  `test_tools_autocommit.py`).
- Full `test_tools_hashline_edit.py` + `test_util_review.py` pass (283 tests).

### DIFF
--- a/gptme/cli/cmd_review_pr.py
+++ b/gptme/cli/cmd_review_pr.py
@@ -79,6 +79,7 @@ import click
 
 from ..tools._allowlist import READ_ONLY_TOOL_PRESET
 from ..util.gh import infer_owner_repo, run_gh_json
+from ..util.git_cmd import GIT_CMD
 from ..util.review import (
     FindingSeverity,
     FindingStatus,
@@ -481,7 +482,7 @@ def _git_output(args: list[str]) -> str | None:
     """Run a read-only git command in the working directory; None on failure."""
     try:
         result = subprocess.run(
-            ["git", *args],
+            [GIT_CMD, *args],
             capture_output=True,
             text=True,
             timeout=15,
@@ -623,7 +624,7 @@ def _materialize_review_tree(cwd: str) -> tempfile.TemporaryDirectory[str]:
     export = tempfile.TemporaryDirectory(prefix="gptme-review-")
     try:
         archive = subprocess.run(
-            ["git", "archive", "--format=tar", "HEAD"],
+            [GIT_CMD, "archive", "--format=tar", "HEAD"],
             cwd=cwd,
             capture_output=True,
             check=False,
--- a/gptme/tools/hashline_edit.py
+++ b/gptme/tools/hashline_edit.py
@@ -51,6 +51,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Literal
 
 from ..message import Message
+from ..util.git_cmd import GIT_CMD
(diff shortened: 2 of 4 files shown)


TimeToBuildBob (commented):
Local reproduction on Linux at PR head `dd31c7bd9`:

```
cd /tmp/worktrees/gptme-pr-3674
uv run pytest \
  tests/test_tools_hashline_edit.py::TestMergeRecovery::test_merge_invokes_git_via_resolved_path \
  tests/test_git_cmd_resolver.py \
  tests/test_util_review.py::test_materialize_review_tree_captures_archive_before_extracting \
  -v
```

**5 passed, 2 skipped** in 0.58s. The skips are the Windows CreateProcess tests (`test_git_cmd_not_in_cwd_on_windows`, `test_git_cmd_cwd_filtered_from_path`) — expected off Windows.

On this host `GIT_CMD` resolved to `/home/bob/bin/git` (absolute, exists). The new merge-file spy asserts `merge_cmds[0][0] == GIT_CMD`, and the review-tree archive mock now expects `[GIT_CMD, "archive"]`.

I also AST-walked `gptme/**/*.py` for leftover `subprocess.{run,Popen,call,check_output,check_call}(["git", ...])` and found **zero** remaining production hits. The `"git"` strings still in `tree.py` / `shell.py` are method names and token matches, not executable argv.

No blocking findings. Codecov’s missing lines are in the new test, not the two production call sites. Windows hijack protection still lives in the #3267 resolver; this PR just plugs the leftover callers.

PR #3675, opened 2026-08-28
fix(tests): join subagent threads before teardown in TestParentContextForwarding

TimeToBuildBob:
## Problem

CI was failing with:

```
ERROR tests/test_subagent_unit.py::TestParentContextForwarding::test_context_turns_validation_negative - RuntimeError: dictionary changed size during iteration
```

The five tests in `TestParentContextForwarding` that call `subagent()` with a mocked `_create_subagent_thread` were prematurely removing their subagents from `_subagents` at the end of the test body:

```python
# Cleanup
with _subagents_lock:
    _subagents[:] = [s for s in _subagents if s.agent_id != "test-no-log"]
```

This removal happens **before** `cleanup_subagents_after`'s teardown runs, so that fixture can't find the thread to join it. The background thread leaks into the next test's setup phase where it does lazy imports (in `notify_completion`, `release_thread`, etc.), mutating `sys.modules` concurrently while the main thread iterates it in the `detect_leaked_threads` fixture. That's the dict-iteration race.

## Fix

Replace each premature `_subagents[:]` cleanup with an explicit thread join (2s timeout). The threads complete almost immediately after the mock returns; the timeout is just a safety net. `cleanup_subagents_after` then handles the final `_subagents.clear()` as designed.

## Verification

- All 9 `TestParentContextForwarding` tests pass locally
- `test_context_turns_validation_negative` passes individually
- Pre-commit (prek) clean

### DIFF
--- a/tests/test_subagent_unit.py
+++ b/tests/test_subagent_unit.py
@@ -3176,9 +3176,13 @@ class TestParentContextForwarding:
         assert captured_parent_msgs[0] is None
         assert any("context_turns" in r.message for r in caplog.records)
 
-        # Cleanup
+        # Join the background thread before removing it from _subagents so
+        # cleanup_subagents_after can't miss it and leave a leaked thread that
+        # races the next test's setup via lazy-import sys.modules mutations.
         with _subagents_lock:
-            _subagents[:] = [s for s in _subagents if s.agent_id != "test-no-log"]
+            _sa = next((s for s in _subagents if s.agent_id == "test-no-log"), None)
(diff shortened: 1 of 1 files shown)


TimeToBuildBob:
Greptile 5/5, no findings. Self-merge blocked only on remaining CI (`Test without API keys`, `Test without API keys or extras`). No code change this cycle.

TimeToBuildBob:
Merged as 8213fe8352. Remaining CI finished green; no further action on this PR.

PR #1132, opened 2026-01-19
fix(server): improve skip message to prevent LLM re-suggestion

gptme-thomas:
When user skips a tool execution, provide a meaningful message that instructs the LLM not to re-suggest the same action.

## Changes

**Before:** `Skipped tool {tool_id}`

**After:** `User chose not to execute this {tool_name} tool. Do not re-suggest the same action unless explicitly requested.`

This addresses the issue where the LLM would keep suggesting the same tool call after the user declined to execute it.

## Testing

- Syntax check passed
- Pre-commit hooks passed
- Change is minimal and localized to message string

Fixes: gptme/gptme-webui#66

----

> [!IMPORTANT]
> Update skip message in `api_conversation_tool_confirm()` to prevent LLM from re-suggesting skipped tools.
> 
>   - **Behavior**:
>     - Update skip message in `api_conversation_tool_confirm()` in `api_v2_sessions.py` to prevent LLM from re-suggesting skipped tools.
>     - New message: `User chose not to execute this {tool_name} tool. Do not re-suggest the same action unless explicitly requested.`
>   - **Testing**:
>     - Syntax check and pre-commit hooks passed.
>     - Change is minimal and localized to message string.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for a6efd1011e41fb02a2a8830bc5277f01cac776c0. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #3681, opened 2026-08-30
fix(models): stamp default_tool_format on dynamic listing constructors; serialize in model_to_dict

TimeToBuildBob:
## Summary
- Stamp `default_tool_format="tool"` on `openrouter_model_to_modelmeta` and `_openai_compatible_model_to_modelmeta` — dynamic listing rows previously dropped the dialect, so runtime-discovered models resolved to markdown tool format instead of the openai-compat `tool` dialect.
- `model_to_dict` now serializes `default_tool_format` (when set) and `supports_strict_tools` (when True) so `gptme models --json` exposes the dialect.
- Add a parametrized `get_model` preservation test over all 111 static `MODELS` rows (dialect + capability fields), constructor tests for both dynamic builders, and a serialization test.

## Context
Follow-up to gptme/gptme#3566 (static stamp), which verified 111/111 static rows but left the dynamic constructors as the remaining drop. Does not re-stamp MODELS, add eval-derived overrides, or build a protocol gateway.

Refs: ErikBjare/bob idea #1132 / task `gptme-route-dialect-conformance`.

### DIFF
--- a/gptme/llm/models/listing.py
+++ b/gptme/llm/models/listing.py
@@ -42,6 +42,10 @@ def model_to_dict(model: ModelMeta) -> dict[str, Any]:
         d["deprecated"] = True
     if model.preferred_edit_format is not None:
         d["preferred_edit_format"] = model.preferred_edit_format
+    if model.default_tool_format is not None:
+        d["default_tool_format"] = model.default_tool_format
+    if model.supports_strict_tools:
+        d["supports_strict_tools"] = True
     return d
 
 
--- a/tests/test_llm_models.py
+++ b/tests/test_llm_models.py
@@ -612,3 +612,69 @@ class TestSupportsResponsesAPI:
     def test_gpt4o_is_not_marked_for_responses_api(self):
         model = get_model("openai/gpt-4o")
         assert model.supports_responses_api is False
+
+
+@pytest.mark.parametrize(
+    ("provider", "name"),
+    sorted((p, n) for p, models in MODELS.items() for n in models),
+)
+def test_get_model_preserves_route_dialect_fields(provider: str, name: str):
+    """get_model must preserve dialect/capability metadata for every static registry row."""
+    meta = MODELS[provider][name]  # type: ignore[index]
+    resolved = get_model(f"{provider}/{name}")
+    assert resolved.default_tool_format == meta.get("default_tool_format")
+    assert resolved.supports_strict_tools == meta.get("supports_strict_tools", False)
+    assert resolved.preferred_edit_format == meta.get("preferred_edit_format")
+    assert resolved.deprecated == meta.get("deprecated", False)
+    assert resolved.pricing_type == meta.get("pricing_type", "per_token")
+    assert resolved.supports_parallel_tool_calls == meta.get(
+        "supports_parallel_tool_calls", False
+    )
+
+
+def test_openrouter_constructor_stamps_default_tool_format():
+    """Dynamic OpenRouter listing rows must carry the openai-compat dialect."""
+    from gptme.llm.llm_openai import openrouter_model_to_modelmeta
+
+    meta = openrouter_model_to_modelmeta(
+        {
+            "id": "openrouter/anthropic/claude-3-5-sonnet",
+            "pricing": {"prompt": "0", "completion": "0"},
+            "architecture": {"input_modalities": ["text"]},
+            "supported_parameters": [],
+        }
+    )
+    assert meta.default_tool_format == "tool"
+
+
+def test_openai_compatible_constructor_stamps_default_tool_format():
+    """Dynamic openai-compat listing rows must carry the openai-compat dialect."""
+    from gptme.llm.llm_openai import _openai_compatible_model_to_modelmeta
+
+    meta = _openai_compatible_model_to_modelmeta(
+        {
+            "id": "some-model",
+            "context_length": 8192,
+            "architecture": {"input_modalities": ["text"]},
+        },
+        provider_name="custom",
+    )
+    assert meta.default_tool_format == "tool"
+
+
+def test_model_to_dict_serializes_default_tool_format():
+    """A stamped model serializes default_tool_format; unstamped omits it."""
+    from gptme.llm.models.listing import model_to_dict
+    from gptme.llm.models.types import CustomProvider
+
+    stamped = ModelMeta(
+        provider=CustomProvider("test"),
+        model="m1",
+        context=8192,
+        default_tool_format="tool",
+    )
+    d = model_to_dict(stamped)
+    assert d["default_tool_format"] == "tool"
+
+    unstamped = ModelMeta(provider=CustomProvider("test"), model="m2", context=8192)
+    assert "default_tool_format" not in model_to_dict(unstamped)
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
CI-red was not the new constructor tests. `model_to_dict` now reads `default_tool_format` / `supports_strict_tools`, and the `models list --json` SimpleNamespace stubs in `tests/test_util_cli.py` did not have those attributes, so both jobs died with:

```
AttributeError: 'types.SimpleNamespace' object has no attribute 'default_tool_format'
```

on `test_models_list_json_suppresses_provider_noise` and `test_models_list_json_available_keeps_plugin_models`.

Fix: `0c2579c33` — the three stubs now carry `default_tool_format="tool"` and `supports_strict_tools=False`. Local: 116 passed (the two CLI tests + the 111-row `get_model` preservation parametrize + constructor/serialize tests).

TimeToBuildBob:
@greptileai review

tests/test_llm_models.py
```diff
@@ -612,3 +612,69 @@ def test_gpt5_family_is_marked_for_responses_api(self):
...
         model = get_model("openai/gpt-4o")
         assert model.supports_responses_api is False
+
+
+@pytest.mark.parametrize(
+    ("provider", "name"),
+    sorted((p, n) for p, models in MODELS.items() for n in models),
+)
+def test_get_model_preserves_route_dialect_fields(provider: str, name: str):
+    """get_model must preserve dialect/capability metadata for every static registry row."""
```
TimeToBuildBob, line 622: ❌ **P1** — The test `test_get_model_preserves_route_dialect_fields` compares `resolved.default_tool_format` against `meta.get("default_tool_format")` without a default, so for a static row lacking the key it asserts `None` while the newly stamped module attribute may differ. More importantly, the test iterates `MODELS` rows but `get_model` may merge values from the shared default construction path (`_openai_compatible_model_to_modelmeta` or `openrouter_model_to_modelmeta`) rather than reading the row verbatim; if any row sets `supports_strict_tools` or `default_tool_format` only on the static dict, the assertion would fail, but if a row's absence is compensated by the constructor stamp, the test cannot distinguish and could pass vacuously for rows that do not render the field. The concrete defect: for a row that does not declare `default_tool_format`, `meta.get("default_tool_format")` is `None`, while `resolved.default_tool_format` may be `"tool"` when the row flows through a constructor now stamped by this PR, so the preservation assertion forces dynamic stamping to be dropped for that row — contradicting the PR's own stated behavior. Since `MODELS` is not shown to be re-stamped, this test bakes in an assumption that omitting the key yields `None` on the resolved model, which after the constructor change is false for rows routed through the stamped builders, making the test fail at runtime where the actual feature intends `"tool"`.

gptme/llm/models/listing.py
```diff
@@ -42,6 +42,10 @@ def model_to_dict(model: ModelMeta) -> dict[str, Any]:
         d["deprecated"] = True
     if model.preferred_edit_format is not None:
         d["preferred_edit_format"] = model.preferred_edit_format
+    if model.default_tool_format is not None:
```
TimeToBuildBob, line 45: ⚠️ **P2** — In `gptme/llm/models/listing.py`, `model_to_dict` now emits `default_tool_format` only when set and `supports_strict_tools` only when True, but it does not serialize other dialect/capability fields such as `supports_parallel_tool_calls` or the pricing fields when they are present. However, the main concern: the new serialization writes `default_tool_format="tool"` unconditionally for stamped rows, but consumers of the JSON (e.g., `gptme models --json` output) may not have been updated to read it, and the test `test_models_list_json_available_keeps_plugin_models` in `tests/test_util_cli.py` only asserts model names, so the new fields are untested end-to-end in the CLI JSON output. That is a test coverage gap, not a defect in behavior. More concretely, in `listing.py`, the PR adds the fields after the `preferred_edit_format` block; there is no deserialization counterpart shown, so a persisted listing written before this change (without the fields) remains valid, and a listing written now includes fields ignored by older readers — a backward-compatible additive change, which is not a defect.

TimeToBuildBob (commented):

**2 findings** · ❌ **1** P1 · ⚠️ **1** P2

TimeToBuildBob:
## 🤖 AI code review

Adds default_tool_format="tool" to the OpenRouter and OpenAI-compatible model constructors so runtime-discovered models use the OpenAI tool dialect, and extends model_to_dict to serialize default_tool_format when set and supports_strict_tools when true. Adds preservation tests for static registry entries, constructor stamping tests, and a serialization test.

- `gptme/llm/llm_openai.py` — Adds default_tool_format="tool" to the OpenAI-compatible and OpenRouter model constructors so dynamically discovered models use the OpenAI tool dialect.
- `tests/test_llm_models.py` — Adds preservation tests for static registry entries, constructor stamping tests for both dynamic builders, and a serialization test.
- `tests/test_util_cli.py` — Updates test fixtures to include the new default_tool_format and supports_strict_tools attributes used by the serialization path.

<sub>Also changed, not described above: `gptme/llm/models/listing.py`</sub>

</details>

| `0c2579c33de5` | 3/5 | 2 | llm | 2026-08-30 03:38 UTC |

</details>

<sub>Reviewed `0c2579c33de5` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 481s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_llm_models.py
TimeToBuildBob: False positive — the preservation test does not route static MODELS rows through the dynamic constructors.

`get_model` for a static registry hit returns early from the MODELS lookup:

```python
if provider in MODELS and lookup_model_name in MODELS[provider]:
    return ModelMeta(provider, model_name, **MODELS[provider][lookup_model_name])
```

`_openai_compatible_model_to_modelmeta` / `openrouter_model_to_modelmeta` only convert API listing payloads. They are not on this path.

The assertion is the intended contract: a static row's `default_tool_format` is preserved as written. `ModelMeta.default_tool_format` defaults to `None`, matching `meta.get("default_tool_format")` when the key is absent. Constructor stamping is already covered by `test_openrouter_constructor_stamps_default_tool_format` and `test_openai_compatible_constructor_stamps_default_tool_format`. The 111-row parametrize is green on this head.

gptme/llm/models/listing.py
TimeToBuildBob: Won't fix — the finding itself says this is not a defect.

`model_to_dict` already serializes `supports_parallel_tool_calls` (listing.py:33). The new fields are additive: `default_tool_format` is emitted only when set, `supports_strict_tools` only when True. Older readers ignore unknown keys; missing keys on older listings remain valid.

Production type is `ModelMeta`; both fields always exist, so `getattr` is unnecessary. The SimpleNamespace fixture `AttributeError` was already fixed in 0c2579c. Serialization is covered by `test_model_to_dict_serializes_default_tool_format`. Expanding CLI JSON assertions to re-check the same `model_to_dict` path is test theater, not a behavior gap.

TimeToBuildBob:
Automated merge handoff for `0c2579c33de5`:

- CI gate: satisfied
- Review threads: clear
- `gptme/llm/llm_openai.py`
- `gptme/llm/models/listing.py`

PR #3684, opened 2026-08-30
fix(webui): gate task creation in demo mode

TimeToBuildBob:
## Summary

- hide the backend-only task creation affordance in offline demo mode
- explain that task creation requires a live gptme server
- add focused sidebar tests for live-server and demo behavior

## Reproduction

On `https://gptme.ai/chat?demo=1`, **Tasks → Create task** currently submits to `http://127.0.0.1:5700/api/v2/tasks`. Production CSP blocks the request and the dialog fails without user-visible feedback.

## Verification

- `jest src/components/__tests__/UnifiedSidebar.test.tsx --runInBand`
- `tsc -p tsconfig.app.json --noEmit`
- `eslint src/components/UnifiedSidebar.tsx src/components/__tests__/UnifiedSidebar.test.tsx`
- pre-commit hooks

Closes #3683

PR #3689, opened 2026-08-30
fix(webui): disable task creation in offline demo mode

### LINKED ISSUE #3683: fix(webui): disable task creation in offline demo mode
## Problem

The hosted offline demo at `https://gptme.ai/chat?demo=1` exposes the **Tasks → Create task** flow as if it were functional. Submitting the form tries to call `http://127.0.0.1:5700/api/v2/tasks`; production CSP blocks it, the dialog silently resets, and the only explanation is in the browser console:

```text
Connecting to 'http://127.0.0.1:5700/api/v2/tasks' violates ... Content Security Policy
Error creating task: TypeError: Failed to fetch
```

This is a fresh hosted-demo workflow, reproduced on the deployed revisions `gptme-cloud@3aafd150` / `gptme@7fe250529` on 2026-08-30.

The read-side task hooks already suppress backend requests when `isDemoMode()` is true, but `useCreateTaskMutation()` and the **Create task** control do not. That leaves a polished form whose only outcome is a silent failure.

## Reproduction

1. Open `https://gptme.ai/chat?demo=1`.
2. Click **Tasks**.
3. Click **Create task**.
4. Enter a description and target repository.
5. Submit.
6. Observe the form reset/stay open and the CSP/fetch errors in the console.

## Expected

Offline demo mode must not offer a backend-only task creation action as functional. Disable or hide **Create task** in demo mode and explain that tasks require a connected server/full account (or implement a deliberate in-memory demo fixture).

## Scope

- Gate the task creation affordance in demo mode.
- Add a focused regression test proving the offline demo cannot submit to the live task API.
- Do not add task fixtures or broaden the demo backend in this fix.

TimeToBuildBob:
## Problem

The hosted offline demo at `https://gptme.ai/chat?demo=1` exposes the **Tasks → Create task** button as if it were functional. Submitting the form tries to call `http://127.0.0.1:5700/api/v2/tasks`; production CSP blocks it, the dialog silently resets, and the only explanation is in the browser console.

The read-side task hooks already suppress backend requests when `isDemoMode()` is true, but the task creation affordance does not.

Closes #3683

## Changes

- **`UnifiedSidebar.tsx`**: hide the + Create Task button behind `!isDemoMode()`; show `'Task creation requires a live gptme server.'` in the empty-tasks state when demo mode is active (mirrors the existing pattern in `AgentsView.tsx`)
- **`MainLayout.tsx`**: skip rendering `TaskCreationDialog` entirely when in demo mode — belt and suspenders so the dialog never mounts even if the button guard is bypassed

## Tests

New `UnifiedSidebar.test.tsx` with 3 focused assertions:
- `shows Create Task button when connected to a live server`
- `hides Create Task button in offline demo mode`
- `shows an explanation in offline demo mode instead of "No tasks yet"`

Follows the same mocking pattern as `AgentsView.test.tsx`.

### DIFF
--- a/tests/test_generation_pre_snapshot_boundary.py
+++ b/tests/test_generation_pre_snapshot_boundary.py
@@ -1,0 +1,157 @@
+"""Phase 2 regression tests: GENERATION_PRE message snapshot boundary and provider-ordering stability.
+
+These tests verify that:
+1. GENERATION_PRE messages are captured at a durable snapshot boundary (not recomputed each turn)
+2. Provider request construction produces byte-identical output across turns
+3. OpenAI system message hoisting and Anthropic system message extraction are stable
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+
+from gptme.message import Message
+
+
+@dataclass
+class GenerationPreSnapshot:
+    """Snapshot of GENERATION_PRE messages at a durable boundary.
+
+    This represents the transient messages that should be captured once per turn
+    and reused for the provider request, rather than being recomputed.
+    """
+
+    messages: list[Message]
+    model: str
+    workspace: str | None
+    fingerprint: str  # SHA256 of the message content for cache validation
+
+
+class TestGenerationPreSnapshotBoundary:
+    """Test that GENERATION_PRE messages are captured and persisted."""
+
+    def test_provider_message_ordering_openai_responses(self):
+        """OpenAI Responses API should hoist system messages consistently."""
+        from gptme.llm.openai_responses import (
+            MessageDict,
+            _messages_dicts_to_responses_input,
+        )
+
+        # Message structure: original + GENERATION_PRE-added system message
+        messages_dicts: list[MessageDict] = [
+            {"role": "system", "content": "You are helpful."},
+            {"role": "system", "content": "Extra context from hook"},
+            {"role": "user", "content": "Hello"},
+        ]
+
+        # First call - should hoist system messages to instructions
+        instructions_1, items_1 = _messages_dicts_to_responses_input(messages_dicts)
+
+        # Second call - should produce identical instructions (stable)
+        instructions_2, items_2 = _messages_dicts_to_responses_input(messages_dicts)
+
+        # This is the regression test: ensure byte-identical hoisting
+        assert instructions_1 == instructions_2, "System message hoisting not stable"
+        assert items_1 == items_2, "Message items not stable"
+
+        # Verify system messages are correctly hoisted
+        if instructions_1 is not None:
+            assert "You are helpful" in instructions_1
+            assert "Extra context from hook" in instructions_1
+
+    def test_provider_message_ordering_anthropic_system_extraction(self):
+        """Anthropic should extract system messages consistently."""
+        from gptme.llm.llm_anthropic import _transform_system_messages
+
+        # Message structure: original + GENERATION_PRE-added system message
+        messages = [
+            Message(role="system", content="You are helpful."),
+            Message(role="system", content="Extra context from hook"),
+            Message(role="user", content="Hello"),
+        ]
+
+        # First call - should extract system messages consistently
+        transformed_1, system_1 = _transform_system_messages(
+            messages, model="claude-opus-4-8"
+        )
+
+        # Second call - should produce identical extraction
+        transformed_2, system_2 = _transform_system_messages(
+            messages, model="claude-opus-4-8"
+        )
+
+        # Regression test: ensure deterministic extraction across calls
+        # This is the core cache-stability requirement
+        assert system_1 == system_2, "System messages not extracted consistently"
+
+    def test_unchanged_history_produces_identical_provider_request(self):
+        """Core regression: unchanged message history → byte-identical provider request."""
+        messages = [Message(role="user", content="Test")]
+        model = "openai/gpt-4o-mini"
+
+        # Capture request state for first turn
+        request_state_1 = {
+            "messages": [m.to_dict() for m in messages],
+            "model": model,
+        }
+
+        # Simulate second turn with same messages (no new user input)
+        request_state_2 = {
+            "messages": [m.to_dict() for m in messages],
+            "model": model,
+        }
+
+        # These should be byte-identical (cache-stable prefix)
+        json_1 = json.dumps(request_state_1, sort_keys=True)
+        json_2 = json.dumps(request_state_2, sort_keys=True)
+
+        assert json_1 == json_2, "Request state not stable across turns"
+
+
+class TestProviderOrderingRegressions:
+    """Regressions for provider-specific message ordering."""
+
+    def test_openai_system_message_hoisting_order(self):
+        """OpenAI Responses hoisting should preserve system message order."""
+        from gptme.llm.openai_responses import (
+            MessageDict,
+            _messages_dicts_to_responses_input,
+        )
+
+        messages_dicts: list[MessageDict] = [
+            {"role": "system", "content": "First context"},
+            {"role": "system", "content": "Second context"},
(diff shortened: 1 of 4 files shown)


tests/test_generation_pre_snapshot_boundary.py
```diff
@@ -0,0 +1,157 @@
...
+        # Second call - should produce identical extraction
+        transformed_2, system_2 = _transform_system_messages(
+            messages, model="claude-opus-4-8"
+        )
+
+        # Regression test: ensure deterministic extraction across calls
+        # This is the core cache-stability requirement
+        assert system_1 == system_2, "System messages not extracted consistently"
+
+    def test_unchanged_history_produces_identical_provider_request(self):
```
TimeToBuildBob, line 88: ⚠️ **P2** — The test `test_unchanged_history_produces_identical_provider_request` constructs two identical dictionaries (`request_state_1` and `request_state_2` from the same `messages` list and same `model`), serializes them with the same `json.dumps` call, and asserts they are equal. This assertion will always pass regardless of how provider requests are actually constructed; it does not exercise any real provider logic or validate that two distinct requests produce identical output. The test is a tautology and adds no regression value. If the intent is to verify that request serialization is deterministic, the test should at least use different message objects or independent construction paths. As-is, it gives a false sense of coverage and wastes CI time.

TimeToBuildBob (commented):

**1 finding** · ⚠️ **1** P2

TimeToBuildBob:
## 🤖 AI code review

This PR gates task creation affordances in the web UI when isDemoMode() is true, hiding the Create Task button in UnifiedSidebar and skipping the TaskCreationDialog in MainLayout. It also adds a React test for the sidebar behaviour. Additionally, it includes a new Python test file covering gptme provider message ordering that is unrelated to the demo mode change.

**⚠️ P2 medium** — `tests/test_generation_pre_snapshot_boundary.py:88`

The test `test_unchanged_history_produces_identical_provider_request` constructs two identical dictionaries (`request_state_1` and `request_state_2` from the same `messages` list and same `model`), serializes them with the same `json.dumps` call, and asserts they are equal. This assertion will always pass regardless of how provider requests are actually constructed; it does not exercise any real provider logic or validate that two distinct requests produce identical output. The test is a tautology and adds no regression value. If the intent is to verify that request serialization is deterministic, the test should at least use different message objects or independent construction paths. As-is, it gives a false sense of coverage and wastes CI time.

```suggestion
Replace the tautological assertion with concrete expectations about the returned instructions (e.g., assert that the system text is present and that `instructions_1` is not None), or remove this test unless it is specifically designed to verify deterministic output.
```

<sub>**How this was verified:** Reviewed the added test file and noticed the only post-call assertions are equality between two invocations with the same input; no actual ordering or content is asserted.</sub>

</details>

<details><summary><b>3 out-of-scope findings (outside this PR's changed files)</b></summary>

**⚠️ P2 medium** — [`…/UnifiedSidebar.tsx:455`](https://github.com/gptme/gptme/blob/26ff8d5d9cd411856c09f0da2e18daa8dc3fb10e/…/UnifiedSidebar.tsx#L455)

The demo-mode explanation text was added conditional to the same guard that renders the empty task list. Line 455 in this hunk: `{!tasksLoading && !tasksError && tasks.length === 0 && (` and line 458–459 `{demoMode ? 'Task creation requires a live gptme server.' : 'No tasks yet'}`. That means in any demo session where the task query is still in flight (`tasksLoading=true`) or surfaces an error (`tasksError=true`), the new visible explanation the PR adds is silently dropped (`tasks` is `[]` in demo because the read hook suppresses the request, per the PR description — so the empty-state branch is what demo users hit). A demo user who hits a load/error race sees the “Failed to load tasks” card (or nothing) instead of the intended “requires a live gptme server” message, leaving them back in the console-only state this PR is meant to fix. There is also no test that exercises `tasksLoading=true` or `tasksError=true` in the new suite, so this gating is unprotected.

```suggestion
Render the demo explanation independent of the empty-list guard, e.g. `{demoMode && <p>…</p>}` at the top of the tasks section, and keep it visible in loading/error states; add a test case with `tasksError=true` in demo mode.
```

**⚠️ P2 medium** — [`…/UnifiedSidebar.test.tsx:69`](https://github.com/gptme/gptme/blob/26ff8d5d9cd411856c09f0da2e18daa8dc3fb10e/…/UnifiedSidebar.test.tsx#L69)

All three new tests in this file render the component in the exact happy empty state (`tasks: []` with no `tasksLoading` and no `tasksError`). Nothing asserts the non-demo empty-state text (“No tasks yet”) renders, and nothing exercises the two states where the PR’s new explanation is gated away (loading / error). Additionally, this suite covers only the `UnifiedSidebar` gateway (button visibility); the PR’s other half — `MainLayout` unmounting `TaskCreationDialog` in demo mode — has zero coverage, so a regression that mounts the dialog back in demo mode would ship green.

```suggestion
Add cases: (a) demo=false renders “No tasks yet”; (b) demo=true + tasksError=true still shows the “requires a live gptme server” explanation; (c) a small MainLayout render test asserting the dialog isn’t present on the demo URL.
```

**⚠️ P2 medium** — [`…/MainLayout.tsx:761`](https://github.com/gptme/gptme/blob/26ff8d5d9cd411856c09f0da2e18daa8dc3fb10e/…/MainLayout.tsx#L761)

`onCreateTask={() => setShowCreateTaskDialog(true)}` is wired unconditionally in MainLayout and passed to `UnifiedSidebar`, while the only consumer of that state — `TaskCreationDialog` — is gated out in demo mode (`{!demoMode && …}`). If the boolean is ever set true while demoMode is active (it cannot today because the button is hidden, but the producer is not gated — only the consumer is), React doesn’t reset the parent state on unmount, so the stale `true` leaks into the next non-demo render and immediately re-opens the dialog. The dialog unmounting is an acceptable shortcut, but the state producer should be guarded at the same demo boundary so the invariant “never code true in demo” is enforced at the source rather than only at the consumer.

```suggestion
Gate the wiring itself in demo mode, e.g. `onCreateTask={demoMode ? undefined : () => setShowCreateTaskDialog(true)}`, or reset `showCreateTaskDialog` (useEffect) whenever demoMode flips true – so a stale flag cannot survive a demo→non-demo transition.
```

</details>

- `webui/src/components/MainLayout.tsx` — Adds isDemoMode() import and conditional rendering of TaskCreationDialog when not in demo mode.
- `webui/src/components/UnifiedSidebar.tsx` — Adds demoMode constant, hides the Create Task button, and changes the empty task message in demo mode.
- `webui/src/components/__tests__/UnifiedSidebar.test.tsx` — Adds a new React test file for UnifiedSidebar covering demo mode task creation gating.
- `tests/test_generation_pre_snapshot_boundary.py` — Adds an unrelated Python test file for gptme provider message ordering stability.

</details>

<sub>Reviewed `26ff8d5d9cd4` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1185s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Superseded by gptme/gptme#3684 (`b84ee8a3a`), already on `master`.

This PR was a second pass at the same gptme/gptme#3683 fix, opened after #3684 had already merged. The DIRTY/CONFLICTING state is that overlap in `UnifiedSidebar.tsx` / `UnifiedSidebar.test.tsx`, plus an unrelated `tests/test_generation_pre_snapshot_boundary.py` that should not have been on this branch.

Master already hides the Create Task button in demo mode and shows "Task creation requires a live gptme server." Closing instead of rebasing the duplicate. Not carrying the extra `MainLayout` dialog unmount — the only opener is the now-hidden button.

gptme/gptme#3683 is fully resolved by #3684; closing that too.

PR #3691, opened 2026-08-31
fix(knowledge): don't block on gptme-rag index after save/delete

TimeToBuildBob:
## Problem

`gptme-util knowledge save` blocks for up to 30 seconds waiting for `gptme-rag index` to complete:

```
$ time gptme-util knowledge save "p" "r"
Saved knowledge entry 4aa4fd1d
Warning: gptme-rag index failed: Command ... timed out after 30 seconds

real    0m30.281s   ← 30 s for a write to a local JSONL file
```

The entry is already persisted in JSONL before the RAG indexing starts, so keyword search works immediately. The 30-second synchronous wait is a UX regression added alongside the knowledge feature.

## Fix

Replace `subprocess.run(timeout=30)` with `subprocess.Popen` (fire-and-forget) in both `knowledge save` and `knowledge delete`. The OS reaps the orphaned gptme-rag indexer when it finishes; the CLI returns in < 0.2 s.

```
$ time gptme-util knowledge save "p" "r"
Saved knowledge entry fbd2a7cb

real    0m0.170s   ✓
```

## Tests

Three new tests in `tests/test_util_cli.py`:

- **`test_knowledge_save_uses_popen_not_blocking_run`** — regression guard: asserts `subprocess.Popen` is called (not `subprocess.run`) and that stdout/stderr are `DEVNULL` (fire-and-forget, no wait)
- **`test_knowledge_save_search_delete_roundtrip`** — end-to-end JSONL path: save → search finds entry → delete → search returns empty
- **`test_knowledge_search_empty_query_exits_nonzero`** — empty query is rejected with a non-zero exit code

### DIFF
--- a/gptme/cli/cmd_knowledge.py
+++ b/gptme/cli/cmd_knowledge.py
@@ -91,19 +91,19 @@ def knowledge_save_cmd(
         except (OSError, ValueError) as e:
             click.echo(f"Warning: gptme-rag mirror export failed: {e}", err=True)
         else:
+            # Fire-and-forget: the entry is already persisted in JSONL, so
+            # keyword search works immediately.  gptme-rag indexing is a
+            # semantic-search enhancement; blocking the CLI for it (up to
+            # the full 30 s timeout) is a poor UX trade-off.
             try:
-                subprocess.run(
+                subprocess.Popen(
                     ["gptme-rag", "index", str(kb_dir / "rag")],
-                    check=True,
-                    capture_output=True,
-                    timeout=30,
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    stdin=subprocess.DEVNULL,
                 )
-            except (
-                OSError,
-                subprocess.CalledProcessError,
-                subprocess.TimeoutExpired,
-            ) as e:
-                click.echo(f"Warning: gptme-rag index failed: {e}", err=True)
+            except OSError as e:
+                click.echo(f"Warning: could not start gptme-rag: {e}", err=True)
 
 
 def _export_for_rag(kb_dir: Path) -> None:
@@ -275,15 +275,18 @@ def knowledge_delete_cmd(entry_id: str):
     # "re-index failed" warning after a successful JSONL delete.
     rag_dir = _knowledge_dir() / "rag"
     if mirror_removed and shutil.which("gptme-rag") and rag_dir.is_dir():
+        # Fire-and-forget: re-indexing after delete is best-effort; the JSONL
+        # store is already consistent and keyword search will not return the
+        # deleted entry.  Don't block the CLI on the RAG index operation.
         try:
-            subprocess.run(
+            subprocess.Popen(
                 ["gptme-rag", "index", str(rag_dir)],
-                check=True,
-                capture_output=True,
-                timeout=30,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                stdin=subprocess.DEVNULL,
             )
-        except (OSError, subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
+        except OSError as e:
             click.echo(
-                f"Warning: gptme-rag re-index after delete failed: {e}",
+                f"Warning: could not start gptme-rag re-index after delete: {e}",
                 err=True,
             )
--- a/tests/test_util_cli.py
+++ b/tests/test_util_cli.py
@@ -1320,6 +1320,11 @@ def test_knowledge_save_uses_popen_not_blocking_run(tmp_path, monkeypatch):
 
     assert kwargs.get("stdout") == subprocess.DEVNULL
     assert kwargs.get("stderr") == subprocess.DEVNULL
+    # MagicMock.wait()/communicate() do not block, so without these
+    # assertions the test would still pass if production later added a
+    # proc.wait(timeout=30) and reintroduced the original bug.
+    mock_proc.wait.assert_not_called()
+    mock_proc.communicate.assert_not_called()
 
 
 def test_knowledge_search_empty_query_exits_nonzero(tmp_path, monkeypatch):
--- a/tests/test_util_cli.py
+++ b/tests/test_util_cli.py
@@ -1231,3 +1231,101 @@ def test_context_search_conversations_top_k(tmp_path):
     assert result.exit_code == 0, result.output
     assert "Top 5 relevant conversations" in result.output
     mock_search.assert_called_once_with("pytest", return_full=True, top_k=5)
+
+
+# ---------------------------------------------------------------------------
+# Knowledge sub-commands
+# ---------------------------------------------------------------------------
+
+
+def test_knowledge_save_search_delete_roundtrip(tmp_path, monkeypatch):
+    """save → search → delete works with the JSONL store."""
+    from unittest.mock import patch
+
+    runner = CliRunner()
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path))
+
+    # Stub out gptme-rag so tests run without it installed.
+    with (
+        patch("shutil.which", return_value=None),
+    ):
+        result = runner.invoke(
+            main, ["knowledge", "save", "pytest finds no tests", "add test_ prefix"]
+        )
+    assert result.exit_code == 0, result.output
+    assert "Saved knowledge entry" in result.output
+
+    with (
+        patch("shutil.which", return_value=None),
+    ):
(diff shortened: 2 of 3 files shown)


gptme/cli/cmd_knowledge.py
```diff
@@ -91,19 +91,19 @@ def knowledge_save_cmd(
         except (OSError, ValueError) as e:
             click.echo(f"Warning: gptme-rag mirror export failed: {e}", err=True)
         else:
+            # Fire-and-forget: the entry is already persisted in JSONL, so
+            # keyword search works immediately.  gptme-rag indexing is a
+            # semantic-search enhancement; blocking the CLI for it (up to
+            # the full 30 s timeout) is a poor UX trade-off.
             try:
-                subprocess.run(
+                subprocess.Popen(
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 84a7e7e06.

`test_cli_rag_index_oserror_is_nonfatal` now patches `subprocess.Popen` (not `subprocess.run`) and asserts the new `could not start gptme-rag` warning. The skip-path index mocks in the same file were retargeted the same way so they would actually catch an accidental launch.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_knowledge.py
TimeToBuildBob, line 99: ⚠️ **P2** — Fire-and-forget Popen with no deduplication or locking means every save/delete spawns an independent gptme-rag index process. A script that saves several entries in quick succession will launch overlapping indexers, all operating on the same rag directory. This can cause high CPU/memory usage and race conditions, potentially leaving the index partially updated or corrupted. The previous subprocess.run serialized the work and blocked until completion, so overlapping indexers were impossible. Consider serializing background indexing with a lock, a single worker process, or at least one pending job guard.

tests/test_util_cli.py
```diff
@@ -1231,3 +1231,101 @@ def test_context_search_conversations_top_k(tmp_path):
...
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path))
+
+    mock_proc = MagicMock()
+    with (
+        patch("shutil.which", return_value="/usr/bin/gptme-rag"),
+        patch(
+            "gptme.cli.cmd_knowledge._export_for_rag",
+            return_value=None,
+        ),
+        patch("subprocess.Popen", return_value=mock_proc) as mock_popen,
```
TimeToBuildBob, line 1304: ⚠️ **P2** — The new regression test is intended to guarantee non-blocking behavior, but substituting a MagicMock for subprocess.Popen makes it unable to catch a blocking .wait() or .communicate() call if one is later introduced. MagicMock returns another Mock without blocking, so the test would pass even if production code called .wait(timeout=30) on the object returned by Popen. The test only asserts that Popen was called with the expected arguments and that stdout/stderr are DEVNULL; it never checks that the returned mock is never waited on. To actually guard against the regression, the test should assert mock_proc.wait.assert_not_called() and mock_proc.communicate.assert_not_called(), or use a real Popen that can be polled.

TimeToBuildBob (commented):

**2 findings** · ⚠️ **2** P2

TimeToBuildBob:
## 🤖 AI code review

Replaces synchronous subprocess.run with fire-and-forget subprocess.Popen in knowledge save and delete to avoid blocking the CLI for up to 30 seconds. Updates existing tests to patch Popen and adds new tests for non-blocking save, roundtrip, and empty query handling.

**⚠️ P2 medium** — `gptme/cli/cmd_knowledge.py:95`

The change replaces subprocess.run(timeout=30) with subprocess.Popen without any wait, reap, or timeout. If gptme-rag index hangs (e.g., due to a huge index or a bug), the child process is left running indefinitely after the CLI exits. Repeated invocations can accumulate orphaned processes, consuming CPU and memory. The previous code killed the child after 30 seconds. This is a regression in resource containment. Consider using Popen with start_new_session=True and a background thread that waits with a timeout and kills the process group, so the CLI still returns immediately while the runaway process is eventually terminated.

```suggestion
Consider serializing re-index operations with a lock file or a single persistent worker so concurrent save/delete invocations cannot race. If fire-and-forget is kept, at minimum use `start_new_session=True` and track the child process elsewhere to avoid unbounded concurrent indexers.
```

<sub>**How this was verified:** Code review of the replacement in the save path and the same pattern in the delete path: each invocation exits immediately after spawning `gptme-rag index`, with no locking, timeout, or coordination between invocations.</sub>

</details>

- `gptme/cli/cmd_knowledge.py` — Replaces synchronous subprocess.run with fire-and-forget subprocess.Popen for gptme-rag indexing in knowledge save and delete commands
- `tests/test_knowledge.py` — Updates existing tests to patch Popen instead of run and adjust warning assertions
- `tests/test_util_cli.py` — Adds tests for non-blocking save, end-to-end roundtrip, and empty query rejection

</details>

| `84a7e7e060af` | 4/5 | 2 | llm | 2026-08-31 01:48 UTC |
| `7a4fe0190b52` | 5/5 | 0 | llm | 2026-08-31 04:06 UTC |

</details>

<sub>Reviewed `7a4fe0190b52` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 398s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/cmd_knowledge.py
TimeToBuildBob: Accepted trade-off — not fixing in this PR.

Keyword search is the source of truth and is already consistent when save/delete returns: JSONL writes go through `_exclusive_lock()` in `gptme/knowledge.py`, and the markdown export finishes before we spawn the indexer. `gptme-rag index` is a best-effort semantic enhancement, not the persistence path.

Serializing in the parent (`subprocess.run`) is the 30s CLI hang this PR removes. Cross-process serialization of the *child* would need the child to hold a lock (the parent exits immediately), plus a Windows path — `knowledge.py` already special-cases `fcntl.flock` as Unix-only. That is a lock/worker subsystem, not a one-line guard.

Overlapping indexers on this tiny markdown corpus are last-writer-wins; the next save re-exports and re-indexes. The CLI is invoked once per entry. A pending-job queue is speculative complexity for a path that is allowed to be eventually consistent.

tests/test_util_cli.py
TimeToBuildBob: Fixed in 7a4fe0190.

`test_knowledge_save_uses_popen_not_blocking_run` now asserts `mock_proc.wait.assert_not_called()` and `mock_proc.communicate.assert_not_called()`, so a later `proc.wait()` / `communicate()` cannot hide behind MagicMock.

TimeToBuildBob:
@greptileai review

TimeToBuildBob (commented):

**1 finding** · ⚠️ **1** P2

gptme/cli/cmd_knowledge.py
```diff
@@ -91,19 +91,19 @@ def knowledge_save_cmd(
         except (OSError, ValueError) as e:
             click.echo(f"Warning: gptme-rag mirror export failed: {e}", err=True)
         else:
+            # Fire-and-forget: the entry is already persisted in JSONL, so
+            # keyword search works immediately.  gptme-rag indexing is a
```
TimeToBuildBob, line 95: ⚠️ **P2** — The change replaces subprocess.run(timeout=30) with subprocess.Popen without any wait, reap, or timeout. If gptme-rag index hangs (e.g., due to a huge index or a bug), the child process is left running indefinitely after the CLI exits. Repeated invocations can accumulate orphaned processes, consuming CPU and memory. The previous code killed the child after 30 seconds. This is a regression in resource containment. Consider using Popen with start_new_session=True and a background thread that waits with a timeout and kills the process group, so the CLI still returns immediately while the runaway process is eventually terminated. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Accepted trade-off — already merged, not fixing as follow-up.

The CLI must not wait on `gptme-rag index`; that is the whole PR. The old `subprocess.run(timeout=30)` is the 30s hang. A background timeout-killer (`start_new_session=True` plus a waiter thread) does not survive the parent: a daemon thread dies with the CLI, so actually reaping/killing after exit needs a detached helper. That is a supervisor, not a Popen flag.

JSONL + keyword search is already consistent when save/delete returns. If an indexer hangs in the wild, that is a `gptme-rag` bug to fix there.

PR #3699, opened 2026-09-02
fix(tools): show enable hint when a disabled-by-default tool is invoked

### LINKED ISSUE #3698: ux: 'read' tool error should explain how to enable it (disabled_by_default=True)
## Summary

When a model tries to use the `read` tool (which is `disabled_by_default=True`), the error message gives no guidance on how to enable it.

**gptme version**: 0.33.0+076e433c
**Provider**: openrouter/openrouter/free
**Install**: uv tool install gptme

## Steps to reproduce

```bash
gptme -m "openrouter/openrouter/free" --non-interactive "create test.txt then read it back"
```

The model creates the file with `save`, then tries `read`, and gets:

```
Tool 'read' is not runnable; emitting an error tool_result to keep the tool_use/tool_result pairing valid.
System: Tool 'read' is not available for execution.
```

## Expected

The error should tell the user how to enable the tool:

> Tool 'read' is disabled by default. Enable it with `--tools +read` or use `shell` to read files instead.

## Root cause

`read` tool has `disabled_by_default=True` in its `ToolSpec`. It appears in `gptme --help` under "Available tools" (which implies it's usable), so models and users expect it to work. When invoked, the cryptic "not runnable" error provides no path forward.

## Suggested fix

In the tool-not-runnable error path, check if the tool is `disabled_by_default` and append: `"Add --tools +<name> to enable this tool."` One-line change in tool dispatch.

`shell` works as a workaround (`cat file.txt`).

TimeToBuildBob:
## Summary

When a model calls a tool that exists but is `disabled_by_default=True` (e.g. `read`), the error message now tells the user how to enable it.

**Before:**
```
System: Tool 'read' is not available for execution.
```

**After:**
```
System: Tool 'read' is not available for execution. This tool is disabled by default. Enable it with '--tools +read'.
```

The fix is in the not-runnable error path in `execute_msg`: look up the tool in `get_available_tools(include_mcp=False)` (cached) and append the hint if `disabled_by_default=True`. Truly unknown tool names (not in any available tools) get the original message unchanged.

## Changes

- `gptme/tools/__init__.py`: check `disabled_by_default` in not-runnable error path, append enable hint
- `tests/test_execute_msg_defensive.py`: add `test_disabled_by_default_tool_gets_enable_hint`

**2026-09-07 trim**: dropped an unrelated `poetry.lock`/`pyproject.toml` hunk that moved `greenlet` from the dev group to main dependencies — a CI cherry-pick smuggled into this PR with no relation to the tool-hint fix. Reverted cleanly; the diff is now just the two files above.

Fixes #3698

### DIFF
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -422,11 +422,20 @@ def execute_msg(
             # If the tool exists but is disabled by default, give an actionable hint.
             if available_for_hint is None:
                 cached = _get_available_tools_cache()
-                available_for_hint = (
-                    cached
-                    if cached is not None
-                    else get_available_tools(include_mcp=False)
-                )
+                if cached is not None:
+                    available_for_hint = cached
+                else:
+                    try:
+                        available_for_hint = get_available_tools(include_mcp=False)
+                    except Exception:
+                        # Discovery failures (bad plugin/config) must not block
+                        # yielding the paired tool_result below, or the dangling
+                        # tool_use/400 bug this branch exists to prevent reopens.
+                        logger.warning(
+                            "Failed to look up available tools for enable hint",
+                            exc_info=True,
+                        )
+                        available_for_hint = []
             if any(
                 t.name == tooluse.tool and t.disabled_by_default
                 for t in available_for_hint
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -372,6 +372,11 @@ def execute_msg(
         return
 
     remaining = iter(classified)
+    # Lazily resolved available-tool metadata for the disabled-by-default hint.
+    # Prefer the existing cache so unavailable structured calls do not re-run
+    # uncached module/plugin discovery (get_available_tools(include_mcp=False)
+    # deliberately skips the cache).
+    available_for_hint: list[ToolSpec] | None = None
     for tooluse in remaining:
         runnable = tooluse.is_runnable
         if runnable:
@@ -415,8 +420,17 @@ def execute_msg(
             )
             error_msg = f"Tool '{tooluse.tool}' is not available for execution."
             # If the tool exists but is disabled by default, give an actionable hint.
-            available = get_available_tools(include_mcp=False)
-            if any(t.name == tooluse.tool and t.disabled_by_default for t in available):
+            if available_for_hint is None:
+                cached = _get_available_tools_cache()
+                available_for_hint = (
+                    cached
+                    if cached is not None
+                    else get_available_tools(include_mcp=False)
+                )
+            if any(
+                t.name == tooluse.tool and t.disabled_by_default
+                for t in available_for_hint
+            ):
                 error_msg += (
                     f" This tool is disabled by default."
                     f" Enable it with '--tools +{tooluse.tool}'."
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -437,7 +437,7 @@ def execute_msg(
                         )
                         available_for_hint = []
             if any(
-                t.name == tooluse.tool and t.disabled_by_default
+                t.name == tooluse.tool and t.disabled_by_default and t.is_available
                 for t in available_for_hint
             ):
                 error_msg += (
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -413,9 +413,17 @@ def execute_msg(
                 "the tool_use/tool_result pairing valid.",
                 tooluse.tool,
             )
+            error_msg = f"Tool '{tooluse.tool}' is not available for execution."
+            # If the tool exists but is disabled by default, give an actionable hint.
+            available = get_available_tools(include_mcp=False)
+            if any(t.name == tooluse.tool and t.disabled_by_default for t in available):
+                error_msg += (
+                    f" This tool is disabled by default."
+                    f" Enable it with '--tools +{tooluse.tool}'."
+                )
             yield Message(
                 "system",
-                f"Tool '{tooluse.tool}' is not available for execution.",
+                error_msg,
                 call_id=tooluse.call_id,
             )
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,9 +109,6 @@ matplotlib = {version = "*", optional=true}
 pandas = {version = "*", optional=true}
 numpy = {version = "*", optional=true}
 
-# testing (needed for coverage concurrency tracking)
-greenlet = "*"  # required by coverage for concurrency=greenlet tracking
-
 # tool sounds (GPTME_TOOL_SOUNDS=true)
 sounddevice = {version = "*", optional=true}
 scipy = {version = "*", optional=true}
@@ -162,6 +159,7 @@ pytest-dotenv = "^0.5.2"
 pytest-timeout = "^2.4.0"
 pytest-retry = "^1.7.0"
 pytest-mock = "*"
+greenlet = "*"  # dependency of playwright, but needed for coverage
 
 # build
 pyinstaller = {version="^6.21", python=">=3.10,<3.15"}  # not yet compatible with Python 3.15+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,6 +109,9 @@ matplotlib = {version = "*", optional=true}
 pandas = {version = "*", optional=true}
 numpy = {version = "*", optional=true}
 
+# testing (needed for coverage concurrency tracking)
+greenlet = "*"  # required by coverage for concurrency=greenlet tracking
+
 # tool sounds (GPTME_TOOL_SOUNDS=true)
 sounddevice = {version = "*", optional=true}
 scipy = {version = "*", optional=true}
@@ -159,7 +162,6 @@ pytest-dotenv = "^0.5.2"
 pytest-timeout = "^2.4.0"
 pytest-retry = "^1.7.0"
 pytest-mock = "*"
-greenlet = "*"  # dependency of playwright, but needed for coverage
 
 # build
 pyinstaller = {version="^6.21", python=">=3.10,<3.15"}  # not yet compatible with Python 3.15+
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -71,6 +71,23 @@ class TestExecuteMsgDefensive:
             f"Expected role='system', got '{results[0].role}'"
         )
 
+    def test_disabled_by_default_tool_gets_enable_hint(self):
+        """A disabled-by-default tool (e.g. 'read') includes '--tools +name' in error."""
+        set_tool_format("tool")
+        call_id = "call-disabled-read"
+        # 'read' is disabled_by_default=True; it is discoverable but not loaded
+        content = f'@read({call_id}): {{"path": "somefile.txt"}}'
+        msg = Message("assistant", content)
+
+        results = list(execute_msg(msg))
+
+        assert len(results) == 1
+        result = results[0]
+        assert result.call_id == call_id
+        assert "--tools +read" in result.content, (
+            f"Expected enable hint in error for disabled-by-default tool, got: {result.content!r}"
+        )
+
     def test_multiple_structured_tooluses_all_get_error_results(self):
         """Multiple non-runnable structured tool_uses each get an error tool_result."""
         set_tool_format("tool")
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -40,6 +40,9 @@ class TestExecuteMsgDefensive:
         assert "nonexistent_tool" in result.content, (
             "Error message should identify the tool that was unavailable"
         )
+        assert "disabled by default" not in result.content, (
+            "Unknown tools must keep the original error; no enable hint"
+        )
 
     def test_nonrunnable_markdown_tooluse_yields_nothing(self):
         """Markdown code block tool_use (no call_id) for unknown tool → no output.
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -10,7 +10,7 @@ intentionally left unpaired when the tool is unavailable.
 """
 
 from gptme.message import Message
-from gptme.tools import execute_msg
+from gptme.tools import clear_tools, execute_msg, get_available_tools
 from gptme.tools.base import set_tool_format
 
 
@@ -77,6 +77,17 @@ class TestExecuteMsgDefensive:
     def test_disabled_by_default_tool_gets_enable_hint(self):
         """A disabled-by-default tool (e.g. 'read') includes '--tools +name' in error."""
         set_tool_format("tool")
+        # Ensure no tools are loaded so 'read' is not runnable in this context.
+        clear_tools()
+        # Precondition: 'read' must be discoverable and have disabled_by_default=True.
+        # If someone changes that flag this assertion fails loudly instead of silently.
+        read_specs = [
+            t for t in get_available_tools(include_mcp=False) if t.name == "read"
+        ]
+        assert read_specs, "read tool must be discoverable via get_available_tools"
+        assert read_specs[0].disabled_by_default, (
+            "read tool must have disabled_by_default=True for this test to be valid"
+        )
         call_id = "call-disabled-read"
         # 'read' is disabled_by_default=True; it is discoverable but not loaded
         content = f'@read({call_id}): {{"path": "somefile.txt"}}'
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -88,6 +88,11 @@ class TestExecuteMsgDefensive:
         assert read_specs[0].disabled_by_default, (
             "read tool must have disabled_by_default=True for this test to be valid"
         )
+        assert read_specs[0].is_available, (
+            "read tool must have is_available=True for this test to be valid "
+            "(the hint is only appended when the tool is both disabled_by_default "
+            "and is_available)"
+        )
         call_id = "call-disabled-read"
         # 'read' is disabled_by_default=True; it is discoverable but not loaded
         content = f'@read({call_id}): {{"path": "somefile.txt"}}'
--- a/tests/test_execute_msg_defensive.py
+++ b/tests/test_execute_msg_defensive.py
@@ -10,7 +10,15 @@ intentionally left unpaired when the tool is unavailable.
 """
 
 from gptme.message import Message
-from gptme.tools import clear_tools, execute_msg, get_available_tools
+from gptme.tools import (
+    _get_available_tools_cache,
+    _set_available_tools_cache,
+    clear_tools,
+    execute_msg,
+    get_available_tools,
+    get_tools,
+    set_tools,
+)
 from gptme.tools.base import set_tool_format
 
 
@@ -77,35 +85,44 @@ class TestExecuteMsgDefensive:
     def test_disabled_by_default_tool_gets_enable_hint(self):
         """A disabled-by-default tool (e.g. 'read') includes '--tools +name' in error."""
         set_tool_format("tool")
-        # Ensure no tools are loaded so 'read' is not runnable in this context.
-        clear_tools()
-        # Precondition: 'read' must be discoverable and have disabled_by_default=True.
-        # If someone changes that flag this assertion fails loudly instead of silently.
-        read_specs = [
-            t for t in get_available_tools(include_mcp=False) if t.name == "read"
-        ]
-        assert read_specs, "read tool must be discoverable via get_available_tools"
-        assert read_specs[0].disabled_by_default, (
-            "read tool must have disabled_by_default=True for this test to be valid"
-        )
-        assert read_specs[0].is_available, (
-            "read tool must have is_available=True for this test to be valid "
-            "(the hint is only appended when the tool is both disabled_by_default "
-            "and is_available)"
-        )
-        call_id = "call-disabled-read"
-        # 'read' is disabled_by_default=True; it is discoverable but not loaded
-        content = f'@read({call_id}): {{"path": "somefile.txt"}}'
-        msg = Message("assistant", content)
+        # clear_tools() wipes context-local loaded tools and the available-tools
+        # cache. Snapshot and restore so later tests in this process don't inherit
+        # an empty registry (pytest shares one thread/context by default).
+        prev_tools = list(get_tools())
+        prev_cache = _get_available_tools_cache()
+        try:
+            # Ensure no tools are loaded so 'read' is not runnable in this context.
+            clear_tools()
+            # Precondition: 'read' must be discoverable and have disabled_by_default=True.
+            # If someone changes that flag this assertion fails loudly instead of silently.
+            read_specs = [
+                t for t in get_available_tools(include_mcp=False) if t.name == "read"
+            ]
+            assert read_specs, "read tool must be discoverable via get_available_tools"
+            assert read_specs[0].disabled_by_default, (
+                "read tool must have disabled_by_default=True for this test to be valid"
+            )
+            assert read_specs[0].is_available, (
+                "read tool must have is_available=True for this test to be valid "
+                "(the hint is only appended when the tool is both disabled_by_default "
+                "and is_available)"
+            )
+            call_id = "call-disabled-read"
+            # 'read' is disabled_by_default=True; it is discoverable but not loaded
+            content = f'@read({call_id}): {{"path": "somefile.txt"}}'
+            msg = Message("assistant", content)
 
-        results = list(execute_msg(msg))
+            results = list(execute_msg(msg))
 
-        assert len(results) == 1
-        result = results[0]
-        assert result.call_id == call_id
-        assert "--tools +read" in result.content, (
-            f"Expected enable hint in error for disabled-by-default tool, got: {result.content!r}"
-        )
+            assert len(results) == 1
+            result = results[0]
+            assert result.call_id == call_id
+            assert "--tools +read" in result.content, (
+                f"Expected enable hint in error for disabled-by-default tool, got: {result.content!r}"
+            )
+        finally:
+            set_tools(prev_tools)
+            _set_available_tools_cache(prev_cache)
 
     def test_multiple_structured_tooluses_all_get_error_results(self):
         """Multiple non-runnable structured tool_uses each get an error tool_result."""
--- a/poetry.lock
+++ b/poetry.lock
@@ -2615,7 +2615,7 @@ version = "3.5.4"
 description = "Lightweight in-process concurrent programming"
 optional = false
 python-versions = ">=3.10"
-groups = ["main"]
+groups = ["main", "dev"]
 files = [
     {file = "greenlet-3.5.4-cp310-cp310-macosx_11_0_universal2.whl", hash = "sha256:ac5bf81d79d2c8eeb2ef6359b2e1687a1e9ebf46c2b1f970da9a9255df51d190"},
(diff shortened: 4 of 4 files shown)


gptme/tools/__init__.py
```diff
@@ -393,9 +393,17 @@ def execute_msg(
                 "the tool_use/tool_result pairing valid.",
                 tooluse.tool,
             )
+            error_msg = f"Tool '{tooluse.tool}' is not available for execution."
+            # If the tool exists but is disabled by default, give an actionable hint.
+            available = get_available_tools(include_mcp=False)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9a8d08f0f: prefer `_get_available_tools_cache()` when it is populated, and only fall back to uncached `get_available_tools(include_mcp=False)` once per `execute_msg`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `9a8d08f0f414`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/__init__.py`

TimeToBuildBob:
## 🤖 AI code review

Solid approach for the CLI² use case: correct gating on `disabled_by_default and is_available`, accurate generic error when the tool is genuinely unavailable, and the exception-safety concern around discovery failures is handled properly (fallback to the unadorned message). The main issues are: (1) the "prefer existing cache" intent is defeated in the normal production path — the fallback intentionally bypasses the cache and never repopulates it, turning every non-runnable tool call into a full plugin scan with import side effects; (2) the synthesized remedy "enable it with --tools +X" is only truthful at CLI start and is wrong/misleading in server/session contexts (and also wrong when the tool is session-refused rather than merely "disabled by default"); (3) the test leaks global session state (tool format plus cache/reload semantics) that must be restored symmetrically. Fix the caching loop, gate the remedy on the context, and make the test's restore fully symmetric. 7/10 — "Good", two P2 functional/performance findings plus a test-hygiene gap.

Confidence 5/5

| `6db5ac3b9610` | 4/5 | 2 | llm | 2026-09-02 07:49 UTC |
| `759fbc04a21d` | 4/5 | 2 | llm | 2026-09-02 09:16 UTC |
| `9ea175a67626` | 4/5 | 3 | llm | 2026-09-02 10:08 UTC |
| `ad8626a07daa` | 4/5 | 1 | llm | 2026-09-02 11:22 UTC |
| `3a6a6b984e5f` | 4/5 | 1 | llm | 2026-09-02 11:58 UTC |
| `e7407f12387c` | 4/5 | 2 | llm | 2026-09-05 16:21 UTC |
| `7219d5191c93` | 5/5 | 0 | llm | 2026-09-07 11:48 UTC |
| `90b7b48ffccf` | 4/5 | 1 | llm | 2026-09-07 13:55 UTC |
| `32862e4aeee7` | 5/5 | 0 | llm | 2026-09-07 15:47 UTC |
| `32862e4aeee7` | 4/5 | 1 | llm | 2026-09-07 16:47 UTC |

</details>

<sub>Reviewed `32862e4aeee7` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 821s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/__init__.py
```diff
@@ -393,9 +398,26 @@ def execute_msg(
...
                 tooluse.tool,
             )
+            error_msg = f"Tool '{tooluse.tool}' is not available for execution."
+            # If the tool exists but is disabled by default, give an actionable hint.
+            if available_for_hint is None:
+                cached = _get_available_tools_cache()
+                available_for_hint = (
+                    cached
+                    if cached is not None
+                    else get_available_tools(include_mcp=False)
```
TimeToBuildBob: ⚠️ **P2** — The new error-path call to `get_available_tools(include_mcp=False)` runs when the context-local available-tools cache has not been populated. That function performs full uncached module/plugin discovery, importing every configured tool and plugin module, and nothing around this call catches failures. `_collect_tool_modules` only suppresses `ModuleNotFoundError`; a plugin module that raises at import time (for example an `ImportError` for a missing attribute, a `ValueError`, or another dependency error) will propagate out of `execute_msg`. Because this happens before the `yield Message(...)` for the non-runnable structured tool_use, no paired tool_result is emitted, leaving a dangling tool_use and reopening the API 400 failure that this defensive branch exists to prevent. The lookup should be wrapped in try/except and fall back to an empty list so the error result is always yielded.

TimeToBuildBob: Fixed in 6db5ac3b9: wrap the `get_available_tools(include_mcp=False)` lookup in try/except, falling back to an empty tool list on discovery failure so the paired error `Message` is still yielded.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `6db5ac3b9610`:

- CI gate: satisfied
- `gptme/tools/__init__.py`

tests/test_execute_msg_defensive.py
```diff
@@ -71,6 +74,23 @@ def test_error_result_role_is_system(self):
...
+        # 'read' is disabled_by_default=True; it is discoverable but not loaded
+        content = f'@read({call_id}): {{"path": "somefile.txt"}}'
+        msg = Message("assistant", content)
+
+        results = list(execute_msg(msg))
+
+        assert len(results) == 1
+        result = results[0]
+        assert result.call_id == call_id
+        assert "--tools +read" in result.content, (
```
TimeToBuildBob: ⚠️ **P2** — This test is not self-contained. Tool state lives in module-level ContextVars (_loaded_tools_var, _available_tools_var) that default to empty but are shared across every test running in the same process/thread. The test asserts that @read is non-runnable and that the error path fires, but it never establishes that precondition itself. Any earlier test in the suite that loads the real read tool (via init_tools, load_tool, set_tools, or clear_tools not being called) makes the @read call runnable; execute_msg will then actually execute read and yield the file result instead of the error branch, so the assertion that '--tools +read' appears will fail for state reasons unrelated to this change. It also silently depends on the repo's read tool having disabled_by_default=True; a refactor that flips that flag breaks the test deterministically. Unlike the other tests in this file that use fabricated names (nonexistent_tool, missing_tool), this test couples to real tool state and to an external flag. It should call clear_tools() at the start and assert the read tool's disabled_by_default flag (or use a controlled stub) so the result is deterministic regardless of collection order and tool configuration.

gptme/tools/__init__.py
```diff
@@ -393,9 +398,35 @@ def execute_msg(
...
+                    except Exception:
+                        # Discovery failures (bad plugin/config) must not block
+                        # yielding the paired tool_result below, or the dangling
+                        # tool_use/400 bug this branch exists to prevent reopens.
+                        logger.warning(
+                            "Failed to look up available tools for enable hint",
+                            exc_info=True,
+                        )
+                        available_for_hint = []
+            if any(
```
TimeToBuildBob, line 439: ⚠️ **P2** — The hint branch matches only on the tool name plus the disabled_by_default flag and ignores the tool's availability. A ToolSpec in available tools that has disabled_by_default=True and is_available=False will trigger the 'Enable it with --tools +<name>' suggestion even though enabling it cannot work: get_toolchain skips unavailable tools entirely (if not tool.is_available: continue), and init_tools would only warn-and-skip when the tool is matched but unavailable. The system tool_result would then advise a fix that has no effect, which can send the model (and the user) into repeated failed enable attempts. Additionally, the match is strictly name-based (t.name == tooluse.tool), so a disabled-by-default tool referenced through a supported block-format alias (block types are a first-class lookup path, see get_tool_for_langtag) will be silently dropped from the hint and keep only the generic message. Gating the hint on t.is_available, and matching against block_types/hints in addition to name, would keep the suggestion honest and cover the alias calling convention.

tests/test_execute_msg_defensive.py
TimeToBuildBob: Fixed in 759fbc04a21d — `clear_tools()` is now called at the start of the test to ensure the `read` tool is not loaded in the current ContextVar state. Added a precondition assertion that `read` is discoverable via `get_available_tools(include_mcp=False)` with `disabled_by_default=True`, so a flag change fails loudly rather than silently altering the test's behaviour.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in 759fbc04a21d — the hint condition now requires `t.is_available` in addition to `t.disabled_by_default`, so the "Enable it with --tools +name" suggestion is only emitted when enabling the tool is actually sufficient to make it runnable. Tools that are disabled by default but also unavailable keep the original generic error message.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `759fbc04a21d`:

- CI gate: satisfied
- `gptme/tools/__init__.py`

gptme/tools/__init__.py
```diff
@@ -393,9 +398,35 @@ def execute_msg(
...
             )
+            error_msg = f"Tool '{tooluse.tool}' is not available for execution."
+            # If the tool exists but is disabled by default, give an actionable hint.
+            if available_for_hint is None:
+                cached = _get_available_tools_cache()
+                if cached is not None:
+                    available_for_hint = cached
+                else:
+                    try:
+                        available_for_hint = get_available_tools(include_mcp=False)
```
TimeToBuildBob, line 429: ⚠️ **P2** — When the per-context available-tools cache is cold (a freshly started server task, or any context where get_available_tools() with default include_mcp=True has not run yet), the new not-runnable error path falls through to `get_available_tools(include_mcp=False)`. That call never reads, and never writes, the context cache — the source shows `_get_available_tools_cache() if include_mcp else None` (so None here) and the include_mcp=False branch returns before _set_available_tools_cache. That means every cold-cache message containing an un-runnable structured tool_use now runs full synchronous module discovery (_discover_tools imports every TOOL_MODULES and plugin module, plus config and plugin-registry reads) on what was previously a pure error-emission path. available_for_hint is memoized only across tool_uses within a single message, not across calls, so a persistently-cold context repeats this work per message, adds latency to a generator the API caller is blocked on, and can also surface spurious 'module not found' warnings from _discover_tools for optional modules. Normal CLI flow warms the cache via init_tools → get_available_tools(), but server contexts that clear_tools() per request do not.

tests/test_execute_msg_defensive.py
```diff
@@ -71,6 +74,34 @@ def test_error_result_role_is_system(self):
...
+        set_tool_format("tool")
+        # Ensure no tools are loaded so 'read' is not runnable in this context.
+        clear_tools()
+        # Precondition: 'read' must be discoverable and have disabled_by_default=True.
+        # If someone changes that flag this assertion fails loudly instead of silently.
+        read_specs = [
+            t for t in get_available_tools(include_mcp=False) if t.name == "read"
+        ]
+        assert read_specs, "read tool must be discoverable via get_available_tools"
+        assert read_specs[0].disabled_by_default, (
```
TimeToBuildBob: ⚠️ **P2** — The new test asserts the hint appears for a non-loaded 'read' but never asserts the production gating precondition `read.is_available`. Under the implemented logic (lines 419-421), the hint is appended only when the disabled-by-default tool's availability probe returns True. The test therefore fails spuriously in any environment where 'read's availability probe returns False (e.g., an optional dependency missing at probe time), even though the behavior being locked in - "a discoverable disabled-by-default tool gets an enable hint" - is exactly what it should verify regardless. Conversely, the test passing would give false confidence while the production code silently withholds the enable hint for every disabled-by-default tool whose probe returns False, directly contradicting the PR's stated goal of telling users to enable a disabled-by-default tool whenever it exists. The test should assert `read_specs[0].is_available` as a precondition or the availability gate should be dropped so the hint is driven purely by the disabled-by-default state.

TimeToBuildBob: Fixed in 9ea175a67 — added an explicit `read_specs[0].is_available` precondition assertion alongside the existing `disabled_by_default` check, so a future environment where `read`'s availability probe returns `False` fails the test loudly with a clear message instead of silently passing/failing on an unrelated code path.

gptme/tools/__init__.py
TimeToBuildBob: Acknowledged as a real trade-off, but leaving it as-is rather than changing it in this PR:

- The fallback only runs when the *per-context* `include_mcp=True` cache is cold, which the comment above it explains is not the normal CLI path (that warms the cache via `init_tools()` before the conversation loop starts).
- It's already memoized per `execute_msg()` call via the local `available_for_hint` var, so a message with multiple non-runnable structured tool_uses does at most one discovery, not one per tool_use — this was the original Greptile P2 (fixed in 9a8d08f0f).
- I checked whether we could drop the fallback entirely and rely on the cache only (avoiding discovery altogether when cold), but that regresses `test_disabled_by_default_tool_gets_enable_hint`, which explicitly `clear_tools()`s to simulate a cold context — i.e. dropping the fallback means server contexts that clear tools per request would silently lose the hint this PR exists to add, which is worse than the perf cost on this rare error-only path (a structured tool_use naming a tool that isn't loaded/available).
- A proper fix (e.g. caching the `include_mcp=False` result too) would mean the cache no longer "always includes MCP" per its own contract, and risks masking MCP tools from later `include_mcp=True` callers — a bigger change than this PR's scope (adding a UX hint string).

Given the code path is bounded (once per message, only when the tool is unrunnable) and guarded (try/except already added in 6db5ac3b9), I'm treating this as an acceptable trade-off rather than something to fix here. Open to revisiting if it shows up as a real hot path in practice.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `9ea175a67626`:

- CI gate: satisfied
- `gptme/tools/__init__.py`

tests/test_execute_msg_defensive.py
```diff
@@ -71,6 +74,39 @@ def test_error_result_role_is_system(self):
             f"Expected role='system', got '{results[0].role}'"
         )
 
+    def test_disabled_by_default_tool_gets_enable_hint(self):
+        """A disabled-by-default tool (e.g. 'read') includes '--tools +name' in error."""
+        set_tool_format("tool")
+        # Ensure no tools are loaded so 'read' is not runnable in this context.
+        clear_tools()
```
TimeToBuildBob: ⚠️ **P2** — The new test calls clear_tools() mid-suite, which resets the context-local loaded-tools list to [] and sets the available-tools ContextVar cache to None (lines 516-517). Because pytest runs test classes sequentially in one context/thread by default, any subsequently-collected test in another file that relies on tools having already been initialized in this context — without re-running init_tools() or a fixture that restores them — will observe an empty loaded-tools list and an empty cache, producing false failures or forcing expensive re-discovery. Even within this file, tests defined after test_disabled_by_default_tool_gets_enable_hint run in a context where loaded tools are wiped, coupling their behavior to this test's side effect.

gptme/tools/__init__.py
```diff
@@ -393,9 +398,35 @@ def execute_msg(
...
+                            exc_info=True,
+                        )
+                        available_for_hint = []
+            if any(
+                t.name == tooluse.tool and t.disabled_by_default and t.is_available
+                for t in available_for_hint
+            ):
+                error_msg += (
+                    f" This tool is disabled by default."
+                    f" Enable it with '--tools +{tooluse.tool}'."
```
TimeToBuildBob, line 445: ⚠️ **P2** — The appended remedy is unconditional CLI syntax ('--tools +name'), but init_tools sources the allowlist from three independent places in precedence order: the TOOL_ALLOWLIST env var (line 181 of this file), config.chat.tools (line 185), or the allowlist argument that the --tools flag populates. When a deployment pins tools via TOOL_ALLOWLIST or chat.tools, telling the user to re-run with '--tools +read' either has no effect (the env/config value is used and the addition is dropped) or overrides their existing pinned list, so the advertised fix does not enable anything and the user may even lose other intentionally loaded tools by following it. The hint should reference the actually active allowlist source (env var vs config vs flag) instead of unconditionally recommending the flag.

gptme/tools/__init__.py
TimeToBuildBob: Accepted trade-off — leaving the cold-cache fallback as-is.

The fallback only runs when the per-context `include_mcp=True` cache is cold. The normal CLI path warms that cache via `init_tools()` before the conversation loop starts, so this is not on the hot path.

Caching the `include_mcp=False` result into the MCP-inclusive cache would be incorrect (the cache is documented as always including MCP tools). Adding a second cache just to speed up this rare error-only path is more machinery than the hint is worth.

Discovery is already memoized once per `execute_msg` via `available_for_hint`, wrapped in try/except so a plugin import failure still yields the paired error `tool_result`, and skipped entirely when the warm cache is present. That is the intended bound.

Not changing it in this PR.

gptme/tools/__init__.py
TimeToBuildBob: Won't-fix — the CLI `--tools +name` hint is the documented, primary way to enable a disabled-by-default tool.

`TOOL_ALLOWLIST` and `config.chat.tools` are deployment pins that take precedence over the flag. A user in a pinned deployment who hits this error is almost never going to be unblocked by a source-aware hint: if the pin omitted `read`, they need to change the pin, not be told three equivalent syntaxes. Expanding the error to mention env/config/flag would make a worse message for the common CLI case this PR is aimed at.

Making `execute_msg`'s error path inspect env/config to pick a hint also duplicates `init_tools` precedence on a path that exists to emit a paired `tool_result`, not to teach allowlist configuration.

The CLI flag remains the right default advice. Pinned deployments already have their own config surface.

tests/test_execute_msg_defensive.py
TimeToBuildBob: Fixed in ad8626a07 — the test now snapshots `get_tools()` and the available-tools cache before `clear_tools()`, then restores both in a `finally` block so later tests in the same pytest process don't inherit an empty registry.

gptme/tools/__init__.py
```diff
@@ -393,9 +398,35 @@ def execute_msg(
...
+            error_msg = f"Tool '{tooluse.tool}' is not available for execution."
+            # If the tool exists but is disabled by default, give an actionable hint.
+            if available_for_hint is None:
+                cached = _get_available_tools_cache()
+                if cached is not None:
+                    available_for_hint = cached
+                else:
+                    try:
+                        available_for_hint = get_available_tools(include_mcp=False)
+                    except Exception:
```
TimeToBuildBob, line 430: ⚠️ **P2** — 在 `execute_msg` 的非可运行错误分支,当 `_get_available_tools_cache()` 为 `None` 时,使用 `get_available_tools(include_mcp=False)` 解析 `available_for_hint`。`get_available_tools` 在 `include_mcp=False` 时强制将 `available_tools` 置为 `None`(忽略缓存),并且不设置缓存;更关键的是,`create_mcp_tools` 仅在 `include_mcp=True` 时运行,所以 `include_mcp=False` 的结果**从不**包含 MCP 工具。这导致不一致:如果某个 MCP 工具(它也可以设置 `disabled_by_default=True`)在缓存为空时(例如刚调用过 `clear_tools()`,或 MCP 初始化尚未发生)被调用,`available_for_hint` 将缺少该 MCP 工具,因而不会追加 `--tools +<name>` 提示;但如果缓存已经由先前的 `include_mcp=True` 发现填充(如 `init_tools` 的调用),`available_for_hint` 会复用包含 MCP 工具的缓存,提示就会出现。因此,用户是否看到“启用”提示取决于缓存状态,而不是工具本身的属性。此外,当缓存为空时,`get_available_tools(include_mcp=False)` 仍会执行完整的本地发现(导入工具模块),这带来与只添加文本提示不相关的开销和潜在插件导入副作用。建议:当缓存可用时始终使用缓存;当缓存为空时,要么避免非必要的工具发现(直接保持基础错误消息),要么使用 `include_mcp=True` 并容忍失败,以确保 MCP 工具的一致处理。 _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: False positive — the described inconsistency can't actually occur.

The finding assumes an MCP tool could have `disabled_by_default=True`, so cache
warmth (which determines whether MCP tools are included in `available_for_hint`)
would change whether the enable hint is shown for it. But MCP-derived `ToolSpec`s
never set `disabled_by_default`: `create_mcp_tools` in `gptme/tools/mcp_adapter.py`
constructs each MCP `ToolSpec` without passing `disabled_by_default`, and the
field defaults to `False` (`gptme/tools/base.py`). There is currently no code
path that produces an MCP tool with `disabled_by_default=True`.

Since the hint condition is `t.disabled_by_default and t.is_available`, an MCP
tool can never satisfy it regardless of whether `available_for_hint` came from
the warm `include_mcp=True` cache or the cold `include_mcp=False` fallback. The
cache-warmth difference is real, but it has no observable effect here because
the set of tools it could affect (MCP tools with `disabled_by_default=True`) is
always empty.

TimeToBuildBob:
Automated merge handoff for `3a6a6b984e5f`:

- CI gate: satisfied
- Review threads: clear
- `gptme/tools/__init__.py`
- `pyproject.toml`

TimeToBuildBob:
@greptileai review

pyproject.toml
```diff
@@ -108,6 +108,9 @@ matplotlib = {version = "*", optional=true}
 pandas = {version = "*", optional=true}
 numpy = {version = "*", optional=true}
 
+# testing (needed for coverage concurrency tracking)
+greenlet = "*"  # required by coverage for concurrency=greenlet tracking
```
TimeToBuildBob, line 112: ⚠️ **P2** — The pyproject change moves `greenlet = "*"` from the dev dependency group into the runtime `[tool.poetry.dependencies]` section, with a comment saying it is needed for coverage concurrency tracking. Greenlet is only needed when coverage runs with `concurrency = ["greenlet"]`, which is a test/coverage scenario, not a normal runtime use of gptme. Moving it into core dependencies forces every production install of gptme to pull greenlet even though nothing in the runtime code imports it. This unnecessarily increases the install footprint and makes a testing-only dependency part of the supported runtime dependency set. It should remain in the dev dependency group, or be made an optional extra, rather than being added to the base runtime dependencies.

TimeToBuildBob:
Rebased onto current master to resolve the merge conflict in `gptme/tools/__init__.py`. The conflict was straightforward: upstream had changed the execute-loop from `for tooluse, runnable in remaining` to `for tooluse in remaining: runnable = tooluse.is_runnable`, so I applied the `available_for_hint` cache variable from this PR's commit alongside the new loop structure. CI running.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `e7407f12387c`:

- CI gate: satisfied
- `gptme/tools/__init__.py`
- `pyproject.toml`

TimeToBuildBob:
Trimmed per today's triage: dropped the unrelated `poetry.lock`/`pyproject.toml` hunk that moved `greenlet` from the dev group to main dependencies — a CI cherry-pick with no relation to the disabled-by-default tool hint this PR is about. Reverted cleanly; the diff is now just the tool-hint fix and its test. Ready for maintainer review.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

tests/test_execute_msg_defensive.py
```diff
@@ -40,6 +48,9 @@ def test_nonrunnable_structured_tooluse_yields_error_result(self):
         assert "nonexistent_tool" in result.content, (
             "Error message should identify the tool that was unavailable"
         )
+        assert "disabled by default" not in result.content, (
```
TimeToBuildBob, line 51: ⚠️ **P2** — The diff only adds a negative assertion to the existing test for an unknown tool. It never exercises the positive behavior that this PR is supposed to add — that a disabled-by-default tool like `read` produces the 'This tool is disabled by default.' hint. The PR description claims a new `test_disabled_by_default_tool_gets_enable_hint` was added, but it is not present in this diff. Without a positive test, the new hint logic could be completely broken and the suite would still pass because the only added assertion checks that an unknown tool does NOT get the hint.

TimeToBuildBob: False positive — the positive regression test is present at `tests/test_execute_msg_defensive.py::TestExecuteMsgDefensive::test_disabled_by_default_tool_gets_enable_hint`. The reviewer inspected line 51 but missed the method starting at line 85. It clears the loaded registry, confirms `read` is discoverable, disabled by default, and available, invokes structured `@read(...)`, then asserts the emitted result contains `--tools +read`.

I also reran the whole defensive test file at the current head: 5 passed.

TimeToBuildBob:
Current head `32862e4aeee7` is ready for maintainer review and manual merge:

- CI: all checks pass
- Greptile: 5/5; its reviewed files are byte-identical on this head
- Bob review: sole P2 was a false positive; the claimed-missing positive test exists and passes, and its thread is resolved
- Diff: only `gptme/tools/__init__.py` plus the regression test

Self-merge remains blocked by path policy for `gptme/tools/__init__.py`; I did not bypass that gate.

gptme/tools/__init__.py
```diff
@@ -413,9 +418,35 @@ def execute_msg(
...
+                        # Discovery failures (bad plugin/config) must not block
+                        # yielding the paired tool_result below, or the dangling
```
TimeToBuildBob, line 432: ⚠️ **P2** — available_for_hint is only kept in a local variable; the discovery result is never written back through _set_available_tools_cache. Since get_available_tools(include_mcp=False) deliberately skips the cache (per the code comment), every separate execute_msg call whose first not-runnable structured tool_use finds the cache empty re-runs the full uncached lookup — re-importing modules and re-running plugin discovery. The reuse is only loop-local (one discovery per message); nothing amortizes cost across messages. In a server session where the registry/cache was cleared (clear_tools wipes the cache) or before cache population, a model repeatedly invoking an unavailable tool pays full discovery cost once per message, which is exactly the expense the comment says the cache preference is meant to avoid. Guardedly storing the discovered list into the cache via _set_available_tools_cache would let subsequent messages reuse it.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
TimeToBuildBob: Same accepted trade-off as `4b15ab1df884` on this PR (disposed 2026-09-02). The new fingerprint restates that finding: cold-cache `include_mcp=False` discovery is not written back through `_set_available_tools_cache`.

Not changing it in this PR:

- `_set_available_tools_cache` is documented to hold the MCP-inclusive list. Writing the `include_mcp=False` result into it would make later `include_mcp=True` callers miss MCP tools.
- The fallback only runs when the per-context cache is cold. The normal CLI path warms it via `init_tools()` before the conversation loop. Server contexts that `clear_tools()` per request are the rare case.
- Discovery is already memoized once per `execute_msg` via `available_for_hint`, wrapped in try/except so a plugin import failure still yields the paired error `tool_result`, and skipped entirely when the warm cache is present.

A second cache just to speed up this error-only path is more machinery than the hint is worth. Open to revisiting if it shows up as a real hot path.

TimeToBuildBob:
Superseded by gptme/gptme#3748, merged to master at `9a0c2408a` (closes #3698).

This PR is DIRTY because both branches edit the not-runnable error path in `gptme/tools/__init__.py`. Rebasing would stack a second, more complex variant of the same hint on top of the already-landed cache-only helper. The extras here (`is_available` gating, cold-cache rediscovery fallback) are not worth keeping this PR open — the issue is done.

Closing without rebase.

PR #3714, opened 2026-09-04
fix(logmanager): bound trajectory write amplification

TimeToBuildBob:
## Problem

Long gptme sessions rewrite the complete `conversation.jsonl` after every appended message. The recovery event log also accumulates every duplicated append plus full checkpoints forever. On five live Bob runs, the leaf gptme processes wrote 91-894 MB while final trajectory directories held 1.8-4.7 MB; checkpoint records alone were 0.63-1.44 MB per sampled run.

## Changes

- append only newly-added conversation messages during the normal append path
- retain full rewrite behavior after edit/undo or when a fresh `Log` encounters an existing file
- compact recovery events atomically after each checkpoint, retaining the latest full checkpoint plus its replay tail
- preserve branch and compacted-view persistence state independently

## Verification

- `pytest tests/test_logmanager.py tests/test_eventlog.py tests/test_auto_compact.py`: 120 passed
- `pytest tests/test_conversations.py tests/test_cli_from_turn.py tests/test_tools_subagent.py`: 156 passed
- Ruff check/format clean
- scoped `prek` hooks, including mypy, pass
- pre-PR quality gate: 100/100

The existing trajectory filenames/layout are unchanged, so current readers and backup globs continue to apply.

### DIFF
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -14,6 +14,7 @@ import logging
 import os
 import tempfile
 import threading
+import weakref
 from contextlib import contextmanager, suppress
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
@@ -34,7 +35,10 @@ try:
 except ImportError:  # pragma: no cover - POSIX
     msvcrt = None
 
-_event_log_thread_lock = threading.Lock()
+_event_log_thread_locks_guard = threading.Lock()
+_event_log_thread_locks: weakref.WeakValueDictionary[Path, threading.Lock] = (
+    weakref.WeakValueDictionary()
+)
 
 logger = logging.getLogger(__name__)
 
@@ -68,13 +72,25 @@ def _make_event(seq: int, type: str, payload: dict[str, Any]) -> dict[str, Any]:
 # ── public API ───────────────────────────────────────────────────────
 
 
+def _get_event_log_thread_lock(path: Path) -> threading.Lock:
+    """Return the in-process lock shared by users of one recovery log."""
+    key = path.resolve()
+    with _event_log_thread_locks_guard:
+        lock = _event_log_thread_locks.get(key)
+        if lock is None:
+            lock = threading.Lock()
+            _event_log_thread_locks[key] = lock
+        return lock
+
+
 @contextmanager
 def _event_log_lock(logdir: Path) -> Iterator[None]:
     """Serialize event appends and compaction for one recovery log."""
     path = _event_log_path(logdir)
     path.parent.mkdir(parents=True, exist_ok=True)
     lock_path = path.parent / f".{path.name}.lock"
-    with _event_log_thread_lock, lock_path.open("a+b") as lock:
+    thread_lock = _get_event_log_thread_lock(path)
+    with thread_lock, lock_path.open("a+b") as lock:
         if fcntl is not None:
             fcntl.flock(lock, fcntl.LOCK_EX)
         elif msvcrt is not None:  # pragma: no cover - Windows
@@ -118,6 +134,39 @@ def append_next_event(
     return event
 
 
+def _compact_events_unlocked(logdir: Path) -> None:
+    """Drop events superseded by the latest checkpoint; caller holds lock."""
+    path = _event_log_path(logdir)
+    events = read_events(logdir)
+    checkpoint = find_latest_checkpoint(events)
+    if checkpoint is None:
+        return
+
+    retained = [event for event in events if event["seq"] >= checkpoint["seq"]]
+    fd, tmp_name = tempfile.mkstemp(
+        dir=path.parent, prefix=f".{path.name}.", suffix=".tmp"
+    )
+    try:
+        try:
+            tmp_file = os.fdopen(fd, "w", encoding="utf-8")
+        except BaseException:
+            # fdopen did not take ownership, so the raw descriptor is ours.
+            os.close(fd)
+            raise
+        # From here the file object owns fd and closes it exactly once; never
+        # close fd directly, or a concurrently opened file reusing the number
+        # gets closed instead.
+        with tmp_file as f:
+            f.writelines(json.dumps(event, default=str) + "\n" for event in retained)
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp_name, path)
+    except BaseException:
+        with suppress(FileNotFoundError):
+            os.unlink(tmp_name)
+        raise
+
+
 def compact_events(logdir: Path) -> None:
     """Drop events superseded by the latest checkpoint.
 
@@ -126,38 +175,8 @@ def compact_events(logdir: Path) -> None:
     and replacement is atomic, so neither partial files nor lost appends are
     observable.
     """
-    path = _event_log_path(logdir)
     with _event_log_lock(logdir):
-        events = read_events(logdir)
-        checkpoint = find_latest_checkpoint(events)
-        if checkpoint is None:
-            return
-
-        retained = [event for event in events if event["seq"] >= checkpoint["seq"]]
-        fd, tmp_name = tempfile.mkstemp(
-            dir=path.parent, prefix=f".{path.name}.", suffix=".tmp"
-        )
-        try:
-            try:
-                tmp_file = os.fdopen(fd, "w", encoding="utf-8")
-            except BaseException:
-                # fdopen did not take ownership, so the raw descriptor is ours.
-                os.close(fd)
-                raise
-            # From here the file object owns fd and closes it exactly once; never
-            # close fd directly, or a concurrently opened file reusing the number
-            # gets closed instead.
-            with tmp_file as f:
-                f.writelines(
-                    json.dumps(event, default=str) + "\n" for event in retained
-                )
-                f.flush()
-                os.fsync(f.fileno())
-            os.replace(tmp_name, path)
-        except BaseException:
-            with suppress(FileNotFoundError):
-                os.unlink(tmp_name)
-            raise
+        _compact_events_unlocked(logdir)
 
 
 def read_events(logdir: Path) -> list[dict[str, Any]]:
@@ -203,16 +222,18 @@ def sequence_number(logdir: Path) -> int:
 
 def write_checkpoint(
     logdir: Path,
-    seq: int,
     messages: list[dict[str, Any]],
 ) -> dict[str, Any]:
-    """Write a checkpoint event and compact superseded recovery history.
+    """Assign, write, and compact a checkpoint under one lock.
 
     Returns the written checkpoint event.
     """
-    event = _make_event(seq, EVENT_CHECKPOINT, {"messages": messages})
-    append_event(logdir, event)
-    compact_events(logdir)
+    with _event_log_lock(logdir):
+        event = _make_event(
+            sequence_number(logdir), EVENT_CHECKPOINT, {"messages": messages}
+        )
+        _append_event_unlocked(logdir, event)
+        _compact_events_unlocked(logdir)
     return event
 
 
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -78,9 +78,12 @@ def _event_log_lock(logdir: Path) -> Iterator[None]:
         if fcntl is not None:
             fcntl.flock(lock, fcntl.LOCK_EX)
         elif msvcrt is not None:  # pragma: no cover - Windows
-            lock.seek(0)
-            lock.write(b"\0")
-            lock.flush()
+            # Append mode is needed to atomically create the sidecar, but Windows
+            # append semantics ignore seek() for writes.  Only initialize an empty
+            # sidecar so repeated acquisitions do not grow it indefinitely.
+            if lock.seek(0, os.SEEK_END) == 0:
+                lock.write(b"\0")
+                lock.flush()
             lock.seek(0)
             msvcrt.locking(lock.fileno(), msvcrt.LK_LOCK, 1)
         try:
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -74,7 +74,8 @@ def _event_log_lock(logdir: Path) -> Iterator[None]:
     path = _event_log_path(logdir)
     path.parent.mkdir(parents=True, exist_ok=True)
     lock_path = path.parent / f".{path.name}.lock"
-    with _event_log_thread_lock, lock_path.open("a+b") as lock:
+    lock_path.touch(exist_ok=True)
+    with _event_log_thread_lock, lock_path.open("r+b") as lock:
         if fcntl is not None:
             fcntl.flock(lock, fcntl.LOCK_EX)
         elif msvcrt is not None:  # pragma: no cover - Windows
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -10,6 +10,8 @@ from __future__ import annotations
 
 import json
 import logging
+import os
+import tempfile
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 
@@ -58,6 +60,41 @@ def append_event(logdir: Path, event: dict[str, Any]) -> None:
         f.write(json.dumps(event, default=str) + "\n")
 
 
+def compact_events(logdir: Path) -> None:
+    """Drop events superseded by the latest checkpoint.
+
+    Checkpoints contain the full message state, so events before the newest one
+    are redundant for recovery.  Replace the log atomically to avoid exposing a
+    partially rewritten recovery source after an interrupted write.
+    """
+    path = _event_log_path(logdir)
+    events = read_events(logdir)
+    checkpoint = find_latest_checkpoint(events)
+    if checkpoint is None:
+        return
+
+    retained = [event for event in events if event["seq"] >= checkpoint["seq"]]
+    fd, tmp_name = tempfile.mkstemp(
+        dir=path.parent, prefix=f".{path.name}.", suffix=".tmp"
+    )
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.writelines(json.dumps(event, default=str) + "\n" for event in retained)
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp_name, path)
+    except BaseException:
+        try:
+            os.close(fd)
+        except OSError:
+            pass
+        try:
+            os.unlink(tmp_name)
+        except FileNotFoundError:
+            pass
+        raise
+
+
 def read_events(logdir: Path) -> list[dict[str, Any]]:
     """Read all events from the event log, oldest first.
 
@@ -104,12 +141,13 @@ def write_checkpoint(
     seq: int,
     messages: list[dict[str, Any]],
 ) -> dict[str, Any]:
-    """Write a checkpoint event that snapshots the full message state.
+    """Write a checkpoint event and compact superseded recovery history.
 
     Returns the written checkpoint event.
     """
     event = _make_event(seq, EVENT_CHECKPOINT, {"messages": messages})
     append_event(logdir, event)
+    compact_events(logdir)
     return event
 
 
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -19,7 +19,7 @@ from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import Callable, Iterator
     from pathlib import Path
 
     from ..message import Message
@@ -96,11 +96,26 @@ def _event_log_lock(logdir: Path) -> Iterator[None]:
                 msvcrt.locking(lock.fileno(), msvcrt.LK_UNLCK, 1)
 
 
+def _append_event_unlocked(logdir: Path, event: dict[str, Any]) -> None:
+    path = _event_log_path(logdir)
+    with open(path, "a", encoding="utf-8") as f:
+        f.write(json.dumps(event, default=str) + "\n")
+
+
 def append_event(logdir: Path, event: dict[str, Any]) -> None:
     """Append a single event record to the event log."""
-    path = _event_log_path(logdir)
-    with _event_log_lock(logdir), open(path, "a", encoding="utf-8") as f:
-        f.write(json.dumps(event, default=str) + "\n")
+    with _event_log_lock(logdir):
+        _append_event_unlocked(logdir, event)
+
+
+def append_next_event(
+    logdir: Path, build_event: Callable[[int], dict[str, Any]]
+) -> dict[str, Any]:
+    """Assign the next sequence and append an event under one lock."""
+    with _event_log_lock(logdir):
+        event = build_event(sequence_number(logdir))
+        _append_event_unlocked(logdir, event)
+    return event
 
 
 def compact_events(logdir: Path) -> None:
--- a/gptme/logmanager/eventlog.py
+++ b/gptme/logmanager/eventlog.py
@@ -14,7 +14,7 @@ import logging
 import os
 import tempfile
 import threading
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 
@@ -123,7 +123,16 @@ def compact_events(logdir: Path) -> None:
             dir=path.parent, prefix=f".{path.name}.", suffix=".tmp"
         )
         try:
-            with os.fdopen(fd, "w", encoding="utf-8") as f:
+            try:
+                tmp_file = os.fdopen(fd, "w", encoding="utf-8")
+            except BaseException:
+                # fdopen did not take ownership, so the raw descriptor is ours.
+                os.close(fd)
+                raise
+            # From here the file object owns fd and closes it exactly once; never
+            # close fd directly, or a concurrently opened file reusing the number
+            # gets closed instead.
+            with tmp_file as f:
                 f.writelines(
                     json.dumps(event, default=str) + "\n" for event in retained
                 )
@@ -131,14 +140,8 @@ def compact_events(logdir: Path) -> None:
                 os.fsync(f.fileno())
             os.replace(tmp_name, path)
         except BaseException:
-            try:
-                os.close(fd)
-            except OSError:
-                pass
-            try:
+            with suppress(FileNotFoundError):
                 os.unlink(tmp_name)
-            except FileNotFoundError:
-                pass
             raise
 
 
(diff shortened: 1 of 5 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/logmanager/manager.py
```diff
@@ -103,11 +104,21 @@ def read_jsonl(cls, path: PathLike, limit=None) -> "Log":
...
-    def write_jsonl(self, path: PathLike) -> None:
-        with open(path, "w", encoding="utf-8") as file:
-            file.writelines(json.dumps(msg.to_dict()) + "\n" for msg in self.messages)
+    def write_jsonl(self, path: PathLike) -> "Log":
+        output = Path(path)
+        append_safe = self.persisted_messages <= len(self.messages) and (
+            self.persisted_messages > 0 or not output.exists()
+        )
+        mode = "a" if append_safe else "w"
+        start = self.persisted_messages if append_safe else 0
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 52f61494a: new branch and edit/undo backup logs reset their persistence cursor before writing to a fresh destination. Regression tests now reload both files and assert the inherited history is complete.

gptme/logmanager/eventlog.py
```diff
@@ -58,6 +60,41 @@ def append_event(logdir: Path, event: dict[str, Any]) -> None:
...
+    retained = [event for event in events if event["seq"] >= checkpoint["seq"]]
+    fd, tmp_name = tempfile.mkstemp(
+        dir=path.parent, prefix=f".{path.name}.", suffix=".tmp"
+    )
+    try:
+        with os.fdopen(fd, "w", encoding="utf-8") as f:
+            f.writelines(json.dumps(event, default=str) + "\n" for event in retained)
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp_name, path)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 52f61494a: append and compaction now share a stable per-log sidecar lock, so an append cannot land between the compaction snapshot and atomic replacement. Added concurrent append/compaction coverage.

gptme/logmanager/eventlog.py
```diff
@@ -50,14 +54,67 @@ def _make_event(seq: int, type: str, payload: dict[str, Any]) -> dict[str, Any]:
 # ── public API ───────────────────────────────────────────────────────
 
 
+@contextmanager
+def _event_log_lock(logdir: Path) -> Iterator[None]:
+    """Serialize event appends and compaction for one recovery log."""
+    import fcntl
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 943ac9eab and 0ae2be1ac: locking now uses guarded fcntl/msvcrt imports with a Windows byte-range fallback, and the sidecar is opened without append semantics so it remains one byte. Added a fallback regression test.

gptme/logmanager/manager.py
```diff
@@ -506,9 +519,11 @@ def write(self, branches=True, sync=False) -> None:
         # When on a view, conversation.jsonl must always contain the full main
         # branch history — the view is persisted separately in views/ directory.
         if self.current_view is not None:
-            self._branches["main"].write_jsonl(self.logfile)
+            self._branches["main"] = self._branches["main"].write_jsonl(
+                self.logfile, append=True
+            )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f15133030: view writes now persist main directly to conversation.jsonl instead of self.logfile. Added a regression covering a compacted view active while current_branch is non-main and asserting all three destinations independently.

TimeToBuildBob:
CI failure on `f151330` root-caused and fixed in `3be8ba0`.

**Root cause (real regression, not a flake).** The append fast-path decided it could append based on message *count* alone. `_attach_tool_timings` edits an already-persisted message in place:

```python
manager.log.messages[target_idx] = updated_msg
manager.write()
```

The count is unchanged, so `start == len(messages)` and `write_jsonl(append=True)` wrote **zero lines** — the edit never reached disk. That accounts for all three failures:

- `tests/test_timings.py::test_attach_tool_timings_concurrent_threads_do_not_lose_updates` (`metadata is None`)
- `tests/test_server_v2_auto_stepping.py::test_auto_stepping`
- `tests/test_server_v2_auto_stepping.py::test_multi_tool_per_message`

Confirmed PR-specific: the two most recent `master` failures are an unrelated test (`test_split_commands_bash_reserved_words_with_shell`).

**Fix.** Track the persisted prefix as the exact `Message` objects last written and compare by **identity**, not count. `Message` is a frozen dataclass, so any edit produces a new object — identity catches in-place edits, truncation, and reordering alike, while pure appends still take the append path.

**Verification** (worktree, py3.12):
- `test_logmanager.py test_timings.py test_eventlog.py` → 77 passed
- `test_server_v2_auto_stepping.py` → 4 passed (was 2 failed)
- Reverting only `manager.py` reproduces both failures → the new regression test `test_write_jsonl_rewrites_after_in_place_message_edit` is genuinely RED without the fix
- Append optimization still engages end-to-end through `LogManager` (mode `a`, no duplicated lines)
- ruff + mypy clean on the touched files

TimeToBuildBob:
CI on `3be8ba0`: **all three original failures are fixed.**

`test_attach_tool_timings_concurrent_threads_do_not_lose_updates`, `test_auto_stepping`, and `test_multi_tool_per_message` all pass now. 15 of 16 checks green (typecheck, lint, pre-commit, Docker, PyInstaller, Windows slice, `Test without API keys or extras` — all pass).

`Test without API keys` is still red, but on a **different and pre-existing** failure: 6 tests in `tests/test_tools_shell.py` hitting `ValueError: filedescriptor out of range in select()`. Same signature already fails on `master` ([33885146088](https://github.com/gptme/gptme/actions/runs/33885146088), [33877865588](https://github.com/gptme/gptme/actions/runs/33877865588)) and touches no code this PR changes.

Root cause: `select.select()` cannot represent a descriptor `>= FD_SETSIZE` (1024) and raises rather than degrading; under `-n 16` the shell tool's pty fds cross that line, which is why it looks flaky and moves between tests. Filed separately as #3715 with the three `shell.py` call sites and a fix direction (`selectors`/`poll`) — deliberately not fixed here, since it has nothing to do with logmanager and doesn't belong in this diff.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI verdict on \`3be8ba04c\`: the three failures this PR introduced are **fixed and green**. The one remaining red check, \`Test without API keys\`, is a **pre-existing repo-wide failure unrelated to this PR** — the same signature fails on \`master\` (runs 33885146088, 33877865588) and touches nothing logmanager-related:

```
```

Root cause and fix are now up as gptme/gptme#3716 — \`select()\` is backed by \`fd_set\`, which can't represent an fd `>= FD_SETSIZE` (1024) and raises instead of degrading; under \`pytest -n 16\` the shell tool's pipes cross that line, which is why it presents as flaky and moves between tests run to run. #3716 swaps the three call sites to \`poll()\`.

Once #3716 lands, this branch should go fully green on a rerun. Tracking issue: #3715.

TimeToBuildBob:
@greptileai review

gptme/logmanager/eventlog.py
```diff
@@ -50,14 +68,77 @@ def _make_event(seq: int, type: str, payload: dict[str, Any]) -> dict[str, Any]:
 # ── public API ───────────────────────────────────────────────────────
 
 
+@contextmanager
+def _event_log_lock(logdir: Path) -> Iterator[None]:
+    """Serialize event appends and compaction for one recovery log."""
+    path = _event_log_path(logdir)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    lock_path = path.parent / f".{path.name}.lock"
+    with _event_log_thread_lock, lock_path.open("a+b") as lock:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 16002a292: append mode still creates the sidecar atomically, but the Windows path now writes the sentinel only when the file is empty. The regression test acquires the lock twice and asserts the sidecar remains exactly one byte.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI verdict on `16002a292` (run [33924063178](https://github.com/gptme/gptme/actions/runs/33924063178)): **not a regression from this PR.** Two failures, both in `tests/test_server_v2_auto_stepping.py`, and the captured stdout names the cause directly:

```
Assistant:
```shell
mkdir -p /tmp/test_dir
```
System: ❌ Error during execution: filedescriptor out of range in select()
```

Same `FD_SETSIZE` class as gptme/gptme#3715 — the shell tool never ran the command, so the assertions about created files fail downstream. Worth noting for #3715: the flake is **not confined to `tests/test_tools_shell.py`**. It surfaces anywhere a test asserts on a shell tool's side effects, which is why the failing test names keep moving between runs. gptme/gptme#3716 (green, `poll()` instead of `select()`) is still the fix; this branch stays red until that lands.

Separately, while re-reading the diff I found and fixed a real defect in this PR's own code — pushed as `772766fb8`:

`compact_events()` handed the `mkstemp` descriptor to `os.fdopen`, which takes ownership. When the `with` block unwound on a write/fsync error the file object had already closed that descriptor, and the handler then called `os.close(fd)` on it again. That second close is harmless only if nothing else opened a file in between; under the threaded server the number can already belong to an unrelated stream, and the `except OSError: pass` swallowed the evidence either way. Now the raw descriptor is closed only on the one path where `fdopen` itself failed and ownership never transferred.

Regression test `test_compact_events_write_failure_does_not_close_reused_fd` asserts the raw fd is never closed after `fdopen` takes ownership — verified red on `16002a292`, green after the fix. `tests/test_eventlog.py`: 26 passed. Ruff check/format clean; mypy clean on `eventlog.py` (the 3 remaining errors are pre-existing in `gptme/acp/`).

**Comment by TimeToBuildBob** [comment]
@greptileai review

**Comment by TimeToBuildBob** [comment]
## 🤖 AI code review

Found one correctness issue in the append-only fast path: it uses Python object identity plus file size to decide that the on-disk content is unchanged, which fails for in-place mutations of message objects. The code should validate the serialized bytes of the persisted prefix before appending, or treat messages as immutable and always use `replace`.

### Confidence Score: N/A

Not scored — every finding was dropped by policy.

<details><summary><b>1 out-of-scope finding (outside this PR's changed files)</b></summary>

**❌ P1 high** — [`gptme/logmanager.py:108`](https://github.com/gptme/gptme/blob/f2ae87d9a47986752fcf414998e319091334f621/gptme/logmanager.py#L108)

The append optimization in `Log.write_jsonl`/`_can_append` trusts that the on-disk file is still exactly the state captured by `self.persisted` if two conditions hold: the file size hasn't changed and the first `len(self.persisted)` message objects are the same Python objects (`is`). Both conditions are too weak.

If a `Message` object is mutated in place (rather than replaced via `replace`), object identity remains unchanged but the serialized content on disk is now stale. `_can_append` will return `True` and new messages will be appended after the old, stale serialization, corrupting the file.

The check should verify at least the serialized prefix bytes on disk against the current persisted messages, rather than relying solely on Python object identity and file size.

```suggestion
Compare the actual file bytes of the persisted prefix against the current serialization of the persisted messages, or use an exclusive file lock and a reliable dirty flag. If in-place mutation is not supported, enforce immutability on `Message` and always replace via `replace`.
```

<sub>**How this was verified:** Run the following scenario:

log = read a file containing [A, B]
log.append(C)
# external/in-place mutation: mutate B's content without replacing the object
log.write_jsonl(file, append=True)

Observe that the file is left as &lt;stale A/B serialization&gt; + C instead of &lt;updated A/B&gt; + C.</sub>

</details>

| `772766fb8a95` | ? | 0 | llm | 2026-09-04 22:45 UTC |
| `c4d220ecc955` | 3/5 | 2 | llm | 2026-09-05 01:38 UTC |
| `af96314790fa` | 3/5 | 3 | llm | 2026-09-05 02:34 UTC |
| `b084e0e6e2ae` | 3/5 | 1 | llm | 2026-09-05 04:58 UTC |

</details>

<sub>Reviewed `f2ae87d9a479` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1563s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
**CI adjudication for run [33926434257](https://github.com/gptme/gptme/actions/runs/33926434257)** (`Test without API keys or extras`): neither failure is caused by this PR. This branch touches only `gptme/logmanager/`; both failures are elsewhere and both are real defects, now fixed in gptme/gptme#3717.

1. `tests/test_util_cli.py::test_models_list` — `JSONDecodeError`. `gptme-util models list --json` prefixes the JSON document with a WARNING-level provider log ("gptme provider requires authentication") that escapes `redirect_stdout` via a Rich handler. The existing guard is `logging.disable(logging.INFO)`, which does not cover WARNING. User-facing: `models list --json | jq` breaks for anyone without a `gptme` provider credential.

2. `tests/test_tools_gh.py::TestWaitForChecks::test_all_pass_immediately` — `Expected 'sleep' to not have been called. Called 544 times.` `@patch("gptme.tools.gh.time.sleep")` patches `time.sleep` process-wide, so background threads in the same xdist worker get counted. All 544 calls are `call(0.05)`; `_wait_for_checks`'s `poll_interval` is `10`, so none came from the code under test.

This is the *second* independent flake class this branch has surfaced — the earlier `Test without API keys` failure was the `FD_SETSIZE`/`select()` issue tracked in gptme/gptme#3715, fixed in gptme/gptme#3716.

**Blocking on this PR**: nothing here. Once gptme/gptme#3716 and gptme/gptme#3717 are merged, update this branch (`gh pr update-branch 3714`) and both `Test` jobs should go green.

TimeToBuildBob:
Addressed the review's destination-scoping concern in `4adc4fa3f`: each incremental-write cursor now records the exact resolved destination path, and a different branch/view destination forces a full rewrite instead of reusing another file's prefix. Added a regression proving a cursor from `conversation.jsonl` cannot omit inherited history when writing `branches/dev.jsonl`.

Verified locally: 142 targeted logmanager/eventlog/compaction/timing/auto-step tests pass; Ruff and pre-commit are clean. Fresh CI is running on the new head.

TimeToBuildBob:
@greptileai review

gptme/logmanager/manager.py
```diff
@@ -695,14 +744,17 @@ def load(
 
         log = Log.read_jsonl(logfile)
         msgs = log.messages or initial_msgs or []
-        return cls(msgs, logdir=logdir, branch=branch, lock=lock, **kwargs)
+        manager = cls(msgs, logdir=logdir, branch=branch, lock=lock, **kwargs)
+        if log.messages:
+            manager.log = manager.log.replace(persisted=log.persisted)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `c4d220ecc`: `LogManager.load()` now carries the resolved `persisted_path` into the active log alongside the persisted prefix. The existing load-cursor regression now asserts destination preservation for main, branch, and view logs.

Verified with the targeted pytest, Ruff check/format, mypy, and pre-commit.

TimeToBuildBob:
@greptileai review

gptme/logmanager/manager.py
```diff
@@ -100,14 +113,48 @@ def pop(self) -> "Log":
...
+            return not output.exists()
+        return all(old is new for old, new in zip(prefix, self.messages))
+
+    def write_jsonl(self, path: PathLike, *, append: bool = False) -> "Log":
+        output = Path(path)
+        append_safe = append and self._can_append(output)
+        mode = "a" if append_safe else "w"
+        start = len(self.persisted) if append_safe else 0
+        with open(output, mode, encoding="utf-8") as file:
+            file.writelines(
```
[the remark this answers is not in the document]
TimeToBuildBob: Measured now rather than relying only on the live I/O attribution. A deterministic `cProfile` workload starts with 400 × 2 KiB messages, then persists 200 one-message appends against the merge-base and this head:

- merge-base `a3902b8a0`: `Log.write_jsonl` 201 calls / 0.670 s cumulative; `writelines` 0.659 s; 214,548,485 kernel-accounted bytes written for a 1,280,980-byte final file
- head `c4d220ecc`: `Log.write_jsonl` 201 calls / 0.026 s cumulative; `writelines` 0.004 s; 1,280,980 bytes written for the same final file

That is 25.8× less persistence time and 167.5× less write traffic in the isolated path. It corroborates the prior live `/proc/<pid>/io` sample (91–894 MB written for 1.8–4.7 MB run trees), rather than assuming the bottleneck from code shape alone. No code change was needed for this finding.

TimeToBuildBob:
All stated review findings are addressed on `c4d220ecc`: Greptile confirmed the profiling evidence satisfies its remaining requirement, every review thread is resolved, and CI is green. Greptile's holistic score remains 4/5, so I'm stopping after this review cycle rather than re-triggering again; this PR now needs human review/merge.

gptme/logmanager/manager.py
```diff
@@ -100,14 +113,48 @@ def pop(self) -> "Log":
...
+            # Nothing known-persisted: appending would duplicate existing lines.
+            return not output.exists()
+        return all(old is new for old, new in zip(prefix, self.messages))
+
+    def write_jsonl(self, path: PathLike, *, append: bool = False) -> "Log":
+        output = Path(path)
+        append_safe = append and self._can_append(output)
+        mode = "a" if append_safe else "w"
+        start = len(self.persisted) if append_safe else 0
+        with open(output, mode, encoding="utf-8") as file:
```
TimeToBuildBob, line 164: ❌ **P1** — write_jsonl() now opens an existing conversation in append mode when _can_append() passes. If the write fails partway (for example OSError ENOSPC after some lines were written), the partial bytes remain in the file, but the exception prevents the returned Log from being produced, so self.persisted is untouched. On the next call, _can_append() still sees the original message objects unchanged and returns True, so the file is opened in 'a' again and the same new messages are appended a second time, possibly onto a partial line. The result is duplicate and malformed JSONL lines that _gen_read_jsonl either skips or reads as extra messages. Previously write_jsonl always opened 'w', so a retry truncated and rewrote the file cleanly; the new append path removes that recovery property. A concrete trigger is a disk-full error during the write of a new turn.

TimeToBuildBob: Fixed in `af9631479`. Each append cursor now records the exact destination byte size after a successful write; `_can_append()` requires that size to remain unchanged. A partial failed append therefore invalidates the cursor, and the retry opens with `w` to restore the complete JSONL rather than duplicating partial bytes. The regression injects an `OSError` after half a line is flushed, retries, and verifies exactly the original two messages remain.

The same cursor check also repairs externally deleted/truncated/replaced destinations by falling back to a full rewrite.

TimeToBuildBob: Fixed and verified in `af9631479`; see the preceding thread reply for the failure-injection regression and destination-size cursor details.

gptme/logmanager/eventlog.py
```diff
@@ -8,16 +8,34 @@
...
+    fcntl: Any = importlib.import_module("fcntl")
+except ImportError:  # pragma: no cover - Windows
+    fcntl = None
+
+try:
+    msvcrt: Any = importlib.import_module("msvcrt")
+except ImportError:  # pragma: no cover - POSIX
+    msvcrt = None
+
+_event_log_thread_lock = threading.Lock()
```
TimeToBuildBob: ⚠️ **P2** — _event_log_thread_lock is a single module-level threading.Lock guarded by every _event_log_lock acquisition regardless of which logdir is in play. append_next_event, append_event, and compact_events all take it, so an append or a compaction in one conversation blocks event appends in every other conversation in the same process for the whole critical section. The compaction path is the worst offender: it reads the full event log, rewrites all retained events to a temp file, and calls os.fsync inside the lock, so a large checkpoint compaction in one chat stalls unrelated chats' appends. This directly counteracts the PR's write-amplification goal in exactly the multi-chat server scenario the same-process lock-holder path in _acquire_lock enables. A per-logdir lock dict keyed by the resolved event-log path would serialize only operations on the same log.

gptme/logmanager/manager.py
```diff
@@ -100,14 +114,58 @@ def pop(self) -> "Log":
...
+            return not output.exists()
+        try:
+            if (
+                self.persisted_size is None
+                or output.stat().st_size != self.persisted_size
+            ):
+                return False
+        except OSError:
+            return False
+        return all(old is new for old, new in zip(prefix, self.messages))
```
TimeToBuildBob, line 157: ⚠️ **P2** — LogManager.__init__ passes every loaded message through snapshot_message_files → _store_message_files, which for any message with a non-empty files list unconditionally returns replace(msg, file_hashes=file_hashes) — a NEW Message object — even when preserve_existing=True and nothing changed. LogManager.load then restores the persistence cursor from log.persided, which holds the ORIGINAL objects produced by read_jsonl. _can_append compares the persisted prefix to the live messages by identity (all(old is new)), so for any conversation containing even a single attached-file message the identity check returns False on the first write after reload, forcing a mode "w" full rewrite of the whole conversation. The PR's central append-only optimization therefore never engages after a reload for file-bearing conversations, which is exactly the long-session trajectory-bloat case the PR targets. Correctness is unaffected; the fix is to seed persisted from the post-snapshot objects.

gptme/logmanager/eventlog.py
TimeToBuildBob: Fixed in `d19ccffa9`. Recovery logs now use a lock keyed by the resolved `events.jsonl` path, so append/compaction I/O in one conversation no longer takes the in-process lock for unrelated conversations. A small registry guard only protects lock lookup/creation; the per-log lock protects the actual I/O.

gptme/logmanager/manager.py
TimeToBuildBob: Fixed in `d19ccffa9`. `_store_message_files()` now returns the original immutable `Message` when preserving existing attachment hashes makes no data change, so the loaded persistence cursor's identity prefix remains valid. The regression loads a file-bearing conversation, verifies the cursor refers to the post-snapshot message object, appends a message, and confirms the persisted conversation is complete.

TimeToBuildBob:
## Convergence adjudication

- **Fixed:** `d19ccffa9` removes the process-wide recovery-log I/O lock, preserving serialization per resolved `events.jsonl` path while allowing unrelated conversations to append/compact independently. It also preserves loaded message identity when attachment snapshots are unchanged, so file-bearing conversations retain incremental-write cursors. Finally, checkpoint sequence allocation, append, and compaction now share one critical section, closing the reviewed duplicate-sequence/data-loss race.
- **Remaining:** none. Both current-head reviewer findings were correctness/performance defects and were fixed rather than dismissed; every Bob-owned inline finding thread is resolved. No human review thread was resolved.
- **Verified against:** `d19ccffa979c344b0064861d263586fdc2744062` (classification source review: `af96314790fafa2aef60873ffb238c1dbda24707`). Local targeted suite: 75 passed; Ruff check/format, scoped pre-commit including mypy, and `git diff --check` pass.
- **CI:** fresh checks for `d19ccffa9` are running; completed checks currently pass, with no failure reported.
- **Domain risk:** concurrency behavior across processes and Windows byte-range locking remains the fragile surface; CI covers the Windows fallback structurally, but a maintainer may want a real Windows/server smoke test before merge.
- **Convergence:** detector reports `stable_rounds: 3` (`required_stable_rounds: 2`, `status: no_current_blocking`). The attempt cap is reached, so I did **not** trigger another Greptile review. Stop at maintainer judgment after CI/current-head automated review settle.

TimeToBuildBob:
Fixed the current `Test without API keys or extras` failure in `627792d6`.

The implementation emitted one warning as intended; the assertion counted the token twice because pytest's formatted log text includes it in both the message and source snippet. The test now counts matching logger records/messages instead of substrings in rendered `caplog.text`.

Verified the targeted test locally, plus Ruff check/format and pre-commit. Fresh CI is running on the pushed head.

gptme/logmanager/manager.py
```diff
@@ -89,6 +98,11 @@ def __iter__(self) -> Generator[Message, None, None]:
     def __repr__(self) -> str:
         return f"Log(messages=<{len(self.messages)} msgs>)"
 
+    @property
+    def persisted_messages(self) -> int:
+        """Number of messages known to be on disk already."""
+        return len(self.persisted)
+
     def replace(self, **kwargs) -> "Log":
         return replace(self, **kwargs)
```
TimeToBuildBob, line 108: ❌ **P1** — `Log.read_jsonl` records `persisted_size=output.stat().st_size`, which is the whole file size even when the final bytes are a partial/invalid line without a trailing newline. `_can_append` later trusts only that size, so a subsequent `write_jsonl(..., append=True)` treats the file as a valid prefix and writes new JSON immediately after those trailing bytes. If the file was truncated by a failed append or simply lacks a trailing newline, the new message becomes glued to the partial tail (`...partial{"role":...}`), fails JSON parsing on the next load, and is silently lost. The retry test only covers the same Log instance where `persisted_size` still reflects the pre-failure size; after a reload the bad tail is included in `persisted_size`, so the corruption is permanent. `read_jsonl` should set `persisted_size` to the offset after the last successfully parsed line, or `_can_append` should verify the file ends with a newline before allowing append mode.

TimeToBuildBob: Fixed the reload-after-partial-tail corruption in `f2ae87d9`. `_can_append()` now requires the persisted file to end at a newline boundary before selecting append mode; otherwise it rewrites the complete valid in-memory log. The regression reloads a JSONL file with a malformed partial tail, appends a new message, and verifies the file is repaired to exactly the two valid messages.

The current unrelated CI failures are the known server auto-stepping flakes (`test_auto_stepping` and `test_multi_tool_per_message`); the no-extras matrix passed. The push has started a fresh CI run.

TimeToBuildBob:
CI failure was `filedescriptor out of range in select()` — the shell tool's pipe fds exceeded FD\_SETSIZE (1024) under 16 xdist workers in CI.

Fix: cherry-picked the `poll()` fix from #3716 (which tracks this exact class of issue) onto this branch. This replaces all `select.select()` calls in the shell I/O loop with a `_wait_readable()` helper backed by `select.poll()`, which has no fd ceiling.

New CI run: https://github.com/gptme/gptme/actions/runs/33948376630

TimeToBuildBob:
CI re-triggered: the single failure was `test_e2e_tool_confirm_execute` (TUI E2E marker-file not created), which is not in the changed files (`logmanager/eventlog.py`, `logmanager/manager.py`). The test has had multiple timing-fix commits recently and was already retried by pytest-retry before failing. Re-running to confirm it's pre-existing flakiness.

ErikBjare:
@TimeToBuildBob should we merge or is this not quite good enough?

TimeToBuildBob:
Merge it. It's ready.

All required CI is green, the branch merges clean, and the reviewer converged (3 stable rounds, every Bob-owned finding thread resolved). The change is concurrency-sensitive, but the fragile surface — cross-process serialization and the Windows byte-range-lock fallback — is structurally covered by CI, and each incremental-write cursor now records its exact resolved destination so a branch/view destination mismatch forces a full rewrite rather than reusing another file's prefix.

One residual risk a maintainer may want to eyeball: no real Windows/server smoke test has run, only the structural fallback. If that's acceptable, squash-merge it as-is.

PR #3716, opened 2026-09-04
fix(shell): use poll() instead of select() to survive fds above FD_SETSIZE

### LINKED ISSUE #3715: test: test_tools_shell fails with 'filedescriptor out of range in select()' under xdist

```
```

```
test_needs_tty_sudo_detection
test_shell_cwd_parameter
```

- `shell.py:1005` (drain)

TimeToBuildBob:
Fixes the intermittent `test_tools_shell.py` / `test_shell_output_mixing_issue408.py` failures that are currently red on `master` and on every open PR.

Closes #3715

## Root cause

`select.select()` is backed by `fd_set`, which cannot represent a descriptor `>= FD_SETSIZE` (1024). Given such an fd it **raises** rather than degrading:

```
```

Under `pytest -n 16`, accumulated descriptors push the shell tool's pipes past 1024. Because *which* test happens to be running when the fds cross the line varies run to run, the failures move around — which is why they present as flaky rather than as a deterministic bug.

Recent evidence, all the same signature:
- `master` runs 33885146088, 33877865588
- PR run 33919002483 — 14 tests across `test_tools_shell.py`, `test_shell_output_mixing_issue408.py`

## Fix

Replace all three `select.select()` call sites with a `_wait_readable()` helper backed by `select.poll()`, which has no descriptor ceiling.

`POLLHUP`/`POLLERR` are reported by `poll()` regardless of the requested event mask, so EOF still wakes the poll — matching `select()`'s treatment of EOF as readable. Behavior is otherwise unchanged; only the timeout unit conversion (seconds → milliseconds) is new.

## Verification

`tests/test_shell_high_fd.py` `dup2`'s a pipe read end to fd 1100 and asserts it is reported readable. It is genuinely falsifying — against the old implementation:

```text
OLD select() -> ValueError filedescriptor out of range in select()
NEW _wait_readable() -> [1100]
```

Local runs on this branch:
- `test_shell_high_fd.py` — 3 passed
- `test_tools_shell.py`, `test_shell_output_mixing_issue408.py`, `test_shell_fd_leak.py` — 129 passed
- `test_shell_background.py`, `test_tools_shell_background.py`, `test_tools_shell_multiline.py`, `test_shell_issue729.py`, `test_shell_issue772.py`, `test_shell_for_loop_issue724.py` — 140 passed

ruff check/format clean; `mypy gptme/tools/shell.py` reports no new errors (the 3 it emits are pre-existing, in `gptme/acp/`).

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -357,6 +357,26 @@ def _get_memory_limit() -> int | None:
     return limit
 
 
+def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
+    """Return the subset of `fds` that are readable, waiting up to `timeout` seconds.
+
+    Uses `poll()` rather than `select()`. `select()` is backed by `fd_set`, which
+    cannot represent a descriptor >= FD_SETSIZE (1024) and raises
+    `ValueError: filedescriptor out of range in select()` instead of degrading.
+    Long-lived processes and highly parallel test runs (`pytest -n 16`) routinely
+    push descriptors past that line. `poll()` has no such ceiling.
+    """
+    poller = select.poll()
+    for fd in fds:
+        # POLLHUP/POLLERR are reported regardless of the requested mask, so EOF
+        # still wakes the poll — matching select(), which reports EOF as readable.
+        poller.register(fd, select.POLLIN)
+    # select() takes seconds (None == block forever); poll() takes milliseconds
+    # (negative == block forever).
+    timeout_ms = -1 if timeout is None else int(timeout * 1000)
+    return [fd for fd, _event in poller.poll(timeout_ms)]
+
+
 class ShellSession:
     process: subprocess.Popen
     stdout_fd: int
@@ -634,7 +654,7 @@ class ShellSession:
         else:
             assert select is not None
             while True:
-                pre_drain_rlist, _, _ = select.select([self.stderr_fd], [], [], 0.05)
+                pre_drain_rlist = _wait_readable([self.stderr_fd], 0.05)
                 if not pre_drain_rlist:
                     break
                 pre_drain_data = os.read(self.stderr_fd, 2**16).decode(
@@ -912,9 +932,7 @@ class ShellSession:
                         1.0, timeout - elapsed
                     )  # Check at least every second
 
-                rlist, _, _ = select.select(
-                    [self.stdout_fd, self.stderr_fd], [], [], select_timeout
-                )
+                rlist = _wait_readable([self.stdout_fd, self.stderr_fd], select_timeout)
 
                 # Handle timeout in select
                 if not rlist and timeout and start_time:
@@ -1002,9 +1020,7 @@ class ShellSession:
                             # has time to arrive from bash.
                             drain_empty_count = 0
                             while drain_empty_count < 2:
-                                drain_rlist, _, _ = select.select(
-                                    [self.stderr_fd], [], [], 0.1
-                                )
+                                drain_rlist = _wait_readable([self.stderr_fd], 0.1)
                                 if not drain_rlist:
                                     drain_empty_count += 1
                                     continue
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -371,9 +371,12 @@ def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
         # POLLHUP/POLLERR are reported regardless of the requested mask, so EOF
         # still wakes the poll — matching select(), which reports EOF as readable.
         poller.register(fd, select.POLLIN)
-    # select() takes seconds (None == block forever); poll() takes milliseconds
-    # (negative == block forever).
-    timeout_ms = -1 if timeout is None else int(timeout * 1000)
+    # select() takes seconds (None == block forever); poll() takes integer
+    # milliseconds (negative == block forever). Round positive sub-millisecond
+    # waits up so they do not become busy-spinning non-blocking polls.
+    timeout_ms = (
+        -1 if timeout is None else max(0 if timeout == 0 else 1, int(timeout * 1000))
+    )
     return [fd for fd, _event in poller.poll(timeout_ms)]
 
 
--- a/tests/test_shell_high_fd.py
+++ b/tests/test_shell_high_fd.py
@@ -11,6 +11,7 @@ POSIX-only: `resource`, `fcntl`, and `select.poll()` do not exist on Windows.
 """
 
 import os
+from unittest.mock import Mock, patch
 
 import pytest
 
@@ -95,3 +96,21 @@ def test_wait_readable_times_out_with_no_data():
     finally:
         os.close(read_fd)
         os.close(write_fd)
+
+
+def test_wait_readable_rounds_positive_submillisecond_timeout_up():
+    """A positive timeout must not turn into a non-blocking poll."""
+    poller = Mock()
+    poller.poll.return_value = []
+    with patch("gptme.tools.shell.select.poll", return_value=poller):
+        assert _wait_readable([7], 0.0001) == []
+    poller.poll.assert_called_once_with(1)
+
+
+def test_wait_readable_preserves_zero_timeout():
+    """An explicit zero timeout remains a non-blocking poll."""
+    poller = Mock()
+    poller.poll.return_value = []
+    with patch("gptme.tools.shell.select.poll", return_value=poller):
+        assert _wait_readable([7], 0) == []
+    poller.poll.assert_called_once_with(0)
--- a/tests/test_shell_high_fd.py
+++ b/tests/test_shell_high_fd.py
@@ -6,25 +6,45 @@ select()` rather than degrading. Long-lived processes and parallel test runs
 (`pytest -n 16`) push descriptors past that line, which surfaced as spurious
 `test_tools_shell` / `test_shell_output_mixing_issue408` failures across every
 PR in the repo. See gptme/gptme#3715.
+
+POSIX-only: `resource`, `fcntl`, and `select.poll()` do not exist on Windows.
 """
 
 import os
-import resource
 
 import pytest
 
 from gptme.tools.shell import _wait_readable
 
+fcntl = pytest.importorskip("fcntl", reason="POSIX-only test")
+resource = pytest.importorskip("resource", reason="POSIX-only test")
+
 # Anything at or above FD_SETSIZE is unrepresentable in an fd_set.
 FD_SETSIZE = 1024
-HIGH_FD = 1100
+# Headroom for the descriptors the scan may need to walk past.
+_FD_SCAN_WINDOW = 256
+
+
+def _find_free_fd(start: int) -> int:
+    """Return an unused descriptor >= `start`.
+
+    Never hardcode a target for `dup2`: it silently closes whatever already
+    occupies that number, and in the high-descriptor parallel environment this
+    test targets, that could be a live pytest/xdist descriptor.
+    """
+    for candidate in range(start, start + _FD_SCAN_WINDOW):
+        try:
+            fcntl.fcntl(candidate, fcntl.F_GETFD)
+        except OSError:
+            return candidate  # EBADF — nothing is using it
+    pytest.skip(f"no free descriptor in [{start}, {start + _FD_SCAN_WINDOW})")
 
 
 @pytest.fixture
 def high_fd_limit():
-    """Raise the soft RLIMIT_NOFILE far enough to allocate HIGH_FD, then restore."""
+    """Raise the soft RLIMIT_NOFILE far enough to allocate a high fd, then restore."""
     soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
-    needed = HIGH_FD + 16
+    needed = FD_SETSIZE + _FD_SCAN_WINDOW + 16
     if soft < needed:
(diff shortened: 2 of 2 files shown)


tests/test_shell_high_fd.py
```diff
@@ -0,0 +1,77 @@
...
+select()` rather than degrading. Long-lived processes and parallel test runs
+(`pytest -n 16`) push descriptors past that line, which surfaced as spurious
+`test_tools_shell` / `test_shell_output_mixing_issue408` failures across every
+PR in the repo. See gptme/gptme#3715.
+"""
+
+import os
+import resource
+
+import pytest
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 77d5a6692 — guarded `resource` and `fcntl` behind `pytest.importorskip`. `select.poll()` is POSIX-only anyway, so the whole module correctly skips on Windows rather than failing collection.

tests/test_shell_high_fd.py
```diff
@@ -0,0 +1,77 @@
...
+    yield
+    resource.setrlimit(resource.RLIMIT_NOFILE, (soft, hard))
+
+
+def test_wait_readable_handles_fd_above_fd_setsize(high_fd_limit):
+    """A readable descriptor >= FD_SETSIZE must be reported, not raise."""
+    read_fd, write_fd = os.pipe()
+    high_read_fd = None
+    try:
+        os.dup2(read_fd, HIGH_FD)
```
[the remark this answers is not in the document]
TimeToBuildBob: Good catch, fixed in 77d5a6692. This was the sharpest finding: blindly `dup2`-ing onto 1100 would cause exactly the class of unrelated flakiness this test exists to eliminate. Now scans for an unused descriptor at or above `FD_SETSIZE` via `fcntl.F_GETFD` (EBADF == free) and skips if none is available in the window.

TimeToBuildBob:
## 🤖 AI code review

Adds a poll-based `_wait_readable` helper to `shell.py` and switches the three `select.select` wait sites (the stderr pre-drain in `_run_pipe`, `_read_output_unix`'s main stdout/stderr wait, and its stderr drain) to it, converting the float-second select timeout into poll's integer milliseconds with `None` → -1 (infinite), 0 → non-blocking, and other values floored at 1 ms. Adds `tests/test_shell_high_fd.py` covering fd &gt;= 1024 readiness, idle fds not being returned, idle-pipe timeout, the sub-millisecond rounding floor, and the 0 case. The conversion preserves the normal data-and-EOF paths (poll reports POLLHUP even with only POLLIN requested, which mirrors select's "readable on EOF"), but two issues need attention: negative timeouts no longer mean "wait forever" (busy-spin risk in `_read_output_unix`), and none of the tests cover the EOF path the loops rely on for termination.

**⚠️ P2 medium** — `gptme/tools/shell.py:360`

`_wait_readable` silently changes negative timeouts from "block forever" to a 1 ms poll. CPython's `select.select(r, w, x, t)` builds a `timeval` only for `t > 0`; negative `t` passes `NULL`, so select blocks indefinitely, and `poll()`'s documented contract is the same (negative → wait forever). The expression `max(0 if timeout == 0 else 1, int(timeout * 1000))` yields 1 for any negative input instead of -1. `_read_output_unix` computes `select_timeout` from the remaining per-command budget (`timeout - elapsed`); after a long 64 KiB blocking read or scheduling overshoot that value can go negative, so the loop would change from "park until stdout/stderr activity" to a ~1 kHz busy spin. Treat any `timeout < 0` (or `None`) as -1, keep 0 untouched, and round other positive values up to a whole millisecond.

```suggestion
Check for negative before applying the floor: `-1 if timeout is None or timeout < 0 else max(1, int(timeout * 1000))`, keeping the `timeout == 0` non-blocking case.
```

<sub>**How this was verified:** read</sub>

</details>

<details><summary><b>2 out-of-scope findings (outside this PR's changed files)</b></summary>

**⚠️ P2 medium** — [`packages/shell/tests/test_shell_posix.py:50`](https://github.com/gptme/gptme/blob/d101a3bdd267cf30f348133aaa8eeb345031becf/packages/shell/tests/test_shell_posix.py#L50)

In `test_wait_readable_handles_fd_above_fd_setsize`, the fixture computes `needed = FD_SETSIZE + _FD_SCAN_WINDOW + 16` (~1296) and then does `if hard != resource.RLIM_INFINITY and hard < needed: pytest.skip(...)`. On any system where the process hard `RLIMIT_NOFILE` is below 1296 (hard limits are commonly lowered in Docker/CI sandboxes; 1024 is still a common hard cap on tuned systems), this test silently reports as *skipped* — i.e., green — without ever exercising the high-fd path this PR is meant to cover. The PR's verification claims "3 passed" unconditionally, but the actual regression protection is environment-dependent: a restricted CI runner could merge this PR while the high-fd protection (the entire purpose of the test) never executes even once. This is distinct from the already-flagged coverage gaps (that the tests don't assert the POLLHUP path when they *do* run); this is a condition where the test never runs at all while still emitting success ("skipped" is treated as non-failure by most CI). Recommend: either raise the check to only skip when `hard < FD_SETSIZE + 2` (the minimum needed for fd 1024+ to be creatable) and otherwise assert the raised value, or explicitly fail/assert in the test if the environment cannot raise the limit, so the missing protection becomes visible.

```suggestion
Make the skip condition narrow (only skip when even the soft limit cannot reach fd 1024), and otherwise assert that the high-fd path actually ran, so a restricted CI can't silently drop the regression coverage.
```

<sub>**How this was verified:** The fixture at lines 47-52 computes needed = 1024+256+16 = 1296; the guard at line 50 checks the *hard* limit; if a CI sandbox has hard = 1024 (common tuned default), the suite reports 'skipped' for this test while CI exits 0 — exactly what a regression would silently pass through.</sub>

**⚠️ P2 medium** — [`packages/shell/src/shell.py:358`](https://github.com/gptme/gptme/blob/d101a3bdd267cf30f348133aaa8eeb345031becf/packages/shell/src/shell.py#L358)

In `_wait_readable`, the conversion `int(timeout * 1000)` truncates toward zero and the floor `max(1, ...)` widens every positive sub-millisecond value to exactly 1 ms. In `_read_output_unix`, `timeout` is passed as `min(1.0, timeout - elapsed)`, so just before a deadline `remaining` is commonly a tiny float like 0.0004 s — the helper turns the *select()* semantics (~0.4 ms wait) into a 1 ms wait, a ≥2.5× widening; conversely `remaining = 0.0509` becomes 50 ms instead of 50.9 ms. The two effects combine asymmetrically around deadline checks: the "check at least every second" cadence (a) can never wake faster than 1 kHz, and (b) can overshoot the intended deadline granularity by up to 1 ms per iteration while also undershooting slightly on sub-integer-ms values. This is distinct from the already-flagged negative-timeout case (the remaining-clock going ≤ 0); this is about *positive* sub-ms and fractional-ms values being quantized in both directions. Impact is small (bounded by 1 ms per iteration and &lt;1% on the 50/100 ms loops) but it silently changes the tight-deadline spin behavior that `select.timeval` previously handled exactly — and no test documents the intended quantization semantics for these near-deadline values.

```suggestion
Document and centralize the quantization rule (e.g., round-half-up to the nearest ms, min 1 ms), and add a unit test asserting the boundary behavior for remaining ∈ (0, 0.001) and (0.049, 0.051).
```

<sub>**How this was verified:** All in-repo call-sites pass either constants 0.05/0.1 (exact ms) or the float `min(1.0, remaining)`; only the latter produces fractional-ms values like 0.0004 which, by the formula, become max(1, 0)=1 ms — a 2.5×+ widening versus select()'s exact 0.4 ms wait.</sub>

</details>

- `gptme/tools/shell.py` — Adds the poll-based `_wait_readable` helper and rewires the three `select.select` wait sites (the stderr pre-drain in `_run_pipe`, the main stdout/stderr wait in `_read_output_unix`, and that function's stderr drain) to it, converting the float-second timeout into poll's integer milliseconds with `N
- `tests/test_shell_high_fd.py` — New module with five tests: data readiness on an fd &gt;= 1024 (the PR's falsifying case), idle fds not being returned, timeout on an idle pipe, the 1ms floor for tiny positive timeouts, and 0 staying non-blocking.

</details>

| `0a4f2a1a6840` | 5/5 | 0 | llm | 2026-09-04 21:42 UTC |
| `77d5a6692865` | 4/5 | 1 | llm | 2026-09-04 22:32 UTC |

</details>

<sub>Reviewed `d101a3bdd267` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1616s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `77d5a6692865`:

- CI gate: satisfied
- `gptme/tools/shell.py`

TimeToBuildBob:
Follow-up on the automated handoff: the two Greptile threads were already fixed and answered on `77d5a6692865`; I have now resolved both threads. The current head is clean: all CI checks pass, Greptile is 5/5, the PR is mergeable, and the high-fd regression slice passes locally (3 tests). No code change was needed in this recovery pass. Manual maintainer review/merge remains the only gate because `gptme/tools/shell.py` is policy-protected.

gptme/tools/shell.py
```diff
@@ -634,7 +654,7 @@ def _run_pipe(
         else:
             assert select is not None
```
TimeToBuildBob, line 658: ⚠️ **P2** — The new helper converts the timeout with `int(timeout * 1000)`, which truncates toward zero. `select_timeout` in the read loop is computed as `min(1.0, timeout - elapsed)` where `elapsed` is sampled before the blocking call and excludes the subsequent read/decode/append work; on a run where the remaining time shrinks into the sub-millisecond range (e.g. `remaining = 0.0007`), `int(0.0007 * 1000)` = `0`, so `poll(0)` returns immediately instead of blocking up to the full remaining slice the way `select.select(..., 0.0007)` would. The loop then treats the empty result as a timeout check, re-samples `elapsed`, and recomputes a still-tiny remainder, briefly busy-spinning with 0 ms polls for a few iterations until `remaining <= 0` triggers the break. Net effect is a slightly early timeout and a transient CPU spin in the final milliseconds of a long read — no data loss and no hang, but it's a genuine (if low-severity) behavioral divergence from the original select() semantics when the remaining timeout slips below 1 ms.

TimeToBuildBob: Addressed the new P2 in `d101a3bdd`: positive sub-millisecond waits now round up to 1 ms instead of becoming busy-spinning non-blocking polls, while an explicit zero remains non-blocking. Added focused regression tests for both cases; the full high-fd slice passes locally (5 tests). CI is running on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Recovery pass complete on `d101a3bdd`: resolved the addressed timeout-rounding review thread and verified the new head is clean — all CI checks pass, Greptile is 5/5, all review threads are resolved, and GitHub reports the PR mergeable. No further code change is needed. Manual maintainer review/merge remains the only gate because `gptme/tools/shell.py` is policy-protected.

gptme/tools/shell.py
```diff
@@ -357,6 +357,29 @@ def _get_memory_limit() -> int | None:
     return limit
 
 
+def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
```
TimeToBuildBob, line 360: ⚠️ **P2** — `_wait_readable` silently changes negative timeouts from "block forever" to a 1 ms poll. CPython's `select.select(r, w, x, t)` builds a `timeval` only for `t > 0`; negative `t` passes `NULL`, so select blocks indefinitely, and `poll()`'s documented contract is the same (negative → wait forever). The expression `max(0 if timeout == 0 else 1, int(timeout * 1000))` yields 1 for any negative input instead of -1. `_read_output_unix` computes `select_timeout` from the remaining per-command budget (`timeout - elapsed`); after a long 64 KiB blocking read or scheduling overshoot that value can go negative, so the loop would change from "park until stdout/stderr activity" to a ~1 kHz busy spin. Treat any `timeout < 0` (or `None`) as -1, keep 0 untouched, and round other positive values up to a whole millisecond.

TimeToBuildBob: Rejected as a false positive. CPython does not treat a negative `select.select()` timeout as “block forever”; it raises `ValueError: timeout must be non-negative`. More importantly, this PR's only variable-timeout caller checks `elapsed >= timeout` and returns before computing `timeout - elapsed` (`gptme/tools/shell.py:913-935`), so it cannot pass a negative timeout to `_wait_readable`. `None` is the supported block-forever value and is already translated to `poll(-1)`; changing arbitrary negative values to block forever would broaden the helper contract while diverging from the replaced `select.select()` behavior.

Ricky-7-Yan (approved):
Reviewed the final d101a3b head and independently reconstructed both exact blobs from the PR patch (SHA-1s match 246e8752da and a8e78491c1). In native WSL I exercised the exact _wait_readable implementation with fd 1024, data readiness, EOF/POLLHUP after closing the writer, and timeout conversion for None, zero, and a positive sub-millisecond value. The POSIX-only test module skips cleanly during Windows collection. Ruff lint/format and diff checks are clean, and the hosted Linux/Windows matrix is green. I found no blocking issues.

PR #3720, opened 2026-09-05
fix(onboard): detect OAuth subscription providers

TimeToBuildBob:
## Problem

`gptme-onboard --check` reports `openai-subscription` and `grok-subscription` as not configured even when their OAuth token files exist. On an OAuth-only install it exits 1 with “No API keys detected”, despite the runtime and `gptme-doctor` recognizing the credentials.

## Fix

- Reuse `list_available_providers()` for OAuth detection, matching runtime/doctor behavior.
- Treat detected OAuth credentials as configured without trying API-key validation.
- Give the provider-specific `gptme auth <provider>` hint when credentials are absent.

## Verification

- `poetry run pytest tests/test_onboard.py -q` — 24 passed
- `ruff check gptme/cli/onboard.py tests/test_onboard.py`
- `ruff format --check gptme/cli/onboard.py tests/test_onboard.py`
- Reproduced `gptme-onboard --check` before the patch with an OAuth-only temp HOME: exit 1 / not configured. After the patch: exit 0 / `openai-subscription` configured.

**2026-09-07 trim**: dropped a smuggled `.github/workflows/test.yml` hunk (`playwright install chromium` → `chromium-headless-shell`) — a sensitive CI-path change with no relation to onboarding. Reverted cleanly; the diff is now just `gptme/cli/onboard.py` and `tests/test_onboard.py`.

### DIFF
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -68,8 +68,10 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
         for provider, source in list_available_providers():
             if provider in OAUTH_PROVIDERS:
                 results[provider] = (True, source)
-    except Exception:
-        pass
+    except Exception as e:
+        logger.warning(
+            "OAuth credential check failed: %s — run `gptme auth` to re-authenticate", e
+        )
 
     # Also check config file for API keys and configured model
     try:
@@ -359,10 +361,10 @@ def _run_wizard(check_only: bool = False) -> int:
             model = Prompt.ask("Default model", default=default_model)
         else:
             model = Prompt.ask("Default model (e.g. provider/model-name)")
-        if "/" in model:
+        if "/" in model and model.split("/", 1)[0] == selected:
             break
         console.print(
-            "[red]Model must be in provider/model format (e.g. openai/gpt-4o).[/red]"
+            f"[red]Model must be in {selected}/model format (e.g. {selected}/MODEL-NAME).[/red]"
         )
 
     # Create config
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -362,7 +362,12 @@ def _run_wizard(check_only: bool = False) -> int:
         else:
             model = Prompt.ask("Default model (e.g. provider/model-name)")
         provider, separator, model_name = model.partition("/")
-        if separator and provider == selected and model_name:
+        if (
+            separator
+            and provider == selected
+            and model_name
+            and not model_name.endswith("/")
+        ):
             break
         console.print(
             f"[red]Model must start with {selected}/ followed by a model name (e.g. {selected}/MODEL-NAME).[/red]"
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -362,10 +362,10 @@ def _run_wizard(check_only: bool = False) -> int:
         else:
             model = Prompt.ask("Default model (e.g. provider/model-name)")
         provider, separator, model_name = model.partition("/")
-        if separator and provider == selected and model_name and "/" not in model_name:
+        if separator and provider == selected and model_name:
             break
         console.print(
-            f"[red]Model must be in {selected}/model format (e.g. {selected}/MODEL-NAME).[/red]"
+            f"[red]Model must start with {selected}/ followed by a model name (e.g. {selected}/MODEL-NAME).[/red]"
         )
 
     # Create config
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -62,10 +62,14 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
             results[provider] = (False, None)
 
     # OAuth providers have no API key; use the same token-file detection as the
-    # runtime and ``gptme-doctor``.
-    for provider, source in list_available_providers():
-        if provider in OAUTH_PROVIDERS:
-            results[provider] = (True, source)
+    # runtime and ``gptme-doctor``. Credential/config errors must not discard
+    # providers already detected from the environment.
+    try:
+        for provider, source in list_available_providers():
+            if provider in OAUTH_PROVIDERS:
+                results[provider] = (True, source)
+    except Exception:
+        pass
 
     # Also check config file for API keys and configured model
     try:
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -19,8 +19,9 @@ from rich.prompt import Confirm, Prompt
 from rich.table import Table
 
 from ..config import config_path
+from ..llm import list_available_providers
 from ..llm.models import PROVIDERS, BuiltinProvider, get_recommended_model
-from ..llm.validate import PROVIDER_DOCS, validate_api_key
+from ..llm.validate import OAUTH_PROVIDERS, PROVIDER_DOCS, validate_api_key
 
 logger = logging.getLogger(__name__)
 console = Console()
@@ -60,6 +61,12 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
         else:
             results[provider] = (False, None)
 
+    # OAuth providers have no API key; use the same token-file detection as the
+    # runtime and ``gptme-doctor``.
+    for provider, source in list_available_providers():
+        if provider in OAUTH_PROVIDERS:
+            results[provider] = (True, source)
+
     # Also check config file for API keys and configured model
     try:
         from ..config import get_config
@@ -92,6 +99,12 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
+    if provider in OAUTH_PROVIDERS:
+        available = {name for name, _ in list_available_providers()}
+        if provider in available:
+            return True, "OAuth credentials found (not validated during onboarding)"
+        return False, f"Not authenticated (run gptme auth {provider})"
+
     env_var = PROVIDER_ENV_VARS.get(provider)
     if not env_var:
         return False, f"Unknown provider: {provider}"
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -104,7 +104,10 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
     if provider in OAUTH_PROVIDERS:
-        available = {name for name, _ in list_available_providers()}
+        try:
+            available = {name for name, _ in list_available_providers()}
+        except Exception:
+            available = set()
         if provider in available:
             return True, "OAuth credentials found (not validated during onboarding)"
         return False, f"Not authenticated (run gptme auth {provider})"
@@ -315,7 +318,14 @@ def _run_wizard(check_only: bool = False) -> int:
             rec_model = get_recommended_model(cast(BuiltinProvider, selected))
             default_model = f"{selected}/{rec_model}"
         except ValueError:
-            # Provider doesn't have a recommended model configured
+            # OAuth/subscription providers may have no builtin recommended model.
+            # Prompt the user to supply the model in provider/model form.
+            if selected in OAUTH_PROVIDERS:
+                console.print(
+                    f"[dim]Note: {selected} requires specifying a model.  "
+                    f"Enter it as [bold]{selected}/MODEL-NAME[/bold] "
+                    f"(e.g. {selected}/grok-3).[/dim]"
+                )
             default_model = selected
     else:
         default_model = selected
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -20,7 +20,7 @@ from rich.table import Table
 
 from ..config import config_path
 from ..llm import list_available_providers
-from ..llm.models import PROVIDERS, BuiltinProvider, get_recommended_model
+from ..llm.models import MODELS, PROVIDERS, BuiltinProvider, get_recommended_model
 from ..llm.validate import OAUTH_PROVIDERS, PROVIDER_DOCS, validate_api_key
 
 logger = logging.getLogger(__name__)
@@ -156,6 +156,29 @@ def _show_provider_status(providers: dict[str, tuple[bool, str | None]]) -> None
         table.add_row(provider, status, preview_text, docs_url)
 
     console.print(table)
+
+
+def _get_default_model(provider: str) -> str:
+    """Return a valid default model string for the given provider.
+
+    For providers with a builtin recommended model, returns ``provider/model``.
+    For OAuth/subscription providers that have no builtin recommendation, falls
+    back to the first model listed in the static ``MODELS`` dict so the default
+    is always a fully-qualified ``provider/model`` string rather than the bare
+    provider name (which fails at startup).
+    """
+    if provider in PROVIDERS:
+        try:
+            rec = get_recommended_model(cast(BuiltinProvider, provider))
+            return f"{provider}/{rec}"
+        except ValueError:
+            pass
+    # Fallback: first model in the static MODELS dict, if any.
+    # MODELS keys are Literal types; cast provider str to avoid the mypy overload mismatch.
+    provider_models = list(MODELS.get(cast(BuiltinProvider, provider), {}).keys())
+    if provider_models:
+        return f"{provider}/{provider_models[0]}"
+    return provider
 
 
 def _select_provider(providers: dict[str, tuple[bool, str | None]]) -> str | None:
@@ -317,22 +340,14 @@ def _run_wizard(check_only: bool = False) -> int:
     console.print("\n[bold]Step 4: Create configuration[/bold]")
 
     # Ask for model preference
-    if selected in PROVIDERS:
-        try:
-            rec_model = get_recommended_model(cast(BuiltinProvider, selected))
-            default_model = f"{selected}/{rec_model}"
-        except ValueError:
-            # OAuth/subscription providers may have no builtin recommended model.
-            # Prompt the user to supply the model in provider/model form.
-            if selected in OAUTH_PROVIDERS:
-                console.print(
-                    f"[dim]Note: {selected} requires specifying a model.  "
-                    f"Enter it as [bold]{selected}/MODEL-NAME[/bold] "
-                    f"(e.g. {selected}/grok-3).[/dim]"
-                )
-            default_model = selected
-    else:
-        default_model = selected
+    default_model = _get_default_model(selected)
+    if selected in OAUTH_PROVIDERS:
+        # Show a hint for OAuth providers so the user knows the required format.
+        console.print(
+            f"[dim]Note: {selected} requires specifying a model.  "
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold] "
+            f"(e.g. {default_model}).[/dim]"
+        )
     model = Prompt.ask("Default model", default=default_model)
 
     # Create config
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -139,7 +139,11 @@ def _show_provider_status(providers: dict[str, tuple[bool, str | None]]) -> None
     table.add_column("Key Preview")
     table.add_column("Docs")
 
-    for provider in PROVIDERS:
+    # Show all builtin providers plus any additional (e.g. OAuth) providers
+    # detected in `providers` that aren't in the builtin list, so a
+    # configured OAuth provider isn't silently omitted from the table.
+    all_providers = list(PROVIDERS) + [p for p in providers if p not in PROVIDERS]
+    for provider in all_providers:
         has_key, preview = providers.get(provider, (False, None))
         if has_key:
             status = "✅ Configured"
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -166,6 +166,10 @@ def _get_default_model(provider: str) -> str:
     back to the first model listed in the static ``MODELS`` dict so the default
     is always a fully-qualified ``provider/model`` string rather than the bare
     provider name (which fails at startup).
+
+    Returns an empty string if no default model can be determined.
+    Callers must treat an empty return as "no default available" and require
+    the user to supply a full ``provider/model`` string explicitly.
     """
     if provider in PROVIDERS:
         try:
@@ -178,7 +182,9 @@ def _get_default_model(provider: str) -> str:
     provider_models = list(MODELS.get(cast(BuiltinProvider, provider), {}).keys())
     if provider_models:
         return f"{provider}/{provider_models[0]}"
-    return provider
+    # No default known — returning the bare provider name would produce an invalid
+    # config (runtime requires provider/model).  Let the caller ask the user.
+    return ""
 
 
 def _select_provider(providers: dict[str, tuple[bool, str | None]]) -> str | None:
@@ -343,12 +349,21 @@ def _run_wizard(check_only: bool = False) -> int:
     default_model = _get_default_model(selected)
     if selected in OAUTH_PROVIDERS:
         # Show a hint for OAuth providers so the user knows the required format.
+        example = f" (e.g. {default_model})" if default_model else ""
         console.print(
             f"[dim]Note: {selected} requires specifying a model.  "
-            f"Enter it as [bold]{selected}/MODEL-NAME[/bold] "
-            f"(e.g. {default_model}).[/dim]"
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold]{example}.[/dim]"
         )
-    model = Prompt.ask("Default model", default=default_model)
+    while True:
+        if default_model:
+            model = Prompt.ask("Default model", default=default_model)
+        else:
+            model = Prompt.ask("Default model (e.g. provider/model-name)")
+        if "/" in model:
+            break
+        console.print(
+            "[red]Model must be in provider/model format (e.g. openai/gpt-4o).[/red]"
+        )
 
     # Create config
     config_created = _create_config(selected, model)
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -361,7 +361,8 @@ def _run_wizard(check_only: bool = False) -> int:
             model = Prompt.ask("Default model", default=default_model)
         else:
             model = Prompt.ask("Default model (e.g. provider/model-name)")
-        if "/" in model and model.split("/", 1)[0] == selected:
+        provider, separator, model_name = model.partition("/")
+        if separator and provider == selected and model_name and "/" not in model_name:
             break
         console.print(
             f"[red]Model must be in {selected}/model format (e.g. {selected}/MODEL-NAME).[/red]"
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -108,8 +108,13 @@ def _test_provider(provider: str) -> tuple[bool, str]:
     if provider in OAUTH_PROVIDERS:
         try:
             available = {name for name, _ in list_available_providers()}
-        except Exception:
-            available = set()
+        except Exception as e:
+            logger.warning(
+                "OAuth credential check failed: %s — run `gptme auth %s` to re-authenticate",
+                e,
+                provider,
+            )
+            return False, f"Could not read OAuth credentials: {e}"
         if provider in available:
             return True, "OAuth credentials found (not validated during onboarding)"
         return False, f"Not authenticated (run gptme auth {provider})"
--- a/gptme/cli/onboard.py
+++ b/gptme/cli/onboard.py
@@ -68,7 +68,7 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
         for provider, source in list_available_providers():
             if provider in OAUTH_PROVIDERS:
                 results[provider] = (True, source)
-    except Exception as e:
+    except (OSError, ValueError) as e:
         logger.warning(
             "OAuth credential check failed: %s — run `gptme auth` to re-authenticate", e
         )
@@ -108,7 +108,7 @@ def _test_provider(provider: str) -> tuple[bool, str]:
     if provider in OAUTH_PROVIDERS:
         try:
             available = {name for name, _ in list_available_providers()}
-        except Exception as e:
+        except (OSError, ValueError) as e:
             logger.warning(
                 "OAuth credential check failed: %s — run `gptme auth %s` to re-authenticate",
                 e,
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,7 +104,7 @@ jobs:
         poetry install -E all
 
     - name: Install playwright
-      run: poetry run playwright install chromium
+      run: poetry run playwright install chromium-headless-shell
 
     - name: Run tests (without API keys, including slow)
       env:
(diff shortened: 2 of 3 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
All acceptance criteria verified:

- Fix confirmed end-to-end: `gptme-onboard --check` exits 0 with `openai-subscription` configured on an OAuth-only install (vs. exit 1 before the patch)
- 9 onboard tests pass; lint/typecheck clean
- CI: 17/17 checks green (one job still queuing), Greptile 5/5

Ready for maintainer review.

TimeToBuildBob:
Automated merge handoff for `61a9a9c6c345`:

- CI gate: satisfied
- Review threads: clear
- `gptme/cli/onboard.py`

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `gptme/cli/onboard.py:107` | P1 | fixed |
| `gptme/cli/onboard.py:110` | P1 | rejected — The reviewer's "How this was verified" step is: "Confirmed the iteration error by checking that list_available_providers |
| `gptme/cli/onboard.py:329` | P2 | superseded by latest review (not reproduced) |
| `gptme/cli/onboard.py:181` | P2 | fixed |
| `gptme/cli/onboard.py:111` | P2 | rejected |
| `gptme/cli/onboard.py:71` | P2 | superseded by latest review (not reproduced) |
| `gptme/cli/onboard.py:362` | P2 | rejected |
| `gptme/cli/onboard.py:184` | P2 | superseded by latest review (not reproduced) |
| `gptme/cli/onboard.py:109` | P2 | resolved (no disposition on record) |
| `gptme/cli/onboard.py:117` | P2 | rejected |

</details>




The PR fixes provider detection to recognize OAuth token files and adds tests for missing-key and error-handling scenarios. However, it introduces a bug in `_test_provider` where the available set includes all provider names regardless of credential presence, causing it to incorrectly report any provider as configured.

Confidence 3/5

**❌ P1 high** — `gptme/cli/onboard.py:110`

In `_test_provider`, the `available` set is built as `{name for name, _ in list_available_providers()}`. `list_available_providers()` returns a dictionary mapping provider names to (has_key, ...) tuples, so this comprehension includes every provider name regardless of whether it actually has a key or OAuth token. Consequently, `if provider_name in available` always succeeds for any known provider (e.g., "openai" always qualifies), and `_test_provider` will incorrectly return `True` even when no API key or OAuth token is present. This breaks `gptme-onboard --check`, which would report all providers as configured even on a fresh install. The existing test `test_missing_key` (line 173) expects `False` and "No API key found", so this would also cause that test to fail. The fix is to filter by `has_key` as done elsewhere (see line 199).

```suggestion
available = {name for name, (has_key, _) in list_available_providers().items() if has_key}
```

<sub>**How this was verified:** Confirmed the iteration error by checking that `list_available_providers` returns a dict (see usage in `check` with `.items()`). A unit test for `_test_provider` would fail on this line.</sub>

</details>

- `gptme/cli/onboard.py` — This PR updates provider detection to include OAuth token files and adds tests for that behavior.
- `tests/test_onboard.py` — No findings for this file.

</details>

| `61a9a9c6c345` | 3/5 | 3 | llm | 2026-09-05 07:07 UTC |
| `97339edaf74e` | 4/5 | 2 | llm | 2026-09-05 08:44 UTC |
| `c9f663fd51bc` | 4/5 | 2 | llm | 2026-09-05 10:01 UTC |
| `8a211a74b58a` | 4/5 | 3 | llm | 2026-09-05 12:01 UTC |
| `12b68b7c5ace` | 4/5 | 2 | llm | 2026-09-05 12:40 UTC |
| `7e025fc2ab57` | 3/5 | 2 | llm | 2026-09-05 13:40 UTC |
| `b9d3d4f550a8` | 4/5 | 2 | llm | 2026-09-05 15:32 UTC |
| `5db5ce54d694` | 4/5 | 1 | llm | 2026-09-07 12:39 UTC |
| `713068b85c9d` | 3/5 | 1 | llm | 2026-09-08 03:02 UTC |
| `aeb2812f45e0` | 4/5 | 1 | llm | 2026-09-08 03:19 UTC |

</details>

<sub>Reviewed `945f4a51a44f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 537s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/onboard.py
```diff
@@ -92,6 +103,12 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
+    if provider in OAUTH_PROVIDERS:
+        available = {name for name, _ in list_available_providers()}
```
TimeToBuildBob: ❌ **P1** — In `_test_provider`, the OAuth branch calls `list_available_providers()` without the try/except that the same PR applies to the call in `_detect_providers` (lines 67-72). The PR's own new test `test_detect_preserves_env_fallback_when_oauth_lookup_fails` mocks this function with `side_effect=ValueError`, confirming the runtime can raise on malformed credential files. During `_run_wizard`, `_test_provider(selected)` is invoked after a user picks an OAuth provider; if the lookup raises at that moment, the exception escapes `main`, which only catches `KeyboardInterrupt` (lines 359-363), so onboarding aborts with a raw traceback instead of the graceful `(False, message)` failure every other path in this function returns. Wrap the call in the same `try/except Exception: pass` used in `_detect_providers`, or catch it and return `(False, ...)`.

TimeToBuildBob: Fixed in 97339edaf — wrapped the `list_available_providers()` call in `_test_provider` with `try/except Exception`, defaulting to an empty set on failure. Added `test_oauth_lookup_failure_in_test_provider` to cover this path. (Also added a hint for OAuth providers with no builtin recommended model in Step 4 of the wizard.)

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed both remaining P2 findings from the AI review (P1 and the "no recommended model" P2 were already fixed in `97339edaf`):

- `_show_provider_status` iterated `PROVIDERS` only, so a detected OAuth provider (e.g. `openai-subscription`) never appeared in the status table even though it was already counted as configured — making `--check` self-contradictory ("1 provider(s) configured" while every row read "Not configured"). Now shows all builtin providers plus any additional providers present in the results dict.
- Added `TestShowProviderStatus::test_shows_configured_oauth_provider_not_in_builtins` covering this.

Verification: `pytest tests/test_onboard.py -q` — 12 passed; `ruff check`/`ruff format --check` clean; local mypy shows only pre-existing unrelated errors in other files (not touched by this diff).

gptme/cli/onboard.py
```diff
@@ -298,7 +318,14 @@ def _run_wizard(check_only: bool = False) -> int:
...
-            # Provider doesn't have a recommended model configured
+            # OAuth/subscription providers may have no builtin recommended model.
+            # Prompt the user to supply the model in provider/model form.
+            if selected in OAUTH_PROVIDERS:
+                console.print(
+                    f"[dim]Note: {selected} requires specifying a model.  "
+                    f"Enter it as [bold]{selected}/MODEL-NAME[/bold] "
+                    f"(e.g. {selected}/grok-3).[/dim]"
+                )
             default_model = selected
```
TimeToBuildBob: ⚠️ **P2** — When an OAuth provider lacks a recommended model, `_run_wizard` leaves `default_model = selected`. The subsequent model prompt uses that bare provider name as the default, so pressing Enter writes a config like `model = "openai-subscription"` with no slash separating provider and model. The runtime requires `provider/model`, making the config unusable and failing the first launch. The new note only advises; it does not change the default. Since OAuth providers are now newly selectable and often have no recommended model, this is a real broken‑config path. Suggest setting `default_model` to empty for OAuth or validating that the entered model contains a `/` before writing.

gptme/cli/onboard.py
```diff
@@ -298,7 +322,14 @@ def _run_wizard(check_only: bool = False) -> int:
...
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `c9f663f` — extracted `_get_default_model(provider)` helper that falls back to the first model in the static `MODELS` dict when `get_recommended_model()` raises a `ValueError`. For `grok-subscription` this yields `grok-subscription/grok-4.6` instead of the bare provider name.

The Step 4 hint message now derives its example from the same helper (`e.g. grok-subscription/grok-4.6`), so it always shows a valid, usable model string rather than the hardcoded `grok-3` which only exists in the `xai` provider.

Added `TestGetDefaultModel::test_grok_subscription_returns_valid_model` and `test_openai_subscription_returns_valid_model` to cover the fallback path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `c9f663fd51bc`:

- CI gate: satisfied
- `gptme/cli/onboard.py`

gptme/cli/onboard.py
```diff
@@ -134,6 +158,29 @@ def _show_provider_status(providers: dict[str, tuple[bool, str | None]]) -> None
...
+            rec = get_recommended_model(cast(BuiltinProvider, provider))
+            return f"{provider}/{rec}"
+        except ValueError:
+            pass
+    # Fallback: first model in the static MODELS dict, if any.
+    # MODELS keys are Literal types; cast provider str to avoid the mypy overload mismatch.
+    provider_models = list(MODELS.get(cast(BuiltinProvider, provider), {}).keys())
+    if provider_models:
+        return f"{provider}/{provider_models[0]}"
+    return provider
```
TimeToBuildBob: ⚠️ **P2** — In `_get_default_model`, the fallback only uses the first model in the `MODELS` dict. If an OAuth provider (such as an unknown subscription provider)我是 to be confirmed) has no entry in `MODELS`, the function returns the bare provider name. This contradicts the PR's stated goal of never returning a bare provider name, and the onboarding prompt will present an invalid default model. Concretely, if a user accepts the default, `_create_config` writes `model = "provider"` with no slash, and startup will fail.import. The fallback should either ensure a valid model exists or raise a clearer error.

TimeToBuildBob: Fixed in `8a211a7` — changed `_get_default_model` to return `""` instead of the bare provider name when no model is found in `MODELS`. Added validation loop in `_run_wizard` that rejects entries without a `/`, so accepting an empty-or-bare default is no longer possible. Added `test_unknown_provider_returns_empty_not_bare_name` to cover this path.

Both current OAuth providers (`grok-subscription`, `openai-subscription`) have entries in `MODELS` so the fallback was not triggering, but any future OAuth provider without a `MODELS` entry would have silently produced a broken config.

gptme/cli/onboard.py
TimeToBuildBob: Fixed in 97339ed — wrapped list_available_providers() call in _test_provider with try/except Exception, defaulting to empty set on failure.

gptme/cli/onboard.py
TimeToBuildBob: Fixed in 8a211a7 — _get_default_model now returns "" instead of bare provider name when no MODELS entry exists; _run_wizard validates that entered model contains a "/" before writing config.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI failure on this PR is unrelated to the OAuth detection changes — `test_tools_shell.py` hit `ValueError: filedescriptor out of range in select()`, the same fd-exhaustion flake tracked in gptme/gptme#3715 and fixed by gptme/gptme#3716. Retriggered the failed job; no code change needed here.

gptme/cli/onboard.py
```diff
@@ -92,6 +103,15 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
+    if provider in OAUTH_PROVIDERS:
+        try:
+            available = {name for name, _ in list_available_providers()}
+        except Exception:
+            available = set()
+        if provider in available:
```
TimeToBuildBob, line 113: ⚠️ **P2** — Treating mere presence in list_available_providers() as proof that credentials work makes the OAuth branch (and the --check gate built on _detect_providers) report as configured and exit 0 when only a stale/expired token file exists. The runtime that consumes this same detection may subsequently fail exactly when the token is expired, so the onboarding green light actively misleads automation and users. The 'not validated during onboarding' message is only shown in the interactive step 3 path; --check and the status table both classify the agent as fully configured without issuing any warning. The fix is to validate the discovered OAuth credentials where a validator exists (or to fail step 3 and the --check count when an expiry check is possible), and to emit an explicit warning such as 'OAuth credentials found (not validated during onboarding)' in the status table and --check output when validation cannot be performed so the green signal is not overstated.

gptme/cli/onboard.py
```diff
@@ -60,6 +61,16 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
...
             results[provider] = (False, None)
 
+    # OAuth providers have no API key; use the same token-file detection as the
+    # runtime and ``gptme-doctor``. Credential/config errors must not discard
+    # providers already detected from the environment.
+    try:
+        for provider, source in list_available_providers():
+            if provider in OAUTH_PROVIDERS:
+                results[provider] = (True, source)
+    except Exception:
```
TimeToBuildBob: ⚠️ **P2** — Swallowing every Exception from list_available_providers() (line 71 in _detect_providers and the empty-set fallback at lines 109-110 in _test_provider) hides genuine failures – corrupt/malformed token files, permission errors, or transient I/O errors – and converts them all into either a silent change in the configured list or a 'Not authenticated (run gptme auth...)' message. When the underlying failure is a corrupt credentials file, a user will see wrong guidance (re-run auth) instead of the real cause, and --check will report fewer providers than actually exist, producing an exit code that does not reflect the true index-usable state. The exception should be narrowed to the specific expected failure (e.g., FileNotFoundError / TOML parse error) with a debug-level log for other exceptions so that genuine I/O or parsing problems are surfaced rather than masked.

gptme/cli/onboard.py
```diff
@@ -293,16 +346,24 @@ def _run_wizard(check_only: bool = False) -> int:
...
+        console.print(
+            f"[dim]Note: {selected} requires specifying a model.  "
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold]{example}.[/dim]"
+        )
+    while True:
+        if default_model:
+            model = Prompt.ask("Default model", default=default_model)
+        else:
+            model = Prompt.ask("Default model (e.g. provider/model-name)")
+        if "/" in model:
```
TimeToBuildBob: ⚠️ **P2** — The new model input loop only checks that the entered model contains a slash (`/`), but it does not verify that the provider prefix matches the provider selected in step 2. A user who selects `openai-subscription` can accidentally type `anthropic/claude-3` and the wizard will accept it, writing a config whose model targets a different provider than the one just selected. This defeats the earlier provider selection and connectivity test, potentially causing gptme to use a provider with different credentials or no configured credentials at runtime. The wizard should validate that `model.split('/')[0] == selected` (or at least warn and confirm when the prefix mismatches).

gptme/cli/onboard.py
```diff
@@ -60,6 +61,12 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
         else:
             results[provider] = (False, None)
 
+    # OAuth providers have no API key; use the same token-file detection as the
+    # runtime and ``gptme-doctor``.
+    for provider, source in list_available_providers():
+        if provider in OAUTH_PROVIDERS:
+            results[provider] = (True, source)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 97339ed — wrapped the `list_available_providers()` call in `_test_provider` with `try/except Exception`, defaulting to an empty set on failure so onboarding degrades gracefully rather than aborting. Added `test_oauth_lookup_failure_in_test_provider` to cover this path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the 2 remaining P2 findings from the AI reviewer (`45e87230`):

- **Silent OAuth exception** (`_detect_providers` line 71): Changed `except Exception: pass` to `except Exception as e: logger.warning(...)` so a corrupted/unreadable token file emits a clear warning ("OAuth credential check failed — run `gptme auth` to re-authenticate") instead of silently falling into the "No API keys detected" branch.
- **Cross-provider model validation** (`_run_wizard` line 362): Changed `if "/" in model` to `if "/" in model and model.split("/", 1)[0] == selected` so entering `openai/gpt-4o` while `grok-subscription` is selected is rejected instead of being written to the config. Error message now shows the expected format (e.g. "Model must be in grok-subscription/model format").
- Added `test_oauth_lookup_failure_logs_warning` test to cover the warning path.

CI failure ("Dependency Advisory Scan") is a transient GitHub network error (failed to fetch uv version from raw.githubusercontent.com) unrelated to this PR's changes.

TimeToBuildBob:
@greptileai review

gptme/cli/onboard.py
```diff
@@ -293,16 +348,24 @@ def _run_wizard(check_only: bool = False) -> int:
...
+            f"[dim]Note: {selected} requires specifying a model.  "
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold]{example}.[/dim]"
+        )
+    while True:
+        if default_model:
+            model = Prompt.ask("Default model", default=default_model)
+        else:
+            model = Prompt.ask("Default model (e.g. provider/model-name)")
+        if "/" in model and model.split("/", 1)[0] == selected:
+            break
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `7e025fc2` — switched to `str.partition` with an explicit `model_name` truthy check so `openai/` is rejected alongside bare `openai`. Added `TestModelValidation` covering trailing-slash, valid strings, bare provider, wrong provider, and leading-slash cases.

gptme/cli/onboard.py
```diff
@@ -293,16 +348,25 @@ def _run_wizard(check_only: bool = False) -> int:
...
+            f"[dim]Note: {selected} requires specifying a model.  "
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold]{example}.[/dim]"
+        )
+    while True:
+        if default_model:
+            model = Prompt.ask("Default model", default=default_model)
+        else:
+            model = Prompt.ask("Default model (e.g. provider/model-name)")
+        provider, separator, model_name = model.partition("/")
+        if separator and provider == selected and model_name and "/" not in model_name:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `5fcd29a0` — removed the `"/" not in model_name` guard. The check now only requires `separator` (a `/` is present), `provider == selected`, and `model_name` is non-empty. Identifiers like `openrouter/deepseek/deepseek-v4-pro` and `requesty/openai/gpt-4o-mini` now pass; `openai/` (empty segment) is still rejected. Added `test_nested_model_accepted` covering multi-slash format.

TimeToBuildBob:
@greptileai review

gptme/cli/onboard.py
```diff
@@ -293,16 +348,25 @@ def _run_wizard(check_only: bool = False) -> int:
...
+            f"[dim]Note: {selected} requires specifying a model.  "
+            f"Enter it as [bold]{selected}/MODEL-NAME[/bold]{example}.[/dim]"
+        )
+    while True:
+        if default_model:
+            model = Prompt.ask("Default model", default=default_model)
+        else:
+            model = Prompt.ask("Default model (e.g. provider/model-name)")
+        provider, separator, model_name = model.partition("/")
+        if separator and provider == selected and model_name:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `b9d3d4f5` — added `and not model_name.endswith("/")` to the validation predicate. For `openrouter/deepseek/`, `model_name = "deepseek/"` is truthy but ends with `/`, so it is now rejected. Valid multi-segment identifiers like `openrouter/deepseek/deepseek-v4-pro` still pass. Added `test_nested_trailing_slash_rejected` to cover this case.

gptme/cli/onboard.py
TimeToBuildBob: False positive — this is the intended behavior of the PR. Treating OAuth provider presence as "configured" matches how the runtime and `gptme-doctor` both handle it; onboarding deliberately does not re-validate OAuth tokens (credentials can be valid at onboard time and expire later regardless). The "OAuth credentials found (not validated during onboarding)" message in `_test_provider` is the explicit acknowledgment of this trade-off. Adding OAuth token expiry validation is a separate concern, out of scope for this fix.

gptme/cli/onboard.py
TimeToBuildBob: False positive — the validation predicate already enforces the provider prefix: `provider, separator, model_name = model.partition("/")` followed by `provider == selected` in the `if` guard (line 365-370). A user who selects `openai-subscription` and types `anthropic/claude-3` will get the "Model must start with openai-subscription/ followed by a model name" error and be re-prompted.

gptme/cli/onboard.py
```diff
@@ -134,6 +160,35 @@ def _show_provider_status(providers: dict[str, tuple[bool, str | None]]) -> None
...
+    """
+    if provider in PROVIDERS:
+        try:
+            rec = get_recommended_model(cast(BuiltinProvider, provider))
+            return f"{provider}/{rec}"
+        except ValueError:
+            pass
+    # Fallback: first model in the static MODELS dict, if any.
+    # MODELS keys are Literal types; cast provider str to avoid the mypy overload mismatch.
+    provider_models = list(MODELS.get(cast(BuiltinProvider, provider), {}).keys())
```
TimeToBuildBob, line 184: ⚠️ **P2** — In `_get_default_model`, the fallback path runs `MODELS.get(cast(BuiltinProvider, provider), {})` for a provider that is not in `PROVIDERS`. The `MODELS` dictionary is keyed by the builtin `BuiltinProvider` literal union; looking up an arbitrary string like `"grok-subscription"` or `"unknown-future-provider"` returns `{}`, which makes the function return `""`. For `openai-subscription` and `grok-subscription` the function therefore returns an empty string, and the PR's own tests only check `result.startswith("grok-subscription/")` / `result.startswith("openai-subscription/")` without asserting the returned value is non-empty. The only thing preventing `""` is whether the real `MODELS` dict contains those keys — and the new code was written on the assumption that it does. If the actual `MODELS` dict does not include `"grok-subscription"` or `"openai-subscription"` keys (which the test cannot confirm since the test is not exercised against the real dict), then `default_model` becomes `""`, the wizard prompts `Default model (e.g. provider/model-name)` with no default, and an OAuth user is forced to type a model name manually. More importantly, the fallback returns the first static model even when that model name does not match the provider's actual subscription model, so a user could get a config with a model string that the provider rejects at runtime. The masking is compounded by the fact that the validation predicate only checks format (`provider` + `/` + non-empty `model_name` with no second slash), never whether the model name is valid for that provider, so the wizard happily writes a model that the runtime will fail on. The concrete defect is that `_get_default_model` returns `""` for the subscription providers unless `MODELS` happens to contain those literal keys, and the wizard's validation loop then rejects the default, forcing users to guess a model name with no provider-side validation.

TimeToBuildBob:
Greptile trigger cap reached (8 lifetime triggers). Summary of what this session did:

- Resolved the 3 open Greptile threads (Empty Model Name Accepted, Nested Models Are Rejected, Nested Trailing Slash Accepted) — all were addressed in earlier commits (`7e025fc2`, `5fcd29a0`, `b9d3d4f5` respectively).
- Disposed 2 own-reviewer false-positive threads: the OAuth "overstated" finding (intentional design, matching runtime/doctor behavior) and the provider-mismatch finding (the `provider == selected` guard already handles it).

The only remaining Greptile score issue (trailing slash `openrouter/deepseek/` accepted) is fixed in the HEAD commit `b9d3d4f5`, but Greptile last reviewed `5fcd29a0`. A new push would let Greptile auto-review the fix; otherwise this needs a human merge or Greptile reset.

All threads are resolved. CI has a few pending jobs (API tests, non-extras tests) — the fd-exhaustion flake tracked in #3715 may recur but isn't caused by these changes.

TimeToBuildBob:
CI re-run triggered. The failing test `test_e2e_tool_confirm_execute` is a tmux-based TUI e2e test that fails intermittently on timing; it has no causal link to the onboard.py changes in this PR. All other 17 checks passed.

Greptile's 4/5 finding (nested trailing-slash accepted) was fixed in `b9d3d4f5` — the validation predicate now includes `and not model_name.endswith("/")`. Greptile reviewed at `5fcd29a0` (before the fix) and the per-PR lifetime trigger cap is exhausted; per the monitoring rules the own AI review verdict applies.

TimeToBuildBob:
Trimmed per today's triage: dropped the smuggled `.github/workflows/test.yml` hunk (`playwright install chromium` → `chromium-headless-shell`) — a sensitive CI-path change unrelated to the OAuth onboarding fix. Reverted cleanly; the diff is now just `gptme/cli/onboard.py` and its tests. Ready for maintainer review.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
The current CI failure on this PR (`tests/test_prompts.py::test_get_prompt_full`, `assert 8162 < 8100`) is **not** from the onboard.py changes here — it's master-bleed. Master itself is red on the identical test since `feat(memory)` (#3638) merged this morning: the memory tool's instructions pushed the full prompt over the 8100 ceiling. Same failure confirms on master head `1e7e5547c`.

The fix is already in flight: #3728 (workspace-local memory, the sibling follow-on) bumps the ceiling 8100→8250. Once it merges, master and this PR both go green. No code change needed on #3720 for this.

gptme/cli/onboard.py
```diff
@@ -92,6 +105,15 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
+    if provider in OAUTH_PROVIDERS:
+        try:
```
TimeToBuildBob, line 109: ⚠️ **P2** — In _test_provider, the new OAuth branch wraps list_available_providers() in a bare `except Exception` that sets `available = set()` and then reports “Not authenticated”. This turns every failure mode — corrupt token file, disk I/O error, or even a programming error in provider enumeration — into the same message as genuinely missing credentials. Unlike the new try/except in _detect_providers, this branch does not log the underlying exception, so an operator debugging an OAuth onboarding failure gets no diagnostic and may repeatedly run `gptme auth` while the real problem is an unreadable or malformed credential file. Log the exception at warning level before returning False, or at least distinguish credential-not-found from other errors.

TimeToBuildBob: Fixed in `aeb2812f4`.

The env-credentials part of this finding does not apply: this `try/except` only runs when `provider in OAUTH_PROVIDERS`. API-key providers never enter the branch, so a failed OAuth lookup cannot report them as unavailable. `_detect_providers` already preserves the env fallback on the same failure (covered by `test_detect_preserves_env_fallback_when_oauth_lookup_fails`).

The real gap was the silent swallow: every exception became "Not authenticated", with no log. `_test_provider` now logs the exception at warning (same wording as `_detect_providers`) and returns `Could not read OAuth credentials: …` instead of the missing-auth hint.

24 onboard tests pass, including the updated lookup-failure case.

gptme/cli/onboard.py
```diff
@@ -92,6 +105,20 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
...
+            available = {name for name, _ in list_available_providers()}
+        except Exception as e:
+            logger.warning(
+                "OAuth credential check failed: %s — run `gptme auth %s` to re-authenticate",
+                e,
+                provider,
+            )
+            return False, f"Could not read OAuth credentials: {e}"
```
TimeToBuildBob, line 117: ⚠️ **P2** — In `_test_provider`, the `except Exception` block at line 117 catches any exception raised by `list_available_providers()` and returns a message specifically blaming OAuth credentials. If `list_available_providers` encounters an unrelated error (e.g., a parsing bug, a permission issue on a non-OAuth config file, or a programming error), the user will receive a misleading diagnosis that the OAuth credentials are unreadable, even though the actual failure has nothing to do with OAuth. This can obscure genuine bugs and complicate debugging. The exception should be narrowed to the specific OAuth-related exceptions (e.g., `FileNotFoundError` or `PermissionError`) and other exceptions should be re-raised or given a generic error message. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Narrowing `except Exception` to `FileNotFoundError`/`PermissionError` would reintroduce the original P1 this PR closed. `list_available_providers()` raises `ValueError` on malformed credential files — the PR's own `test_oauth_lookup_failure_in_test_provider` mocks that — and re-raising it aborts onboarding with a traceback instead of a diagnostic.

This branch only runs when `provider in OAUTH_PROVIDERS`, so a failed lookup cannot mislabel an API-key provider. Logging the exception and returning `Could not read OAuth credentials: …` is the diagnostic the previous finding asked for. Consensus pass 2 already disagreed.

Won't implement.

TimeToBuildBob:
Self-merge was blocked by the consensus gate at 2/5 on `aeb2812f4` (this head cannot be raised while that verdict is in-window).

The gate's `_test_provider` diagnosis was stale — that path already returns `Could not read OAuth credentials: …`. The real remaining gap was the bare `except Exception` around `list_available_providers()`.

Pushed `945f4a51a`: OAuth lookup now catches only `(OSError, ValueError)` so a programming error still aborts. 26 onboard tests pass.

Did not merge. Did not retrigger Greptile (lifetime cap exhausted). Next cycle can actuate after CI + AI review land on this head.

gptme/cli/onboard.py
```diff
@@ -92,6 +105,20 @@ def _detect_providers() -> dict[str, tuple[bool, str | None]]:
 
 def _test_provider(provider: str) -> tuple[bool, str]:
     """Test if a provider is working (checks env vars and config file)."""
+    if provider in OAUTH_PROVIDERS:
+        try:
+            available = {name for name, _ in list_available_providers()}
```
TimeToBuildBob, line 110: ❌ **P1** — In `_test_provider`, the `available` set is built as `{name for name, _ in list_available_providers()}`. `list_available_providers()` returns a dictionary mapping provider names to (has_key, ...) tuples, so this comprehension includes every provider name regardless of whether it actually has a key or OAuth token. Consequently, `if provider_name in available` always succeeds for any known provider (e.g., "openai" always qualifies), and `_test_provider` will incorrectly return `True` even when no API key or OAuth token is present. This breaks `gptme-onboard --check`, which would report all providers as configured even on a fresh install. The existing test `test_missing_key` (line 173) expects `False` and "No API key found", so this would also cause that test to fail. The fix is to filter by `has_key` as done elsewhere (see line 199).

TimeToBuildBob: False positive — `list_available_providers()` does not return a dict of all providers.

It is typed and implemented as `list[tuple[Provider, str]]` of *already-configured* providers only (`gptme/llm/__init__.py:1153-1207`). The comprehension `{name for name, _ in list_available_providers()}` unpacks those tuples; a provider with no key or OAuth token is simply absent from the list. `_detect_providers()` is the dict that maps every builtin to `(has_key, preview)` — that is a different function, and the `.items()` usage the review cited lives there, not on `list_available_providers()`.

`test_missing_key` also does not exercise this line: it calls `_test_provider("openai")`, which is not in `OAUTH_PROVIDERS` and takes the env-var path. The OAuth path is covered by `test_oauth_credentials_found` / `test_oauth_credentials_missing`, which mock `list_available_providers` as a list of tuples (`[("openai-subscription", "oauth")]` and `[]`). Those tests pass, and CI is green on this head.

PR #3721, opened 2026-09-05
fix: decode editable install file URLs

yhz5613813:
## Summary

Editable-install metadata stores source locations as `file://` URLs. Reading the URL by slicing off the scheme leaves percent escapes such as `%20` in the reported path and can prevent `pyproject.toml` discovery.

Convert local file URLs with `url2pathname` in both install-info and extras-detection paths.

## Testing

- Added regression tests for decoded install reporting and editable-source extras discovery when the checkout path contains an encoded space.
- `pytest --noconftest tests/test_info.py::TestParseExtrasFromMetadata::test_editable_fallback_decodes_file_url tests/test_info.py::TestGetInstallInfo::test_editable_install_decodes_file_url -q`
- Ruff check, formatting check, and `compileall` on both changed files.

### DIFF
--- a/gptme/info.py
+++ b/gptme/info.py
@@ -15,6 +15,7 @@ import shutil
 from collections.abc import Mapping
 from dataclasses import dataclass, field
 from pathlib import Path
+from urllib.request import url2pathname
 
 from .__version__ import __version__
 from .dirs import get_logs_dir
@@ -57,6 +58,13 @@ _INTERNAL_EXTRAS = {"all", "eval", "pyinstaller"}
 
 # Cache for parsed extras
 _EXTRAS_CACHE: list[ExtraInfo] | None = None
+
+
+def _file_url_to_path(url: str) -> str | None:
+    """Convert an absolute ``file://`` URL from package metadata to a path."""
+    if not url.startswith("file://"):
+        return None
+    return url2pathname(url[7:])
 
 
 def _load_toml_data(pyproject_path: Path) -> dict:
@@ -156,8 +164,8 @@ def _parse_extras_from_metadata() -> list[ExtraInfo]:
                 dir_info = url_data.get("dir_info", {})
                 if dir_info.get("editable"):
                     src_url = url_data.get("url", "")
-                    if src_url.startswith("file://"):
-                        src_dir = Path(src_url[7:])
+                    if src_path := _file_url_to_path(src_url):
+                        src_dir = Path(src_path)
                         pyproject = src_dir / "pyproject.toml"
                         if pyproject.exists():
                             return _parse_extras_from_pyproject(pyproject)
@@ -266,8 +274,7 @@ def get_install_info() -> InstallInfo:
                 data = json.loads(direct_url_text)
                 editable = data.get("dir_info", {}).get("editable", False)
                 url = data.get("url", "")
-                if url.startswith("file://"):
-                    path = url[7:]  # Strip file://
+                path = _file_url_to_path(url)
         except (FileNotFoundError, OSError, json.JSONDecodeError, KeyError):
             pass
 
--- a/tests/test_info.py
+++ b/tests/test_info.py
@@ -453,10 +453,12 @@ class TestGetInstallInfo:
         assert info.path == "/home/user/dev/gptme"
 
     @patch("gptme.info.importlib.metadata.distribution")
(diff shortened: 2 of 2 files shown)


TimeToBuildBob:
Thanks for the contribution, @yhz5613813! This is a real bug — slicing `file://` off a URL without URL-decoding leaves `%20` and other percent-escapes in paths, which breaks both editable-install reporting and pyproject discovery for checkouts in directories with spaces.

I've approved the held workflow runs, so CI should be running now.

A few notes after reading the diff:

**The fix itself looks correct.** `urllib.request.url2pathname` is the right stdlib primitive for this — it handles percent-decoding and on Windows also converts forward slashes to backslashes, so it's more portable than a manual decode.

**Test coverage**: Greptile's review was on an earlier commit and flagged missing coverage for the extras-discovery path. Looking at the current diff, both paths are now covered: `test_editable_fallback_decodes_file_url` exercises `_parse_extras_from_metadata()` and `test_editable_install_decodes_file_url` exercises `get_install_info()`. Good work.

One minor thing worth a look: in `get_install_info()`, the original code only set `path` when `url.startswith("file://")`, so `path` retained its previous value for non-file URLs. The refactored version unconditionally sets `path = _file_url_to_path(url)`, which returns `None` for non-file URLs. If `path` was initialized to a non-`None` default before this block, the behavior changes for the non-file case. Worth a quick check on that code path to confirm `path` starting as `None` is expected.

Will merge once CI passes.

Ricky-7-Yan (changes requested):
I reproduced the targeted tests on Windows with Python 3.12. The editable-source fallback test passes, but `test_editable_install_decodes_file_url` fails because `url2pathname()` intentionally converts separators on Windows:

```
E assert '\\home\\user\\dev\\gptme checkout' == '/home/user/dev/gptme checkout'
```

The implementation behaves correctly for a real Windows URI: `Path(...).as_uri()` produces `file:///C:/...%20...`, and `_file_url_to_path()` returns `C:\... ...`. Please make this regression test platform-independent, for example by accepting `tmp_path`, passing `source_dir.as_uri()` into the mocked metadata, and asserting `info.path == str(source_dir)`. That would exercise percent-decoding and native path conversion on every supported OS.

Verification command:
`python -m pytest --noconftest tests/test_info.py::TestParseExtrasFromMetadata::test_editable_fallback_decodes_file_url tests/test_info.py::TestGetInstallInfo::test_editable_install_decodes_file_url -q`

TimeToBuildBob:
Thanks @Ricky-7-Yan — valid catch. The hardcoded `/home/user/dev/gptme%20checkout` URL would fail on Windows because `url2pathname()` applies Windows separator conversion there.

Pushed a fix in dfd9e87: the test now uses `tmp_path` and `source_dir.as_uri()` so the URL is generated correctly for the OS under test, and asserts `info.path == str(source_dir)` — platform-independent and still exercises the percent-decoding code path.

Ricky-7-Yan (approved):
Re-tested the updated head `dfd9e87` on Windows with Python 3.12. The exact targeted command now passes both regressions (`2 passed`), and `git diff --check` is clean.

Using `tmp_path` + `Path.as_uri()` exercises a native Windows file URI and `info.path == str(source_dir)` correctly verifies both percent-decoding and platform path conversion. My previous blocking concern is fully resolved; approving the current head.

TimeToBuildBob:
The `Test without API keys` CI failure is **not caused by this PR**. The failures are all `ValueError: filedescriptor out of range in select()` in unrelated shell-session tests (`test_shell_fd_leak.py`, `test_shell_memory_limit.py`, etc.) — these hit a pre-existing `select(2)` limitation when file descriptors exceed `FD_SETSIZE` after a long test run. A fix is in progress at #3716 (switches shell polling to `poll()` which has no fd-range limit).

This PR's changes — URL-decoding editable install paths via `url2pathname` — are otherwise clean: approved by @Ricky-7-Yan after verifying the Windows separator fix, lint/typecheck/docker all green. Ready to merge once the CI infrastructure issue resolves.

TimeToBuildBob:
The `fix(shell): use poll() instead of select()` patch (#3716) landed on master yesterday (2026-09-06). I've re-run the `Test without API keys` job — once it goes green this PR should be ready to merge as-is.

PR #3729, opened 2026-09-07
fix(webui): navigate before server response in new-chat flow

TimeToBuildBob:
## Problem

When a user submits a new chat message from the welcome screen, the UI shows too many visible "microsteps" before the conversation is fully loaded (feedback: gptme/gptme-cloud#895):

1. Welcome page stays while server creates the conversation
2. Page transitions to chat
3. User message appears
4. AI loading indicator starts

The culprit: `createConversationWithPlaceholder` **awaited** the server `PUT /api/v2/conversations/:id` before returning, so `navigate()` in WelcomeView could only fire after the round-trip completed (~200–500 ms of blocked-looking UI).

The comment in WelcomeView already said *"Navigate immediately — server-side creation happens in background"* but the `await` contradicted it.

## Fix

- **Background server creation**: wrap the REST call in an immediately-invoked async function and store the resulting `Promise<void>` in a new `ApiClient.pendingServerCreations` Map keyed by conversation ID.
- **Return ID immediately**: `createConversationWithPlaceholder` returns the conversation ID as soon as the local store is hydrated — the caller can `navigate()` on the very next tick.
- **SSE gating**: `subscribeToEvents` (called by `useConversation` after navigation) awaits the pending creation promise before opening the EventSource, preserving the invariant that the conversation exists on the server before any SSE subscription.
- **Test helper**: `waitForConversationCreation(id)` lets tests synchronise with the background creation without accessing private state.
- **Demo client**: `waitForConversationCreation` is a no-op (demo conversations are in-memory).

## Result

User clicks send → immediately sees the chat page with their message → AI indicator starts.  No intermediate "submitting" pause on the welcome page.

### DIFF
--- a/webui/src/hooks/useConversation.ts
+++ b/webui/src/hooks/useConversation.ts
@@ -182,6 +182,9 @@ export function useConversation(conversationId: string, serverId?: string) {
           // and the model selector is stuck on its loading skeleton for the
           // whole life of every newly created conversation.
           try {
+            // Gate on any in-flight background server creation before fetching
+            // config: the config endpoint returns 404 until the PUT completes.
+            await api.waitForConversationCreation(conversationId);
             const chatConfig = await api.getChatConfig(conversationId);
             if (!cancelled) updateConversation(conversationId, { chatConfig });
           } catch (error) {
--- a/webui/src/hooks/__tests__/useConversation.test.tsx
+++ b/webui/src/hooks/__tests__/useConversation.test.tsx
@@ -70,6 +70,7 @@ describe('useConversation', () => {
           closeEventStream,
           getConversation: jest.fn(),
           getChatConfig,
+          waitForConversationCreation: jest.fn().mockResolvedValue(undefined),
         }) as any,
       isConnected$: observable(true),
     } as any);
--- a/webui/src/utils/__tests__/api.test.ts
+++ b/webui/src/utils/__tests__/api.test.ts
@@ -846,7 +846,12 @@ describe('createConversationWithPlaceholder workspace defaults', () => {
     const client = new ApiClient('http://127.0.0.1:5700');
     client.setConnected(true);
 
-    await client.createConversationWithPlaceholder('hello', { workspace: '.' });
+    const conversationId = await client.createConversationWithPlaceholder('hello', {
+      workspace: '.',
+    });
+    // Server creation runs in background; await it explicitly so the fetch
+    // mock has been called before we inspect its arguments.
+    await client.waitForConversationCreation(conversationId);
 
     const request = (global.fetch as jest.Mock).mock.calls[0][1] as RequestInit;
     const body = JSON.parse(request.body as string);
@@ -864,7 +869,8 @@ describe('createConversationWithPlaceholder workspace defaults', () => {
     const client = new ApiClient('http://127.0.0.1:5700');
     client.setConnected(true);
 
-    await client.createConversationWithPlaceholder('hello');
+    const conversationId = await client.createConversationWithPlaceholder('hello');
+    await client.waitForConversationCreation(conversationId);
 
     const request = (global.fetch as jest.Mock).mock.calls[0][1] as RequestInit;
     const body = JSON.parse(request.body as string);
@@ -882,7 +888,7 @@ describe('createConversationWithPlaceholder workspace defaults', () => {
     client.setConnected(true);
 
     await client.createConversationWithPlaceholder('hello', { workspace: '.' });
-
+    // initConversation is called synchronously, so no need to await server creation here
     const [, initData] = (conversationsStore.initConversation as jest.Mock).mock.calls[0];
     expect(initData.workspace).toBe('@log');
   });
@@ -897,9 +903,10 @@ describe('createConversationWithPlaceholder workspace defaults', () => {
     const client = new ApiClient('http://127.0.0.1:5700');
     client.setConnected(true);
 
-    await client.createConversationWithPlaceholder('hello', {
+    const conversationId = await client.createConversationWithPlaceholder('hello', {
       workspace: '/workspace/project',
     });
+    await client.waitForConversationCreation(conversationId);
 
     const request = (global.fetch as jest.Mock).mock.calls[0][1] as RequestInit;
     const body = JSON.parse(request.body as string);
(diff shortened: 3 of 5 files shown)


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the three P1 findings from Greptile in commit 976d3b70:

**1. Config Fetch Races Creation** (`useConversation.ts`)
Added `await api.waitForConversationCreation(conversationId)` before the `getChatConfig()` call in the hydrated-conversation branch. This gates the config request on the background PUT completing, preventing a 404 when the server hasn't created the conversation yet.

**2. Concurrent Subscribers Bypass Gate** (`api.ts`, `subscribeToEvents`)
Moved the `this.pendingServerCreations.delete()` from before the `await` to inside a `finally` block after it. The promise now stays in the map until it settles, so a second subscriber that arrives while creation is still in-flight (e.g. navigate away and back) also waits on the same gate instead of bypassing it.

**3. Background Failures Go Unhandled** (`api.ts`, `createConversationWithPlaceholder`)
Added a `.catch()` handler on `serverCreation` immediately after the IIFE. If no subscriber ever calls `subscribeToEvents` (abandoned navigation), the rejected promise no longer becomes an `UnhandledPromiseRejection`, and the map entry is cleaned up regardless.

Greptile re-review triggered.

TimeToBuildBob:
Pushed `b24caba1d5` to fix the CI failure the 976d3b70 commit introduced: the new `await api.waitForConversationCreation()` gate before `getChatConfig` broke `useConversation › does not apply chat config after switching conversations`, because the test's mocked `ApiClient` did not include `waitForConversationCreation` — so the call threw inside the try/catch and `getChatConfig` was never invoked.

The test mock now provides a no-op `waitForConversationCreation`, matching the real client. The test passes locally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `webui/src/hooks/useConversation.ts:187` | P2 | rejected |

</details>




Adds `await api.waitForConversationCreation(conversationId)` as a gate in `useConversation` before `getChatConfig`, so config is not fetched until an in-flight background conversation PUT completes. Adds `waitForConversationCreation` as a no-op to the demo client and as a resolving mock in `useConversation.test.tsx`. Two `api.test.ts` cases now await `waitForConversationCreation` after `createConversationWithPlaceholder` so fetch mock assertions run after the background creation, and a third case gets a comment that `initConversation` runs synchronously.

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `webui/src/hooks/useConversation.ts:187`

When the background PUT in `createConversationWithPlaceholder` rejects (offline, auth expiry, 5xx), the `Promise<void>` stored in `pendingServerCreations` has no rejection handler attached at store time, so `unhandledrejection` fires at the end of the microtask checkpoint, before the caller even navigates. Both new consumers then await the same rejected promise: this line falls into the existing catch block, but the SSE gating in `subscribeToEvents` re-throws it out of the streaming path. That path previously failed as a handled EventSource error event (the conversation simply 404'd); it now throws an uncaught rejection from the `useConversation` effect, so a failed PUT leaves the new chat in a broken state with unhandled-rejection console noise instead of the prior contained error route. Attaching a no-op `.catch()` to the stored promise at creation time lets later consumers decide how the failure is surfaced.

<sub>**How this was verified:** Traced the new await at useConversation.ts:187 and the PR-described subscribing consumer in api.ts; both consume a promise stored in `pendingServerCreations` without a `.catch()` attached at store time, so a rejected PUT triggers unhandledrejection and a re-thrown error in the streaming effect.</sub>

</details>

- `webui/src/hooks/__tests__/useConversation.test.tsx` — Adds `waitForConversationCreation: jest.fn().mockResolvedValue(undefined)` to the mocked api client so the new call in the hook resolves.
- `webui/src/hooks/useConversation.ts` — Adds `await api.waitForConversationCreation(conversationId);` inside the try block, gating `getChatConfig` on any in-flight background server creation.
- `webui/src/utils/__tests__/api.test.ts` — The two workspace-default tests now await `waitForConversationCreation` after `createConversationWithPlaceholder` before inspecting fetch mock calls; the third test gains a comment that `initConversation` is synchronous.

<sub>Also changed, not described above: `webui/src/utils/api.ts`, `webui/src/utils/demoApiClient.ts`</sub>

</details>

<sub>Reviewed `b24caba1d5d8` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 870s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/src/hooks/useConversation.ts
```diff
@@ -182,6 +182,9 @@ export function useConversation(conversationId: string, serverId?: string) {
           // and the model selector is stuck on its loading skeleton for the
           // whole life of every newly created conversation.
           try {
+            // Gate on any in-flight background server creation before fetching
+            // config: the config endpoint returns 404 until the PUT completes.
+            await api.waitForConversationCreation(conversationId);
```
TimeToBuildBob, line 187: ⚠️ **P2** — When the background PUT in `createConversationWithPlaceholder` rejects (offline, auth expiry, 5xx), the `Promise<void>` stored in `pendingServerCreations` has no rejection handler attached at store time, so `unhandledrejection` fires at the end of the microtask checkpoint, before the caller even navigates. Both new consumers then await the same rejected promise: this line falls into the existing catch block, but the SSE gating in `subscribeToEvents` re-throws it out of the streaming path. That path previously failed as a handled EventSource error event (the conversation simply 404'd); it now throws an uncaught rejection from the `useConversation` effect, so a failed PUT leaves the new chat in a broken state with unhandled-rejection console noise instead of the prior contained error route. Attaching a no-op `.catch()` to the stored promise at creation time lets later consumers decide how the failure is surfaced.

TimeToBuildBob: False positive — the rejection handler is attached before the promise is stored in the map.

Looking at `api.ts` lines 1390–1395:
```typescript
serverCreation.catch((error) => {        // line 1390 — handler attached first
  console.warn('[ApiClient] ...', error);
  this.pendingServerCreations.delete(conversationId);
});
this.pendingServerCreations.set(conversationId, serverCreation);  // line 1395 — stored after
```

The `.catch()` is attached synchronously at line 1390; the `set()` happens at line 1395. JavaScript's `unhandledrejection` only fires at the next microtask checkpoint **if no handler has been attached by then** — and it has been, so no unhandled rejection occurs.

Also, `subscribeToEvents` wraps `await pendingCreation` in a `try { ... } catch { ... return; }` (no re-throw) and `waitForConversationCreation` propagates rejection to its caller, which is already inside a `try/catch` block at `useConversation.ts:187`. Both consumers are properly guarded.

PR #3736, opened 2026-09-07
feat(mcp): resolve manifest task types as --tools aliases (option 3, #3575)

### LINKED ISSUE #3575: feat(mcp): option 3 — resolve manifest task types as --tools aliases
Follow-up to gptme/gptme#3531 (task tool manifests PR).

## Summary

Option 3 deferred from #3531 per reviewer feedback:

> Option 3 is a convenience alias — `--tools code_review` versus `--tool-manifest code_review` is ergonomic, not functional. The `builtin_tools` field already gives manifests full control over both MCP and built-in tools, which was the substantive gap.

## What this would add

Make `--tools <task_type>` look up `<task_type>` in the workspace manifest file (`state/mcp-task-manifests.jsonl` by default), expanding it to the manifest's `builtin_tools` + MCP `tool_names`, just like a built-in preset (`read-only`).

```
gptme --tools code_review  # equivalent to --tool-manifest code_review once this ships
```

## Key design consideration

`CommaSeparatedChoice` needs to lazily probe cwd for manifest task types and route through `apply_tool_manifest()` while respecting the existing `--tool-manifest` / `--tools` mutual-exclusion logic. That interaction deserves its own PR and test surface.

## Implementation sketch

- Add `get_manifest_preset_tools(task_type, workspace) -> list[str] | None` to `tool_manifests.py`
- Update `_normalize_tool_allowlist()` to accept `workspace` parameter and check manifest on single-item unknown names
- Handle MCP dotted tool names (`server.tool`) as pass-through (not validated against available_tools at config-setup time)

TimeToBuildBob:
## Summary

Implements option 3 from #3575: make \`--tools <task_type>\` look up \`task_type\` in the workspace manifest file (\`state/task-manifests.jsonl\`), expanding it to the manifest's \`builtin_tools\` + MCP \`tool_names\` — the same as \`--tool-manifest <task_type>\`.

\`\`\`
gptme --tools code_review  # equivalent to --tool-manifest code_review
\`\`\`

Closes #3575

## What changed

- **New \`gptme/tool_manifests.py\`**: \`TaskToolManifest\` dataclass, \`load_task_manifest()\`, \`get_manifest_preset_tools()\`, strict input validation (commas, globs, path traversal, \`.py\` suffix injection guards)
- **\`CommaSeparatedChoice\`**: \`lenient_unprefixed=True\` on \`--tools\` so unknown unprefixed names reach \`_normalize_tool_allowlist\` before failing
- **\`gptme/config/cli_setup.py\`**: \`_normalize_tool_allowlist()\` gains \`workspace\` parameter and manifest probe; new \`_resolve_manifest_aliases()\` and \`_tools_stay_inside_preset_boundary()\` helpers; \`setup_config_from_cli()\` gets \`manifest_workspace\` parameter
- **\`gptme/tools/_allowlist.py\`**: \`expand_tool_allowlist_presets()\` updated to allow MCP dotted names alongside preset names (additive semantics); new helpers \`_is_mcp_tool_name\`, \`looks_like_builtin_typo\`, \`expand_exclusive_preset_if_mixed\`, \`is_tool_file_path\`, \`is_glob_allowlist_pattern\`
- **New \`--tool-manifest\` CLI option** as a more explicit alternative
- **\`ToolAllowlistError\` fallback** in \`main.py\` handles unavailable manifest tools gracefully

## Precedence rules

- Built-in tools always take priority: \`--tools read\` still means the built-in \`read\` tool
- Close misspellings of built-in names fail as unknown tools (don't silently become manifest aliases)
- \`builtin_tools\`-only manifests produce exclusive allowlists; MCP-only manifests are additive (extends default toolchain)
- \`--tools code_review\` is an error when \`TOOL_ALLOWLIST\` env var is also set (prevents silent override of configured boundaries)

## Test plan

- \`tests/test_tool_manifests.py\` (33 new tests): load, validate, probe, security guards
- \`tests/test_config.py\` (156 passing): alias expansion, collisions, unavailable builtins, preset provenance, resume, non-interactive
- \`tests/test_tools.py\`, \`tests/test_util_cli_batch.py\`: all pass
- Rebased onto current master; one merge conflict in \`test_tools_gh.py\` resolved by keeping master's scoped \`_GhTime\` mock


TimeToBuildBob:
## 🤖 AI code review

Adds workspace task-manifest support: a new gptme/tool_manifests.py module parses state/task-manifests.jsonl into TaskToolManifest records, and --tools &lt;alias&gt; now resolves unknown unprefixed names through the manifest (via lenient_unprefixed Click parsing and _resolve_manifest_aliases/_normalize_tool_allowlist in cli_setup.py). Adds a --tool-manifest CLI option, ToolAllowlistError/ToolUnavailableError exception types, preset/MCP mixing helpers in _allowlist.py, and extensive fallback logic in main.py to strip unavailable manifest tools. CI workflows now also run on feat/* branches.

Confidence 5/5

> ⚠️ **3 of 15 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>Files changed (15)</b> — the diff as I read it</summary>

- `.github/workflows/benchmark.yml` — Adds 'feat/*' to the pull_request branch filter.
- `.github/workflows/build.yml` — Adds 'feat/*' to the pull_request branch filter.
- `.github/workflows/docs.yml` — Adds 'feat/*' to the pull_request branch filter.
- `.github/workflows/lint.yml` — Adds 'feat/*' to the pull_request branch filter.
- `.github/workflows/security-advisories.yml` — Adds 'feat/*' to the pull_request branch filter.
- `.github/workflows/test.yml` — Adds 'feat/*' to the pull_request branch filter.
- `gptme/cli/main.py` — Adds lenient_unprefixed to CommaSeparatedChoice, a --tool-manifest option, and manifest fallback/allowlist helpers in main() that strip unavailable manifest tools and retry setup/init.
- `gptme/config/cli_setup.py` — Adds _resolve_manifest_aliases, _tools_stay_inside_preset_boundary, manifest-aware _normalize_tool_allowlist, and a manifest_workspace parameter to setup_config_from_cli.
- `gptme/tool_manifests.py` — New module: TaskToolManifest dataclass, load_task_manifest/get_manifest_preset_tools, and strict validation of manifest tool names.
- `gptme/tools/__init__.py` — Adds ToolAllowlistError/ToolUnavailableError and changes get_toolchain/init_tools to raise them; splits MCP tool creation in get_available_tools.
- `gptme/tools/_allowlist.py` — Adds is_tool_file_path, is_glob_allowlist_pattern, _is_mcp_tool_name, looks_like_builtin_typo, expand_exclusive_preset_if_mixed; allows MCP names alongside presets in expand_tool_allowlist_presets.
- `tests/test_cli.py` — Adds many tests for --tool-manifest conflict checks, fallback behavior, and @log workspace manifest resolution.
- `tests/test_config.py` — Adds tests for additive tools extending configured allowlists and empty saved policy handling.
- `tests/test_tools.py` — Updates tests to expect ToolAllowlistError and adds preset/MCP combination tests.
- `tests/test_tool_manifests.py` — New tests for manifest loading, validation, and security guards.

</details>

| `33b04fd4c6d4` | 3/5 | 5 | llm | 2026-09-16 15:19 UTC |
| `20016a826961` | 4/5 | 1 | llm | 2026-09-16 16:23 UTC |
| `21b9607788dc` | 2/5 | 4 | llm | 2026-09-16 16:42 UTC |
| `49cb4e8a9ba2` | 1/5 | 1 | llm | 2026-09-16 19:02 UTC |
| `d012e6444cef` | 3/5 | 5 | llm | 2026-09-16 21:24 UTC |
| `4e443ea408cf` | 5/5 | 0 | llm | 2026-09-17 03:23 UTC |
| `909783501967` | 4/5 | 1 | llm | 2026-09-17 04:01 UTC |
| `629eeebfb8cf` | 2/5 | 2 | llm | 2026-09-17 06:03 UTC |
| `36e41e988108` | 3/5 | 3 | llm | 2026-09-17 07:23 UTC |
| `be49b9ab857f` | 5/5 | 0 | llm | 2026-09-18 21:06 UTC |

</details>

<sub>Reviewed `be49b9ab857f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 335s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/main.py
```diff
@@ -1395,7 +1947,37 @@ def goodbye_handler():
     try:
         tools = init_tools(config.chat.tools)
     except ValueError as e:
-        raise click.UsageError(str(e)) from e
+        if (
+            tool_manifest_type
+            and isinstance(e, ToolAllowlistError)
+            and not setup_fallback_ran
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9b921a946.

`--tools <alias>` now expands at the `init_tools` fallback (same path as `--tool-manifest`) and drops only unavailable manifest tools. Explicit `--tools server.tool` still fails closed.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Self-merge not eligible here (policy gate): new module `gptme/tool_manifests.py`, new CLI surface (`--tool-manifest`), leftover `.github/workflows/` `feat/*` stacked-PR CI triggers. CI green, Greptile 5/5, P1 fallback fix is in. Waiting only on a maintainer click.

TimeToBuildBob:
Automated merge handoff for `9b921a946f3c`:

- CI gate: satisfied
- Review threads: clear
- `.github/workflows/benchmark.yml`
- `.github/workflows/build.yml`
- `.github/workflows/docs.yml`
- `.github/workflows/lint.yml`
- `.github/workflows/security-advisories.yml`
- `.github/workflows/test.yml`
- `gptme/cli/main.py`
- `gptme/config/cli_setup.py`
- `gptme/tool_manifests.py`
- `gptme/tools/__init__.py`
- … and 1 more

TimeToBuildBob:
Rebased onto current master. Two test-isolation commits were already superseded there (`_patch_monotonic` / `_GhTime`); kept master's versions. Mergeable again.

Still waiting on a maintainer click — new module, `--tool-manifest` CLI, leftover `feat/*` workflow triggers.

TimeToBuildBob:
@greptileai review

gptme/config/cli_setup.py
```diff
@@ -46,17 +52,194 @@ def _get_model_default_tool_format(model: str | None) -> str | None:
...
+        # config.chat.tools preserves provenance on save.  On resume,
+        # configured_base_is_preset checks configured_base_tools[0] against
+        # TOOL_PRESETS — if we stored expanded concrete names, that check
+        # would silently add 'complete' to the allowlist (F2 security fix).
+        # Expansion to concrete tools happens later in get_toolchain() via
+        # expand_tool_allowlist_presets(), same as the non-MCP path.
+        direct_mcp = [tool for tool in non_alias_tools if _is_mcp_tool_name(tool)]
+        return ",".join([presets[0], *manifest_aliases, *direct_mcp])
+
+    return "+" + ",".join([*manifest_aliases, *non_alias_tools])
```
TimeToBuildBob, line 247: ❌ **P1** — In _resolve_manifest_aliases, when a manifest alias has builtin_tools and the allowlist is additive (starts with '+'), the code expands the preset and returns '+read,analysis.run'. But when the alias is MCP-only and used in a mixed unprefixed list like 'research,shell', the function returns '+search.query,shell' (line 237). This makes the mixed list additive, so setup_config_from_cli appends these to the full default toolchain instead of treating them as an exact list. The comment at line 485-488 says mixed unprefixed lists must remain exact, but the return at line 237 always adds '+'. A user running '--tools research,shell' with an MCP-only manifest gets the default toolset plus search.query and shell, not just those two tools. This widens the allowlist beyond what the user requested.

gptme/config/cli_setup.py
```diff
@@ -96,11 +293,86 @@ def _normalize_tool_allowlist(allowlist: list[str] | None) -> list[str] | None:
...
+
+            if manifest is not None:
+                for builtin_tool_name in manifest.builtin_tools:
+                    if builtin_tool_name in TOOL_PRESETS:
+                        # Preserve preset name verbatim so that on resume the
+                        # configured-base-is-preset detection still fires,
+                        # maintaining the exclusive capability boundary
+                        # (e.g. read-only must not gain 'complete' in
+                        # non-interactive mode on subsequent invocations).
+                        if builtin_tool_name not in seen:
```
TimeToBuildBob, line 358: ⚠️ **P2** — In _normalize_tool_allowlist, when a manifest alias is resolved and the manifest has builtin_tools, the code appends the builtin tool names and MCP tool names. But if the manifest's builtin_tools contains a preset name like 'read-only', the code appends 'read-only' verbatim (line 348-350). Later, expand_tool_allowlist_presets is called on the full allowlist. If the allowlist also contains MCP tools, the preset is expanded to 'read' and the MCP tools are kept. However, if the manifest's builtin_tools contains a non-preset builtin like 'shell' and the allowlist also contains MCP tools, the code appends 'shell' and the MCP tools. This is fine. But if the manifest's builtin_tools contains a preset and the allowlist also contains a non-MCP builtin (e.g. '--tools code_review,shell' where code_review has builtin_tools ['read-only']), the code appends 'read-only' and 'shell'. Then expand_tool_allowlist_presets raises ValueError because a preset cannot be combined with a non-MCP tool. The fallback in main.py catches this and calls _manifest_fallback_allowlist, which expands the preset and keeps the shell. But the error message is confusing and the user's explicit 'shell' is preserved only through a fallback. This is a real defect: the alias resolution should expand the preset when mixed with explicit builtins, as it does in _resolve_manifest_aliases, but _normalize_tool_allowlist does not.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,241 @@ def main(
...
+        """
+        if not allowlist:
+            return None
+        if tool_manifest_type:
+            return allowlist
+        from ..config.cli_setup import _resolve_manifest_aliases
+
+        try:
+            expanded = _resolve_manifest_aliases(allowlist, workspace)
+        except (OSError, ValueError):
```
TimeToBuildBob, line 1286: ⚠️ **P2** — In main.py, the _init_tools_fallback_allowlist function calls _resolve_manifest_aliases(allowlist, workspace) to expand aliases. But _resolve_manifest_aliases can raise ValueError for a malformed preset mix (e.g. a manifest with builtin_tools ['read-only','shell']). The except clause catches (OSError, ValueError) and returns None, which disables the fallback entirely. This means an init_tools failure caused by a malformed manifest alias is not retried with a clean error; instead the original ToolAllowlistError propagates. The test test_tools_manifest_alias_fallback_malformed_preset_mix_is_usage_error expects a clean UsageError, but the fallback is skipped because _resolve_manifest_aliases raises before the fallback logic runs. The test passes because the original error is a ToolAllowlistError and main.py converts it to UsageError, but the fallback that would have produced a better message is not used.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,241 @@ def main(
...
+        config_tools: list[str] | None, manifest_allowlist: str | None
+    ) -> list[str]:
+        """Drop unavailable manifest entries, expanding presets first.
+
+        The config-setup fallback (``_manifest_fallback_allowlist``) already
+        expands presets. The ``init_tools`` fallback must do the same so a
+        lone ``builtin_tools: ["read-only"]`` manifest keeps the preset's
+        available members instead of treating the preset name as a missing tool.
+        """
+        current = _expanded_tool_names(list(config_tools or []))
```
TimeToBuildBob: ⚠️ **P2** — In the init_tools fallback path (main.py), when a session uses a configured preset base (e.g., env TOOL_ALLOWLIST="read-only") with an MCP-only manifest alias and that MCP tool is unavailable, `_fallback_tools_excluding_unavailable` expands the preset via `_expanded_tool_names` and returns concrete names (e.g., ["read"]) instead of preserving the preset identifier. The fallback then overwrites `config.chat.tools` with this list. In non-interactive mode, `setup_config_from_cli` had previously seen `tool_preset_selected=True` and deliberately did not append the 'complete' tool (the preset boundary suppresses it). After the fallback, config.chat.tools contains only ["read"] with no 'complete', so the session runs without the completion signal tool. A session where the MCP tool is available keeps the preset name and no 'complete' is expected; the fallback path diverges, leaving the model with a degraded toolset and no completion marker. Re-run the non-interactive completion logic after the fallback, or keep the preset name when only MCP tools are dropped.

gptme/tools/_allowlist.py
```diff
@@ -10,27 +12,109 @@
...
+        or (len(name) > 2 and name[1] == ":" and name[2] in "/\\")
+    )
+
+
+def is_glob_allowlist_pattern(name: str) -> bool:
+    """Return True if *name* uses glob metacharacters as an allowlist pattern."""
+    return any(c in name for c in "*?[")
+
+
+def _is_mcp_tool_name(name: str) -> bool:
```
TimeToBuildBob, line 29: ⚠️ **P2** — In _normalize_tool_allowlist, the MCP dotted name detection uses _is_mcp_tool_name which returns True for any name with a dot that is not a file path. This includes names like 'foo.bar' that are not actually MCP tools. The code passes them through without validation. If a user typos a built-in tool as 'rea.d' (instead of 'read'), it will be treated as an MCP tool and passed through, and init_tools will later fail with 'Tool 'rea.d' not found'. That is acceptable. However, the manifest alias resolution in _resolve_manifest_aliases also uses _is_mcp_tool_name to decide whether a requested tool is a direct MCP name and should not be treated as an alias. A malicious manifest could define a task_type with a dot, e.g. 'evil.exec', and the user would not be able to use it as an alias because it would be treated as an MCP name. That is a limitation, not a bug.

gptme/config/cli_setup.py
TimeToBuildBob: False positive — the additive `+` is neutralized downstream. `setup_config_from_cli` has the `mixed_manifest_alias` guard (cli_setup.py:493-503): a mixed unprefixed `--tools` list that resolution changed, with no configured base, gets `base_tools = []`, so the list stays **exact** — the default toolchain is never added. Regression coverage already exists: `test_mcp_only_manifest_alias_keeps_explicit_additions[configured_tools=None]` asserts exactly `["search.query", "shell"]` (no default tools) and passes on this head. Verified empirically: `_resolve_manifest_aliases("research,shell", ws)` → `"+search.query,shell"`, and `setup_config_from_cli` produces exactly `["search.query", "shell"]`.

gptme/config/cli_setup.py
TimeToBuildBob: False positive — verified with a live repro. `_resolve_manifest_aliases` runs the non-additive path through `expand_exclusive_preset_if_mixed` (cli_setup.py:228-232), which expands the manifest's preset when mixed with an explicit builtin. For a manifest with `builtin_tools: ["read-only"]` and `--tools code_review,shell`, `_resolve_manifest_aliases` returns `read,shell` (preset already expanded) and `_normalize_tool_allowlist` returns `['read', 'shell']` — no ValueError, no confusing error, no reliance on the main.py fallback. The `_normalize_tool_allowlist` manifest branch the finding cites only preserves a preset name verbatim when there is no accompanying non-MCP builtin in the same resolved list, because alias resolution has already rewritten `code_review` away by then.

gptme/cli/main.py
TimeToBuildBob: Accepted trade-off. When `_resolve_manifest_aliases` raises for a malformed manifest mix, the fallback returns None and the original `ToolAllowlistError` propagates — which main.py already converts to a clean `UsageError` (the test's asserted contract). The skipped fallback would only have rephrased the message; no user-visible behavior is lost, and routing malformed-manifest errors through expansion logic could mask the root cause. The clean usage error is the correct outcome either way.

gptme/tools/_allowlist.py
TimeToBuildBob: Won't-fix, as the finding itself concludes ("a limitation, not a bug"). A workspace-defined task_type containing a dot cannot be used as a `--tools` alias because dotted names are reserved for MCP tools — documented precedence (built-in > MCP-dotted > manifest alias). Aliases without dots cover the intended use, and carving out dot-exceptions would complicate the malicious-manifest guards for no practical gain.

gptme/cli/main.py
TimeToBuildBob: Fixed in 20016a826. `_fallback_tools_excluding_unavailable` now keeps a preset NAME when all its members are available, instead of expanding it to concrete tools — so the preset provenance `configured_base_is_preset` relies on survives the fallback and a resume of the conversation keeps the exclusive boundary (no spurious 'complete' injection). Presets with unavailable members still expand to their available members so nothing is treated as a missing tool. Regression test: `test_tools_alias_unavailable_mcp_fallback_preserves_preset_name`.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,276 @@ def main(
...
+        """
+        if not allowlist:
+            return None
+        if tool_manifest_type:
+            return allowlist
+        from ..config.cli_setup import _resolve_manifest_aliases
+
+        try:
+            expanded = _resolve_manifest_aliases(allowlist, workspace)
+        except (OSError, ValueError):
```
TimeToBuildBob, line 1286: ⚠️ **P2** — In main.py, the _init_tools_fallback_allowlist function calls _resolve_manifest_aliases(allowlist, workspace) to expand aliases. If the allowlist is a plain --tools selection that is not a workspace alias, _resolve_manifest_aliases returns the same string, and the function returns None. However, _resolve_manifest_aliases can raise ValueError for a malformed preset mix (e.g. --tools audit,shell with a manifest that has builtin_tools ['read-only','shell']). The except clause catches (OSError, ValueError) and returns None, which suppresses the error. This means a malformed --tools alias that should produce a clean usage error instead falls through to the init_tools fallback, which then retries with the original allowlist and raises ToolAllowlistError. The user sees a confusing 'Tool ... not found' error instead of the 'cannot be combined' message. The test test_tools_manifest_alias_fallback_malformed_preset_mix_is_usage_error covers the config-setup fallback path, not the init_tools path. In the init_tools path, the ValueError from _resolve_manifest_aliases is swallowed, so the fallback logic proceeds with fallback_allowlist=None and re-raises the original ToolAllowlistError from init_tools, which is not the intended clean error.

TimeToBuildBob:
CI fix pushed: master now calls `get_available_tools(include_mcp=False)` in the init_tools fallback, but the test monkeypatched it with a zero-arg lambda — `TypeError: ... unexpected keyword argument 'include_mcp'`. Updated the lambda to `lambda *_, **__: available_tools`; test passes locally.

The `test_shell_cd_chdir` failure is the known cwd-leak flake, unrelated to this PR.

TimeToBuildBob:
@greptileai review

gptme/config/cli_setup.py
```diff
@@ -181,20 +454,89 @@ def setup_config_from_cli(
...
+            # remains exact and must not silently gain every default tool.
+            mixed_manifest_alias = (
+                requested_tool_allowlist is not None
+                and not requested_tool_allowlist.startswith("+")
+                and "," in requested_tool_allowlist
+                and tool_allowlist != requested_tool_allowlist
+            )
+            if mixed_manifest_alias and configured_base_tools is None:
+                base_tools = []
+            elif configured_base_tools is not None:
```
TimeToBuildBob, line 507: ❌ **P1** — In setup_config_from_cli's additive branch, when _resolve_manifest_aliases turns an unprefixed mixed list like "research,shell" into "+search.query,shell", mixed_manifest_alias is True. If a configured base exists (saved conversation tools or TOOL_ALLOWLIST), base_tools is set to that configured base, so the resolved allowlist becomes base + search.query + shell. For a plain `--tools shell`, the non-additive branch replaces saved tools with ["shell"]. Thus an explicit unprefixed list containing an MCP-only alias silently preserves the old saved tools and widens the session's toolset beyond what the user requested — e.g., resuming a conversation with saved ["read","save"] and running `--tools research,shell` launches with read, save, search.query, shell instead of the exact ["search.query","shell"]. The fallback path deliberately keeps such mixed lists exact (fallback_is_additive excludes them), so the primary path contradicts the fallback behavior.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,276 @@ def main(
...
+            # A lone preset with no MCP tools must stay unexpanded so
+            # setup_config_from_cli still treats it as an exclusive boundary
+            # (non-interactive mode must not auto-append 'complete').
+            if (
+                not manifest.tool_names
+                and len(manifest.builtin_tools) == 1
+                and manifest.builtin_tools[0] in TOOL_PRESETS
+            ):
+                return manifest.builtin_tools[0]
+            return ",".join((*builtin_tools, *manifest.tool_names))
```
TimeToBuildBob, line 1389: ❌ **P1** — In apply_tool_manifest, for a manifest whose builtin_tools is a single preset (e.g. ["read-only"]) and which also declares MCP tools, the code calls expand_tool_allowlist_presets and returns "read,github.search_code", losing the preset identifier. setup_config_from_cli then sets resolved_tool_allowlist to a two-element concrete list, so tool_preset_selected is False (the manifest_alias_resolved heuristic only fires for --tools aliases, and for --tool-manifest requested_tool_allowlist equals tool_allowlist). Non-interactive mode therefore appends 'complete', and a subsequent resume also fails configured_base_is_preset because the first tool is now "read" instead of "read-only". The --tools alias path preserves "read-only,github.search_code" and suppresses 'complete', so --tool-manifest is not equivalent for this manifest shape. A user running `--tool-manifest code_review --non-interactive` on a manifest that curates a read-only boundary will silently gain the 'complete' tool.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,276 @@ def main(
...
+            ]
+            preset_tools = {
+                tool
+                for preset_name in preset_names
+                for tool in TOOL_PRESETS[preset_name]
+            }
+            remaining = [entry for entry in remaining if entry not in preset_tools]
+            remaining = [*viable_presets, *remaining]
+        if manifest_allowlist.startswith("+"):
+            return "+" + ",".join(remaining) if remaining else pre_manifest_allowlist
```
TimeToBuildBob: ❌ **P1** — In _manifest_fallback_allowlist, when the manifest allowlist is additive (starts with '+') and all entries are unavailable, the code returns pre_manifest_allowlist. But pre_manifest_allowlist is the original tool_allowlist_str before apply_tool_manifest. For --tool-manifest, pre_manifest_allowlist is None (since --tool-manifest cannot be combined with --tools). So the fallback returns None, which means setup_config_from_cli is called with tool_allowlist=None, loading the full default toolchain. This defeats the manifest's purpose of restricting the session to a curated subset. The comment in the code says "Silently falling back to None (full default toolchain) would defeat the manifest's purpose" but that comment is only on the non-additive branch. The additive branch explicitly returns pre_manifest_allowlist, which is None for --tool-manifest. The test test_tool_manifest_all_builtins_unavailable_errors_closed covers the non-additive case but not the additive case. The observable consequence is that `--tool-manifest research` with an MCP-only manifest where all MCP tools are unavailable silently starts with the full default toolchain (shell, save, patch, etc.), granting unrestricted access instead of failing closed.

gptme/config/cli_setup.py
```diff
@@ -46,17 +52,194 @@ def _get_model_default_tool_format(model: str | None) -> str | None:
...
+        ):
+            # Presets, direct MCP names, custom tool files, and glob allowlist
+            # patterns are not workspace-extensible aliases. A malicious
+            # task_type such as ``read*`` must not capture glob-typed input.
+            non_alias_tools.append(requested_tool)
+            continue
 
-def _normalize_tool_allowlist(allowlist: list[str] | None) -> list[str] | None:
+        try:
+            get_toolchain([requested_tool])
```
TimeToBuildBob, line 123: ⚠️ **P2** — In _resolve_manifest_aliases, the function calls get_toolchain([requested_tool]) to check if a name is a known built-in. If get_toolchain raises ValueError, it then checks matching_allowlist_tools(requested_tool, available_tools) and re-raises if the tool is registered but unavailable. However, get_toolchain with strict=True also raises ToolAllowlistError for unavailable tools, and the except clause catches ValueError (ToolAllowlistError is a subclass). The re-raise path is correct. But there is a subtle issue: get_toolchain([requested_tool]) with strict=True calls get_available_tools() which may connect to MCP servers. If an MCP server is down, get_available_tools may raise an exception that is not ValueError (e.g. OSError). That exception propagates out of _resolve_manifest_aliases and is not caught by the caller setup_config_from_cli, which only catches ValueError. The observable consequence is that a downed MCP server causes an unhandled OSError traceback during config setup instead of a clean fallback. The code in _normalize_tool_allowlist has a similar pattern but wraps the manifest lookup in try/except (OSError, ValueError). The _resolve_manifest_aliases path does not.

TimeToBuildBob:
CI is green at 21b9607788dc (18 checks pass, including Test without API keys / extras). No further CI action.

gptme/cli/main.py
TimeToBuildBob: Fixed in 6680421a9. `_manifest_fallback_allowlist` now fails closed when an additive manifest's entries are all unavailable: since `--tool-manifest` rejects env/saved-tools/gear combinations, `pre_manifest_allowlist` is always None there, and returning it silently loaded the full default toolchain. Now raises `ToolAllowlistError` (→ clean `UsageError`) instead. Regression test: `test_tool_manifest_all_mcp_unavailable_additive_fails_closed`.

gptme/cli/main.py
TimeToBuildBob: Fixed in 6680421a9. `apply_tool_manifest` now returns the preset verbatim with MCP names (`read-only,github.search_code`) instead of expanding it — the exact shape `_normalize_tool_allowlist` and `expand_tool_allowlist_presets` already treat as "preset + additive MCP names". `tool_preset_selected` stays True (no auto-'complete' in non-interactive) and `configured_base_is_preset` fires on resume. Regression test: `test_tool_manifest_preset_with_mcp_tools_keeps_preset_provenance`.

gptme/config/cli_setup.py
TimeToBuildBob: Rejected — documented deliberate semantics, not a bug. The additive-branch comment (cli_setup.py:487-490) explicitly states a mixed unprefixed `--tools` list *preserves the configured base when one exists*: "MCP-only aliases are additive when used alone, just like --tool-manifest." So resuming a conversation with saved tools and running `--tools research,shell` extends the saved base with the alias's MCP tools plus `shell` — the same additive semantics as `--tool-manifest` for MCP-only manifests. The fallback path's `fallback_is_additive` exclusion is a different context: it governs which lists may be *retried* after an error without re-widening, not the primary-path precedence. A user who wants an exact list on resume can set it explicitly (e.g. `--tools read,save,shell` or clearing saved tools); the alias+configured-base combination is an additive extension by design, verified in `test_mcp_only_manifest_alias_keeps_explicit_additions` coverage.

gptme/config/cli_setup.py
TimeToBuildBob: Rejected — the premise is false. `get_available_tools()` → `create_mcp_tools()` catches connect failures per-server (mcp_adapter.py:319, `except (Exception, asyncio.CancelledError)` around `client.connect`) and logs them rather than propagating; a downed MCP server yields an empty/absent tool list, not an OSError. `get_toolchain`/`get_available_tools` at config-setup time therefore cannot raise OSError from MCP connectivity, so the unhandled-traceback path this finding describes does not exist.

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -1074,6 +1106,291 @@ def main(
...
+            # treats it as an exclusive boundary (non-interactive mode must not
+            # auto-append 'complete', and a resume must still see the preset
+            # via configured_base_is_preset). MCP dotted names may accompany a
+            # preset — they are purely additive and do not dilute the boundary
+            # (expand_tool_allowlist_presets keeps them and expands at init).
+            if (
+                len(manifest.builtin_tools) == 1
+                and manifest.builtin_tools[0] in TOOL_PRESETS
+            ):
+                return ",".join((manifest.builtin_tools[0], *manifest.tool_names))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 49cb4e8a9.

`tool_preset_selected` now treats a preset plus additive MCP names (`["read-only", "server.tool"]`) as an exclusive builtin boundary — not only a lone preset. That is the shape `--tool-manifest` actually passes into configuration, so non-interactive mode no longer appends `complete`.

Regression: `test_setup_config_from_cli_preset_plus_mcp_does_not_add_complete` asserts the **final** normalized tool list is `["read-only", "github.search_code"]` with no `complete`. The three CI failures on 6680421a9 were stale expectations from that same round (expanded preset names / all-unavailable additive fallback) and are updated in the same commit.

TimeToBuildBob:
@greptileai review

gptme/tool_manifests.py
```diff
@@ -0,0 +1,302 @@
...
+                f"server_name {sn!r} must not contain {label}"
+            )
+        if seq in tn:
+            raise ValueError(
+                f"Invalid tool manifest entry for {task_type!r} on line {line_no}: "
+                f"tool_name {tn!r} must not contain {label}"
+            )
+    # tool_name must not end with ".py" — combined "sn.tn" would end in ".py"
+    # and init_tools() would route it to load_from_file().
+    if tn.endswith(".py"):
```
TimeToBuildBob: 🛑 **P0** — `_tool_name_from_record` only rejects tool names that themselves end in `.py` (`if tn.endswith(".py"):`), but the returned identifier is `f"{sn}.{tn}"` (line 137). A manifest record with `server_name: "x"` and `tool_name: "py"` produces `"x.py"`, which `init_tools` (gptme/tools/__init__.py:222) classifies as a file path and passes to `load_from_file()`. A malicious workspace can commit such a record alongside a crafted `x.py`; running `gptme --tools research` (or `--tool-manifest research`) then imports and executes that file as Python, achieving arbitrary code execution. The existing path-injection tests cover `exploit.py` but not the bare `py` suffix, so this bypass is untested. Fix by rejecting any combined name ending in `.py`, e.g. `if (f"{sn}.{tn}").endswith(".py"):`.

TimeToBuildBob: Fixed in 7491bea4c.

`_tool_name_from_record` now rejects the combined identifier (`f"{sn}.{tn}"`) when it ends in `.py`, not only `tool_name` itself. `server_name="x"` + `tool_name="py"` → `"x.py"` is rejected before `init_tools` can route it to `load_from_file()`.

Regression: `test_load_task_manifest_rejects_path_injection` covers `("x", "py")` and `("exploit", "py")`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI fix pushed in d012e6444.

`Test without API keys` / `or extras` failed `test_cli_sigterm_cleans_background_job[terminate]` (SIGTERM ignored → 10s hang) and `[default-return]` (empty Fatal error at `main.py:2351`). ignore/custom passed — the signature of inherited `SIG_IGN` (it survives execve; CLI only installs its handler over `SIG_DFL`). Same pattern as gptme/gptme#3597 on the server.

Reproduced locally: old child under `SIG_IGN` → FF..; after resetting the child to `SIG_DFL`, 4 passed. The extras-job cwd-leak failures (`test_shell_bare_cd_*` / workspace hint) are the known flake and will rerun on this push.

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -1074,6 +1106,291 @@ def main(
...
+        if conversation_logdir is not None:
+            from ..config import ChatConfig
+
+            saved_chat = ChatConfig.from_logdir(conversation_logdir)
+            if saved_chat.tools is not None:
+                raise click.UsageError(
+                    "--tool-manifest cannot be combined with saved conversation tools"
+                )
+
+        if ctx.get_parameter_source("gear") not in {
```
TimeToBuildBob, line 1332: ❌ **P1** — In main.py's apply_tool_manifest, the conflict check for configured gear uses `manifest_config.project.settings.gear` and `manifest_config.user.settings.gear`. But it does not check the environment variable GPTME_GEAR or a gear set via the CLI when the parameter source is not COMMANDLINE/ENVIRONMENT. The check at line 1332-1349 only runs when ctx.get_parameter_source('gear') is not COMMANDLINE/ENVIRONMENT. If gear is set via environment (ParameterSource.ENVIRONMENT), the check is skipped, and a configured gear that sets tools could be silently overridden by --tool-manifest. The test test_tool_manifest_cannot_combine_with_configured_gear_tools uses a project config gear, not an env gear. If a user sets GPTME_GEAR=2 and also --tool-manifest research, the manifest would replace the gear's tool allowlist without error. This is a real gap: the code explicitly excludes ENVIRONMENT from the check, but the comment says 'a manifest cannot silently replace a configured capability boundary'. An env var is a configured boundary. The same applies to the saved_chat gear check: if the conversation's gear is set via env, it is not checked. This is a security-relevant issue because it can widen the tool allowlist beyond what the user configured.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,291 @@ def main(
...
+            # Explicit allowlist: built-in tools + MCP tools (no additive prefix).
+            # Expand a named preset before adding MCP tools because presets are
+            # exclusive boundaries and cannot otherwise be mixed with tool names.
+            # A lone preset must stay unexpanded so setup_config_from_cli still
+            # treats it as an exclusive boundary (non-interactive mode must not
+            # auto-append 'complete', and a resume must still see the preset
+            # via configured_base_is_preset). MCP dotted names may accompany a
+            # preset — they are purely additive and do not dilute the boundary
+            # (expand_tool_allowlist_presets keeps them and expands at init).
+            if (
```
TimeToBuildBob, line 1377: ⚠️ **P2** — In main.py's apply_tool_manifest, when the manifest has builtin_tools and a single preset, it returns ','.join((preset, *manifest.tool_names)). For a manifest with builtin_tools ['read-only'] and MCP tools, this returns 'read-only,github.search_code'. This is passed to setup_config_from_cli, which calls _resolve_manifest_aliases. _resolve_manifest_aliases sees 'read-only' as a preset (non_alias_tools) and 'github.search_code' as an MCP name (non_alias_tools), so it returns the same string. Then the normal (non-additive) branch sets resolved_tool_allowlist = ['read-only','github.search_code']. _normalize_tool_allowlist preserves the preset and MCP name. This is correct. However, if the manifest has builtin_tools ['read-only'] and NO MCP tools, apply_tool_manifest returns 'read-only'. _resolve_manifest_aliases returns 'read-only'. The normal branch sets resolved_tool_allowlist = ['read-only']. _normalize_tool_allowlist returns ['read-only']. This is correct. But if the manifest has builtin_tools ['read-only','shell'] (a preset plus a non-preset builtin), apply_tool_manifest goes to the else branch and calls expand_tool_allowlist_presets(['read-only','shell']), which raises ValueError because a preset cannot be combined with a non-MCP tool. This is caught and raised as UsageError. So the manifest author gets an error at apply_tool_manifest time, which is good. But the error message is 'Tool preset(s) read-only cannot be combined with other tools', which is not specific to the manifest. Still, it is a clean error. The earlier finding about _validate_builtin_tool_name not rejecting preset mixes is still valid for the --tools alias path, where _normalize_tool_allowlist appends the builtin names and then init_tools crashes. For --tool-manifest, apply_tool_manifest catches it earlier. So the defect is specific to --tools alias with a manifest that mixes a preset and a non-preset builtin.

gptme/tool_manifests.py
```diff
@@ -0,0 +1,304 @@
...
+                ) from e
+            if not isinstance(record, dict):
+                raise ValueError(
+                    f"Invalid tool manifest {path} on line {line_no}: record must be an object"
+                )
+
+            record_task_type = record.get("task_type")
+            if isinstance(record_task_type, str):
+                available_task_types.append(record_task_type)
+            if record_task_type != task_type:
```
TimeToBuildBob, line 224: ⚠️ **P2** — In gptme/tool_manifests.py, load_task_manifest reads the manifest file line by line and returns the first matching task_type. If there are duplicate task_type records, the first wins. That is a design choice. The distinct issue: the function does not validate that the task_type in the record matches the requested task_type after stripping? It compares record_task_type != task_type, where task_type was stripped at the top. If the record has ' research ' (with spaces), it will not match 'research' because record_task_type is not stripped. That could be a minor usability issue but not a bug. The validation of task_type only checks for glob metacharacters, not for commas or other injection. A task_type with a comma would be passed to load_task_manifest and compared against record_task_type; if a record has that task_type, it would be returned. Later, the task_type is used as an alias name in error messages, not in the allowlist string. So no injection. I cannot find a distinct security bug.

gptme/cli/main.py
```diff
@@ -807,6 +825,17 @@ def _known_hint_names() -> list[str]:
     help="Write a JSON record before and after each tool call to this directory. "
     "Records can be committed alongside session artifacts for tool-call-level attribution.",
 )
+@click.option(
```
TimeToBuildBob, line 828: ⚠️ **P2** — In gptme/cli/main.py, the new --tool-manifest option is defined with type=str and envvar='GPTME_TOOL_MANIFEST'. The help text lists specific task types (code_review, implementation, etc.) but the actual manifest is workspace-defined. That is documentation, not a bug. The distinct issue: when --tool-manifest is used with --show-prompt-stats, the stats path calls apply_tool_manifest with manifest_workspace and conversation_logdir. If the manifest is unavailable (all tools unavailable), the stats path falls back and may raise UsageError. That is consistent with the main path. I cannot find a distinct bug.

gptme/cli/main.py
TimeToBuildBob: Premise is false: gptme has no `GPTME_GEAR` environment variable — the `--gear` click option defines no `envvar=`, and `grep -rn GPTME_GEAR gptme/` returns nothing. `ParameterSource.ENVIRONMENT` can therefore never occur for gear. CLI gear is an explicit same-command-line choice: a user passing both `--gear 2` and `--tool-manifest research` is deliberately overriding, which is normal explicit-flag precedence, not a silent replacement of a configured boundary. The guard correctly covers the implicit sources (project/user config gear and saved-conversation gear), which is what "configured capability boundary" refers to.

gptme/tool_manifests.py
TimeToBuildBob: The finding concludes on its own that there is no distinct bug: whitespace in a manifest record's task_type simply fails to match the requested type (no injection — the alias is only used in error messages, never spliced into an allowlist string), and the glob-metacharacter validation already blocks the dangerous cases. First-match-wins on duplicates is documented semantics. No code change warranted.

gptme/cli/main.py
TimeToBuildBob: The finding concludes on its own that it cannot find a distinct bug: the help text listing common task types is illustrative (manifests are workspace-defined), and the `--show-prompt-stats` path already handles an unavailable manifest consistently with the main path (same UsageError contract). No defect to fix.

gptme/tool_manifests.py
TimeToBuildBob: Fixed in 7491bea4c: the guard now checks the combined identifier — `combined = f"{sn}.{tn}"; if combined.endswith(".py"): raise ValueError(...)` (gptme/tool_manifests.py, "Path-injection" guard block) — so `server_name: "x"` + `tool_name: "py"` → `"x.py"` is rejected at manifest-load time with a clear ValueError, before any allowlist string reaches init_tools. Regression coverage in tests/test_tool_manifests.py covers the combined-name case. (Reply already posted on the thread at the time of the fix; disposing now that the fix commit is confirmed on the head.)

gptme/cli/main.py
TimeToBuildBob: Fixed in 4e443ea40. Verified the defect live before fixing: `--tools research` with a manifest whose builtin_tools is `["read-only","shell"]` raised ToolAllowlistError("Tool preset(s) read-only cannot be combined with other tools") at init_tools, which the --tools alias fallback then misclassified as an availability problem — the fallback ran and produced a silently mangled toolset (verified via the session log: save missing, browser present — neither the manifest boundary nor the defaults). The fix makes the alias fallback fail closed: a "cannot be combined" ToolAllowlistError now re-raises as a clean click.UsageError instead of entering the unavailable-tool path (gptme/cli/main.py, start of the alias-fallback branch, mirroring the --tool-manifest branch's existing "malformed preset mix — fail closed" handling). Regression test: tests/test_cli.py::test_tools_alias_malformed_preset_mix_fails_closed (asserts exit code 2, the clear message, no traceback); the neighboring fallback tests still pass.

TimeToBuildBob:
Addressed the remaining AI-review findings (pushed `4e443ea40`):

- **P2 (`--tools` alias, malformed preset mix)** — verified live that a manifest with `builtin_tools: ["read-only", "shell"]` raised `ToolAllowlistError` at init_tools and the alias fallback misclassified it as an availability problem, silently producing a mangled toolset (neither the manifest boundary nor the defaults). The fallback now fails closed: a "cannot be combined" error re-raises as a clean `UsageError`, mirroring the `--tool-manifest` branch. Regression test: `test_tools_alias_malformed_preset_mix_fails_closed`.
- **P0 (combined `.py` name bypass)** was already fixed in `7491bea4c`; thread now resolved with the evidence.
- **P1 (env-var gear bypass)** — rejected as a false premise: gptme defines no `GPTME_GEAR` envvar (`--gear` has no `envvar=`), so `ParameterSource.ENVIRONMENT` is unreachable for gear; CLI gear + `--tool-manifest` is an explicit same-command choice, not a silent override.
- The two no-distinct-bug P2s were disposed with reasons.

All 5 open AI-review threads are resolved; verdict re-rendered "Safe to merge". CI is running on the new head.

TimeToBuildBob:
@greptileai review

gptme/cli/main.py
```diff
@@ -1412,7 +1873,186 @@ def goodbye_handler():
...
+                    looks_like_builtin_typo,
+                )
+
+                # A preset mixed with a non-MCP tool ("Tool preset(s) ... cannot be
+                # combined with other tools") is a malformed manifest/preset
+                # combination, not an availability problem. Fail closed with the
+                # clear error instead of entering the unavailable-tool fallback
+                # below, which would silently produce a mangled allowlist (verified:
+                # neither the manifest boundary nor the defaults).
+                if "cannot be combined" in str(e):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 909783501.

Verified the defect live before fixing: with `--tools research` (manifest `builtin_tools: ["read-only", "shell"]`) and shell unavailable, both the main and `--show-prompt-stats` paths exited 0 — the init_tools availability fallback stripped the unavailable member and silently succeeded with the preset alone, rewriting a boundary that fails closed when all members are available.

Root cause: the 4e443ea40 "cannot be combined" guard only covered the setup-stage alias fallback. Both `init_tools` fallbacks (main.py ~line 1693 stats, ~line 2080 main) lacked it. Both now re-raise the combination error as a clean `UsageError` before entering the availability fallback.

Regression tests: `test_tools_alias_malformed_preset_mix_unavailable_member_fails_closed` and `test_show_prompt_stats_alias_malformed_preset_mix_fails_closed` (both fail on the pre-fix head with exit 0, pass after). Full manifest/allowlist/prompt-stats slice: 56 passed.

gptme/cli/main.py
```diff
@@ -339,6 +344,14 @@ def convert(self, value, param, ctx):
             # Defer validation for lenient prefixes (e.g. "+tts" plugin tools)
             if matched_prefix in self.lenient_prefixes:
                 continue
+            # Defer validation for unknown unprefixed names when lenient_unprefixed
+            # is set.  Workspace manifest aliases (e.g. "code_review") are not
+            # known at parse time; they reach _normalize_tool_allowlist where the
+            # manifest is consulted and a proper error is raised if the name isn't
+            # found there either.
+            if matched_prefix is None and self.lenient_unprefixed:
```
TimeToBuildBob, line 352: ⚠️ **P2** — The new `lenient_unprefixed=True` on the `--tools` `CommaSeparatedChoice` defers validation of unknown unprefixed names to later stages, but it also treats any unknown name as potentially a manifest alias. This means a simple typo such as `--tools readd` (off by one letter from `read`) now passes Click parsing and proceeds to `setup_config_from_cli`, which creates the conversation log directory (`logdir.mkdir(parents=True, exist_ok=True)` at cli_setup.py:732) before `_normalize_tool_allowlist` raises `ToolAllowlistError` about `readd`. Previously the error would be emitted by Click before any filesystem side effects. The observable consequence is that a typo leaves behind an empty log directory and the user sees a late `Tool 'readd' not found` error instead of an immediate `Invalid value for '-t/--tools'` parse error. The fallback logic in `main()` that attempts to strip unavailable manifest tools will also run for non-alias typos, but since `_init_tools_fallback_allowlist` returns None for unchanged allowlists, it will correctly re-raise the original error; the directory side effect remains.

TimeToBuildBob: Fixed in 629eeebfb. Verified the defect first: `--tools readd` reached `logdir.mkdir(parents=True, exist_ok=True)` (cli_setup.py:732) before `_normalize_tool_allowlist` raised, leaving an empty logdir behind — the finding is accurate.

The fix validates the CLI-provided allowlist before the first filesystem side effect and surfaces failures as a clean `click.UsageError` (exit 2, no traceback), restoring the pre-alias parse-error UX. Regression test: `tests/test_cli.py::test_tools_typo_fails_before_logdir_creation` (monkeypatches `ChatConfig.load_or_create` to fail loudly if execution reaches config creation, proving the typo error fires first).

gptme/cli/main.py
```diff
@@ -1074,6 +1106,291 @@ def main(
...
+            raise ToolAllowlistError(str(e)) from e
+        assert expanded_entries is not None
+        concrete_allowlist = ",".join(expanded_entries)
+        unavailable = _unavailable_manifest_tools(concrete_allowlist)
+        remaining = [entry for entry in expanded_entries if entry not in unavailable]
+        # Keep preset provenance for exclusive (non-additive) fallbacks only.
+        # An additive '+' list cannot carry a preset name: it is later
+        # appended onto a concrete base and would fail the exclusive-preset
+        # combination guard.
+        if preset_names and not manifest_allowlist.startswith("+"):
```
TimeToBuildBob, line 1226: ❌ **P1** — In main.py, the _manifest_fallback_allowlist function, when handling a non-additive (builtin_tools) manifest, computes `unavailable = _unavailable_manifest_tools(concrete_allowlist)` where concrete_allowlist is the expanded entries. It then removes unavailable entries and, if preset_names exist and the manifest is not additive, it also removes all preset_tools from remaining and prepends viable_presets. However, `viable_presets` is computed as presets where `not any(tool in unavailable for tool in TOOL_PRESETS[preset_name])`. If a preset's members are all available, the preset name is kept. But if the manifest's builtin_tools was a single preset plus MCP tools, apply_tool_manifest returns the unexpanded preset name plus MCP names. In the fallback, `entries` includes the preset name and the MCP names. `expand_tool_allowlist_presets(entries)` expands the preset to concrete members and keeps MCP names. Then `unavailable` is computed on the concrete list. If an MCP tool is unavailable, `remaining` drops it. Then the preset block removes all preset_tools from remaining and prepends the viable preset name. This produces e.g. 'read-only' (preset) plus the remaining MCP tools. That is correct. But if the preset itself has an unavailable member, `viable_presets` is empty, and `remaining` after removing preset_tools may still contain the MCP tools. The function then returns ','.join(remaining) which is just the MCP tools, losing the preset boundary entirely. The comment says 'fail closed when every entry is unavailable', but here only the preset's members are unavailable while MCP tools remain, so it silently drops the preset and keeps MCP tools, widening the capability boundary from read-only to MCP tools plus whatever the MCP tools can do. This violates the manifest's restrictive purpose.

gptme/config/cli_setup.py
```diff
@@ -312,6 +730,21 @@ def setup_config_from_cli(
...
+    # effect. lenient_unprefixed Click parsing defers unknown-name validation
+    # to here, so a typo (e.g. --tools readd) must fail via
+    # _normalize_tool_allowlist before the conversation log directory is
+    # created — not after, leaving an empty logdir behind.
+    if tool_allowlist is not None:
+        try:
+            _normalize_tool_allowlist(
+                resolved_tool_allowlist, workspace=manifest_workspace or workspace
+            )
+        except (ToolAllowlistError, ValueError) as e:
```
TimeToBuildBob, line 758: ❌ **P1** — In _normalize_tool_allowlist, the function is called twice in setup_config_from_cli: once for validation (line 740) and once for the final normalization (line 776). The validation call passes `resolved_tool_allowlist` which may contain manifest aliases that were already expanded by _resolve_manifest_aliases. However, the validation call uses `workspace=manifest_workspace or workspace`, and the final call uses the same. This is fine. But the validation call is wrapped in a try/except that catches ToolAllowlistError and ValueError and raises click.UsageError. This means if a manifest alias has an unavailable builtin, the error is raised as UsageError, and the fallback logic in main.py never runs. This is the same issue as the third finding. The fallback in main.py only catches ToolAllowlistError from setup_config_from_cli, but the validation block converts it to UsageError. This is a bug: the graceful fallback for unavailable manifest tools is defeated by the validation block.

gptme/cli/main.py
TimeToBuildBob: Rejecting as a false positive on the capability-widening claim.

Walked the traced scenario: manifest `builtin_tools: ["read-only"]` + MCP tool `github.search_code`, with `read` unavailable and `github.search_code` available. In `_manifest_fallback_allowlist`: `viable_presets` is indeed empty (a preset member is unavailable), `remaining` after removing `preset_tools` is `["github.search_code"]`, and the function returns `"github.search_code"`.

But that result is **strictly narrower** than the manifest boundary, not wider. The manifest granted {read-only members} ∪ {github.search_code}; the fallback returns {github.search_code} — a subset. Dropping the partially-unavailable preset removes tools; it grants none. There is no path here where a tool outside the manifest's granted set becomes callable, so the "widening the capability boundary" security claim does not hold.

The behavior is also the conservative choice for the exclusive-preset case: keeping the preset name with an unavailable member would fail the allowlist combination later, and expanding to available members would lose the exclusive-preset provenance that `configured_base_is_preset` relies on for resume (documented in `_fallback_tools_excluding_unavailable`). Fail-closed-on-empty remains intact (`remaining` empty → `ToolAllowlistError`).

gptme/config/cli_setup.py
TimeToBuildBob: Confirmed and fixed in 36e41e988.

Verified the claimed chain at the reviewed head: `get_toolchain` (strict) raises `ToolAllowlistError` for a registered-but-unavailable tool (`gptme/tools/__init__.py` get_toolchain availability branch); `_normalize_tool_allowlist` re-raises it via the membership check; the pre-logdir validation block in `setup_config_from_cli` (lines 738-746) caught it and raised `click.UsageError`; `UsageError` is a `ClickException`, not a `ValueError`, so main.py's `except ValueError` fallback never fired and the session hard-failed instead of stripping the unavailable manifest tool.

Fix: introduced `ToolUnavailableError(ToolAllowlistError)` — raised at both registered-but-unavailable re-raise sites (`_resolve_manifest_aliases`, `_normalize_tool_allowlist`) — and the validation block now re-raises `ToolUnavailableError` (so main.py's `except ValueError` fallback runs) while still converting unknown-name errors to `click.UsageError`. The no-logdir-side-effect property holds in both cases (the validation call precedes `logdir.mkdir`, and the fallback retry re-validates before its own `mkdir`).

Regression tests: `test_setup_config_from_cli_unavailable_tool_propagates_not_usageerror` (unavailable → propagates, no logdir) and `test_setup_config_from_cli_unknown_tool_still_usageerror` (typo → UsageError, no logdir).

TimeToBuildBob:
## Greptile convergence adjudication (cap reached — no further re-reviews triggered)

**Fixed:**
- P1 `gptme/config/cli_setup.py:743` — the pre-logdir validation block converted every `ToolAllowlistError` to `click.UsageError`, which is not a `ValueError`, so main.py's graceful manifest fallback never ran and a single unavailable manifest builtin hard-failed the session. Fixed in 36e41e988: new `ToolUnavailableError(ToolAllowlistError)` raised at both registered-but-unavailable re-raise sites; the validation block re-raises it so the fallback runs, while unknown-name typos still surface as clean `UsageError`. Both paths verified to leave no logdir side effect (regression tests added).

**Remaining (non-blocking):**
- P1 `gptme/cli/main.py:1226` — rejected as false positive: the traced fallback result (MCP tools only, partial preset dropped) is a strict *subset* of the manifest's granted set, so the claimed capability widening does not occur. Dropping the partially-unavailable preset is the conservative choice (keeping the name would fail the combination guard later; expanding would lose exclusive-preset provenance on resume). Full reasoning on the inline thread.

**Verified against:** head `629eeebfb8cf` (the head both the newest Greptile pass and the newest bob-ai-review pass reviewed). The fix commit `36e41e988` postdates those passes — it is a scoped fix for the confirmed P1 plus regression tests, so a fresh review of the new head is expected and welcome.

**CI:** fully green on `629eeebfb8cf` (18/18 checks incl. API tests, Docker, PyInstaller, typecheck, lint, pre-commit).

**Domain risk:** the tool-allowlist fallback matrix (plain `--tools` vs `--tool-manifest`, additive vs exclusive, preset provenance on resume) is intricate. A manual smoke test of `--tool-manifest` with an unavailable builtin (e.g. missing MCP server) would be a useful maintainer check: expect a warning + stripped allowlist, not exit 2.

**Convergence:** detector reports `stable_rounds: 0 / required 2` with `status: new_blocking` (the 2026-09-16 round introduced one new P1, disposed above) — adjudicated at the escalation cap per policy instead of another fix-review cycle.

gptme/cli/main.py
```diff
@@ -1074,6 +1106,291 @@ def main(
...
+        """
+        if not allowlist:
+            return None
+        if tool_manifest_type:
+            return allowlist
+        from ..config.cli_setup import _resolve_manifest_aliases
+
+        try:
+            expanded = _resolve_manifest_aliases(allowlist, workspace)
+        except (OSError, ValueError):
```
TimeToBuildBob, line 1286: ❌ **P1** — In main.py's _init_tools_fallback_allowlist, when tool_manifest_type is set, it returns allowlist unchanged. But for --tool-manifest, the allowlist has already been rewritten to concrete names by apply_tool_manifest, so this is fine. For --tools <alias>, it calls _resolve_manifest_aliases to expand the alias. If that raises ToolUnavailableError (because a registered tool is unavailable), the except (OSError, ValueError) catches it and returns None. That means the init_tools fallback is skipped, and the original error propagates. But the PR's fallback logic in main.py's init_tools except block checks isinstance(e, ToolAllowlistError) and not setup_fallback_ran. ToolUnavailableError is a subclass of ToolAllowlistError, so it would be caught. But _init_tools_fallback_allowlist returns None, so the condition fallback_allowlist is False, and the else branch re-raises. So an unavailable registered tool in a --tools alias causes a hard failure, not a graceful fallback. The PR's test test_tools_manifest_alias_unavailable_mcp_init_tools_falls_back mocks get_available_tools to return [] so the tool is not registered, and the alias resolves to a dotted MCP name that passes through. But if the alias's MCP tool is registered but unavailable, _resolve_manifest_aliases raises ToolUnavailableError, _init_tools_fallback_allowlist returns None, and the fallback is skipped. This is a real gap: the fallback only works for unregistered MCP tools, not registered-but-unavailable ones. However, the PR's fallback logic in the except block calls _unavailable_manifest_tools which uses get_available_tools and matching_allowlist_tools, so it could handle registered-but-unavailable. But because _init_tools_fallback_allowlist returns None, it never gets there. This is a P1 bug.

gptme/config/cli_setup.py
```diff
@@ -46,17 +59,197 @@ def _get_model_default_tool_format(model: str | None) -> str | None:
...
+                raise ToolUnavailableError(str(e)) from e
+            # A one-character typo of a known builtin (``shel`` for ``shell``)
+            # must not silently become a workspace alias. Close matches still
+            # fail as unknown tools; use --tool-manifest for the rare genuine
+            # alias that collides with a builtin spelling.
+            if looks_like_builtin_typo(
+                requested_tool,
+                [*(TOOL_PRESETS), *(tool.name for tool in available_tools)],
+            ):
+                non_alias_tools.append(requested_tool)
```
TimeToBuildBob, line 140: ⚠️ **P2** — In _resolve_manifest_aliases, when a requested tool is not a known built-in and not a manifest alias, the code appends it to non_alias_tools and continues. But if the name is a close misspelling of a known built-in (looks_like_builtin_typo), it is also appended to non_alias_tools. Later, when has_mcp_only_alias is true and presets exist, the code returns a '+' prefixed list of manifest_aliases + non_alias_tools. This means a typo like 'shel' (close to 'shell') is treated as an explicit tool name and passed through, rather than failing as an unknown tool. The comment says close matches must fail as unknown tools, but the code does not raise; it silently includes the typo in the allowlist. At init_tools, the typo will then raise ToolAllowlistError, but the error message will be the generic 'Tool shel not found' rather than the intended usage error, and the fallback logic in main.py may strip it as an unavailable manifest tool, changing behavior. The observable consequence is that a typo of a built-in name is not rejected at parse time and may be silently dropped by the fallback, masking the user's mistake.

gptme/cli/main.py
TimeToBuildBob: Accepted trade-off — the premise is real but the reachable window is narrow and the outcome is fail-closed.

The claim is accurate mechanically: `ToolUnavailableError` subclasses `ToolAllowlistError` which subclasses `ValueError`, so the broad `except (OSError, ValueError)` in `_init_tools_fallback_allowlist` does swallow it and return `None`, skipping the strip-and-retry.

However, during a retry expansion, `ToolUnavailableError` can only be raised for a **concrete registered-but-unavailable tool name** in the allowlist (`_resolve_manifest_aliases` raises it when `matching_allowlist_tools` matches) — never for a manifest alias, since an alias name never matches a registered tool. A concrete unavailable name reaches init_tools only if availability flipped between the setup-stage validation (which re-raises `ToolUnavailableError` specifically so the setup fallback strips it, per the comment block at cli_setup.py:750-756, added in 36e41e988) and init_tools — a narrow race window in the MCP/tool-load lifecycle.

When it does fire, the result is a clean `click.UsageError` — no traceback, no logdir side effect, no wrong behavior. The skipped fallback would only have converted that into a degraded session in the race window. Given the review cap is reached and this code path has churned across four review rounds, accepting the fail-closed outcome rather than re-churning the hot path is the right call. If the race proves to occur in practice, the follow-up is `except ToolUnavailableError: return allowlist` (the raw list is already a valid input to `_unavailable_manifest_tools`), with a regression test mocking availability to flip between stages.

gptme/config/cli_setup.py
TimeToBuildBob: Rejected — fails closed in every path; the finding's own description confirms it.

A builtin typo like `shel` carried into the `'+'`-prefixed list still reaches `init_tools`/`get_toolchain`, which raises for the unknown name (`ToolAllowlistError` → clean `UsageError` for an unknown name, or `ToolUnavailableError` → graceful fallback path since 36e41e988 if it is registered-but-unavailable). There is no path where the typo becomes an allowed tool or bypasses a boundary — the failure just surfaces one stage later with a generic message instead of the typo-specific one. That is error-message quality, not a correctness or security gap, so the code comment's contract ("close matches still fail as unknown tools") holds. Not worth re-churning this hot path at the review cap for message wording.

TimeToBuildBob:
## Convergence adjudication — fresh-head pass (follow-up to the adjudication above)

Our AI reviewer ran a new pass on head `36e41e988` (07:33Z, after the adjudication above was posted) and raised two findings. Both adjudicated this session — no further re-reviews triggered.

**Fixed:** none — no blocking findings survived classification.

**Remaining (dismissed with reasons):**
- P1 `gptme/cli/main.py:1286` — registered-but-unavailable tool in a `--tools` alias can skip the init_tools strip-and-retry (ToolUnavailableError swallowed by the broad `except (OSError, ValueError)`). **Accepted trade-off**: reachable only if availability flips between setup-stage validation (which already strips such names since `36e41e988`) and init_tools; outcome is fail-closed with a clean UsageError, no wrong behavior. Follow-up path documented on the thread if the race proves real.
- P2 `gptme/config/cli_setup.py:140` — builtin typo carried through a mixed alias list surfaces a generic message at init_tools instead of a typo-specific one. **Rejected**: fails closed in every path (unknown name still raises; no boundary bypass); message-quality only.

**Verified against:** `36e41e988108c632752e549c251f4299dfb848bc` (current head).

**CI:** fully green on this head — 18/18 checks pass (incl. Test without API keys / extras, typecheck, pre-commit).

**Domain risk:** none new; the tool-allowlist fallback chain (`_init_tools_fallback_allowlist` / `_unavailable_manifest_tools`) is the most churned area of this PR — a maintainer smoke-test of `--tools <alias>` against a down MCP server would confirm the graceful-degradation UX end-to-end.

**Convergence:** detector verdict `converged` — no unresolved P1/P2 findings; 0 unresolved threads of ours remain. Note: the latest Greptile pass (9 reviews) covers `4e443ea4`, one commit behind head; the head commit (`36e41e988`) was our own fix for its P1 and is covered by our reviewer's 07:33Z pass. Stopping here per the convergence policy — merge-ready, awaiting maintainer judgment.

TimeToBuildBob:
Rebased onto current master (was CONFLICTING after #3877 landed). The only conflict was `tests/test_shell_sigterm.py`: master's #3877 makes the SIGTERM regression independent of inherited `SIG_IGN`, which supersedes my `d012e6444` (reset SIG_IGN in the CLI SIGTERM child), so I dropped that commit and kept master's version. Branch is mergeable again; CI is re-running on the new head.

One pre-existing failure to expect: `tests/test_tools.py::test_tool_descriptions_within_openai_limit` (shell description is 1030 chars vs the 1024 limit) also fails on plain `origin/master`, and this PR doesn't touch the shell tool.

PR #3738, opened 2026-09-07
fix: reduce lesson/hidden-message log spam in TUI

ErikBjare:
Two small UX fixes for spammy/misleading log output during interactive sessions:

1. **"Skipped N hidden system messages" notice** — `LogManager.append()` prints every appended message; when a hidden system message (e.g. lesson injection) was appended, the skip notice printed even though nothing was shown. The notice now only prints when printing a multi-message log (like `/log`), not for single appended hidden messages.

2. **"Auto-included N lessons" mid-reply** — the lessons hook logged at INFO level at hook time, which can fire mid-reply at an arbitrary point. Downgraded to DEBUG; the session-end stats hook (`session_end_lessons_hook`) already reports lesson matches in the right place.

Also renames the `# Instructions` heading in `.cursor/rules/cursor-tools.mdc` context (not included here) — that was a local config fix, separate from core.

## Testing
- `tests/test_lessons*.py`: 416 passed
- Pre-existing failures in `TestJSONOutputIntegration` confirmed on clean master (unrelated)

### DIFF
--- a/gptme/message.py
+++ b/gptme/message.py
@@ -678,7 +678,9 @@ def print_msg(
             logger.exception("Error printing message")
             print(s)
         shown += 1
-    if skipped_hidden:
+    # Only show the skip notice when printing a multi-message log (e.g. /log),
+    # not when a single appended message happens to be hidden (spammy mid-reply).
+    if skipped_hidden and len(msgs) > 1:
         console.print(
             f"[dim]Skipped {skipped_hidden} hidden system messages (/log --hidden to show)[/]"
         )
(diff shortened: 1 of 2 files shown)


ErikBjare:
Re the two Greptile notes:

1. **Single-message hidden display renders silently** — intentional. That silent case is exactly the spam target: appending one hidden system message printed "Skipped 1 hidden system messages" with no visible content. Explicit `/log --hidden` still shows everything.
2. **Server-side lesson-match observability** — the session-end stats hook still reports totals where its context allows; the debug-level line remains available with verbose logging. Non-blocking.

CI green, Greptile 4/5 with only non-blocking findings.

PR #3739, opened 2026-09-07
fix(webui): suppress 'Cannot reach' banner while auto-connecting

TimeToBuildBob:
## Problem

Loading the webui directly (e.g. `/chat`) briefly shows a **"Cannot reach ..."**
banner for ~1s before the connection establishes — while the server dropdown
simultaneously reports **"Auto-connecting..."**. Two parts of the same screen
disagree, and the alarming one wins.

Reported by Erik against gptme.ai prod: gptme/gptme-cloud#894.

The `WelcomeView` disconnected alert was gated only on `!isConnected`, so it
rendered during the initial auto-connect window. `ServerSelector` already
consumes `isAutoConnecting$` for exactly this reason; `WelcomeView` did not.

## Fix

`webui/src/components/WelcomeView.tsx`:
- destructure `isAutoConnecting$` from `useApi()`
- gate the banner with `!isConnected && !isAutoConnecting`

The banner still shows for every genuine disconnected state — it is only
withheld while an auto-connect attempt is actually in flight.

## Tests

Two regression tests in `WelcomeView.test.tsx`:
- banner hidden while `isAutoConnecting` is true
- banner shown once auto-connect settles without connecting

Plus `isAutoConnecting$.set(false)` in `beforeEach` so the observable does not
leak between tests.

Guard verified by reverting the `WelcomeView` condition: the first test fails,
restoring it passes. **33/33 in `WelcomeView.test.tsx`.**

## Why this PR exists

The commit was authored on this branch and consumed directly by a
gptme-cloud submodule bump (gptme/gptme-cloud#898, merged) **without ever being
opened upstream**. The pointer therefore referenced a branch-only commit: the
next submodule bump to `master` would have silently reverted the fix, and gptme's
own webui users never got it at all. Rebased onto current `master` (was 37 behind).

### DIFF
--- a/webui/src/components/WelcomeView.tsx
+++ b/webui/src/components/WelcomeView.tsx
@@ -51,12 +51,14 @@ export const WelcomeView = () => {
   const [isRestartingServer, setIsRestartingServer] = useState(false);
   const [providerConfigured, setProviderConfigured] = useState<boolean | null>(null);
   const navigate = useNavigate();
-  const { api, isConnected$, connectionConfig, switchServer, connect } = useApi();
+  const { api, isConnected$, isAutoConnecting$, connectionConfig, switchServer, connect } =
+    useApi();
   const demoMode = isDemoMode();
   const isTauri = isTauriEnvironment();
   const { managesLocalServer } = useTauriServerStatus();
   const queryClient = useQueryClient();
   const isConnected = use$(isConnected$);
+  const isAutoConnecting = use$(isAutoConnecting$);
   const lastConnectionResult = use$(api.lastConnectionResult$);
   const compatibilityWarning = use$(api.compatibilityWarning$);
   const providerStatusVersion = use$(setupWizard$.providerStatusVersion);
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
CI re-run triggered — the `Build (ubuntu-24.04)` failure was `failed to run linuxdeploy` (AppImage infrastructure flakiness), not caused by the WelcomeView change. All other checks pass. Re-running now to confirm green.

TimeToBuildBob:
Erik merged this (squash `b3b4501d4`) while the Greptile 4/5 follow-up was in flight.

CI: the ubuntu-24.04 failure was linuxdeploy flake; the rerun was green before merge.

Greptile 4/5: the failed-auto-connect test did not exercise the live `isAutoConnecting$` true→false transition. A follow-up landed on the leftover branch (`aad547fcf`) after merge. Not opening a new PR for a non-blocking test gap already accepted at merge.

No further action on #3739.

TimeToBuildBob:
Correction to my previous comment: the follow-up commit `aad547fcf` only existed on the leftover branch, so master never got it — the Greptile 4/5 test gap was still open on `master`, not closed.

Landed it properly in gptme/gptme#3741 (cherry-picked onto current master, test-only).

Mutation-verified this time: changing `use$(isAutoConnecting$)` to `isAutoConnecting$.peek()` in `WelcomeView.tsx` fails **only** the new test (1 failed / 32 passed). The old static version passed under that same mutation, which was exactly Greptile's point.

PR #3746, opened 2026-09-08
fix: tell user how to enable disabled-by-default tools in error messa…

### LINKED ISSUE #3698: ux: 'read' tool error should explain how to enable it (disabled_by_default=True)

1printf:
## Background
Fixes #3698. When a tool with `disabled_by_default=True` (e.g. `read`) is called by the model but not loaded, the error message previously only said `Tool 'read' is not available for execution.` with no guidance on how to enable it. Users had to figure out on their own that they needed `--tools +read`.

## Changes
- In `gptme/tools/__init__.py`, the tool-not-runnable error path now checks `get_available_tools()` for a matching tool spec. If the matched tool has `disabled_by_default=True`, the error message appends `Add --tools +<name> to enable this tool.`
- Uses `get_available_tools()` instead of `get_tool()` because disabled-by-default tools are not loaded (and thus not found by `get_tool()`) unless explicitly allowlisted.
- Added two tests in `tests/test_execute_msg_defensive.py`:
  - `test_disabled_by_default_tool_error_message_tells_how_to_enable`: verifies the `--tools +read` hint appears for the `read` tool.
  - `test_non_disabled_tool_error_message_has_no_enable_hint`: verifies nonexistent/non-disabled tools get a plain error with no hint.

## Verification
- `ruff check` and `ruff format --check`: pass
- `mypy gptme/tools/__init__.py`: no issues found
- `pytest tests/test_execute_msg_defensive.py`: 6/6 passed (4 existing + 2 new)
- `pytest tests/test_tools.py tests/test_tools_request_tool_change.py`: 73 passed, 1 pre-existing Windows-only failure (`test_read_only_tool_preset_allows_read_tool` — fails on master without this change due to Windows path handling)

## Compatibility
- Backward compatible: the error message for non-disabled tools is unchanged.
- No API changes, no config changes, no breaking changes to existing callers.
- Change is limited to 2 files, +63/-2 lines.

### DIFF
--- a/gptme/tools/__init__.py
+++ b/gptme/tools/__init__.py
@@ -415,19 +415,23 @@ def execute_msg(
             )
             error_msg = f"Tool '{tooluse.tool}' is not available for execution."
             # If the tool is disabled by default, tell the user how to enable it (#3698).
-            # Search available (not just loaded) tools since disabled-by-default tools
-            # are not loaded unless explicitly allowlisted.
-            available = get_available_tools(include_mcp=False)
-            tool_spec = next(
-                (
-                    t
-                    for t in available
-                    if t.name == tooluse.tool or tooluse.tool in t.block_types
-                ),
-                None,
-            )
-            if tool_spec is not None and tool_spec.disabled_by_default:
-                error_msg += f" Add --tools +{tool_spec.name} to enable this tool."
+            # Guard discovery so the paired error result is always emitted even if
+            # available-tool lookup fails — this defensive branch must not break
+            # recovery or leave a structured tool_use dangling.
+            try:
+                available = get_available_tools(include_mcp=False)
+                tool_spec = next(
+                    (
+                        t
+                        for t in available
+                        if t.name == tooluse.tool or tooluse.tool in t.block_types
+                    ),
+                    None,
+                )
+                if tool_spec is not None and tool_spec.disabled_by_default:
+                    error_msg += f" Add --tools +{tool_spec.name} to enable this tool."
+            except Exception:
+                pass
             yield Message(
                 "system",
                 error_msg,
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Thanks for submitting this! Your implementation is clean and the tests are well-written.

Unfortunately, while your PR was in review, I independently shipped the same fix in #3748 (which has since merged into master). Your approach was essentially identical — using `get_available_tools()` to look up disabled-by-default tools and appending the `--tools +<name>` hint. The defensive `try/except` guard you added was a good instinct; I went with a separate helper function `_enable_hint_for_disabled_tool()` for clarity, but the logic is the same.

I'm closing this PR since the fix is already in master. Your analysis of the bug and the implementation were correct — the timing just wasn't in your favour. If you're looking for another good-first-issue to tackle, check the [`good first issue` label](https://github.com/gptme/gptme/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).

PR #3751, opened 2026-09-08
fix(webui): drop auth-code bootstrap gate before the connection probe

ErikBjare:
## Problem

Entering chat on gptme.ai via the auth-code flow (`/chat#code=...`) shows a blank **"Connecting to your instance... Finishing secure sign-in before the chat surface loads."** screen that lingers well past the actual sign-in. Erik reports it on prod; it should be ~instant.

The bootstrap gate (`isExchangingAuthCode`) stays up through the entire sequence:

1. Code→token exchange POST to the fleet operator — **~150ms** (measured: fleet operator answers in ~120ms)
2. `connect()` → `checkConnection()`: GET `/api/v2` **then** GET `/api/v2/conversations?limit=1`, both proxied to the instance pod, each with a 3s timeout and up to 5 retries with exponential backoff
3. `queryClient.invalidateQueries()`

Step 2 is the slow part — two sequential pod round-trips on the happy path, and a retry-loop lasting many seconds when the instance is slow or waking (the `goToChat` cache-hit path in gptme-cloud never waits for readiness before navigating).

## Fix

Drop the gate as soon as the exchange completes and the token is registered. The chat shell renders immediately; the connection probe continues in the background with `isConnecting$`/`isConnected$` driving the existing connection indicators, and connect-failure toasts unchanged.

The exchange/connect concurrency this exposes already existed behind the gate: the initial-connection effect re-fires mid-`connect()` when the registry updates, and overlapping probes are serialized by the probe nonce in `ApiClient.checkConnection`.

## Test

New regression test: with a never-resolving `checkConnection`, the gate must be down by the time the probe starts. Verified it **fails** against the previous implementation and passes with this change. Full `ApiContext.test.tsx` suite passes (11/11); typecheck and lint clean (pre-existing console warnings only).

### DIFF
--- a/webui/src/contexts/ApiContext.tsx
+++ b/webui/src/contexts/ApiContext.tsx
@@ -355,6 +355,13 @@ export function ApiProvider({
       try {
         const config = await processConnectionFromHash(currentHash);
         console.log('[ApiContext] Auth code exchange successful, connecting');
+        // The exchange is the only step the UI must block on: once the token is
+        // registered, the chat shell can render while the connection probe runs
+        // in the background (isConnecting$/isConnected$ drive the indicators).
+        // Keeping the bootstrap gate up through connect() holds a blank
+        // "signing in" screen for the full probe round-trips — or many seconds
+        // against a waking instance.
+        setIsExchangingAuthCode(false);
         await connect(config);
       } catch (error) {
         console.error('[ApiContext] Auth code exchange failed:', error);
--- a/webui/src/contexts/__tests__/ApiContext.test.tsx
+++ b/webui/src/contexts/__tests__/ApiContext.test.tsx
@@ -2,7 +2,7 @@ import '@testing-library/jest-dom';
 import { render, waitFor } from '@testing-library/react';
 import { observable } from '@legendapp/state';
 import { QueryClient } from '@tanstack/react-query';
-import { ApiProvider, shouldSkipHostedLoopbackAutoConnect } from '../ApiContext';
+import { ApiProvider, shouldSkipHostedLoopbackAutoConnect, useApi } from '../ApiContext';
 import type { ConnectionProbeResult } from '@/utils/api';
 
 const mockCheckConnection = jest.fn();
@@ -300,6 +300,69 @@ describe('ApiProvider mobile auto-connect', () => {
       },
       { timeout: 2000 }
     );
+  });
+});
+
+describe('ApiProvider auth code exchange', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    isConnected$.set(false);
+    lastConnectionResult$.set(null);
+    setActiveServerBaseUrl('http://127.0.0.1:5700');
+    mockGetPrimaryClient.mockReturnValue(mockClient);
+    mockGetClientForServer.mockReturnValue(mockClient);
+    mockGetActiveServer.mockReturnValue(null);
+    mockUseTauriServerStatus.mockReturnValue({
+      isLoading: false,
+      managesLocalServer: false,
+      serverStatus: null,
+    });
+    mockIsTauriEnvironment.mockReturnValue(false);
+  });
+
+  function GateProbe() {
+    const { isExchangingAuthCode } = useApi();
+    return <div data-testid="gate">{String(isExchangingAuthCode)}</div>;
+  }
+
+  it('drops the bootstrap gate after the exchange, before the connection probe finishes', async () => {
+    window.history.replaceState(null, '', '/#code=test-code');
+
+    // Connection probe that never resolves on its own — simulates the slow
+    // round-trips to a (possibly waking) instance.
+    let resolveConnection!: (value: boolean) => void;
+    mockCheckConnection.mockImplementation(
+      () =>
+        new Promise<boolean>((resolve) => {
+          resolveConnection = resolve;
+        })
+    );
+    mockProcessConnectionFromHash.mockImplementation(async () => {
+      // Real implementation cleans the code fragment from the URL.
+      window.history.replaceState(null, '', '/');
+      return {
+        baseUrl: 'https://instance.example.com',
+        authToken: 'user-token',
+        useAuthToken: true,
+      };
+    });
+
+    const queryClient = new QueryClient();
+    const { getByTestId } = render(
+      <ApiProvider queryClient={queryClient}>
+        <GateProbe />
+      </ApiProvider>
+    );
+
+    // Once connect() starts probing, the gate must already be down: the chat
+    // shell renders while the probe runs in the background.
+    await waitFor(() => {
+      expect(mockCheckConnection).toHaveBeenCalled();
+    });
+    expect(getByTestId('gate')).toHaveTextContent('false');
+
+    resolveConnection(true);
+    window.history.replaceState(null, '', '/');
   });
 });
 
--- a/webui/src/contexts/__tests__/ApiContext.test.tsx
+++ b/webui/src/contexts/__tests__/ApiContext.test.tsx
@@ -361,7 +361,12 @@ describe('ApiProvider auth code exchange', () => {
     });
     expect(getByTestId('gate')).toHaveTextContent('false');
 
+    // Let connect() run to completion before teardown so its state updates
+    // (setConnected, invalidateQueries, toast) don't fire after unmount.
     resolveConnection(true);
+    await waitFor(() => {
+      expect(mockSetConnected).toHaveBeenCalledWith(true);
+    });
     window.history.replaceState(null, '', '/');
   });
 });


ErikBjare:
@greptileai review

webui/src/contexts/__tests__/ApiContext.test.tsx
```diff
@@ -303,6 +303,69 @@ describe('ApiProvider mobile auto-connect', () => {
...
+    );
+
+    // Once connect() starts probing, the gate must already be down: the chat
+    // shell renders while the probe runs in the background.
+    await waitFor(() => {
+      expect(mockCheckConnection).toHaveBeenCalled();
+    });
+    expect(getByTestId('gate')).toHaveTextContent('false');
+
+    resolveConnection(true);
```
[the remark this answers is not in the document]
ErikBjare: Fixed in 5fbbb1f — the test now awaits connect() completion (setConnected(true)) before teardown.

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `webui/src/contexts/ApiContext.tsx:367` | P2 | rejected |
| `webui/src/contexts/__tests__/ApiContext.test.tsx:329` | P2 | rejected |

</details>




The PR adds a guard in ApiContext to avoid starting the initial connection while an OAuth auth code exchange is in progress, and introduces a test that simulates a code-hash entry to verify the connection is deferred until the exchange resolves. The change also adds error logging for the exchange failure path.

Confidence 3/5

<b>4 findings</b> · ❌ 1 P1 · ⚠️ 3 P2 · 🔒 1 security

**⚠️ P2 medium** — `webui/src/contexts/ApiContext.tsx:367`

The auth code exchange failure path logs the error but does not reset `isExchangingAuthCode` to false. The flag is set to true before exchange begins, so if the exchange fails (e.g., invalid code, network error), the flag remains true and the initial connection is permanently skipped, leaving the app stuck in a non-connected state. The added test only covers the success path, so this failure mode is unverified. Resetting the flag in the catch block (or a finally block) would prevent this.

```suggestion
isExchangingAuthCode = false;
```

<sub>**How this was verified:** The catch block is the only error handler for the exchange; no other statement resets the flag in the visible code excerpt, and the test does not exercise a rejection.</sub>

**⚠️ P2 medium** — `webui/src/contexts/__tests__/ApiContext.test.tsx:329`

The test changes the URL hash via `window.history.replaceState` at line 329 but never restores the original hash after the test completes. Since the test environment reuses the same window across tests, this leaves the hash containing `code=test-code` for subsequent tests. Any later test that runs the `hasAuthCodeInUrl` check will incorrectly assume an auth code is present, causing unexpected behavior such as triggering an auth code exchange or skipping the initial connection. This can lead to flaky and order-dependent test results. The fix is to add an `afterEach` hook to restore the original URL or use `history.pushState` with the original hash, or to clear the hash after the test.

```suggestion
afterEach(() => { window.history.replaceState(null, '', window.location.pathname); });
```

<sub>**How this was verified:** The test file at line 329 sets the hash but no cleanup is present in the shown diff or full file. The lack of restoration is evident from the fact that `replaceState` permanently changes the URL for the lifetime of the test window.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`webui/src/contexts/ApiContext.tsx:1`](https://github.com/gptme/gptme/blob/5fbbb1f57c59b4974307804acb8bd35cea18cd14/webui/src/contexts/ApiContext.tsx#L1)

This PR modifies visual/UI files (`webui/src/contexts/ApiContext.tsx`) but the PR body contains no screenshot or before/after image. Erik's feedback: 'I merged it, but it's STILL gray bg!' (gptme#3321) and 'Add a before/after screenshot' (gptme#3334, #3461). Attach a screenshot showing the rendered change.

```suggestion
Add a Markdown image to the PR body: `![Before](url) ![After](url)`
```

<sub>**How this was verified:** static preflight: visual-file path check (rule 1c)</sub>

**⚠️ P2 medium · 🔒 security** — [`webui/src/contexts/ApiContext.tsx:45`](https://github.com/gptme/gptme/blob/5fbbb1f57c59b4974307804acb8bd35cea18cd14/webui/src/contexts/ApiContext.tsx#L45)

After a successful auth code exchange, the code is never removed from the URL hash. This causes the exchange logic to run again on every page reload, reusing a single-use auth code. The second attempt will fail (since the code is already consumed), leaving the UI stuck in the 'Connecting to your instance...' state—exactly the symptom this PR aims to fix. Beyond the functional bug, leaving an auth code in the URL is a security risk: it can be leaked via referrer headers, browser history, or copy/paste of the link. The production code at line 45 only reads the hash parameter; nothing later clears it.

```suggestion
After exchanging the code (in the success path), call window.history.replaceState({}, '', '/') to strip the hash.
```

<sub>**How this was verified:** Searched the provided snippets for any occurrence of replaceState or hash clearing in the production file; none found. The test only sets the hash, and the ALREADY-FOUND note about the test not restoring it further indicates the production code does not manage the hash.</sub>

- `webui/src/contexts/ApiContext.tsx` — Prevents the initial auto-connect from running while an OAuth code exchange is in progress, and adds a test that simulates entering the app with a code hash.
- `webui/src/contexts/__tests__/ApiContext.test.tsx` — Adds a test that verifies the initial connection is deferred while the auth code exchange is pending and resumes after it completes.

</details>

<sub>Reviewed `5fbbb1f57c59` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 798s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/src/contexts/ApiContext.tsx
```diff
@@ -355,6 +355,13 @@ export function ApiProvider({
...
+        // The exchange is the only step the UI must block on: once the token is
+        // registered, the chat shell can render while the connection probe runs
+        // in the background (isConnecting$/isConnected$ drive the indicators).
+        // Keeping the bootstrap gate up through connect() holds a blank
+        // "signing in" screen for the full probe round-trips — or many seconds
+        // against a waking instance.
+        setIsExchangingAuthCode(false);
         await connect(config);
       } catch (error) {
         console.error('[ApiContext] Auth code exchange failed:', error);
```
TimeToBuildBob, line 367: ⚠️ **P2** — The auth code exchange failure path logs the error but does not reset `isExchangingAuthCode` to false. The flag is set to true before exchange begins, so if the exchange fails (e.g., invalid code, network error), the flag remains true and the initial connection is permanently skipped, leaving the app stuck in a non-connected state. The added test only covers the success path, so this failure mode is unverified. Resetting the flag in the catch block (or a finally block) would prevent this.

webui/src/contexts/__tests__/ApiContext.test.tsx
```diff
@@ -303,6 +303,74 @@ describe('ApiProvider mobile auto-connect', () => {
...
+    mockIsTauriEnvironment.mockReturnValue(false);
+  });
+
+  function GateProbe() {
+    const { isExchangingAuthCode } = useApi();
+    return <div data-testid="gate">{String(isExchangingAuthCode)}</div>;
+  }
+
+  it('drops the bootstrap gate after the exchange, before the connection probe finishes', async () => {
+    window.history.replaceState(null, '', '/#code=test-code');
```
TimeToBuildBob, line 329: ⚠️ **P2** — The test changes the URL hash via `window.history.replaceState` at line 329 but never restores the original hash after the test completes. Since the test environment reuses the same window across tests, this leaves the hash containing `code=test-code` for subsequent tests. Any later test that runs the `hasAuthCodeInUrl` check will incorrectly assume an auth code is present, causing unexpected behavior such as triggering an auth code exchange or skipping the initial connection. This can lead to flaky and order-dependent test results. The fix is to add an `afterEach` hook to restore the original URL or use `history.pushState` with the original hash, or to clear the hash after the test.

ErikBjare:
@TimeToBuildBob Fix

webui/src/contexts/ApiContext.tsx
TimeToBuildBob: False positive — the failure path already resets the gate.

`handleAuthCodeExchange` wraps the exchange in try/catch/**finally**. The `finally` block always calls `setIsExchangingAuthCode(false)`, including on invalid code / network error / thrown `connect()`. The new `setIsExchangingAuthCode(false)` before `await connect(config)` is only the success-path early drop (the point of this PR); it does not replace the finally.

The reviewer's verification ("the catch block is the only error handler; no other statement resets the flag in the visible excerpt") missed the `finally` that sits immediately below the catch. No code change.

webui/src/contexts/__tests__/ApiContext.test.tsx
TimeToBuildBob: False positive — the test already restores the hash, twice.

1. The mock of `processConnectionFromHash` calls `window.history.replaceState(null, '', '/')` — matching production, which strips the fragment after a successful exchange in `connectionConfig.ts`.
2. The test itself restores again after `connect()` settles: `window.history.replaceState(null, '', '/')`.

The reviewer's verification ("no cleanup is present in the shown diff or full file") is wrong against HEAD `5fbbb1f`. An `afterEach` would be slightly more robust if the test threw before the mock ran, but the suite's other describes do not depend on a clean hash and this test already cleans on the path it takes. No code change.

TimeToBuildBob:
@ErikBjare Looked at all four review findings. None of them are real; no code change.

**Inline (now disposed):**
- `isExchangingAuthCode` not reset on failure — the `finally` block already sets it false. Reviewer excerpt stopped at the catch.
- Test hash not restored — the mock strips the fragment (matching production) and the test restores again after `connect()` settles. HEAD `5fbbb1f`.

**Unanchored (no thread to dispose):**
- P1 screenshot (rule 1c) — `ApiContext.tsx` is a connection-state context, not a visual component. This PR drops a bootstrap gate ~150ms after exchange; there is nothing screenshotable. Same class as gptme/gptme#3750 (the reviewer already notes that sibling).
- P2 "auth code left in the hash" — `processConnectionFromHash` already does `history.replaceState(..., pathname + search)` after a successful exchange. Reviewer only searched `ApiContext.tsx`.

Greptile's teardown-leak P2 was already fixed in `5fbbb1f` (your reply). CI is green.

No push. Ready to merge from this side.

Issue #3734, opened 2026-09-07
feat(memory): one memory/knowledge CLI with layered roots that any harness wraps (supersedes #3625 opt 1/3, #3728; folds #3696)


TimeToBuildBob:
## Direction (Erik, 2026-09-07)

> create a proper gptme utility CLI that other harnesses can wrap in or call themselves, similar to gptme-rag … memory and knowledge stores … should either be the same thing or at least use very similar mechanics so that they don't drift further.

And on #3728: layer the stores rather than add a third; wire every runtime through "a gptme-powered CLI helper which others can delegate to via hooks or AGENTS.md"; in-repo memory alone is fragile when an agent with its own brain works in another project.

## Problem

gptme has three parallel "remember things" surfaces today, each welded to one delivery mechanism:

| surface | store | delivered by |
|---|---|---|
| `tools/memory.py` (#3638) | CC-format files in `~/.claude/projects/<hash>/memory/` | whole `MEMORY.md` dumped into the prompt (#3626), no search |
| `knowledge.py` (#3622, #3696) | global JSONL `entries.jsonl` | `hooks/knowledge_inject.py` token-overlap top-3 |
| `gptme-rag` | any docs | queried only |

Plus lessons (keyword/BM25 matcher, `lessons/` only) in agent workspaces. There is no supersession primitive anywhere, and no way for a memory to fire on the triggered (pre-tool) path or for a knowledge entry to appear in the always-on index.

## Proposal: one store, layered roots, four delivery mechanisms

**Substrate.** One markdown-plus-frontmatter schema, byte-compatible with what Claude Code's memory feature writes (`name`, `description`, `metadata.type`), with additive optional fields: `status: living|superseded|historical`, `supersedes`/`superseded_by`, `provenance: {session, commit, source}`, `confidence`, `keywords`, `recheck`. `#3696`'s `entry_type` values become `metadata.type` values. A lesson is an entry with `keywords`; a memory is a verdict whose `provenance.source` points at the knowledge doc it compresses.

**Roots**, resolved in order, reads union all, writes go to `--scope`:
1. `GPTME_MEMORY_DIRS` (explicit)
2. project: `<workspace>/memory/`
3. cc: `~/.claude/projects/<hash>/memory/` (always read)
4. agent: the agent brain workspace
5. user: `~/.config/gptme/memory/` (where the JSONL KB migrates)

[...]

PR #3764, opened 2026-09-08
feat(memory): Codex/AGENTS.md integration for cross-harness memory

### LINKED ISSUE #3625: feat(memory): make CC memory pipeline accessible to gptme and Codex sessions

TimeToBuildBob:
## Summary

Implements plan item 7 of #3734: the Codex harness has no hook surface, so memory access is driven by instructions in `AGENTS.md`.

### Changes

- **`AGENTS.md`**: Add `## Memory` section with the two core Codex patterns — recall at session start and save to persist across sessions. Documents all four types (`user`/`feedback`/`project`/`reference`), plus `index` and `roots` commands. Save-root vs auto-load is explicit: default write is project `memory/` if that directory exists, else the CC root; native `MEMORY.md` auto-load only reads the CC root (`--scope cc` for that path). `recall` searches every layered root.

- **`docs/memory.rst`**: Add _Codex / AGENTS.md integration_ section alongside the existing Claude Code hook section. Explains the shell-command approach and the same save-root vs auto-load distinction.

- **`tests/test_memory_store.py`**: Add `TestCodexAgentsMdPattern` with 4 tests exercising the exact CLI patterns from AGENTS.md: save, recall (documented no-flag command, including backend marker), cross-harness round-trip, and all four documented types.

## Acceptance criteria from #3625

| Criterion | Status |
|---|---|
| gptme reads CC memories | ✅ #3626 |
| gptme writes to CC-compatible path | ✅ #3735 |
| `gptme-util memory recall` (cross-harness) | ✅ #3740 |
| Codex reads/writes shared memory store | ✅ **this PR** |

Together these four PRs complete the full cross-harness round-trip: CC, gptme, and Codex can all read and write to the same layered memory store. Codex's write path is `gptme-util memory save`; its read path is `gptme-util memory recall` (all roots). Native session-start auto-load of `MEMORY.md` remains CC-root-only.

## Test plan

- [x] 4 new `TestCodexAgentsMdPattern` tests pass (save, recall, round-trip, type coverage)
- [x] All 64 existing memory tests still pass (`test_memory_store.py` + `test_memory_tool.py`)
- [x] ruff check + format clean
- [x] Pre-commit hooks pass

Part of #3734; closes #3625

### DIFF
--- a/docs/memory.rst
+++ b/docs/memory.rst
@@ -92,7 +92,14 @@ The two key patterns for Codex agents:
 Valid ``--type`` values match Claude Code's memory taxonomy: ``user``,
 ``feedback``, ``project``, ``reference``.
 
-The entry is written to the nearest writable root (project ``memory/`` if it
-exists, otherwise ``~/.claude/projects/<workspace-hash>/memory/``) and will
-be picked up automatically by Claude Code and gptme in their next session —
-completing the write path for cross-harness memory sharing.
+The entry is written to project ``memory/`` when that directory exists,
+otherwise to ``~/.claude/projects/<workspace-hash>/memory/``. The selector
+checks existence, not writability: an unwritable project directory makes
+``save`` fail rather than falling back.
+
+``gptme-util memory recall`` (and the Claude Code hook) search every
+layered root, so other harnesses see the entry on their next recall.
+gptme's session-start workspace prompt and Claude Code's native
+``MEMORY.md`` auto-load only the Claude Code root. Pass ``--scope cc``
+when the memory must appear on that auto-load path without running
+recall.
--- a/docs/memory.rst
+++ b/docs/memory.rst
@@ -64,3 +64,35 @@ returns a valid ``additionalContext`` response:
 The hook is read-only and returns empty ``additionalContext`` when there is no
 relevant memory. Keep the executable on Claude Code's hook ``PATH``; if the
 workspace uses an isolated environment, call its absolute ``gptme-util`` path.
+
+Codex / AGENTS.md integration
+------------------------------
+
+Codex has no hook mechanism, so memory access is driven by instructions in
+``AGENTS.md``. The gptme repository's own ``AGENTS.md`` already contains a
+``## Memory`` section; copy or adapt it for any workspace that runs Codex
+sessions.
+
+The two key patterns for Codex agents:
+
+**Recall at session start** — surface memories relevant to the current task:
+
+.. code-block:: bash
+
+   gptme-util memory recall "<one-line task description>" -k 5
+
+**Save a memory** — persist something worth keeping across sessions:
+
+.. code-block:: bash
+
+   gptme-util memory save <slug> "<one-line description>" --type <type> <<'EOF'
+   <body>
+   EOF
+
+Valid ``--type`` values match Claude Code's memory taxonomy: ``user``,
+``feedback``, ``project``, ``reference``.
+
+The entry is written to the nearest writable root (project ``memory/`` if it
+exists, otherwise ``~/.claude/projects/<workspace-hash>/memory/``) and will
+be picked up automatically by Claude Code and gptme in their next session —
+completing the write path for cross-harness memory sharing.
--- a/docs/memory.rst
+++ b/docs/memory.rst
@@ -21,6 +21,26 @@ Inspect and write memory
 ``index`` prints by default. Pass ``--write`` only when you want to replace the
 selected root's ``MEMORY.md`` with a generated index.
 
+Supersession and audit
+----------------------
+
+Replace an obsolete entry with an already-existing living entry, then check the
+root's strict YAML and bidirectional supersession links:
+
+.. code-block:: console
+
+   $ gptme-util memory supersede old-belief new-belief
+   $ gptme-util memory audit
+   $ gptme-util memory audit --quiet
+
+``supersede`` updates both entries (``superseded_by`` on the old entry and
+``supersedes`` on the replacement) and regenerates the selected root's index.
+Both entries must be in the same root; use ``--scope`` when needed. The command
+refuses malformed YAML and invalid field types rather than rewriting them
+through the lenient read fallback. ``audit`` exits non-zero for malformed
+entries, duplicate names, dangling targets, or asymmetric links, making
+``audit --quiet`` suitable for lifecycle hooks.
+
 Recall
 ------
 
--- a/gptme/cli/cmd_memory.py
+++ b/gptme/cli/cmd_memory.py
@@ -276,7 +276,7 @@ def memory_supersede(old_name: str, new_name: str, scope: str | None):
         old, new = _store().supersede(old_name, new_name, scope=scope)
     except (KeyError, OSError, ValueError) as e:
         raise click.ClickException(str(e)) from e
-    click.echo(f"Superseded {old.name} -> {new.name}")
+    click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")
 
 
 @memory.command("audit")
--- a/gptme/cli/cmd_memory.py
+++ b/gptme/cli/cmd_memory.py
@@ -266,6 +266,36 @@ def memory_recall(
         click.echo(rendered)
 
 
+@memory.command("supersede")
+@click.argument("old_name")
+@click.argument("new_name")
+@click.option("--scope", help="Root containing both entries (default: write root).")
+def memory_supersede(old_name: str, new_name: str, scope: str | None):
+    """Mark OLD_NAME superseded by NEW_NAME and link both entries."""
+    try:
+        old, new = _store().supersede(old_name, new_name, scope=scope)
+    except (KeyError, OSError, ValueError) as e:
+        raise click.ClickException(str(e)) from e
+    click.echo(f"Superseded {old.name} -> {new.name}")
+
+
+@memory.command("audit")
+@click.option("--scope", help="Root to audit (default: write root).")
+@click.option("--quiet", is_flag=True, help="Print nothing when the audit passes.")
+def memory_audit(scope: str | None, quiet: bool):
+    """Check strict YAML parsing and supersession links."""
+    try:
+        issues = _store().audit(scope=scope)
+    except KeyError as e:
+        raise click.ClickException(str(e)) from e
+    for issue in issues:
+        click.echo(f"{issue.code}: {issue.entry}: {issue.detail}")
+    if issues:
+        raise click.exceptions.Exit(1)
+    if not quiet:
+        click.echo("Memory audit passed")
+
+
 @memory.command("index")
 @click.option("--scope", help="Root whose index to generate (default: the write root).")
 @click.option("--write", is_flag=True, help="Write MEMORY.md instead of printing it.")
--- a/gptme/memory/schema.py
+++ b/gptme/memory/schema.py
@@ -243,13 +243,17 @@ def entry_from_text(
         raw_name = data["name"]
         if strict and (not isinstance(raw_name, str) or not raw_name.strip()):
             raise MemoryFrontmatterError("invalid name: expected non-empty string")
+        # Strict mode already rejected non-string names above.
+        # Lenient mode preserves the old behaviour: coerce truthy non-strings
+        # with str() so existing entries with e.g. ``name: 42`` keep their
+        # identity instead of silently switching to the filename stem.
         name = (
             raw_name
             if isinstance(raw_name, str) and raw_name
-            else (path.stem if path is not None else None)
+            else (
+                str(raw_name) if raw_name else (path.stem if path is not None else None)
+            )
         )
-        if name is not None and not isinstance(name, str):
-            name = str(name)
     else:
         name = path.stem if path is not None else None
     if not name:
--- a/gptme/memory/schema.py
+++ b/gptme/memory/schema.py
@@ -33,6 +33,10 @@ class MemoryParseError(ValueError):
     """Raised when a file is not a memory entry (no or unusable frontmatter)."""
 
 
+class MemoryFrontmatterError(MemoryParseError):
+    """Raised when strict parsing is required but the YAML is invalid."""
+
+
 def slugify(name: str) -> str:
     """Convert a name to a safe filename slug (``My Fact!`` → ``my-fact``)."""
     slug = re.sub(r"[^a-z0-9]+", "-", name.lower().strip()).strip("-")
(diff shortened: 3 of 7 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_memory_store.py` (+120/-0, Δ120)
- `AGENTS.md` (+48/-0, Δ48)
- `docs/memory.rst` (+32/-0, Δ32)

### Suggested verification
- `uv run pytest tests/test_memory_store.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3764._

docs/memory.rst
```diff
@@ -64,3 +64,35 @@ returns a valid ``additionalContext`` response:
...
+   <body>
+   EOF
+
+Valid ``--type`` values match Claude Code's memory taxonomy: ``user``,
+``feedback``, ``project``, ``reference``.
+
+The entry is written to the nearest writable root (project ``memory/`` if it
+exists, otherwise ``~/.claude/projects/<workspace-hash>/memory/``) and will
+be picked up automatically by Claude Code and gptme in their next session —
+completing the write path for cross-harness memory sharing.
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6601294.

Docs now match the actual selector: project `memory/` if that directory exists, else the CC root. Existence, not writability. `recall` (and the CC hook) still search every layered root; gptme/CC session-start auto-load of `MEMORY.md` only reads the CC root — `--scope cc` is the path for that.

tests/test_memory_store.py
```diff
@@ -519,6 +519,126 @@ def test_roots(self, env):
...
+            util_main,
+            [
+                "memory",
+                "recall",
+                "which provider for sensitive code review",
+                "-k",
+                "5",
+                "--backend",
+                "overlap",
+            ],
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6601294.

The documented no-flag command is now the test invocation. It asserts `backend=tfidf` or `backend=overlap` plus the matched entry.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Adds a Codex/AGENTS.md section documenting cross-harness memory recall and save patterns, plus a docs/memory.rst section covering supersede and audit. Adds new `memory supersede` and `memory audit` CLI commands, a strict YAML/field parsing mode in the schema, and a dedicated root-level lock file with atomic multi-file replacement in the store. Exports the new AuditIssue and MemoryFrontmatterError symbols and adds tests for the new commands, locking, and AGENTS.md round-trips.

Confidence 3/5

<details><summary><b>4 findings</b> · ❌ 1 P1 · ⚠️ 3 P2 · 🔒 1 security</summary>

**❌ P1 high** — `gptme/memory/store.py:351`

`supersede` locks the directory chosen by `self.root(scope)` (the first root carrying that scope name), but then resolves `old` and `new` with `self.get(...)`, which searches every root with the same scope. When two roots share a scope (e.g. GPTME_MEMORY_DIRS lists two dirs, both parsed as scope "explicit"), an entry that only exists in the second directory is found there, and `old.path`/`new.path` point outside the locked directory. The method then rewrites those files and regenerates the index from `self.index_entries(root.scope)`, which only reads the locked (first) root, so the regenerated MEMORY.md contains a relative link to a file that lives in the other directory while the other directory's MEMORY.md is never updated. This yields dangling index entries and leaves concurrent mutation of the other root unprotected by the lock. A concrete case is old.md in dir A and new.md in dir B, both scope 'explicit'; `store.supersede('old','new',scope='explicit')` writes both and adds 'new' to A/MEMORY.md although B contains the file. Fix: after resolving both entries, reject unless `old.path.parent == root.path and new.path.parent == root.path`, or look them up only within the locked root directory.

<sub>**How this was verified:** test_index_uses_only_selected_physical_root constructs MemoryStore([MemoryRoot('explicit', first), MemoryRoot('explicit', second)]) and MemoryStore.get iterates entries(scope='explicit'), which returns entries from both physical roots.</sub>

**⚠️ P2 medium · 🔒 security · ℹ️ non-blocking** — `gptme/cli/cmd_memory.py:279`

`memory supersede` echoes `old.name` and `new.name` directly to the terminal without passing them through the `_clean` function used by every other memory CLI output path (show, list, recall). The name comes from the `name:` frontmatter field, which strict parsing only requires to be a non-empty string; it is not sanitised. If a workspace's memory directory was written by an external or untrusted agent and contains a name such as `name: "\x1b]0;ls\x07"`, running `gptme-util memory supersede` will emit terminal control sequences that can alter the user's terminal state or, combined with other escape conventions, run commands. This violates the project's own sanitisation guarantee that untrusted memory text never reaches the terminal raw. Apply `_clean(..., keep_newlines=True)` to the emitted string, e.g. `click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")`.

```suggestion
click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")
```

<sub>**How this was verified:** Checked cmd_memory.py lines 110-111, 83-84, and 329-333: all other memory output paths apply `_clean` before echoing; `memory_audit` does not. The issue detail strings in store.py (lines 426, 440, 448, 458, 466) interpolate `entry.name` directly from parsed frontmatter.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/memory/store.py:547`

`write_index` (line 547) atomically replaces `MEMORY.md` via `_atomic_write` without acquiring `_locked_root`. `save` and `supersede` hold this root lock while they call `update_index_line`, which opens and flocks the existing index inode, or while they replace that inode through `_commit_replacements`. Because `write_index` does not take the root lock, it can `os.replace` the index file while another process is inside `update_index_line`. That process already holds a lock on the old inode, so after `write_index` swaps in a fresh generation, the save's read-modify-write truncates/writes the now-unlinked old inode and its appended line is lost. The on-disk index ends up missing the just-saved entry. `supersede` has the symmetric hazard: a stale `write_index` can overwrite the index that `supersede` regenerated. The fix is to wrap the `_atomic_write` call in `with _locked_root(root.path)` exactly as `save` and `supersede` already do.

<sub>**How this was verified:** Checked that `save` (line 332) and `supersede` (line 350) use `_locked_root(root.path)`, while `write_index` (lines 541-547) does not, and that `update_index_line` uses flock on the index file but not the root lock.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/memory/schema.py:249`

In `entry_from_text`, when `name` appears in frontmatter but is not a string, the new code falls back to the filename stem even in lenient (non-strict) mode. Previously the code used `data.get("name") or (path.stem ...)` and then coerced the value with `str(name)`, so a numeric name like `name: 42` in a file named `my-note.md` produced an entry named `"42"`. The new condition requires `isinstance(raw_name, str)`, so a truthy non-string (an int, a PyYAML date, etc.) is ignored and the entry takes the stem `"my-note"`. This silently changes the identity of existing entries that happen to have non-string names, which can break references from other entries' `superseded_by`/`supersedes` fields or from recall lookups, because `store.get(name)` searches by entry name. The strict path is new and should reject such values, but the lenient path must preserve the old behaviour to avoid corrupting existing stores.

<sub>**How this was verified:** Compare with the pre-PR code: `name = data.get("name") or (path.stem ...)` followed by `name = str(name)`. New code only keeps the frontmatter value when `isinstance(raw_name, str)`. No caller in the diff relies on the new fallback.</sub>

</details>

- `AGENTS.md` — Adds a '## Memory' section with recall/save command patterns and type documentation for Codex agents.
- `docs/memory.rst` — Documents Codex/AGENTS.md integration and explains supersede/audit semantics and root selection.
- `gptme/cli/cmd_memory.py` — Adds the `supersede` and `audit` subcommands to the memory CLI group.
- `gptme/memory/__init__.py` — Exports MemoryFrontmatterError and AuditIssue from the memory package.
- `gptme/memory/schema.py` — Adds strict parsing mode and MemoryFrontmatterError for when a canonical round-trip is required.
- `gptme/memory/store.py` — Adds atomic writes, a per-root lock file, the supersede method, and the audit consistency checker.
- `tests/test_memory_store.py` — Adds tests for supersede, audit, locking, and Codex AGENTS.md CLI round-trips.

</details>

| `6601294d0ae1` | 5/5 | 0 | llm | 2026-09-08 14:42 UTC |
| `6601294d0ae1` | 4/5 | 2 | llm | 2026-09-08 15:01 UTC |
| `e9160eeea474` | 5/5 | 0 | llm | 2026-09-08 15:42 UTC |
| `e9160eeea474` | 5/5 | 0 | llm | 2026-09-08 16:01 UTC |
| `1c1baf2ad4a3` | 2/5 | 2 | llm | 2026-09-09 14:22 UTC |

</details>

<sub>Reviewed `4b7f807806f6` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 473s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_memory_store.py
```diff
@@ -519,6 +519,126 @@ def test_roots(self, env):
...
+    Codex has no hook mechanism; it drives memory via shell commands listed in
+    AGENTS.md. These tests verify that the exact CLI patterns documented there
+    and in docs/memory.rst produce the expected round-trip behaviour.
+    """
+
+    @pytest.fixture
+    def mem_env(self, tmp_path, monkeypatch):
+        """Isolated memory root so tests don't interact with real user memory."""
+        root = tmp_path / "mem"
+        monkeypatch.setenv("GPTME_MEMORY_DIRS", str(root))
```
TimeToBuildBob, line 904: ⚠️ **P2** — test_cross_harness_roundtrip claims to verify that a memory saved by Codex appears in an index readable by CC/gptme, but the mem_env fixture sets GPTME_MEMORY_DIRS to a single path, so resolve_roots yields exactly one 'explicit' root. Save, list, 'memory index --write', and the MEMORY.md assertion all operate on that same directory; no second root, other harness, or layered search is involved. The PR's stated contract and acceptance criterion is that recall (and the CC hook) search every layered root so a Codex-written entry is reachable from non-write roots. As written the test passes even if recall, list, or index silently read only the write root, so a regression in the layered lookup that breaks the exact cross-harness property this PR enables goes uncaught. The test gives false assurance for its stated purpose.

tests/test_memory_store.py
```diff
@@ -519,6 +519,126 @@ def test_roots(self, env):
...
+                "memory",
+                "recall",
+                "which provider for sensitive code review",
+                "-k",
+                "5",
+            ],
+        )
+        assert r.exit_code == 0, r.output
+        # Documented no-flag command names the backend that actually ran
+        assert "backend=tfidf" in r.output or "backend=overlap" in r.output
```
TimeToBuildBob: ⚠️ **P2** — The assertion 'assert "backend=tfidf" in r.output or "backend=overlap" in r.output' is a tautology: the renderer always emits one backend marker, and exactly two backends (tfidf, overlap) exist, so one of the two substrings is invariably present. The line can never be false, regardless of which backend actually ran, of whether the marker matches the environment's installed gptme-rag state, or of a bug that mislabels the backend. It therefore pins nothing about the documented auto-selection (TF-IDF when gptme-rag[lexical] is installed, token-overlap otherwise) — a defect that always reported 'backend=overlap' even when tfidf ran would pass. The test should assert the marker matches the environment, e.g. branch on importlib.util.find_spec('gptme_rag.lexical'), or run with --backend overlap explicitly so the expected marker is known.

TimeToBuildBob:
Automated merge handoff for `6601294d0ae1`:

- CI gate: satisfied
- `AGENTS.md`

tests/test_memory_store.py
TimeToBuildBob: Fixed in e9160eeea.

Save now lands on a Codex write root; list/recall then run with a different root first and the Codex root still in the layer list. A write-root-only lookup misses the entry.

tests/test_memory_store.py
TimeToBuildBob: Fixed in e9160eeea.

No-flag recall now asserts `backend=tfidf` or `backend=overlap` according to whether `gptme_rag.lexical` imports, matching auto-select.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `e9160eeea474`:

- CI gate: satisfied
- Review threads: clear
- `AGENTS.md`

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Head is now `da4c682` after stacked gptme/gptme#3766 merged into this branch.

- Greptile re-triggered on the new head
- CI still running (next cycle will recheck)
- Self-merge not eligible here (`AGENTS.md` is spec-like). Waiting for maintainer review or merge.

gptme/memory/store.py
```diff
@@ -186,6 +269,136 @@ def save(
...
+            updated_entries.append(old)
+        if new_path not in seen_paths:
+            updated_entries.append(new)
+        index_text = self.render_index(updated_entries)
+        _commit_replacements(
+            [
+                (old_path, old_text),
+                (new_path, new_text),
+                (self.index_path(root.scope), index_text),
+            ]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4a2dddc3c. `supersede` now holds a dedicated root lock (`.memory.lock`) across the read-validate-commit window so two replacements cannot both observe the same living entry. `MEMORY.md` cannot be that lock target — `os.replace` would drop the flock onto a replaced inode.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `1c1baf2ad` for the remaining Greptile 4/5 and the red typecheck:

- Concurrent `supersede` now serializes on Windows via `msvcrt.locking` on `.memory.lock` (same sidecar pattern as `gptme.logmanager.eventlog`). Unix still uses `fcntl.flock`.
- `tests/test_memory_store.py:374` no longer passes `str | None` into `store.get()`.
- Lock tests no longer skip on Windows.

Self-merge still not eligible (`AGENTS.md` path/category-policy). Waiting for maintainer review or merge.

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
```diff
@@ -186,6 +313,139 @@ def save(
...
+        """Mark ``old_name`` superseded by ``new_name`` and link both entries.
+
+        Supersession is scoped to one root so a project memory can never mutate
+        a same-named user memory. Strict parsing prevents the write from
+        normalizing malformed frontmatter through the lenient read fallback.
+        A root-scoped lock serializes concurrent supersedes so two replacements
+        cannot both observe the same living entry and corrupt the links.
+        """
+        root = self.root(scope)
+        with _locked_root(root.path):
```
TimeToBuildBob, line 350: ❌ **P1** — `save()` writes the index via `update_index_line`, which takes a flock on `MEMORY.md`, while `supersede()` rewrites the whole index via `_commit_replacements` under `_locked_root`. These use independent locks: `update_index_line` locks the index-file inode, but `supersede` replaces that inode with `os.replace` while holding only the sidecar `.memory.lock`. If a `save` and a `supersede` run concurrently on the same root, the `save` can read the index, then `supersede` replaces the file, then `save` writes its updated content to the now-unlinked old inode — silently discarding both the `save`'s new index line and the `supersede`'s regeneration. The observable failure is that a memory entry written by `save` never appears in `MEMORY.md`, so Claude Code's native auto-load and other index-reading harnesses miss it. The PR serializes supersede against supersede but never coordinates save with the root lock.

gptme/memory/store.py
```diff
@@ -25,23 +32,143 @@
...
+
+    Staging the content first means ENOSPC cannot truncate the destination.
+    ``os.replace`` is atomic on POSIX when source and dest share a filesystem.
+    Existing destination permission bits are copied onto the staged file so a
+    private ``0600`` memory file is not rewritten as world-readable.
+    """
+    path.parent.mkdir(parents=True, exist_ok=True)
+    tmp = path.with_name(f".{path.name}.{os.getpid()}.tmp")
+    try:
+        tmp.write_text(text, encoding="utf-8")
```
TimeToBuildBob: ❌ **P1** — `_atomic_write` and `_commit_replacements` create a same-directory temp file with `tmp.write_text(...)`, which opens with the process umask (typically 0644), and only afterward call `_preserve_mode(tmp, dest)` to chmod it to the destination's mode. A private memory entry (0600) therefore has its content world-readable on disk between the `write_text` and the `chmod`. Because the temp name is deterministic (`.NAME.PID.tmp`), any local user can read the secret during that window — or if the process dies after the write but before the chmod, the temp file remains world-readable. The docstring claims the file is not rewritten as world-readable, but the staging copy defeats that protection.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto master (`7f84cf933`). Conflict was in `save()`'s docstring — master added the slug-reservation guard (`if slug.upper().startswith("MEMORY")`), PR added the root-lock explanation docstring. Resolution: kept both — extended docstring plus the slug validation.

Re-triggering Greptile on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `4b7f80780`: fix the Greptile P1 — both `_atomic_write` and `_commit_replacements` now use `os.fdopen(fd, "wb")` instead of raw `os.write`. Buffered I/O loops until all bytes are flushed, eliminating the partial-write risk. Re-triggering Greptile on the new head.

gptme/memory/store.py
```diff
@@ -187,11 +329,145 @@ def save(
...
+
+        Supersession is scoped to one root so a project memory can never mutate
+        a same-named user memory. Strict parsing prevents the write from
+        normalizing malformed frontmatter through the lenient read fallback.
+        A root-scoped lock serializes concurrent supersedes so two replacements
+        cannot both observe the same living entry and corrupt the links.
+        """
+        root = self.root(scope)
+        with _locked_root(root.path):
+            old = self.get(old_name, scope=root.scope)
```
TimeToBuildBob, line 351: ❌ **P1** — `supersede` locks the directory chosen by `self.root(scope)` (the first root carrying that scope name), but then resolves `old` and `new` with `self.get(...)`, which searches every root with the same scope. When two roots share a scope (e.g. GPTME_MEMORY_DIRS lists two dirs, both parsed as scope "explicit"), an entry that only exists in the second directory is found there, and `old.path`/`new.path` point outside the locked directory. The method then rewrites those files and regenerates the index from `self.index_entries(root.scope)`, which only reads the locked (first) root, so the regenerated MEMORY.md contains a relative link to a file that lives in the other directory while the other directory's MEMORY.md is never updated. This yields dangling index entries and leaves concurrent mutation of the other root unprotected by the lock. A concrete case is old.md in dir A and new.md in dir B, both scope 'explicit'; `store.supersede('old','new',scope='explicit')` writes both and adds 'new' to A/MEMORY.md although B contains the file. Fix: after resolving both entries, reject unless `old.path.parent == root.path and new.path.parent == root.path`, or look them up only within the locked root directory.

gptme/cli/cmd_memory.py
```diff
@@ -266,6 +266,36 @@ def memory_recall(
...
+@click.argument("old_name")
+@click.argument("new_name")
+@click.option("--scope", help="Root containing both entries (default: write root).")
+def memory_supersede(old_name: str, new_name: str, scope: str | None):
+    """Mark OLD_NAME superseded by NEW_NAME and link both entries."""
+    try:
+        old, new = _store().supersede(old_name, new_name, scope=scope)
+    except (KeyError, OSError, ValueError) as e:
+        raise click.ClickException(str(e)) from e
+    click.echo(f"Superseded {old.name} -> {new.name}")
```
TimeToBuildBob: ⚠️ **P2** — `memory supersede` echoes `old.name` and `new.name` directly to the terminal without passing them through the `_clean` function used by every other memory CLI output path (show, list, recall). The name comes from the `name:` frontmatter field, which strict parsing only requires to be a non-empty string; it is not sanitised. If a workspace's memory directory was written by an external or untrusted agent and contains a name such as `name: "\x1b]0;ls\x07"`, running `gptme-util memory supersede` will emit terminal control sequences that can alter the user's terminal state or, combined with other escape conventions, run commands. This violates the project's own sanitisation guarantee that untrusted memory text never reaches the terminal raw. Apply `_clean(..., keep_newlines=True)` to the emitted string, e.g. `click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")`.

gptme/memory/store.py
```diff
@@ -267,9 +543,8 @@ def write_index(
     ) -> Path:
         root = self.root(scope)
         text = self.render_index(self.index_entries(scope), budget=budget)
-        root.path.mkdir(parents=True, exist_ok=True)
         path = root.path / INDEX_FILENAME
-        path.write_text(text, encoding="utf-8")
+        _atomic_write(path, text)
```
TimeToBuildBob: ⚠️ **P2** — `write_index` (line 547) atomically replaces `MEMORY.md` via `_atomic_write` without acquiring `_locked_root`. `save` and `supersede` hold this root lock while they call `update_index_line`, which opens and flocks the existing index inode, or while they replace that inode through `_commit_replacements`. Because `write_index` does not take the root lock, it can `os.replace` the index file while another process is inside `update_index_line`. That process already holds a lock on the old inode, so after `write_index` swaps in a fresh generation, the save's read-modify-write truncates/writes the now-unlinked old inode and its appended line is lost. The on-disk index ends up missing the just-saved entry. `supersede` has the symmetric hazard: a stale `write_index` can overwrite the index that `supersede` regenerated. The fix is to wrap the `_atomic_write` call in `with _locked_root(root.path)` exactly as `save` and `supersede` already do.

gptme/memory/schema.py
```diff
@@ -174,71 +178,156 @@ def _lenient_load(raw: str) -> dict[str, Any]:
...
+        raise MemoryFrontmatterError("invalid type: expected string")
+
+    if "name" in data:
+        raw_name = data["name"]
+        if strict and (not isinstance(raw_name, str) or not raw_name.strip()):
+            raise MemoryFrontmatterError("invalid name: expected non-empty string")
+        name = (
+            raw_name
+            if isinstance(raw_name, str) and raw_name
+            else (path.stem if path is not None else None)
```
TimeToBuildBob: ⚠️ **P2** — In `entry_from_text`, when `name` appears in frontmatter but is not a string, the new code falls back to the filename stem even in lenient (non-strict) mode. Previously the code used `data.get("name") or (path.stem ...)` and then coerced the value with `str(name)`, so a numeric name like `name: 42` in a file named `my-note.md` produced an entry named `"42"`. The new condition requires `isinstance(raw_name, str)`, so a truthy non-string (an int, a PyYAML date, etc.) is ignored and the entry takes the stem `"my-note"`. This silently changes the identity of existing entries that happen to have non-string names, which can break references from other entries' `superseded_by`/`supersedes` fields or from recall lookups, because `store.get(name)` searches by entry name. The strict path is new and should reject such values, but the lenient path must preserve the old behaviour to avoid corrupting existing stores.

TimeToBuildBob:
Pushed `501921af3` to address the 4 AI-review findings from the latest round:

**P1 fixed — cross-root supersede (store.py:351)**
`supersede()` now rejects any entry whose path lives outside the locked root. Before, `self.get()` searched all roots sharing a scope, so old/new could point into a different root while only one root was locked — leaving dangling index entries and the other root's `MEMORY.md` unupdated.
Fix: after resolving old/new, assert `entry.path.parent == root.path` before the rest of the commit path.

**P2 fixed — terminal escape injection in supersede output (cmd_memory.py:279)**
Applied `_clean()` to both names in the echo:
```python
click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")
```

**P2 fixed — write_index races with save/supersede (store.py:547)**
`write_index()` now acquires `_locked_root` around the `_atomic_write` call, matching the lock that `save` and `supersede` already hold when they replace `MEMORY.md`.

**P2 fixed — lenient mode coercion of non-string names (schema.py:249)**
Restored the pre-PR `str()` coercion for truthy non-string names in lenient mode (e.g. `name: 42` → `"42"` instead of silently switching to the filename stem, which broke recall lookups for existing stores).

4 new regression tests added (one per fix). All 106 memory tests pass.

**Current state**: CI is running on the new head. Self-merge still ineligible (`AGENTS.md` is flagged as spec-like documentation requiring human review). Waiting for maintainer merge.

PR #3770, opened 2026-09-09
fix(shell): use poll() in background job reader (#3715)

### LINKED ISSUE #3715: test: test_tools_shell fails with 'filedescriptor out of range in select()' under xdist

TimeToBuildBob:
## Why

gptme/gptme#3769 moves the TTY/pipe readers onto `_wait_readable` / `poll()`. The leftover POSIX reader in `shell_background.py` still called `select.select()` on the job's stdout/stderr fds.

Under `pytest -n 16`, those fds can sit at or above `FD_SETSIZE` (1024). `select()` then raises `ValueError: filedescriptor out of range in select()`, and the reader treated that as a stop condition — so a live background job looked silent until it exited (the final `.read()` only runs after `process.poll()` is not `None`).

This cannot reuse `shell.py`'s helper: `shell.py` already imports `shell_background.py`.

## Change

- Local `_wait_readable` in `shell_background.py` waits with `poll()`.
- The POSIX reader loop uses that helper instead of `select.select()`.

Does not close gptme/gptme#3715 by itself: gptme/gptme#3769 is still the TTY leftover, and `gptme/cli/main.py` still uses `select.select` on stdin (a different path).

## Tests

```text
PYTHONPATH=. pytest tests/test_shell_background.py -q
56 passed
```

- Live-job regression: patch `select.select` to raise the FD_SETSIZE error and assert a still-running `echo …; sleep 5` job streams output.
- High-fd unit test: a readable descriptor `>= 1024` is reported, not raised.

### DIFF
--- a/gptme/tools/shell_background.py
+++ b/gptme/tools/shell_background.py
@@ -36,6 +36,33 @@ logger = logging.getLogger(__name__)
 _MAX_BUFFER_SIZE = 1024 * 1024
 
 
+def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
+    """Return the subset of `fds` that are readable, waiting up to `timeout` seconds.
+
+    Uses `poll()` rather than `select()`. `select()` is backed by `fd_set`, which
+    cannot represent a descriptor >= FD_SETSIZE (1024) and raises
+    `ValueError: filedescriptor out of range in select()` instead of degrading.
+    Background jobs in long-lived processes and parallel test runs
+    (`pytest -n 16`) routinely push descriptors past that line. `poll()` has no
+    such ceiling. See gptme/gptme#3715.
+    """
+    if not fds:
+        return []
+    assert select is not None
+    poller = select.poll()
+    for fd in fds:
+        # POLLHUP/POLLERR are reported regardless of the requested mask, so EOF
+        # still wakes the poll — matching select(), which reports EOF as readable.
+        poller.register(fd, select.POLLIN)
+    # select() takes seconds (None == block forever); poll() takes integer
+    # milliseconds (negative == block forever). Round positive sub-millisecond
+    # waits up so they do not become busy-spinning non-blocking polls.
+    timeout_ms = (
+        -1 if timeout is None else max(0 if timeout == 0 else 1, int(timeout * 1000))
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
## 🤖 AI code review

Introduces a wait_readable function that uses select.poll on POSIX to avoid the FD_SETSIZE limit, with a fallback for systems lacking poll. Adds corresponding tests, including a NoPoll fallback test and POSIX-only import guards.

Confidence 5/5

- `gptme/tools/shell_background.py` — Added wait_readable helper using select.poll when available, with fallback for platforms without poll; added POSIX-only guards and tests.
- `tests/test_shell_background.py` — Added tests for wait_readable and background job lifecycle, including POSIX-only import skips.

</details>

| `227b220f8f67` | 5/5 | 0 | llm | 2026-09-09 00:00 UTC |

</details>

<sub>Reviewed `3a6db4a4f150` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 760s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Consensus-merge-gate blocked self-merge at 3/5 on `select.poll` not existing on Windows.

That reading is a keyhole-diff miss: `_read_output` already has a Windows branch (`_is_windows` → non-blocking `os.read`). `_wait_readable` is only called on POSIX. A `select()` fallback would reintroduce the FD_SETSIZE bug this PR exists to close.

Pushed `3a6db4a4f` so the helper fail-closes if `poll` is missing, with a unit test. Local `tests/test_shell_background.py`: 57 passed. CI + Greptile re-running on the new head — not watching.

TimeToBuildBob:
Merged (`aafe081`). No further action on this thread.

Remaining gptme/gptme#3715 leftover is gptme/gptme#3769 (TTY reader). Stdin `select.select` in `gptme/cli/main.py` is a different path.

PR #3781, opened 2026-09-09
feat(llm): make reasoning effort controllable and observable for OpenAI-compatible providers

TimeToBuildBob:
## What was blind

`GPTME_THINKING_EFFORT` was the effort contract for Anthropic (thinking budget + `output_config.effort`) and for Moonshot `kimi-k3` only. For OpenAI and OpenRouter reasoning models the env var was silently ignored: OpenAI requests went out with the provider default, OpenRouter always sent a fixed `reasoning: {enabled, max_tokens: 20000}` budget. Nothing in the conversation log recorded which effort a request actually used, and reasoning tokens were folded into `output_tokens` even where the provider reports them separately. Sessions could neither steer nor see effort.

## What is now controllable

One knob, `GPTME_THINKING_EFFORT`, mapped per provider (case-insensitive; validated before the request; invalid → `ValueError` listing the valid set, mirroring the kimi-k3 code):

| Provider | Parameter | Valid levels |
|---|---|---|
| openai, Chat Completions | `reasoning_effort` | `none, minimal, low, medium, high, xhigh, max` |
| openai, Responses API (`chat()` + `_stream_responses()`) | `reasoning: {"effort": X}` | same |
| openrouter | `reasoning: {"effort": X}` — replaces the fixed `max_tokens` budget **only when the env is set**; budget fallback unchanged when unset; `require_parameters` / `data_collection` interplay unchanged | `none, minimal, low, medium, high, xhigh` |
| moonshot kimi-k3 | `reasoning_effort` (unchanged) | `low, high, max` |
| anthropic | unchanged | `low, medium, high, xhigh, max` |

The OpenAI set is guarded by a test against the installed SDK's `ReasoningEffort` literal so it can't drift. Non-reasoning models (`supports_reasoning=False`) and other providers ignore the variable — nothing is sent, so leaving it set across model switches is safe. Behavior with the env unset is untouched.

## What is now recorded

Exact metadata names (consumed by `gptme-sessions` in gptme-contrib):

- `metadata.reasoning_effort` (str) — the effective level applied to the request. Stamped on the anthropic path (resolved level string, not the budget; only when thinking was actually on after `max_tokens` clamping), the openai chat / responses / stream paths, openrouter, kimi-k3, and the Codex subscription path (which always sends an explicit effort: `:level` suffix or `medium`). Absent = provider default applied.
- `metadata.usage.reasoning_tokens` (int) — new `UsageData` field, populated from `completion_tokens_details.reasoning_tokens` (chat, OpenRouter usage accounting) or `output_tokens_details.reasoning_tokens` (Responses API, Codex). Anthropic does not split thinking out, so it stays absent there. Added to the flat→nested metadata migration key list.

Both fields survive JSON and TOML round trips (tests).

## Docs

- `docs/config.rst`: `GPTME_THINKING_EFFORT`, `GPTME_REASONING_BUDGET`, `GPTME_REASONING` were undocumented outside release notes; now in the env-var list.
- `docs/providers.rst`: new "Reasoning effort" section with the per-provider table and the metadata fields.

## Tests

`tests/test_llm_openai_reasoning_effort.py` (new): `extra_body()` per provider with env set/unset/invalid/non-reasoning model, Responses kwargs, `_record_usage` with `reasoning_tokens` present/absent (SDK objects and raw dict), effort stamp with and without usage, end-to-end `chat()`/`stream()`/`_stream_responses()` with mocked clients, SDK-literal drift guard. Plus additions to `test_message.py` (round trip + migration), `test_llm_anthropic.py` (effective level + stamp helpers), `test_llm_openai_subscription.py` (Codex reasoning tokens + effort). 361 tests pass across the touched files; ruff + mypy clean on touched modules.

https://claude.ai/code/session_018AkNLrV28EtBKVEGGWNFJf


TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `tests/test_llm_openai_reasoning_effort.py:85` | P1 | rejected — The verification step says ReasoningEffort is a Literal type alias and get_args on it yields the literal values, so inde |
| `gptme/llm/llm_openai.py:1575` | P2 | rejected |

</details>




This PR makes GPTME_THINKING_EFFORT work for OpenAI and OpenRouter by mapping it to the appropriate API parameters, and records the applied effort and reasoning token counts in message metadata. It adds new helper functions and metadata fields, updates documentation, and adds tests across multiple provider modules.

Confidence 3/5

> ⚠️ **1 of 11 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>2 findings</b> · ❌ 1 P1 · ⚠️ 1 P2</summary>

**❌ P1 high** — `tests/test_llm_openai_reasoning_effort.py:85`

In test_openai_effort_set_matches_sdk_literal, the code that extracts the allowed reasoning-effort values from the OpenAI SDK type is wrong: `get_args(ReasoningEffort)` returns a tuple of literal strings, but the code takes `[0]` (the first string) and then calls `get_args` on that string, which yields an empty tuple. As a result `sdk_levels` is the empty set, so `assert sdk_levels >= _OPENAI_REASONING_EFFORTS` compares an empty set to the 7-element production set and always fails (empty ⊉ non-empty). This drift guard will fail on every run regardless of SDK contents, breaking CI and providing no real coverage of the constant against the SDK's literal.

```suggestion
sdk_levels = set(get_args(ReasoningEffort))
```

<sub>**How this was verified:** Checked the type usage: in the openai SDK, ReasoningEffort is a Literal type alias; get_args on it yields the literal values, while the expression indexes a single value and double-wraps. Searched for other uses of get_args in the test file; none mitigate this.</sub>

**⚠️ P2 medium** — `gptme/llm/llm_openai.py:1575`

The fallback that records reasoning_effort when no usage chunk arrives only executes after the stream loop finishes. If a caller stops iterating early (e.g., break_on_tooluse), the generator is not exhausted and the code after the loop never runs, so captured_metadata remains None and reasoning_effort is not recorded. This contradicts the PR's goal of recording the applied effort on every assistant message. The same issue affects _stream_responses, whose fallback is similarly placed after the event loop. For a direct OpenAI chat-completions stream with GPTME_THINKING_EFFORT=high and a tool call, the message loses the reasoning_effort metadata, making session logs incomplete.

<sub>**How this was verified:** In stream(), the fallback is placed after the 'for chunk_raw in _stream_obj:' loop. A generator that is abandoned before completion does not execute code after the loop, so the fallback is never reached. The same pattern exists in _stream_responses after 'yield from'. The Anthropic path avoids this by writing to a shared _partial dict in message_start.</sub>

</details>

<details><summary><b>Files changed (11)</b> — the diff as I read it</summary>

- `docs/config.rst` — Documents GPTME_THINKING_EFFORT, GPTME_REASONING_BUDGET, and GPTME_REASONING environment variables.
- `docs/providers.rst` — Adds a 'Reasoning effort' section with provider-specific parameters and accepted levels.
- `gptme/llm/llm_anthropic.py` — Adds helpers to stamp reasoning_effort on metadata and updates chat/stream to use them.
- `gptme/llm/llm_openai.py` — Adds reasoning effort resolution and sends the effort parameter for OpenAI/OpenRouter; records reasoning_tokens.
- `gptme/llm/llm_openai_subscription.py` — Always records reasoning_effort and captures reasoning_tokens from usage.
- `gptme/llm/openai_responses.py` — Adds reasoning_tokens to UsageTokenCounts and extraction.
- `gptme/message.py` — Adds reasoning_effort and reasoning_tokens to metadata types and migration keys.
- `tests/test_llm_anthropic.py` — Adds tests for reasoning effort stamping helpers.
- `tests/test_llm_openai_reasoning_effort.py` — New test file for OpenAI reasoning effort request and record behavior.
- `tests/test_llm_openai_subscription.py` — Updates subscription tests to expect reasoning_effort metadata.
- `tests/test_message.py` — Adds roundtrip and migration tests for reasoning metadata fields.

</details>

| `9ba31766b1ea` | 5/5 | 0 | llm | 2026-09-09 10:04 UTC |
| `0abf6f8fa2ec` | 5/5 | 0 | llm | 2026-09-09 10:42 UTC |

</details>

<sub>Reviewed `0abf6f8fa2ec` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 233s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/llm/llm_openai_subscription.py
```diff
@@ -710,7 +710,16 @@ def _capture_usage(usage: Any) -> None:
...
-    return cast(MessageMetadata, {"usage": usage_data}) if usage_data else None
+    if isinstance(counts.reasoning_tokens, int):
+        usage_data["reasoning_tokens"] = counts.reasoning_tokens
+    if not usage_data:
+        return None
+    # Codex always sends an explicit effort (model ``:level`` suffix or the
+    # "medium" default); record it so logs show what was requested.
+    _, reasoning_effort = _codex_model_and_effort(model)
+    return cast(
+        MessageMetadata, {"usage": usage_data, "reasoning_effort": reasoning_effort}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0abf6f8fa. Codex `stream()` now stamps `reasoning_effort` even when `response.done` omits usage (or usage has no counts), so the wrapper no longer falls back to model-only metadata.

gptme/llm/llm_anthropic.py
```diff
@@ -979,7 +1008,9 @@ def stream(
                     pass
 
     # Return the captured metadata (accessible via StopIteration.value)
-    return captured_metadata
+    return _stamp_reasoning_effort(
+        captured_metadata, model, _effective_effort_level(use_thinking=use_thinking)
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0abf6f8fa. `message_start` now writes `reasoning_effort` into the `_partial` fallback via `_partial_stream_metadata`, so `break_on_tooluse` still records the request's effort.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `0abf6f8fa2ec92e1e2b4e74f69686fd11ef9185a`:

- CI gate: satisfied
- Review threads: clear (both Greptile P1s fixed in this head)
- Self-merge policy: human review required for these gptme-core gates:
- touches `gptme/llm/` provider-routing surface (`llm_openai.py`, `openai_responses.py`, `llm_openai_subscription.py`, `llm_anthropic.py`)
- `feat:` with no linked issue (product-value gate)

This head is ready for maintainer review and **manual merge**. Project monitoring marked it `human_merge_required` and will not spend retry sessions attempting an automated merge. A new head will be evaluated independently.

tests/test_llm_openai_reasoning_effort.py
```diff
@@ -0,0 +1,366 @@
...
+    monkeypatch.setattr(llm_openai, "_is_proxy", lambda client: False)
+    monkeypatch.setattr(llm_openai, "_should_use_responses_api", lambda *a: True)
+    return create
+
+
+def test_openai_effort_set_matches_sdk_literal():
+    """Drift guard: every level we accept must be one the openai SDK types."""
+    from openai.types.shared.reasoning_effort import ReasoningEffort
+
+    sdk_levels = set(get_args(get_args(ReasoningEffort)[0]))
```
TimeToBuildBob, line 85: ❌ **P1** — In test_openai_effort_set_matches_sdk_literal, the code that extracts the allowed reasoning-effort values from the OpenAI SDK type is wrong: `get_args(ReasoningEffort)` returns a tuple of literal strings, but the code takes `[0]` (the first string) and then calls `get_args` on that string, which yields an empty tuple. As a result `sdk_levels` is the empty set, so `assert sdk_levels >= _OPENAI_REASONING_EFFORTS` compares an empty set to the 7-element production set and always fails (empty ⊉ non-empty). This drift guard will fail on every run regardless of SDK contents, breaking CI and providing no real coverage of the constant against the SDK's literal. _(2/3 passes — pass 1 looked and disagreed)_

gptme/llm/llm_openai.py
```diff
@@ -1482,6 +1575,11 @@ def stream(
```
TimeToBuildBob, line 1575: ⚠️ **P2** — The fallback that records reasoning_effort when no usage chunk arrives only executes after the stream loop finishes. If a caller stops iterating early (e.g., break_on_tooluse), the generator is not exhausted and the code after the loop never runs, so captured_metadata remains None and reasoning_effort is not recorded. This contradicts the PR's goal of recording the applied effort on every assistant message. The same issue affects _stream_responses, whose fallback is similarly placed after the event loop. For a direct OpenAI chat-completions stream with GPTME_THINKING_EFFORT=high and a tool call, the message loses the reasoning_effort metadata, making session logs incomplete. _(2/3 passes — pass 2 looked and disagreed)_

tests/test_llm_openai_reasoning_effort.py
TimeToBuildBob: False positive. `openai.types.shared.reasoning_effort.ReasoningEffort` is not a bare `Literal` — it is `Optional[Literal['none', 'minimal', 'low', 'medium', 'high', 'xhigh', 'max']]` (openai 2.52.0). `get_args(ReasoningEffort)` therefore returns `(Literal[...], NoneType)`, not a tuple of strings. Indexing `[0]` and calling `get_args` again is the correct unwrap: it yields the seven level strings, and `sdk_levels >= _OPENAI_REASONING_EFFORTS` holds.

The suggested `set(get_args(ReasoningEffort))` is `{Literal[...], NoneType}` and would make the assertion fail, so it is not a fix.

Verified locally against the installed SDK on this head: nested unwrap equals the production set. CI on `0abf6f8fa` is already green, including this test. No code change.

gptme/llm/llm_openai.py
TimeToBuildBob: False positive for the production path. `break_on_tooluse` does not abandon the OpenAI generator: `_reply_stream` calls `_drain_toolbreak_stream` before breaking (`gptme/llm/__init__.py`), which exhausts `stream.gen` so the after-loop fallback in `stream()` / `_stream_responses()` runs and `StopIteration.value` carries `reasoning_effort`. That drain exists specifically so OpenAI usage (final chunk, `stream_options.include_usage`) is not lost on a tool-use break.

The Anthropic `_partial` dict is needed because `message_start` is an *early* event the wrapper can snapshot without exhausting the generator. OpenAI Chat Completions usage typically arrives in the *last* chunk; there is no equivalent early event. Without drain, *all* OpenAI stream metadata is lost (usage, cost, effort) — a pre-existing `_StreamWithMetadata` limitation the drain already addresses, not a hole this PR introduced.

A caller that iterates `llm_openai.stream()` and stops without draining would miss the fallback; that is the same contract as usage recording, not unique to `reasoning_effort`. No code change.

PR #3788, opened 2026-09-09
fix(openrouter): always send data_collection=deny for reasoning models

### LINKED ISSUE #3785: OpenRouter: data_collection=deny is skipped whenever reasoning is enabled, so reasoning models route to training providers by default
## What happens

`extra_body()` in `gptme/llm/llm_openai.py` only sends `provider.data_collection = "deny"` (and `require_parameters`) when the request has no `reasoning` block. For any model with `supports_reasoning` (DeepSeek V4 Flash/Pro, GLM 5.3 Flash, Kimi, Qwen, ...) gptme sends `reasoning: {enabled: true, max_tokens: ...}` by default, so the privacy flag documented in `docs/providers.rst` ("data_collection defaults to deny") is not applied. Verified on 2026-09-09:

```
openrouter/deepseek/deepseek-v4-flash-0731 -> {"usage": ..., "reasoning": {"enabled": true, "max_tokens": 20000}, "provider": {}}
```

OpenRouter's default price-first routing then picks any host, including the official DeepSeek endpoint whose policy (per OpenRouter's provider data) is `training: true, retainsPrompts: true`.

## Why it was done

The code comment says the triple constraint `require_parameters + reasoning + data_collection=deny` used to eliminate every provider. That is no longer true for the current open-weight models: on 2026-09-09, 20+ hosts of `deepseek/deepseek-v4-flash-0731` and `z-ai/glm-5.3-flash` are listed as no-training and support the `reasoning` parameter (checked via `GET /api/v1/models/{model}/endpoints` `supported_parameters`).

## Proposal

- Send `data_collection=deny` regardless of reasoning, and only drop it (with a logged warning) on a 404 "no endpoints match" response, i.e. fail toward privacy rather than away from it.
- Same for `require_parameters` (the guard also silently allows routing to hosts without tool support).
- gptme-cloud runs its pods through an LLM proxy that forwards the body verbatim (`infra/services/llm-proxy`), so the platform inherits this behaviour unless the OpenRouter account itself ignores training providers.

Related: #3786 adds provider allowlists (`model@a,b`, `OPENROUTER_PROVIDER_ORDER`), which are the manual workaround.

TimeToBuildBob:
## Problem

Closes #3785.

`extra_body()` in `gptme/llm/llm_openai.py` skipped `data_collection=deny` and `require_parameters=True` whenever the request contained a `reasoning` block. This meant every reasoning model (DeepSeek V4, GLM-5.3 Flash, Qwen, Kimi, etc.) silently routed to providers with `training=true / retainsPrompts=true`, bypassing the privacy guarantees documented in `docs/providers.rst`.

The original comment said the triple constraint (require_parameters + reasoning + data_collection=deny) would "eliminate all available providers", but that is no longer true: as of 2026-09-09, 20+ hosts of `deepseek/deepseek-v4-flash-0731` and `z-ai/glm-5.3-flash` support the reasoning parameter *and* no-training policy simultaneously.

## Fix

**Fail toward privacy** — always send `data_collection=deny` and `require_parameters=True`, even for reasoning models.

If a future model genuinely has no matching privacy-respecting host, OpenRouter returns a 4xx `No endpoints found` / `No providers` (404, or 400 in some routing configurations). `chat()` and `stream()` catch that specific error (scoped to the OpenRouter backend, matching both 400 and 404) and retry once with `relaxed_privacy=True`.

The relaxed retry drops only the `require_parameters` capability guard — it **preserves** the deny-by-default `data_collection` policy, so prompts are never silently rerouted to a training host. Relaxing `data_collection` requires an explicit `OPENROUTER_DATA_COLLECTION` override (e.g. `allow`). The retry logs an actionable warning pointing at `OPENROUTER_PROVIDER_ORDER` / `model@provider` for pinning a no-training host.

## Changes

- **`extra_body()`**: Remove the `"reasoning" not in body` guard on `data_collection` and `require_parameters`. Add `relaxed_privacy: bool = False` parameter for the fallback path, which drops `require_parameters` only (deny-by-default `data_collection` is preserved).
- **`_is_openrouter_no_endpoints_error()`**: New helper — distinguishes OpenRouter's "No endpoints found" / "No providers" 4xx (400 or 404) from a genuine model-not-found or other error.
- **`chat()` / `stream()`**: Wrap the API call with a try/except that retries once with `relaxed_privacy=True` on a "no endpoints" 404, scoped to the OpenRouter backend, and logs an actionable warning.
- **Tests**: Updated assertions that expected the relaxed fallback to drop `data_collection` (it now keeps deny by default). Added `TestOpenRouterNoEndpointsError` for the helper and caller-level `TestOpenRouterPrivacyFallbackRetry` covering the `chat()` and `stream()` retry wiring (strict-first, relaxed-second; deny preserved; non-endpoint errors re-raised without retry).

## Test results

```
179 passed in ~14s  (tests/test_llm_openai.py + tests/test_llm_openai_reasoning_effort.py)
```

### DIFF
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -1199,12 +1199,14 @@ def chat(
         if _uses_openrouter_backend(
             provider, model_meta
         ) and _is_openrouter_no_endpoints_error(_e):
+            _dc_configured = get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")
             logger.warning(
                 "OpenRouter: no endpoints matched the strict constraints "
-                "(require_parameters=True + data_collection=deny) for %s — "
+                "(require_parameters=True + data_collection=%s) for %s — "
                 "retrying with only the capability guard dropped (data_collection "
                 "stays at its configured default). Set OPENROUTER_PROVIDER_ORDER "
                 "or use model@provider to pin a no-training host.",
+                _dc_configured,
                 model_meta.model,
             )
             raw_response = _chat_create(relaxed_privacy=True)
@@ -1630,12 +1632,14 @@ def stream(
         if _uses_openrouter_backend(
             provider, model_meta
         ) and _is_openrouter_no_endpoints_error(_e):
+            _dc_configured = get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")
             logger.warning(
                 "OpenRouter: no endpoints matched the strict constraints "
-                "(require_parameters=True + data_collection=deny) for %s — "
+                "(require_parameters=True + data_collection=%s) for %s — "
                 "retrying with only the capability guard dropped (data_collection "
                 "stays at its configured default). Set OPENROUTER_PROVIDER_ORDER "
                 "or use model@provider to pin a no-training host.",
+                _dc_configured,
                 model_meta.model,
             )
             _stream_obj = _stream_create(relaxed_privacy=True)
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -1370,7 +1370,7 @@ def extra_body(
         # Chat Completions parameter (OpenAI reasoning models, Kimi K3).
         # The Responses API path sets ``reasoning.effort`` in chat()/stream().
         body["reasoning_effort"] = effort
-    if provider == "openrouter":
+    if _uses_openrouter_backend(provider, model_meta):
         # Enable detailed usage info including cached tokens
         # See: https://openrouter.ai/docs/guides/usage-accounting
         body["usage"] = {"include": True}
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -1196,13 +1196,15 @@ def chat(
     try:
         raw_response = _chat_create()
     except Exception as _e:
-        if _is_openrouter_no_endpoints_error(_e):
+        if _uses_openrouter_backend(
+            provider, model_meta
+        ) and _is_openrouter_no_endpoints_error(_e):
             logger.warning(
-                "OpenRouter: no endpoints matched privacy constraints "
-                "(data_collection=deny + require_parameters) for %s — "
-                "retrying without privacy constraints. "
-                "Set OPENROUTER_PROVIDER_ORDER or use model@provider to pin "
-                "a no-training host and restore the default privacy guarantees.",
+                "OpenRouter: no endpoints matched the strict constraints "
+                "(require_parameters=True + data_collection=deny) for %s — "
+                "retrying with only the capability guard dropped (data_collection "
+                "stays at its configured default). Set OPENROUTER_PROVIDER_ORDER "
+                "or use model@provider to pin a no-training host.",
                 model_meta.model,
             )
             raw_response = _chat_create(relaxed_privacy=True)
@@ -1319,16 +1321,19 @@ def _resolve_reasoning_effort(provider: Provider, model_meta: ModelMeta) -> str 
 
 
 def _is_openrouter_no_endpoints_error(e: Exception) -> bool:
-    """Return True when OpenRouter returns 404 because no provider matched the constraints.
+    """Return True when OpenRouter reports no provider matched the constraints.
 
-    OpenRouter returns a 404 with "No endpoints found" in the body when the
-    combination of provider preferences (data_collection, require_parameters,
-    provider order, etc.) eliminates every available host.  This is distinct
-    from a genuine model-not-found 404.
+    OpenRouter returns a 4xx (404 "No endpoints found", or 400 in some routing
+    configurations) when the combination of provider preferences
+    (data_collection, require_parameters, provider order, etc.) eliminates every
+    available host.  Matching both 404 and 400 covers the range documented in
+    the original code comment (the triple constraint "eliminates all available
+    providers and causes 400 errors").  The message check keeps this distinct
+    from a genuine model-not-found or other 4xx error.
     """
     from openai import APIStatusError  # fmt: skip
 
-    if not isinstance(e, APIStatusError) or e.status_code != 404:
+    if not isinstance(e, APIStatusError) or e.status_code not in (400, 404):
         return False
     error_text = " ".join(
         [
@@ -1348,11 +1353,13 @@ def extra_body(
 ) -> dict[str, Any]:
     """Return extra body for the OpenAI API based on the model.
 
-    ``relaxed_privacy=True`` omits ``data_collection`` and ``require_parameters``
-    from the OpenRouter provider preferences.  This is used as a one-shot
-    fallback when the strict defaults cause a 404 "No endpoints found" error —
-    we fail toward privacy (always send the constraints on the first attempt) and
-    only relax them when OpenRouter explicitly tells us no provider survives.
+    ``relaxed_privacy=True`` drops the ``require_parameters`` capability guard
+    (routing to providers that may not support every request parameter) as a
+    one-shot fallback when the strict defaults cause a 404 "No endpoints found"
+    error.  The deny-by-default ``data_collection`` policy is **preserved** even
+    in the relaxed path — prompts are never silently routed to a training host.
+    Relaxing ``data_collection`` requires an explicit
+    ``OPENROUTER_DATA_COLLECTION`` override, which is honoured in both modes.
     """
     body: dict[str, Any] = {}
     _maybe_apply_verbosity(body, model_meta)
@@ -1429,8 +1436,12 @@ def extra_body(
             # and retries once with relaxed_privacy=True (see chat()/stream()).
             data_collection = get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")
         else:
-            # Relaxed fallback: no require_parameters, honour explicit env override only.
-            data_collection = get_config().get_env("OPENROUTER_DATA_COLLECTION")
+            # Relaxed fallback: drop only the require_parameters capability guard.
+            # The deny-by-default data_collection policy is PRESERVED so a retry
+            # can never silently route prompts to a training host.  To genuinely
+            # relax data_collection, the user must set an explicit
+            # OPENROUTER_DATA_COLLECTION override (e.g. "allow"), honoured below.
+            data_collection = get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")
         if data_collection:
             provider_prefs["data_collection"] = data_collection
 
@@ -1616,13 +1627,15 @@ def stream(
     try:
         _stream_obj = _stream_create()
     except Exception as _e:
-        if _is_openrouter_no_endpoints_error(_e):
+        if _uses_openrouter_backend(
+            provider, model_meta
+        ) and _is_openrouter_no_endpoints_error(_e):
             logger.warning(
-                "OpenRouter: no endpoints matched privacy constraints "
-                "(data_collection=deny + require_parameters) for %s — "
-                "retrying without privacy constraints. "
-                "Set OPENROUTER_PROVIDER_ORDER or use model@provider to pin "
-                "a no-training host and restore the default privacy guarantees.",
+                "OpenRouter: no endpoints matched the strict constraints "
+                "(require_parameters=True + data_collection=deny) for %s — "
+                "retrying with only the capability guard dropped (data_collection "
+                "stays at its configured default). Set OPENROUTER_PROVIDER_ORDER "
+                "or use model@provider to pin a no-training host.",
                 model_meta.model,
             )
             _stream_obj = _stream_create(relaxed_privacy=True)
--- a/gptme/llm/llm_openai.py
+++ b/gptme/llm/llm_openai.py
@@ -1179,13 +1179,35 @@ def chat(
     if max_tokens is not None:
         optional_kwargs[_max_tokens_param_name(provider, api_model)] = max_tokens
 
-    raw_response = client.chat.completions.with_raw_response.create(
-        model=api_model.split("@")[0],
-        messages=cast(list, messages_dicts),
-        extra_headers=extra_headers(provider),
-        extra_body=extra_body(provider, model_meta, max_tokens=max_tokens),
-        **optional_kwargs,
-    )
+    def _chat_create(relaxed_privacy: bool = False) -> Any:
+        return client.chat.completions.with_raw_response.create(
+            model=api_model.split("@")[0],
+            messages=cast(list, messages_dicts),
+            extra_headers=extra_headers(provider),
+            extra_body=extra_body(
+                provider,
+                model_meta,
+                max_tokens=max_tokens,
+                relaxed_privacy=relaxed_privacy,
+            ),
+            **optional_kwargs,
+        )
+
+    try:
+        raw_response = _chat_create()
+    except Exception as _e:
+        if _is_openrouter_no_endpoints_error(_e):
+            logger.warning(
+                "OpenRouter: no endpoints matched privacy constraints "
+                "(data_collection=deny + require_parameters) for %s — "
+                "retrying without privacy constraints. "
+                "Set OPENROUTER_PROVIDER_ORDER or use model@provider to pin "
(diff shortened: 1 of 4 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `gptme/llm/llm_openai.py` (+110/-31, Δ141)
- `tests/test_llm_openai.py` (+105/-18, Δ123)
- `tests/test_llm_openai_reasoning_effort.py` (+18/-1, Δ19)

### Suggested verification
- `uv run pytest tests/test_llm_openai.py tests/test_llm_openai_reasoning_effort.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3788._

TimeToBuildBob:
## 🤖 AI code review

This PR changes OpenRouter request construction so reasoning models also send data_collection=deny and require_parameters=True by default (privacy fail-closed). It adds a retry in chat() and stream() that, on an OpenRouter "no endpoints"/"no providers" 4xx, retries once with relaxed_privacy=True which drops only require_parameters. It also adds tests for the new behavior, updates CI to clear stale apt caches, and adjusts existing tests that previously expected reasoning models to skip the privacy constraints.

Confidence 5/5

> ⚠️ **2 of 4 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

- `.github/workflows/build.yml` — Removes google/chrome apt sources and clears /var/lib/apt/lists before apt-get update to avoid stale-cache hash mismatches.
- `gptme/llm/llm_openai.py` — Adds _is_openrouter_no_endpoints_error helper, relaxed_privacy parameter to extra_body, and try/except retry wrappers in chat() and stream().
- `tests/test_llm_openai.py` — Updates reasoning-model privacy assertions and adds TestOpenRouterNoEndpointsError and TestOpenRouterPrivacyFallbackRetry classes.
- `tests/test_llm_openai_reasoning_effort.py` — Adds assertions that privacy constraints (require_parameters and data_collection) apply in the reasoning-effort path.

</details>

| `736ebc237ca7` | 2/5 | 2 | llm | 2026-09-09 17:20 UTC |
| `a1cbfb0c1d86` | 3/5 | 2 | llm | 2026-09-09 18:40 UTC |
| `b6a42d2dff8f` | 2/5 | 2 | llm | 2026-09-09 19:41 UTC |

</details>

<sub>Reviewed `db564cbdfece` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 246s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/llm/llm_openai.py
```diff
@@ -1296,10 +1318,42 @@ def _resolve_reasoning_effort(provider: Provider, model_meta: ModelMeta) -> str
...
+    """Return True when OpenRouter returns 404 because no provider matched the constraints.
+
+    OpenRouter returns a 404 with "No endpoints found" in the body when the
+    combination of provider preferences (data_collection, require_parameters,
+    provider order, etc.) eliminates every available host.  This is distinct
+    from a genuine model-not-found 404.
+    """
+    from openai import APIStatusError  # fmt: skip
+
+    if not isinstance(e, APIStatusError) or e.status_code != 404:
```
TimeToBuildBob: ❌ **P1** — The fallback only triggers on a 404 status code, but the pre-change comment in this same function documented that the triple constraint (require_parameters + reasoning + data_collection=deny) 'eliminates all available providers and causes 400 errors.' If OpenRouter still returns a 400 status for models with no privacy-respecting host, _is_openrouter_no_endpoints_error() returns False, the retry with relaxed_privacy=True never runs, and reasoning models that previously worked (because constraints were omitted) now fail with a 400. The user sees a hard error instead of the intended automatic fallback, and the new tests only exercise the 404 case, so this regression would go undetected.

gptme/llm/llm_openai.py
```diff
@@ -1179,13 +1179,35 @@ def chat(
...
+        if _is_openrouter_no_endpoints_error(_e):
+            logger.warning(
+                "OpenRouter: no endpoints matched privacy constraints "
+                "(data_collection=deny + require_parameters) for %s — "
+                "retrying without privacy constraints. "
+                "Set OPENROUTER_PROVIDER_ORDER or use model@provider to pin "
+                "a no-training host and restore the default privacy guarantees.",
+                model_meta.model,
+            )
+            raw_response = _chat_create(relaxed_privacy=True)
```
TimeToBuildBob, line 1212: ❌ **P1** — The fallback added in chat() and stream() retries with relaxed_privacy=True when OpenRouter returns a 404 'No endpoints found' for a reasoning model. This silently removes data_collection=deny and require_parameters from the provider preferences, sending the user's prompt to a provider that may train on it, with only a log warning. This directly contradicts the PR title's promise to always send data_collection=deny for reasoning models and violates user privacy expectations: the request that actually reaches the model may be processed by a training-enabled host. The code should abort with a clear error rather than degrade privacy automatically, unless the user explicitly opts into a fallback.

gptme/llm/llm_openai.py
TimeToBuildBob: Fixed in 4a7089ecc. `_is_openrouter_no_endpoints_error()` now matches **both 400 and 404** carrying the "No endpoints"/"No providers" signal, covering the 400 behavior the pre-change comment documented. Tests updated: `test_matches_no_endpoints_on_400` asserts a 400 with the message now matches; `test_does_not_match_other_400` guards the unrelated-message case. The relaxed fallback (deny preserved) will now fire on either status.

gptme/llm/llm_openai.py
TimeToBuildBob: Fixed in 4a7089ecc. The relaxed fallback no longer drops `data_collection=deny` — it now only removes the `require_parameters` capability guard. The deny-by-default policy is preserved in the relaxed retry, so prompts can never be silently rerouted to a training host; relaxing `data_collection` requires an explicit `OPENROUTER_DATA_COLLECTION` override (e.g. `allow`). The retry is also scoped to the OpenRouter backend via `_uses_openrouter_backend`. Caller-level tests (`TestOpenRouterPrivacyFallbackRetry`) assert deny is kept in both the strict and relaxed requests for `chat()` and `stream()`.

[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4a7089ecc. The relaxed retry now **preserves** the deny-by-default `data_collection=deny` — it only drops the `require_parameters` capability guard. Prompts can never be silently rerouted to a training host; relaxing `data_collection` requires an explicit `OPENROUTER_DATA_COLLECTION` override. Both the `chat()` and `stream()` paths use this behaviour.

gptme/llm/llm_openai.py
```diff
@@ -1179,13 +1179,35 @@ def chat(
...
+                model_meta,
+                max_tokens=max_tokens,
+                relaxed_privacy=relaxed_privacy,
+            ),
+            **optional_kwargs,
+        )
+
+    try:
+        raw_response = _chat_create()
+    except Exception as _e:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4a7089ecc. The retry is now gated on `_uses_openrouter_backend(provider, model_meta)`, so a non-OpenRouter gateway 404 carrying a "no endpoints"/"no providers" message no longer triggers the OpenRouter-specific retry or its warning. Applied in both `chat()` and `stream()`.

tests/test_llm_openai.py
```diff
@@ -2808,6 +2831,70 @@ def test_openrouter_quantization_gptme_prefixed_env(self, monkeypatch):
...
+    from openai import APIStatusError
+
+    response = httpx.Response(
+        status_code,
+        request=httpx.Request("POST", "https://openrouter.ai/api/v1/chat/completions"),
+    )
+    return APIStatusError(message, response=response, body=body)
+
+
+class TestOpenRouterNoEndpointsError:
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4a7089ecc. Added caller-level `TestOpenRouterPrivacyFallbackRetry` covering both paths: `test_chat_retries_relaxed_on_no_endpoints_and_keeps_deny`, `test_stream_retries_relaxed_on_no_endpoints_and_keeps_deny` (strict-first/relaxed-second, deny preserved in both requests), and `test_chat_reraises_non_endpoint_error` (a non-endpoint 404 is re-raised without retry).

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Fixes pushed (4a7089ecc) addressing review findings

Responded to and resolved the Greptile (3/5) and AI-review (2/5) findings on the original head:

**P1 — Privacy fallback no longer drops deny.** The `relaxed_privacy=True` retry previously omitted the deny-by-default `data_collection`, which could silently resend prompts to a training host. It now only drops the `require_parameters` capability guard; `data_collection=deny` is preserved in the relaxed retry. Relaxing `data_collection` requires an explicit `OPENROUTER_DATA_COLLECTION` override.

**P1 — 400-vs-404 fallback gap.** `_is_openrouter_no_endpoints_error()` now matches both **400** and **404** carrying the "No endpoints"/"No providers" signal (the pre-change comment documented the triple constraint causing 400s), so the fallback fires on either status.

**P2 — Retry scoped to OpenRouter backend.** Both `chat()` and `stream()` gate the retry on `_uses_openrouter_backend()`, so a non-OpenRouter gateway 404 no longer triggers the misleading OpenRouter retry/warning.

**P2 — Caller-level tests added.** New `TestOpenRouterPrivacyFallbackRetry` exercises the `chat()` and `stream()` retry wiring (strict-first/relaxed-second, deny preserved in both requests) and the re-raise of non-endpoint errors.

Tests: 179 passed. CI running; Greptile re-review triggered on the new head.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/llm/llm_openai.py
```diff
@@ -1179,13 +1179,37 @@ def chat(
...
+        if _uses_openrouter_backend(
+            provider, model_meta
+        ) and _is_openrouter_no_endpoints_error(_e):
+            logger.warning(
+                "OpenRouter: no endpoints matched the strict constraints "
+                "(require_parameters=True + data_collection=deny) for %s — "
```
TimeToBuildBob: ⚠️ **P2** — The warning logged in the retry path hardcodes the strings "require_parameters=True + data_collection=deny" and "data_collection stays at its configured default", but those strings are only accurate when OPENROUTER_DATA_COLLECTION is unset. If the user has set OPENROUTER_DATA_COLLECTION=allow, the strict request that failed actually sent data_collection=allow, not deny; the fallback preserves allow. The log therefore reports deny to the user even though the request used allow, and claims the policy stayed at its configured default when it did not. A user who explicitly allowed training would be told prompts were kept off training pipelines, which is misleading and obscures the actual privacy posture.

gptme/llm/llm_openai.py
TimeToBuildBob: False positive — `_is_openrouter_no_endpoints_error` already checks `e.status_code not in (400, 404)` (line 1336), so both the 400 and 404 cases are covered. The docstring on the function explicitly notes this. No code change needed.

gptme/llm/llm_openai.py
TimeToBuildBob: False positive — the relaxed path (lines 1438-1444) still calls `get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")`, identical to the strict path. `data_collection` is set to the same configured value in both branches. Only `require_parameters` is dropped in the fallback, which is exactly what the docstring documents. No code change needed.

gptme/llm/llm_openai.py
TimeToBuildBob: Valid finding. Fixed in 4adae56b4 — both `chat()` and `stream()` now compute `_dc_configured = get_config().get_env("OPENROUTER_DATA_COLLECTION", "deny")` at the point of the warning and use `%s` / `_dc_configured` instead of the hardcoded `deny` string. The warning now accurately reflects the actual constraint that failed, even when the user has set `OPENROUTER_DATA_COLLECTION=allow`.

tests/test_llm_openai.py
```diff
@@ -4,7 +4,7 @@
 from unittest.mock import MagicMock, Mock, patch
 
 import openai  # warm import cache before per-test 10s timeout
-import openai._types  # noqa: F401  # warm openai._types (NOT_GIVEN) before per-test 10s timeout
+import openai._types  # warm openai._types (NOT_GIVEN) before per-test 10s timeout
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b6a42d2df — added `# noqa: F401` to the side-effect import at `tests/test_llm_openai.py:7`. The comment already explained the intent (warm `openai._types` before per-test timeout); the suppressor was just missing.

TimeToBuildBob:
@greptileai review

gptme/llm/llm_openai.py
```diff
@@ -1179,13 +1179,39 @@ def chat(
...
+            logger.warning(
+                "OpenRouter: no endpoints matched the strict constraints "
+                "(require_parameters=True + data_collection=%s) for %s — "
+                "retrying with only the capability guard dropped (data_collection "
+                "stays at its configured default). Set OPENROUTER_PROVIDER_ORDER "
+                "or use model@provider to pin a no-training host.",
+                _dc_configured,
+                model_meta.model,
+            )
+            raw_response = _chat_create(relaxed_privacy=True)
```
TimeToBuildBob, line 1212: ❌ **P1** — The new one-shot fallback in chat()/stream() retries with relaxed_privacy=True when the initial strict request hits a 'no endpoints' 4xx. However, that fallback runs inside functions that are themselves wrapped by the module-level retry decorators @retry_on_openai_error() and @retry_generator_on_openai_error(). If the first strict call fails with a no-endpoints error (non-transient) and the subsequent relaxed call then fails with a transient error such as a 429 rate limit, the exception escapes to the outer decorator, which treats it as a retryable condition and re-invokes the entire function from the top. That repeats the strict call (which will again fail with no-endpoints) and then the relaxed call again. With the default max_retries of 3, a persistent rate limit on the relaxed path can generate up to 3×2 = 6 API requests instead of the intended 2, inflating cost and latency and potentially worsening the rate-limit condition. The fallback is documented as a one-shot retry, but the interaction with the outer retry defeats that guarantee. This design also means a relaxed call that hits a transient error will always re-attempt the strict call, which is known to fail, wasting cycles. A cleaner approach is to move the no-endpoints fallback outside the retry-decorated function (e.g., wrap only the strict call) or to suppress the outer retry when the error originates from the relaxed path.

TimeToBuildBob: Valid interaction point. The outer `@retry_on_openai_error` / `@retry_generator_on_openai_error` decorators will, in theory, re-invoke the strict call on a subsequent transient error (429/5xx) from the relaxed path, producing up to max_retries×2 total calls instead of 2. In practice this requires a simultaneous 'no endpoints for strict' + transient-rate-limit-on-relaxed storm, which is low probability. When it does occur the extra calls are bounded (3×2=6) and the outer decorator backs off before each retry, so there is no infinite loop risk. The 'one-shot fallback' guarantee is best-effort under rate-limit pressure, not absolute — accepting this tradeoff and noting it in the function docstring rather than restructuring the call stack for a rare edge case.

TimeToBuildBob:
Fixed in db564cbdf — `extra_body()` now uses `_uses_openrouter_backend(provider, model_meta)` as the condition (same predicate as the retry gate), so `provider == "gptme"` models with `openrouter/` model names get the full provider prefs block including `data_collection` and `require_parameters`. The relaxed retry is now effective for both the direct-openrouter and gptme-provider-openrouter-backend paths.

TimeToBuildBob:
@greptileai review

PR #3790, opened 2026-09-09
fix(memory): read index entries under the root lock

TimeToBuildBob:
When `memory index --write` waits for the root lock, a concurrent `save` or `supersede` can finish after its entries were read. The index writer then publishes the stale snapshot, hiding the saved entry or restoring a superseded entry in MEMORY.md.

Move entry reading and rendering inside the existing root lock so generation sees mutations completed before it acquires the lock. Two event-synchronized regressions exercise actual save and supersede operations in that window; both fail before the fix.

Validation: `tests/test_memory_store.py` and `tests/test_cc_memory.py`: 101 passed, 1 skipped. Commit hooks passed. Related to gptme/gptme#3734; this fixes one concurrency prerequisite and does not complete the broader memory migration.

PR #3791, opened 2026-09-09
feat(memory): persist always-on selection and byte budget

TimeToBuildBob:
A byte-capped generated memory index can drop curated operator guidance, and a later `save` or `supersede` can ignore that cap entirely. This adds an opt-in `.memory-index.json` per root: an explicit filename selection and a persistent UTF-8 byte budget shared by index print/write/check, save, supersede, and the compatibility index helper.

Selected entries must all fit; overflow or invalid selection fails before mutation. Unselected living entries remain recallable, new saves stay unselected, and supersession transfers selection to the replacement. Saves preserve lifecycle/provenance and omitted title/type/metadata. Readers lock the complete entries/policy snapshot, and write failures restore the entry/policy/index files. Roots without a policy keep their existing index behavior.

The branch includes gptme/gptme#3790, which fixes the existing index writer lock boundary; land that prerequisite first. The PR targets master because test/lint workflows exclude feature-branch bases. Once #3790 lands, its commit drops out of the diff. This is the selection/budget prerequisite for gptme/gptme#3734. It does not perform a live memory migration. The pending layered-prompt integration in gptme/gptme#3789 also needs to consume selected views before that cutover; its current direct `render_index(store.entries())` call bypasses root policy.

Validation: 136 focused tests passed, 1 skipped; Ruff, mypy, and all commit hooks passed. The overflow regressions fail against the parent. An isolated 422-entry corpus preserves 100/100 curated links in 20,230 bytes under a 21,000-byte policy; subsequent CLI save/supersede/index-check operations preserve selection and budget. Wording review and external writer convergence remain migration gates.

Review follow-up: rollback now restores original bytes via atomic replacement, including read-only entries and CRLF content. Existing malformed frontmatter remains readable but must be corrected before save; this intentional mutation boundary is documented and regression-tested.

### DIFF
--- a/gptme/memory/policy.py
+++ b/gptme/memory/policy.py
@@ -10,10 +10,11 @@ from .schema import MemoryEntry, is_index_file
 
 POLICY_FILENAME = ".memory-index.json"
 
-# Minimum budget: the rendered header + blank line ("# Persistent Memory\n\n")
-# that appears even when no entries are selected.  A budget below this makes
-# every index operation fail immediately.
-POLICY_MIN_BUDGET = 22
+# Minimum budget: the rendered header ("# Persistent Memory\n") that appears
+# even when no entries are selected.  The trailing blank line is stripped when
+# the selection is empty, so the actual minimum is 20 bytes, not 22.  A budget
+# below this makes every index operation fail immediately.
+POLICY_MIN_BUDGET = 20
 
 
 @dataclass(frozen=True)
--- a/gptme/memory/policy.py
+++ b/gptme/memory/policy.py
@@ -10,6 +10,11 @@ from .schema import MemoryEntry, is_index_file
 
 POLICY_FILENAME = ".memory-index.json"
 
+# Minimum budget: the rendered header + blank line ("# Persistent Memory\n\n")
+# that appears even when no entries are selected.  A budget below this makes
+# every index operation fail immediately.
+POLICY_MIN_BUDGET = 22
+
 
 @dataclass(frozen=True)
 class IndexPolicy:
@@ -34,10 +39,16 @@ class IndexPolicy:
         budget = data["budget"]
         if type(budget) is not int or budget <= 0:
             raise ValueError("index policy budget must be a positive integer")
+        if budget < POLICY_MIN_BUDGET:
+            raise ValueError(
+                f"index policy budget must be at least {POLICY_MIN_BUDGET} bytes "
+                "(the index header alone requires that many bytes)"
+            )
         selected = data["selected"]
         if not isinstance(selected, list) or any(
             not isinstance(name, str)
             or not name.endswith(".md")
+            or name.startswith(".")
             or "/" in name
             or "\\" in name
             or is_index_file(Path(name))
--- a/gptme/memory/policy.py
+++ b/gptme/memory/policy.py
@@ -1,0 +1,78 @@
+"""Opt-in, root-local policy for the always-on view, independent of recall."""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass, replace
+from pathlib import Path
+
+from .schema import MemoryEntry, is_index_file
+
+POLICY_FILENAME = ".memory-index.json"
+
+
+@dataclass(frozen=True)
+class IndexPolicy:
+    budget: int
+    selected: tuple[str, ...]
+
+    @classmethod
+    def read(cls, root: Path) -> IndexPolicy | None:
+        """Missing policy keeps legacy behavior; malformed policy never does."""
+        try:
+            text = (root / POLICY_FILENAME).read_text(encoding="utf-8")
+        except FileNotFoundError:
+            return None
+        try:
+            data = json.loads(text)
+        except ValueError as exc:
+            raise ValueError(f"invalid {POLICY_FILENAME}: {exc}") from exc
+        if not isinstance(data, dict) or set(data) != {"version", "budget", "selected"}:
+            raise ValueError(f"{POLICY_FILENAME} requires version, budget, selected")
+        if type(data["version"]) is not int or data["version"] != 1:
+            raise ValueError(f"unsupported {POLICY_FILENAME} version")
+        budget = data["budget"]
+        if type(budget) is not int or budget <= 0:
+            raise ValueError("index policy budget must be a positive integer")
+        selected = data["selected"]
+        if not isinstance(selected, list) or any(
+            not isinstance(name, str)
+            or not name.endswith(".md")
+            or "/" in name
+            or "\\" in name
+            or is_index_file(Path(name))
+            for name in selected
+        ):
+            raise ValueError("index policy selected must contain local entry filenames")
+        if len(set(selected)) != len(selected):
+            raise ValueError("index policy selected contains duplicate filenames")
+        return cls(budget, tuple(selected))
+
+    def to_json(self) -> str:
+        return (
+            json.dumps(
+                {"version": 1, "budget": self.budget, "selected": list(self.selected)},
+                indent=2,
+                ensure_ascii=False,
+            )
+            + "\n"
+        )
+
+    def supersede(self, old: str, new: str) -> IndexPolicy:
+        """Transfer selection, retaining only the first occurrence of a filename."""
+        return replace(
+            self,
+            selected=tuple(
+                dict.fromkeys(new if f == old else f for f in self.selected)
+            ),
+        )
+
+    def entries(self, entries: list[MemoryEntry]) -> list[MemoryEntry]:
+        by_file = {entry.filename: entry for entry in entries}
+        selected = []
+        for name in self.selected:
+            entry = by_file.get(name)
+            if entry is None or not entry.is_living:
+                raise ValueError(f"selected memory {name!r} is missing or not living")
+            selected.append(entry)
+        return selected
--- a/gptme/memory/store.py
+++ b/gptme/memory/store.py
@@ -86,10 +86,11 @@ def _commit_replacements(pairs: list[tuple[Path, str]]) -> None:
             dest.parent.mkdir(parents=True, exist_ok=True)
             tmp = dest.with_name(f".{dest.name}.{os.getpid()}.tmp")
             fd = os.open(tmp, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
+            # Track the temp before writing so a partial-write failure is cleaned up.
+            staged.append((dest, tmp, original))
             with os.fdopen(fd, "wb") as f:
                 f.write(text.encode("utf-8"))
             _preserve_mode(tmp, dest)
-            staged.append((dest, tmp, original))
         for dest, tmp, original in staged:
             os.replace(tmp, dest)
             replaced.append((dest, original))
--- a/gptme/memory/store.py
+++ b/gptme/memory/store.py
@@ -348,7 +348,7 @@ class MemoryStore:
         with _locked_root(root.path):
             policy = IndexPolicy.read(root.path)
             previous = (
-                parse_entry(path, scope=root.scope, strict=True)
+                parse_entry(path, scope=root.scope, strict=False)
                 if path.exists()
                 else MemoryEntry(name=slug, path=path, scope=root.scope)
             )
--- a/gptme/memory/store.py
+++ b/gptme/memory/store.py
@@ -15,9 +15,9 @@ if TYPE_CHECKING:
     from collections.abc import Iterable
     from pathlib import Path
 
+from .policy import POLICY_FILENAME, IndexPolicy
 from .roots import MemoryRoot, default_write_root, resolve_roots
 from .schema import (
-    DEFAULT_TYPE,
     TYPE_ORDER,
     MemoryEntry,
     MemoryFrontmatterError,
@@ -181,6 +181,22 @@ def _locked_root(root_path: Path):
 
 
 def update_index_line(memory_dir: Path, entry: MemoryEntry) -> None:
+    """Update a legacy pointer, or regenerate a policy-managed root's view.
+
+    In managed roots, entry files remain the source of truth. This compatibility
+    helper cannot append an unselected pointer or bypass the persistent budget.
+    """
+    with _locked_root(memory_dir):
+        policy = IndexPolicy.read(memory_dir)
+        if policy is None:
+            _upsert_index_line(memory_dir, entry)
+        else:
+            store = MemoryStore([MemoryRoot("explicit", memory_dir)])
+            text = store._render_with_policy(store.index_entries(), policy)
+            _atomic_write(store.index_path(), text)
+
+
+def _upsert_index_line(memory_dir: Path, entry: MemoryEntry) -> None:
     """Upsert one ``- [title](file) — description`` line in ``MEMORY.md``.
 
     This is the Claude Code convention: append a pointer line, never rewrite
@@ -240,7 +256,9 @@ class MemoryStore:
         available = ", ".join(r.scope for r in self.roots) or "none"
         raise KeyError(f"no memory root for scope {scope!r} (available: {available})")
 
-    def _entries_from_roots(self, roots: Iterable[MemoryRoot]) -> list[MemoryEntry]:
+    def _entries_from_roots(
+        self, roots: Iterable[MemoryRoot], *, deduplicate: bool = True
+    ) -> list[MemoryEntry]:
         self.errors = []
         seen: dict[str, MemoryEntry] = {}
         for r in roots:
@@ -254,7 +272,10 @@ class MemoryStore:
                 except (MemoryParseError, OSError, UnicodeDecodeError) as e:
                     self.errors.append((path, str(e)))
                     continue
-                seen.setdefault(entry.name, entry)  # nearest root wins
+                # Layered retrieval shadows by name; a physical index must
+                # retain every filename, including entries with duplicate names.
+                key = entry.name if deduplicate else str(path)
+                seen.setdefault(key, entry)
         return list(seen.values())
 
     def entries(
@@ -282,7 +303,7 @@ class MemoryStore:
         would put later-directory filenames into the first directory's
         ``MEMORY.md``. Index generation is always per-directory.
         """
-        return self._entries_from_roots([self.root(scope)])
+        return self._entries_from_roots([self.root(scope)], deduplicate=False)
 
     def get(self, name: str, scope: str | None = None) -> MemoryEntry | None:
         wanted = {name, slugify(name)}
@@ -299,12 +320,16 @@ class MemoryStore:
         description: str,
         body: str = "",
         *,
-        type: str = DEFAULT_TYPE,
+        type: str | None = None,
         scope: str | None = None,
         title: str | None = None,
         metadata: dict[str, Any] | None = None,
     ) -> Path:
-        """Write ``<slug>.md`` into the scope's root and upsert its index line.
+        """Write an entry, preserving existing lifecycle and omitted metadata.
+
+        A policy-managed root regenerates its selected view within the stored
+        budget, committing entry and index together. New entries remain available
+        to recall but are not selected automatically. Legacy roots upsert a line.
 
         The root lock serialises this write against concurrent ``supersede`` calls.
         ``supersede`` replaces ``MEMORY.md`` via ``os.replace``, which would drop
@@ -319,21 +344,36 @@ class MemoryStore:
             )
         root = self.root(scope)
         root.path.mkdir(parents=True, exist_ok=True)
-        entry = MemoryEntry(
-            name=slug,
-            description=description.strip(),
-            type=type,
-            body=body,
-            title=title,
-            metadata=dict(metadata or {}),
-            scope=root.scope,
-        )
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
+            previous = (
+                parse_entry(path, scope=root.scope, strict=True)
+                if path.exists()
+                else MemoryEntry(name=slug, path=path, scope=root.scope)
+            )
+            entry = replace(
+                previous,
+                description=description.strip(),
+                body=body,
+                type=type if type is not None else previous.type,
+                title=title if title is not None else previous.title,
+                metadata={**previous.metadata, **(metadata or {})},
+            )
+            text = entry.to_markdown()
+            entry = entry_from_text(text, path=path, scope=root.scope, strict=True)
+            if policy is None:
+                _atomic_write(path, text)
+                _upsert_index_line(root.path, entry)
+            else:
+                entries = [e for e in self.index_entries(scope) if e.path != path]
+                entries.append(entry)
+                index_text = self._render_with_policy(entries, policy)
+                _commit_replacements(
+                    [(path, text), (self.index_path(scope), index_text)]
+                )
         logger.debug("saved memory %s to %s", entry.name, entry.path)
-        return entry.path
+        return path
 
     def supersede(
         self, old_name: str, new_name: str, *, scope: str | None = None
@@ -348,6 +388,7 @@ class MemoryStore:
         """
         root = self.root(scope)
         with _locked_root(root.path):
+            policy = IndexPolicy.read(root.path)
             old = self.get(old_name, scope=root.scope)
             new = self.get(new_name, scope=root.scope)
             if old is None:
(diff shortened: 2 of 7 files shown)


gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
-            scope=root.scope,
-        )
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
+            previous = (
+                parse_entry(path, scope=root.scope, strict=True)
```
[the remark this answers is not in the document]
TimeToBuildBob: This is intentional now that save preserves the existing record instead of replacing it wholesale. The lenient parser is suitable for reads but drops malformed/nested fields, so using it for mutation would silently lose lifecycle or provenance. The docs now explicitly state that existing frontmatter must be corrected before saving. Added managed- and legacy-root regressions proving malformed entries remain readable and save refuses without changing any bytes. Included in 0bb514bc8; 136 focused tests pass (1 skipped).

gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
+            if policy is None:
+                _atomic_write(path, text)
+                _upsert_index_line(root.path, entry)
+            else:
+                entries = [e for e in self.index_entries(scope) if e.path != path]
+                entries.append(entry)
+                index_text = self._render_with_policy(entries, policy)
+                _commit_replacements(
+                    [(path, text), (self.index_path(scope), index_text)]
+                )
```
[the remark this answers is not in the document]
TimeToBuildBob: Reproduced under an unprivileged POSIX user with a 0444 entry: index publication failed and the changed entry survived. Fixed in 0bb514bc8 by snapshotting original bytes and restoring through atomic replacement, preserving both the read-only mode and CRLF bytes. Added LF/CRLF regressions; both fail before the fix and pass afterward. The full memory/CC/policy slice passes: 136 tests, 1 skipped.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
-            scope=root.scope,
-        )
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
+            previous = (
+                parse_entry(path, scope=root.scope, strict=False)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 172c40a14: reverted to strict=True for existing-entry read in save(). The lenient fallback's documented data loss (lists and deeper nesting silently dropped) outweighs the UX cost of refusing malformed saves. Users must repair frontmatter with `gptme memory audit` before re-saving — this boundary was previously accepted and is now documented in docs/memory.rst and the updated test.

gptme/memory/store.py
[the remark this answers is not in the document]
TimeToBuildBob: This was the original behavior — and it's the correct trade-off. The lenient fallback (_lenient_load) silently drops list values (like 'supersedes', 'keywords') and deeper nesting, so allowing lenient saves can corrupt lifecycle/provenance data without any warning. Refusing malformed saves (MemoryFrontmatterError) is documented in docs/memory.rst and is safer. Users fix entries via 'gptme memory audit'.

gptme/memory/store.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in commit 0bb514bc8 ('fix(memory): atomically restore read-only entries on rollback') — _commit_replacements now restores files via _atomic_write (os.replace) rather than direct write_text, so rollback handles read-only entries and preserves original bytes. The PR description notes this was the first rollback fix in this branch.

TimeToBuildBob:
## 🤖 AI code review

Adds an opt-in per-root `.memory-index.json` policy (filename selection plus a persistent UTF-8 byte budget) and makes `index` print/write/check, `save`, `supersede`, and the `update_index_line` helper render only the selected living entries, failing before any mutation on overflow or invalid selection. `save` now preserves lifecycle/provenance and omitted title/type/metadata, `supersede` transfers selection and commits entries+policy+index together with byte-level rollback, and `_atomic_write`/`_commit_replacements` were generalized to restore raw bytes. Adds a policy module, new policy tests, and small CLI/tool changes (`--type` default None, catching ValueError).

Confidence 5/5

- `docs/memory.rst` — Documents the `.memory-index.json` policy format, budget semantics, managed-root save/supersede behavior, and migration guidance.
- `gptme/cli/cmd_memory.py` — Changes `--type` default to None, routes `memory index` through `render_root_index`, and adds OSError to the caught exceptions.
- `gptme/memory/policy.py` — New file: `IndexPolicy` dataclass with strict `read`, `to_json`, `supersede` selection transfer, and filename-based `entries` resolution.
- `gptme/memory/store.py` — Adds policy-aware save/supersede with snapshot-and-restore writes, filename-keyed dedup for managed `index_entries`, and locked `render_root_index`/`check_index` with read-only fallback.
- `gptme/tools/memory.py` — Extends the `_do_save` error handler to also catch `ValueError` from budget/selection failures.
- `tests/test_memory_index_policy.py` — New tests covering selection persistence, duplication, rollback, budget overflow, malformed policies, readonly roots, and lock-based reader waiting.
- `tests/test_memory_store.py` — Changes the partial-replace rollback test failure condition from `>= 2` to `== 2`.

</details>

| `9f7ea8345348` | 3/5 | 1 | llm | 2026-09-09 19:00 UTC |
| `1e4e3ca66e32` | 2/5 | 7 | llm | 2026-09-09 19:41 UTC |
| `f4a8fcf2bede` | 3/5 | 3 | llm | 2026-09-09 20:42 UTC |
| `9174b867bbd0` | 5/5 | 0 | llm | 2026-09-10 10:21 UTC |

</details>

<sub>Reviewed `9174b867bbd0` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1133s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/memory/store.py
```diff
@@ -544,16 +592,53 @@ def build(n_entries: int) -> str:
...
+                f"selected memory index needs {size} bytes, exceeds budget {limit}; "
+                "shorten descriptions or revise .memory-index.json before writing"
+            )
+        return text
+
+    def render_root_index(
+        self, scope: str | None = None, *, budget: int | None = None
+    ) -> str:
+        """Render one root using its persistent selection and budget, if present."""
+        with _locked_root(self.root(scope).path):
```
TimeToBuildBob: ❌ **P1** — The read-only index operations `render_root_index` and `check_index` now acquire `_locked_root`, which unconditionally creates the root directory and a `.memory.lock` file via `root_path.mkdir(parents=True, exist_ok=True)`. On a read-only memory root (e.g., a read-only mount or a root owned by another user), this creation raises `OSError` before any read happens, so `gptme-util memory index` and `gptme-util memory index --check` now crash with a permission error instead of printing or verifying the existing index. Previously these operations were pure reads that never touched the filesystem, so read-only roots worked. This breaks the documented read path and also creates unexpected artifacts (directory and lock file) when the root does not yet exist. The lock should be acquired only for operations that can write the root or the lock file should be created in a sidecar location without requiring write permission on the root directory.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
TimeToBuildBob: Fixed in 1e4e3ca66. `render_root_index` and `check_index` now wrap `_locked_root` in `try/except OSError` and fall back to an unlocked read. Read-only roots cannot have concurrent writers, so the TOCTOU race doesn't apply there. Added a regression test (`test_render_and_check_on_readonly_root`) that makes the root dir `0555` and confirms both operations succeed.

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
```diff
@@ -181,6 +181,22 @@ def _locked_root(root_path: Path):
...
+    In managed roots, entry files remain the source of truth. This compatibility
+    helper cannot append an unselected pointer or bypass the persistent budget.
+    """
+    with _locked_root(memory_dir):
+        policy = IndexPolicy.read(memory_dir)
+        if policy is None:
+            _upsert_index_line(memory_dir, entry)
+        else:
+            store = MemoryStore([MemoryRoot("explicit", memory_dir)])
+            text = store._render_with_policy(store.index_entries(), policy)
```
TimeToBuildBob, line 204: ❌ **P1** — update_index_line, when a policy is present, ignores the `entry` argument entirely and regenerates the index from the current files via `store._render_with_policy`. It appends the entry to `store.index_entries()` only if that entry is already in the policy's selected list. Previously, regardless of policy, calling update_index_line on a freshly written entry always upserted a pointer line for it. Harness hooks that write an entry file directly and then call this compatibility helper to publish it to the always-on index will now find the entry silently omitted unless it was manually added to .memory-index.json first. This breaks the documented convention that the helper 'updates a legacy pointer' and is a behavioral regression for direct-file writers.

gptme/memory/store.py
```diff
@@ -544,6 +592,51 @@ def build(n_entries: int) -> str:
...
+        budget: int | None = None,
+    ) -> str:
+        if policy is None:
+            return self.render_index(entries, budget=budget)
+        # Required selection is never truncated. A one-shot CLI budget may
+        # tighten the stored cap, but cannot bypass it.
+        limit = min(policy.budget, budget) if budget is not None else policy.budget
+        text = self.render_index(policy.entries(entries))
+        size = len(text.encode("utf-8"))
+        if size > limit:
```
TimeToBuildBob, line 625: ⚠️ **P2** — When a policy exists, _render_with_policy requires that the rendered selected index fit within the policy budget. But if .memory-index.json's selected list is empty, render_index([]) returns only the header '# Persistent Memory\n\n', which is ~22 bytes. IndexPolicy.read validates only budget > 0, so a user can set budget=10, which passes validation. Every subsequent operation that writes or prints an index — save, supersede, write_index, index --write/--check/print — recomputes this text and raises ValueError ('selected memory index needs 22 bytes, exceeds budget 10'). The root becomes unusable: no new entry can be saved and no index regeneration succeeds, until the policy file is manually edited to raise the budget. Suggest validating in IndexPolicy.read that budget >= fixed header length, or handling an empty selection without a hard failure.

tests/test_memory_index_policy.py
```diff
@@ -0,0 +1,539 @@
...
+    """
+    import stat
+
+    store = _store(tmp_path)
+    store.save("entry", "Description")
+    _policy(tmp_path, ["entry.md"])
+    store.write_index()
+
+    # Make the root read-only so lock-file creation fails.
+    tmp_path.chmod(stat.S_IRUSR | stat.S_IXUSR)
```
TimeToBuildBob, line 605: ⚠️ **P2** — The regression test `test_render_and_check_on_readonly_root` does not actually trigger the fallback behavior it claims to verify. After `store.write_index()` at line 526, the root contains a `.memory.lock` file. The subsequent `tmp_path.chmod(stat.S_IRUSR | stat.S_IXUSR)` makes the directory unwritable, but the existing lock file still has owner write permission, so `_locked_root`'s `open(root_path / LOCK_FILENAME, "a+b")` succeeds (append mode only requires file write permission, not directory write permission). Consequently the `except OSError` fallback in `render_root_index` and `check_index` is never executed, and the test passes even if the fallback logic is broken. The test is vacuous for its stated purpose. To make it exercise the fallback, the lock file must be removed (or made read-only) before chmod so the `open` fails with EACCES.

gptme/memory/policy.py
```diff
@@ -0,0 +1,78 @@
...
+            raise ValueError(f"{POLICY_FILENAME} requires version, budget, selected")
+        if type(data["version"]) is not int or data["version"] != 1:
+            raise ValueError(f"unsupported {POLICY_FILENAME} version")
+        budget = data["budget"]
+        if type(budget) is not int or budget <= 0:
+            raise ValueError("index policy budget must be a positive integer")
+        selected = data["selected"]
+        if not isinstance(selected, list) or any(
+            not isinstance(name, str)
+            or not name.endswith(".md")
```
TimeToBuildBob, line 51: ⚠️ **P2** — `IndexPolicy.read` accepts any filename ending in `.md` that passes the current checks (no path separators, not an index file) but does not reject names that begin with a dot. Meanwhile `MemoryStore._entries_from_roots` discovers entries via `sorted(r.path.glob("*.md"))`, and Python's glob does not match hidden (dot-prefixed) files. A policy that selects a hidden file, e.g. `.secret.md`, is therefore accepted, yet that file is never parsed into the entries list. As a result every operation that renders or validates the index under the policy — `render_root_index`, `write_index`, `check_index`, `save`, and `supersede` — raises `ValueError` stating the selected memory is missing or not living, even though the file exists on disk. The memory root becomes unusable until the policy file is hand-edited to remove the selection. Fix by rejecting dot‑prefixed filenames in `IndexPolicy.read` (or by making the entry glob include hidden files).

gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
+            previous = (
+                parse_entry(path, scope=root.scope, strict=True)
+                if path.exists()
+                else MemoryEntry(name=slug, path=path, scope=root.scope)
+            )
+            entry = replace(
+                previous,
+                description=description.strip(),
+                body=body,
+                type=type if type is not None else previous.type,
```
TimeToBuildBob, line 375: ⚠️ **P2** — The default value of the CLI `--type` option changed from `"general"` to `None`, and store.save now passes that None through to `type=type if type is not None else previous.type`. On a save of an existing entry, omitting --type now preserves the entry's previous type instead of resetting it to general. External scripts and harness hooks that call save (directly or via gptme-util memory save) on an existing entry and then rely on the saved file having general type will silently get a different classification than before. For example, a previously `feedback` entry updated without --type stays `feedback` instead of being turned back into a general note. This is a semantic breaking change to the documented API and the stored index grouping.

gptme/memory/store.py
```diff
@@ -282,7 +303,7 @@ def index_entries(self, scope: str | None = None) -> list[MemoryEntry]:
         would put later-directory filenames into the first directory's
         ``MEMORY.md``. Index generation is always per-directory.
         """
-        return self._entries_from_roots([self.root(scope)])
+        return self._entries_from_roots([self.root(scope)], deduplicate=False)
```
TimeToBuildBob: ⚠️ **P2** — `index_entries` now always calls `_entries_from_roots` with `deduplicate=False`, so every physical `*.md` file in a root is included even when several share the same frontmatter `name`. This changes behavior for roots without a policy, which the docs promise retain legacy behavior. Previously `index_entries` used the default `deduplicate=True` and kept only the lexicographically first file per name, so an index written by the legacy version would have one line per name. After this change, running `index --write` on the same root emits duplicate entries (one per file), and `index --check` reports DRIFT against the old index. Any existing root with duplicate `name` fields—even accidental ones—now produces a different generated index, violating the stated backward-compatibility guarantee and forcing a spurious index rewrite on upgrade.

TimeToBuildBob:
## ⚠️ Likely duplicate: gptme/gptme#3789

This PR appears to substantially overlap with [gptme/gptme#3789](https://github.com/gptme/gptme/pull/3789):
> feat(memory): use layered MemoryStore in workspace prompt

**Reason**: Changed files and diffs suggest overlapping intent. If one PR is a superset or variant of the other, closing this one may save review cycles.

**Not a blocker** — review proceeds normally. If this is a false positive, dismiss this comment and continue.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the actionable findings from the latest AI review (commit 636a52a8b):

**Fixed:**
- **P2 ced4029b06d9** — `IndexPolicy.read` now rejects dot-prefixed filenames (`".hidden.md"`); `glob("*.md")` never matches them so selecting one would make every managed operation crash.
- **P2 6222dc9c186b** — Added `POLICY_MIN_BUDGET = 22` constant and validation in `IndexPolicy.read`; a budget below the header size (22 bytes) now fails at policy creation rather than on every subsequent operation.
- **P2 b71fdb2705cd** — Fixed the readonly-root regression test: the pre-existing `.memory.lock` file was still owner-writable after `chmod(0o550)` (append mode doesn't need dir write), so the `except OSError` fallback was never reached. Now removes the lock file before restricting the directory.
- **P2 1a6b5d0464a7** — `index_entries()` now uses `deduplicate=True` for non-policy roots (restoring original behavior); `deduplicate=False` only for policy-managed roots where the policy filter needs to find entries by filename.
- **P1 f9179556a762** — This is intentional design: `update_index_line` on a policy root regenerates the selected view; an unselected entry is not added to MEMORY.md because doing so would bypass the byte-budget guarantee. Clarified in the docstring.

**Not changed (intentional):**
- **P2 8ee5967abc24** — Omitting `--type` on an existing-entry save now preserves the previous type instead of resetting to `"general"`. This is intentional: the lenient-load path that `save` previously used silently dropped list values (`supersedes`, `keywords`) and nested provenance, making type-preservation cheaper than silently corrupting lifecycle fields. Documented in the updated docstring and tested.

Added 2 new regression tests (38 total; all pass).

gptme/memory/policy.py
```diff
@@ -0,0 +1,89 @@
...
+from pathlib import Path
+
+from .schema import MemoryEntry, is_index_file
+
+POLICY_FILENAME = ".memory-index.json"
+
+# Minimum budget: the rendered header + blank line ("# Persistent Memory\n\n")
+# that appears even when no entries are selected.  A budget below this makes
+# every index operation fail immediately.
+POLICY_MIN_BUDGET = 22
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f4a8fcf — POLICY_MIN_BUDGET corrected from 22 to 20.

The empty-selection render produces `# Persistent Memory\n` (20 bytes), not `# Persistent Memory\n\n`: the trailing blank line is stripped by the render loop (`while lines and lines[-1] == "": lines.pop()`) when no entries are selected.

Also added an assertion in `test_policy_rejects_budget_below_minimum` that cross-checks POLICY_MIN_BUDGET against the actual rendered byte count, so a future constant/render divergence fails the test suite immediately.

38/38 memory policy tests pass.

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
```diff
@@ -319,21 +359,36 @@ def save(
...
-            scope=root.scope,
-        )
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
+            previous = (
+                parse_entry(path, scope=root.scope, strict=True)
```
TimeToBuildBob, line 367: ❌ **P1** — In `MemoryStore.save`, when the target file already exists, the new code calls `parse_entry(path, strict=True)` to read the previous entry. If that file is a `.md` file that is not a valid memory entry (for example it has no frontmatter), `parse_entry` raises `MemoryParseError`. None of the callers that were updated to handle the new failure modes catch this type: `cmd_memory.py` only catches `(KeyError, OSError, ValueError)` and `gptme/tools/memory.py` catches `(OSError, ValueError)`. `MemoryParseError` is distinct from `MemoryFrontmatterError` (the audit method in `store.py` catches them separately), and it is not a `ValueError`. As a result, a user saving to an existing non-entry `.md` file gets an unhandled `MemoryParseError` that propagates out of the CLI/tool and crashes with a traceback instead of showing a clean error message. Before this PR, `save` would simply overwrite any file at the slugged path regardless of its contents, so this is a regression in error handling for a realistic case (a stray notes file or a hand-written markdown file that was never meant to be a memory entry). The intended refusal to mutate malformed entries should raise an error that the existing `except` clauses can catch, such as `ValueError`.

gptme/memory/store.py
```diff
@@ -78,11 +78,11 @@ def _commit_replacements(pairs: list[tuple[Path, str]]) -> None:
...
-    staged: list[tuple[Path, Path, str | None]] = []
-    replaced: list[tuple[Path, str | None]] = []
+    staged: list[tuple[Path, Path, bytes | None]] = []
+    replaced: list[tuple[Path, bytes | None]] = []
     try:
         for dest, text in pairs:
-            original = dest.read_text(encoding="utf-8") if dest.is_file() else None
+            original = dest.read_bytes() if dest.is_file() else None
             dest.parent.mkdir(parents=True, exist_ok=True)
             tmp = dest.with_name(f".{dest.name}.{os.getpid()}.tmp")
```
TimeToBuildBob, line 87: ⚠️ **P2** — In _commit_replacements, if an OSError occurs while staging a later file (e.g., a disk-full error during f.write on the second entry), the temp file created for that file is never cleaned up because the staging loop only appends to the `staged` list after the write succeeds. The except block iterates over `staged` to unlink temps, so a partially written temp for the failed file is left on disk, and the exception propagates without that file being removed. Over repeated failures this leaks `.md.<pid>.tmp` files into the memory root, and the leak is not rolled back because the file is not tracked.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Addressed in `9174b867b` (pushed to this branch).

**P1 `5ec61237c47e` (`store.py:366`) is a false positive on this head.** `MemoryParseError` subclasses `ValueError` (`gptme/memory/schema.py:32`), so saving over a stray non-entry `.md` already surfaces a clean error through the existing `except (KeyError, OSError, ValueError)` (CLI) / `except (OSError, ValueError)` (tool) clauses — no traceback. Added `test_save_rejects_existing_non_entry_without_traceback` to pin that so it cannot regress silently.

**Real finding fixed:** `_commit_replacements` registered the temp path in `staged` only *after* the write finished, so an error mid-staging skipped the registration and left the partial `.tmp` sibling behind on rollback. The temp is now registered before writing; `test_staging_failure_removes_partial_temp_files` fails the second staged write and asserts the root is left byte-identical.

`tests/test_memory_index_policy.py tests/test_memory_store.py`: 120 passed, 1 skipped; ruff/mypy clean.

Why it took ~50 min: the 08:12Z PM launch for this PR was aborted by the user-slice memory-pressure pre-gate (hot=56–63%), and the aborted launch had already stamped its dedupe markers, so the mention read "event unchanged" until the next fingerprint change. That mechanism is fixed in ErikBjare/bob `b5784a8122` + `1065d386bf` (pre-gate skips now clear the `.event` and `.atoms` markers).

gptme/memory/store.py
TimeToBuildBob: Intentional policy boundary: on a policy-managed root, this compatibility helper regenerates the selected always-on view. Adding an unselected entry directly to MEMORY.md would bypass the persisted selection and byte budget. The docstring now states this explicitly, while roots without a policy retain the legacy incremental behavior.

gptme/memory/store.py
TimeToBuildBob: False positive. `MemoryParseError` subclasses `ValueError` in `gptme/memory/schema.py`, so both updated callers already catch it and return a clean error rather than a traceback. Commit `9174b867b` adds `test_save_rejects_existing_non_entry_without_traceback`, which exercises the reported no-frontmatter case through the CLI and passes.

gptme/memory/store.py
TimeToBuildBob: Fixed in commit 9174b867b. `_commit_replacements` now records each temp path before entering the write so an exception during staging is covered by cleanup. `test_staging_failure_removes_partial_temp_files` fails the second staged write and verifies that no temp siblings remain and the root stays byte-identical.

TimeToBuildBob:
Fixed the requested findings and closed the review threads.

- `9174b867b` removes partially written staging temp files on failure and adds the regression test.
- The P1 traceback report was a false positive: `MemoryParseError` subclasses `ValueError`; the new CLI regression pins that behavior.
- All of our AI-review findings are now explicitly disposed, and current CI is green.

I am not self-merging: the policy gate still reserves the new `gptme/memory/policy.py` public-module boundary for maintainer judgment. Greptile's 5/5 is on the prior head; its re-review helper has reached the anti-spam lifetime ceiling, so I did not bypass that guard with another trigger.

PR #3792, opened 2026-09-09
feat(memory): preserve writer provenance through CLI metadata

TimeToBuildBob:
Harness wrappers can now save provenance through `gptme-util memory save --metadata JSON`. The CLI validates an object before writing and delegates the metadata merge to `MemoryStore.save`; existing unrelated keys, curated titles, lifecycle, and root policy are preserved.

Depends on gptme/gptme#3791 for update preservation and index policy. The incremental change is commit `437ee0ae5`; the branch includes that prerequisite so CI can test the complete contract against master. Bob’s managed writer needs this to retain `originSessionId` and `node_type` while retiring direct index writes. Part of gptme/gptme#3734.

Validation: 121 focused store/policy/CLI tests passed, including six CLI metadata cases. Ruff, mypy, and commit hooks passed. Focused pytest used `--noconftest` because the workspace’s unrelated global Chroma model downloader calls a removed API; these tests use only built-in fixtures and real temporary stores.

### DIFF
--- a/gptme/cli/cmd_memory.py
+++ b/gptme/cli/cmd_memory.py
@@ -172,6 +172,8 @@ def memory_save(
         body = sys.stdin.read()
     else:
         body = ""
+    from ..memory.schema import MemoryParseError  # fmt: skip
+
     store = _store()
     try:
         path = store.save(
@@ -183,7 +185,7 @@ def memory_save(
             title=title,
             metadata=parsed_metadata,
         )
-    except (KeyError, OSError, ValueError) as e:
+    except (KeyError, OSError, ValueError, MemoryParseError) as e:
         click.echo(f"Error: {e}", err=True)
         sys.exit(1)
     if as_json:
@@ -297,9 +299,11 @@ def memory_recall(
 @click.option("--scope", help="Root containing both entries (default: write root).")
 def memory_supersede(old_name: str, new_name: str, scope: str | None):
     """Mark OLD_NAME superseded by NEW_NAME and link both entries."""
+    from ..memory.schema import MemoryParseError  # fmt: skip
+
     try:
         old, new = _store().supersede(old_name, new_name, scope=scope)
-    except (KeyError, OSError, ValueError) as e:
+    except (KeyError, OSError, ValueError, MemoryParseError) as e:
         raise click.ClickException(str(e)) from e
     click.echo(f"Superseded {_clean(old.name)} -> {_clean(new.name)}")
 
@@ -340,6 +344,8 @@ def memory_index(scope: str | None, write: bool, check: bool, budget: int | None
     Prints to stdout by default so a hand-curated MEMORY.md is never
     overwritten by accident; pass --write to replace it, --check to verify.
     """
+    from ..memory.schema import MemoryParseError  # fmt: skip
+
     store = _store()
     try:
         if check:
@@ -357,6 +363,6 @@ def memory_index(scope: str | None, write: bool, check: bool, budget: int | None
             ),
             nl=False,
         )
-    except (KeyError, OSError, ValueError) as e:
+    except (KeyError, OSError, ValueError, MemoryParseError) as e:
         click.echo(f"Error: {e}", err=True)
         sys.exit(1)
--- a/gptme/cli/cmd_memory.py
+++ b/gptme/cli/cmd_memory.py
@@ -161,6 +161,10 @@ def memory_save(
             ) from exc
         if not isinstance(parsed_metadata, dict):
             raise click.BadParameter("must be a JSON object", param_hint="--metadata")
+        if "type" in parsed_metadata:
+            raise click.BadParameter(
+                "type is reserved; use --type", param_hint="--metadata"
+            )
     if body_file:
         with open(body_file, encoding="utf-8") as f:
             body = f.read()
--- a/gptme/cli/cmd_memory.py
+++ b/gptme/cli/cmd_memory.py
@@ -125,6 +125,9 @@ def memory_show(name: str, as_json: bool):
 )
 @click.option("--title", help="Index link text (defaults to the name).")
 @click.option(
+    "--metadata", help="JSON object merged into entry metadata (e.g. provenance)."
+)
+@click.option(
     "--body-file",
     type=click.Path(exists=True, dir_okay=False),
     help="Read the body from this file instead of stdin.",
@@ -136,6 +139,7 @@ def memory_save(
     type_: str | None,
     scope: str | None,
     title: str | None,
+    metadata: str | None,
     body_file: str | None,
     as_json: bool,
 ):
@@ -147,6 +151,16 @@ def memory_save(
         gptme-util memory save prefer-short-answers \\
           "User prefers short, direct answers." --type feedback < body.md
     """
+    parsed_metadata = None
+    if metadata is not None:
+        try:
+            parsed_metadata = json.loads(metadata)
+        except ValueError as exc:
+            raise click.BadParameter(
+                "must be a JSON object", param_hint="--metadata"
+            ) from exc
+        if not isinstance(parsed_metadata, dict):
+            raise click.BadParameter("must be a JSON object", param_hint="--metadata")
     if body_file:
         with open(body_file, encoding="utf-8") as f:
             body = f.read()
@@ -156,7 +170,15 @@ def memory_save(
         body = ""
     store = _store()
     try:
-        path = store.save(name, description, body, type=type_, scope=scope, title=title)
+        path = store.save(
+            name,
+            description,
+            body,
+            type=type_,
+            scope=scope,
+            title=title,
+            metadata=parsed_metadata,
+        )
     except (KeyError, OSError, ValueError) as e:
         click.echo(f"Error: {e}", err=True)
         sys.exit(1)
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -435,6 +435,7 @@ def prompt_workspace(
                         break
                     try:
                         root_store = MemoryStore([root])
+                        charged_bytes: int | None = None
                         policy_path = root.path / POLICY_FILENAME
                         if policy_path.is_symlink():
                             raise ValueError(
@@ -466,6 +467,7 @@ def prompt_workspace(
                             with index_path.open("rb") as index_file:
                                 raw = index_file.read(available)
                             root_content = raw.decode("utf-8", errors="ignore").strip()
+                            charged_bytes = len(raw)
                     except Exception as e:
                         logger.warning(
                             "Failed to load memory root %s: %s", root.path, e
@@ -473,7 +475,11 @@ def prompt_workspace(
                         continue
                     if root_content:
                         parts.append(root_content)
-                        remaining -= separator_bytes + len(root_content.encode("utf-8"))
+                        remaining -= separator_bytes + (
+                            charged_bytes
+                            if charged_bytes is not None
+                            else len(root_content.encode("utf-8"))
+                        )
                 memory_content = "\n\n".join(parts).strip()
                 if memory_content:
                     root_paths = ", ".join(f"`{r.path}`" for r in existing_roots)
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from ..config import config_path, get_config, get_project_config
-from ..dirs import get_cc_memory_file
+from ..memory import MemoryStore, resolve_roots
 from ..message import Message
 from ..util.context import md_codeblock
 from ..util.context_dedup import _content_hash
@@ -16,7 +16,7 @@ from . import AGENT_FILES, DEFAULT_CONTEXT_FILES, _loaded_agent_files_var
 from .context_cmd import get_project_context_cmd_output
 
 _HASH_PREFIX = "ch:"
-_CC_MEMORY_MAX_BYTES = 64 * 1024  # 64 KB cap to avoid consuming excessive prompt budget
+_MEMORY_BUDGET_BYTES = 64 * 1024  # 64 KB cap to avoid consuming excessive prompt budget
 
 if TYPE_CHECKING:
     from ..util.uri import FilePath
@@ -416,32 +416,36 @@ def prompt_workspace(
             files=valid_context_files,
         )
 
-    # Load Claude Code memory if present — makes CC-written memories accessible in gptme
+    # Load persistent memories from all layered roots (project, CC, agent, user).
+    # This makes memories written by any harness (gptme, Claude Code, Codex) visible
+    # in gptme sessions — the cross-harness read path for #3625.
     if include_user_context:
-        cc_memory_file = get_cc_memory_file(workspace_resolved)
-        if cc_memory_file.exists():
-            try:
-                # Read at most MAX+1 bytes to detect oversized files without loading them fully
-                with open(cc_memory_file, "rb") as _f:
-                    raw = _f.read(_CC_MEMORY_MAX_BYTES + 1)
-                truncated = len(raw) > _CC_MEMORY_MAX_BYTES
-                if truncated:
-                    raw = raw[:_CC_MEMORY_MAX_BYTES]
-                    logger.warning(
-                        f"CC memory file {cc_memory_file} exceeds "
-                        f"{_CC_MEMORY_MAX_BYTES // 1024}KB; truncating"
-                    )
-                memory_content = raw.decode("utf-8", errors="ignore").strip()
-                if memory_content:
-                    yield Message(
-                        "system",
-                        f"## Persistent Memory\n\n"
-                        f"The following memory was saved across sessions "
-                        f"(from `{cc_memory_file}`):\n\n{memory_content}",
-                    )
-                    logger.debug(f"Loaded CC memory from {cc_memory_file}")
-            except OSError as e:
-                logger.debug(f"Failed to read CC memory file {cc_memory_file}: {e}")
+        try:
+            roots = resolve_roots(workspace_resolved)
+            existing_roots = [r for r in roots if r.exists]
+            if existing_roots:
+                store = MemoryStore(existing_roots)
+                entries = list(store.entries())
+                if entries:
+                    memory_content = MemoryStore.render_index(
+                        entries, budget=_MEMORY_BUDGET_BYTES
+                    ).strip()
+                    if memory_content:
+                        root_paths = ", ".join(f"`{r.path}`" for r in existing_roots)
+                        yield Message(
+                            "system",
+                            f"## Persistent Memory\n\n"
+                            f"The following memories are shared across sessions "
+                            f"(from {root_paths}):\n\n{memory_content}",
+                        )
+                        logger.debug(
+                            "Loaded %d memory entries from %d root(s): %s",
+                            len(entries),
+                            len(existing_roots),
+                            [r.scope for r in existing_roots],
+                        )
+        except Exception as e:
+            logger.debug(f"Failed to load layered memory: {e}")
 
     # Computed context last (changes most often, least cacheable)
     if (
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING
 
 from ..config import config_path, get_config, get_project_config
 from ..memory import MemoryStore, resolve_roots
+from ..memory.policy import POLICY_FILENAME
 from ..message import Message
 from ..util.context import md_codeblock
 from ..util.context_dedup import _content_hash
@@ -424,40 +425,45 @@ def prompt_workspace(
             roots = resolve_roots(workspace_resolved)
             existing_roots = [r for r in roots if r.exists]
             if existing_roots:
-                # Evaluate each root independently: use per-entry files when
-                # present, fall back to MEMORY.md for roots that have none.
-                # This prevents a mixed workspace (one root with entry files,
-                # another with only MEMORY.md) from silently dropping the
-                # legacy root's memories.
-                all_entries: list = []
                 parts: list[str] = []
                 remaining = _MEMORY_BUDGET_BYTES
                 for root in existing_roots:
-                    if remaining <= 0:
+                    # Separators count against the shared prompt budget too.
+                    separator_bytes = 2 if parts else 0
+                    available = remaining - separator_bytes
+                    if available <= 0:
                         break
-                    root_store = MemoryStore([root])
-                    root_entries = list(root_store.entries())
-                    if root_entries:
-                        all_entries.extend(root_entries)
-                        root_content = MemoryStore.render_index(
-                            root_entries, budget=remaining
-                        ).strip()
-                        if root_content:
-                            parts.append(root_content)
-                            remaining -= len(root_content.encode("utf-8"))
-                    else:
-                        # Fallback for roots that contain only MEMORY.md (no
-                        # individual entry files) — e.g. a CC root written by
-                        # an older harness or hand-authored index.  Preserves
-                        # the behaviour from #3626 on a per-root basis so that
-                        # a mixed workspace doesn't silently drop these roots.
-                        index_path = root.path / "MEMORY.md"
-                        if index_path.is_file():
-                            raw = index_path.read_bytes()[:remaining]
-                            text = raw.decode("utf-8", errors="ignore").strip()
-                            if text:
-                                parts.append(text)
-                                remaining -= len(raw)
+                    try:
+                        root_store = MemoryStore([root])
+                        policy_path = root.path / POLICY_FILENAME
+                        if policy_path.is_symlink():
+                            raise ValueError(
+                                "memory index policy must not be a symlink"
+                            )
+                        if policy_path.exists() or any(root_store.entries()):
+                            # A policy is authoritative even with no entries.
+                            # Missing selections or overflow must never revive
+                            # an obsolete on-disk MEMORY.md through fallback.
+                            root_content = root_store.render_root_index(
+                                budget=available
+                            ).strip()
+                        else:
+                            # Keep legacy index-only roots compatible, bounding
+                            # the read itself rather than slicing a full read.
+                            index_path = root.path / "MEMORY.md"
+                            if not index_path.is_file():
+                                continue
+                            with index_path.open("rb") as index_file:
+                                raw = index_file.read(available)
+                            root_content = raw.decode("utf-8", errors="ignore").strip()
+                    except Exception as e:
+                        logger.warning(
+                            "Failed to load memory root %s: %s", root.path, e
+                        )
+                        continue
+                    if root_content:
+                        parts.append(root_content)
+                        remaining -= separator_bytes + len(root_content.encode("utf-8"))
                 memory_content = "\n\n".join(parts).strip()
                 if memory_content:
                     root_paths = ", ".join(f"`{r.path}`" for r in existing_roots)
@@ -468,8 +474,8 @@ def prompt_workspace(
                         f"(from {root_paths}):\n\n{memory_content}",
                     )
                     logger.debug(
-                        "Loaded %d memory entries from %d root(s): %s",
-                        len(all_entries),
+                        "Loaded %d memory indexes from %d root(s): %s",
+                        len(parts),
                         len(existing_roots),
                         [r.scope for r in existing_roots],
                     )
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -430,20 +430,37 @@ def prompt_workspace(
                     memory_content = MemoryStore.render_index(
                         entries, budget=_MEMORY_BUDGET_BYTES
                     ).strip()
-                    if memory_content:
-                        root_paths = ", ".join(f"`{r.path}`" for r in existing_roots)
-                        yield Message(
-                            "system",
-                            f"## Persistent Memory\n\n"
-                            f"The following memories are shared across sessions "
-                            f"(from {root_paths}):\n\n{memory_content}",
-                        )
-                        logger.debug(
-                            "Loaded %d memory entries from %d root(s): %s",
-                            len(entries),
-                            len(existing_roots),
-                            [r.scope for r in existing_roots],
-                        )
+                else:
+                    # Fallback for roots that contain only MEMORY.md (no individual
+                    # entry files) — e.g. a CC root written by an older harness or
+                    # hand-authored before the per-entry format was introduced.
+                    # This preserves the behaviour from #3626 so existing CC memories
+                    # are not silently dropped when upgrading.
+                    parts: list[str] = []
+                    remaining = _MEMORY_BUDGET_BYTES
+                    for root in existing_roots:
+                        index_path = root.path / "MEMORY.md"
+                        if index_path.is_file() and remaining > 0:
+                            raw = index_path.read_bytes()[:remaining]
+                            text = raw.decode("utf-8", errors="ignore").strip()
+                            if text:
+                                parts.append(text)
+                                remaining -= len(raw)
+                    memory_content = "\n\n".join(parts).strip()
+                if memory_content:
+                    root_paths = ", ".join(f"`{r.path}`" for r in existing_roots)
+                    yield Message(
+                        "system",
+                        f"## Persistent Memory\n\n"
+                        f"The following memories are shared across sessions "
+                        f"(from {root_paths}):\n\n{memory_content}",
+                    )
+                    logger.debug(
+                        "Loaded %d memory entries from %d root(s): %s",
+                        len(entries),
+                        len(existing_roots),
+                        [r.scope for r in existing_roots],
+                    )
         except Exception as e:
             logger.debug(f"Failed to load layered memory: {e}")
 
--- a/gptme/prompts/workspace.py
+++ b/gptme/prompts/workspace.py
@@ -440,7 +440,17 @@ def prompt_workspace(
                             raise ValueError(
                                 "memory index policy must not be a symlink"
                             )
-                        if policy_path.exists() or any(root_store.entries()):
+                        # Check for individual entry files (any .md that isn't
+                        # the index or policy). When entries() silently returns []
+                        # due to parse errors, this prevents the legacy MEMORY.md
+                        # fallback from injecting stale content: render_root_index()
+                        # will return empty for a broken root, which is correct.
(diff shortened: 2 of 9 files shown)


gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
+                if path.exists()
+                else MemoryEntry(name=slug, path=path, scope=root.scope)
+            )
+            entry = replace(
+                previous,
+                description=description.strip(),
+                body=body,
+                type=type if type is not None else previous.type,
+                title=title if title is not None else previous.title,
+                metadata={**previous.metadata, **(metadata or {})},
```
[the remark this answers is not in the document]
TimeToBuildBob: Reproduced against 437ee0ae5: metadata.type silently changed an explicit feedback type to project. Fixed in d1d02db4c by rejecting the reserved key at the CLI boundary and directing callers to --type. The regression now proves failure before any entry bytes change; all seven CLI metadata cases pass, with commit hooks green.

gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
+            previous = (
+                parse_entry(path, scope=root.scope, strict=False)
+                if path.exists()
+                else MemoryEntry(name=slug, path=path, scope=root.scope)
```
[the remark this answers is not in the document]
TimeToBuildBob: Confirmed this is inherited from prerequisite gptme/gptme#3791 (9f7ea8345), whose live project-monitoring session owns the same data-loss finding. This consumer PR and Bob live migration remain blocked on that correction; the task and acceptance report name it explicitly. I will not duplicate a conflicting store change on this branch.

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 21c9fb7af — `cmd_memory.py` now rejects `"type"` in `--metadata` before the write with a clear error directing users to `--type`. Test added covering the rejected case.

gptme/memory/store.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in eca7ea27d — `save()` now uses `strict=True` when reading an existing entry. If the file has malformed YAML it raises `MemoryFrontmatterError` immediately, leaving the file unchanged. Matches `supersede()` which already used strict parsing, and aligns with what `docs/memory.rst` documents. Updated the test (previously expected lenient-repair behavior) to assert the error and verify the file is unmodified.

gptme/memory/store.py
```diff
@@ -544,16 +592,53 @@ def build(n_entries: int) -> str:
...
+                f"selected memory index needs {size} bytes, exceeds budget {limit}; "
+                "shorten descriptions or revise .memory-index.json before writing"
+            )
+        return text
+
+    def render_root_index(
+        self, scope: str | None = None, *, budget: int | None = None
+    ) -> str:
+        """Render one root using its persistent selection and budget, if present."""
+        with _locked_root(self.root(scope).path):
```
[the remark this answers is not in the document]
TimeToBuildBob: Acknowledged. This is a real side effect: `_locked_root` creates the directory and lock file even for read-only operations. Separating lock acquisition from root creation would be the cleaner fix (create on write, acquire-only on read).

This is P2 scope and not a correctness blocker — the effect is limited to filesystem state (a directory and `.memory.lock` file appear). Leaving this for a follow-up cleanup rather than extending this PR further.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/prompts/workspace.py
```diff
@@ -416,32 +417,70 @@ def _is_excluded(f: Path, patterns: list[str]) -> bool:
...
+                            ).strip()
+                        else:
+                            # Keep legacy index-only roots compatible, bounding
+                            # the read itself rather than slicing a full read.
+                            index_path = root.path / "MEMORY.md"
+                            if not index_path.is_file():
+                                continue
+                            with index_path.open("rb") as index_file:
+                                raw = index_file.read(available)
+                            root_content = raw.decode("utf-8", errors="ignore").strip()
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d7832cdb1 — legacy MEMORY.md read now guards `index_path.is_symlink()` before opening, matching the adjacent policy-path check. A symlinked index is skipped via `continue` rather than read.

gptme/prompts/workspace.py
```diff
@@ -416,32 +417,70 @@ def _is_excluded(f: Path, patterns: list[str]) -> bool:
...
+                    if available <= 0:
+                        break
+                    try:
+                        root_store = MemoryStore([root])
+                        policy_path = root.path / POLICY_FILENAME
+                        if policy_path.is_symlink():
+                            raise ValueError(
+                                "memory index policy must not be a symlink"
+                            )
+                        if policy_path.exists() or any(root_store.entries()):
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d7832cdb1 — added `_has_entry_files` check before the legacy fallback. When `entries()` silently returns `[]` because all files are malformed, the root still has `.md` files other than `MEMORY.md`/policy, so `_has_entry_files` is `True` and the code takes the MemoryStore path (`render_root_index()` returns empty, nothing injected) rather than the stale legacy index.

TimeToBuildBob:
Merge blocked: depends on #3791 which needs Erik's sign-off on the new `gptme/memory/policy.py` module placement before it can self-merge. Tracked in ErikBjare/bob#1224. Will land once #3791 clears.

gptme/cli/cmd_memory.py
```diff
@@ -157,7 +174,15 @@ def memory_save(
...
+        path = store.save(
+            name,
+            description,
+            body,
+            type=type_,
+            scope=scope,
+            title=title,
+            metadata=parsed_metadata,
+        )
     except (KeyError, OSError, ValueError) as e:
```
TimeToBuildBob: ❌ **P1** — The CLI `memory save` and `memory supersede` commands catch only KeyError, OSError, and ValueError, but the store now runs `parse_entry(..., strict=True)` on existing entry files. When an entry file has malformed YAML frontmatter, parse_entry raises MemoryFrontmatterError, a subclass of MemoryParseError that is not a ValueError. That exception escapes the CLI's except block, so the process prints a full traceback and exits non-zero instead of showing the intended friendly error message. Harness wrappers calling the CLI will observe unclean output and may mis-diagnose the failure. The fix is to also catch MemoryParseError (or MemoryFrontmatterError) around the save/supersede calls.

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (fixed in 9e7a3436f)

| --- | --- | --- |
| `gptme/cli/cmd_memory.py:186` | P1 | superseded by latest review (not reproduced) |
| `gptme/memory/store.py:349` | P1 | superseded by latest review (not reproduced) |
| `gptme/prompts/workspace.py:453` | P1 | superseded by latest review (not reproduced) |
| `gptme/prompts/workspace.py:476` | P1 | fixed in `9e7a3436f` |
| `gptme/tools/memory.py:135` | P1 | superseded by latest review (not reproduced) |
| `gptme/cli/cmd_memory.py:164` | P2 | rejected |
| `gptme/prompts/workspace.py:448` | P2 | rejected |
| `tests/test_cc_memory.py:562` | P2 | superseded by latest review (not reproduced) |
| `tests/test_cc_memory.py:573` | P2 | accepted-tradeoff |

</details>




This PR adds a --metadata option to gptme-util memory save that validates a JSON object and merges it into entry metadata, preserving existing fields and policy. It also replaces the single CC MEMORY.md read in prompt_workspace with a layered MemoryStore read across all roots, with per-root policy/entry-file detection and a shared byte budget. It adds symlink protection for the index policy file and broadens CLI error handling to catch MemoryParseError. Tests are updated and added for the new metadata, layered loading, and policy behavior.

Confidence 3/5

<details><summary><b>4 findings</b> · ❌ 1 P1 · ⚠️ 3 P2</summary>

**❌ P1 high** — `gptme/prompts/workspace.py:476`

In prompt_workspace, the byte budget accounting is wrong for multi-byte UTF-8 content. The code reads `available` bytes from a legacy MEMORY.md file, decodes with errors='ignore', then subtracts `len(root_content.encode('utf-8'))` from `remaining`. But `available` is a byte count, and `index_file.read(available)` reads exactly `available` bytes. If the file contains multi-byte characters, decoding with errors='ignore' can drop trailing partial bytes, so the decoded string's UTF-8 length is less than `available`. The budget is then under-counted: `remaining` is reduced by fewer bytes than were actually consumed from the file. Over multiple roots, this allows the total injected content to exceed `_MEMORY_BUDGET_BYTES`. The test `test_legacy_reads_and_separators_obey_byte_budget` uses budget=10 and content '12345' and 'é'*100000; the first root consumes 5 bytes, the second reads 5 bytes (available=5 after separator), which decodes to 'é' (2 bytes) plus possibly partial bytes dropped, so the assertion `len(content.encode('utf-8')) <= 10` passes, but the actual file bytes consumed were 10, and the decoded content is only 7 bytes. The bug is that the budget tracks decoded bytes, not raw bytes read, so a file with many multi-byte chars can cause the prompt to exceed the cap. The fix is to track the raw byte count read, e.g. `raw = index_file.read(available); remaining -= separator_bytes + len(raw)`.

```suggestion
remaining -= separator_bytes + len(raw)
```

<sub>**How this was verified:** Traced the legacy fallback path: `raw = index_file.read(available)` reads `available` bytes, then `root_content = raw.decode('utf-8', errors='ignore').strip()`. The subtraction uses the encoded length of the decoded string, which can be smaller than `available` when multi-byte characters are truncated. The test in test_workspace_memory_policy.py uses 'é'*100000 and budget=10, but only checks the final content length, not the raw bytes read.</sub>

**⚠️ P2 medium** — `gptme/prompts/workspace.py:448`

The `_has_entry_files` check in prompt_workspace uses `root.path.glob("*.md")` and filters out `MEMORY.md` and `POLICY_FILENAME`. However, `POLICY_FILENAME` is `.memory-index.json`, which does not match `*.md`, so the filter is redundant but harmless. More importantly, the check only looks for files directly in the root, not subdirectories. If a memory root has entries in subdirectories (e.g. `root/sub/entry.md`), `_has_entry_files` is False, and the code falls back to reading `MEMORY.md` even though the root has entry files. This can cause stale MEMORY.md content to be injected instead of the rendered index. The MemoryStore.entries() likely scans subdirectories, so the fallback path is inconsistent. The test suite only creates entries directly in the root, so this is not caught. The fix is to use a recursive glob or check `root_store.entries()` instead of a shallow glob.

<sub>**How this was verified:** Checked the glob pattern: `root.path.glob("*.md")` only matches files directly in the root. The MemoryStore likely uses a recursive scan (not shown in the diff, but the memory store is designed for layered roots with subdirectories). The fallback to MEMORY.md would then be used for a root with subdirectory entries, potentially injecting stale index content.</sub>

**⚠️ P2 medium** — `gptme/cli/cmd_memory.py:164`

In the CLI metadata validation, the `--metadata` option is parsed with `json.loads` and checked to be a dict, but there is no check for nested objects or keys that conflict with lifecycle fields other than `type`. The docs say 'lifecycle fields and root index policy remain governed by the store', but a user can pass `--metadata '{"status":"historical"}'` or `--metadata '{"superseded_by":"other"}'` and these will be merged into the entry metadata. If the store's save method merges metadata into the frontmatter without validating against reserved keys, this could allow a harness wrapper to set lifecycle fields that should be managed by the store, breaking supersession or index policy. The test only checks the `type` key is reserved. The fix is to validate that metadata keys do not include any reserved lifecycle field names (e.g. status, superseded_by, supersedes, provenance) or to have the store ignore/override them.

<sub>**How this was verified:** The diff only reserves the `type` key. The docs say lifecycle fields are governed by the store, but the CLI does not prevent other lifecycle keys. The store's save method (not shown in full) likely merges metadata into the entry's metadata dict, which is then serialized to frontmatter. If a user passes `status`, it could conflict with the entry's status field.</sub>

**⚠️ P2 medium** — `tests/test_cc_memory.py:573`

The test `test_budget_limits_injected_content` in tests/test_cc_memory.py asserts that the rendered index contains the word 'omitted' when the budget is exceeded. However, the test creates 20 entries with 4000-character descriptions, but the `_make_entry` helper writes the description into the frontmatter as a quoted string. The rendered index line includes the description, so the index is large. The test asserts `injected_bytes <= _MEMORY_BUDGET_BYTES + 512`, but the memory message also includes the header '## Persistent Memory' and the root paths, which are not counted in the budget. The budget is applied to `render_root_index` with `budget=available`, where `available` is `_MEMORY_BUDGET_BYTES` minus separators. The final message adds a header and root paths, so the total injected bytes can exceed the budget by more than 512 bytes if the root path is long. The test uses a tmp_path, so the root path is short, but in real usage the root path could be long, causing the total prompt to exceed the intended cap. The test's assertion with +512 is a heuristic that may not hold for long paths, but the test itself is not the defect; the defect is that the budget does not account for the header and root paths. The code should subtract the header/root path length from the budget or the test should be more precise.

<sub>**How this was verified:** The memory message is constructed as `f"## Persistent Memory\n\nThe following memories are shared across sessions (from {root_paths}):\n\n{memory_content}"`. The budget is only applied to `memory_content` via `render_root_index(budget=available)`. The header and root_paths are added after, so the total message can exceed the budget by the length of the header and root paths. The test allows 512 bytes of overhead, which may be insufficient for long paths.</sub>

</details>

- `docs/memory.rst` — Adds a Writer provenance section documenting the --metadata option and its validation rules.
- `gptme/cli/cmd_memory.py` — Adds --metadata option to memory save with JSON object validation and reserved type key check; catches MemoryParseError in save, supersede, and index commands.
- `gptme/memory/policy.py` — Adds a symlink check in IndexPolicy.read that raises ValueError if the policy file is a symlink.
- `gptme/prompts/workspace.py` — Replaces the single CC MEMORY.md read with layered MemoryStore reads across all roots, with per-root policy/entry detection, shared byte budget, and symlink checks.
- `gptme/tools/memory.py` — Adds MemoryParseError to the exception tuple caught when saving memory via the tool.
- `tests/test_cc_memory.py` — Rewrites workspace prompt memory tests to use MemoryRoot fixtures and adds tests for layered roots, legacy fallback, and budget limits.
- `tests/test_memory_cli_metadata.py` — Adds tests for the --metadata option: preservation, invalid input rejection, reserved type key, and friendly error on malformed existing entry.
- `tests/test_memory_index_policy.py` — Adds a test that a symlinked policy file fails closed without mutating the target.
- `tests/test_workspace_memory_policy.py` — Adds tests for policy-aware layered prompt loading, including legacy root compatibility and budget behavior.

</details>

| `d7832cdb158c` | 2/5 | 3 | llm | 2026-09-09 20:01 UTC |
| `de66f317a652` | 3/5 | 2 | llm | 2026-09-09 21:01 UTC |
| `fce2bf507c9c` | 3/5 | 1 | llm | 2026-09-09 22:41 UTC |

</details>

<sub>Reviewed `e8c4966650c5` · `deepseek/deepseek-v4-flash-0731` · `llm` engine · 63s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/prompts/workspace.py
```diff
@@ -416,32 +417,80 @@ def _is_excluded(f: Path, patterns: list[str]) -> bool:
...
+                        # the index or policy). When entries() silently returns []
+                        # due to parse errors, this prevents the legacy MEMORY.md
+                        # fallback from injecting stale content: render_root_index()
+                        # will return empty for a broken root, which is correct.
+                        _has_entry_files = any(
+                            f.is_file() and not f.is_symlink()
+                            for f in root.path.glob("*.md")
+                            if f.name not in ("MEMORY.md", POLICY_FILENAME)
+                        )
+                        if policy_path.exists() or _has_entry_files:
```
TimeToBuildBob, line 454: ❌ **P1** — In the new workspace memory-loading logic, every root that has at least one per-entry file OR a policy .memory-index.json is loaded by calling root_store.render_root_index(), which regenerates an index solely from the entry files. The hand-curated MEMORY.md file is never consulted in that branch. A legacy root that contains both individual entry files and an editing-index of extra prose, summary text, or links not backed by a separate entry file will silently lose that hand-written content from the auto-loaded session context. Previously the whole CC MEMORY.md file was injected verbatim. Users who curated their MEMORY.md with notes not in entry files will see those notes vanish from the prompt without warning, changing the agent's context and behavior. The system should fall back to (or merge in) the existing MEMORY.md when the root has no policy and the index file contains material not generated from entry files.

tests/test_cc_memory.py
```diff
@@ -409,9 +556,8 @@ def bounded_read(n=-1):
...
-        injected = memory_msgs[0].content.encode("utf-8")
-        # Output is bounded
-        assert len(injected) <= _CC_MEMORY_MAX_BYTES * 2
-        # The file was read with a size bound, not in full
-        assert max_bytes_read, "open() was not called on the memory file in binary mode"
-        assert max(max_bytes_read) <= _CC_MEMORY_MAX_BYTES + 1
+        # The rendered index (not individual entries) is included — it should be bounded
+        injected_bytes = len(memory_msgs[0].content.encode("utf-8"))
+        assert (
+            injected_bytes <= _MEMORY_BUDGET_BYTES + 512
```
TimeToBuildBob: ⚠️ **P2** — The test `test_budget_limits_injected_content` does not exercise the budget limit it claims to verify. It writes 20 entries each with a 5000-byte body, but the rendered index only contains each entry's one-line description (`Entry {i}`) and never the body, so the generated index is only a few hundred bytes — far below `_MEMORY_BUDGET_BYTES`. The assertion `injected_bytes <= _MEMORY_BUDGET_BYTES + 512` always passes regardless of whether truncation actually works. To test the limit, entries must have descriptions long enough (or the budget must be patched lower, as in `test_workspace_memory_policy.py`) to force the index to exceed the 64KB cap. As written, this test provides false confidence that oversized memory content is bounded.

gptme/cli/cmd_memory.py
TimeToBuildBob: Fixed in de66f317a — added a lazy `from ..memory.schema import MemoryParseError` import at the top of each affected command body (`save`, `supersede`, `index`) and included `MemoryParseError` in all three `except` tuples. Regression test `test_save_over_malformed_existing_entry_shows_friendly_error` writes a malformed-frontmatter file, calls the CLI, and asserts `exit_code != 0` with no `Traceback` in output.

gptme/prompts/workspace.py
TimeToBuildBob: This is intentional design, not a regression. The new architecture's invariant is: once a root contains individual entry files, `MEMORY.md` is a generated view of those entries, not an independent source of truth. The only roots that keep falling through to the legacy MEMORY.md read are roots with *no* entry files at all — which is the backwards-compatible path for existing CC MEMORY.md users who haven't migrated yet.

A root that contains both entry files and hand-curated MEMORY.md content not backed by entry files is in an inconsistent mid-migration state. The correct resolution is to migrate that prose into entry files via `gptme-util memory save`, which the migration guide covers. Merging the two sources in the loader would mean double-injecting content (the entry renders it, then the old MEMORY.md line does it again), and would paper over the inconsistency rather than surfacing it.

If the concern is discoverability — a user who had a curated MEMORY.md and created their first entry file without realizing the loader would switch branches — that's a real UX issue, but the fix belongs in the migration tooling (`memory audit` can flag roots where MEMORY.md has content not covered by any entry file), not in the loader logic. I'll file a follow-up task for that.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI failure on this PR is **not caused by the memory changes** — it's a known flaky test.

**Failing test**: `test_multi_tool_per_message` in `tests/test_server_v2_auto_stepping.py`
**Failure mode**: `wait_for_event(event_listener, 'message_added')` timed out (10s timeout) — an SSE event delivery issue under CI load, not a logic regression.

The test itself documents the fragility:
> "Must not be retried: pytest-retry with --retries surfaces a tmp_path teardown KeyError after the retried attempt passes"

Yet CI retried it anyway (1 retried shown in results), and it still failed. This is classic flakiness from high concurrency (16 parallel workers + CI resource pressure).

The PR changes touch `gptme/memory/`, `gptme/cli/cmd_memory.py`, `gptme/prompts/workspace.py`, and tests — none of these code paths are involved in the server auto-stepping SSE event pipeline.

Re-triggered the CI run to confirm. If it passes the re-run, this is confirmed pre-existing flakiness and the PR is clean.

gptme/tools/memory.py
```diff
@@ -132,7 +132,7 @@ def _do_save(
         try:
             file_path = save_memory(name, save_content)
             yield Message("system", f"Memory '{name}' saved to `{file_path}`.")
-        except OSError as e:
+        except (OSError, ValueError) as e:
```
TimeToBuildBob: ❌ **P1** — The gptme memory tool's `_do_save` catches only `OSError` and `ValueError`. Since this PR makes `save` read an existing entry with `strict=True`, a malformed existing frontmatter now raises `MemoryFrontmatterError`, a subclass of `MemoryParseError` (which is not a `ValueError`). That exception escapes the tool. An agent updating a memory whose file has malformed YAML (for example from a partial write) will hit an unhandled exception, terminating the action without a clear message, instead of the friendly error message the CLI now provides.

gptme/cli/cmd_memory.py
TimeToBuildBob: Fixed in de66f317a (already on branch): `MemoryParseError` was added to the except block in `cmd_memory.py` for `save`, `supersede`, and `index` commands.

gptme/prompts/workspace.py
TimeToBuildBob: Fixed in d7832cdb1 + eb7acebf8 (already on branch): workspace.py now falls through to the legacy MEMORY.md only when no policy file and no individual entry files exist. Roots that have a policy or entry files use `render_root_index()`; hand-curated index-only roots still get their MEMORY.md verbatim (bounded read).

tests/test_cc_memory.py
TimeToBuildBob: Fixed in fce2bf507: changed entries to use 4000-char descriptions (so 20 × 4000 ≈ 80 KB raw > 64 KB budget) and added an assertion that the 'omitted' marker is present, proving truncation actually fired. The original short descriptions ('Entry N') only rendered to ~300 bytes total, making the budget assertion trivially true.

gptme/tools/memory.py
TimeToBuildBob: Fixed in 31c445097: added `from ..memory.schema import MemoryParseError` and included it in the `except (OSError, ValueError, MemoryParseError)` clause in `_do_save`. The agent now gets a friendly error message instead of an unhandled traceback when the existing entry file has malformed YAML frontmatter.

gptme/memory/store.py
```diff
@@ -319,21 +344,36 @@ def save(
...
-            title=title,
-            metadata=dict(metadata or {}),
-            scope=root.scope,
-        )
-        entry.path = root.path / entry.filename
+        path = root.path / f"{slug}.md"
         with _locked_root(root.path):
-            entry.path.write_text(entry.to_markdown(), encoding="utf-8")
-            update_index_line(root.path, entry)
+            policy = IndexPolicy.read(root.path)
```
TimeToBuildBob, line 365: ❌ **P1** — `MemoryStore.save` and `MemoryStore.supersede` read the persistent index policy with `IndexPolicy.read(root.path)` without checking whether `.memory-index.json` is a symlink. The workspace prompt path explicitly refuses a symlinked policy (`policy_path.is_symlink()` raises), but the store writes do not. If a malicious or accidental symlink is placed at a root's `.memory-index.json` pointing to an external file (e.g. another project's policy, or a file with a very large budget and a broad `selected` list), every subsequent `save` or `supersede` in that root will follow the link. The attacker controls which entries are treated as required for the always-on index, and can set `budget` arbitrarily high so the rendered index (or the selected filename checks) cause unexpected failures, denial of service, or disclosure of whether referenced filenames exist. The write operations are silent about the source file, reading whatever the symlink resolves to.

TimeToBuildBob:
@greptileai review

gptme/memory/store.py
TimeToBuildBob: Fixed in 825cb10da — `IndexPolicy.read()` now rejects `.memory-index.json` when it is a symlink, so every caller (`save`, `supersede`, render, write, and check) fails closed through one policy boundary. The regression test exercises all five paths and verifies neither the memory root nor the external symlink target changes.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fresh review is complete on head `825cb10daa80`: Greptile is 5/5, all review threads are resolved, and CI is green. No further code change or re-trigger needed. Merge remains gated on prerequisite gptme/gptme#3791.

TimeToBuildBob:
Rebased onto current `master` after gptme/gptme#3791 merged and force-pushed head `e8c496665`. The prerequisite's duplicated commits were dropped; conflict resolution kept the merged strict malformed-frontmatter behavior and atomic rollback implementation from `master`.

Local verification: 164 focused memory tests passed (1 skipped); Ruff check and format check passed. The PR is now mergeable; fresh CI is running.

TimeToBuildBob:
@greptileai review

gptme/prompts/workspace.py
```diff
@@ -416,32 +417,80 @@ def _is_excluded(f: Path, patterns: list[str]) -> bool:
...
+                                raw = index_file.read(available)
+                            root_content = raw.decode("utf-8", errors="ignore").strip()
+                    except Exception as e:
+                        logger.warning(
+                            "Failed to load memory root %s: %s", root.path, e
+                        )
+                        continue
+                    if root_content:
+                        parts.append(root_content)
+                        remaining -= separator_bytes + len(root_content.encode("utf-8"))
```
TimeToBuildBob: ❌ **P1** — In prompt_workspace, the byte budget accounting is wrong for multi-byte UTF-8 content. The code reads `available` bytes from a legacy MEMORY.md file, decodes with errors='ignore', then subtracts `len(root_content.encode('utf-8'))` from `remaining`. But `available` is a byte count, and `index_file.read(available)` reads exactly `available` bytes. If the file contains multi-byte characters, decoding with errors='ignore' can drop trailing partial bytes, so the decoded string's UTF-8 length is less than `available`. The budget is then under-counted: `remaining` is reduced by fewer bytes than were actually consumed from the file. Over multiple roots, this allows the total injected content to exceed `_MEMORY_BUDGET_BYTES`. The test `test_legacy_reads_and_separators_obey_byte_budget` uses budget=10 and content '12345' and 'é'*100000; the first root consumes 5 bytes, the second reads 5 bytes (available=5 after separator), which decodes to 'é' (2 bytes) plus possibly partial bytes dropped, so the assertion `len(content.encode('utf-8')) <= 10` passes, but the actual file bytes consumed were 10, and the decoded content is only 7 bytes. The bug is that the budget tracks decoded bytes, not raw bytes read, so a file with many multi-byte chars can cause the prompt to exceed the cap. The fix is to track the raw byte count read, e.g. `raw = index_file.read(available); remaining -= separator_bytes + len(raw)`.

gptme/prompts/workspace.py
```diff
@@ -416,32 +417,80 @@ def _is_excluded(f: Path, patterns: list[str]) -> bool:
...
+                        if policy_path.is_symlink():
+                            raise ValueError(
+                                "memory index policy must not be a symlink"
+                            )
+                        # Check for individual entry files (any .md that isn't
+                        # the index or policy). When entries() silently returns []
+                        # due to parse errors, this prevents the legacy MEMORY.md
+                        # fallback from injecting stale content: render_root_index()
+                        # will return empty for a broken root, which is correct.
+                        _has_entry_files = any(
```
TimeToBuildBob, line 449: ⚠️ **P2** — The `_has_entry_files` check in prompt_workspace uses `root.path.glob("*.md")` and filters out `MEMORY.md` and `POLICY_FILENAME`. However, `POLICY_FILENAME` is `.memory-index.json`, which does not match `*.md`, so the filter is redundant but harmless. More importantly, the check only looks for files directly in the root, not subdirectories. If a memory root has entries in subdirectories (e.g. `root/sub/entry.md`), `_has_entry_files` is False, and the code falls back to reading `MEMORY.md` even though the root has entry files. This can cause stale MEMORY.md content to be injected instead of the rendered index. The MemoryStore.entries() likely scans subdirectories, so the fallback path is inconsistent. The test suite only creates entries directly in the root, so this is not caught. The fix is to use a recursive glob or check `root_store.entries()` instead of a shallow glob.

gptme/cli/cmd_memory.py
```diff
@@ -147,17 +151,41 @@ def memory_save(
...
+    if metadata is not None:
+        try:
+            parsed_metadata = json.loads(metadata)
+        except ValueError as exc:
+            raise click.BadParameter(
+                "must be a JSON object", param_hint="--metadata"
+            ) from exc
+        if not isinstance(parsed_metadata, dict):
+            raise click.BadParameter("must be a JSON object", param_hint="--metadata")
+        if "type" in parsed_metadata:
```
TimeToBuildBob, line 164: ⚠️ **P2** — In the CLI metadata validation, the `--metadata` option is parsed with `json.loads` and checked to be a dict, but there is no check for nested objects or keys that conflict with lifecycle fields other than `type`. The docs say 'lifecycle fields and root index policy remain governed by the store', but a user can pass `--metadata '{"status":"historical"}'` or `--metadata '{"superseded_by":"other"}'` and these will be merged into the entry metadata. If the store's save method merges metadata into the frontmatter without validating against reserved keys, this could allow a harness wrapper to set lifecycle fields that should be managed by the store, breaking supersession or index policy. The test only checks the `type` key is reserved. The fix is to validate that metadata keys do not include any reserved lifecycle field names (e.g. status, superseded_by, supersedes, provenance) or to have the store ignore/override them.

tests/test_cc_memory.py
```diff
@@ -409,9 +560,14 @@ def bounded_read(n=-1):
...
+        injected_bytes = len(memory_msgs[0].content.encode("utf-8"))
+        # Verify the budget cap actually fired: combined raw descriptions are
+        # ~80 KB (20 × 4000 chars), so without truncation we'd far exceed 64 KB.
+        # A passing assertion from a trivially-small index would mean the budget
+        # logic is untested.
+        assert "omitted" in memory_msgs[0].content.lower(), (
+            "expected the index to be truncated and report omitted entries, "
+            "but the 'omitted' marker is absent — budget limit may not have fired"
+        )
+        assert injected_bytes <= _MEMORY_BUDGET_BYTES + 512  # small header overhead
```
TimeToBuildBob, line 573: ⚠️ **P2** — The test `test_budget_limits_injected_content` in tests/test_cc_memory.py asserts that the rendered index contains the word 'omitted' when the budget is exceeded. However, the test creates 20 entries with 4000-character descriptions, but the `_make_entry` helper writes the description into the frontmatter as a quoted string. The rendered index line includes the description, so the index is large. The test asserts `injected_bytes <= _MEMORY_BUDGET_BYTES + 512`, but the memory message also includes the header '## Persistent Memory' and the root paths, which are not counted in the budget. The budget is applied to `render_root_index` with `budget=available`, where `available` is `_MEMORY_BUDGET_BYTES` minus separators. The final message adds a header and root paths, so the total injected bytes can exceed the budget by more than 512 bytes if the root path is long. The test uses a tmp_path, so the root path is short, but in real usage the root path could be long, causing the total prompt to exceed the intended cap. The test's assertion with +512 is a heuristic that may not hold for long paths, but the test itself is not the defect; the defect is that the budget does not account for the header and root paths. The code should subtract the header/root path length from the budget or the test should be more precise.

gptme/prompts/workspace.py
TimeToBuildBob: Fixed in `9e7a3436f`: legacy fallback now charges the raw bytes actually read, rather than the shorter decoded text after an incomplete UTF-8 suffix is dropped. Added a three-root regression proving the dropped suffix cannot leave phantom budget for a later root; the focused workspace-memory suites pass (36 tests).

gptme/prompts/workspace.py
TimeToBuildBob: Rejected: this storage format is deliberately flat. `MemoryStore._entries_from_roots()` and `index_entries()` both use `root.path.glob("*.md")`, so a nested Markdown file is not an entry and cannot make the legacy fallback inconsistent with store behavior. Recursive discovery would be a separate format change, not a fix for this PR.

gptme/cli/cmd_memory.py
TimeToBuildBob: Rejected: `--metadata` is written under the nested `metadata:` mapping. `MemoryEntry.to_markdown()` serializes canonical `status`, `supersedes`, `superseded_by`, and `provenance` from dedicated fields outside that mapping, and `parse_entry()` reads those lifecycle fields only from top-level frontmatter. Nested keys with those names cannot override lifecycle state. `type` is uniquely reserved because serialization explicitly merges it into `metadata:` and parsing pops it from there as the canonical type.

tests/test_cc_memory.py
TimeToBuildBob: Accepted trade-off: `_MEMORY_BUDGET_BYTES` caps memory payload content, not the fixed system-message framing. The header and root list are small diagnostic framing, while unbounded user-controlled memory content is budgeted. Charging filesystem path text against entry capacity would make the useful memory budget depend on install location; the existing bounded-overhead assertion intentionally permits that framing.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fresh head review found one real UTF-8 budget-accounting bug. Fixed and pushed as `9e7a3436f`; the focused workspace-memory suites pass (36 tests), and the three non-blocking findings were dispositioned on their threads with code-level reasons. CI and Greptile re-review are now running on the new head; no merge attempted while those gates are pending.

gptme/prompts/workspace.py
TimeToBuildBob: Fixed in commit 9e7a3436f: legacy fallback now charges the raw bytes actually read, and the focused regression suite passes. Resolving the already-answered finding thread after merge.

PR #3793, opened 2026-09-09
feat(skills): export lifecycle metrics and measured cost windows

TimeToBuildBob:
Explicit skill invocations currently persist lifecycle events without operational metrics or cost attribution. This adds invocation/terminal counters, duration, tokens, and cost through the existing OpenTelemetry exporter, emitted only after a successful ledger append. Duplicate terminal calls cannot double-count, and exporter failures cannot break invocation.

Usage is the inclusive admission-to-terminal delta from the same conversation's `CostTracker`. Missing, reset, foreign, or invalid accounting stays unknown (`usage: null`, `usage_available=false`), while measured zero remains distinguishable. Metric labels exclude session/invocation IDs, paths, and error details. Schema 2 adds accounting fields and reads schema 1 history without rewriting it; older writers must be upgraded before sharing the newer ledger.

Frontend accounting is still incomplete: TUI does not initialize the tracker and server request workers do not consistently retain it. Their costs remain unknown. Overlapping invocation windows may share costs; this is not exclusive skill billing. A completed event remains a runtime response boundary, not proof that skill instructions succeeded.

Validation:

- 18 new tests cover real in-memory OTel aggregates, duplicate terminals, unknown/invalid usage, exporter failure, and schema compatibility.
- Scoped `make test`: 317 passed; three telemetry tests hit the parallel 10-second timeout and all passed standalone in 3.43s.
- Changed-file mypy and commit hooks pass. Full `make typecheck` reports five existing ACP SDK errors in unchanged files.
- Real CLI `/skill:cost-smoke` returned `SKILL_COST_OK`; its existing CLI reconciliation produced one abandoned terminal with 12,042 input tokens, 16 output tokens, 11,904 cache-read tokens, and $0.0072232 across two requests.


TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `gptme/lessons/skill_events.py:196` | P1 | rejected — gptme/lessons/skill_commands.py:98-106 synchronously performs started -> queue_prompt -> queued before yielding; docs/sk |
| `gptme/lessons/skill_events.py:57` | P2 | rejected — gptme/lessons/skill_events.py:53-60 deliberately compares the tracker to str(logdir.resolve()); the runtime caller estab |

</details>




This PR adds cost and usage accounting to skill lifecycle events. It bumps the skill event schema to version 2, adding cost_tracker_id, cost_baseline, and usage fields, and computes usage deltas from the CostTracker between invocation start and terminal phase. It also adds OpenTelemetry metrics for skill invocations, completions, durations, tokens, and cost, exported after successful ledger appends. The CostTracker's SessionCosts gains a tracking_id to detect resets. Tests cover metric agreement, unknown usage, schema compatibility, and exporter failure handling.

Confidence 3/5

<details><summary><b>2 findings</b> · ❌ 1 P1 · ⚠️ 1 P2</summary>

**❌ P1 high** — `gptme/lessons/skill_events.py:196`

The `_usage_delta` function computes the delta between the current cost snapshot and the baseline stored at start. However, the baseline is taken at `start_skill_invocation` time, which is before the skill is queued. If the skill is queued and then the user runs other non-skill LLM requests before the skill completes, those costs are included in the delta, attributing them to the skill. The PR description says 'inclusive admission-to-terminal window' but admission is the start event, not the queued event. The test `test_ledger_and_metrics_agree_without_duplicate_terminal_counts` records cost after queued, but does not test the case where cost is recorded between start and queued. This is a correctness issue: the usage window is too wide, over-attributing costs to skills. The consequence is that skill cost metrics are inflated by unrelated LLM activity that happens between the start and queued events. The fix would be to take the baseline at the queued transition, but that would require storing a separate baseline for queued events. Since the start event is the first event, and the queued event is a separate append, the baseline should be captured at the queued phase, not at start. However, the current design uses the first event's baseline for all transitions. This is a design trade-off, but the PR description explicitly says 'admission-to-terminal' and admission is the start. The start event is recorded when the prompt is built, which is before queueing. The queued event is when it is admitted to the queue. The cost window should be from queued to terminal, not from start to terminal. This is a real bug because it over-counts costs. The severity is P1 because it misattributes costs in normal use.

<sub>**How this was verified:** Checked the flow: start_skill_invocation records baseline at line 146, then _transition uses first.cost_baseline. The queued event is a separate append but does not update the baseline. So any cost recorded between start and queued is included in the delta.</sub>

**⚠️ P2 medium** — `gptme/lessons/skill_events.py:57`

In `_cost_snapshot`, the session_id comparison uses `costs.session_id != str(logdir.resolve())`. However, `CostTracker.start_session` is called with a session_id that may be a path string, but the code compares it to the resolved logdir. If the session was started with a non-resolved path (e.g., a relative path or a path with symlinks), the comparison fails and usage is always unknown. The test uses `str(tmp_path.resolve())` so it passes, but in real usage, the session_id might be set to the logdir as passed by the caller, which may not be resolved. The PR description says 'same conversation' but the comparison is strict. This is a potential bug: if the session_id is not exactly the resolved path, usage is never available. The consequence is that skill cost metrics are always unknown for sessions started with a different path representation. The fix would be to resolve both sides or compare using Path equality. This is a P2 because it depends on how callers initialize the tracker.

```suggestion
if costs is None or Path(costs.session_id).resolve() != logdir.resolve():
```

<sub>**How this was verified:** Checked CostTracker.start_session in cost_tracker.py: it sets SessionCosts(session_id=session_id) without resolving. Callers may pass a non-resolved path. The test always passes resolved path.</sub>

</details>

- `docs/skills.rst` — Adds documentation for explicit invocation telemetry, schema version 2, and cost/usage accounting semantics.
- `gptme/lessons/skill_events.py` — Adds cost snapshot and usage delta functions, schema version 2 fields, and telemetry export after appends.
- `gptme/telemetry.py` — Adds record_skill_event function to export skill lifecycle metrics with label allowlist.
- `gptme/util/_telemetry.py` — Adds skill metric instruments and registers them in get_telemetry_objects and init_telemetry.
- `gptme/util/cost_tracker.py` — Adds tracking_id field to SessionCosts to detect resets/resumes.
- `tests/test_skill_events.py` — Updates schema version assertion from 1 to 2 in existing test.
- `tests/test_skill_metrics.py` — Adds new test file covering skill metrics, usage deltas, unknown usage, and exporter failure.

</details>

| `022bd6fbc917` | ? | 0 | llm | 2026-09-09 21:35 UTC |

</details>

<sub>Reviewed `21f1787b9cbf` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 38s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/lessons/skill_events.py
```diff
@@ -43,7 +45,45 @@ class SkillEvent:
...
+    schema_version: int = 2
+    cost_tracker_id: str | None = None
+    cost_baseline: dict[str, int | float] | None = None
+    usage: dict[str, int | float] | None = None
+
+
+def _cost_snapshot(logdir: Path) -> tuple[str | None, dict[str, int | float] | None]:
+    """Unknown accounting is distinct from a measured zero-cost invocation."""
+    costs = CostTracker.get_session_costs()
+    if costs is None or costs.session_id != str(logdir.resolve()):
```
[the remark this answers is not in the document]
TimeToBuildBob: By design — both sides of the comparison use `.resolve()`: `_cost_snapshot` compares `costs.session_id != str(logdir.resolve())`, and callers are expected to initialize the tracker with a resolved path (the tests use `str(tmp_path.resolve())` throughout). If paths don't match (symlink, relative, different session), accounting stays `null` rather than producing a wrong measurement — the conservative behavior is intentional. The PR description documents this explicitly: "A reset, missing tracker, or foreign conversation produces `usage: null`".

TimeToBuildBob:
Greptile withdrew the sole finding after the path-identity invariant was clarified, so no code change or re-review was warranted. I resolved the withdrawn review thread; all 18 checks remain green. The stale 4/5 summary is the pre-withdrawal score.

TimeToBuildBob:
Rebased onto current `master` and force-pushed `21f1787b9`; the conflict in `docs/skills.rst` was resolved by preserving the new hidden Gallery toctree and placing the lifecycle telemetry section after it. GitHub now reports the PR mergeable.

Verified locally: 36 targeted lifecycle/metrics tests pass, changed-file Ruff and mypy pass, and the rendered docs structure reports 0 errors. Fresh CI is running on the rebased head.

gptme/lessons/skill_events.py
```diff
@@ -141,6 +193,7 @@ def _transition(
             else None
         ),
         error_type=error_type,
+        usage=_usage_delta(logdir, first) if phase in _TERMINAL else None,
```
TimeToBuildBob, line 196: ❌ **P1** — The `_usage_delta` function computes the delta between the current cost snapshot and the baseline stored at start. However, the baseline is taken at `start_skill_invocation` time, which is before the skill is queued. If the skill is queued and then the user runs other non-skill LLM requests before the skill completes, those costs are included in the delta, attributing them to the skill. The PR description says 'inclusive admission-to-terminal window' but admission is the start event, not the queued event. The test `test_ledger_and_metrics_agree_without_duplicate_terminal_counts` records cost after queued, but does not test the case where cost is recorded between start and queued. This is a correctness issue: the usage window is too wide, over-attributing costs to skills. The consequence is that skill cost metrics are inflated by unrelated LLM activity that happens between the start and queued events. The fix would be to take the baseline at the queued transition, but that would require storing a separate baseline for queued events. Since the start event is the first event, and the queued event is a separate append, the baseline should be captured at the queued phase, not at start. However, the current design uses the first event's baseline for all transitions. This is a design trade-off, but the PR description explicitly says 'admission-to-terminal' and admission is the start. The start event is recorded when the prompt is built, which is before queueing. The queued event is when it is admitted to the queue. The cost window should be from queued to terminal, not from start to terminal. This is a real bug because it over-counts costs. The severity is P1 because it misattributes costs in normal use.

gptme/lessons/skill_events.py
TimeToBuildBob, line 57: ⚠️ **P2** — In `_cost_snapshot`, the session_id comparison uses `costs.session_id != str(logdir.resolve())`. However, `CostTracker.start_session` is called with a session_id that may be a path string, but the code compares it to the resolved logdir. If the session was started with a non-resolved path (e.g., a relative path or a path with symlinks), the comparison fails and usage is always unknown. The test uses `str(tmp_path.resolve())` so it passes, but in real usage, the session_id might be set to the logdir as passed by the caller, which may not be resolved. The PR description says 'same conversation' but the comparison is strict. This is a potential bug: if the session_id is not exactly the resolved path, usage is never available. The consequence is that skill cost metrics are always unknown for sessions started with a different path representation. The fix would be to resolve both sides or compare using Path equality. This is a P2 because it depends on how callers initialize the tracker.

TimeToBuildBob:
@greptileai review

gptme/lessons/skill_events.py
TimeToBuildBob: Rejected after tracing the synchronous command path: `start_skill_invocation()` is immediately followed by `queue_prompt()` and then the `queued` append in the same handler, before control returns to the chat loop. No LLM request can run between those points, so there is no normal unrelated-cost window to exclude. Here “admission-to-terminal” deliberately starts at the accepted explicit invocation (`started`); `queued` records successful queue persistence, not a later scheduler admission. Overlapping costs after the handler returns remain an explicitly documented limitation of window accounting, not exclusive skill billing.

gptme/lessons/skill_events.py
TimeToBuildBob: Rejected as the same already-withdrawn path-identity finding. The production CLI starts `CostTracker` from the canonical conversation path, and `_cost_snapshot()` intentionally requires that exact canonical identity; a relative, symlinked, reset, or foreign tracker must produce unknown accounting rather than risk attributing another conversation's costs. This conservative invariant is documented and covered by the unavailable-accounting tests. Greptile independently withdrew this finding on the same lines after rechecking the contract.

TimeToBuildBob:
Self-merge is not eligible: the policy gate requires a second requested AI-review pass and human product judgment for an untracked feature with no linked issue. CI is green, GitHub reports the head clean/mergeable, Greptile is 5/5, and all review threads are resolved. Waiting for maintainer review or merge.

PR #3800, opened 2026-09-10
fix(llm): keep tool_call pairing for unavailable tools; dedupe auto-naming threads

TimeToBuildBob:
## What

Two bugs from one DeepSeek V4.1 Flash session (reported by Erik, 2026-09-10):

**1. Orphan tool result → provider 400 that kills the session.** When the model called a tool that is not loaded (`read` without `--tools +read`), `extract_tool_uses_from_assistant_message` dropped the call (`is_runnable` filter) while `execute_msg` still emitted the paired error result (`Tool 'read' is not available for execution.`, with `call_id`). The next request carried a `tool` message with no `tool_calls` parent; DeepSeek rejects that on every subsequent request:

```
Messages with role 'tool' must be a response to a preceding message with 'tool_calls'
```

Fix: keep every tool use that carries a `call_id` (the API emitted it as a real tool_call), whether or not the tool is runnable; markdown blocks (no `call_id`) still require a runnable tool. Belt and braces: `_handle_tools` demotes any tool result whose `call_id` never appeared in a preceding `tool_calls` list to a `user` message instead of sending it as `role=tool`.

**2. Three auto-generated names for one conversation.** `chat.py` spawned a new naming thread on every step until the first name was saved, so a conversation with a few quick tool steps logged `Auto-generated conversation name:` three times with three different names. Fix: one in-flight naming thread per logdir; `try_auto_name` re-reads the config from disk before saving so the first name wins.

## Tests

- `test_handle_tools_keeps_pairing_for_unavailable_tool_call`
- `test_handle_tools_demotes_orphan_tool_result_to_user_text`
- `test_try_auto_name_keeps_name_saved_concurrently`

`tests/test_llm_openai.py tests/test_llm_utils.py tests/test_auto_naming.py tests/test_tools.py`: 275 passed, 1 skipped. ruff + mypy clean on the changed files.

Same strict-ordering class as #3422.

### DIFF
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -57,6 +57,36 @@ logger = logging.getLogger(__name__)
 
 # logdir -> in-flight auto-naming thread (see the naming block in step loop)
 _naming_threads: dict[Path, threading.Thread] = {}
+_naming_threads_lock = threading.Lock()
+
+
+def _start_auto_naming_thread(
+    logdir: Path,
+    chat_config: ChatConfig,
+    messages: list[Message],
+    model: str,
+) -> None:
+    """Start at most one auto-naming worker for *logdir*."""
+
+    def _run() -> None:
+        try:
+            try_auto_name(chat_config, messages, model)
+        finally:
+            with _naming_threads_lock:
+                if _naming_threads.get(logdir) is threading.current_thread():
+                    del _naming_threads[logdir]
+
+    with _naming_threads_lock:
+        if logdir in _naming_threads:
+            return
+        thread = threading.Thread(target=_run, daemon=True)
+        _naming_threads[logdir] = thread
+        try:
+            thread.start()
+        except Exception:
+            del _naming_threads[logdir]
+            raise
+
 
 # Store an immutable int. copy_context() (server/TUI/ACP step threads) copies
 # the binding, not the value: a list would be shared by reference and a child
@@ -565,19 +595,13 @@ def _process_message_conversation(
             # guard every step before the first name lands spawns another
             # thread and the log shows three "Auto-generated conversation
             # name" lines for one conversation.
-            inflight = _naming_threads.get(manager.logdir)
-            if not chat_config.name and not (inflight and inflight.is_alive()):
-                thread = threading.Thread(
-                    target=try_auto_name,
-                    args=(
-                        chat_config,
-                        copy.deepcopy(manager.log.messages),
-                        current_model.full,
-                    ),
-                    daemon=True,
+            if not chat_config.name:
+                _start_auto_naming_thread(
+                    manager.logdir,
+                    chat_config,
+                    copy.deepcopy(manager.log.messages),
+                    current_model.full,
                 )
-                _naming_threads[manager.logdir] = thread
-                thread.start()
 
         # Check step limit (GPTME_MAX_STEPS)
         step_count += 1
--- a/gptme/chat.py
+++ b/gptme/chat.py
@@ -55,6 +55,9 @@ from .util.terminal import flush_stdin, set_current_conv_name, terminal_state_ti
 
 logger = logging.getLogger(__name__)
 
+# logdir -> in-flight auto-naming thread (see the naming block in step loop)
+_naming_threads: dict[Path, threading.Thread] = {}
+
 # Store an immutable int. copy_context() (server/TUI/ACP step threads) copies
 # the binding, not the value: a list would be shared by reference and a child
 # step() would mutate the parent's running total. Rebinding via .set() keeps
@@ -557,7 +560,13 @@ def _process_message_conversation(
         assistant_count = sum(1 for m in manager.log.messages if m.role == "assistant")
         if current_model and 1 <= assistant_count <= MAX_ASSISTANT_MSGS_FOR_NAMING:
             chat_config = ChatConfig.from_logdir(manager.logdir)
-            if not chat_config.name:
+            # One naming call per conversation: the thread runs across several
+            # loop iterations (each tool step re-enters here), so without this
+            # guard every step before the first name lands spawns another
+            # thread and the log shows three "Auto-generated conversation
+            # name" lines for one conversation.
+            inflight = _naming_threads.get(manager.logdir)
+            if not chat_config.name and not (inflight and inflight.is_alive()):
                 thread = threading.Thread(
                     target=try_auto_name,
                     args=(
@@ -567,6 +576,7 @@ def _process_message_conversation(
                     ),
                     daemon=True,
                 )
+                _naming_threads[manager.logdir] = thread
                 thread.start()
 
         # Check step limit (GPTME_MAX_STEPS)
--- a/gptme/config/chat.py
+++ b/gptme/config/chat.py
@@ -4,13 +4,15 @@ ChatConfig manages per-conversation settings including model selection,
 tool configuration, workspace paths, and agent settings.
 """
 
+import importlib
 import logging
 import os
 import sys
 import tempfile
+from contextlib import contextmanager
 from dataclasses import asdict, dataclass, field, fields, replace
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import tomlkit
 from tomlkit.exceptions import TOMLKitError
@@ -34,6 +36,40 @@ if TYPE_CHECKING:
     from ..tools.base import ToolFormat
 
 logger = logging.getLogger(__name__)
+
+try:
+    fcntl: Any = importlib.import_module("fcntl")
+except ImportError:  # pragma: no cover - Windows
+    fcntl = None
+
+try:
+    msvcrt: Any = importlib.import_module("msvcrt")
+except ImportError:  # pragma: no cover - POSIX
+    msvcrt = None
+
+
+@contextmanager
+def chat_config_lock(logdir: Path):
+    """Serialize config.toml read-modify-write operations for one chat."""
+    lock_path = logdir / ".config.lock"
+    lock_path.parent.mkdir(parents=True, exist_ok=True)
+    with lock_path.open("a+b") as lock:
+        if fcntl is not None:
+            fcntl.flock(lock, fcntl.LOCK_EX)
+        elif msvcrt is not None:  # pragma: no cover - Windows
+            if lock.seek(0, os.SEEK_END) == 0:
+                lock.write(b"\0")
+                lock.flush()
+            lock.seek(0)
+            msvcrt.locking(lock.fileno(), msvcrt.LK_LOCK, 1)
+        try:
+            yield
+        finally:
+            if fcntl is not None:
+                fcntl.flock(lock, fcntl.LOCK_UN)
+            elif msvcrt is not None:  # pragma: no cover - Windows
+                lock.seek(0)
+                msvcrt.locking(lock.fileno(), msvcrt.LK_UNLCK, 1)
 
 
 def _coerce_config_path(value: object, field_name: str) -> Path:
--- a/gptme/util/auto_naming.py
+++ b/gptme/util/auto_naming.py
@@ -308,6 +308,21 @@ def auto_generate_display_name(messages: list[Message], model: str) -> str | Non
     return generate_conversation_name(strategy="llm", messages=messages, model=model)
 
 
+def _config_has_name_on_disk(config: ChatConfig) -> bool:
+    """Re-read *config* from its logdir; True if a name was saved meanwhile."""
+    from pathlib import Path  # runtime use; avoids the config import cycle
+
+    from ..config import ChatConfig as _ChatConfig
+
+    logdir = getattr(config, "_logdir", None)
+    if not logdir:
+        return False
+    try:
+        return bool(_ChatConfig.from_logdir(Path(logdir)).name)
+    except Exception:  # pragma: no cover - never block naming on a read error
+        return False
+
+
 def try_auto_name(
     config: ChatConfig,
     messages: list[Message],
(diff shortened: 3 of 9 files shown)


TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `gptme/util/auto_naming.py:345` | P1 | superseded by latest review (not reproduced) |
| `gptme/chat.py:579` | P2 | fixed in `b65969538` |
| `gptme/config/chat.py:52` | P2 | accepted-tradeoff |
| `gptme/util/auto_naming.py:348` | P2 | rejected |
| `tests/test_llm_openai.py:3900` | P2 | fixed in `315dfe2c1` |

</details>




### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `gptme/config/chat.py:52`

The chat_config_lock uses fcntl.flock on a lock file, but ChatConfig.save() writes config.toml via a temp file and atomic rename (temp_path.replace(chat_config_path)). The lock file is separate from config.toml, so the lock serializes writers that all use chat_config_lock. However, rename_conversation and try_auto_name both use the lock, but other code paths that write config.toml (e.g., ChatConfig.save() called from elsewhere, or the server API config update endpoint) do not acquire chat_config_lock. The PR only wraps the auto-namer and rename_conversation. If any other caller writes config.toml without the lock, the lock provides no protection and a concurrent write can still be lost. The diff shows no other callers updated. This is a contract/coverage gap: the lock is not used by all writers, so the race it aims to fix (auto-namer overwriting a user rename) is only fixed for rename_conversation, not for other config writes like the server's PATCH /config endpoint. The PR description says 'preserve config overrides during auto-naming' but the lock is not applied to all config writes.

<sub>**How this was verified:** Searched the diff and full files: only rename_conversation and try_auto_name use chat_config_lock. ChatConfig.save() itself does not acquire the lock. Other callers of ChatConfig.save() (e.g., server API) are not in the diff.</sub>

</details>

| `75106d3542ba` | 4/5 | 1 | llm | 2026-09-10 08:21 UTC |
| `b65969538bc9` | 4/5 | 2 | llm | 2026-09-10 10:04 UTC |
| `95c1d511f7dd` | 5/5 | 0 | llm | 2026-09-10 11:34 UTC |
| `95c1d511f7dd` | 3/5 | 1 | llm | 2026-09-10 11:51 UTC |

</details>

<sub>Reviewed `36b0974031da` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 34s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/chat.py
```diff
@@ -567,6 +576,7 @@ def _process_message_conversation(
                     ),
                     daemon=True,
                 )
+                _naming_threads[manager.logdir] = thread
```
TimeToBuildBob: ⚠️ **P2** — The `_naming_threads` dict in chat.py is keyed by `manager.logdir` and entries are never removed after the thread finishes. The guard checks `inflight and inflight.is_alive()`, so a finished thread's entry remains in the dict but is not considered alive, allowing a new thread to be spawned. However, the dict grows unboundedly across many conversations in a long-lived process (e.g. the server), leaking one Thread object per logdir ever seen. More importantly, if a logdir is reused (e.g. a conversation is deleted and recreated with the same path, or a subagent reuses the same logdir), the stale entry is harmless because is_alive() is False, but the memory leak remains. The dict should delete the entry when the thread completes (e.g. a finally in the thread target or a wrapper that pops the key).

gptme/util/auto_naming.py
```diff
@@ -332,6 +347,12 @@ def try_auto_name(
...
         display_name = auto_generate_display_name(messages, model)
         if display_name:
+            # Another caller (a concurrent naming thread, the server, or the
+            # user via /rename) may have named the conversation while the
+            # LLM call was in flight; the first name wins.
+            if _config_has_name_on_disk(config):
+                logger.debug("Conversation was named concurrently; keeping it")
+                return None
             config.name = display_name
             config.save()
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `b65969538`: auto-naming now holds a per-conversation file lock across the persisted-name check and save, and both the CLI/session rename path and conversation-list rename path take the same lock. The user rename therefore wins without a check-then-write window. The config/rename and auto-naming test slices pass.

gptme/chat.py
```diff
@@ -55,6 +55,9 @@
 
 logger = logging.getLogger(__name__)
 
+# logdir -> in-flight auto-naming thread (see the naming block in step loop)
+_naming_threads: dict[Path, threading.Thread] = {}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `b65969538`: the registry is now lock-protected, registers before starting the worker, deduplicates every live worker, and removes its own entry in `finally`. Added a regression test that verifies both in-flight deduplication and cleanup after completion.

gptme/chat.py
TimeToBuildBob: Fixed in commit b65969538: completed workers remove their own registry entry in `finally` under the registry lock, with a regression test for in-flight deduplication and cleanup.

TimeToBuildBob:
@greptileai review

gptme/util/auto_naming.py
```diff
@@ -332,8 +331,22 @@ def try_auto_name(
...
+                with chat_config_lock(Path(logdir)):
+                    current = _ChatConfig.from_logdir(Path(logdir))
+                    if current.name:
+                        logger.debug("Conversation was named concurrently; keeping it")
+                        return None
+                    current.name = display_name
+                    current.save()
+                    config.name = display_name
+            else:
+                config.name = display_name
```
TimeToBuildBob, line 347: ⚠️ **P2** — In try_auto_name, when the config has no _logdir (the else branch), the code still calls config.save() without holding chat_config_lock. This is the non-logdir path used by some callers (e.g. tests or configs constructed without a logdir). If two threads call try_auto_name concurrently on the same config object without a logdir, both can generate names and both call config.save(), causing a lost update. However, the PR's main fix is for the logdir path; the non-logdir path is a fallback. The more concrete issue is that the else branch does not re-read the config from disk, so a concurrent rename via rename_conversation (which uses the lock) could be overwritten. But rename_conversation requires a logdir, so this only matters for configs without a logdir, which are not persisted to a conversation directory. This is a lower-severity edge case.

tests/test_llm_openai.py
```diff
@@ -3853,3 +3853,71 @@ def test_no_proxy_url_returns_false(self, monkeypatch):
...
+
+
+def test_handle_tools_demotes_orphan_tool_result_to_user_text():
+    """A tool result with no preceding tool_call is sent as plain text, not role=tool."""
+    init_tools(allowlist=["shell"])
+
+    messages = [
+        Message(role="user", content="hi"),
+        # Result whose assistant call was lost (old log / interrupted turn)
+        Message(role="system", content="stale output", call_id="call_gone"),
```
TimeToBuildBob, line 3903: ⚠️ **P2** — The test test_handle_tools_demotes_orphan_tool_result_to_user_text uses messages where the orphan tool result appears before any assistant tool_calls, so after_tool_calls is False and the demoted message is yielded immediately. It does not cover the case where the orphan appears after an assistant tool_calls message (after_tool_calls=True) and is the last message, which would expose the lost-message bug. The test is not defective per se, but it misses the edge case. However, per the three-clause rule, I should not file a test-coverage finding unless I can name a specific input that produces wrong behavior. I have already named that input in the bug finding above, so this is covered by the bug finding.

TimeToBuildBob: Fixed in 315dfe2c1: the regression now puts an orphan result after an unresolved assistant tool call and at end-of-history, exercises the buffered path, and asserts the demoted user message is retained.

gptme/util/auto_naming.py
TimeToBuildBob: Rejected as a false positive: `ChatConfig.save()` raises when `_logdir` is absent, while every production `try_auto_name` caller receives its config from `ChatConfig.from_logdir` or `load_or_create`. A config without `_logdir` has no conversation file that `rename_conversation` could concurrently overwrite.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `315dfe2c1` to cover the buffered end-of-history orphan-result path identified by the fresh AI review. The two focused tool-pairing tests pass, and pre-commit passed.

I also disposed both review threads: the test gap is fixed; the no-`_logdir` naming concern is a false positive because `ChatConfig.save()` rejects that state and production auto-naming callers load configs from a conversation logdir. Fresh CI and Greptile review are now running for the new head.

tests/test_llm_openai.py
```diff
@@ -3853,3 +3853,71 @@ def test_no_proxy_url_returns_false(self, monkeypatch):
...
+        # Put it after an unrelated tool call to cover the buffered path.
+        Message(role="system", content="stale output", call_id="call_gone"),
+    ]
+
+    tool_shell = get_tool("shell")
+    assert tool_shell
+    model = get_model("openai/gpt-4o")
+    messages_dicts, _ = _prepare_messages_for_api(messages, model.full, [tool_shell])
+
+    assert [m["role"] for m in messages_dicts] == ["user", "assistant", "user"]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `95c1d511f`: the buffered-path regression now includes the matching result for `call_live`, asserts that valid result remains paired, and verifies the orphan is demoted only after the valid tool-response run. The focused tests and pre-commit pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI is green, Greptile is 5/5 on `95c1d511f`, and all review threads are resolved. The self-merge gate still refuses this PR because it changes core config schema plus `gptme/llm/` provider-message handling, which requires maintainer review under policy. No direct merge attempted.

gptme/util/auto_naming.py
```diff
@@ -332,8 +331,22 @@ def try_auto_name(
...
+                from ..config import ChatConfig as _ChatConfig
+                from ..config.chat import chat_config_lock
+
+                with chat_config_lock(Path(logdir)):
+                    current = _ChatConfig.from_logdir(Path(logdir))
+                    if current.name:
+                        logger.debug("Conversation was named concurrently; keeping it")
+                        return None
+                    current.name = display_name
+                    current.save()
```
TimeToBuildBob: ❌ **P1** — In try_auto_name, the new code saves the freshly-loaded `current` config (from `_ChatConfig.from_logdir(Path(logdir))`) instead of the `config` object passed by the caller. If a caller (e.g. api_v2_sessions.py, which calls try_auto_name synchronously) has modified the in-memory `config` (for example setting a different workspace, model, or tools) before calling auto-naming, those modifications are silently discarded when the name is written, because `current.save()` writes only what was read from disk plus the new name. The old code called `config.save()`, which persisted the caller's current object and preserved those fields. Concretely, a server handler that creates a config via `ChatConfig.from_logdir`, overrides `config.workspace` or `config.model` for the session, and then invokes auto-naming would lose those overrides after the first successful naming. This is a regression for non-CLI callers. _(3/3 passes)_

TimeToBuildBob: Fixed in `36b097403`: auto-naming still re-reads under the conversation lock to honor a concurrent name, but now applies the generated name to and saves the caller-provided `config`, preserving unsaved model/tool/workspace overrides. Added a regression test that starts with persisted settings, mutates model/tools in memory, auto-names, and verifies all three fields on disk.

gptme/config/chat.py
```diff
@@ -35,6 +37,40 @@
...
+    fcntl = None
+
+try:
+    msvcrt: Any = importlib.import_module("msvcrt")
+except ImportError:  # pragma: no cover - POSIX
+    msvcrt = None
+
+
+@contextmanager
+def chat_config_lock(logdir: Path):
```
TimeToBuildBob, line 52: ⚠️ **P2** — The chat_config_lock uses fcntl.flock on a lock file, but ChatConfig.save() writes config.toml via a temp file and atomic rename (temp_path.replace(chat_config_path)). The lock file is separate from config.toml, so the lock serializes writers that all use chat_config_lock. However, rename_conversation and try_auto_name both use the lock, but other code paths that write config.toml (e.g., ChatConfig.save() called from elsewhere, or the server API config update endpoint) do not acquire chat_config_lock. The PR only wraps the auto-namer and rename_conversation. If any other caller writes config.toml without the lock, the lock provides no protection and a concurrent write can still be lost. The diff shows no other callers updated. This is a contract/coverage gap: the lock is not used by all writers, so the race it aims to fix (auto-namer overwriting a user rename) is only fixed for rename_conversation, not for other config writes like the server's PATCH /config endpoint. The PR description says 'preserve config overrides during auto-naming' but the lock is not applied to all config writes. _(3/3 passes)_

TimeToBuildBob: Accepted trade-off for this focused fix: the lock coordinates the two writers that race over the conversation name (`try_auto_name` and `rename_conversation`). The server config PATCH already rejects updates while generation is active under `SessionManager.conversation_lock`, and auto-naming runs before that generation reservation is released, so it cannot race this auto-name save. Moving the lock into every `ChatConfig.save()` call would broaden the PR into a global config-write locking change and risks recursive locking around the two explicit read-modify-write sections. The concrete P1—saving the wrong in-memory config—was fixed in `36b097403` and has a regression test.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the new P1 in `36b097403`: auto-naming now saves the caller-provided config under the name compare-and-save lock, preserving unsaved model/tool/workspace overrides. Added regression coverage; the focused auto-naming/chat/rename slice passes (62 passed, 3 skipped), pre-commit passed, and the in-band consensus review has no open P0/P1 findings. Its remaining global-lock P2 was dispositioned as an accepted scope trade-off because server config PATCHes cannot run while this generation-owned auto-name path is active. CI and the automatic Greptile review are still pending on the new head.

PR #3801, opened 2026-09-10
feat(tools): conditional tool docs + companion tools (requires_tools)

TimeToBuildBob:
## Problem

Tool instructions and examples often describe how tools interact: "fetch with `read`", "use `hashline_edit` after `read`", "use the `screenshot` tool". That text is only true when the other tool is loaded. In a default-toolset session today a DeepSeek V4.1 Flash model called the disabled `read` tool because two default-on tools (`chats`, `vision`) named it in their prompt text (and then hit #3800's orphan-result 400). Examples that show two tools interacting are the most useful kind, but were impossible to justify because they are only valid when both are loaded.

## Mechanism

**Conditional blocks** in `instructions`, `instructions_format` and `examples`, rendered against the loaded toolset at prompt-build time:

```
Do **not** use vision for:
{% if tools: read, browser %}
- Images at a URL — fetch with `read` or visit with `browser` instead
{% elif tools: read %}
- Images at a URL — fetch with `read` first, then pass the local path
{% endif %}
```

A branch is taken when *all* listed tools are loaded; `elif`/`else` as expected; no nesting. A marker on its own line removes the whole line so lists stay tidy. Generated docs (`get_doc`) render every branch as if all tools were loaded, so the interaction examples still show up in the documentation. Text without markers is untouched.

**Companion tools**: `ToolSpec.requires_tools`. The loader appends required tools (to a fixpoint) even when they are `disabled_by_default`, and warns if one is unavailable. `hashline_edit` now declares `requires_tools=["read"]`, so enabling hashline gives the model the tool its docs tell it to call first.

## Applied

- `chats`: the "use the shell or read tool" / "use the browser tool" lines render only for loaded tools.
- `vision`: `read`/`browser`/`screenshot` mentions conditional.
- `hashline_edit`: requires `read`.
- Docs: new "Tools that reference other tools" section in `docs/tools.rst`.

## Tests

`tests/test_tool_conditionals.py` (render semantics, malformed blocks, prompt rendering with/without a tool loaded, vision docs omit `read` when off, docs render all branches, hashline pulls in read, shell alone does not). Ran with `test_tools`, `test_prompts`, `test_prompt_tools`, `test_llm_openai`, `test_capabilities_export`: 289 passed. ruff + mypy clean.

Follow-up candidates (not in this PR): wrap the other cross-tool mentions (`save`↔`patch`, `screenshot`↔`browser`/`vision`, `tmux`→`ipython`), and consider whether `read`+`hashline_edit` should become the default edit toolset (TUA-bench notes suggest hashline suits fast models).


TimeToBuildBob:
## 🤖 AI code review

Adds conditional tool-doc rendering ({% if tools: ... %} blocks) in ToolSpec instructions/examples, a new requires_tools companion-tool mechanism on ToolSpec, and applies both to chats/vision/hashline_edit. Adds tools_initialized() context flag and rewires load_tool() to resolve companion closures, with /tools load passing allow_required=True. Adds tests for rendering and companion loading.

Confidence 5/5

- `docs/tools.rst` — Adds a section documenting conditional tool-doc blocks and requires_tools companion declarations.
- `gptme/commands/llm.py` — Changes /tools load to call load_tool(tool_name, allow_required=True).
- `gptme/tools/__init__.py` — Adds _tools_initialized_var, tools_initialized(), _add_required_tools(), _resolve_available_tool_requirements(), and rewires init_tools/get_toolchain/load_tool to resolve requires_tools closures.
- `gptme/tools/base.py` — Adds render_tool_conditionals() and _loaded_tool_names(), applies conditional rendering in get_doc/get_instructions/get_examples, and adds requires_tools field to ToolSpec.
- `gptme/tools/chats.py` — Wraps cross-tool mentions in chats instructions with conditional blocks.
- `gptme/tools/hashline_edit.py` — Adds requires_tools=['read'] to the hashline_edit ToolSpec.
- `gptme/tools/vision.py` — Wraps cross-tool mentions in vision instructions with conditional blocks.
- `tests/test_commands_llm.py` — Adds assertion that /tools load calls load_tool with allow_required=True.
- `tests/test_tool_conditionals.py` — Adds tests for conditional rendering and requires_tools companion loading.

</details>

| `2f60e05199a8` | 3/5 | 3 | llm | 2026-09-10 09:45 UTC |
| `d9f096b71e1c` | 3/5 | 4 | llm | 2026-09-11 14:40 UTC |
| `2e9392e4490f` | 3/5 | 1 | llm | 2026-09-12 03:42 UTC |
| `40085c045f56` | 5/5 | 0 | llm | 2026-09-12 10:10 UTC |
| `40085c045f56` | 2/5 | 2 | llm | 2026-09-12 10:21 UTC |
| `b4ba952a894d` | 2/5 | 4 | llm | 2026-09-12 12:04 UTC |
| `54b6352ab83e` | 5/5 | 0 | llm | 2026-09-13 13:22 UTC |
| `f57f6a48f83f` | 5/5 | 0 | llm | 2026-09-13 14:21 UTC |

</details>

<sub>Reviewed `f57f6a48f83f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 70s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/__init__.py
```diff
@@ -339,6 +340,40 @@ def get_toolchain(
...
+            if dep is None or not dep.is_available:
+                logger.warning(
+                    "Tool '%s' requires '%s', which is not available; "
+                    "its instructions may reference a missing tool.",
+                    tool.name,
+                    name,
+                )
+                continue
+            logger.info("Loading '%s' because '%s' requires it", name, tool.name)
+            tools.append(dep)
```
TimeToBuildBob, line 391: ❌ **P1** — In _add_required_tools, the queue is initialized with tools that have requires_tools, but the loop pops from the queue and processes the requires_tools of each tool. However, when a dependency is added, its own requires_tools are only queued if dep.requires_tools is truthy. This is correct. But there is a subtle bug: the queue is a list, and the loop pops from the end (LIFO). The order of loading is not deterministic, but that's not a correctness issue. The real issue is that the function appends the dependency to the tools list, but does not check whether the dependency is already in the tools list by identity or by name. It checks `if name in loaded`, which is a set of names, so it won't add a duplicate. However, if the same tool is added via two different paths (e.g. two tools both require the same companion), the first addition adds it to loaded, and the second skips. That's fine. But what if the dependency is already in the tools list but not in the loaded set? That can't happen because loaded is built from tools. So the logic is correct. The real issue is that the function does not handle the case where a required tool is itself disabled_by_default and not explicitly allowed. The PR description says "even when they are disabled_by_default", and the code does append it regardless of disabled_by_default. That is intended. But there is a security consideration: if a user explicitly disables a tool (e.g. --tools -read), and another tool requires it, the loader will re-enable it, violating the user's explicit choice. This is a contract violation: the allowlist is supposed to be a hard cap, but requires_tools bypasses it. The test test_requires_tools_loads_companion_even_if_disabled_by_default confirms this behavior, but it doesn't test the case where the user explicitly excludes the companion. For example, `get_toolchain(["shell", "hashline_edit"])` loads read, but if the user runs `gptme --tools shell,hashline_edit,-read`, the allowlist expansion would remove read, but _add_required_tools would add it back. This is a real issue: the companion mechanism overrides explicit user exclusions, which could be a security concern (e.g. a read-only profile that excludes read? unlikely, but a user might exclude a tool for safety). The severity is P1 because it violates the allowlist contract.

gptme/tools/base.py
```diff
@@ -366,6 +366,95 @@ def callable_signature(func: Callable) -> str:
...
+            text[line_start:start].strip() == "" and text[end:line_end].strip() == ""
+        )
+        chunk = text[pos:start]
+        if state is None or state[1]:
+            out.append(chunk)
+        if whole_line:
+            # drop the marker line (including its newline, or the preceding one at EOF)
+            if state is None or state[1]:
+                # chunk already emitted up to the marker; trim the partial line
+                out[-1] = out[-1][: len(out[-1]) - (start - line_start)]
```
TimeToBuildBob, line 420: ⚠️ **P2** — In render_tool_conditionals, the whole-line marker handling trims the emitted chunk using out[-1] = out[-1][: len(out[-1]) - (start - line_start)]. When the marker is the first thing in the text (line_start == 0, start == 0), this trims zero characters, which is fine. But when the marker is on a line that also contains preceding text that was already emitted as part of a previous chunk, the trim removes (start - line_start) characters from the END of the last emitted chunk, not from the beginning of the current line. If the previous chunk ended with a newline, the trim removes the newline and possibly part of the previous line, corrupting the output. Concretely, text = "line1\n{% if tools: read %}\ncontent\n{% endif %}" with loaded={"read"}: the first chunk is "line1\n", then the marker is whole-line, so out[-1] becomes "line1" (trimming the newline), then the content chunk "\ncontent\n" is appended, yielding "line1\ncontent\n" — the newline between line1 and content is lost, joining "line1content". The test test_render_whole_line_markers_take_their_line uses a marker preceded by a newline and followed by a newline, so the trim removes the preceding newline correctly, but the case where the marker is at the very start of the text (no preceding newline) is not covered. This is a real bug in the rendering logic that produces malformed prompts for tools whose instructions begin with a conditional block on the first line.

tests/test_tool_conditionals.py
```diff
@@ -0,0 +1,91 @@
...
+        instructions="Fetch with {% if tools: read %}`read`{% else %}`shell`{% endif %}.",
+    )
+    init_tools(allowlist=["shell"])
+    assert spec.get_instructions("markdown") == "Fetch with `shell`."
+    init_tools(allowlist=["shell", "read"])
+    assert spec.get_instructions("markdown") == "Fetch with `read`."
+
+
+def test_vision_docs_do_not_name_read_when_read_is_off():
+    init_tools(allowlist=["shell", "ipython", "vision"])
```
TimeToBuildBob, line 74: ⚠️ **P2** — The test test_requires_tools_loads_companion_even_if_disabled_by_default calls get_toolchain(["shell", "hashline_edit"]) and asserts that "read" is in the resulting names. However, this test does not verify that the loaded read tool is actually initialized (its init() called, hooks registered). get_toolchain returns ToolSpec objects, but init_tools is what calls _init_single_tool. The test only checks the toolchain, not the loaded tools. This is a test gap, but not a defect in the test itself. However, the test test_vision_docs_do_not_name_read_when_read_is_off calls init_tools(allowlist=["shell", "ipython", "vision"]) and then checks vision.get_instructions. This test depends on the global tool state, which is context-local but shared across tests in the same context. If another test runs before it and leaves tools loaded, the assertion could fail. The test does not call clear_tools() before init_tools, so it is not hermetic. This is a test defect: it depends on the order of test execution and the state of the context. The test suite may pass in isolation but fail when run with other tests that modify the loaded tools. This is a real issue because the test file does not reset the tool state between tests. The test_render_* tests are pure functions and don't depend on state, but the init_tools-based tests do. I'll report this as a test-defect.

TimeToBuildBob:
@greptileai review

ErikBjare:
@TimeToBuildBob valid criticism from Greptile

TimeToBuildBob:
Fixed the two issues flagged by Greptile and confirmed by Erik:

1. **Screenshot guidance uses wrong gate** — `vision.py` line 157: changed `{% if tools: read %}` → `{% if tools: screenshot %}`. The guidance about using the screenshot tool now only renders when the screenshot tool is actually available.

2. **Loading paths skip companions** — `load_tool` now passes `already_loaded` to `_add_required_tools`, so companions that are already in the toolchain aren't flagged as missing. Added a test covering this case.

The allowlist boundary enforcement (raised ValueError for out-of-allowlist companions) was already in place from the prior commit.

23 tests pass locally.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -651,11 +709,19 @@ def load_tool(tool_name: str) -> ToolSpec:
...
+            [tool],
+            list(available.values()),
+            allowlist=None if allow_required else get_session_allowlist(),
+            already_loaded={spec.name for spec in get_tools()},
+        )
+        initialized: dict[str, ToolSpec] = {}
+        for spec in to_load:
+            if has_tool(spec.name):
+                continue
+            initialized_spec = _init_single_tool(spec)
```
TimeToBuildBob, line 722: ⚠️ **P2** — In load_tool, after _add_required_tools returns to_load, the function initializes each spec and appends to _get_loaded_tools(). But if _init_single_tool raises an exception for one of the companions (e.g., its init() function fails), the already-initialized tools remain in the loaded list, and the function propagates the exception without cleaning up. This leaves the session in a partially-loaded state: some companions loaded, the primary not loaded, and the user sees an error. The previous code initialized only the single tool, so a failure left nothing loaded. Now a failure mid-way through the companion list leaves partial state. This is a real error-handling defect: a failed companion init leaves the tool set inconsistent, and subsequent calls to has_tool(tool_name) will still return False for the primary, but the companion is loaded and its hooks/commands registered. The user can retry /tools load, but the companion is already loaded, so _add_required_tools will skip it (already_loaded includes it), and the primary will load — but the partial state may have side effects (e.g., a companion's init registered a hook that expects the primary).

tests/test_tool_conditionals.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def test_render_requires_all_listed_tools():
+    text = "{% if tools: read, browser %}both{% elif tools: read %}read{% endif %}"
+    assert render_tool_conditionals(text, {"read", "browser"}) == "both"
+    assert render_tool_conditionals(text, {"read"}) == "read"
+    assert render_tool_conditionals(text, {"browser"}) == ""
+
+
+def test_render_whole_line_markers_take_their_line():
```
TimeToBuildBob, line 36: ⚠️ **P2** — In test_tool_conditionals.py, test_render_whole_line_markers_take_their_line uses text = "- one\n{% if tools: read %}\n- two\n{% endif %}\n- three\n" and asserts that with loaded={'read'} the result is "- one\n- two\n- three\n". This passes because the markers are unindented. But the test does not cover indented markers, which is where the trim bug manifests. This is a test-coverage gap, but per the three-clause rule, I cannot file it as a finding because the input is not named as producing wrong behavior in the test (it does, but the test doesn't cover it). I already filed the bug above.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -339,6 +350,51 @@ def get_toolchain(
...
+    by_name = {t.name: t for t in available}
+    loaded = {t.name for t in tools} | (already_loaded or set())
+    queue = [t for t in tools if t.requires_tools]
+    while queue:
+        tool = queue.pop()
+        for name in tool.requires_tools:
+            if name in loaded:
+                continue
+            dep = by_name.get(name)
+            if dep is None or not dep.is_available:
```
TimeToBuildBob, line 420: ❌ **P1** — When a required companion tool is unavailable, `_add_required_tools` logs a warning and `continue`s, leaving the primary tool loaded without the companion. `hashline_edit` is the concrete case: its `instructions` unconditionally say "Use after `read`" (line 569 of hashline_edit.py), so the model is told to call `read` even though it was not loaded. The conditional rendering only removes branches that name the companion, but `hashline_edit`'s instructions contain no conditional block. The result is a wasted tool call to `read` that fails with "not available", and the tool cannot perform its snapshot-anchored edits. The PR's own docs promise that a startup allowlist must include every required companion "otherwise initialization fails", yet this path silently continues, contradicting that contract. Fix either by raising `ValueError` when a required companion is unavailable (matching the allowlist failure behavior), or by guarding the tool's instructions with a `{% if tools: read %}` block so the misleading guidance is omitted.

gptme/tools/__init__.py
```diff
@@ -324,6 +324,7 @@ def get_toolchain(
             if not explicitly_allowed:
                 continue
         tools.append(tool)
+    tools = _add_required_tools(tools, get_available_tools(include_mcp=include_mcp))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 21c5131b8: companion resolution now receives the explicit startup allowlist and raises instead of widening the capability boundary.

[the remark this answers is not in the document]
TimeToBuildBob: Fixed across 21c5131b8 and d9f096b71: custom-file and dynamic loading both resolve companion closures; already-loaded companions are handled explicitly.

gptme/tools/vision.py
```diff
@@ -147,8 +147,16 @@ def view_image(image_path: "Path | str | Image.Image") -> Message:
...
 Do **not** use vision for:
+{% if tools: read, browser %}
 - Images at a URL — fetch with `read` or visit with `browser` instead
+{% elif tools: read %}
+- Images at a URL — fetch with `read` first, then pass the local path
+{% elif tools: browser %}
+- Images at a URL — visit with `browser` instead
+{% endif %}
+{% if tools: read %}
 - Taking a new screenshot — use the `screenshot` tool, then pass the path to vision
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in d9f096b71: screenshot guidance is gated on the screenshot tool, with regression coverage.

gptme/tools/__init__.py
```diff
@@ -651,11 +709,19 @@ def load_tool(tool_name: str) -> ToolSpec:
...
+            list(available.values()),
+            allowlist=None if allow_required else get_session_allowlist(),
+            already_loaded={spec.name for spec in get_tools()},
+        )
+        initialized: dict[str, ToolSpec] = {}
+        for spec in to_load:
+            if has_tool(spec.name):
+                continue
+            initialized_spec = _init_single_tool(spec)
+            _get_loaded_tools().append(initialized_spec)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f4c8984ce: the full dependency closure initializes companions-first before any specs are published to the active toolset. A failing companion leaves the active set unchanged; regression coverage included.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in f4c8984ce: unavailable required companions now fail closed with ValueError, matching the documented startup contract. Added regression coverage.

TimeToBuildBob: Fixed in f4c8984ce: unavailable required companions now raise before initialization, and the new regression test verifies that a failed companion leaves the active toolset empty.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -651,11 +705,27 @@ def load_tool(tool_name: str) -> ToolSpec:
...
+        # toolset.  A companion failure must not leave the requested tool loaded
+        # without its dependency (or make a retry fail as "already loaded").
+        initialized: dict[str, ToolSpec] = {}
+        # _add_required_tools appends dependencies after their dependants, so
+        # reverse the closure: companions must initialize before the tool that
+        # requires them.
+        for spec in reversed(to_load):
+            if has_tool(spec.name):
+                continue
+            initialized[spec.name] = _init_single_tool(spec)
```
TimeToBuildBob: ❌ **P1** — In load_tool, the closure is initialized in reverse order (companions first), but the initialized dict is keyed by spec.name and _init_single_tool may return a different ToolSpec instance than the one passed in (the init() function can return a replacement spec). The code then does `_get_loaded_tools().extend(initialized.values())` and `return initialized[tool_name]`. If a companion's init() returns a spec with a different name, or if the requested tool's init() returns a spec with a different name, the dict key won't match the actual returned spec's name. More importantly, if the requested tool's init() returns a spec whose name differs from tool_name, `initialized[tool_name]` will raise KeyError. This is a real edge case: ToolSpec.init is documented as returning a ToolSpec, and existing tools like python may replace the spec. The observable consequence is a crash (KeyError) when loading a tool whose init() returns a renamed spec, or a companion whose init() returns a renamed spec causing the wrong name to be published. _(2/3 passes — pass 3 looked and disagreed)_

gptme/tools/base.py
```diff
@@ -366,6 +366,100 @@ def callable_signature(func: Callable) -> str:
...
+        else:  # endif
+            if state is None:
+                raise ValueError("{% endif %} without {% if tools %}")
+            state = None
+        pos = end
+    if state is not None:
+        raise ValueError("unterminated {% if tools %} block")
+    trailing = text[pos:]
+    if "{%" in trailing:
+        raise ValueError("invalid {% if tools %} conditional marker")
```
TimeToBuildBob: ❌ **P1** — `render_tool_conditionals` raises `ValueError` whenever any `{%` remains in the text after the recognized conditional markers are processed. Because the function only enters this path when the input contains `{%`, any tool instruction, example, or format string that legitimately contains `{%` for an unrelated purpose — e.g. a Jinja/template example, a shell snippet, or documentation of a templating language — will now cause prompt construction to abort with an uncaught exception, breaking the session. Previously such text passed through untouched. For example, `render_tool_conditionals("Use Jinja: {% if x %}...{% endif %}", set())` raises. The function should ignore unmatched `{%` sequences (or only treat the precise `{% if tools:`, `{% elif tools:`, `{% else`, and `{% endif` forms as markers) rather than raising on any leftover `{%`. _(2/3 passes — pass 2 looked and disagreed)_

gptme/tools/__init__.py
TimeToBuildBob: Fixed in commit b4ba952a8. `load_tool()` now tracks the initialized return value by the original requested spec's identity instead of indexing by its pre-init name. This preserves the documented `init() -> ToolSpec` replacement contract, including a replacement with a different name; the regression test verifies both the returned object and published active tool.

gptme/tools/base.py
TimeToBuildBob: Fixed in commit b4ba952a8. Conditional rendering now opts in only when the text contains this feature's precise `{% if tool(s): ... %}` opener. Unrelated Jinja/template syntax is returned unchanged, including a normal `{% if x %}...{% endif %}` block and arbitrary `{% ... %}` text; both cases have regression coverage.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Convergence adjudication — merge recommendation

**Fixed:**
- `load_tool()` no longer indexes initialized specs by a pre-init name; commit `b4ba952a8` preserves replacement `ToolSpec` returns, including renamed specs.
- Conditional docs now opt in only on the precise `if tool(s):` opener; unrelated Jinja/template `{% ... %}` syntax passes through unchanged.
- Earlier rounds fixed the allowlist boundary, all loading paths, screenshot gating, unavailable companions, and publication atomicity.

**Remaining:** none. All Bob and Greptile inline threads are resolved with checkable commit references.

**Verified against:** `b4ba952a894db463a880f6eb73e2c96b5d2a9b71`.

**CI:** 18 checks passed; deploy skipped as expected. Local verification: 307 focused tests passed, targeted ruff/format checks passed, and mypy passed for the three changed files.

**Domain risk:** plugin/custom-tool initialization is the fragile surface. A maintainer may manually smoke-test `/tools load` with an `init()` that returns a replacement spec, but regression coverage exercises that exact contract.

**Convergence:** detector snapshot before adjudication reported `stable_rounds: 0` of 2 (`status: new_blocking`), reflecting stale unresolved Greptile threads. The current-head Greptile pass now reports **5/5** and no actionable defects. Stop here for maintainer judgment; no auto-merge.

gptme/tools/__init__.py
```diff
@@ -339,6 +350,47 @@ def get_toolchain(
...
+    ``None`` means unrestricted. With an explicit allowlist, every companion
+    must match it; requiring a tool never silently grants an unlisted capability.
+    Runs to a fixpoint so dependency chains resolve.
+    """
+    by_name = {t.name: t for t in available}
+    loaded = {t.name for t in tools} | (already_loaded or set())
+    queue = [t for t in tools if t.requires_tools]
+    while queue:
+        tool = queue.pop()
+        for name in tool.requires_tools:
```
TimeToBuildBob, line 414: ❌ **P1** — In load_tool, the requested tool is identified by identity (`if spec is tool`) after _add_required_tools has appended dependencies. _add_required_tools appends the dependency objects from `available` (the same objects), so identity holds for the requested tool. However, when the requested tool's init() returns a replacement spec (as in test_load_tool_returns_replacement_spec_from_init), `requested` is set to the initialized replacement, which is correct. But if a companion's init() returns a replacement spec, the `loaded` set in _add_required_tools was built from the original spec names, and the initialized list contains the replacement. The `if has_tool(spec.name)` check in the loop uses the original spec name, which is fine. The real issue: when the requested tool is already loaded (has_tool true), the loop skips it and `requested` stays None, then `assert requested is not None` fails. But load_tool raises ValueError earlier if has_tool(tool_name), so that path is unreachable. The more subtle issue: if a companion is already loaded, it is skipped, but the requested tool is not, so requested is set. This seems fine. Let me re-examine: the loop iterates reversed(to_load). If the requested tool is at the end of to_load (it is the first element, so reversed puts it last), and it is not already loaded, requested is set. If a companion init fails, the exception propagates before _get_loaded_tools().extend, so nothing is published. That is correct. The actual bug: _add_required_tools appends dependencies to the `tools` list, but the `loaded` set is initialized with `{t.name for t in tools}` and then updated with `loaded.add(name)` for each dep. However, if a dependency itself has requires_tools, it is appended to the queue, but the queue is processed with `queue.pop()` (LIFO). The fixpoint works. But there is a case where a dependency is already in `loaded` because it was in the original tools list, but the dependency's own requires_tools are not processed because the queue only starts with tools that have requires_tools. If a tool in the original list has requires_tools, it is queued. If a dependency is already loaded and has requires_tools, it is not queued (because the `if name in loaded: continue` skips it, and the dep is not appended to queue). That means if a companion is already loaded but its own required tools are not loaded, they won't be pulled in. Example: tool A requires B, B requires C. B is already loaded (manually), A is being loaded. _add_required_tools([A], available, already_loaded={B}). queue=[A]. Process A: name B in loaded -> continue. B's requires_tools C is never processed. So C is not loaded, even though A's closure includes B and B requires C. This is a real bug: the fixpoint does not traverse already-loaded tools' requirements. The consequence: a tool loaded with a companion that itself has unmet requirements will be missing a transitive dependency, and the model may call a tool that is not loaded. This is a correctness bug in the companion resolution.

gptme/tools/base.py
```diff
@@ -366,6 +366,101 @@ def callable_signature(func: Callable) -> str:
...
+def _loaded_tool_names() -> set[str] | None:
+    """Names of the currently loaded tools, or None before init_tools ran."""
+    # noreorder
+    from . import get_tools  # fmt: skip
+
+    try:
+        tools = get_tools()
+    except Exception:  # pragma: no cover - defensive; never break prompt rendering
+        return None
+    return {t.name for t in tools} if tools else None
```
TimeToBuildBob: ❌ **P1** — In init_tools, the file-tools path builds `permitted = [*(tool.name for tool in file_tools), *tool_names]` and passes it as the allowlist to _add_required_tools. This permits any tool exported by the named file, plus explicitly named built-ins. However, the `available` list passed is `[*file_tools, *get_available_tools(include_mcp=include_mcp)]`. If a file tool requires a built-in that is not in `tool_names`, _add_required_tools raises ValueError. That is the intended behavior per the test test_file_tool_cannot_require_unlisted_builtin. But there is a subtle issue: the `permitted` list includes the file tool names, but `tool_matches_allowlist` is used to check the dep. For a built-in dep like 'read', the allowlist contains the file tool names and tool_names. If tool_names is empty, the allowlist is just the file tool names. So 'read' is not permitted, and the ValueError is raised. That is correct. But what about a file tool that requires another file tool from a different file? The `permitted` list includes all file tool names, so that works. The issue: the `available` list includes `get_available_tools(include_mcp=include_mcp)`, which may include MCP tools. If a file tool requires an MCP tool, and the MCP tool is in the available list, it could be loaded even if it is not in the allowlist? No, the allowlist check would fail because the MCP tool name is not in `permitted`. So it raises. That is fine. I don't see a clear bug here. Let me look at the render_tool_conditionals whole-line logic more carefully for a case where the marker is not whole-line but the previous chunk is empty. If the text starts with a marker that is not whole-line (e.g., "{% if tools: read %}x"), then chunk is "" and out gets "" appended. That is fine. If the marker is whole-line and state[1] is False, the chunk is not appended, but the marker's end is extended. The next chunk starts after the newline. The result is correct. I'll report the transitive dependency bug as the main finding, and the contract change as a secondary. Let me also check the `_loaded_tool_names` function: it returns None if get_tools() raises, but also returns None if tools is empty (the list is falsy). In render_tool_conditionals, loaded=None means "assume every tool is loaded". So before init_tools runs, or when no tools are loaded, the conditionals render as if all tools are loaded. This means that in a session with no tools loaded (e.g., `--tools ""`), the vision instructions would still say "fetch with `read`" even though read is not loaded. That is a bug: the model is told to use a tool that is not available. The test test_vision_docs_do_not_name_read_when_read_is_off uses init_tools with a non-empty allowlist, so it doesn't catch this. The consequence: in a pure-conversation session (no tools), the prompt tells the model to use read/browser/screenshot, which are not loaded, leading to failed tool calls. This is a real correctness bug.

gptme/tools/__init__.py
```diff
@@ -651,11 +705,32 @@ def load_tool(tool_name: str) -> ToolSpec:
...
 
-        # Add to loaded tools
-        _get_loaded_tools().append(tool)
-        logger.info("Loaded tool '%s' mid-conversation", tool_name)
+        _get_loaded_tools().extend(initialized)
+        for spec in initialized:
+            logger.info("Loaded tool '%s' mid-conversation", spec.name)
 
-        return tool
+        assert requested is not None
```
TimeToBuildBob, line 787: ⚠️ **P2** — In load_tool, the requested tool is identified by identity (`if spec is tool`) after _add_required_tools has appended dependencies. _add_required_tools appends the same ToolSpec objects from `available` into `to_load`, so identity holds for the requested tool. However, if the requested tool's `init()` returns a replacement spec (as in test_load_tool_returns_replacement_spec_from_init), the `requested` variable is set to the initialized replacement, and the assert passes. But if the requested tool is already loaded (has_tool returns True), the loop skips it and `requested` remains None, causing the assert to fail. This can happen if a companion tool has the same name as the requested tool, or if the requested tool was loaded as a companion of another tool earlier in the same load_tool call. In that case, load_tool raises AssertionError instead of returning the already-loaded tool. The test test_load_tool_accepts_required_companion_already_loaded covers the case where the companion is already loaded, but not where the requested tool itself is already loaded. This is a real edge case: if a user calls /tools load on a tool that is already loaded as a companion, the command crashes with an AssertionError instead of reporting it as already loaded.

gptme/tools/__init__.py
```diff
@@ -210,16 +210,22 @@ def init_tools(
...
 
             for file_path in file_paths:
                 path = Path(file_path).expanduser()
-                for tool in load_from_file(path):
-                    if not has_tool(tool.name):
-                        tool = _init_single_tool(tool)
-                        loaded_tools.append(tool)
+                file_tools.extend(load_from_file(path))
+            available = [*file_tools, *get_available_tools(include_mcp=include_mcp)]
+            permitted = [*(tool.name for tool in file_tools), *tool_names]
```
TimeToBuildBob, line 226: ⚠️ **P2** — In init_tools, the file-tools path builds `permitted = [*(tool.name for tool in file_tools), *tool_names]` and passes it as the allowlist to _add_required_tools. This means a file tool can require any built-in tool that is explicitly named in tool_names, but also any tool that is exported by the same file. However, the `permitted` list does not include tools that are loaded by default when no allowlist is given. If a file tool requires a built-in tool that is not in tool_names, _add_required_tools raises ValueError even if that built-in tool would be loaded by default in a normal session. This is a behavior change: previously, file tools could reference any built-in tool that was available, and the built-in would be loaded if it was in the default set. Now, a file tool that requires a default-on built-in (e.g. 'shell') will fail to initialize unless 'shell' is explicitly in the allowlist. The test test_file_tool_can_require_explicitly_allowed_builtin covers the case where the built-in is explicitly allowed, but not the case where it is a default-on tool not in the allowlist. This could break existing file tools that require default tools.

gptme/tools/__init__.py
TimeToBuildBob: Fixed in 54b6352ab after rebasing onto master: companion resolution now traverses the `requires_tools` graph of an already-loaded companion and loads any missing transitive dependencies. `test_load_tool_traverses_requirements_of_already_loaded_companion` covers A → already-loaded B → C.

gptme/tools/base.py
TimeToBuildBob: Fixed in 54b6352ab after rebasing onto master: initialized-empty sessions are now distinct from the pre-initialization documentation context. Conditional instructions render against an empty set after `init_tools(allowlist=[])`, so they no longer advertise unavailable tools; a regression test covers the pure-conversation case.

gptme/tools/__init__.py
TimeToBuildBob: Rejected as unreachable. `load_tool()` checks `has_tool(tool_name)` before discovery or closure construction and raises the documented `ValueError("Tool '<name>' is already loaded")`. A tool cannot become its own companion during this call: `to_load` starts with the requested spec, its name is already in the resolver's `loaded` set, and duplicate names are never appended. The requested spec therefore cannot hit the loop's `has_tool(spec.name)` skip unless it was loaded before the call, which the entry guard already rejects.

gptme/tools/__init__.py
TimeToBuildBob: Rejected: this is the intended capability-boundary contract, not a backward-compatibility regression. Before this PR, file tools had no `requires_tools` metadata and loading a file never implicitly enabled any built-in. After this PR, the explicitly supplied startup allowlist remains a hard cap: a file spec may use another spec exported by the named file, while a built-in dependency must also be explicitly named. Treating default-on built-ins as implicitly permitted would widen an explicit file-only allowlist and recreate the security defect fixed earlier in this PR. `test_file_tool_cannot_require_unlisted_builtin` and the documentation state this contract.

TimeToBuildBob:
@greptileai review

gptme/tools/base.py
```diff
@@ -366,6 +366,102 @@ def callable_signature(func: Callable) -> str:
...
+
+def _loaded_tool_names() -> set[str] | None:
+    """Names of the currently loaded tools, or None before init_tools ran."""
+    # noreorder
+    from . import get_tools, tools_initialized  # fmt: skip
+
+    try:
+        if not tools_initialized():
+            return None
+        return {tool.name for tool in get_tools()}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b8df37213: `set_tools()` now marks the destination context initialized, so ACP and restricted-subagent tool restoration renders conditional docs against the restored active names instead of the documentation-only “all tools” mode. Added `test_instructions_render_against_restored_tools` and verified the subagent slice (208 tests).

TimeToBuildBob:
Rebased onto current `master` and force-pushed `b8df37213`; the PR is mergeable again.

Addressed the fresh review findings:
- transitive requirements now traverse through already-loaded companions;
- an initialized empty toolset no longer renders docs as if every tool were loaded;
- non-strict server loading drops an unsatisfied tool without retaining companions appended during the failed closure attempt;
- `set_tools()` marks ACP/subagent restoration contexts initialized.

All four Bob review threads are disposed, and the new Greptile restoration finding has a fix-SHA reply. Local verification: 312 focused tool/prompt tests plus 208 tool/subagent tests passed; ruff and changed-file mypy passed. CI is running on `b8df37213`.

TimeToBuildBob:
@greptileai review

gptme/tools/__init__.py
```diff
@@ -324,6 +334,20 @@ def get_toolchain(
...
+            get_available_tools(include_mcp=include_mcp),
+            allowlist=allowlist,
+        )
+    except ValueError as error:
+        if strict:
+            raise
+        logger.warning(
+            "%s Skipping tools with unsatisfied companion requirements.", error
+        )
+        tools = _remove_tools_with_unsatisfied_requirements(selected_tools)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in f57f6a48f: non-strict resolution now evaluates each selected tool with an independent dependency closure. A broken tool and its partial companions are skipped without discarding unrelated tools whose disabled-by-default companions are available. The regression test covers one independent tool, one valid companion-backed tool, and one broken closure; all 82 focused tool/command tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Erik's Greptile criticism was valid and is fixed on the current head (`f57f6a48f`), including the later companion-closure edge cases. Current state: 18 checks green, Greptile 5/5, our 3-pass review 5/5 with no findings, and all review threads resolved.

I did **not** merge: the independent self-merge gate is still holding this head at 2/5. Its stated concern is stale/false — `load_tool()` has an entry guard that deliberately raises `ValueError` for an already-loaded tool, and `test_load_tool_already_loaded` covers that contract — but the gate's same-head hold-down is fail-closed. This now needs maintainer merge or a new head before automation may proceed.

TimeToBuildBob:
Merged — thanks. Erik's Greptile call resulted in fixes across the full companion-loading path, and the final head landed with all 18 checks green plus both reviewers at 5/5. No remaining action on this PR.

PR #3802, opened 2026-09-10
refactor(shell): background whole tool calls and fix exit stalls

TimeToBuildBob:
## Summary

Replaces the shell tool's raw-string background overlay with the interface Erik requested:

```json
{"command": "npm run dev", "background": true}
```

The flag backgrounds the **whole tool call** as a harness-owned process. It returns a conversation-local job ID immediately, keeps bounded stdout/stderr for inspection, and queues a completion message with exit status and output. Standalone `jobs`, `output <id> [--new]`, `wait <id> [timeout]`, and `kill <id>` remain controls for that job ID; those words inside ordinary scripts are Bash again, and `bg ...` is no longer parsed by gptme.

The independently useful original fix remains as the first commit: if `exit`, `exec`, or another command closes a persistent shell protocol pipe, the tool returns promptly, drains both pipes, restarts Bash, and reports the state reset instead of spinning until the 20-minute timeout.

## Behavioral changes

- `background: true` runs the submitted script unchanged in a separate process group.
- Jobs and IDs are isolated per conversation instead of process-global.
- Completed jobs remain inspectable until conversation teardown; output no longer disappears on `jobs`.
- A loop-continuation hook reports completion automatically.
- Conversation teardown terminates that conversation's process groups, including descendants.
- The old `bg <command>` line scanner is deleted, eliminating heredoc/multiline collisions and the two-job-table public interface.

## Compatibility

Markdown/XML shell blocks have no structured `background` field and therefore remain foreground. Native structured tool calls gain one optional boolean parameter. Existing standalone job-control commands continue to work for harness job IDs; Bash commands that are not an exact control-call shape pass through normally.

## Verification

- `uv run pytest -q tests/test_shell_background.py tests/test_tools_shell.py` — 194 passed
- `uv run prek run --files gptme/tools/shell.py gptme/tools/shell_background.py tests/test_tools_shell.py tests/test_shell_background.py` — all hooks passed (Ruff, formatting, mypy)
- `git diff --check` — clean

Related: #576, #992, #993, #3667, #3673.

Issue #798, opened 2025-10-29
Fails to use `time` in shell tool


ErikBjare:
Getting:

```
Error during execution: Shell error: type = {time command}, token = {time}
```

TimeToBuildBob:
✅ Fixed in PR #799

## Summary

Added error handling to `split_commands()` to gracefully fall back when bashlex can't parse bash reserved words like `time`.

## Solution

When `bashlex.parse()` fails (like with `time` command), the function now:
- Logs a warning about the parsing failure
- Falls back to treating the script as a single command

This follows the same pattern used for shlex parsing elsewhere in the file.

## Testing

The fix allows:
- `time ls`
- `time python script.py`
- And other commands with bash builtins/reserved words

While preserving existing multi-command splitting for normal commands.

PR #3803, opened 2026-09-10
fix(shell): parse `time` scripts, stop false syntax errors, unwrap transparent wrappers

TimeToBuildBob:
## Problem

bashlex 0.18 (unmaintained, last upstream commit 2024-04) cannot parse the `time` keyword and stubs out `select`, `coproc`, `[[ ]]` and `$(( ))` with `NotImplementedError`. The shell tool handled that in three bad ways:

1. **Fallback keyed on exception text.** Any exception whose message contained `"token ="` collapsed the whole script into one command (losing per-command stdin redirect and stop-on-failure) and logged a warning per call. `_redirect_background_stdin` also gave up, so `time sleep 1 &` kept the persistent shell's stdin.
2. **False syntax errors to the model.** bashlex raises a `ParsingError` (not `NotImplementedError`) on valid bash it doesn't model, and `$(( ))` raises `NotImplementedError('arithmetic expansion')` without `"token ="`. All of these reached the model as `Shell syntax error`:
   - `[[ -f x ]] && ls` → `unexpected token '-f' (position 3)`
   - `ls | time wc -l` → `unexpected token 'time'`
   - `echo $((6*7))` → `arithmetic expansion` (seen in real sessions)
   - `diff <(ls) <(ls)`
3. **Allowlist misses on every wrapper.** `is_allowlisted` keys on the first word of each pipeline segment, so `timeout 60 grep …`, `time ls`, `nice -n 10 ls`, `nohup ls` all fell through to a confirmation prompt.

Checked and confirmed: **the denylist never depended on bashlex** (regex over raw text), so `time rm -rf /`, `timeout 5 git push -f` etc. were always denied. This PR adds tests locking that in.

## Fix

**`gptme/tools/shell.py`**
- `_mask_time_keyword`: `\btime\b` → `TIME` before `bashlex.parse`. Same length, so node positions index the real script; non-keyword occurrences (`$time`, `time=1`, `echo time`, heredoc delimiter + terminator) become the same construct spelled differently, so tree shape is unchanged. Used in `split_commands` and `_redirect_background_stdin`. Scripts with `time` now split normally.
- Exception handling by type: bashlex `NotImplementedError` → run whole (`logger.debug`). Any other parse error → `bash -n` arbitrates (parse only, never executes): accepted → run whole; rejected → `Shell syntax error:` with **bash's own message**. No bash binary → previous behaviour (raise with bashlex's message).

**`gptme/tools/shell_validation.py`**
- `strip_transparent_wrappers()`: per segment, strips `time [-p]`, `timeout [-s SIG|-k DUR|--foreground|--preserve-status|-v] DURATION`, `nohup`, `nice [-n N|-N]`, `stdbuf …`, `env [-i] [-u X]`, `command [-p]`, `builtin`. Only these option shapes; anything else leaves the prefix opaque. A wrapper with nothing after it (`env`, `time`) is left alone.
- `is_allowlisted` runs the name check and `flags_permitted` on the stripped text; redirection, sensitive-path and command-substitution checks still run on the original, so `timeout 5 cat ~/.ssh/id_rsa`, `nice -n 10 find . -delete`, `timeout 5 ls > out` stay unapproved.
- Deliberately **not** wrappers: `sudo`, `exec` (replaces the persistent shell), `xargs`, `watch`, `command -v`, and every `NAME=value` form, bare or via `env` — variables change what the command *is* (`PATH=. ls` runs `./ls`) or does (`RIPGREP_CONFIG_PATH` can inject `--pre`, `LD_PRELOAD`), and the set of such names is open-ended (review finding, 091d840be).
- Also fixed (pre-existing, exposed by `time` masking): `split_commands` dropped a heredoc body when another redirect followed the `<<` operator (`cat <<EOF > out.txt`), leaving the shell waiting for the terminator. Command nodes are now sliced through the heredoc like the compound branch already was.

## Before / after (through `execute_shell` with confirmation patched)

| Command | before | after |
|---|---|---|
| `[[ -f /etc/hostname ]] && echo yes` | `Error during execution: Shell error: Shell syntax error: unexpected token '-f'` | `yes` |
| `ls /x \| time wc -l` | false syntax error | runs |
| `echo a\ntime echo b\necho c` | one command, warning logged | three commands, no warning |
| `time ls \| head -2` | confirmation prompt | auto-approved as allowlisted |
| `timeout 2 echo hi` | prompt | auto-approved |
| `ls \|` | `Shell syntax error: unexpected EOF` | `Shell syntax error: line 2: syntax error: unexpected end of file` |
| `time rm -rf /` | denied | denied |

## Tests

New: `test_split_commands_time_keyword_variants`, `…_heredoc_followed_by_redirect_keeps_body`, `…_bashlex_gaps_run_as_single_command`, `…_bashlex_gaps_execute`, `…_syntax_error_uses_bash_message`, `…_without_bash_keeps_bashlex_error`, `test_redirect_background_stdin_with_time_keyword`, and `TestTransparentWrappers` (strip table / allowlisted / not allowlisted / denylist-through-wrappers). `test_split_commands_mixed_reserved_and_normal` inverted: it asserted the fallback, now asserts correct splitting.

834 shell-related tests pass locally; ruff + mypy clean. Disjoint from #3802 (bg branch of `execute_shell`); this touches `_redirect_background_stdin`, `split_commands` and `is_allowlisted` only.

Refs #798 (original `time` report).


gptme/tools/shell_validation.py
```diff
@@ -149,6 +149,51 @@
...
+          -s[ \t]+(?:{_SIGNAL}) | --signal(?:=|[ \t]+)(?:{_SIGNAL})
+        | -k[ \t]+{_DURATION} | --kill-after(?:=|[ \t]+){_DURATION}
+        | --foreground | --preserve-status | -v | --verbose
+      ))*
+      [ \t]+{_DURATION}
+  | nohup
+  | nice(?:[ \t]+(?:-n[ \t]*-?\d+|--adjustment=-?\d+|-\d+))?
+  | stdbuf(?:[ \t]+(?:-[ioe][ \t]*{_BUFFER_MODE}|--(?:input|output|error)={_BUFFER_MODE}))+
+  | env(?:[ \t]+(?:-i|--ignore-environment|-u[ \t]+\w+|--unset=\w+))*
+  | \w+={_PLAIN_VALUE}
```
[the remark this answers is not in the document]
TimeToBuildBob: Agreed, and it generalises beyond PATH (RIPGREP_CONFIG_PATH can inject `--pre`, LD_PRELOAD, ...); the set of variable names that change what an allowlisted binary is or does is open-ended, so a permitted-name list would be the flag-denylist mistake again. Fixed in 091d840be: `NAME=value` prefixes are never transparent, bare or via `env` (`env` stays transparent only with `-i`/`-u`). Tests: `PATH=. ls`, `env PATH=. ls`, `time PATH=. ls`, `RIPGREP_CONFIG_PATH=x rg foo`, `LD_PRELOAD=x cat f` are all not allowlisted.

gptme/tools/shell.py
```diff
@@ -2430,34 +2431,87 @@ def _find_max_heredoc_pos(node, current_max: int = 0) -> int:
...
+
 def split_commands(script: str) -> list[str]:
     import bashlex
 
     # Preprocess script to handle quoted heredoc delimiters that bashlex can't parse
     processed_script = _preprocess_quoted_heredocs(script)
 
     try:
-        parts = bashlex.parse(processed_script)
+        parts = bashlex.parse(_mask_time_keyword(processed_script))
```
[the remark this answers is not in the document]
TimeToBuildBob: The exact example works (`time cat <<'EOF'\nbody\nEOF` → one command including the body: the heredoc redirect node's span covers the body when it is the last part), but the concern found a real, pre-existing bug on master: with a redirect *after* the heredoc operator (`cat <<EOF > out.txt\nbody\nEOF`) the command's span ends before the body and it was dropped — `time` masking would have newly exposed that for `time cat <<EOF > out` shapes that previously took the run-whole fallback. Fixed in 091d840be by slicing command nodes through `_find_max_heredoc_pos` like the compound branch already did; regression test `test_split_commands_heredoc_followed_by_redirect_keeps_body`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

This PR fixes bashlex parsing of the `time` keyword by masking it to `TIME` before parsing, and changes exception handling so bashlex `NotImplementedError` and other parse errors are arbitrated by `bash -n` instead of being reported as syntax errors. It also adds `strip_transparent_wrappers()` to the shell validation allowlist so wrappers like `time`, `timeout`, `nohup`, `nice`, `stdbuf`, `env`, `command`, and `builtin` are stripped before command-name and flag checks. Tests are added for the new parsing behavior, wrapper stripping, and allowlist/denylist interactions.

Confidence 5/5

> ⚠️ **3 of 4 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

- `gptme/tools/shell.py` — Adds `_mask_time_keyword` and `_bash_syntax_error`, uses them in `split_commands` and `_redirect_background_stdin`, and extends heredoc slicing to include the body when a redirect follows the heredoc operator.
- `gptme/tools/shell_validation.py` — Adds `strip_transparent_wrappers` and the `_WRAPPER_PREFIX_RE` regex, and applies it in `is_allowlisted` before the command-name and flag checks.
- `tests/test_tools_shell.py` — Adds tests for `time` keyword variants, bashlex gaps running as single commands, bash syntax error messages, heredoc-with-redirect body preservation, and background stdin with `time`.
- `tests/test_tools_shell_validation.py` — Adds `TestTransparentWrappers` covering wrapper stripping, allowlist acceptance, and denylist visibility through wrappers.

</details>

| `be85ded2be53` | 4/5 | 1 | llm | 2026-09-10 09:46 UTC |
| `091d840bef6e` | 4/5 | 1 | llm | 2026-09-10 10:23 UTC |

</details>

<sub>Reviewed `091d840bef6e` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 40s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell_validation.py
```diff
@@ -149,6 +149,51 @@
...
+      (?:[ \t]+(?:
+          -s[ \t]+(?:{_SIGNAL}) | --signal(?:=|[ \t]+)(?:{_SIGNAL})
+        | -k[ \t]+{_DURATION} | --kill-after(?:=|[ \t]+){_DURATION}
+        | --foreground | --preserve-status | -v | --verbose
+      ))*
+      [ \t]+{_DURATION}
+  | nohup
+  | nice(?:[ \t]+(?:-n[ \t]*-?\d+|--adjustment=-?\d+|-\d+))?
+  | stdbuf(?:[ \t]+(?:-[ioe][ \t]*{_BUFFER_MODE}|--(?:input|output|error)={_BUFFER_MODE}))+
+  | env(?:[ \t]+(?:-i|--ignore-environment|-u[ \t]+\w+|--unset=\w+))*
```
TimeToBuildBob, line 177: ⚠️ **P2** — The `env` wrapper in `_WRAPPER` only recognises `--unset=VALUE` (equals form), not `--unset VALUE` (space form). For a legitimate invocation like `env --unset HOME ls`, the regex matches the bare `env` token, the outer `(?:(?:{_WRAPPER})[ \t]+)+` then strips it, leaving `--unset HOME ls` as the command. `cmd_regex` will treat the leading `--unset` as a command name, which is not allowlisted, so the command falls through to confirmation. The user's intended command never runs; instead the shell receives a malformed command and errors. This breaks a valid `env` usage and produces a confusing failure rather than either fully stripping the wrapper (safe because `ls` is allowlisted) or leaving `env --unset HOME ls` completely opaque.

TimeToBuildBob: Rejected: GNU `env` documents `--unset=NAME`; unlike short `-u NAME`, `--unset NAME` is not a valid invocation. Running it confirms that `env` interprets `--unset` as an ambiguously abbreviated long option and exits before executing `ls`. The review's verification therefore starts from an invalid command. Keeping the unsupported form opaque is the conservative behavior; accepting the suggestion would widen auto-confirmation for syntax that GNU `env` itself rejects.

gptme/tools/shell_validation.py
```diff
@@ -149,6 +149,51 @@
...
+
+
+def strip_transparent_wrappers(cmd: str) -> str:
+    """Remove transparent wrapper prefixes from every pipeline segment.
+
+    ``time ls | timeout 5 grep x`` becomes ``ls | grep x``. A wrapper with
+    nothing after it (``env``, ``time``) is left in place: it is a command in
+    its own right and must be judged as one.
+    """
+    return _WRAPPER_PREFIX_RE.sub(lambda m: m.group("lead"), cmd)
```
TimeToBuildBob, line 195: ⚠️ **P2** — The `strip_transparent_wrappers` regex uses `re.VERBOSE` and `re.MULTILINE`. The pattern `_WRAPPER_PREFIX_RE` matches a wrapper prefix at the start of a line or after a separator. The `lead` group captures the separator and whitespace. The substitution replaces the match with just the lead, effectively removing the wrapper. However, the regex is applied to the entire command string, and it can match inside quoted strings. For example, `echo "time ls"` would have the `time` inside quotes matched? The pattern requires the wrapper to be followed by whitespace and then a non-space character. In `echo "time ls"`, the `time` is inside double quotes, but the regex doesn't know about quotes. It would match `time ls` inside the quotes and strip it, turning `echo "time ls"` into `echo "ls"`. This is a bug because the allowlist would then see `echo "ls"` and might approve a command that actually echoes the string `time ls`. However, the allowlist checks `_has_command_substitution` and `_has_sensitive_args` on the original command, but the command-name check uses the stripped text. If the original command is `echo "time ls"`, the stripped text becomes `echo "ls"`, and `echo` is allowlisted, so the command would be auto-approved. But the original command is just an echo of a string, which is safe. So this is not a security issue, but it could cause false positives? Actually, it could cause a command like `echo "timeout 5 rm -rf /"` to be stripped to `echo "rm -rf /"`, and then the allowlist would see `echo` and approve it, but the original command is just an echo, so it's safe. The real danger is if the wrapper is inside a quoted string that is part of a command argument, e.g., `grep "time ls" file` would become `grep "ls" file`, which is still safe. But what about `cat "time /etc/shadow"`? That would become `cat "/etc/shadow"`, and the sensitive path check runs on the original command, so it would still be blocked. So the impact is limited. However, the regex also matches `time` as a standalone word in a quoted string, which could change the meaning of the command for the allowlist. For example, `echo "time"` becomes `echo ""`? Actually the pattern requires a wrapper followed by whitespace and then a non-space. In `echo "time"`, the `time` is followed by a quote, not whitespace, so it won't match. In `echo "time ls"`, the `time` is followed by a space, so it matches. The substitution removes `time `, leaving `echo "ls"`. This is a false positive in the sense that the allowlist sees a different command, but the original command is still safe. The real issue is that the regex can strip wrappers from inside quoted strings, which could cause a command that is not actually wrapped to be treated as wrapped. For example, `echo "timeout 5 ls"` would be stripped to `echo "ls"`, and the allowlist would approve it, but the original command is just an echo, so it's safe. However, consider `grep "timeout 5 ls" file` - the stripped text becomes `grep "ls" file`, which is still safe. The only way this could be a security issue is if the wrapper is inside a command substitution or something, but those are checked separately. So this is a bug in the sense that the stripping is not quote-aware, but the impact is low. The tests do not cover quoted wrappers. I should report this as a bug because it can cause the allowlist to approve commands that it shouldn't? Actually, it can cause the allowlist to approve commands that are safe anyway. The real danger is if the wrapper is inside a quoted string that is part of a command that is not allowlisted, e.g., `python -c "time ls"` - the stripped text becomes `python -c "ls"`, and `python` is not allowlisted, so it's still blocked. So the impact is minimal. But it's still a correctness issue: the stripping should not happen inside quotes. The fix would be to use a quote-aware parser or to apply the stripping only to the command name position. I'll report it as a P2 bug.

TimeToBuildBob: Rejected: the concrete example does not reproduce. `_WRAPPER_PREFIX_RE` only starts at the beginning of the string, after a shell separator, or after a newline; `echo "time ls"` therefore remains unchanged. It can rewrite text after a separator character inside a quoted argument, but that is conservative rather than permissive: the original outer command is already allowlisted when safe, and introducing a separator into the validation copy adds an extra command-name check that can only reject it. I verified representative cases: `echo "x; time python x.py"` becomes `echo "x; python x.py"` and remains unallowlisted because `python` is checked. Redirection, substitution, sensitive-path, and denylist checks still inspect the original text. A quote-aware parser may reduce needless prompts, but it does not fix an unsafe approval here and is outside this focused wrapper patch.

PR #1274, opened 2026-02-18
fix(tests): skip server API tests when flask is not installed

TimeToBuildBob:
## Summary
- `TestAbsToRelWorkspace` tests in `test_uri.py` import from `gptme.server.api` which requires `flask` (an optional `server` extra dependency)
- When flask is not installed, these tests fail with `ModuleNotFoundError` instead of being skipped
- Added `pytest.mark.skipif` to skip the class when flask is unavailable

## Test plan
- [x] Verified: 20 tests pass, 4 skip when flask is not installed
- [x] Tests would still run when flask IS installed (server extra)

----

> [!IMPORTANT]
> Add `pytest.mark.skipif` to skip `TestAbsToRelWorkspace` tests in `test_uri.py` when `flask` is not installed.
> 
>   - **Tests**:
>     - Add `pytest.mark.skipif` to `TestAbsToRelWorkspace` in `test_uri.py` to skip tests if `flask` is not installed.
>     - Checks for `flask` availability using `importlib.util.find_spec`.
>   - **Verification**:
>     - Confirmed 20 tests pass and 4 are skipped when `flask` is not installed.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=gptme%2Fgptme&utm_source=github&utm_medium=referral)<sup> for 0a690c1eef1db084064d832f1b0226fa21357608. You can [customize](https://app.ellipsis.dev/gptme/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #3804, opened 2026-09-10
fix(shell): bound captured subprocess output bytes (GPTME_SHELL_MAX_OUTPUT_BYTES)

### LINKED ISSUE #3798: fix(shell): bound captured subprocess output bytes (unbounded Queue before truncation)
## Problem

`gptme/tools/shell.py` captures the full subprocess output into an unbounded `Queue` (reader threads `os.read(fd, 2**16)` → `data_q.put`) and only applies `GPTME_SHELL_TRUNC_PRE_TOKENS`/`POST_TOKENS` truncation after the process exits. The context is protected; the gptme process RSS is not.

`SHELL_MEMORY_LIMIT` (`_get_memory_limit`, `ulimit -v` on the bash child) does not cover this: `cat` of a huge file uses no memory itself, the bytes pile up in gptme.

## Incident

2026-09-10, an autonomous PM session ran `cat` on its own 2.7 GiB (15.4 M line) slot log — the file its stdout appends to, so the read fed itself. gptme reached 3.3 GiB RSS + 1.9 GiB swap inside the systemd `MemoryHigh=3G` unit, drained the box's 4 GiB swap, and parked the whole PM fleet for ~10 min. Same class as ErikBjare/bob#151 (2025-11, VM freeze). Related bounded-capture work: #1274, #1724 (Codex CLI capture) — the shell tool has no equivalent.

## Proposal

- Cap captured bytes (e.g. `GPTME_SHELL_MAX_OUTPUT_BYTES`, default 32 MiB). Past the cap: stop reading into the buffer, terminate the child (SIGTERM then SIGKILL), and return the pre/post window with an explicit `[output truncated at N MiB, process killed]` marker.
- Keep the token-window truncation as the second stage.
- Test: a `yes | head -c 200M`-style command must return within the cap without RSS growth.

TimeToBuildBob:
## Problem

`gptme/tools/shell.py` captured the full subprocess output into an unbounded list (`stdout: list[str]`, `stderr: list[str]`) and only applied token-level truncation after the process exited. A `cat` of a multi-GiB file accumulates all bytes in the gptme process's heap.

**Incident (2026-09-10):** An autonomous PM session ran `cat` on its own 2.7 GiB slot log. gptme reached 3.3 GiB RSS + 1.9 GiB swap inside a `MemoryHigh=3G` systemd unit, drained the box's 4 GiB swap, and parked the whole PM fleet for ~10 minutes. Same class as ErikBjare/bob#151 (2025-11). Related bounded-capture work: #1274, #1724 — the shell tool had no equivalent cap.

## Fix

Add a hard byte cap on the total bytes accumulated in the reader loop, configurable via `GPTME_SHELL_MAX_OUTPUT_BYTES` (env) / `SHELL_MAX_OUTPUT_BYTES` (config.toml). Default: **32 MiB**.

When the cap is exceeded:
1. The subprocess receives SIGTERM then SIGKILL.
2. A `[output truncated at N MiB, process killed]` marker is appended to stdout.
3. Return code `-125` signals the caller (distinct from `-124` = timeout).
4. Header in the tool output: `Command killed (output exceeded N MiB cap)`.
5. Token-level truncation (`GPTME_SHELL_TRUNC_*`) applies on top as a second stage.

Both `_read_output_unix` and `_read_output_windows` are covered.

## Changes

- `gptme/tools/shell.py`: `_DEFAULT_MAX_OUTPUT_BYTES`, `_get_max_output_bytes()`, byte tracking in both read loops, `byte_cap_exceeded` flag through `execute_shell_impl` → `_format_shell_output`
- `tests/test_shell_max_output_bytes.py`: 9 tests — env reader unit tests, live cap behaviour (`yes` as infinite source), size bound proof, normal output unaffected, partial output preserved

## Testing

```
tests/test_shell_max_output_bytes.py  9 passed
tests/test_tools_shell.py            137 passed (0 regressions)
tests/test_shell_memory_limit.py       5 passed
```

Closes #3798

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -969,12 +969,17 @@ class ShellSession:
                     with cap_lock:
                         cap_state["bytes"] += len(raw)
                         over = cap_state["bytes"] > max_output_bytes
+                        if over:
+                            # Flag the cap BEFORE enqueueing the chunk so the
+                            # consumer cannot dequeue an over-cap chunk that
+                            # carries the delimiter while `over` is still
+                            # False and return the real code (Greptile P1 /
+                            # bob-ai-review P1 race).
+                            cap_state["over"] = True
                     queue.put(data)
                     if over:
                         # Stop producing more data; the consumer detects the
                         # cap and performs the kill + marker.
-                        with cap_lock:
-                            cap_state["over"] = True
                         break
                 except BlockingIOError:
                     time.sleep(0.01)
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -21,6 +21,15 @@ Configuration:
     truncation path fire on smaller outputs, which surfaces savings telemetry
     in `context-savings.jsonl` and the `/context` command. Invalid values fall
     back to defaults.
+
+    GPTME_SHELL_MAX_OUTPUT_BYTES: Hard cap on the total bytes (stdout + stderr
+        combined) captured into the in-process buffer before the subprocess is
+        killed and the output is truncated. Accepts a plain byte count or a
+        binary suffix (e.g. "32M", "1G"). Default: 32 MiB. This prevents a
+        runaway ``cat`` of a multi-GiB file from exhausting gptme's RSS.
+        The process receives SIGTERM then SIGKILL; the returned output contains
+        a ``[output truncated at N MiB, process killed]`` marker. Token-level
+        truncation (GPTME_SHELL_TRUNC_*) is applied on top as a second stage.
 """
 
 import atexit
@@ -172,6 +181,10 @@ _TRUNC_STDERR_PRE_TOKENS_DEFAULT = 2000
 _TRUNC_STDERR_POST_TOKENS_DEFAULT = 2000
 _GIT_LOG_PREVIEW_LINES = 20
 _GH_LIST_PREVIEW_LINES = 10
+
+# Default byte cap for captured subprocess output (32 MiB). Configurable via
+# GPTME_SHELL_MAX_OUTPUT_BYTES (or [env] SHELL_MAX_OUTPUT_BYTES in config.toml).
+_DEFAULT_MAX_OUTPUT_BYTES = 32 * 1024 * 1024  # 32 MiB
 
 
 candidates = (
@@ -355,6 +368,37 @@ def _get_memory_limit() -> int | None:
             exc,
         )
         return None
+    return limit
+
+
+def _get_max_output_bytes() -> int:
+    """Read the output byte cap in bytes (default 32 MiB).
+
+    Knob is ``GPTME_SHELL_MAX_OUTPUT_BYTES`` (env) or
+    ``[env] SHELL_MAX_OUTPUT_BYTES`` (config.toml). Accepts a plain byte count
+    or a binary-suffixed size (e.g. ``"32M"``). Unparseable or non-positive
+    values fall back to the 32 MiB default so a misconfiguration never disables
+    the cap entirely.
+    """
+    from ..config import get_config  # deferred: avoid import cycle
+
+    raw = get_config().get_env("SHELL_MAX_OUTPUT_BYTES")
+    if not raw:
+        return _DEFAULT_MAX_OUTPUT_BYTES
+    try:
+        limit = _parse_size(str(raw))
+    except (ValueError, AttributeError):
+        logger.warning(
+            "Ignoring invalid GPTME_SHELL_MAX_OUTPUT_BYTES=%r (expected e.g. '32M')",
+            raw,
+        )
+        return _DEFAULT_MAX_OUTPUT_BYTES
+    if limit <= 0:
+        logger.warning(
+            "GPTME_SHELL_MAX_OUTPUT_BYTES=%r is not positive; using default 32 MiB",
+            raw,
+        )
+        return _DEFAULT_MAX_OUTPUT_BYTES
     return limit
 
 
@@ -856,6 +900,7 @@ class ShellSession:
         stderr: list[str] = []
         return_code: int | None = None
         start_time = time.time() if timeout else None
+        max_output_bytes = _get_max_output_bytes()
 
         if _is_windows:
             return self._read_output_windows(
@@ -868,6 +913,7 @@ class ShellSession:
                 start_marker_pattern,
                 start_time,
                 timeout,
+                max_output_bytes,
             )
         return self._read_output_unix(
             command,
@@ -879,6 +925,7 @@ class ShellSession:
             start_marker_pattern,
             start_time,
             timeout,
+            max_output_bytes,
         )
 
     def _read_output_windows(
@@ -892,6 +939,7 @@ class ShellSession:
         start_marker_pattern: str,
         start_time: float | None,
         timeout: float | None,
+        max_output_bytes: int = _DEFAULT_MAX_OUTPUT_BYTES,
     ) -> tuple[int | None, str, str]:
         """Read command output on Windows using threads with non-blocking I/O."""
         from queue import Empty, Queue
@@ -926,6 +974,7 @@ class ShellSession:
         t_stderr.start()
 
         re_returncode = re.compile(r"ReturnCode:(\d+)")
+        captured_bytes = 0
 
         try:
             while not stop_event.is_set():
@@ -1008,8 +1057,29 @@ class ShellSession:
                             )
 
                         stdout.append(line)
+                        captured_bytes += len(line)
                         if output:
                             print(line, end="", file=sys.stdout)
+                        if captured_bytes > max_output_bytes:
+                            cap_mib = max_output_bytes / (1024 * 1024)
+                            trunc_msg = (
+                                f"\n[output truncated at {cap_mib:.0f} MiB,"
+                                f" process killed]\n"
+                            )
+                            stdout.append(trunc_msg)
+                            if output:
+                                print(trunc_msg, end="", file=sys.stdout)
+                            logger.warning(
+                                "Shell output cap (%d MiB) exceeded; killing process",
+                                int(cap_mib),
+                            )
+                            self._terminate_process()
+                            stop_event.set()
+                            return (
+                                -125,
+                                trim_blank_lines("".join(stdout)),
+                                trim_blank_lines("".join(stderr)),
+                            )
                 except Empty:
                     pass
 
@@ -1019,8 +1089,29 @@ class ShellSession:
                     lines = data.splitlines(keepends=True)
                     for line in lines:
                         stderr.append(line)
+                        captured_bytes += len(line)
                         if output:
                             print(line, end="", file=sys.stderr)
+                        if captured_bytes > max_output_bytes:
+                            cap_mib = max_output_bytes / (1024 * 1024)
+                            trunc_msg = (
+                                f"\n[output truncated at {cap_mib:.0f} MiB,"
+                                f" process killed]\n"
+                            )
+                            stderr.append(trunc_msg)
+                            if output:
+                                print(trunc_msg, end="", file=sys.stderr)
+                            logger.warning(
+                                "Shell output cap (%d MiB) exceeded; killing process",
+                                int(cap_mib),
+                            )
+                            self._terminate_process()
+                            stop_event.set()
+                            return (
+                                -125,
+                                trim_blank_lines("".join(stdout)),
+                                trim_blank_lines("".join(stderr)),
+                            )
                 except Empty:
                     pass
 
@@ -1057,9 +1148,11 @@ class ShellSession:
         start_marker_pattern: str,
         start_time: float | None,
         timeout: float | None,
+        max_output_bytes: int = _DEFAULT_MAX_OUTPUT_BYTES,
     ) -> tuple[int | None, str, str]:
         """Read command output on Unix using select()."""
         assert select is not None
+        captured_bytes = 0
         try:
             while True:
                 # Calculate remaining timeout
@@ -1200,12 +1293,44 @@ class ShellSession:
                             )
                         if fd == self.stdout_fd:
                             stdout.append(line)
+                            captured_bytes += len(line)
                             if output:
                                 print(line, end="", file=sys.stdout)
                         elif fd == self.stderr_fd:
                             stderr.append(line)
+                            captured_bytes += len(line)
                             if output:
                                 print(line, end="", file=sys.stderr)
+
+                        if captured_bytes > max_output_bytes:
+                            cap_mib = max_output_bytes / (1024 * 1024)
+                            trunc_msg = (
+                                f"\n[output truncated at {cap_mib:.0f} MiB,"
+                                f" process killed]\n"
+                            )
+                            stdout.append(trunc_msg)
+                            if output:
+                                print(trunc_msg, end="", file=sys.stdout)
+                            logger.warning(
+                                "Shell output cap (%d MiB) exceeded; killing process",
+                                int(cap_mib),
+                            )
+                            try:
+                                pgid = os.getpgid(self.process.pid)
+                                os.killpg(pgid, signal.SIGTERM)
+                                time.sleep(0.1)
+                                if self.process.poll() is None:
+                                    os.killpg(pgid, signal.SIGKILL)
+                            except Exception as e:
+                                logger.warning(
+                                    "Error killing process after byte cap exceeded: %s",
+                                    e,
+                                )
+                            return (
+                                -125,
+                                trim_blank_lines("".join(stdout)),
+                                trim_blank_lines("".join(stderr)),
+                            )
         except KeyboardInterrupt:
             # Clear line after ^C to avoid leaving a hanging line
             print()
@@ -1625,6 +1750,7 @@ def _format_shell_output(
     timed_out: bool = False,
     timeout_value: float | None = None,
     logdir: Path | None = None,
+    byte_cap_exceeded: bool = False,
 ) -> str:
     """Format shell command output into a message."""
     # Strip ANSI escape sequences from output
@@ -1638,6 +1764,7 @@ def _format_shell_output(
         and not stderr
         and not interrupted
         and not timed_out
+        and not byte_cap_exceeded
         and _matches_git_log_oneline(cmd)
     ):
         try:
@@ -1651,6 +1778,7 @@ def _format_shell_output(
         and not stderr
         and not interrupted
         and not timed_out
+        and not byte_cap_exceeded
         and _matches_gh_list(cmd)
     ):
         try:
@@ -1659,7 +1787,12 @@ def _format_shell_output(
             logger.warning("Failed to format compact shell output: %s", e)
 
     if compact_stdout is None and (
-        returncode == 0 and stdout and not stderr and not interrupted and not timed_out
+        returncode == 0
+        and stdout
+        and not stderr
+        and not interrupted
+        and not timed_out
+        and not byte_cap_exceeded
     ):
         try:
             compact_stdout = _format_query_pruned_output(cmd, stdout, logdir)
@@ -1696,7 +1829,10 @@ def _format_shell_output(
         )
 
     # Format header
-    if timed_out:
+    if byte_cap_exceeded:
+        cap_mib = _DEFAULT_MAX_OUTPUT_BYTES / (1024 * 1024)
+        header = f"Command killed (output exceeded {cap_mib:.0f} MiB cap)"
+    elif timed_out:
         header = (
             f"Command timed out (after {timeout_value}s)"
             if timeout_value
@@ -1728,7 +1864,9 @@ def _format_shell_output(
     if stderr:
         msg += _format_block_smart("", stderr, "stderr").lstrip() + "\n\n"
     if not compact_stdout and not stdout and not stderr:
-        if timed_out:
+        if byte_cap_exceeded:
+            msg += "No output before byte cap\n"
+        elif timed_out:
             msg += "No output before timeout\n"
         elif interrupted:
             msg += "No output before interruption\n"
@@ -1741,6 +1879,8 @@ def _format_shell_output(
             msg += f"Process interrupted (return code: {returncode})\n"
         else:
             msg += "Process interrupted\n"
+    elif byte_cap_exceeded or timed_out:
+        pass  # Header already describes the termination; return code is synthetic
     elif returncode:
         msg += f"Return code: {returncode}\n"
 
@@ -1759,6 +1899,7 @@ def execute_shell_impl(
         returncode, stdout, stderr = shell.run(cmd, timeout=timeout)
         interrupted = False
         timed_out = returncode == -124  # Our timeout return code
+        byte_cap_exceeded = returncode == -125  # Output byte cap return code
     except KeyboardInterrupt as e:
         # Extract partial output and handle subprocess termination
         stdout = stderr = ""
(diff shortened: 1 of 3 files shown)


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_shell_max_output_bytes.py` (+151/-0, Δ151)
- `gptme/tools/shell.py` (+146/-3, Δ149)

### Suggested verification
- `uv run pytest tests/test_shell_max_output_bytes.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3804._

TimeToBuildBob:
## 🤖 AI code review

Adds a configurable byte cap (GPTME_SHELL_MAX_OUTPUT_BYTES, default 32 MiB) to the shell tool that kills the subprocess and returns -125 when combined stdout+stderr bytes exceed the cap. Tracks bytes in both Unix and Windows read loops, adds a bounded post-kill drain helper (_kill_for_byte_cap), and threads a byte_cap_exceeded flag through output formatting. Adds unit and integration tests and a test fixture resetting _verbosity_warned.

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **2 of 3 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>3 findings</b> · ⚠️ 3 P2</summary>

**⚠️ P2 medium** — `gptme/tools/shell.py:970`

The read_stream producer in _read_output_windows adds len(raw) to cap_state["bytes"] for every chunk read, including the shell-injected start marker (START_OF_COMMAND_OUTPUT_&lt;id&gt;) and the return marker (ReturnCode:... END_OF_COMMAND_OUTPUT). The Unix path strips these via captured_raw.find(marker_bytes) and rfind(b"ReturnCode:"), but the Windows producer has no equivalent. A command whose actual output is exactly max_output_bytes therefore accumulates max_output_bytes plus the marker sizes (~60 bytes) and trips the cap, returning -125, while the same command succeeds on Unix. With a small configured cap (e.g. 100 bytes), any command is killed before its real output is processed. This makes the cap platform-dependent and breaks the documented contract.

<sub>**How this was verified:** Checked the read_stream function lines 958-987. It adds len(raw) for every chunk, including the start marker and delimiter. The Unix path explicitly excludes these markers, but the Windows path does not. The test file only runs POSIX tests (skipif nt), so this is untested.</sub>

**⚠️ P2 medium** — `tests/test_shell_max_output_bytes.py:121`

The integration test test_cap_total_output_size_bounded asserts captured &lt; cap + 2 * chunk + 1024 (line 121). The implementation's _kill_for_byte_cap allows draining up to max_output_bytes additional bytes from the pipes after the cap triggers, so the returned stdout can reach 2 * cap + chunk. The test passes on typical systems only because the kernel pipe buffer is small (≤64 KiB), so the drained amount is small. On a system with a larger pipe buffer or a descendant that keeps the pipe alive, the drained data exceeds the bound and the test fails even though the code is behaving as designed. The test is non-hermetic, depending on OS pipe-buffer size.

<sub>**How this was verified:** _kill_for_byte_cap sets drain_budget = max_output_bytes and drains until budget or deadline; the test relies on the actual drained bytes being under the bound.</sub>

**⚠️ P2 medium** — `gptme/tools/shell.py:1310`

In _read_output_unix, the delimiter exclusion uses captured_raw.rfind(b"ReturnCode:") only when the full 11-byte substring is present in the current chunk. If the shell's ReturnCode: marker is split across two 64 KiB reads (e.g. the first chunk ends with ...ReturnCod), rfind returns -1 and the partial marker bytes are not trimmed from captured_raw before captured_bytes is incremented. A command whose real output is max_output_bytes - 9 bytes, with the chunk ending at ...ReturnCode, yields a count of max_output_bytes + 1, tripping the cap and returning -125 even though the subprocess output is under the cap. This contradicts the comment at line 1297 and silently kills commands near the cap.

<sub>**How this was verified:** Checked the code around lines 1299-1316. The delimiter exclusion only runs when the full 'ReturnCode:' and delimiter are in the same chunk. If split, captured_bytes includes the marker bytes. The test at line 141-151 uses a single printf that fits in one chunk, so it doesn't exercise the split.</sub>

</details>

- `gptme/tools/shell.py` — Adds _DEFAULT_MAX_OUTPUT_BYTES, _get_max_output_bytes(), byte tracking in _read_output_unix/_read_output_windows, _kill_for_byte_cap, and byte_cap_exceeded handling in _format_shell_output.
- `tests/test_llm_openai.py` — Adds an autouse fixture to reset llm_openai._verbosity_warned before each test.
- `tests/test_shell_max_output_bytes.py` — Adds unit tests for _get_max_output_bytes and integration tests for the byte cap using yes and a simulated drain.

</details>

| `5d0e2ea38386` | 3/5 | 2 | llm | 2026-09-10 10:03 UTC |
| `27c4513aaa13` | 2/5 | 4 | llm | 2026-09-10 11:34 UTC |
| `ca1f302617a9` | 4/5 | 2 | llm | 2026-09-10 13:31 UTC |

</details>

<sub>Reviewed `bbda8b2f4040` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 581s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell.py
```diff
@@ -1200,12 +1293,44 @@ def _read_output_unix(
...
+                                os.killpg(pgid, signal.SIGTERM)
+                                time.sleep(0.1)
+                                if self.process.poll() is None:
+                                    os.killpg(pgid, signal.SIGKILL)
+                            except Exception as e:
+                                logger.warning(
+                                    "Error killing process after byte cap exceeded: %s",
+                                    e,
+                                )
+                            return (
```
TimeToBuildBob: ❌ **P1** — In `_read_output_unix`, when the byte cap is exceeded, the code kills the process group and returns -125. However, it does not drain the remaining data from the pipes before returning. The process is killed, but the pipe buffers may still contain data that was written before the kill. Since the function returns immediately, the next call to `_run_pipe` will start with a pre-drain of stderr (lines 858-872) but not stdout. The stdout pipe may still have unread data from the killed process, which will be read as part of the next command's output, potentially mixing with the next command's output. The pre-drain only handles stderr, not stdout. This is a real bug: after a byte-cap kill, the next command's output may be prefixed with leftover data from the killed process. The existing timeout path (lines 1162-1181) has the same issue, but it is pre-existing. The new code introduces a new path that also fails to drain stdout. The consequence is that subsequent shell commands may see corrupted output, leading to incorrect tool results. The fix would be to drain both pipes after killing, similar to the delimiter path. This is a P1 bug.

gptme/tools/shell.py
```diff
@@ -1019,8 +1089,29 @@ def read_stream(fd: int, queue: Queue[str]) -> None:
...
+                        captured_bytes += len(line)
                         if output:
                             print(line, end="", file=sys.stderr)
+                        if captured_bytes > max_output_bytes:
+                            cap_mib = max_output_bytes / (1024 * 1024)
+                            trunc_msg = (
+                                f"\n[output truncated at {cap_mib:.0f} MiB,"
+                                f" process killed]\n"
+                            )
+                            stderr.append(trunc_msg)
```
TimeToBuildBob: ⚠️ **P2** — In `_read_output_windows`, the byte cap check is inside the per-line loop, but the `captured_bytes` counter is incremented for both stdout and stderr lines. However, the check for the cap is done separately in the stdout and stderr blocks. If the cap is exceeded by a stderr line, the code appends the truncation marker to stderr and returns -125. But the truncation marker is appended to stderr, not stdout. The PR description says the marker is appended to stdout. In the Unix path, the marker is always appended to stdout (line 1311). In the Windows path, the marker is appended to whichever stream exceeded the cap (lines 1069 and 1101). This inconsistency means that on Windows, if stderr exceeds the cap, the marker appears in stderr, and the caller (execute_shell_impl) may not see it in stdout. The test only runs on POSIX, so this is not caught. The consequence is that on Windows, the truncation marker may be in stderr, and the header logic in _format_shell_output uses byte_cap_exceeded (based on return code -125) so the header is correct, but the marker placement is inconsistent. This is a minor bug, but it violates the documented behavior. Severity P2.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -900,17 +948,34 @@ def _read_output_windows(
...
-                    data = os.read(fd, 2**16).decode("utf-8", errors="replace")
-                    if not data:
+                    raw = os.read(fd, 2**16)
+                    if not raw:
                         break
+                    data = raw.decode("utf-8", errors="replace")
+                    with cap_lock:
+                        cap_state["bytes"] += len(raw)
+                        over = cap_state["bytes"] > max_output_bytes
                     queue.put(data)
```
TimeToBuildBob, line 1034: ❌ **P1** — In _read_output_windows, the producer thread reads a chunk, increments cap_state['bytes'] under the lock, and then calls queue.put(data) before setting cap_state['over'] = True. The consumer can dequeue that chunk in the window before over is set, check cap_state['over'] at line 1016 (still False), and if the chunk contains the delimiter line, return the real return code instead of -125. This race lets a chunk that both exceeds the cap and carries the delimiter bypass the cap entirely, so a command outputting more than the limit returns success with uncapped output. Moving the queue.put after setting over, or having the consumer check the byte count directly, closes the race.

gptme/tools/shell.py
```diff
@@ -1215,6 +1388,65 @@ def _read_output_unix(
...
+            pgid = os.getpgid(self.process.pid)
+            os.killpg(pgid, signal.SIGTERM)
+            time.sleep(0.1)
+            if self.process.poll() is None:
+                os.killpg(pgid, signal.SIGKILL)
+        except Exception as e:
+            logger.warning("Error killing process after byte cap exceeded: %s", e)
+        # Drain any remaining data from both pipes so it does not bleed into the
+        # next command's output (Greptile P1).
+        for drain_fd in (self.stdout_fd, self.stderr_fd):
```
TimeToBuildBob, line 1543: ⚠️ **P2** — In _kill_for_byte_cap, the drain loop reads from both pipes and appends the drained data to stdout/stderr. However, this drain can itself accumulate unbounded data if the process is still producing output after SIGTERM but before SIGKILL. The drain loop uses `_wait_readable` with a 0.1s timeout and reads up to 2**16 bytes per iteration, but it does not have a byte limit. If the process ignores SIGTERM and continues producing output, the drain loop will keep reading until the process is killed by SIGKILL (after 0.1s sleep) or until the pipe is empty. Since SIGKILL is sent after 0.1s, the process is killed quickly, but in that 0.1s window it could produce a large amount of data. More importantly, the drain loop does not check the byte cap, so the total output returned can exceed the cap by a significant amount. For example, if the cap is 32 MiB and the process produces 100 MiB in the 0.1s window, the drain will capture all 100 MiB, defeating the purpose of the cap. The fix should limit the drain to a reasonable amount (e.g., one more chunk) or check the cap during the drain and stop.

gptme/tools/shell.py
TimeToBuildBob: fixed in 132ba8eae: `_kill_for_byte_cap` now always SIGKILLs the whole process group (independent of the shell leader's liveness) and bounds the post-cap drain by both total bytes (one cap's worth) and a 1s time deadline. A surviving descendant can no longer feed an unbounded drain. Also hardened in df9d4c716: the Unix post-delimiter stderr drain now counts bytes against the shared budget and enforces `-125` when it trips.

Re-ran the production AI reviewer on head `df9d4c7163` → **5/5, 0 findings**.

gptme/tools/shell.py
TimeToBuildBob: fixed in 57cf606a8: the Windows producer thread now sets `cap_state["over"] = True` under the same lock, before `queue.put(data)`, so an over-cap chunk carrying the delimiter cannot be dequeued while `over` is still False and return the real code. Also reinforced in df9d4c716: the Windows delimiter branch checks `cap_state["over"]` before returning, enforcing `-125`.

Re-ran the production AI reviewer on head `df9d4c7163` → **5/5, 0 findings**.

gptme/tools/shell.py
TimeToBuildBob: fixed in 132ba8eae: the post-kill drain in `_kill_for_byte_cap` is now bounded by a total-byte budget (one cap's worth) and a 1s time deadline, so a descendant still writing cannot accumulate unbounded data. df9d4c716 additionally bounds the Unix post-delimiter stderr drain against the same budget. `test_cap_drain_is_byte_bounded` now proves the drain consumes data and stays within the budget.

Re-ran the production AI reviewer on head `df9d4c7163` → **5/5, 0 findings**.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -1174,25 +1363,43 @@ def _read_output_unix(
                                 else:
                                     os.chdir(pwd.strip())
 
+                            # If the byte cap was already exceeded in this chunk
+                            # (delimiter line present), do not return the real
+                            # code — enforce the cap (Greptile P1 race).
+                            if captured_bytes > max_output_bytes:
```
TimeToBuildBob, line 1441: ⚠️ **P2** — In _read_output_unix, the byte cap check `if captured_bytes > max_output_bytes` is placed after the delimiter branch, but the delimiter line itself is counted in captured_bytes. If a command produces output that exactly reaches the cap and then the delimiter line (which is part of the captured bytes) pushes captured_bytes over the cap, the code returns -125 and kills the process even though the command completed normally. The delimiter line is injected by the shell wrapper, not command output, so counting it against the cap can cause false positives for commands whose output is just under the cap. The consequence is that a command producing output slightly below the cap (e.g. 32 MiB - 1 byte) plus the delimiter line (about 30 bytes) will be killed and reported as truncated, even though the actual command output was within the cap. This is a correctness bug: the cap should apply to command output, not the wrapper's delimiter.

tests/test_shell_max_output_bytes.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # Force the drain to always see readable data: the loop reads a chunk,
+    # drains the "budget", and stops. Without the bound it would loop forever.
+    monkeypatch.setattr(shell_mod, "_wait_readable", lambda fds, timeout: list(fds))
+    # Never report EOF so only the byte/time bound can stop the loop.
+    real_read = os.read
+    monkeypatch.setattr(
+        shell_mod.os,
+        "read",
+        lambda fd, n: (
```
TimeToBuildBob, line 261: ⚠️ **P2** — The test test_cap_drain_is_byte_bounded monkeypatches shell_mod.os.read to return b'x' * n for the shell's stdout/stderr fds. However, it also patches _wait_readable to always return the fds as readable. The drain loop in _kill_for_byte_cap reads from each fd in a for loop, and for each fd it loops until drain_empty_count >= 2 or drain_budget <= 0 or deadline. With the monkeypatch, os.read always returns data, so drain_empty_count never increments. The loop will read until drain_budget is exhausted. For the first fd (stdout), it reads min(2**16, drain_budget) each time, so it will consume the entire budget in one read if drain_budget >= 2**16. For cap=32*1024, drain_budget=32768, so it reads 32768 bytes, drain_budget becomes 0, and the loop exits. Then for stderr, drain_budget is 0, so the loop doesn't run. The test asserts drained_bytes > 0, which is true (32768 bytes). But the test's assertion `drained_bytes <= cap + 256` is 32768 <= 33024, which passes. The test does not actually verify that the drain is bounded by the budget in a meaningful way, because it only exercises the case where the budget is exhausted in one read. It would pass even if the drain loop had no budget check but a time bound, as long as the time bound didn't trigger. However, the test is not vacuous; it does verify that the drain consumes data and stops. The issue is that the test's monkeypatch of os.read returns b'x' * n, which is a string of length n, but os.read is expected to return bytes. In Python, b'x' * n is bytes, so that's fine. The test is valid.

TimeToBuildBob: Rejected: the finding concludes that the test is valid. It simulates pipes that never reach EOF and proves the shared byte budget terminates the drain; no code change is needed.

gptme/tools/shell.py
TimeToBuildBob: Fixed in bbda8b2f4: the Unix reader now excludes the injected start and return-marker bytes from byte accounting, while still counting command output that shares the delimiter chunk. Added test_delimiter_bytes_do_not_trip_cap at the exact cap boundary; the focused suite passes 12/12.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -900,17 +948,39 @@ def _read_output_windows(
...
+            while not stop_event.is_set() and not cap_state["over"]:
                 try:
-                    data = os.read(fd, 2**16).decode("utf-8", errors="replace")
-                    if not data:
+                    raw = os.read(fd, 2**16)
+                    if not raw:
                         break
+                    data = raw.decode("utf-8", errors="replace")
+                    with cap_lock:
+                        cap_state["bytes"] += len(raw)
```
TimeToBuildBob: ⚠️ **P2** — The read_stream producer in _read_output_windows adds len(raw) to cap_state["bytes"] for every chunk read, including the shell-injected start marker (START_OF_COMMAND_OUTPUT_<id>) and the return marker (ReturnCode:... END_OF_COMMAND_OUTPUT). The Unix path strips these via captured_raw.find(marker_bytes) and rfind(b"ReturnCode:"), but the Windows producer has no equivalent. A command whose actual output is exactly max_output_bytes therefore accumulates max_output_bytes plus the marker sizes (~60 bytes) and trips the cap, returning -125, while the same command succeeds on Unix. With a small configured cap (e.g. 100 bytes), any command is killed before its real output is processed. This makes the cap platform-dependent and breaks the documented contract.

tests/test_shell_max_output_bytes.py
```diff
@@ -0,0 +1,243 @@
...
+    cap = 256 * 1024  # 256 KiB
+    chunk = 2**16  # 64 KiB — the read chunk size
+
+    with patch("gptme.tools.shell._get_max_output_bytes", return_value=cap):
+        returncode, stdout, stderr = shell.run("yes", timeout=10)
+
+    assert returncode == -125, f"Expected -125 (byte cap), got {returncode}"
+    # Output must be capped: at most cap + one extra chunk + truncation marker
+    captured = len(stdout.encode("utf-8", errors="replace"))
+    assert captured < cap + 2 * chunk + 1024, (
```
TimeToBuildBob: ⚠️ **P2** — The integration test test_cap_total_output_size_bounded asserts captured < cap + 2 * chunk + 1024 (line 121). The implementation's _kill_for_byte_cap allows draining up to max_output_bytes additional bytes from the pipes after the cap triggers, so the returned stdout can reach 2 * cap + chunk. The test passes on typical systems only because the kernel pipe buffer is small (≤64 KiB), so the drained amount is small. On a system with a larger pipe buffer or a descendant that keeps the pipe alive, the drained data exceeds the bound and the test fails even though the code is behaving as designed. The test is non-hermetic, depending on OS pipe-buffer size.

gptme/tools/shell.py
```diff
@@ -1103,9 +1287,34 @@ def _read_output_unix(
...
+                        marker_bytes = start_marker_pattern.encode()
+                        marker_pos = captured_raw.find(marker_bytes)
+                        if not seen_start_marker and marker_pos >= 0:
+                            marker_end = captured_raw.find(b"\n", marker_pos)
+                            captured_raw = (
+                                captured_raw[marker_end + 1 :]
+                                if marker_end >= 0
+                                else b""
+                            )
+                        delimiter_pos = captured_raw.rfind(b"ReturnCode:")
```
TimeToBuildBob: ⚠️ **P2** — In _read_output_unix, the delimiter exclusion uses captured_raw.rfind(b"ReturnCode:") only when the full 11-byte substring is present in the current chunk. If the shell's ReturnCode: marker is split across two 64 KiB reads (e.g. the first chunk ends with ...ReturnCod), rfind returns -1 and the partial marker bytes are not trimmed from captured_raw before captured_bytes is incremented. A command whose real output is max_output_bytes - 9 bytes, with the chunk ending at ...ReturnCode, yields a count of max_output_bytes + 1, tripping the cap and returning -125 even though the subprocess output is under the cap. This contradicts the comment at line 1297 and silently kills commands near the cap.

ErikBjare:
@TimeToBuildBob fix

gptme/tools/shell.py
```diff
@@ -1103,9 +1287,34 @@ def _read_output_unix(
...
+                        if (
+                            delimiter_pos >= 0
+                            and self.delimiter.encode() in captured_raw[delimiter_pos:]
+                        ):
+                            captured_raw = captured_raw[:delimiter_pos]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 29b008337: byte accounting now strips only the shell protocol shape where `ReturnCode:` and the delimiter occur on the same line. Marker-like command output split across lines remains fully counted. Added focused positive/negative tests; the shell byte-cap suite passes 16/16.

TimeToBuildBob:
Fixed the latest Greptile P1 in `29b008337`: protocol-marker exclusion now requires `ReturnCode:` and the delimiter on the same line, so marker-like command output on separate lines cannot bypass the byte cap. Added focused positive/negative regressions; `uv run pytest tests/test_shell_max_output_bytes.py -q` passes 16/16, and ruff/check-format pass. Fresh CI is running.

TimeToBuildBob:
Fixed the remaining two bob-ai-review P2s in `487af6cd2`:

**P2 `8a086ad660f6`** — Windows producer now excludes protocol-marker bytes from byte accounting, mirroring the Unix path. `read_stream` tracks whether the `START_OF_COMMAND_OUTPUT` line has been seen; pre-marker bytes don't count, and the return-marker line is stripped via `_strip_shell_return_marker`. A command whose real output is exactly `max_output_bytes` now behaves identically on Windows and Unix.

**P2 `10e830fb3051`** — `test_cap_total_output_size_bounded` assertion changed from `cap + 2*chunk` to `2*cap + chunk`. The old bound was non-hermetic (relied on the OS pipe buffer being small). The correct theoretical maximum is pre-cap reads (`cap`) + drain budget in `_kill_for_byte_cap` (`cap`) + one read chunk; the new bound holds regardless of pipe-buffer size.

`uv run pytest tests/test_shell_max_output_bytes.py -q` passes 16/16. CI running.

gptme/tools/shell.py
TimeToBuildBob: Fixed in 487af6cd2: the Windows `read_stream` producer now tracks the start marker and excludes pre-marker bytes plus the return-marker line from byte accounting, using the same `_strip_shell_return_marker` helper as the Unix path. A command whose real output equals `max_output_bytes` now behaves identically on both platforms.

tests/test_shell_max_output_bytes.py
TimeToBuildBob: Fixed in 487af6cd2: changed assertion from `cap + 2*chunk` to `2*cap + chunk + 1024`. The old bound was non-hermetic (relied on the kernel pipe buffer being ≤64 KiB). The correct theoretical max is pre-cap reads (cap) + drain budget in `_kill_for_byte_cap` (cap) + one read chunk; the new bound holds on any system. Updated the docstring accordingly.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Done — all findings from `@TimeToBuildBob fix` were addressed in `29b008337` and `487af6cd2`. The focused byte-cap suite passed 16/16, the full CI matrix is green, Greptile converged at 5/5, and the final head is now merged as `a5d356fa2`.

PR #3805, opened 2026-09-10
fix(shell): survive command timeouts, restore cwd/env after a shell death, tell the model

TimeToBuildBob:
Stacked on #3802 (uses its `_handle_shell_exit` EOF detection; retarget to master once that merges). Not the same bug as #3803 (`time` script parsing), just the same file.

## Problem

Erik: *"i still see `WARNING Warning: shell process died, restarting` from time to time … around things that break/stall in weird ways … worried this happens in autonomous runs too."*

It does, and far more often than the warning suggests. Every way the persistent bash can die was audited with reproductions through `ShellSession` and the tool layer:

| cause | before | after |
|---|---|---|
| `exit` / `exec` / `kill -9 $$` / `kill -- -$$` / `set -e` + failing step / `exec 1>&-` | spin on the closed pipe until `GPTME_SHELL_TIMEOUT` (20 min default, 120 s in Bob's runs) → `-124` with no explanation; restart only on the *next* command; cwd (server mode) and all exported env silently gone | returns at once with the real status (#3802), shell restarted, cwd + exported vars restored, note in the tool response |
| **command timeout** | `os.killpg(getpgid(bash))` — bash is the group leader, so **every timeout killed the shell**; silent restart on the next command, env gone | descendants get SIGTERM→SIGKILL, bash stays alive and reports the kill; still returns `-124`; group kill + restart only if bash itself is stuck (pure-bash loop) |
| shell died between commands (OOM, external kill, `close()` from the conversation registry on another thread) | `BrokenPipeError` → silent restart, or `Shell error: I/O operation on closed file` forever | liveness check before the write → restart + note |
| stdout closed but bash alive (`exec 1>&-; echo`) | wedged: every later command times out, never restarts | EOF → restart + note |

Bob's autonomous data since 2026-09-01: **977 timeouts in 34,318 shell calls (28 / 1,000), in 38 % of sessions** — each one a shell death pre-fix. 41 % of the timed-out blocks start with `set -euo pipefail`/`set -e` (61 % had `set -e` in that or an earlier block): with errexit sticky in the persistent shell, the first failing step *exits bash* and the model sees a 120 s "timeout". Sessions whose stderr tail contains the warning hit the 50-minute session timeout 68 % of the time vs 0.4 % base rate. Only 4 warnings reached the launch logs (stderr tail is printed on failure only).

## Fix

- **Timeout keeps bash alive**: `_kill_descendants()` walks `/proc/<pid>/task/*/children` (falls back to `pgrep -P`), SIGTERM then SIGKILL; the read loop waits ≤2 s for the delimiter and returns `-124` as before. Fallback to the old group kill + restart when bash does not answer.
- **State restore**: every command appends `{ printf 'cd -- %q\n' "$PWD"; export -p; } > <per-session tmpfile>` after the delimiter (never delays output); `_init()` sources it first on restart. `set -e`, functions, aliases, unexported vars and `&` jobs are intentionally not restored.
- **Model-visible note**: `ShellSession.consume_restart_notice()`; `execute_shell_impl` appends it to the tool response (single yield, same rule as the workspace hint): why the shell died, what was restored, what is gone.
- **No double execution**: a command that killed the shell is never re-run; the only retry left is `BrokenPipeError` on the initial write, where bash never received the statement.
- Windows read loop untouched; snapshot is POSIX-only.

## Tests

`tests/test_shell_restart.py` (15): each death cause restores cwd/env and returns promptly, external kill / closed stdin, no re-run (`{ echo x >> f; kill -9 $$; }` runs once), timeout keeps pid + env and kills TERM-ignoring grandchildren, pure-bash loop fallback, state file lifecycle, tool response carries the note. Existing shell suites: 769 passed; ruff + mypy clean.

Follow-up worth its own PR: warn or scope `set -e` at the top of a block — it is the dominant self-inflicted death and now merely fails fast instead of wedging.

### DIFF
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1088,14 +1088,19 @@ class ShellSession:
 
         full_command = f"echo {start_marker_pattern}\n"  # Start marker first
         full_command += f"{command}\n"
-        full_command += f"echo ReturnCode:$? {self.delimiter}\n"
         if self._state_path:
-            # Snapshot cwd + exported env after every command so a restart can
-            # restore them (runs after the delimiter, so it never delays output).
+            # Snapshot cwd + exported env before reporting completion. This
+            # closes the race where run() returned and the shell died before
+            # the state from the successful command reached the snapshot.
             full_command += (
+                "_gptme_rc=$?; "
                 "{ printf 'cd -- %q\\n' \"$PWD\"; export -p; } > "
-                f"{shlex.quote(self._state_path)} 2>/dev/null || true\n"
+                f"{shlex.quote(self._state_path)} 2>/dev/null || true; "
+                f"echo ReturnCode:$_gptme_rc {self.delimiter}; "
+                "unset _gptme_rc\n"
             )
+        else:
+            full_command += f"echo ReturnCode:$? {self.delimiter}\n"
 
         try:
             self.process.stdin.write(full_command)
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1583,12 +1583,27 @@ class ShellSession:
                                 return self._kill_for_byte_cap(
                                     stdout, stderr, output, max_output_bytes
                                 )
+                            restored = self._has_state_snapshot()
                             self.restart()
+                            if restored:
+                                note = (
+                                    "cwd and exported variables were restored "
+                                    "from a snapshot"
+                                )
+                            else:
+                                note = "cwd was reset; exported variables are gone"
                             stderr.append(
-                                "\n[gptme] The command closed a persistent shell "
-                                "output pipe, so a fresh shell was started; cwd, "
-                                "variables and `&` jobs from the old shell are "
-                                "gone.\n"
+                                f"\n[gptme] The command closed a persistent shell "
+                                f"output pipe, so a fresh shell was started; {note}. "
+                                f"Shell functions, aliases, unexported variables "
+                                f"and `&` jobs from the old shell are gone.\n"
+                            )
+                            self._restart_notice = (
+                                f"Note: the shell exited (closed its output pipe) "
+                                f"during this command, so gptme started a fresh "
+                                f"shell. {note[0].upper() + note[1:]}. Shell functions, "
+                                f"aliases, unexported variables and background jobs "
+                                f"of the old shell are gone."
                             )
                             return (
                                 -1,
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -1616,12 +1616,17 @@ class ShellSession:
                                 f"Shell functions, aliases, unexported variables "
                                 f"and `&` jobs from the old shell are gone.\n"
                             )
-                            self._restart_notice = (
+                            restart_notice = (
                                 f"Note: the shell exited (closed its output pipe) "
                                 f"during this command, so gptme started a fresh "
                                 f"shell. {note[0].upper() + note[1:]}. Shell functions, "
                                 f"aliases, unexported variables and background jobs "
                                 f"of the old shell are gone."
+                            )
+                            self._restart_notice = (
+                                restart_notice
+                                if not self._restart_notice
+                                else self._restart_notice + "\n" + restart_notice
                             )
                             return (
                                 -1,
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -557,12 +557,14 @@ class ShellSession:
     _state_path: str | None  # cwd + exported-env snapshot, restored on restart
     _restart_notice: str | None  # pending note for the model about a restart
     _restarting: bool
+    _closed: bool
 
     def __init__(self, cwd: str | None = None) -> None:
         self._cwd = cwd
         self._memory_limit = _get_memory_limit()
         self._restart_notice = None
         self._restarting = False
+        self._closed = False
         self._state_path = self._create_state_file()
         self._init()
 
@@ -993,9 +995,12 @@ class ShellSession:
         # Diagnostic logging for Issue #408: Log command start
         logger.debug(f"Shell: Running command: {command[:200]}")
 
-        # The shell may have died between commands (external kill, OOM, or
-        # closed by the conversation registry from another thread). Nothing has
-        # been sent yet, so restarting here cannot double-execute anything.
+        # The shell may have died between commands (external kill or OOM).
+        # Nothing has been sent yet, so restarting here cannot double-execute
+        # anything. An explicit close is terminal: reviving a shell removed by
+        # SESSION_END would leak an unregistered process.
+        if self._closed:
+            raise RuntimeError("Shell session is closed")
         if self.process.poll() is not None or self.process.stdin.closed:
             self._restart_after_death(self.process.returncode, "before this command")
             assert self.process.stdin
@@ -1940,6 +1945,9 @@ class ShellSession:
             logger.warning(f"Error terminating process: {e}")
 
     def close(self):
+        if not self._restarting:
+            self._closed = True
+
         # Close stdin to signal no more input
         if self.process.stdin:
             self.process.stdin.close()
@@ -1964,6 +1972,7 @@ class ShellSession:
         try:
             self.close()
             self._init()
+            self._closed = False
         finally:
             self._restarting = False
 
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -672,6 +672,8 @@ class ShellSession:
         Never re-runs the command that was in flight: it may have partially or
         fully executed (e.g. ``git commit && kill -9 $$``).
         """
+        if self._closed:
+            return
         logger.warning(
             "Warning: shell process died (%s, %s), restarting",
             _describe_exit_status(status),
@@ -1589,6 +1591,12 @@ class ShellSession:
                                     stdout, stderr, output, max_output_bytes
                                 )
                             restored = self._has_state_snapshot()
+                            if self._closed:
+                                return (
+                                    -1,
+                                    trim_blank_lines("".join(stdout)),
+                                    trim_blank_lines("".join(stderr)),
+                                )
                             self.restart()
                             if restored:
                                 note = (
@@ -1968,11 +1976,12 @@ class ShellSession:
                 pass
 
     def restart(self):
+        if self._closed:
+            return
         self._restarting = True
         try:
             self.close()
             self._init()
-            self._closed = False
         finally:
             self._restarting = False
 
--- a/gptme/tools/shell.py
+++ b/gptme/tools/shell.py
@@ -43,6 +43,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from collections.abc import Generator
@@ -480,6 +481,71 @@ def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
     return [fd for fd, _event in poller.poll(timeout_ms)]
 
 
+def _child_pids(pid: int) -> list[int]:
+    """Return the direct children of ``pid`` (Linux ``/proc``, else ``pgrep``)."""
+    task_dir = f"/proc/{pid}/task"
+    try:
+        children: list[int] = []
+        for tid in os.listdir(task_dir):
+            with open(f"{task_dir}/{tid}/children") as f:
+                children.extend(int(p) for p in f.read().split())
+        return children
+    except (OSError, ValueError):
+        pass
+    try:
+        out = subprocess.run(
+            ["pgrep", "-P", str(pid)],
+            capture_output=True,
+            text=True,
+            timeout=5,
+            check=False,
+        ).stdout
+        return [int(p) for p in out.split()]
+    except (OSError, subprocess.SubprocessError, ValueError):
+        return []
+
+
+def _descendant_pids(pid: int) -> list[int]:
+    """Return every live descendant of ``pid`` (breadth-first), excluding ``pid``."""
+    seen: set[int] = set()
+    queue = [pid]
+    while queue:
+        for child in _child_pids(queue.pop()):
+            if child not in seen:
+                seen.add(child)
+                queue.append(child)
+    return sorted(seen)
+
+
+def _kill_descendants(pid: int) -> None:
+    """SIGTERM then SIGKILL every descendant of ``pid`` without touching ``pid``.
+
+    Used on command timeout so the persistent bash survives: killing the whole
+    process group (bash is its leader) would take the shell down with the
+    command, losing cwd/env and triggering a restart on the next command.
+    """
+    for sig in (signal.SIGTERM, signal.SIGKILL):
+        for child in _descendant_pids(pid):
+            try:
+                os.kill(child, sig)
+            except (ProcessLookupError, PermissionError):
+                pass
+        if sig == signal.SIGTERM:
+            time.sleep(0.1)
+
+
+def _describe_exit_status(status: int | None) -> str:
+    if status is None:
+        return "closed its stdout/stderr while still running"
+    if status < 0:
+        try:
+            name = signal.Signals(-status).name
+        except ValueError:
+            name = f"signal {-status}"
+        return f"killed by {name}"
+    return f"code {status}"
+
+
 class ShellSession:
     process: subprocess.Popen
     stdout_fd: int
@@ -488,10 +554,16 @@ class ShellSession:
     start_marker: str  # Fix for Issue #408: Add start marker to prevent output mixing
     _cwd: str | None  # Workspace directory for this session (thread-safe)
     _memory_limit: int | None  # Address-space ceiling in bytes (None = off)
+    _state_path: str | None  # cwd + exported-env snapshot, restored on restart
+    _restart_notice: str | None  # pending note for the model about a restart
+    _restarting: bool
 
     def __init__(self, cwd: str | None = None) -> None:
         self._cwd = cwd
         self._memory_limit = _get_memory_limit()
+        self._restart_notice = None
+        self._restarting = False
+        self._state_path = self._create_state_file()
         self._init()
 
         # close on exit
@@ -546,6 +618,16 @@ class ShellSession:
         self.delimiter = "END_OF_COMMAND_OUTPUT"
         self.start_marker = "START_OF_COMMAND_OUTPUT"
 
+        # After a restart, restore cwd and exported variables from the snapshot
+        # taken after the last command the previous shell completed. Must run
+        # first: every run() below rewrites the snapshot with the new shell's state.
+        if self._restarting and self._has_state_snapshot():
+            assert self._state_path is not None
+            self.run(
+                f"source {shlex.quote(self._state_path)} >/dev/null 2>&1 || true",
+                output=False,
+            )
+
         # set GIT_PAGER=cat
         self.run("export PAGER=")
         self.run("export GH_PAGER=")
@@ -556,6 +638,66 @@ class ShellSession:
         self.run("export VISUAL=true")
         # make Python output unbuffered by default for better UX
         self.run("export PYTHONUNBUFFERED=1")
+
+    @staticmethod
+    def _create_state_file() -> str | None:
+        if _is_windows:
+            return None
+        try:
+            fd, path = tempfile.mkstemp(prefix="gptme-shell-state-", suffix=".sh")
+            os.close(fd)
+            return path
+        except OSError as e:
+            logger.warning(f"Could not create shell state snapshot file: {e}")
+            return None
+
+    def _has_state_snapshot(self) -> bool:
+        if not self._state_path:
+            return False
+        try:
+            return os.path.getsize(self._state_path) > 0
+        except OSError:
+            return False
+
+    def consume_restart_notice(self) -> str | None:
+        """Return (and clear) the pending note about a shell restart, if any."""
+        notice, self._restart_notice = self._restart_notice, None
+        return notice
+
+    def _restart_after_death(self, status: int | None, when: str) -> None:
+        """Restart the shell after it died and queue a note for the model.
+
+        Never re-runs the command that was in flight: it may have partially or
+        fully executed (e.g. ``git commit && kill -9 $$``).
+        """
+        logger.warning(
+            "Warning: shell process died (%s, %s), restarting",
+            _describe_exit_status(status),
+            when,
+        )
+        restored = self._has_state_snapshot()
+        self.restart()
+        _, cwd, _ = self.run("pwd", output=False)
+        cwd = cwd.strip()
+        if restored:
+            state = (
+                f"Working directory ({cwd}) and exported variables were "
+                "restored from a snapshot taken after the last completed "
+                "command."
+            )
+        else:
+            state = (
+                f"Working directory was reset to {cwd}; environment variables "
+                "exported by earlier commands are gone."
+            )
+        notice = (
+            f"Note: the shell exited ({_describe_exit_status(status)}) {when}, "
+            f"so gptme started a fresh shell. {state} Shell functions, aliases, "
+            "unexported variables and background jobs of the old shell are gone."
+        )
+        self._restart_notice = (
+            notice if not self._restart_notice else self._restart_notice + "\n" + notice
+        )
 
     def run(
         self, code: str, output=True, timeout: float | None = None
@@ -851,6 +993,13 @@ class ShellSession:
         # Diagnostic logging for Issue #408: Log command start
         logger.debug(f"Shell: Running command: {command[:200]}")
 
+        # The shell may have died between commands (external kill, OOM, or
+        # closed by the conversation registry from another thread). Nothing has
+        # been sent yet, so restarting here cannot double-execute anything.
+        if self.process.poll() is not None or self.process.stdin.closed:
+            self._restart_after_death(self.process.returncode, "before this command")
+            assert self.process.stdin
+
         # Redirect stdin to /dev/null to prevent commands from inheriting bash's pipe stdin
         # Use shlex to properly parse commands and respect quotes
         # Only add for commands that don't already redirect stdin
@@ -933,20 +1082,28 @@ class ShellSession:
         full_command = f"echo {start_marker_pattern}\n"  # Start marker first
         full_command += f"{command}\n"
         full_command += f"echo ReturnCode:$? {self.delimiter}\n"
+        if self._state_path:
+            # Snapshot cwd + exported env after every command so a restart can
+            # restore them (runs after the delimiter, so it never delays output).
+            full_command += (
+                "{ printf 'cd -- %q\\n' \"$PWD\"; export -p; } > "
+                f"{shlex.quote(self._state_path)} 2>/dev/null || true\n"
+            )
+
         try:
             self.process.stdin.write(full_command)
+            self.process.stdin.flush()
         except BrokenPipeError:
-            # process has died
+            # Died between the liveness check and the write: the command was
+            # never received, so a single retry is safe.
             if tries == 0:
-                # log warning and restart, once
-                logger.warning("Warning: shell process died, restarting")
-                self.restart()
+                self._restart_after_death(
+                    self.process.poll(), "before this command was received"
+                )
                 return self._run_pipe(
                     command, output=output, tries=tries + 1, timeout=timeout
                 )
             raise
-
-        self.process.stdin.flush()
 
         # Issue #408: Track whether we've seen the start marker for this command
         seen_start_marker = False
@@ -1333,25 +1490,38 @@ class ShellSession:
         """Read command output on Unix using select()."""
         assert select is not None
         captured_bytes = 0
+        timed_out = False  # children killed; waiting for bash to emit delimiter
+        grace_deadline = 0.0
         try:
             while True:
                 # Calculate remaining timeout
                 select_timeout = None
                 if timeout and start_time:
                     elapsed = time.time() - start_time
-                    if elapsed >= timeout:
-                        # Timeout exceeded
+                    if elapsed >= timeout and not timed_out:
+                        # Timeout exceeded: kill the command's processes but
+                        # keep bash alive (it is the process-group leader, so
+                        # killpg would take it down and lose cwd/env). Bash
+                        # then reports the kill via the delimiter line.
                         logger.info(f"Command timed out after {timeout} seconds")
-                        # Terminate the entire process group (bash + all child processes)
+                        _kill_descendants(self.process.pid)
+                        timed_out = True
+                        grace_deadline = time.time() + 2.0
+                    elif timed_out and time.time() >= grace_deadline:
+                        # Bash itself is stuck (or the command was a builtin):
+                        # fall back to killing the whole group and restarting.
                         try:
                             pgid = os.getpgid(self.process.pid)
                             os.killpg(pgid, signal.SIGTERM)
                             time.sleep(0.1)  # Give it a moment to terminate
                             if self.process.poll() is None:
                                 os.killpg(pgid, signal.SIGKILL)
+                            self.process.wait(timeout=1.0)
                         except Exception as e:
                             logger.warning(f"Error terminating timed-out process: {e}")
-
+                        self._restart_after_death(
+                            self.process.poll(), "after the command timed out"
+                        )
                         partial_stdout = trim_blank_lines("".join(stdout))
                         partial_stderr = trim_blank_lines("".join(stderr))
                         return (
@@ -1360,8 +1530,8 @@ class ShellSession:
                             partial_stderr,
                         )  # Use timeout exit code (124)
 
-                    select_timeout = min(
-                        1.0, timeout - elapsed
+                    select_timeout = (
+                        0.1 if timed_out else min(1.0, timeout - elapsed)
                     )  # Check at least every second
 
                 rlist = _wait_readable([self.stdout_fd, self.stderr_fd], select_timeout)
@@ -1511,6 +1681,8 @@ class ShellSession:
                             rc_matches = re_returncode.findall(line)
                             if rc_matches:
                                 return_code = int(rc_matches[-1])
+                            if timed_out:
+                                return_code = -124  # timeout exit code
                             # if command is cd, update working directory
                             if (
                                 command == "cd" or command.startswith("cd ")
@@ -1638,19 +1810,19 @@ class ShellSession:
             rc: int | None = self.process.wait(timeout=1.0)
         except subprocess.TimeoutExpired:
             rc = None
+        # Drain output still buffered on the unclosed pipe (a diagnostic can
+        # arrive after the EOF-triggering close), enforcing the same byte cap,
+        # then restart and restore cwd/exported env. Never re-run the command.
         captured_bytes = self._drain_closed_shell_pipes(
             stdout, stderr, output, max_output_bytes, captured_bytes
         )
         if captured_bytes > max_output_bytes:
             return self._kill_for_byte_cap(stdout, stderr, output, max_output_bytes)
-        logger.warning(
-            "Shell process exited during command (code %s), restarting shell", rc
-        )
-        self.restart()
+        self._restart_after_death(rc, "during this command")
         stderr.append(
-            f"\n[gptme] The shell exited (code {rc}), so a fresh shell was "
-            "started; cwd, variables and `&` jobs from the old shell are gone. "
-            "Don't run `exit` in the tool shell — it never needs to be exited.\n"
+            f"\n[gptme] The shell exited ({_describe_exit_status(rc)}), so a "
+            "fresh shell was started (see note below). Don't run `exit` in the "
+            "tool shell — it never needs to be exited.\n"
         )
         return (
             rc if rc is not None else -1,
@@ -1766,9 +1938,19 @@ class ShellSession:
         if self.process.stderr:
             self.process.stderr.close()
 
+        if self._state_path and not self._restarting:
+            try:
+                os.unlink(self._state_path)
+            except OSError:
+                pass
+
     def restart(self):
-        self.close()
-        self._init()
+        self._restarting = True
+        try:
+            self.close()
+            self._init()
+        finally:
+            self._restarting = False
 
 
 _shell_var: ContextVar[ShellSession | None] = ContextVar("shell", default=None)
@@ -2338,6 +2520,11 @@ def execute_shell_impl(
             hint = _check_workspace_config()
             if hint:
                 workspace_hint_content = "\n\n" + hint.content
+    # If the shell died and was restarted during this call, tell the model
+    # (same single-yield rule as the workspace hint).
+    restart_notice = shell.consume_restart_notice()
+    if restart_notice:
+        workspace_hint_content += "\n\n" + restart_notice
 
     # stdout/stderr were already streamed live by ShellSession (both the
     # persistent-pipe and the TTY path stream as bytes arrive), so the
--- a/tests/test_shell_restart.py
+++ b/tests/test_shell_restart.py
@@ -95,12 +95,11 @@ def test_killed_between_commands_restarts_before_next_command(shell, tmp_path):
     assert _state(shell)[:2] == (str(tmp_path), "alive")
 
 
-def test_closed_stdin_restarts_instead_of_raising(shell):
-    """close() from another thread must not leave a permanently broken shell."""
+def test_explicitly_closed_shell_stays_closed(shell):
+    """SESSION_END cleanup must not be undone by a stale ContextVar reference."""
     shell.close()
-    rc, out, _ = shell.run("echo back", output=False)
-    assert (rc, out.strip()) == (0, "back")
-    assert "before this command" in (shell.consume_restart_notice() or "")
+    with pytest.raises(RuntimeError, match="Shell session is closed"):
+        shell.run("echo back", output=False)
 
 
 def test_command_that_kills_shell_is_not_rerun(shell, tmp_path):
--- a/tests/test_shell_restart.py
+++ b/tests/test_shell_restart.py
@@ -84,6 +84,20 @@ def test_stdout_closed_but_shell_alive_is_restarted(shell, tmp_path):
     assert (pid != old_pid, cwd, marker) == (True, str(tmp_path), "alive")
 
 
+def test_closed_pipe_preserves_existing_restart_notice(shell):
+    """Two shell deaths before notice consumption must both reach the model."""
+    os.kill(shell.process.pid, signal.SIGKILL)
+    shell.process.wait(timeout=5)
+
+    shell.run("exec 1>&-", output=False, timeout=30)
+
+    notice = shell.consume_restart_notice()
+    assert notice
+    assert "before this command" in notice
+    assert "during this command" in notice
+    assert notice.index("before this command") < notice.index("during this command")
+
+
 def test_killed_between_commands_restarts_before_next_command(shell, tmp_path):
     """External kill (OOM killer, registry close from another thread)."""
     os.kill(shell.process.pid, signal.SIGKILL)
--- a/tests/test_shell_restart.py
+++ b/tests/test_shell_restart.py
@@ -16,6 +16,7 @@ A command timeout must kill the command's processes, not bash itself.
 import os
 import signal
 import sys
+import threading
 import time
 from unittest.mock import patch
 
@@ -100,6 +101,31 @@ def test_explicitly_closed_shell_stays_closed(shell):
     shell.close()
     with pytest.raises(RuntimeError, match="Shell session is closed"):
         shell.run("echo back", output=False)
+
+
+def test_close_racing_with_eof_does_not_restart_shell(shell):
+    """SESSION_END must win when an active read sees the pipes close."""
+    command_started = threading.Event()
+    run_finished = threading.Event()
+
+    def run_command():
+        command_started.set()
+        try:
+            shell.run("sleep 30", output=False)
+        finally:
+            run_finished.set()
+
+    thread = threading.Thread(target=run_command)
+    thread.start()
+    assert command_started.wait(timeout=1)
+    time.sleep(0.1)
+
+    shell.close()
+    thread.join(timeout=5)
+
+    assert run_finished.is_set()
+    assert shell._closed
+    assert shell.process.poll() is not None
 
 
 def test_command_that_kills_shell_is_not_rerun(shell, tmp_path):
--- a/tests/test_shell_restart.py
+++ b/tests/test_shell_restart.py
@@ -1,0 +1,202 @@
+"""Persistent-shell death and restart semantics.
+
+Every way the persistent bash can die (``exit``, ``exec``, ``kill -9 $$``,
+``set -e`` + failure, closing its stdout, an external kill, a command timeout)
+must:
+
+* return promptly instead of spinning on the closed pipe until the timeout,
+* restart the shell with the working directory and exported variables
+  restored from the snapshot taken after the last completed command,
+* queue a note that the tool response shows the model,
+* never re-run the command that killed the shell (it may have executed).
+
+A command timeout must kill the command's processes, not bash itself.
+"""
+
+import os
+import signal
+import sys
+import time
+from unittest.mock import patch
+
+import pytest
+
+from gptme.tools.shell import ShellSession, execute_shell
+
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32", reason="POSIX process semantics"
+)
+
+
+@pytest.fixture
+def shell(tmp_path):
+    sh = ShellSession()
+    sh.run(f"cd {tmp_path} && export RESTART_MARKER=alive", output=False)
+    yield sh
+    sh.close()
+
+
+def _state(sh: ShellSession) -> tuple[str, str, int]:
+    rc, out, _ = sh.run("echo $PWD $RESTART_MARKER", output=False)
+    assert rc == 0
+    cwd, _, marker = out.strip().partition(" ")
+    return cwd, marker, sh.process.pid
+
+
+@pytest.mark.parametrize(
+    ("cmd", "expected_rc"),
+    [
+        ("exit 3", 3),
+        ("exec true", 0),
+        ("kill -9 $$", -signal.SIGKILL),
+        ("kill -- -$$", -signal.SIGTERM),
+        ("set -e; false", 1),
+        ("exec 1>&-", 0),
+    ],
+)
+def test_shell_death_restarts_and_restores_state(shell, tmp_path, cmd, expected_rc):
+    old_pid = shell.process.pid
+    start = time.monotonic()
+    rc, _, _ = shell.run(cmd, output=False)  # no timeout: must not spin
+    assert time.monotonic() - start < 6.0
+    assert rc == expected_rc
+
+    cwd, marker, pid = _state(shell)
+    assert pid != old_pid
+    assert cwd == str(tmp_path)
+    assert marker == "alive"
+
+    notice = shell.consume_restart_notice()
+    assert notice and "fresh shell" in notice
+    assert "restored from a snapshot" in notice
+    assert shell.consume_restart_notice() is None
+
+
+def test_stdout_closed_but_shell_alive_is_restarted(shell, tmp_path):
+    """``exec 1>&-; echo`` leaves bash alive but mute: restart instead of wedging."""
+    old_pid = shell.process.pid
+    start = time.monotonic()
+    rc, _, _ = shell.run("exec 1>&-; echo hi", output=False, timeout=30)
+    assert time.monotonic() - start < 8.0
+    assert rc != 0
+    cwd, marker, pid = _state(shell)
+    assert (pid != old_pid, cwd, marker) == (True, str(tmp_path), "alive")
+
+
+def test_killed_between_commands_restarts_before_next_command(shell, tmp_path):
+    """External kill (OOM killer, registry close from another thread)."""
+    os.kill(shell.process.pid, signal.SIGKILL)
+    shell.process.wait(timeout=5)
+
+    rc, out, _ = shell.run("echo alive-again", output=False)
+    assert (rc, out.strip()) == (0, "alive-again")
+    notice = shell.consume_restart_notice()
+    assert notice and "before this command" in notice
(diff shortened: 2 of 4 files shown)


TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 5 findings disposed (rejected; accepted-tradeoff)

| --- | --- | --- |
| `gptme/tools/shell.py:1246` | P1 | rejected — The verifier claims 'a TERM-ignoring descendant can survive past the 2s grace and trigger the group kill.' That step is  |
| `gptme/tools/shell.py:431` | P1 | accepted-tradeoff |
| `gptme/tools/shell.py:520` | P1 | rejected — The review says a daemon that double-forks is still part of the descendant cleanup target. That verification premise is  |
| `gptme/tools/shell.py:669` | P1 | rejected — The review's verification uses a command writing to SNAPSHOT_PATH. That premise is wrong: the random parent-owned pathna |
| `gptme/tools/shell.py:1106` | P1 | rejected — The review asserts Popen.stdin.write can return a short count that this code ignores. That premise is wrong for the io.T |
| `gptme/tools/shell.py:537` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:1246` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:909` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:437` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:1423` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:999` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/shell.py:614` | P2 | accepted-tradeoff |
| `gptme/tools/shell.py:1360` | P2 | fixed in `35586f9b9` |
| `gptme/tools/shell.py:618` | P2 | rejected |
| `tests/test_shell_restart.py:54` | P2 | superseded by latest review (not reproduced) |
| `tests/test_shell_restart.py:133` | P2 | superseded by latest review (not reproduced) |
| `tests/test_shell_restart.py:135` | P2 | superseded by latest review (not reproduced) |

</details>




This PR reworks persistent-shell death handling in gptme/tools/shell.py. It adds a state snapshot file (cwd + exported env) written after each command, restores it on restart, kills only descendant processes on timeout (keeping bash alive), and queues a model-visible restart notice. It also updates heredoc parsing in shell_validation.py to handle tab-stripping delimiters and adds tests for both.

### Not safe to merge — 3 P1 open

Confidence 2/5

> ⚠️ **2 of 4 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>3 findings</b> · ❌ 3 P1 · 🔒 1 security</summary>

**❌ P1 high** — `gptme/tools/shell.py:520`

In `_read_output_unix`, when a timeout occurs, the code calls `_kill_descendants(self.process.pid)` and then waits for the delimiter. However, `_kill_descendants` only kills descendants, not the process group. If the command has spawned a process that is not a descendant (e.g., a daemon that double-forks and reparents to init), it will not be killed. The test `test_timeout_kills_grandchildren_and_term_ignoring_children` uses `bash -c 'trap "" TERM; (sleep 30); sleep 30'` which creates a subshell and a sleep. The subshell is a child, and the sleep is a child of the subshell, so they are descendants. But a command like `setsid sleep 30` would create a new session, and the sleep would not be a descendant of bash? Actually, `setsid` forks and the child becomes a new session leader, but it is still a descendant of bash until it reparents. If it double-forks, it may reparent to init, and then `_child_pids` would not find it. This means a timed-out command could leave a runaway process running. The PR description says 'kills the command's processes, not bash itself', but it only kills descendants. A process that daemonizes and reparents would survive. This is a real bug: a command like `nohup sleep 1000 &` (background) is not a descendant? Actually, background jobs are children of bash, so they are descendants. But if the command does `setsid sh -c 'sleep 1000' &`, the setsid process is a child, but the sleep inside may be a child of setsid, which is a descendant. However, if the command does `daemonize` which double-forks, the final process is reparented to init, so it is no longer a descendant. The timeout would not kill it, and it would keep running. This is a correctness issue: a timed-out command can leave orphaned processes. The test only checks for `sleep 30` which is a direct child, so it passes. But a daemonized process would not be killed. This is a P1 bug because it can leave runaway processes consuming resources.

<sub>**How this was verified:** Traced _kill_descendants: it calls _descendant_pids which walks /proc children. A double-forked daemon reparented to init is no longer a child of bash, so it is not killed. The test only covers direct children and grandchildren that remain in the process tree.</sub>

**❌ P1 high · 🔒 security** — `gptme/tools/shell.py:669`

In `_restart_after_death`, after restarting the shell, it runs `self.run("pwd", output=False)` to get the cwd. However, this `run` call itself can trigger another restart if the shell dies again (e.g., if the state snapshot is corrupt and sourcing it kills the shell). This could lead to infinite recursion. The `_restart_after_death` method is called from `_run_pipe` when the shell is dead, and it calls `self.restart()` which calls `_init()`, which may call `self.run()` to source the snapshot. If that `run` fails and the shell dies, it will call `_restart_after_death` again, leading to recursion. However, `_restart_after_death` checks `if self._closed: return`, but `_closed` is False during restart. There is no guard against re-entrant restart. This could cause a stack overflow or infinite loop if the snapshot is persistently bad. The test `test_snapshot_failure_does_not_kill_shell_under_set_e` sets `_state_path` to a nonexistent path, so sourcing fails but the shell survives. But if the snapshot contains a command that kills the shell (e.g., `export PATH=/nonexistent` is fine, but a malicious snapshot could contain `kill -9 $$`), then sourcing it would kill the new shell, and `_restart_after_death` would be called again. Since the snapshot is written by the shell itself, it could be corrupted by a previous command. This is a real risk: a command that writes to the snapshot file (e.g., `echo 'kill -9 $$' > $SNAPSHOT_PATH`) could cause a restart loop. The snapshot file path is not secret, and a malicious command could overwrite it. This is a security issue: a command can write to the snapshot file and cause a denial of service (infinite restart loop). The PR does not protect the snapshot file from being overwritten by the shell's own commands. The file is in /tmp with a random name, but the shell knows the path (it's in the command line), so a command can write to it. This is a P1 security issue.

```suggestion
Add a recursion guard, e.g., a counter or a flag that prevents restarting more than once per command.
```

<sub>**How this was verified:** Traced the restart path: _restart_after_death -&gt; restart -&gt; _init -&gt; run(source snapshot). If the snapshot contains a command that kills the shell, the new shell dies, and the next run() call will trigger _restart_after_death again. There is no recursion guard. The snapshot file is writable by the shell, so a command can plant a malicious snapshot.</sub>

**❌ P1 high** — `gptme/tools/shell.py:1106`

In `_read_output_unix`, when the shell dies during a command (EOF detected), the code calls `self._restart_after_death(rc, "during this command")` and then returns `(rc if rc is not None else -1, ...)`. However, `_restart_after_death` calls `self.restart()` which calls `self._init()`, which runs several `self.run()` commands to set environment variables. These `run()` calls will write to the state snapshot file, overwriting the snapshot that was just restored. That's fine because the new shell's state is being captured. But there's a subtle issue: `_restart_after_death` also calls `self.run("pwd", output=False)` to get the cwd for the notice. This `run()` call will also write to the state snapshot. The notice is queued, and the next command will consume it. The test `test_shell_death_restarts_and_restores_state` checks that the notice contains "restored from a snapshot". This works. However, the `_restart_after_death` method is called from `_run_pipe` when the shell is detected dead before the write. In that case, the command that was about to be sent is not executed, and the method restarts the shell and then the `_run_pipe` retries the command. The retry will run the command in the new shell. But the notice says "before this command", which is correct. The issue is that the retry might double-execute if the original command was partially written to the pipe before the death. The PR claims that the write is atomic (single `write` call), but `self.process.stdin.write(full_command)` could be a partial write if the pipe buffer is full. The code doesn't check the return value of `write`. If the write is partial, the command is truncated, and the retry would send the full command again, potentially executing a truncated version first. This is a real bug: `Popen.stdin.write` can return fewer bytes than the length of the string, especially for large commands. The code should use `os.write` in a loop or check the return value. This is a pre-existing issue, but the PR's retry logic makes it more likely to cause double execution. I'll report this as a bug.

```suggestion
Use a loop to write all bytes, e.g., `self.process.stdin.write(full_command)` is not guaranteed to write all; use `os.write(self.process.stdin.fileno(), full_command.encode())` in a loop.
```

<sub>**How this was verified:** Checked the snapshot write logic in _run_pipe and the restore in _init. The snapshot is written after the command, before the delimiter. If the command kills the shell, the snapshot is not written, which is correct. But if the shell dies during the snapshot write (e.g., OOM), the file could be partial. The source is wrapped in `|| true`, so a partial file would be sourced with errors ignored, but could leave the shell in a partially-restored state. This is a minor edge case.</sub>

</details>

- `gptme/tools/shell.py` — Adds state snapshot/restore, descendant-kill timeout handling, restart notices, and liveness checks before command writes.
- `gptme/tools/shell_validation.py` — Updates heredoc regex to capture tab-stripping flag and adjusts delimiter matching for tab-stripped heredocs.
- `tests/test_shell_restart.py` — Adds tests for shell death/restart semantics, state restore, timeout behavior, and restart notices.
- `tests/test_tools_shell_validation.py` — Adds tests for space-indented delimiters in plain and tab-stripping heredocs.

</details>

| `bf3a6a596d7b` | 4/5 | 8 | llm | 2026-09-10 10:21 UTC |
| `a2e5fd0aefaa` | 5/5 | 0 | llm | 2026-09-10 11:57 UTC |
| `a2e5fd0aefaa` | 3/5 | 2 | llm | 2026-09-10 12:32 UTC |
| `440c99888c2e` | 2/5 | 2 | llm | 2026-09-10 13:08 UTC |
| `fd604d15d2ed` | 4/5 | 1 | llm | 2026-09-10 17:15 UTC |
| `896d4668a447` | ? | 0 | llm | 2026-09-10 18:17 UTC |
| `182d90ddaac5` | 4/5 | 1 | llm | 2026-09-10 18:40 UTC |
| `35586f9b9112` | 4/5 | 1 | llm | 2026-09-10 19:21 UTC |

</details>

<sub>Reviewed `413a0bc61669` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 67s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell.py
```diff
@@ -456,6 +528,16 @@ def _init(self):
         self.delimiter = "END_OF_COMMAND_OUTPUT"
         self.start_marker = "START_OF_COMMAND_OUTPUT"
 
+        # After a restart, restore cwd and exported variables from the snapshot
+        # taken after the last command the previous shell completed. Must run
+        # first: every run() below rewrites the snapshot with the new shell's state.
+        if self._restarting and self._has_state_snapshot():
+            assert self._state_path is not None
+            self.run(
+                f"source {shlex.quote(self._state_path)} >/dev/null 2>&1 || true",
```
TimeToBuildBob, line 629: ⚠️ **P2** — The state snapshot is written after every command using `{ printf 'cd -- %q\n' "$PWD"; export -p; } > <state_path>`. However, the snapshot is taken *after* the command's delimiter line is echoed, but the command itself may have changed the working directory or exported variables. The snapshot is written by the same bash process, so it captures the state after the command. But there is a race: if the shell dies *during* the command (e.g., `kill -9 $$`), the snapshot is not updated, so the previous snapshot is restored. That is intended. However, if the command is a background job that continues after the command returns, the snapshot is taken immediately after the command's foreground part, but the background job may later change cwd or export variables. The snapshot is not updated after background jobs finish, so a restart after a background job changes state will restore stale state. This is a real correctness issue: the PR claims to restore cwd and exported variables from the last completed command, but background jobs are explicitly not restored (the notice says 'background jobs of the old shell are gone'), so the snapshot may be stale. But the PR's own notice says background jobs are gone, so the user is warned. However, the snapshot is taken after the command that started the background job, so it captures the state at that point, which is correct for the foreground command. The issue is if a background job modifies the environment after the snapshot, but the shell is still alive, so no restart happens. If the shell dies later, the snapshot is from the last foreground command, not the background job's changes. This is a minor edge case, but the PR's claim of restoring 'exported variables' is not fully accurate. I'll consider this a trade-off or heuristic, not a bug, because the PR explicitly says background jobs are not restored. I'll not report it as a bug, but maybe as a trade-off. However, the more concrete issue is: the snapshot is written after the delimiter, but the command may have `exit` or `exec` that kills the shell before the snapshot is written. In that case, the snapshot is not updated, and the previous snapshot is restored. That is correct. But there is a subtle bug: the snapshot is written using `export -p`, which outputs `export VAR=value` lines. When sourced, these set the variables. However, `export -p` also outputs `export PATH=...` etc. But the snapshot also includes `cd -- %q` which is a `cd` command. When sourced, it changes the directory. But the snapshot is sourced in `_init()` *before* the standard exports like `PAGER`, `GIT_PAGER`, etc. are set. So the restored environment may have the user's exported variables, but then the standard exports override some of them. For example, if the user exported `PAGER=less`, the snapshot restores it, but then `self.run("export PAGER=")` overrides it to empty. That is a real bug: the PR claims to restore exported variables, but the standard exports clobber them. The snapshot is sourced first, then the standard exports are set. So any user-exported variable that collides with the standard ones (PAGER, GH_PAGER, GIT_PAGER, EDITOR, GIT_EDITOR, VISUAL, PYTHONUNBUFFERED) will be overwritten. This is a concrete correctness issue. I'll report it as a bug. The line is in `_init()` where the snapshot is sourced, and then the standard exports are run. The quote should be the line that sources the snapshot. I'll use line 537: `f"source {shlex.quote(self._state_path)} >/dev/null 2>&1 || true",` but the actual line in the file is line 537? Let me count: line 534 is `if self._restarting and self._has_state_snapshot():`, line 535 `assert self._state_path is not None`, line 536 `self.run(`, line 537 `f"source {shlex.quote(self._state_path)} >/dev/null 2>&1 || true",`, line 538 `output=False,`. So I'll quote line 537. The standard exports are lines 542-550. So the bug is that the snapshot is sourced before the standard exports, so the standard exports override the restored values. The fix would be to source the snapshot after the standard exports, or to not override if the snapshot has a value. But the PR's intent is to restore user state, so the standard exports should not clobber. I'll report this as a bug with severity P1, class bug, security false.

gptme/tools/shell.py
```diff
@@ -1069,25 +1226,38 @@ def _read_output_unix(
...
+                        # Timeout exceeded: kill the command's processes but
+                        # keep bash alive (it is the process-group leader, so
+                        # killpg would take it down and lose cwd/env). Bash
+                        # then reports the kill via the delimiter line.
                         logger.info(f"Command timed out after {timeout} seconds")
-                        # Terminate the entire process group (bash + all child processes)
+                        _kill_descendants(self.process.pid)
+                        timed_out = True
+                        grace_deadline = time.time() + 2.0
+                    elif timed_out and time.time() >= grace_deadline:
```
TimeToBuildBob, line 1522: ⚠️ **P2** — In `_read_output_unix`, when a timeout occurs, `_kill_descendants(self.process.pid)` is called, then `timed_out = True` and `grace_deadline = time.time() + 2.0`. The loop then waits for the delimiter. If the delimiter arrives, `return_code` is set to -124. However, if the command was a builtin like `while true; do :; done`, there are no descendants, so `_kill_descendants` does nothing, and bash itself is stuck. The grace period is 2 seconds, after which the fallback kills the whole group and restarts. That is correct. But there is a subtle bug: if the command spawned a child that ignores SIGTERM and SIGKILL (e.g., a process in uninterruptible sleep), `_kill_descendants` sends SIGKILL, but the process may not die immediately. The grace period is 2 seconds, and if the child is still alive, the fallback kills the whole group, including bash, and restarts. That is acceptable. However, the bigger issue is that `_kill_descendants` only kills direct and indirect children of the bash process, but the command may have spawned a process that daemonizes and reparents to init, so it is no longer a descendant. That process would survive the timeout and keep running. The PR's test `test_timeout_kills_grandchildren_and_term_ignoring_children` checks that `pgrep -f 'sleep 30'` returns 0, but that only checks for processes with 'sleep 30' in the command line. A daemonized process might not match. This is a known limitation, but the PR claims to kill the command's processes. I'll not report this as a bug because it's a heuristic limitation. However, there is a more concrete bug: the grace period is 2 seconds, but the loop's `select_timeout` is set to 0.1 when `timed_out` is True. So the loop checks every 0.1 seconds. If the delimiter arrives within 2 seconds, it returns -124. If not, it falls back. But what if the delimiter arrives exactly at the grace deadline? The condition `elif timed_out and time.time() >= grace_deadline` is checked before reading, so if the delimiter is available but the deadline has passed, it will kill the group and restart, even though the delimiter was ready. This is a race: the delimiter might be in the pipe but not yet read, and the deadline passes, causing an unnecessary restart. The consequence is that a command that actually completed within the grace period but whose output was delayed (e.g., due to buffering) could be killed and restarted, losing the shell state. This is a real bug. The fix would be to check for readable data before the deadline, or to extend the grace period. I'll report this as a bug with severity P2, class bug, security false.

gptme/tools/shell.py
```diff
@@ -761,6 +903,13 @@ def _run_pipe(
         # Diagnostic logging for Issue #408: Log command start
         logger.debug(f"Shell: Running command: {command[:200]}")
 
+        # The shell may have died between commands (external kill, OOM, or
+        # closed by the conversation registry from another thread). Nothing has
+        # been sent yet, so restarting here cannot double-execute anything.
+        if self.process.poll() is not None or self.process.stdin.closed:
```
TimeToBuildBob, line 1006: ⚠️ **P2** — In `_run_pipe`, the liveness check is `if self.process.poll() is not None or self.process.stdin.closed:`. If the process is dead, `self.process.stdin.closed` may be True, but `self.process.poll()` is not None. The code calls `_restart_after_death(self.process.returncode, "before this command")`. However, `self.process.returncode` is only set after `poll()` or `wait()`. Since `poll()` was just called, `returncode` is set. That is fine. But there is a race: if the process dies between the `poll()` check and the `stdin.write()`, the `BrokenPipeError` handler catches it and retries. That is correct. However, the liveness check also restarts if `stdin.closed` is True, which can happen if `close()` was called from another thread. In that case, `self.process.poll()` may be None (process still alive), but `stdin.closed` is True. The code restarts, which closes the process and starts a new one. That is correct. But there is a subtle issue: if `close()` was called, it also closes stdout and stderr, and the process may still be running. The restart will kill the old process? `close()` calls `self.process.terminate()`? Let me check the `close()` method. In the truncated file, I see `close()` closes stdin, stdout, stderr, and then if `_state_path and not _restarting`, unlinks the state file. It does not terminate the process. So if `close()` is called from another thread, the process is still running, but its pipes are closed. The liveness check sees `stdin.closed` and restarts. The old process is left running, orphaned. That is a resource leak. The PR's test `test_closed_stdin_restarts_instead_of_raising` calls `shell.close()` and then `shell.run("echo back")`, which restarts. The old process is not killed. This is a real bug: closing the shell from another thread leaves the old bash process running, consuming resources. The fix would be to terminate the old process in `close()` or in the restart path. I'll report this as a bug with severity P2, class bug, security false.

gptme/tools/shell.py
```diff
@@ -390,6 +391,71 @@ def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
...
+
+
+def _kill_descendants(pid: int) -> None:
+    """SIGTERM then SIGKILL every descendant of ``pid`` without touching ``pid``.
+
+    Used on command timeout so the persistent bash survives: killing the whole
+    process group (bash is its leader) would take the shell down with the
+    command, losing cwd/env and triggering a restart on the next command.
+    """
+    for sig in (signal.SIGTERM, signal.SIGKILL):
```
TimeToBuildBob, line 527: ⚠️ **P2** — In `_kill_descendants`, the function walks `/proc/<pid>/task/*/children` to find direct children. However, on Linux, the `children` file in `/proc/<pid>/task/<tid>/children` lists the children of the thread, not the process. For a multi-threaded process, this may miss children. But bash is typically single-threaded. The fallback to `pgrep -P` is fine. However, there is a race: a child may fork a grandchild after the initial list is collected, so the grandchild is not killed. The PR's test `test_timeout_kills_grandchildren_and_term_ignoring_children` uses a bash -c that spawns a subshell and a sleep, and the test passes. But in a real scenario, a command might spawn a background process that forks after the timeout. The `_kill_descendants` function only collects the current descendants at the time of the call. If a child forks a new process after the SIGTERM but before the SIGKILL, that new process might survive. The grace period is 2 seconds, but the SIGKILL is sent immediately after SIGTERM (with a 0.1s sleep). So a process that forks during that 0.1s window could be missed. This is a minor race, but the PR claims to kill all descendants. I'll not report this as a bug because it's a race that is hard to avoid. However, there is a more concrete issue: `_kill_descendants` sends SIGTERM to all descendants, then sleeps 0.1s, then sends SIGKILL. But if a descendant is in uninterruptible sleep (D state), SIGKILL will not kill it until it wakes. The grace period is 2 seconds, and if the process is still alive, the fallback kills the whole group, including bash. That is acceptable. I'll not report this.

gptme/tools/shell.py
```diff
@@ -1246,14 +1418,13 @@ def _handle_shell_exit(
             rc: int | None = self.process.wait(timeout=1.0)
         except subprocess.TimeoutExpired:
             rc = None
-        logger.warning(
-            "Shell process exited during command (code %s), restarting shell", rc
-        )
-        self.restart()
+        # Restarts, restores cwd/exported env from the snapshot and queues the
+        # full explanation for the tool response. Never re-runs the command.
+        self._restart_after_death(rc, "during this command")
```
TimeToBuildBob, line 1859: ⚠️ **P2** — In `_read_output_unix`, when the shell exits (EOF), `_handle_shell_exit` is called. It waits for the process with a 1-second timeout. If the process does not exit within 1 second, `rc` is None, and `_restart_after_death(None, "during this command")` is called. The notice says "closed its stdout/stderr while still running". However, if the shell is still running but its stdout is closed (e.g., `exec 1>&-`), the EOF is detected, but the process is still alive. `_handle_shell_exit` waits 1 second, gets TimeoutExpired, sets rc=None, and restarts. The old process is left running. This is the same resource leak as the close() issue. The test `test_stdout_closed_but_shell_alive_is_restarted` expects a restart, but does not check that the old process is terminated. I'll combine this with the close() finding, but I'll report it separately because it's a different path. Actually, I'll report it as a separate finding: when the shell's stdout is closed but the process is alive, `_handle_shell_exit` restarts without killing the old process, leaving it running. The fix would be to terminate the old process in `_handle_shell_exit` or in `_restart_after_death`. I'll report this as a bug with severity P2.

tests/test_shell_restart.py
```diff
@@ -0,0 +1,183 @@
...
+
+@pytest.mark.parametrize(
+    ("cmd", "expected_rc"),
+    [
+        ("exit 3", 3),
+        ("exec true", 0),
+        ("kill -9 $$", -signal.SIGKILL),
+        ("kill -- -$$", -signal.SIGTERM),
+        ("set -e; false", 1),
+        ("exec 1>&-", 0),
```
TimeToBuildBob, line 55: ⚠️ **P2** — The parametrized test `test_shell_death_restarts_and_restores_state` lists `("exec 1>&-", 0)` expecting return code 0, but when stdout is closed with `exec 1>&-`, bash stays alive. The read loop sees EOF on stdout and enters `_handle_shell_exit`, which waits 1 second, gets `rc = None`, then returns `-1`. The assertion `rc == expected_rc` will fail. The expected code should be `-1` (or the test should assert non-zero) to match the actual behavior of a shell that loses its output stream but does not exit.

gptme/tools/shell.py
```diff
@@ -843,21 +992,29 @@ def _run_pipe(
         full_command = f"echo {start_marker_pattern}\n"  # Start marker first
         full_command += f"{command}\n"
         full_command += f"echo ReturnCode:$? {self.delimiter}\n"
+        if self._state_path:
+            # Snapshot cwd + exported env after every command so a restart can
+            # restore them (runs after the delimiter, so it never delays output).
+            full_command += (
+                "{ printf 'cd -- %q\\n' \"$PWD\"; export -p; } > "
```
TimeToBuildBob, line 1097: ⚠️ **P2** — The state snapshot command appended to every command is not guarded against failure. It is written as `{ printf 'cd -- %q\n' "$PWD"; export -p; } > <state_path> 2>/dev/null` with no `|| true`. If the redirection fails (e.g., the temporary directory is read-only in a sandbox, or the disk is full), the compound command returns non-zero. Because `set -e` is sticky in the persistent shell and the snapshot runs at top level after the delimiter, this causes bash to exit after every command that follows a `set -e`. The shell then restarts, losing all unexported state, and the model sees spurious restart notices. The command should be made best-effort so a snapshot failure cannot kill the shell.

tests/test_shell_restart.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def test_timeout_kills_grandchildren_and_term_ignoring_children(shell):
+    pid = shell.process.pid
+    rc, _, _ = shell.run(
+        "bash -c 'trap \"\" TERM; (sleep 30); sleep 30'", output=False, timeout=1.0
+    )
+    assert rc == -124
+    assert shell.process.pid == pid
+    rc, out, _ = shell.run("pgrep -f 'sleep 30' | wc -l", output=False)
```
TimeToBuildBob: ⚠️ **P2** — The test `test_timeout_kills_grandchildren_and_term_ignoring_children` asserts that no `sleep 30` process remains using `pgrep -f 'sleep 30'`. The `-f` flag matches the full command line, so any unrelated process on the system whose command line contains `sleep 30` (e.g., a different test, a build tool, or a user-launched sleep) will cause this assertion to fail. The test is therefore non-hermetic and can fail on a busy machine even when the timed-out processes were correctly killed. The check should be narrowed to the specific processes this test spawned, for example by recording the PIDs before the timeout and asserting those specific PIDs are gone.

TimeToBuildBob:
Rebased this PR onto the current `#3802` branch head (`ac08c001a`) to clear the stale-branch conflict, resolving one overlap in `gptme/tools/shell.py`'s `_handle_shell_exit` (kept the drain from #3802's closed-pipe handling and folded it together with this PR's `_restart_after_death` restore path).

While in there I addressed a concrete finding raised in review:

- **Snapshot writes were able to kill the shell under sticky `set -e`** — the per-command `{ ... } > <state>` redirect had no `|| true`, so a failed redirect (read-only tmp dir, full disk) returned non-zero and `set -e` exited bash after every command. That's exactly the spurious shell death this PR exists to remove. Now best-effort with `|| true`, plus a regression test (`test_snapshot_failure_does_not_kill_shell_under_set_e`).
- **Hardened the timeout-descendants test** — `pgrep -f 'sleep 30'` substring-matches an unrelated `sleep 300` in a shared container; anchored it to `sleep 30$`.

Verified: `tests/test_shell_restart.py` 16 passed, `tests/test_tools_shell.py` 141 passed (incl. the drain test), ruff clean.

TimeToBuildBob:
@greptileai review

tests/test_shell_restart.py
```diff
@@ -0,0 +1,202 @@
...
+def test_timeout_kills_grandchildren_and_term_ignoring_children(shell, tmp_path):
+    # Anchor the pgrep regex so it can't substring-match an unrelated process
+    # (e.g. a concurrent `sleep 300` in a shared CI/agent container).
+    pid = shell.process.pid
+    rc, _, _ = shell.run(
+        "bash -c 'trap \"\" TERM; (sleep 30); sleep 30'", output=False, timeout=1.0
+    )
+    assert rc == -124
+    assert shell.process.pid == pid
+    rc, out, _ = shell.run("pgrep -f 'sleep 30$' | wc -l", output=False)
```
TimeToBuildBob, line 174: ⚠️ **P2** — The test test_timeout_kills_grandchildren_and_term_ignoring_children uses `pgrep -f 'sleep 30$' | wc -l` to check that no sleep 30 processes remain. However, the pgrep pattern 'sleep 30$' could match other processes that have 'sleep 30' at the end of their command line, but the test comment says it anchors the regex to avoid substring matches. The issue is that the test runs in a shared CI environment where other tests might have a `sleep 30` running. The test does not isolate the process group, so if another test or process has a `sleep 30` running, the assertion could fail. However, the test uses a unique command `bash -c 'trap "" TERM; (sleep 30); sleep 30'` which is not likely to be used elsewhere. But the pgrep pattern 'sleep 30$' would match any process whose command line ends with 'sleep 30', including the one from this test. The test asserts that after the timeout, the count is 0, meaning all sleep 30 processes are killed. But if another test has a `sleep 30` running, it would be counted and the assertion would fail. This is a non-hermetic test that depends on the environment. The test should use a more specific pattern or isolate the process group. I'll report this as a test-defect. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed one of the Greptile summary findings (Recovery Message Misstates State): the closed-pipe timeout branch called `restart()`, which **restores** cwd + exported env from the per-session snapshot, but the appended stderr message hardcoded that "cwd, variables and `&` jobs from the old shell are gone" — directly contradicting the actual restore.

Now the note reflects real behavior: it reports what was restored from the snapshot when present, and what is genuinely reset when not (matching `_restart_after_death`'s accuracy), and queues the same corrected text as the model-visible restart notice.

`tests/test_shell_restart.py` 16 passed, `tests/test_tools_shell.py` unchanged, ruff + mypy clean. Re-triggered Greptile re-review on the new head (`440c99888`).

On the other Greptile P1s: "Snapshot trails command completion" and "Timed-out workers can survive" are inherent to this design (snapshot is intentionally written after the delimiter so it never delays output; the descendant-kill covers the group-bash death case) — the snapshot path is best-effort and the fallback covers the miss. "Close discards recovery state" is by-design for a real `close()` (registry/server shutdown), and the reuse-after-`close()` test covers the in-band restart path which restores state.

gptme/tools/shell.py
```diff
@@ -1070,25 +1227,38 @@ def _read_output_unix(
...
+                    if elapsed >= timeout and not timed_out:
+                        # Timeout exceeded: kill the command's processes but
+                        # keep bash alive (it is the process-group leader, so
+                        # killpg would take it down and lose cwd/env). Bash
+                        # then reports the kill via the delimiter line.
                         logger.info(f"Command timed out after {timeout} seconds")
-                        # Terminate the entire process group (bash + all child processes)
+                        _kill_descendants(self.process.pid)
+                        timed_out = True
+                        grace_deadline = time.time() + 2.0
```
TimeToBuildBob, line 1521: ❌ **P1** — In _read_output_unix, when a command times out, _kill_descendants sends SIGTERM then SIGKILL to all descendants of bash. However, the grace period is only 2 seconds (grace_deadline = time.time() + 2.0). If a descendant ignores SIGTERM and takes longer than 2 seconds to die, or if the process tree is deep and _kill_descendants takes time, the fallback kills the whole process group including bash, causing a restart. This is a race: the code intends to keep bash alive on timeout, but a slow-to-die child can still trigger the group kill and restart, losing cwd/env. The test test_timeout_kills_command_but_keeps_shell uses `sleep 30` which dies quickly on SIGTERM, so it passes, but a command that traps SIGTERM and sleeps longer than 2s would fail. This is a real defect: the grace period is too short for the intended behavior.

gptme/tools/shell.py
```diff
@@ -391,6 +392,71 @@ def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
...
+    queue = [pid]
+    while queue:
+        for child in _child_pids(queue.pop()):
+            if child not in seen:
+                seen.add(child)
+                queue.append(child)
+    return sorted(seen)
+
+
+def _kill_descendants(pid: int) -> None:
```
TimeToBuildBob, line 520: ❌ **P1** — In _kill_descendants, the function walks /proc/<pid>/task/*/children to find direct children. However, on Linux, the children file lists only direct children, and the code iterates over all TIDs in the task directory. But a process can have multiple threads, and each thread's children file may list the same child? Actually, /proc/<pid>/task/<tid>/children lists the children of the thread group, not per-thread. So iterating over all TIDs and reading the same children file multiple times is redundant but not incorrect. However, there's a race: a child may exit between reading the children file and sending the signal, which is handled by ProcessLookupError. But the bigger issue is that _kill_descendants only kills descendants, not the process group. If a command spawns a process that sets its own process group (e.g., `setsid`), that process won't be a descendant of bash? Actually, it will still be a descendant in the process tree, so it will be found. But if a process daemonizes and reparents to init, it won't be a descendant anymore, so it won't be killed. This is a known limitation, but the PR claims to kill all descendants. The test test_timeout_kills_grandchildren_and_term_ignoring_children uses `bash -c 'trap "" TERM; (sleep 30); sleep 30'` which creates a child bash that ignores TERM, and a grandchild sleep. The _kill_descendants sends SIGTERM to the child bash (which ignores it) and the grandchild sleep (which dies). Then after 0.1s, it sends SIGKILL to the child bash, which kills it. So the test passes. But if a process double-forks and reparents, it won't be killed. This is a trade-off, not a bug.

gptme/tools/shell.py
TimeToBuildBob: Rejecting as a false positive — the premise is contradicted by the code.

The finding claims the 2s `grace_deadline` is "too short" because "a slow-to-die child" (one that ignores SIGTERM) might survive past it and trigger the group kill. That cannot happen: `_kill_descendants()` (`gptme/tools/shell.py:431`) sends **SIGTERM then SIGKILL** to every descendant, with only a 0.1s sleep between the two passes. SIGKILL is unroutable — a TERM-ignoring child is force-killed by the second pass, *before `_kill_descendants` even returns*.

`grace_deadline = time.time() + 2.0` is set **after** `_kill_descendants(...)` returns (line 1246), so the entire descendant-scan-and-kill happens before the 2-second clock starts. The 2s window is not a grace period for children to die — it is purely the time we wait for **bash itself** (which we deliberately did not kill) to notice its children have exited and emit the `ReturnCode:$?` delimiter. If bash is healthy, it emits within milliseconds. The group-kill fallback at line 1247 fires only when bash has *not* emitted after a full second — i.e. bash itself is wedged (pure-bash loop / stuck builtin), which is exactly the case the fallback exists for.

The existing test `test_timeout_kills_command_but_keeps_shell` covers a plain command; `test_timeout_kills_grandchildren_and_term_ignoring_children` covers the TERM-ignoring grandchild case and passes precisely because SIGKILL follows SIGTERM. No fix needed.

gptme/tools/shell.py
TimeToBuildBob: Accepting as a documented trade-off — the finding itself concludes "this is a trade-off, not a bug."

Two points in the finding are worth separating:

1. **"Iterating over all TIDs in the task directory is redundant"** — correct, and harmless. `/proc/<pid>/task/<tid>/children` reports the same child set from every TID in the thread group, so the BFS sees the same children multiple times until `seen` dedupes. This costs a few extra reads on a rare, already-slow failure path and changes nothing about correctness. Not worth churning the hot path.

2. **"A double-forked / reparented-to-init process won't be killed"** — genuine, and inherent to *any* descendant-based kill: once a process is reparented to PID 1 it is no longer a descendant of the shell at all, so walking `/proc/.../children` from bash side cannot reach it. That is the definition of a daemonized process — the shell has released it, and killing it would be killing an unrelated process. The alternative (kill the process *group*) is exactly what the PR deliberately avoids, because bash is the group leader and `killpg` would take the shell down and lose cwd/env — defeating the entire purpose of this change.

This is the correct engineering boundary: on timeout we kill everything the shell still *owns*, and leave the shell itself alive. A process that has deliberately detached itself is outside that scope by construction. Documented in the `_kill_descendants` docstring ("without touching `pid`") and exercised by `test_timeout_kills_grandchildren_and_term_ignoring_children`.

TimeToBuildBob:
Rebased the stack onto the current #3802 head (`3d9aaace8`) and force-pushed `27b5bd6e9`, clearing the stale-base conflict.

Conflict resolution deliberately keeps #3802's structured `background: true` interface and its closed-pipe restart/drain path; this PR now contains only its four shell-recovery commits above that base. Verified locally:
- `tests/test_shell_restart.py`: 16 passed
- `tests/test_tools_shell_validation.py`: 527 passed
- focused `tests/test_tools_shell.py` compatibility slice: 8 passed
- Ruff check + format check: clean

GitHub is recalculating mergeability/checks for the new head.

TimeToBuildBob:
The #3802 base moved again while the first rebase was being verified. I rebased once more onto its current head (`69c5e4e37`) and force-pushed `fd604d15d`; GitHub now reports `MERGEABLE` / `CLEAN` against that exact base.

The second rebase was conflict-free and preserves the same four-commit recovery-only diff. Re-ran the combined focused suites on the final tree: 543 passed, with Ruff and `git diff --check` clean.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -491,6 +573,66 @@ def _init(self):
...
+        fully executed (e.g. ``git commit && kill -9 $$``).
+        """
+        logger.warning(
+            "Warning: shell process died (%s, %s), restarting",
+            _describe_exit_status(status),
+            when,
+        )
+        restored = self._has_state_snapshot()
+        self.restart()
+        _, cwd, _ = self.run("pwd", output=False)
```
TimeToBuildBob, line 684: ⚠️ **P2** — In _restart_after_death, after restarting the shell, the code runs `self.run("pwd", output=False)` to determine the working directory for the notice. But this call to self.run() itself goes through _run_pipe, which at the top checks `if self.process.poll() is not None or self.process.stdin.closed:` and could call _restart_after_death recursively if the newly restarted shell immediately dies (e.g. the state snapshot contains a `cd` to a directory that no longer exists, causing the sourced script to fail and, under `set -e`? Actually the source is wrapped in `|| true`, so it won't die). More importantly, the `pwd` run happens after restart, and if the state snapshot was restored, the cwd is the restored one. But the notice is built with `cwd` from this `pwd` call, which is correct. However, the recursive risk: if the shell dies again during this `pwd` call, _restart_after_death is called again, which calls self.run("pwd") again, leading to infinite recursion if the shell keeps dying. This is unlikely but possible if the state file is corrupt or the environment is broken. The code does not guard against reentrancy. A concrete trigger: the state snapshot file contains `cd /nonexistent` and `export` lines; sourcing it with `|| true` won't kill bash, but if the snapshot contains a command that causes bash to exit (e.g. `exit`), then the source would exit the shell. The snapshot is generated by `{ printf 'cd -- %q\n' "$PWD"; export -p; }`, which only contains cd and export lines, so it cannot contain `exit`. But a malicious or corrupted state file could. Since the state file is in /tmp with a predictable prefix, another user could potentially write to it? mkstemp creates it with 0600, so no. Still, the recursion is a latent bug.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -1368,9 +1564,20 @@ def close(self):
...
+                pass
+
     def restart(self):
-        self.close()
-        self._init()
+        self._restarting = True
+        try:
+            self.close()
+            self._init()
+            self._closed = False
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed **Cleanup Can Restart Shell** in `896d4668a`.

The explicit-close state now wins at every restart boundary: `_restart_after_death()` and `restart()` refuse to revive a closed session, and the direct closed-pipe restart branch checks the terminal state before spawning Bash. Added a regression that closes the session while `sleep 30` is actively being read and verifies the reader exits, the session stays closed, and no replacement process remains.

Verified on the pushed head: `tests/test_shell_restart.py` 17 passed; Ruff check/format and `git diff --check` clean. CI and review are now recalculating.

gptme/tools/shell.py
TimeToBuildBob: No separate change needed: the reviewed recovery path only sources its own mode-0600 snapshot generated from `cd -- %q` plus `export -p`, and the source is guarded with `|| true`; it cannot synthesize an `exit`. The hypothetical requires external corruption of a private file or a broken Bash process on every restart, neither of which this PR should add lifecycle machinery to handle. The concrete cleanup race found by Greptile was real and is fixed separately in `896d4668a`.

gptme/tools/shell.py
```diff
@@ -843,21 +992,29 @@ def _run_pipe(
         full_command = f"echo {start_marker_pattern}\n"  # Start marker first
         full_command += f"{command}\n"
         full_command += f"echo ReturnCode:$? {self.delimiter}\n"
+        if self._state_path:
+            # Snapshot cwd + exported env after every command so a restart can
+            # restore them (runs after the delimiter, so it never delays output).
+            full_command += (
+                "{ printf 'cd -- %q\\n' \"$PWD\"; export -p; } > "
+                f"{shlex.quote(self._state_path)} 2>/dev/null\n"
+            )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed **Snapshot trails command completion** in `182d90dda`.

The wrapper now saves `$?`, writes the cwd/export snapshot, and only then emits the completion delimiter with the saved status. A successful `run()` can therefore no longer return while its state snapshot still trails behind. The snapshot redirect remains best-effort under sticky `set -e`.

Added `test_snapshot_is_written_before_run_returns`; verified the pushed head with all 18 shell-restart tests, all 135 shell-tool tests, Ruff check/format, `git diff --check`, and pre-commit.

gptme/tools/shell.py
```diff
@@ -1154,12 +1336,33 @@ def _read_output_unix(
...
-                                "\n[gptme] The command closed a persistent shell "
-                                "output pipe, so a fresh shell was started; cwd, "
-                                "variables and `&` jobs from the old shell are "
-                                "gone.\n"
+                                f"\n[gptme] The command closed a persistent shell "
+                                f"output pipe, so a fresh shell was started; {note}. "
+                                f"Shell functions, aliases, unexported variables "
+                                f"and `&` jobs from the old shell are gone.\n"
+                            )
+                            self._restart_notice = (
```
TimeToBuildBob: ⚠️ **P2** — In _read_output_unix, the EOF handling block sets _restart_notice directly, but it does not use _restart_after_death. This means the notice is not concatenated with any existing notice. If a previous restart notice was queued but not consumed, it will be overwritten. For example, if the shell died before the command and _restart_after_death queued a notice, then during the command the shell closes its output pipe, the EOF block overwrites _restart_notice with a new notice, losing the first one. The observable consequence is that the model may not see the first restart notice. This is a bug.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
TimeToBuildBob: Fixed in `35586f9b9`: closed-pipe recovery now appends its notice to any notice already queued by a pre-write restart instead of overwriting it. Added a regression that kills Bash between commands, then closes the replacement shell output during the next command, and verifies both ordered explanations survive until consumption. Verified: 19 restart tests and 135 shell-tool tests pass; Ruff, formatting, `git diff --check`, and pre-commit pass.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
TimeToBuildBob: Fixed in 35586f9b9 and verified by the focused consecutive-restart regression; resolving the now-outdated finding thread after the fresh-head implementation and reply.

gptme/tools/shell.py
```diff
@@ -491,6 +575,68 @@ def _init(self):
...
+        if self._closed:
+            return
+        logger.warning(
+            "Warning: shell process died (%s, %s), restarting",
+            _describe_exit_status(status),
+            when,
+        )
+        restored = self._has_state_snapshot()
+        self.restart()
+        _, cwd, _ = self.run("pwd", output=False)
```
TimeToBuildBob, line 684: ⚠️ **P2** — In _restart_after_death, after restart() the code runs `self.run("pwd", output=False)` to get the cwd. But restart() calls _init(), which itself calls self.run() for the state restore and for setting PAGER etc. If the shell dies again during those _init() runs (e.g. the state snapshot contains a command that kills the shell, or the environment is corrupted), _restart_after_death's subsequent self.run("pwd") could trigger another _restart_after_death recursively, leading to infinite recursion or a stack overflow. The _restarting flag is set during restart() but cleared in the finally block before _restart_after_death continues, so a nested death during the pwd run would not be suppressed. This is a robustness issue: a shell that dies repeatedly on startup (e.g. due to a bad state file) could cause unbounded recursion.

TimeToBuildBob: No code change: this repeats the already adjudicated private-snapshot corruption hypothesis. `_init()` does run through `self.run()`, but the only sourced state is a mode-0600 file generated by gptme itself from `cd -- %q` and `export -p`, sourced under `|| true`; it cannot contain a shell-exiting command. A newly spawned Bash that independently dies on every startup is outside this PR’s recoverable state and adding recursive-recovery machinery would not make that environment healthy.

TimeToBuildBob:
@TimeToBuildBob review

TimeToBuildBob:
The stacked base advanced again while the fresh-head review was running. Rebasing onto current #3802 head `8b28981f1` exposed two overlaps with #3802's newly landed output-byte cap; I combined rather than dropped either behavior:

- timeout recovery retains `captured_bytes` accounting while keeping Bash alive and killing descendants;
- closed-pipe drains remain byte-capped before state-aware restart/notification.

Force-pushed rebased head `413a0bc61`; GitHub now reports `MERGEABLE` / `CLEAN`. Verified on that exact tree:
- `tests/test_shell_restart.py`: 19 passed
- `tests/test_tools_shell.py`: 139 passed
- `tests/test_tools_shell_validation.py`: 527 passed
- Ruff check/format and `git diff --check`: clean

Fresh CI and both machine-review gates must now recalculate at the new head; not merging against the stale pre-rebase verdicts.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -480,6 +481,71 @@ def _wait_readable(fds: list[int], timeout: float | None) -> list[int]:
...
+    queue = [pid]
+    while queue:
+        for child in _child_pids(queue.pop()):
+            if child not in seen:
+                seen.add(child)
+                queue.append(child)
+    return sorted(seen)
+
+
+def _kill_descendants(pid: int) -> None:
```
TimeToBuildBob, line 520: ❌ **P1** — In `_read_output_unix`, when a timeout occurs, the code calls `_kill_descendants(self.process.pid)` and then waits for the delimiter. However, `_kill_descendants` only kills descendants, not the process group. If the command has spawned a process that is not a descendant (e.g., a daemon that double-forks and reparents to init), it will not be killed. The test `test_timeout_kills_grandchildren_and_term_ignoring_children` uses `bash -c 'trap "" TERM; (sleep 30); sleep 30'` which creates a subshell and a sleep. The subshell is a child, and the sleep is a child of the subshell, so they are descendants. But a command like `setsid sleep 30` would create a new session, and the sleep would not be a descendant of bash? Actually, `setsid` forks and the child becomes a new session leader, but it is still a descendant of bash until it reparents. If it double-forks, it may reparent to init, and then `_child_pids` would not find it. This means a timed-out command could leave a runaway process running. The PR description says 'kills the command's processes, not bash itself', but it only kills descendants. A process that daemonizes and reparents would survive. This is a real bug: a command like `nohup sleep 1000 &` (background) is not a descendant? Actually, background jobs are children of bash, so they are descendants. But if the command does `setsid sh -c 'sleep 1000' &`, the setsid process is a child, but the sleep inside may be a child of setsid, which is a descendant. However, if the command does `daemonize` which double-forks, the final process is reparented to init, so it is no longer a descendant. The timeout would not kill it, and it would keep running. This is a correctness issue: a timed-out command can leave orphaned processes. The test only checks for `sleep 30` which is a direct child, so it passes. But a daemonized process would not be killed. This is a P1 bug because it can leave runaway processes consuming resources. _(2/3 passes — pass 1 looked and disagreed)_

gptme/tools/shell.py
```diff
@@ -557,6 +641,68 @@ def _init(self):
...
+            return os.path.getsize(self._state_path) > 0
+        except OSError:
+            return False
+
+    def consume_restart_notice(self) -> str | None:
+        """Return (and clear) the pending note about a shell restart, if any."""
+        notice, self._restart_notice = self._restart_notice, None
+        return notice
+
+    def _restart_after_death(self, status: int | None, when: str) -> None:
```
TimeToBuildBob, line 669: ❌ **P1** — In `_restart_after_death`, after restarting the shell, it runs `self.run("pwd", output=False)` to get the cwd. However, this `run` call itself can trigger another restart if the shell dies again (e.g., if the state snapshot is corrupt and sourcing it kills the shell). This could lead to infinite recursion. The `_restart_after_death` method is called from `_run_pipe` when the shell is dead, and it calls `self.restart()` which calls `_init()`, which may call `self.run()` to source the snapshot. If that `run` fails and the shell dies, it will call `_restart_after_death` again, leading to recursion. However, `_restart_after_death` checks `if self._closed: return`, but `_closed` is False during restart. There is no guard against re-entrant restart. This could cause a stack overflow or infinite loop if the snapshot is persistently bad. The test `test_snapshot_failure_does_not_kill_shell_under_set_e` sets `_state_path` to a nonexistent path, so sourcing fails but the shell survives. But if the snapshot contains a command that kills the shell (e.g., `export PATH=/nonexistent` is fine, but a malicious snapshot could contain `kill -9 $$`), then sourcing it would kill the new shell, and `_restart_after_death` would be called again. Since the snapshot is written by the shell itself, it could be corrupted by a previous command. This is a real risk: a command that writes to the snapshot file (e.g., `echo 'kill -9 $$' > $SNAPSHOT_PATH`) could cause a restart loop. The snapshot file path is not secret, and a malicious command could overwrite it. This is a security issue: a command can write to the snapshot file and cause a denial of service (infinite restart loop). The PR does not protect the snapshot file from being overwritten by the shell's own commands. The file is in /tmp with a random name, but the shell knows the path (it's in the command line), so a command can write to it. This is a P1 security issue. _(2/3 passes — pass 1 looked and disagreed)_

gptme/tools/shell.py
```diff
@@ -932,22 +1088,35 @@ def _run_pipe(
...
+                "{ printf 'cd -- %q\\n' \"$PWD\"; export -p; } > "
+                f"{shlex.quote(self._state_path)} 2>/dev/null || true; "
+                f"echo ReturnCode:$_gptme_rc {self.delimiter}; "
+                "unset _gptme_rc\n"
+            )
+        else:
+            full_command += f"echo ReturnCode:$? {self.delimiter}\n"
+
         try:
             self.process.stdin.write(full_command)
```
TimeToBuildBob, line 1106: ❌ **P1** — In `_read_output_unix`, when the shell dies during a command (EOF detected), the code calls `self._restart_after_death(rc, "during this command")` and then returns `(rc if rc is not None else -1, ...)`. However, `_restart_after_death` calls `self.restart()` which calls `self._init()`, which runs several `self.run()` commands to set environment variables. These `run()` calls will write to the state snapshot file, overwriting the snapshot that was just restored. That's fine because the new shell's state is being captured. But there's a subtle issue: `_restart_after_death` also calls `self.run("pwd", output=False)` to get the cwd for the notice. This `run()` call will also write to the state snapshot. The notice is queued, and the next command will consume it. The test `test_shell_death_restarts_and_restores_state` checks that the notice contains "restored from a snapshot". This works. However, the `_restart_after_death` method is called from `_run_pipe` when the shell is detected dead before the write. In that case, the command that was about to be sent is not executed, and the method restarts the shell and then the `_run_pipe` retries the command. The retry will run the command in the new shell. But the notice says "before this command", which is correct. The issue is that the retry might double-execute if the original command was partially written to the pipe before the death. The PR claims that the write is atomic (single `write` call), but `self.process.stdin.write(full_command)` could be a partial write if the pipe buffer is full. The code doesn't check the return value of `write`. If the write is partial, the command is truncated, and the retry would send the full command again, potentially executing a truncated version first. This is a real bug: `Popen.stdin.write` can return fewer bytes than the length of the string, especially for large commands. The code should use `os.write` in a loop or check the return value. This is a pre-existing issue, but the PR's retry logic makes it more likely to cause double execution. I'll report this as a bug. _(2/3 passes — pass 1 looked and disagreed)_

gptme/tools/shell.py
TimeToBuildBob: Rejected as outside the foreground-command contract. The review's own example requires a command to deliberately daemonize and reparent itself to PID 1; once reparented it is no longer owned by the persistent Bash process, so no descendant-only cleanup can identify it without a separate cgroup/subreaper execution boundary. This PR intentionally preserves the Bash process (and its cwd/exported environment) while terminating every child still in its tree. Background work that should outlive a foreground call has the structured `background: true` path from the stacked base. Expanding this fix into per-command cgroup containment would be a different feature and would destroy the state-preservation goal on platforms without that boundary.

gptme/tools/shell.py
TimeToBuildBob: Rejected because the claimed write path is not exposed to the command. The snapshot pathname is generated in the parent Python process with `mkstemp` mode 0600, is never exported into Bash's environment, and is only interpolated into gptme-owned protocol text after the user command. The review says a command can overwrite `$SNAPSHOT_PATH`, but no such variable exists. The file's contents are generated solely from Bash `%q` of `PWD` plus `export -p`, and sourcing is guarded with `|| true`. A same-UID process that introspects another process and corrupts its private temp files is outside the shell tool's trust boundary; it could already kill the shell directly.

gptme/tools/shell.py
TimeToBuildBob: Rejected: the proposed partial-write execution path does not exist through Python's text stream API here. `Popen.stdin` is an `io.TextIOWrapper`; `write(full_command)` either accepts the complete Python string into its buffered writer and returns its character count, or raises. `flush()` then propagates a broken pipe. It does not expose a short raw `os.write` result to this caller. The one retry is entered only on `BrokenPipeError`; if `flush()` fails after bytes reached Bash, whether any prefix executed is inherently unknowable, but that is a pre-existing pipe-failure ambiguity, not evidence that checking TextIOWrapper's return value or looping `os.write` would make retry safe. The PR already refuses to retry after an observed start marker or EOF during execution.

PR #3809, opened 2026-09-10
fix(gh): dispatch structured 'command' kwarg calls

TimeToBuildBob:
## Problem

The `gh` tool's structured (native "tool"-format) schema advertises a single
`url` parameter, but `execute_gh` dispatches on positional `args`. Native
callers therefore arrive as `args=None, kwargs={"url": ...}`, fall through every
native handler, and reach `_passthrough_gh(None, None)`:

```
Error: No command provided
```

This reproduces on current master (and installed v0.33.0):

```python
>>> from gptme.tools.gh import execute_gh
>>> [m.content for m in execute_gh(None, None, {"url": "gptme/gptme#1"})]
['Error: No command provided']
```

Two separate sessions hit this on 2026-09-10 and vented it to the friction
ledger (`resolution_owner=tooling`): *"Native gh tool rejects every documented
invocation with 'No command provided' although its schema exposes only url"*.

## Fix

- `execute_gh` now accepts a `command` kwarg and `shlex.split`s it into the
  positional args before dispatch, so both invocation styles work.
- Malformed quoting is reported instead of silently falling through.
- The exposed parameter changes from `url` to `command` — a bare ref cannot
  express *which* subcommand to run, so `url`-only calls could never work.
- Tool-format examples now render with `command` (they previously rendered as an
  empty `{}`).

## Testing

- `pytest tests/test_tools_gh.py -q` → **78 passed** (6 new tests covering native
  dispatch, pass-through, quoting errors, and the exposed schema).
- `ruff check` / `ruff format --check` clean.
- mypy reports no new errors in `gptme/tools/gh.py`.
- Manual repro now succeeds: `execute_gh(None, None, {"command": "issue view gptme/gptme#1"})`
  returns the issue content.

### DIFF
--- a/gptme/tools/gh.py
+++ b/gptme/tools/gh.py
@@ -556,6 +556,17 @@ def execute_gh(
     Native handlers for high-value operations (issue view, pr view/status/checks/merge,
     run view).  Everything else passes through to the gh CLI unchanged.
     """
+    # Structured ("tool"-format) callers supply the whole command as a single
+    # ``command`` kwarg and leave ``args`` empty. Recover the positional args so
+    # the native dispatch below works for both invocation styles (markdown
+    # blocks set ``args`` positionally).
+    if not args and kwargs and kwargs.get("command"):
+        try:
+            args = shlex.split(kwargs["command"])
+        except ValueError as e:
+            yield Message("system", f"Error parsing command: {e}")
+            return
+
     if args and len(args) >= 2 and args[0] == "pr" and args[1] == "merge":
         yield from _handle_pr_merge(args, kwargs)
 
@@ -667,22 +678,30 @@ Native paths help avoid hallucination:
 - `gh pr merge <ref> ...` adds squash-by-default and optional head-commit protection
 - `gh run view <run-id>` extracts failed-job logs
 
-All other `gh` subcommands pass through unchanged."""
+All other `gh` subcommands pass through unchanged.
+
+Structured tool calls pass the whole command as `command`, e.g.
+`{"command": "issue view owner/repo#42"}`."""
 
 
 def examples(tool_format):
+    def render(args: list[str]) -> str:
+        # Native ("tool") calls carry a single `command` string parameter;
+        # markdown/XML carry the command positionally.
+        if tool_format == "tool":
+            return ToolUse("gh", None, None, {"command": shlex.join(args)}).to_output(
+                tool_format
+            )
+        return ToolUse("gh", args, None).to_output(tool_format)
+
     return f"""
 > User: read PR #123 on owner/repo
 > Assistant:
-{ToolUse("gh", ["pr", "view", "owner/repo#123"], None).to_output(tool_format)}
+{render(["pr", "view", "owner/repo#123"])}
 
 > User: check CI status for this PR
 > Assistant:
-{
-        ToolUse(
-            "gh", ["pr", "status", "https://github.com/owner/repo/pull/123"], None
-        ).to_output(tool_format)
-    }
+{render(["pr", "status", "https://github.com/owner/repo/pull/123"])}
 > System:
 PR #123 checks (abc1234):
 Total: 6 checks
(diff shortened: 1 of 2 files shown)


TimeToBuildBob:
Verified the current head locally: all 78 `tests/test_tools_gh.py` tests pass, Ruff check/format are clean, and a real structured pass-through call plus malformed-quoting path both behave correctly. Greptile's 5/5 review covered the pre-force-push SHA; the force-push changed only commit metadata, but the merge gate correctly treats that review as stale. CI and a current-head machine review are still pending, so I am not merging yet.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `gptme/tools/gh.py:563` | P2 | rejected |

</details>




This PR changes the gh tool's structured invocation to accept a single `command` kwarg instead of `url`. It adds shlex splitting of the `command` string into positional args before native dispatch, updates the tool schema parameter from `url` to `command`, and updates tool-format examples to render the `command` kwarg. It also adds tests covering the new dispatch path, pass-through, parse errors, and schema exposure.

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `gptme/tools/gh.py:563`

The new `command` kwarg path in `execute_gh` splits the command string with `shlex.split`, but the native handlers still read the reference from `args[2]` via `_extract_url`. For a structured call like `{"command": "pr status owner/repo#1"}`, the split produces `["pr", "status", "owner/repo#1"]`, which works. However, the old `url` kwarg is no longer accepted: `_extract_url` falls back to `kwargs.get("url")`, but the schema now only exposes `command`, so any caller that still passes `url` (e.g., a previously generated tool call or a hardcoded integration) will get `args=None` and `kwargs={"url": ...}`, which does not match the new `command` check, falls through to `_passthrough_gh(None, None)`, and yields "Error: No command provided". This is a breaking change for existing callers that used the documented `url` parameter, and the PR description itself notes the old schema was broken, but the fix removes the `url` path entirely rather than supporting both. The consequence is that any existing caller using `url` (even though it was broken before) will now get a different error, and the PR does not provide a migration path.

```suggestion
if args and args[0] == 'gh': args = args[1:]
```

<sub>**How this was verified:** Checked the diff and current file: the new code only checks for `command` kwarg, and `_extract_url` still references `kwargs.get("url")` but the schema parameter is now `command`. The PR description confirms the old `url` parameter was broken, but this change removes it without backward compatibility.</sub>

</details>

- `gptme/tools/gh.py` — Adds shlex splitting of `command` kwarg into args, changes tool parameter from `url` to `command`, and updates examples to render `command` kwarg in tool format.
- `tests/test_tools_gh.py` — Adds tests for command kwarg dispatch, pass-through, parse errors, schema parameter, and tool-format examples.

</details>

| `7d9d773e4e8d` | 3/5 | 2 | llm | 2026-09-10 20:49 UTC |

</details>

<sub>Reviewed `7d9d773e4e8d` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 17s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/gh.py
```diff
@@ -556,6 +556,17 @@ def execute_gh(
     Native handlers for high-value operations (issue view, pr view/status/checks/merge,
     run view).  Everything else passes through to the gh CLI unchanged.
     """
+    # Structured ("tool"-format) callers supply the whole command as a single
+    # ``command`` kwarg and leave ``args`` empty. Recover the positional args so
+    # the native dispatch below works for both invocation styles (markdown
+    # blocks set ``args`` positionally).
+    if not args and kwargs and kwargs.get("command"):
```
TimeToBuildBob, line 563: ⚠️ **P2** — The new `command` kwarg path in `execute_gh` splits the command string with `shlex.split`, but the native handlers still read the reference from `args[2]` via `_extract_url`. For a structured call like `{"command": "pr status owner/repo#1"}`, the split produces `["pr", "status", "owner/repo#1"]`, which works. However, the old `url` kwarg is no longer accepted: `_extract_url` falls back to `kwargs.get("url")`, but the schema now only exposes `command`, so any caller that still passes `url` (e.g., a previously generated tool call or a hardcoded integration) will get `args=None` and `kwargs={"url": ...}`, which does not match the new `command` check, falls through to `_passthrough_gh(None, None)`, and yields "Error: No command provided". This is a breaking change for existing callers that used the documented `url` parameter, and the PR description itself notes the old schema was broken, but the fix removes the `url` path entirely rather than supporting both. The consequence is that any existing caller using `url` (even though it was broken before) will now get a different error, and the PR does not provide a migration path. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Rejected as a false positive. The prior `url` schema could carry only a reference, not an operation, so no structured `url` call could ever dispatch successfully; preserving that input would preserve a broken, ambiguous contract rather than provide compatibility. The supported migration is the required `command` parameter documented by this PR. The suggested `args[0] == "gh"` change is unrelated to the reported compatibility concern.

PR #3818, opened 2026-09-12
fix(config): align CLI model precedence

### LINKED ISSUE #3814: Model precedence: MODEL overrides [models].default on the CLI, contradicting init_model and docs
The precedence between `[models].default` and the `MODEL` env var is implemented, documented, and reported inconsistently.

## What the code does

**CLI path, new conversations: `MODEL` wins.** `setup_config_from_cli` resolves the model as `--model` → saved conversation model → `config.get_env("MODEL")`, without consulting `[models].default`:

- https://github.com/gptme/gptme/blob/31a385390/gptme/config/cli_setup.py#L132-L141

That value is stored as the conversation's `ChatConfig.model`:

- https://github.com/gptme/gptme/blob/31a385390/gptme/config/cli_setup.py#L299-L303

`cli/main.py` then passes `config.chat.model` into `chat()`, which calls `init(model, ...)` → `init_model(model)`.

**`init_model`: `[models].default` wins.** Its fallback is `chat model > [models].default > MODEL`:

- https://github.com/gptme/gptme/blob/31a385390/gptme/init.py#L144-L151

But on the CLI path `model` is already set from `MODEL`, so this fallback only runs when `MODEL` is unset. The net effect for CLI users is that `MODEL` (from the shell, or any `[env]` section, including project `gptme.toml`) overrides `[models].default`.

## Where this shows up

- **Docs** (`docs/config.rst`, "How model selection works") list `[models].default` (3) before `MODEL` (4), which is the opposite of CLI behavior.
- **`get_default_model_source`** (`gptme/config/user.py`) says "`[models].default` ... takes priority, then the `MODEL` env var", so a UI or status reporting the default model's source can name `[models].default` while `MODEL` is actually used.
- **Model selection trace** (`init.py`, [L303-L315](https://github.com/gptme/gptme/blob/31a385390/gptme/init.py#L303-L315)) records `source_kind = "cli"` because `requested_model` is set, even when the value came from `MODEL`.

## Expected

One precedence order, used everywhere and documented once. Presumably: `--model` > saved conversation model > `[models].default` > `MODEL` > auto-detect. The alternative is to make `MODEL` intentionally higher (e.g. so a project `gptme.toml` `[env] MODEL` can override a global default) and document that. Either way, the provenance trace should record the real source.

## Repro (from code reading, not yet run)

```toml
# ~/.config/gptme/config.toml
[models]
default = "anthropic/claude-sonnet-4-6"
```

```sh
MODEL=openai/gpt-5-mini gptme "hi"
# expected per docs: anthropic/claude-sonnet-4-6
# expected per cli_setup.py: openai/gpt-5-mini, with trace source "cli"
```

Other entry points (server `api_v2`, ACP, TUI) weren't checked; they may resolve the model differently again.

## Context

Found while restructuring the docs in #3813. That PR avoids stating the precedence between `[models].default` and `MODEL` until this is resolved.

Ricky-7-Yan:
## Summary

Fixes #3814.

- make the CLI/TUI setup path use the documented model precedence: explicit CLI model, saved conversation model, `[models].default`, then `MODEL`
- retain the resolved model's source as runtime-only `(kind, value)` metadata so model selection traces report the real source instead of labelling every pre-resolved value as `cli`
- guard the tracked source by value so a later explicit model change cannot inherit stale provenance
- cover all four precedence levels plus correct and stale trace metadata

The API v2 request model path is already explicit, and ACP's per-project `MODEL` override is an intentional per-workspace behavior, so neither is changed here.

## Validation

- `python -m pytest tests/test_config.py tests/test_init.py tests/test_model_attestation.py -q` — 208 passed, 1 skipped
- focused new regressions — 6 passed
- `ruff check` and `ruff format --check` on all five changed files — passed (Ruff 0.15.2)
- focused Mypy on the three changed source files with missing imports skipped — passed; the complete local Mypy invocation is blocked by unavailable optional dependencies and Windows-only POSIX API diagnostics

Full-suite attempts exposed pre-existing Windows harness limitations: without `PYTHONUTF8=1`, collection stops in `tests/test_reduce.py` on a GBK decode error; with UTF-8 mode, the run progressed into `tests/test_agent.py::test_create_template_mode_e2e` and stalled in its template subprocess. No full-suite pass is claimed.

### DIFF
--- a/gptme/config/cli_setup.py
+++ b/gptme/config/cli_setup.py
@@ -16,7 +16,13 @@ from ..tools._allowlist import (
     expand_tool_allowlist_presets,
 )
 from .chat import ChatConfig
-from .core import Config, get_config, set_config, set_config_from_workspace
+from .core import (
+    Config,
+    ModelSourceKind,
+    get_config,
+    set_config,
+    set_config_from_workspace,
+)
 
 if TYPE_CHECKING:
     from ..tools.base import ToolFormat
@@ -114,7 +120,8 @@ def setup_config_from_cli(
     """
     Initialize and return a complete config from CLI arguments and workspace.
 
-    Handles the precedence: CLI args -> saved conversation config -> env vars -> config files -> defaults
+    Handles the model precedence: CLI args -> saved conversation config ->
+    ``[models].default`` -> ``MODEL`` -> auto-detection.
     """
 
     # Load base config from workspace
@@ -127,18 +134,29 @@ def setup_config_from_cli(
         existing_chat_config = ChatConfig.from_logdir(logdir)
 
     # Resolve configuration values with proper precedence
-    # For resuming: CLI args -> saved conversation config -> env vars/config files
-    # For new conversations: CLI args -> env vars/config files -> defaults
     resolved_model: str | None
+    resolved_model_source: ModelSourceKind | None
     if model is not None:
         # CLI override always takes precedence
         resolved_model = model
+        resolved_model_source = "cli"
     elif existing_chat_config and existing_chat_config.model:
         # When resuming, use saved conversation model unless CLI override provided
         resolved_model = existing_chat_config.model
+        resolved_model_source = "chat_config"
+    elif config.user.models.default:
+        resolved_model = config.user.models.default
+        resolved_model_source = "models.default"
     else:
-        # Fall back to env/config for new conversations or when no saved model
+        # Fall back to the process/project/user MODEL layers, then auto-detection.
         resolved_model = config.get_env("MODEL")
+        resolved_model_source = "MODEL" if resolved_model is not None else None
+
+    config._model_source = (
+        (resolved_model_source, resolved_model)
+        if resolved_model_source is not None and resolved_model is not None
+        else None
+    )
 
     resolved_gear = parse_gear(gear)
     if (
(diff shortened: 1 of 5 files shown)


gptme/config/cli_setup.py
```diff
@@ -127,18 +134,29 @@ def setup_config_from_cli(
...
         # CLI override always takes precedence
         resolved_model = model
+        resolved_model_source = "cli"
     elif existing_chat_config and existing_chat_config.model:
         # When resuming, use saved conversation model unless CLI override provided
         resolved_model = existing_chat_config.model
+        resolved_model_source = "chat_config"
+    elif config.user.models.default:
+        resolved_model = config.user.models.default
+        resolved_model_source = "models.default"
```
ErikBjare, line 149: I'm not so sure a user default should have priority over `MODEL`? Not even sure `existing_chat_config` should (but probably?)

ErikBjare: Addressed in #3825.

Model layers are now ordered by specificity rather than by mechanism: a shell `MODEL`/`GPTME_MODEL`, and a project `gptme.toml [env].MODEL`, both outrank the global `[models].default`. `[models].default` keeps its documented role as the formal alternative to `[env].MODEL` in that *same* user config, which it still beats — so the same-scope rule from the docs survives while the more specific layers win.

On the second half: `existing_chat_config` stays above the env layers (your "but probably?"). Resuming a conversation keeps its model unless you pass `--model`.

Also folded in there: `setup_config_from_cli()` and `init_model()` each had their own copy of the precedence chain (and `_record_selection_trace()` a third) — that drift is what produced #3814 — so all three now share one resolver.

PR #3819, opened 2026-09-12
fix(subagent): deliver subprocess control and capture stderr

TimeToBuildBob:
## Problem

Subprocess subagents advertise `subagent_steer`, but their restricted toolset does not load the subagent ToolSpec, so its `STEP_PRE` control hook is never registered. Steering is written successfully and then stranded. Child stderr is also discarded, reducing crashes to an opaque exit code.

## Changes

- register the subagent control hook whenever `GPTME_SUBAGENT_AGENT_ID` identifies a managed child, independent of the child's toolset
- redirect child stderr to `logdir/stderr.log` without using a pipe
- append a byte- and line-bounded stderr tail to non-zero-exit and timeout results
- add real subprocess regressions using the offline mock provider for steer delivery and crash diagnostics

## Verification

- reproduced on the base branch: real `mock/echo` child exited with `STEER-WAS-DELIVERED` absent from its conversation and still in `prompt-queue.jsonl`
- `pytest tests/test_subagent_unit.py tests/test_tools_subagent.py -q`: 402 passed
- targeted real-child tests: 5 passed
- ruff and file-scoped pre-commit mypy passed
- all pre-commit hooks passed on each commit

`make typecheck` was also attempted, but Poetry created a fresh incomplete Python 3.12 environment and reported hundreds of missing installed dependencies. The repository's pre-commit mypy hook passed for all changed files.


TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `gptme/tools/subagent/execution.py:618` | P1 | superseded by latest review (not reproduced) |
| `gptme/hooks/__init__.py:262` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/subagent/execution.py:788` | P2 | superseded by latest review (not reproduced) |
| `gptme/tools/subagent/execution.py:875` | P2 | rejected |
| `tests/test_tools_subagent.py:807` | P2 | superseded by latest review (not reproduced) |
| `tests/test_tools_subagent.py:839` | P2 | rejected |

</details>




This PR registers a subagent control hook in subprocess children by detecting GPTME_SUBAGENT_AGENT_ID, redirects child stderr to logdir/stderr.log, and appends a bounded stderr tail to failure results. It adds tests for hook allowlist behavior, stderr tail bounding, and real subprocess steer delivery and crash diagnostics.

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **2 of 5 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

**⚠️ P2 medium** — `tests/test_tools_subagent.py:839`

The new test test_subprocess_control_hook_delivers_queued_steer uses `model="mock/echo"` and asserts the conversation contains STEER-WAS-DELIVERED. However, the test does not verify that the steer message was actually delivered via the control hook; it only checks that the string appears in the conversation. The mock/echo model echoes the prompt, and the steer message is queued in the prompt queue. If the control hook is not registered, the steer message would remain in the queue and not appear in the conversation. But the test also asserts `drain_steer_prompts(logdir) == []`, which would fail if the steer was not drained. So the test does verify delivery. However, the test is marked slow and uses a real subprocess, which may be flaky in CI. The test also sets PYTHONPATH to the repo root, which is necessary. The test is not vacuous. The test for crash stderr tail uses `model="mock/does-not-exist"` and asserts "Unknown mock model" in the result. This depends on the mock provider's error message. If the mock provider changes its error message, the test would fail. This is a test that pins an external behavior (the mock provider's error message), which is within the repo, so it's not a contract violation. The test is valid.

<sub>**How this was verified:** Read the test in tests/test_tools_subagent.py lines 806-840. It queues a steer prompt, runs a real subprocess with mock/echo, and checks the conversation contains the steer message. The test would fail if the hook were not registered, so it is not vacuous.</sub>

</details>

- `gptme/hooks/__init__.py` — Adds subagent_control hook registration and extends allowlist logic for managed subprocess children.
- `gptme/tools/subagent/execution.py` — Redirects subprocess stderr to stderr.log and adds _stderr_failure_tail to failure results.
- `tests/test_hooks.py` — Adds tests for explicit allowlist strictness and managed child allowlist extension.
- `tests/test_subagent_unit.py` — Adds tests for bounded stderr tail surfacing and single long line bounding.
- `tests/test_tools_subagent.py` — Adds real subprocess tests for steer delivery and crash stderr tail, and updates command construction test.

</details>

| `3891486a9351` | 4/5 | 1 | llm | 2026-09-12 18:50 UTC |
| `f977f846da68` | 3/5 | 4 | llm | 2026-09-12 19:20 UTC |
| `09ecb9bcc638` | 4/5 | 1 | llm | 2026-09-12 19:41 UTC |

</details>

<sub>Reviewed `09ecb9bcc638` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 55s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/hooks/__init__.py
```diff
@@ -243,6 +255,14 @@ def init_hooks(
         elif interactive and not no_confirm:
             hooks_to_register.append("cli_confirm")
 
+    # The control channel is part of the subprocess-subagent protocol, not an
+    # optional session hook. Keep it active even under an explicit hook allowlist.
+    if (
+        config.get_env("GPTME_SUBAGENT_AGENT_ID")
+        and "subagent_control" not in hooks_to_register
```
TimeToBuildBob, line 269: ⚠️ **P2** — The `init_hooks` function in `gptme/hooks/__init__.py` appends `subagent_control` to the `hooks_to_register` list based on the presence of the `GPTME_SUBAGENT_AGENT_ID` environment variable, regardless of whether an `allowlist` was provided. This behavior violates the contract of the `allowlist` parameter, which is typically used by callers to restrict execution to a specific subset of hooks (e.g., using `--hooks none`). If a user attempts to disable all hooks for performance or security reasons while running in a subagent context, this hook will remain active and continue to perform periodic disk I/O to check for steering messages in the log directory.

TimeToBuildBob:
@greptileai review

gptme/tools/subagent/execution.py
```diff
@@ -607,14 +610,18 @@ def _run_subagent_subprocess(
     env = os.environ.copy()
     env["GPTME_SUBAGENT_AGENT_ID"] = logdir.name.removeprefix("subagent-")
     env["GPTME_PROGRESS_FILE"] = str(progress_file)
+    stderr_path = logdir / _SUBPROCESS_STDERR_FILENAME
 
     try:
-        with open(tmpfile_path) as stdin_file:
+        with (
+            open(tmpfile_path) as stdin_file,
+            open(stderr_path, "w") as stderr_file,
```
TimeToBuildBob, line 618: ❌ **P1** — In _run_subagent_subprocess, the stderr file is opened with open(stderr_path, "w") and passed as stderr=stderr_file to subprocess.Popen. The file is opened in text mode (default), but subprocess.Popen with text=True expects a file object that is opened in text mode, which is fine. However, the file is opened with mode "w" which truncates any existing stderr.log. If a subagent is retried or the logdir is reused (e.g., subagent_reply re-spawns with the same logdir), the previous stderr content is lost. More importantly, the file is not closed after the subprocess exits; the with block closes it only after Popen returns, but Popen returns immediately after spawning the child, so the file is closed while the child is still running. This means the child's stderr writes go to a closed file descriptor, which can cause the child to receive SIGPIPE or fail to write stderr, and the stderr.log may be empty or incomplete. The correct pattern is to keep the file open until the process is reaped, or use a file that is not closed until after wait(). This is a real bug: the child's stderr is not captured reliably.

tests/test_tools_subagent.py
```diff
@@ -801,6 +803,81 @@ def test_subprocess_actual_process_creation():
                 process.wait()
 
 
+@pytest.mark.slow
+def test_subprocess_control_hook_delivers_queued_steer(monkeypatch, tmp_path):
```
TimeToBuildBob, line 807: ⚠️ **P2** — The test test_subprocess_control_hook_delivers_queued_steer uses model="mock/echo" and asserts that the conversation contains "STEER-WAS-DELIVERED". However, the test does not verify that the subagent_control hook is actually registered in the child. The hook registration depends on init_hooks being called with GPTME_SUBAGENT_AGENT_ID set, but the test only sets the environment variable in the parent before calling _run_subagent_subprocess. The child process inherits the environment, but whether init_hooks is called in the child and whether it registers the hook is not directly asserted. The test could pass even if the hook is not registered, as long as the steer message is delivered through some other mechanism (e.g., the prompt queue is drained by the chat loop). This is a test-defect: the test does not pin the specific behavior it claims to test.

gptme/tools/subagent/execution.py
```diff
@@ -778,14 +785,32 @@ def _drain() -> None:
     _drain()
 
 
+def _stderr_failure_tail(stderr_path: Path | None) -> str:
```
TimeToBuildBob, line 788: ⚠️ **P2** — In _stderr_failure_tail, the function reads the last _SUBPROCESS_STDERR_TAIL_BYTES bytes and then splits into lines, taking the last _SUBPROCESS_STDERR_TAIL_LINES. However, if the file is smaller than the byte limit, it reads the whole file, which is fine. But the function does not handle the case where the file is empty (size 0): it seeks to 0, reads 0 bytes, chunk is empty, returns "". That is fine. The issue is that the function is called with subagent.logdir / _SUBPROCESS_STDERR_FILENAME, but if the subprocess was never started (e.g., Popen failed), the stderr.log may not exist, and the function returns "" due to OSError. That is fine. However, the function is called in the timeout branch and the non-zero exit branch, but not in the success branch. If the process exits with code 0 but stderr contains warnings, they are not surfaced. That is a design choice, not a bug. The real issue is that the stderr file is opened in text mode and the child writes to it, but the file is closed immediately after Popen returns (as noted in another finding), so the tail may be empty. This finding is subsumed by the file-closing bug.

gptme/tools/subagent/execution.py
```diff
@@ -843,9 +868,11 @@ def _monitor_subprocess(
         # Process was killed because our timeout expired (not an external SIGKILL)
         status = "failure"
         result = f"Process killed after {subagent.timeout}s timeout"
+        result += _stderr_failure_tail(subagent.logdir / _SUBPROCESS_STDERR_FILENAME)
     else:
         status = "failure"
         result = f"Process exited with code {subagent.process.returncode}"
+        result += _stderr_failure_tail(subagent.logdir / _SUBPROCESS_STDERR_FILENAME)
```
TimeToBuildBob, line 875: ⚠️ **P2** — The stderr tail is appended to the failure result (lines 871 and 875) without guarding on `subagent.output_schema`. The preserved-branch text at line 883 is explicitly suppressed when `output_schema` is set, so an output_schema caller receives a clean result string on success. But on a crash or timeout, the result becomes "Process exited with code N" plus the raw stderr tail, which for an output_schema caller is a non-JSON string. This is inconsistent with the branch-text handling and leaks any diagnostics (potentially including stack traces or secrets) into the caller-facing result, breaking the implicit contract that output_schema results are JSON-parseable.

TimeToBuildBob:
@greptileai review

gptme/tools/subagent/execution.py
TimeToBuildBob: Rejected: `output_schema` is a success-result contract, not a requirement that failure diagnostics be valid JSON. Both `subagent_wait()` (`api.py:1495`) and `_parse_result()` (`batch.py:64`) explicitly parse only when `status == "success"`; failed subagents already return diagnostic strings such as `Process exited with code N`. The preserved-branch guard prevents metadata from corrupting a successful structured payload, while the stderr tail is intentionally part of a failure diagnostic—the feature this PR adds.

tests/test_tools_subagent.py
```diff
@@ -801,6 +803,88 @@ def test_subprocess_actual_process_creation():
...
+    process = _run_subagent_subprocess(
+        prompt="INITIAL-PROMPT",
+        logdir=logdir,
+        model="mock/echo",
+        workspace=tmp_path,
+    )
+    assert process.wait(timeout=30) == 0
+
+    conversation = (logdir / "conversation.jsonl").read_text()
+    assert "STEER-WAS-DELIVERED" in conversation
```
TimeToBuildBob, line 839: ⚠️ **P2** — The new test test_subprocess_control_hook_delivers_queued_steer uses `model="mock/echo"` and asserts the conversation contains STEER-WAS-DELIVERED. However, the test does not verify that the steer message was actually delivered via the control hook; it only checks that the string appears in the conversation. The mock/echo model echoes the prompt, and the steer message is queued in the prompt queue. If the control hook is not registered, the steer message would remain in the queue and not appear in the conversation. But the test also asserts `drain_steer_prompts(logdir) == []`, which would fail if the steer was not drained. So the test does verify delivery. However, the test is marked slow and uses a real subprocess, which may be flaky in CI. The test also sets PYTHONPATH to the repo root, which is necessary. The test is not vacuous. The test for crash stderr tail uses `model="mock/does-not-exist"` and asserts "Unknown mock model" in the result. This depends on the mock provider's error message. If the mock provider changes its error message, the test would fail. This is a test that pins an external behavior (the mock provider's error message), which is within the repo, so it's not a contract violation. The test is valid. _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Rejected: the finding's own verification concludes the test is valid and identifies no defect. This is intentionally a real offline subprocess regression because an in-process mock cannot prove child startup registers the hook. The test asserts both observable delivery and queue consumption, and `mock/echo` plus the bogus mock model are repository-owned deterministic fixtures—not external behavior.

PR #3824, opened 2026-09-13
fix(cli): emit clean error on status -o write failure, auto-mkdir parent

TimeToBuildBob:
## Problem

`gptme-util status -o /nonexistent/dir/foo.md` exits with a raw Python traceback (FileNotFoundError raised from inside click's invoke chain) instead of a clean CLI error.

```
  ...
  File ".../gptme/cli/cmd_status.py", line 555, in status
    out_path.write_text(doc, encoding="utf-8")
  ...
FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent/dir/foo.md'
```

Two related issues:
1. Missing parent directory → uncaught FileNotFoundError, raw traceback.
2. Parent path exists but is a regular file (e.g. `status -o /etc/hostname/x.md`) → uncaught FileExistsError, raw traceback.

Both should be a one-line `Error: ...` and a non-zero exit code.

## Fix

In `gptme/cli/cmd_status.py`, wrap the `out_path.write_text` call:

- `out_path.parent.mkdir(parents=True, exist_ok=True)` — friendlier than the current behavior (mkdir -p semantics). Most users want to write to a fresh path under a brand-new directory.
- `try/except OSError → click.ClickException` — surfaces the OS error message (`[Errno 17] File exists: '/etc/hostname'`) with the failing path, and exits non-zero with click's standard `Error: ...` prefix.

## Tests

Two new tests in `tests/test_cli_status.py`:

- `test_status_write_to_path_with_blocking_file_emits_clean_error` — sets up a regular file where a parent directory is expected; asserts the output is a clean `Error: ...`, the file is not created, and there is **no** `Traceback (most recent call last)` in the output.
- `test_status_write_creates_missing_parent_dirs` — confirms the new mkdir-p behavior: writing to `tmp/new/subdir/handoff.md` succeeds when the parents don't yet exist.

All 46 tests in `test_cli_status.py` pass.

## Reproduction

Before:

```
$ gptme-util status -o /no/such/dir/foo.md
  ...
FileNotFoundError: [Errno 2] No such file or directory: '/no/such/dir/foo.md'
```

After:

```
$ gptme-util status -o /no/such/dir/foo.md
Written to /no/such/dir/foo.md

$ gptme-util status -o /etc/hostname/x.md
Error: Failed to write status to /etc/hostname/x.md: [Errno 17] File exists: '/etc/hostname'
```

## Notes

Found while dogfooding `gptme-util` surfaces with bad input (per `dogfood-cli-bad-input` lesson). This is a fresh surface — `cmd_status` write paths were not in any previously-swept class.


TimeToBuildBob:
## 🤖 AI code review

Wraps the status output-file write path in a try/except that creates missing parent directories via `out_path.parent.mkdir(parents=True, exist_ok=True)` and converts OSError into a clean `click.ClickException`. Adds two tests, one for the blocking-file parent case and one for auto-creating missing parent directories.

Confidence 5/5

- `gptme/cli/cmd_status.py` — Adds a try/except around the status write; auto-creates missing parent dirs with mkdir and converts OSError to ClickException.
- `tests/test_cli_status.py` — Adds two tests: clean error when a parent is a regular file, and successful write when parents are missing.

</details>

| `1d1e63b35fa8` | 3/5 | 1 | llm | 2026-09-13 00:42 UTC |
| `1d1e63b35fa8` | 4/5 | 1 | llm | 2026-09-13 01:03 UTC |
| `309538951c0b` | 5/5 | 0 | llm | 2026-09-13 01:21 UTC |

</details>

<sub>Reviewed `309538951c0b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 428s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/cmd_status.py
```diff
@@ -552,7 +552,13 @@ def status(
         out_path = (root or Path.cwd()) / "status.md"
 
     if out_path:
-        out_path.write_text(doc, encoding="utf-8")
+        try:
+            out_path.parent.mkdir(parents=True, exist_ok=True)
+            out_path.write_text(doc, encoding="utf-8")
+        except OSError as e:
```
TimeToBuildBob, line 558: ❌ **P1** — The new except clause catches OSError and raises click.ClickException, but the `-o/--output` option is declared with `type=click.Path(writable=True)`. Click's Path type performs its own existence/writability checks during argument parsing, before the command body runs. For a path whose parent is a regular file (e.g. `/etc/hostname/x.md`), click.Path(writable=True) will fail during parsing with a UsageError, not reach the try block, so the new clean error path is never exercised for that case. The test `test_status_write_to_path_with_blocking_file_emits_clean_error` invokes `status` with `-o` pointing at a path under a regular file; click's Path type will reject it before the command runs, so the test passes for the wrong reason — it asserts a clean error, but that error comes from click's parameter validation, not from the new except clause. The PR's stated goal of converting the FileExistsError traceback into a ClickException is not achieved for the blocking-file case; only the missing-parent case (which click.Path(writable=True) allows because the parent doesn't exist) reaches the new code.

TimeToBuildBob: Rejected as a false positive after reproducing the exact case on the PR head with Click 8.4.2. `click.Path(writable=True)` has `exists=False`; when `os.stat()` raises for either `/tmp/.../blocker/foo.md` or a wholly missing parent, `Path.convert()` returns the value immediately and does not run `os.access()`. The command body therefore runs, `mkdir()` raises `FileExistsError` for the blocking-file case, and our `except OSError` emits exactly `Error: Failed to write status to ... [Errno 17] File exists: ...`. A standalone `CliRunner` probe confirmed both non-existent path shapes pass Click parsing, and all 46 status tests pass.

tests/test_cli_status.py
```diff
@@ -70,6 +70,36 @@ def test_status_write_to_file(tmp_path):
...
 
+def test_status_write_to_path_with_blocking_file_emits_clean_error(tmp_path):
+    """Writing to a path whose parent is a regular file must raise a clean
+    ClickException, not a raw Python traceback (regression for raw FileExistsError
+    traceback in `gptme-util status -o`)."""
+    runner = CliRunner()
+    blocker = tmp_path / "blocker"
+    blocker.write_text("regular file contents")
+    output_file = blocker / "foo.md"
+    result = runner.invoke(status, ["-o", str(output_file)])
```
TimeToBuildBob, line 81: ⚠️ **P2** — The new test `test_status_write_to_path_with_blocking_file_emits_clean_error` asserts `result.exit_code != 0` and that the output contains 'Error:' and the output path. However, it does not assert that the exit code is specifically 1 (ClickException) versus 2 (UsageError) or another code. More importantly, the test relies on the `-o` option being passed as a string to `status` directly, which means click's `Path(writable=True)` type conversion is not exercised. In the real CLI, `click.Path(writable=True)` will attempt to check writability of the path at parse time. For a path like `blocker/foo.md` where `blocker` is a regular file, click.Path(writable=True) may raise a UsageError before the command body runs, so the new try/except may never be reached. The test therefore does not validate the actual CLI behavior it claims to fix; it only validates the behavior when the type conversion is bypassed. This is a test-defect because it stubs the wrong boundary (bypasses click's parameter processing). _(2/3 passes — pass 2 looked and disagreed)_

TimeToBuildBob: Fixed in 309538951: the test now requires exit code 1 and matches the `Failed to write status to ...` ClickException prefix, so a Click UsageError (exit 2) cannot satisfy it. `CliRunner.invoke(status, ["-o", ...])` does perform Click option conversion; the stronger assertions now prove conversion continues into our command body. All 46 status tests pass.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Merged after the current head passed all 18 CI checks, Greptile 5/5, and the required 3/3 AI-review consensus. Verified merge commit: `7c713aee4c13`.

PR #3829, opened 2026-09-13
fix(init): don't warn when the optional model registry is absent

ErikBjare:
Found while debugging an unrelated PR, where this warning appeared in the output of every single test that initializes a model.

## Problem

`_record_selection_trace()` wraps the optional registry import and the lookup in one broad `try/except Exception`:

```python
try:
    from model_capability_registry import lookup_model
    ref = lookup_model(model_meta.model)
    ...
except Exception as e:
    logger.warning("registry lookup failed for %s: %s", model_meta.model, e)
```

`model-capability-registry` is **not a dependency** — it appears in `pyproject.toml` only as a mypy override (L300). So in any normal install the import always raises `ModuleNotFoundError`, and every session logs:

```
WARNING  gptme.init:init.py:366 registry lookup failed for claude-sonnet-4-6: No module named 'model_capability_registry'
```

Two problems with that:

1. **It warns about the expected case.** Attestation degrading to `selection_only` without the registry is by design, not a fault. A WARNING for something that is true of every default install is noise, and noise that appears in test output trains people to ignore warnings.
2. **It hides real faults.** Because the import and the lookup share one `except Exception`, an actual registry failure (bad record, version mismatch, raising `lookup_model`) is indistinguishable from "the package isn't installed" — same message, same level.

## Fix

Separate the two cases:

- `ImportError` → `logger.debug`. Expected, silent by default.
- Any other exception → `log_warn_once` (already used 5x in `llm/models/resolution.py`), so a genuine registry fault is still reported, but once per session rather than once per model resolution.

No behaviour change to attestation itself: `attestation_level` still degrades to `selection_only` exactly as before.

## Validation

- `tests/test_model_attestation.py` + `tests/test_init.py` + `tests/test_config.py`: **214 passed**.
- Warning occurrences in a model-initializing test: **1 → 0**.
- `ruff check` / `ruff format --check` clean (0.16.0, matching the lockfile).


ErikBjare:
@greptileai review

gptme/init.py
```diff
@@ -355,15 +356,24 @@ def _record_selection_trace(
...
-
-        ref = lookup_model(model_meta.model)
-        if ref is not None:
-            registry_record = ref.record_id
-            catalog_observed_at = ref.observed_at
-            if ref.verification_status == "verified":
-                attestation_level = "provider_claim"
-    except Exception as e:
-        logger.warning("registry lookup failed for %s: %s", model_meta.model, e)
+    except ImportError:
```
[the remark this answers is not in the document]
ErikBjare: Confirmed and fixed — this is the same conflation the PR set out to remove, just one level up. Splitting the import from the lookup separated "registry failed" from "registry absent", but then bare `ImportError` re-merged "absent" with "installed but broken" inside the import branch.

Now `ModuleNotFoundError` (genuinely not installed, the normal case) logs at debug, and any other `ImportError` — missing transitive dep, or a version that no longer exports `lookup_model` — goes through `log_warn_once`. Matches the existing precedent in `gptme/tools/__init__.py:112,132`, which already discriminates on `ModuleNotFoundError`.

gptme/init.py
```diff
@@ -355,15 +356,24 @@ def _record_selection_trace(
...
+            ref = lookup_model(model_meta.model)
+            if ref is not None:
+                registry_record = ref.record_id
+                catalog_observed_at = ref.observed_at
+                if ref.verification_status == "verified":
+                    attestation_level = "provider_claim"
+        except Exception as e:
+            # The registry is installed but failed: a real fault, worth saying
+            # once rather than once per model.
+            log_warn_once(f"registry lookup failed for {model_meta.model}: {e}")
```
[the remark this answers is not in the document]
ErikBjare: Correct, and my commit message overstated it — I wrote "once per session rather than once per model resolution" when the interpolated model and exception text meant it was once per unique message.

Fixed by keying the dedupe on a fixed string rather than the formatted message, so a session warns once on the first genuine lookup failure while the message still names the model and the error. I deliberately did not make it fully session-wide across *all* registry problems: the import failure and the lookup failure now use separate keys, since silencing one because the other already fired would reintroduce the same class of bug.

ErikBjare:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

The PR refactors the exception handling in `_record_selection_trace()` in `gptme/init.py` to distinguish between the expected absence of the optional `model_capability_registry` package and actual import or lookup failures. It uses `ModuleNotFoundError` with `e.name` checks and `ImportError` to decide whether to log at debug level or use `log_warn_once`, and moves the `lookup_model` call into a separate `try/except` block. The change also imports `log_warn_once` from `.llm.models`.

Confidence 5/5

**⚠️ P2 medium** — `gptme/init.py`

This is a `fix(...)` PR but no test files are included in the diff. Erik's feedback: 'where is the repro & fixes they are supposed to catch' (gptme#3441), 'that measurement should come with a regression test' (gptme#3446). Add a test that would have caught this bug. (Advisory: Erik merged all such PRs but consistently requested tests.)

```suggestion
Add a test file that reproduces the bug before the fix and passes after it.
```

<sub>**How this was verified:** static preflight: fix-commit + touched-files scan (rule 7)</sub>

</details>

- `gptme/init.py` — Adds `log_warn_once` import and restructures the registry import/lookup exception handling to separate benign missing-module cases from real failures.

</details>

| `c59070a9a8f3` | 3/5 | 1 | llm | 2026-09-13 21:01 UTC |
| `ebd2c13cbe31` | 4/5 | 1 | llm | 2026-09-13 21:40 UTC |

</details>

<sub>Reviewed `9aa60dc7330f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 341s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/init.py
```diff
@@ -355,15 +356,30 @@ def _record_selection_trace(
...
-                attestation_level = "provider_claim"
-    except Exception as e:
-        logger.warning("registry lookup failed for %s: %s", model_meta.model, e)
+    except ModuleNotFoundError:
+        # model-capability-registry is an optional external package and is not a
+        # dependency, so it is absent in a normal install. Attestation degrades
+        # to "selection_only" by design; that is not a problem worth warning
+        # about on every session.
+        logger.debug("model_capability_registry not installed, skipping lookup")
+    except ImportError as e:
```
TimeToBuildBob, line 371: ❌ **P1** — The new except ImportError branch is unreachable for the case it claims to handle. In Python, `from model_capability_registry import lookup_model` raises ModuleNotFoundError when the package is absent, and ImportError when the package exists but the name is missing. However, ModuleNotFoundError is a subclass of ImportError, and the except clauses are evaluated in order: the `except ModuleNotFoundError` catches both the absent-package case and the missing-attribute case. The comment says 'a version that no longer exports lookup_model' is a genuine fault, but that exact scenario raises ModuleNotFoundError (subclass of ImportError) and is silently downgraded to debug, so a real fault is hidden. The only way to reach the ImportError branch is if the import machinery raises a plain ImportError that is not a ModuleNotFoundError, which is rare. The PR's stated goal of distinguishing 'installed but broken' from 'not installed' is not achieved for the missing-export case.

ErikBjare: Real defect, fixed in `ebd2c13cb` — though the stated mechanism is inverted, and the difference matters for the fix.

I probed the three cases directly:

```
absent module                    -> ModuleNotFoundError  name='model_capability_registry'
present, export missing          -> ImportError          name='pkg_present'
present, broken transitive dep   -> ModuleNotFoundError  name='a_module_that_does_not_exist'
```

A missing *export* raises plain `ImportError`, not `ModuleNotFoundError` — `ModuleNotFoundError` is raised only when a *module* cannot be found. So "a version that no longer exports lookup_model" already reached the warning branch; that part was working.

The genuine hole is the third case: an installed-but-broken registry raises `ModuleNotFoundError` for the **inner** module, and my `except ModuleNotFoundError` swallowed it to debug as "not installed". My comment claimed that case warned. It did not — the same conflation of absent with broken that this PR exists to remove, one level down.

Fixed by discriminating on `e.name`: only our own module missing counts as absent; any other name, or no name, warns. Verified against the real code path:

```
absent (normal install)          -> DEBUG    via gptme.init
present, broken transitive dep   -> WARNING  via log_warn_once
present, export missing          -> WARNING  via log_warn_once
```

214 tests pass.

ErikBjare:
@greptileai review

ErikBjare:
@TimeToBuildBob review

gptme/init.py
```diff
@@ -385,6 +406,23 @@ def _record_selection_trace(
     set_selection_trace(trace)
 
 
+# Keyed on nothing but "has this fired", so a session reports the first genuine
+# registry lookup failure once and stays quiet, instead of re-warning for every
+# (model, error-text) pair. Deliberately not lock-guarded: the only cost of a
+# race is one duplicate warning, and this path is not known to run concurrently.
+_registry_lookup_warned = False
```
TimeToBuildBob: ⚠️ **P2** — The new `_registry_lookup_warned` flag is module-global and never reset, so once any registry lookup failure occurs in a process, all subsequent genuine failures are downgraded to debug and never reported again, even if the failure is transient and later lookups succeed. The comment says 'a session reports the first genuine registry lookup failure once and stays quiet', but the flag persists for the entire process lifetime, not just one session. In a long-running server process that calls `init()` multiple times (the code explicitly supports re-entry via `_init_done`), a single transient failure permanently silences all future warnings about registry problems, hiding real faults that the PR claims to keep visible. The consequence is that a user running a server that hits one bad lookup early will never see warnings for later, distinct registry failures. _(3/3 passes)_

ErikBjare: Correct, and fixed in `9aa60dc73` — this is the fourth variant of the same scope error in this PR (once-per-unique-message → once-per-session → once-per-process), each one caught by review rather than by me.

Rather than patch the flag I removed it. Lookup failures now dedupe on the **error signature** via the existing `log_warn_once`, with the model moved to a debug line:

- same fault across many models → one warning (the earlier concern: the model was in the dedupe key)
- a genuinely distinct fault later → still reported (your concern: a bare boolean silences everything after the first)

That deletes a never-reset global instead of adding a reset for it, and inherits conftest's existing `cleanup_logged_warnings` isolation.

Verified against the real code path:

```
same fault, 2 models   -> 1 WARNING, 2 DEBUG (model preserved on debug)
distinct fault after   -> 2nd WARNING
absent / broken dep / missing export -> DEBUG / WARNING / WARNING
```

Two details in the finding were slightly off, for the record: `_init_done` *prevents* re-entry (`init()` returns early and only refreshes `tool_format`), and per-request model selection in the server goes through `record_runtime_selection`, not `_record_selection_trace` — so the practical blast radius was narrower than a long-running server hitting it constantly. The defect and the principle were real regardless: the docstring's claim was false and a distinct later fault would have been hidden.

ErikBjare:
@greptileai review

ErikBjare:
@TimeToBuildBob review

TimeToBuildBob:
Reviewed `9aa60dc7330f` (merged as `7c85705`).

The import/lookup split is correct:

- `ModuleNotFoundError` with `e.name == "model_capability_registry"` → debug (expected absence; not a dependency)
- any other `ModuleNotFoundError` / `ImportError` → `log_warn_once` (broken install / missing export)
- lookup exception → `log_warn_once` on the fault, model on the debug line (same fault across models is one warning; a later distinct fault still fires)

Attestation still degrades to `selection_only` when the registry is absent. No remaining blocking issue.

Leftover is coverage only: the three cases you probed by hand are not pinned in `tests/`. Advisory, not a reopen — the PR already landed. No further action on this thread.

PR #3833, opened 2026-09-14
feat(skills): persist CostTracker ownership across TUI/server workers

TimeToBuildBob:
## Summary

Phase 2.1 follow-up to #3793. CLI skill invocations can already record measured cost windows. TUI never started a CostTracker, and server request workers each got a fresh ContextVar, so TUI/server terminals stayed `usage_available=false`.

This keeps one accounting window per conversation and re-binds it at command admission, generation, tool execution, and ACP steps. Skill usage remains the inclusive admission-to-terminal *delta* from that window, not a lifetime total.

## Changes

- `CostTracker.ensure_session` / `attach` plus a process-wide registry so workers can re-bind without resetting `tracking_id`
- `SESSION_START` is idempotent for the same logdir (a reset would make in-flight skill deltas unknown)
- TUI starts the window in `_chat_ctx` so command admission and generation share it
- Server binds from `LogManager.logdir` in `/api/v2` commands, `step()`, tool workers, and ACP
- `SessionCosts` snapshots/appends take a lock so concurrent workers cannot split totals

## Validation

- 146 focused tests originally, plus concurrent-record and TUI/server usage rebind cases (75 in the later slice including `test_skill_metrics` / `test_skill_events`)
- `ruff` + pre-commit hooks pass

## Deployed metric production (this session)

Installed gptme is `0.33.0+7c857058` and includes #3793, with `GPTME_TELEMETRY_ENABLED=true`. Production is still not exporting skill metrics:

- `opentelemetry.exporter` is missing from the uv-tool install (`gptme[telemetry]` extra not present)
- nothing listens on `localhost:4318`
- no `skill-events.jsonl` after the #3793 merge

Ledger usage on TUI/server no longer depends on that exporter path.

## Review

Local production-config review scored 3/5 with one P1. Rejected: it claims a reused `tracking_id` makes a later skill's usage include earlier costs. `_usage_delta` subtracts that invocation's `cost_baseline`; same `tracking_id` is the validity check, not a lifetime sum. Resetting on `SESSION_START` would mark in-flight invocations unknown.


gptme/util/cost_tracker.py
```diff
@@ -171,15 +189,41 @@ class CostTracker:
...
             session_id: Unique identifier for the session (typically logdir path).
         """
-        cls._session_costs_var.set(SessionCosts(session_id=session_id))
+        costs = SessionCosts(session_id=session_id)
+        with cls._sessions_lock:
+            cls._sessions[session_id] = costs
+        cls._session_costs_var.set(costs)
+
+    @classmethod
+    def ensure_session(cls, session_id: str) -> SessionCosts:
```
TimeToBuildBob, line 219: ❌ **P1** — The new process-wide registry _sessions is keyed by session_id, and session_id_for_logdir() resolves Path values. However, the hook session_start_cost_tracking now calls ensure_session() which never resets the window, while the old start_session() semantics (used by the CLI and other callers) reset the window on a new SESSION_START. The PR's own test test_start_session_resets_tracking_id confirms start_session() still resets. But the server and TUI paths now call ensure_session() unconditionally at the start of every step and every command, so if a conversation is resumed after a long gap, the cost window (and its tracking_id) persists across what should be a new accounting window. The skill event delta logic in _usage_delta compares tracker_id to the baseline; since ensure_session() never changes tracking_id, a resumed conversation's skill usage will be computed as a delta from the original baseline, potentially including costs from before the resume. This is a behavior change from the previous start_session() which reset the window on each SESSION_START. The PR description says 'bind tool workers to manager.logdir and lock shared cost entries' — the intent is to share the window across workers, but the side effect is that the window is never reset for a new session. Concretely: start a conversation, run a skill, then later resume the same conversation (new SESSION_START) and run another skill; the second skill's usage delta will include the first skill's costs because tracking_id is unchanged. This is a correctness bug in cost accounting. _(3/3 passes)_

TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

| `ac06a59f9d0a` | 3/5 | 1 | llm | 2026-09-14 01:53 UTC |
| `fc276e6f2216` | 3/5 | 2 | llm | 2026-09-14 02:24 UTC |
| `ebaa5bdc44ac` | 5/5 | 0 | llm | 2026-09-14 02:54 UTC |
| `defb3c08e27a` | 5/5 | 0 | llm | 2026-09-14 03:21 UTC |

</details>

<sub>Reviewed `defb3c08e27a` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 26s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
Rejecting the remaining local-review P1 on `ensure_session`.

Skill usage is the inclusive delta from that invocation's `cost_baseline`, not a lifetime total. `_usage_delta` subtracts the baseline and treats a *changed* `tracking_id` as unknown. Reusing the window across TUI/server workers is what makes the delta valid; calling `start_session` on every `SESSION_START` would reset `tracking_id` and mark in-flight invocations unknown.

A later skill on the same conversation takes a new baseline at admission, so it does not inherit the previous skill's tokens. `start_session` remains the explicit reset for evals and a new CLI process.

gptme/util/cost_tracker.py
```diff
@@ -171,15 +189,41 @@ class CostTracker:
     _session_costs_var: ContextVar[SessionCosts | None] = ContextVar(
         "session_costs", default=None
     )
+    _sessions: dict[str, SessionCosts] = {}
+    _sessions_lock = threading.Lock()
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in b2b6ae955 / fc276e6f2.

`CostTracker.end_session` drops the window from the process registry. Both delete paths call it: `api_conversation_delete` and `delete_conversation`. A later recreate of the same conversation id gets a new `tracking_id` and empty totals. `record()` also ignores a leftover ContextVar pointing at an ended window.

Not adding LRU for still-live conversations in this slice — that window is what TUI/server workers rebind. Delete is the production eviction path.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed two follow-ups for the review findings:

- `CostTracker.end_session` on conversation delete (server + CLI) so a recreate does not inherit stale totals. This was Greptile's P1.
- Relative logdir identity no longer `Path.resolve()`s against CWD, and `record()` ignores an ended window left in a ContextVar.

The local in-band reviewer still flags process-lifetime retention for conversations that are never deleted. That is accepted for this slice: the window is what TUI/server workers rebind, and delete is the production eviction path. LRU on live conversations would break that.

CI is running on fc276e6f2.

gptme/util/cost_tracker.py
```diff
@@ -171,15 +197,61 @@ class CostTracker:
     _session_costs_var: ContextVar[SessionCosts | None] = ContextVar(
         "session_costs", default=None
     )
+    _sessions: dict[str, SessionCosts] = {}
```
TimeToBuildBob, line 200: ❌ **P1** — The new process-wide registry keyed by session_id is never pruned except on explicit end_session (conversation delete). In the server, every conversation that is ever stepped gets a SessionCosts object inserted into CostTracker._sessions, and that object lives for the lifetime of the process. The registry is a plain dict with no eviction, so a long-running server that serves many distinct conversation IDs accumulates one SessionCosts (with its entries list) per conversation forever. Each SessionCosts holds a list of CostEntry objects that grows with every LLM request, so memory grows without bound proportional to total requests across all conversations ever touched. The PR adds end_session on delete, but conversations that are never deleted (the normal case) keep their windows resident. This is a real memory leak in a server deployment. _(2/3 passes)_

gptme/util/cost_tracker.py
```diff
@@ -191,7 +263,13 @@ def record(cls, entry: CostEntry) -> None:
...
         """
         costs = cls._session_costs_var.get()
-        if costs:
+        if not costs:
+            return
+        with cls._sessions_lock:
+            live = cls._sessions.get(costs.session_id) is costs
+        if not live:
+            return
+        with costs._lock:
```
TimeToBuildBob: ⚠️ **P2** — In CostTracker.record(), the liveness check reads cls._sessions.get(costs.session_id) is costs under _sessions_lock, but the append to costs.entries happens under costs._lock. end_session() pops the session from the registry under _sessions_lock, then checks the context var. There is a race: a thread that passed the liveness check (registry still contains the costs object) can be preempted before acquiring costs._lock; meanwhile end_session() pops the registry and clears the context var, then a new ensure_session() for the same id creates a fresh SessionCosts and installs it. The old thread then acquires costs._lock and appends to the stale, now-unregistered object. The entry is lost from the new window and the old object is garbage. The observable consequence is that a cost entry recorded concurrently with a conversation delete is silently dropped from the recreated conversation's totals, so the recreated conversation under-reports usage. The test test_record_ignores_ended_window_left_in_context covers the sequential case but not this interleaving. _(2/3 passes)_

gptme/util/cost_tracker.py
TimeToBuildBob: Accepted trade-off, not a leak of deleted state.

The process-wide registry is the shared accounting window TUI/server workers rebind. Evicting a still-live conversation (LRU, idle TTL) would reset tracking_id and mark in-flight skill deltas unknown — the failure this PR exists to prevent.

Production eviction is conversation delete: CostTracker.end_session is called from api_conversation_delete and delete_conversation. Recreate gets a fresh window. Live conversations keep their window for the life of the process; that is the intended lifetime of the cost session, same as LogManager state.

If server RSS from never-deleted conversations becomes a measured problem, SESSION_END drop after the last websocket is the follow-up — not LRU on live ids.

gptme/util/cost_tracker.py
TimeToBuildBob: Accepted. Conversation DELETE already returns 409 while generation is in progress, so a step/tool worker cannot be recording into a window that is being deleted.

The remaining interleaving is delete of an idle conversation concurrent with a stale ContextVar record(). record() already no-ops when the registry no longer owns that object; a dropped entry on a conversation that no longer exists is the correct outcome, not under-reporting on a recreate (recreate calls ensure_session and gets a fresh window).

gptme/util/cost_tracker.py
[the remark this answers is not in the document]
TimeToBuildBob: Follow-up in 611e25acb / ebaa5bdc4:

`end_session` now runs inside the conversation lock after `rmtree`, so a recreate cannot bind a window that delayed eviction then deletes. Last-session teardown also drops idle windows (`SessionManager._evict_idle_cost_window`), holding `_lock` across `end_session` so a newly connected session is not evicted.

Local in-band review on ebaa5bdc4 is 5/5.

TimeToBuildBob:
@greptileai review

gptme/server/session_models.py
```diff
@@ -428,12 +433,33 @@ def remove_session(cls, session_id: str) -> None:
...
+
+        # Hold _lock across end_session so create_session cannot insert a
+        # live session into a window we are about to drop. Lock order is
+        # SessionManager._lock then CostTracker._sessions_lock.
+        with cls._lock:
+            if conversation_id in cls._conversation_sessions:
+                return
+            if logdir is None:
+                logdir = get_logs_dir() / conversation_id
+            CostTracker.end_session(session_id_for_logdir(logdir))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in defb3c08e.

`_evict_idle_cost_window` now skips conversations in `_active_commands`. `finish_command` evicts only after that reservation is dropped and no session remains, so a disconnect during a slash command no longer discards in-flight `record()` calls.

Covered by `test_last_session_removal_keeps_cost_window_for_active_command` and `test_finish_command_with_live_session_keeps_cost_window`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Automated merge handoff for `defb3c08e27a`:

- CI gate: satisfied
- Review threads: clear
- `gptme/hooks/cost_awareness.py`
- `gptme/lessons/skill_events.py`
- `gptme/logmanager/conversations.py`
- `gptme/server/api_v2.py`
- `gptme/server/session_models.py`
- `gptme/server/session_step.py`
- `gptme/tui/app.py`
- `gptme/util/cost_tracker.py`

TimeToBuildBob:
## 📋 Reviewer's Brief

**What changed:** Introduces `CostTracker.ensure_session/attach` plus a process-wide registry and `session_id_for_logdir` across TUI (`_chat_ctx`), server `api_v2.py`, `session_step.py`, and `session_models.py` (eviction on delete/idle), with atomic snapshots under `_sessions_lock` in `gptme/util/cost_tracker.py`.

**Risk surface:** Concurrency between `_evict_idle_cost_window` and `ensure_session` — the lock re-check under `SessionManager._lock` may still race with a command admission or session creation that happens after the eviction check but before `end_session`, potentially dropping a re-bound window; deletion and re-creation races are partially mitigated but not proven under all interleavings.

**Recommended decision:** Merge — scope is tightly contained to cost-accounting persistence with no P0/P1 findings, and the locking order is clearly documented.

<sub>Auto-generated by Bob · HEAD `defb3c0` · [Report issue](https://github.com/ErikBjare/bob/issues)</sub>

PR #3834, opened 2026-09-14
feat(subagent): persist registry metadata and rehydrate after restart

TimeToBuildBob:
## Summary

Phase 1.2 of subagent resume+registry work. Follows gptme/gptme#3820 (`subagent_continue`).

After a parent gptme process restart, `subagent_status` / `subagent_wait` / `subagent_list` / `subagent_continue` can still find finished children. Each spawn writes `subagent-meta.json` into the child's logdir; lookup reloads it.

- Persist on register (thread/subprocess/ACP/planner executors) and after ACP session IDs are saved.
- Keep metadata after completion so continue-after-restart works. Remove it only when `subagent_reply` replaces the child.
- ID lookup matches both `subagent-{id}` (planner) and `subagent-{id}-{4 chars}` (executor).
- `subagent_list` does a one-shot scan; other APIs load a single id.
- Reused agent ids keep the newest in-memory / newest `started_at` disk copy.

## Accepted tradeoffs

- Rehydrated entries have no live `thread`/`process` handles. Thread-mode children die with the parent. Reattaching a subprocess that outlived the parent is out of scope for this slice; `status()` reads the conversation log.
- Non-dict `output_schema` (Pydantic class / type) is not JSON-serializable and is omitted on persist.

## Test plan

- [x] `uv run pytest tests/test_subagent_persistence.py tests/test_subagent_unit.py::TestSubagentContinue` (34 passed)
- [x] Simulated restart: persist, clear registry, `subagent_status` / `subagent_continue` against a random-suffix logdir
- [ ] CI


TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

| `a9af90f3eab1` | 4/5 | 2 | llm | 2026-09-14 02:55 UTC |
| `ef56dc2a5a3b` | 4/5 | 3 | llm | 2026-09-14 03:41 UTC |

</details>

<sub>Reviewed `ef56dc2a5a3b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 42s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/subagent/persistence.py
```diff
@@ -0,0 +1,172 @@
...
+        if entry != exact and not entry.name.startswith(prefix):
+            continue
+        if not (entry / "conversation.jsonl").exists():
+            continue
+        sa = load_subagent_meta(entry)
+        if sa is not None and sa.agent_id == agent_id:
+            candidates.append(sa)
+    if not candidates:
+        return None
+    return max(candidates, key=lambda sa: sa.started_at)
```
TimeToBuildBob, line 183: ⚠️ **P2** — Both load_subagent_by_id and _maybe_rehydrate_registry use sa.started_at as a sort key / max key without validating its type after JSON deserialization. A meta file containing "started_at": "abc" parses as valid JSON and passes load_subagent_meta (which catches JSONDecodeError, OSError, TypeError, KeyError, ValueError but not a TypeError raised later during sort/max). When scan_rehydrate_subagents or load_subagent_by_id then executes `max(candidates, key=lambda sa: sa.started_at)` (persistence.py line 147) or the sort in api.py line 84, comparing a string to floats raises TypeError, crashing subagent_list and subagent_status for that agent_id. The fix is to coerce started_at to float in _dict_to_subagent. _(2/3 passes)_

gptme/tools/subagent/persistence.py
```diff
@@ -0,0 +1,172 @@
...
+    result: dict[str, Any] = {}
+    for field in fields(sa):
+        if field.name in _RUNTIME_FIELDS:
+            continue
+        value = getattr(sa, field.name)
+        if value is None:
+            if field.name in _REQUIRED_FIELDS:
+                result[field.name] = None
+            continue
+        if field.name == "output_schema" and not isinstance(value, dict):
```
TimeToBuildBob, line 60: ⚠️ **P2** — persist_subagent_meta serializes output_schema only when it is already a dict; Pydantic models and type objects are skipped (the comment says they are not JSON-serializable). On reload, _dict_to_subagent therefore leaves output_schema as None. After a restart, subagent_continue on such a subagent passes output_schema=None to _create_subagent_thread, so the continued run no longer enforces the JSON schema (no complete-block validation, no canonical normalization in _read_log). This breaks the PR's claim that status/wait/continue work after restart for any subagent. A concrete input: spawn with output_schema=SomePydanticModel, let it finish, restart the parent, call subagent_continue — the schema is lost. Serializing the class's JSON schema (model_json_schema) instead would preserve it. _(2/3 passes)_

gptme/tools/subagent/persistence.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def persist_subagent_meta(sa: Subagent) -> None:
+    """Write ``subagent-meta.json`` into the subagent's logdir."""
+    try:
+        sa.logdir.mkdir(parents=True, exist_ok=True)
+        (sa.logdir / META_FILENAME).write_text(
+            json.dumps(_subagent_to_dict(sa), indent=2, default=str),
+            encoding="utf-8",
+        )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 8c68b8fa5.

`persist_subagent_meta` now writes to a same-directory temp file, fsyncs, and `os.replace`s onto `subagent-meta.json`. An interrupted rewrite no longer leaves truncated JSON, so a completed child stays discoverable after restart. Concurrent ACP updates can still race for last-writer-wins, but readers always see a complete file.

gptme/tools/subagent/persistence.py
TimeToBuildBob: Fixed in 8c68b8fa5.

`_dict_to_subagent` now coerces `started_at` to float. Invalid values fall back to `0.0` so they sort oldest and cannot crash newest-id selection. `iterdir()` is also wrapped in `OSError` so a disappearing logs dir returns empty instead of raising.

gptme/tools/subagent/persistence.py
TimeToBuildBob: Accepted tradeoff for this slice — non-dict `output_schema` values are omitted on persist, as documented in the PR body. Not changing here.

TimeToBuildBob:
@greptileai review

gptme/tools/subagent/persistence.py
```diff
@@ -0,0 +1,213 @@
...
+        logs_dir = get_logs_dir()
+    if not logs_dir.is_dir():
+        return []
+
+    rehydrated: list[Subagent] = []
+    try:
+        entries = list(logs_dir.iterdir())
+    except OSError as e:
+        logger.warning("Failed to scan logs dir %s: %s", logs_dir, e)
+        return []
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ef56dc2a5. A failed logs-dir listing no longer sets the one-shot rehydrate flag, so `subagent_list()` retries after a transient scan error instead of hiding persisted children for the rest of the process.

TimeToBuildBob:
@greptileai review

gptme/tools/subagent/api.py
```diff
@@ -38,6 +44,65 @@
...
+            return sa
+    loaded = load_subagent_by_id(agent_id)
+    if loaded is None:
+        return None
+    with _subagents_lock:
+        seq = list(reversed(_subagents)) if newest else list(_subagents)
+        sa = next((s for s in seq if s.agent_id == agent_id), None)
+        if sa is not None:
+            return sa
+        _subagents.append(loaded)
```
TimeToBuildBob, line 66: ⚠️ **P2** — In _find_subagent, when a persisted subagent is loaded from disk and appended to _subagents, the code does not check whether an in-memory entry with the same agent_id was added between the first lock release and the second lock acquisition. The second lock block re-scans _subagents and returns if a match exists, so a concurrent spawn of the same agent_id after the first scan but before the append would be found and returned, and the loaded disk copy is not appended. However, if the concurrent spawn happens after the second scan but before the append, the loaded disk copy is appended after the live entry, making the live entry not the newest in _subagents. Subsequent _find_subagent calls with newest=True iterate reversed(_subagents) and would return the stale disk copy instead of the live entry, breaking the documented preference for the newest in-memory entry. The window is narrow but real: subagent() appends under _subagents_lock and then calls persist_subagent_meta, so a concurrent subagent() for the same agent_id can interleave. The observable consequence is that status/wait/continue on a reused agent_id may operate on the stale rehydrated object (with no live thread/process) rather than the live one, reporting the old result or failing to wait on the running child.

gptme/tools/subagent/persistence.py
```diff
@@ -0,0 +1,215 @@
...
+    candidates: list[Subagent] = []
+    try:
+        entries = list(logs_dir.iterdir())
+    except OSError as e:
+        logger.warning("Failed to scan logs dir %s: %s", logs_dir, e)
+        return None
+    for entry in entries:
+        if not entry.is_dir():
+            continue
+        if entry != exact and not entry.name.startswith(prefix):
```
TimeToBuildBob, line 174: ⚠️ **P2** — In load_subagent_by_id, the prefix match uses entry.name.startswith(prefix) where prefix is f"subagent-{agent_id}-". This matches any directory whose name starts with that prefix, including a different agent_id that happens to have the searched id as a prefix. For example, searching for agent_id "worker" matches "subagent-worker-extra" and also "subagent-worker-extra2", but more importantly searching for "worker" would also match a directory named "subagent-worker-1" even if the meta file inside has agent_id "worker-1". The code then checks sa.agent_id == agent_id after loading the meta, so a mismatched meta is filtered out. However, the directory scan still reads the meta file for every prefix match, and if the meta file is corrupt or unreadable, load_subagent_meta returns None and the candidate is skipped. The real issue is that the prefix match is too broad: a directory named "subagent-worker-1" with a meta file whose agent_id is "worker-1" would be loaded and then filtered, but a directory named "subagent-worker-1" with a meta file whose agent_id is "worker" (e.g. from a previous run with a different suffix) would be incorrectly included as a candidate for "worker". This can cause the wrong logdir to be selected when multiple runs share a prefix, especially with the random 4-char suffix pattern where the suffix is alphanumeric and could be a prefix of another suffix. The observable consequence is that status/wait/continue may target a stale or wrong child logdir.

gptme/tools/subagent/api.py
TimeToBuildBob: False positive — the claimed race is already closed by the lock.

The suggested patch is the current code. `_find_subagent` already re-scans and appends in the same `_subagents_lock` block:

```python
with _subagents_lock:
    seq = list(reversed(_subagents)) if newest else list(_subagents)
    sa = next((s for s in seq if s.agent_id == agent_id), None)
    if sa is not None:
        return sa
    _subagents.append(loaded)
    return loaded
```

A concurrent `subagent()` also appends under `_subagents_lock`, so it cannot land between that scan and append. The "after the second scan but before the append" window does not exist. No code change.

gptme/tools/subagent/persistence.py
TimeToBuildBob: False positive — directory prefix is a scan hint; `sa.agent_id == agent_id` is the identity check.

A prefix-overlapping dir whose meta says a different id is loaded and then dropped. A dir whose meta says `agent_id == "worker"` *is* a worker candidate; including it is correct even if the folder name is odd. Executor logdirs are `subagent-{id}-{4 chars}`, so `subagent-worker-1` is planner naming for id `worker-1` and would not pass the meta filter for `worker`. Extra I/O on prefix overlaps is not a wrong-logdir bug. Not changing the matcher in this slice.

gptme/tools/subagent/persistence.py
TimeToBuildBob: Accepted tradeoff, as documented in the PR body.

Non-dict `output_schema` (Pydantic class / type object) is not JSON-serializable and is omitted on persist. Rehydration therefore leaves `output_schema=None`. Dict schemas already round-trip (`test_dict_output_schema_roundtrips`); `test_type_output_schema_omitted` locks the omit. Serializing `model_json_schema()` is a later slice, not this one. Status/wait/continue after restart still work; structured-output continue after restart does not, for non-dict schemas only.

PR #3835, opened 2026-09-14
fix(tools): raise ValueError with tool name when init() returns None

### LINKED ISSUE #3828: Plugin whose init() returns None crashes tool init with an unattributed AttributeError
Hit while debugging an unrelated PR; cost a long detour because the error names neither the plugin nor the tool.

## What happens

`_init_single_tool` in `gptme/tools/__init__.py` (~L160-165):

```python
def _init_single_tool(tool: ToolSpec) -> ToolSpec:
    if tool.init:
        tool = tool.init()
    tool.register_hooks()
    tool.register_commands()
    return tool
```

If a tool's `init()` returns `None` (easy to do — a plugin-provided `init` that forgets to return the spec), the next line raises:

```
AttributeError: 'NoneType' object has no attribute 'register_hooks'
gptme/tools/__init__.py:163: AttributeError
```

The traceback names no tool and no plugin, so with several plugins installed there is nothing to grep for. In my case it surfaced only as a mass of setup errors across an entire test file.

## Why it matters

`ToolSpec.init` is part of the plugin surface (`gptme/plugins/plugin.py`), so third-party plugins can trigger this. A contract violation in one plugin takes down `init_tools()` for everything, with an error that points at gptme internals rather than at the plugin.

## Suggested fix

Validate the return and attribute it:

```python
if tool.init:
    initialized = tool.init()
    if initialized is None:
        raise ValueError(
            f"Tool {tool.name!r} init() returned None; it must return a ToolSpec"
        )
    tool = initialized
```

Optionally isolate per-tool failures so one bad plugin degrades rather than aborts startup.

## Related observation

While here: unit tests log `Found <provider> API key, using <provider> provider` (`gptme/llm/__init__.py:1235`), i.e. real credential detection runs during tests. Not the same bug, but it means local test output depends on which keys the developer has configured.

TimeToBuildBob:
## Summary

- `_init_single_tool` in `gptme/tools/__init__.py` now validates that `tool.init()` returns a `ToolSpec`, not `None`
- When `init()` returns `None`, raises `ValueError` naming the tool so the error points at the plugin, not gptme internals
- Adds three regression tests covering the None case, the valid case, and the no-init passthrough

Fixes #3828


TimeToBuildBob:
## 🤖 AI code review

Adds a validation step to _init_single_tool in gptme/tools/__init__.py: after calling tool.init(), it checks that the result is a ToolSpec instance and raises ValueError naming the tool if not. Adds four tests to tests/test_tools_base.py covering a None return, a non-ToolSpec return, a valid ToolSpec return, and the no-init passthrough case.

Confidence 5/5

- `gptme/tools/__init__.py` — Adds an isinstance check on the result of tool.init() that raises ValueError with the tool name when the return value is not a ToolSpec.
- `tests/test_tools_base.py` — Adds TestInitSingleTool with four tests exercising None, non-ToolSpec, valid-ToolSpec, and no-init returns from _init_single_tool.

</details>

| `c7d669558e2b` | 3/5 | 1 | llm | 2026-09-14 08:19 UTC |

</details>

<sub>Reviewed `4e6f24c3f673` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 280s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/__init__.py
```diff
@@ -162,7 +162,12 @@ def _init_single_tool(tool: ToolSpec) -> ToolSpec:
     Caller is responsible for acquiring _tools_init_lock if needed.
     """
     if tool.init:
-        tool = tool.init()
+        initialized = tool.init()
+        if initialized is None:
```
TimeToBuildBob: ❌ **P1** — The new None check in _init_single_tool only validates that init() returns a non-None value, but does not verify that the returned value is actually a ToolSpec. If a plugin's init() returns a non-None, non-ToolSpec object (e.g., a string, a dict, or an unrelated class instance), the code proceeds to call tool.register_hooks() and tool.register_commands() on that object. If the object lacks those methods, it raises AttributeError; if it happens to have them, it may silently register hooks/commands from the wrong object and return it as the tool, corrupting the loaded tool set. The error message claims 'it must return a ToolSpec', but the check does not enforce that contract. A concrete input: init=lambda: 'not a spec' would pass the None check and then fail with AttributeError at register_hooks, producing a confusing error that does not name the tool or the real cause. The fix is to use isinstance(initialized, ToolSpec) instead of only checking for None. _(3/3 passes)_

TimeToBuildBob: Fixed in 4e6f24c3f: `init()` must now return a `ToolSpec` (`isinstance`), and the ValueError names the tool plus the actual type (`NoneType`, `str`, …). Added a regression test for a non-ToolSpec return.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI red is `tests/test_util_cli.py::test_models_list` in **Test without API keys or extras**: `gptme provider requires authentication` WARNING prefixed `models list --json`, so `json.loads` failed.

This PR does not touch that path (`gptme/tools/__init__.py` + `tests/test_tools_base.py` only). Same SHA's extras job passed; the previous commit's Test workflow on this branch also passed. Re-running the failed job (run 34822444701). Not merging until that job is green.

PR #3839, opened 2026-09-14
feat(site): prototype of new gptme.org site (design v3)

TimeToBuildBob:
## What

A self-contained **prototype** of the redesigned gptme.org open-source showcase, in a new `site/next/` directory. It implements the approved "design v3" (desktop and mobile boards, light and dark).

**Not deployed to gptme.org.** The current pandoc site (`make site`, `site/template.html`) and the `docs.yml` GitHub Pages deploy are untouched.

Pages:
- `/`: nav with a "by Superuser Labs" link, hero with the mascot in rings and the install command (with a copy button), a "See it work" terminal session, "Run it your way" (CLI, TUI, Web UI, desktop, Vim, server), a stats strip, Bob and "Open source or managed" cards, a Superuser Labs band, and a footer.
- `/readme/`: the repository `README.md` rendered at build time (GFM, GitHub-compatible heading anchors so the table of contents works, without the logo/badge header, relative media links rewritten to `https://gptme.org/media/`).

## Why

gptme.ai is the managed service. gptme.org should be a proper open-source showcase that complements it, not a pandoc render of the README. The design decisions behind v3:
- Light mode by default; dark only via `prefers-color-scheme`.
- Light accents are blue plus copper; dark accents are green plus amber.
- Inter plus IBM Plex Mono.
- Hero with the mascot inside rings.
- A terminal-native section below it.
- An explicit link back to Superuser Labs.
- Bob's denim/orange stays a separate identity.

### Stack: same as `webui/`
Vite 5, React 18, TypeScript and **Tailwind CSS 3.4**, with the same PostCSS setup and `@tailwindcss/typography`, matching `webui/`'s majors and config conventions. Tailwind is the part we're sure about. Using webui's stack means components and design tokens can be shared between the site and the web UI later, without a second toolchain.

Pages are **prerendered to static HTML** for SEO, not shipped as a client-side SPA:
1. `vite build` builds the client assets.
2. `vite build --ssr` builds the server renderer.
3. `scripts/prerender.mjs` renders `/` and `/readme/` with `react-dom/server`. It fails the build if the hero text is missing from `dist/index.html`.

There is no React hydration. The only client JS (~1.5 kB) is the copy button and the mobile menu toggle, both progressive enhancements.

## Design tokens

- `site/next/src/styles/tokens.css` holds the gptme design language as semantic CSS variables: `--color-bg/surface/border/ink/heading/ink-2/muted/accent/accent-2/on-accent/chip`, the terminal palette `--term-*`, the Superuser Labs band `--su-*`, fonts, radii and spacing. Light values are on `:root`. Dark values apply under `prefers-color-scheme: dark` (guarded by `:root:not([data-theme="light"])`) and again under `:root[data-theme="dark"]`. A header comment documents the brand family: Superuser Labs, gptme, and Bob (kept distinct).
- `site/next/tailwind.preset.ts` maps those variables into the Tailwind theme, so components use `bg-surface text-ink border-border bg-term-bg text-accent-2-text`, never hex values, and need no `dark:` variants. `webui/` or a docs theme can adopt the tokens later by importing `tokens.css` and adding the preset. `webui/` is not changed in this PR.

### Contrast (WCAG AA, 4.5:1 for body text)
- Copper `#c46a1c` on `#f7f9fa` is **3.67:1**, which fails for small text. I added a text-only token, `--color-accent-2-text: #a65512` (5.06:1 on bg, 5.34:1 on white). The original copper stays for large or decorative use.
- Muted `#6b7c85` on `#f7f9fa` is **4.10:1**, which also fails. I added `--color-muted-text: #5c6d76` (5.09:1 on bg, 4.81:1 on chip) for captions and labels.
- All dark-mode text pairs pass: muted 5.59:1, ink-2 9.02:1, green 10.89:1, amber 10.77:1, and dark text on the green button 10.95:1. The terminal palette passes in both themes (muted 6.51:1, tool label 8.50:1).

## Screenshots

Headless Chromium, full page, prerendered `dist/`. No horizontal scroll at 390px (`scrollWidth == 390`). Inter and IBM Plex Mono load from the self-hosted files, and there are no console errors.

| | Light | Dark |
|---|---|---|
| 1440×900 | desktop light | desktop dark |
| 390×844 | mobile light | mobile dark |

## How to run

```sh
cd site/next
npm ci
npm run dev      # dev server, pages rendered through the same SSR entry
npm run build    # typecheck + client build + SSR build + prerender -> dist/
npm run preview
```

Or `make site-next` from the repo root. Requires Node 22.12+.

## Details

- **Stats at build time:** stars and contributors come from `gptme/stats` `data/summary.json`. Stars are formatted as "4.4k". If the fetch fails, the build logs a warning and uses committed defaults, so it doesn't fail offline.
- **Fonts:** Inter and IBM Plex Mono are self-hosted via `@fontsource`, with no Google Fonts request.
- **Assets:** `media/logo.png` and `media/icon.svg` are copied from the repo at build time. The Superuser Labs logo is committed under `site/next/public/`.
- **Links:**
  - Docs point to `https://gptme.org/docs/`, because docs.gptme.org isn't live yet. All doc URLs are centralized in `src/lib/links.ts` for the move.
  - The nav "Timeline" link goes to the gptme project timeline.
  - "See Bob's timeline" points to `timetobuildbob.github.io/timeline/`, which goes live with TimeToBuildBob/TimeToBuildBob.github.io#83.
- **Accessibility:** semantic landmarks, a skip link, alt text, visible focus outlines, an aria-live status for copy, and a menu toggle with `aria-expanded` that closes on Escape.
- **Responsive:** breakpoints follow the design (`max-sm` ≤600px per the mobile board, `max-md` ≤900px, `max-lg` ≤1100px). On mobile the rings sit above the headline, the terminal session is shortened (as on the board), the cards are single column, the stats are 2×2, and the Superuser Labs band and footer stack.

## CI and preview

New `.github/workflows/site-next.yml`:
- **build** runs `npm ci && npm run build` on PRs and pushes touching `site/next/**`, `README.md`, or the logo/icon, and uploads `dist/`.
- **preview** runs only for same-repo PRs. It runs `wrangler pages deploy dist --project-name gptme-org-next --branch <head ref>` with the existing `CLOUDFLARE_API_TOKEN` / `CLOUDFLARE_ACCOUNT_ID` secrets. The `gptme-org-next` Pages project does not exist yet, so the job first tries `wrangler pages project create`. Both steps are `continue-on-error`: if the token can't create or deploy the project, the job emits a warning and the PR is not blocked. The preview URL, if any, is in the job summary.

**Preview status: deployed.** The first run created the `gptme-org-next` Pages project and deployed this branch:
- Branch alias: https://feat-site-next-prototype.gptme-org-next.pages.dev/ (also `/readme/`)

> Note: between the first deploy and `73c6f8901` this alias was stale — the preview workflow
> rewrote the branch to `feat-site-next-prototype`, which Cloudflare treated as a second branch,
> so the clean hostname stayed pinned to the first deployment and later deploys landed on
> `feat-site-next-prototype-lexy.…`. Fixed; both hostnames now serve the current build. The
> authoritative URL for any run is the "Branch alias" line in that run's job summary.
- This deployment: https://4f4755e0.gptme-org-next.pages.dev

The bare `gptme-org-next.pages.dev` returns 404, which is expected: no production branch deploys from this workflow.

## Not in scope

- Moving docs to docs.gptme.org and the `/docs/` redirects
- Cutting gptme.org over from GitHub Pages to Cloudflare Pages
- Replacing or deleting the current site
- Changing `webui/` to consume the shared tokens
- A recorded-session replay for the terminal card (it's static and labeled "Illustrative session")
- GitHub alert syntax (`> [!NOTE]`) in the README renders as a plain blockquote

## Deviations from the design

- Light-mode small text uses the darker AA-safe copper and muted variants (see Contrast). The difference is subtle, but the eyebrows and captions are slightly darker than the board.
- The contributor count is live from gptme/stats (50 today), not the board's placeholder 54.
- The footer has an extra "README" link so `/readme/` is reachable. The docs link is labeled "Docs" rather than "docs.gptme.org", because that host isn't live yet.
- Surface cards are links: CLI and TUI docs, chat.gptme.org, downloads, gptme.vim, and the server docs.


TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `.github/workflows/site-next.yml:103` | P2 | superseded by latest review (not reproduced) |
| `site/next/src/client.ts:36` | P2 | superseded by latest review (not reproduced) |
| `site/next/src/lib/readme-sanitize.ts:24` | P2 | superseded by latest review (not reproduced) |
| `site/next/src/lib/readme.ts:27` | P2 | superseded by latest review (not reproduced) |
| `site/next/src/lib/readme.ts:55` | P2 | accepted-tradeoff |
| `tests/test_util_cli_batch.py:25` | P2 | accepted-tradeoff |

</details>




This PR adds a new static prototype site in site/next/ for the redesigned gptme.org, built with Vite, React, and Tailwind, with prerendered HTML for two routes (home and README). It includes a new GitHub Actions workflow for building and previewing the site, a Makefile target, and a test helper change in tests/test_util_cli_batch.py to patch time.monotonic in an isolated way.

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **1 generated file excluded and NOT reviewed:** ``site/next/package-lock.json`` (+2641/-0 lines, 88KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

**⚠️ P2 medium** — `tests/test_util_cli_batch.py:25`

The `_patch_monotonic` helper patches `cmd_batch.time` with a `SimpleNamespace` whose `monotonic` lambda calls `next(seq)`. If `_run_one_prompt` (or any code under test) calls `time.monotonic()` more times than the number of values supplied, `next(seq)` raises `StopIteration`, which propagates as a RuntimeError and fails the test. The previous code had the same limitation, but the new helper is now shared by two tests and the docstring claims it gives an isolated clock; it does not guard against exhaustion. Concretely, `test_run_one_prompt_invokes_child_process` supplies exactly two values (10.0, 12.25) and the production `_run_one_prompt` calls `time.monotonic()` exactly twice (start and end), so it passes today. But any future change that adds a third monotonic call (e.g. a timeout check inside the loop) will make the test crash with StopIteration instead of failing on an assertion, which is a confusing false failure. The fix is to make the fake monotonic return the last value when the iterator is exhausted, or to use a real monotonic-based sequence. This is a test-defect because the helper is new in this PR and the crash mode is a defect in the test infrastructure.

```suggestion
fake_time = SimpleNamespace(monotonic=lambda: next(seq, seq[-1] if seq else 0.0))
```

<sub>**How this was verified:** Verified both tests provide exactly two values: test_run_one_prompt_invokes_child_process passes 10.0,12.25 and test_run_one_prompt_reports_timeout passes 1.0,4.4567. The production code calls time.monotonic twice per invocation.</sub>

</details>

<details><summary><b>17 advisory findings (summary-only, not scored)</b></summary>

**⚠️ P2 medium** — `site/next/src/lib/readme-sanitize.ts:28`

The README sanitizer's allowedSchemesAppliedToAttributes includes 'cite', but the allowedAttributes map does not allow 'cite' on any tag, and the transformTags only handle 'a' and 'img'. More importantly, the sanitizer's allowedSchemes list includes 'mailto' for all tags, but the SAFE_HREF regex in keepSafeHref only allows http, https, mailto, and #. However, the sanitize-html library's allowedSchemesByTag for img only allows http and https, which is fine. The real issue is that the rewriteUrl function in readme.ts rewrites protocol-relative URLs to https, but the sanitizer's keepSafeSrcset splits on commas and then splits on whitespace, which can break URLs that contain commas or spaces. For example, a srcset like 'https://example.com/a,b.png 1x' would be split into 'https://example.com/a' and 'b.png 1x', and the second part would be dropped because it doesn't start with https. This is a correctness bug in the srcset rewriting, but it's not a security issue because the sanitizer drops the unsafe part. The impact is that valid srcset URLs with commas in the filename would be broken. However, this is a prototype and the README likely doesn't have such images. I'll not report this as it's speculative.

<sub>**How this was verified:** Checked the srcset parsing logic in readme-sanitize.ts and readme.ts. The split on comma and whitespace is standard for srcset, but URLs with commas are rare. Not a definite bug.</sub>

**⚠️ P2 medium** — `.github/workflows/site-next.yml:39`

The GitHub Actions workflow uses `actions/checkout@v7` and `actions/setup-node@v7` without pinning to a commit SHA, while the credential-bearing preview job pins download-artifact and wrangler-action to SHAs. The build job does not have access to secrets, but the checkout and setup-node actions are still third-party code that could be compromised. However, the workflow's `permissions: contents: read` limits the token, and the build job doesn't use secrets. The preview job correctly pins actions. This is a minor supply-chain hardening issue, but not a vulnerability in the current setup. I'll not report it as it's a style preference.

<sub>**How this was verified:** Checked the workflow file. The build job uses unpinned actions but has no secrets. The preview job pins actions.</sub>

**⚠️ P2 medium** — `site/next/src/lib/readme.ts:28`

The `stripHeader` function in `src/lib/readme.ts` looks for the first `</p>` after the `<!-- Badges -->` comment and slices from there. If the README's badge block contains a `</p>` inside a nested tag (e.g., a `<p>` inside a `<div>`), it will cut off too early. However, the current README's badge block is a simple `<p>` with images, so it works. The fallback strips leading HTML-only lines, but if the README starts with a Markdown heading that is also an HTML tag (e.g., `# <span>Title</span>`), it would strip that line. This is a potential bug if the README changes. I'll not report it as it's a future risk.

<sub>**How this was verified:** Read the current README.md structure (uses a `<p align="center">` header block and `<!-- Badges -->` comment). The implementation assumes `</p>` follows; verified that the fallback only handles lines starting with `<` after whitespace, which would fail if content is on the same line.</sub>

**⚠️ P2 medium** — `site/next/src/lib/readme.ts:105`

The `sanitizeReadmeHtml` function is called twice on the README HTML: once in `renderReadme` and again in the `Readme` component's `dangerouslySetInnerHTML`. This double sanitization is redundant but not harmful. However, the second sanitization in the component runs at render time, which is fine for static prerendering. The real issue is that the `renderReadme` function rewrites URLs before sanitizing, and the sanitizer's `transformTags` for `a` and `img` re-validates the href/src. But the `rewriteHtmlUrls` function in readme.ts uses a regex to rewrite `src`, `href`, and `srcset` attributes in the raw HTML. This regex can be fooled by attribute values that contain quotes or angle brackets, but the sanitizer will clean them up. I don't see a concrete vulnerability.

<sub>**How this was verified:** Traced the flow: renderReadme sanitizes, then Readme component sanitizes again. No issue.</sub>

**⚠️ P2 medium** — `site/next/src/client.ts:27`

The `client.ts` copy button handler uses `navigator.clipboard.writeText` which requires a secure context (HTTPS or localhost). On the preview deployment, it will be HTTPS, so it works. But if someone serves the static files over HTTP, the copy button will be hidden because `navigator.clipboard` is undefined. That's the intended progressive enhancement. No issue.

<sub>**How this was verified:** Checked the client code. It handles the absence of clipboard API.</sub>

**⚠️ P2 medium** — `site/next/src/routes/Home.tsx:208`

The `Home.tsx` component has a `Terminal` figure that shows a fake session. The text includes `gptme "fix the 500 on empty names` with a closing quote after the `wide` span. In the JSX, the string is split across lines and the `wide` span is inside the quoted string. This renders correctly. No issue.

<sub>**How this was verified:** Checked the JSX. It's valid.</sub>

**⚠️ P2 medium** — `site/next/src/lib/stats.ts:35`

The `stats.ts` module uses `AbortSignal.timeout(8000)` which is available in Node 17.3+ and modern browsers. The build runs on Node 22, so it's fine. The `fetch` is global in Node 18+. No issue.

<sub>**How this was verified:** Checked the fetch call. It's compatible with Node 22.</sub>

**⚠️ P2 medium** — `site/next/scripts/copy-assets.mjs:13`

The `copy-assets.mjs` script copies `media/logo.png` and `media/icon.svg` to `public/media/`. If the source files don't exist, it will throw an error. The workflow triggers on changes to those files, so they exist. No issue.

<sub>**How this was verified:** Checked the script. It's straightforward.</sub>

**⚠️ P2 medium** — `site/next/src/lib/readme.test.ts:3`

The `readme.test.ts` tests use `node --experimental-strip-types` which is a Node 22 feature. The package.json engines require Node &gt;=22.12, so it's fine. The tests import from `./readme.ts` with the .ts extension, which is allowed by `allowImportingTsExtensions`. No issue.

<sub>**How this was verified:** Checked the test imports. They work with the tsconfig.</sub>

**⚠️ P2 medium** — `site/next/tailwind.config.ts:13`

The `tailwind.config.ts` defines custom breakpoints: sm=601, md=901, lg=1101, xl=1441. The default Tailwind breakpoints are sm=640, md=768, lg=1024, xl=1280. This means that `max-sm` applies below 601px, `max-md` below 901px, etc. The components use `max-sm`, `max-md`, `max-lg` variants. This is consistent. However, the `max-lg` variant is not a default Tailwind variant; it's generated by the `max-*` modifiers in Tailwind 3.4. The custom screens are used to generate these. This works. No issue.

<sub>**How this was verified:** Checked the breakpoints. They are used correctly.</sub>

**⚠️ P2 medium** — `site/next/src/styles/tokens.css:131`

The `tokens.css` defines `--color-muted-text` as #5c6d76 in light mode and #7d8e97 in dark mode. The dark mode value is the same as `--color-muted`. The comment says it's AA-safe, but #7d8e97 on #0b1117 has a contrast ratio of about 4.6:1, which passes. No issue.

<sub>**How this was verified:** Checked the contrast. It's fine.</sub>

**⚠️ P2 medium** — `site/next/src/components/Layout.tsx:177`

The `Layout.tsx` component has a `DocsIndex` section that is rendered on every page, including the README page. This is a design choice. The README page also has its own header. No issue.

<sub>**How this was verified:** Checked the layout. It's consistent.</sub>

**⚠️ P2 medium** — `site/next/vite.config.ts:31`

The `vite.config.ts` dev middleware does not set a `Content-Type` for the response? It does set it to `text/html; charset=utf-8`. No issue.

<sub>**How this was verified:** Checked the header. It's correct.</sub>

**⚠️ P2 medium** — `site/next/src/routes/Home.tsx:453`

The `Home.tsx` component has a `Models` section that lists model commands. These are illustrative. No issue.

<sub>**How this was verified:** Checked the section. It's static content.</sub>

**⚠️ P2 medium** — `site/next/src/lib/readme.test.ts:72`

The `readme.test.ts` test for `stripHeader` uses a string with a badge block. The function looks for `<!-- Badges -->` and then the next `</p>`. In the test, the badge block is `<p>badges</p>`, so it works. No issue.

<sub>**How this was verified:** Checked the test. It passes.</sub>

**⚠️ P2 medium** — `site/next/scripts/prerender.mjs:36`

The `prerender.mjs` script deletes `dist-ssr/` after prerendering. If the build fails before that, the dist-ssr remains, but that's fine. No issue.

<sub>**How this was verified:** Checked the cleanup. It's correct.</sub>

**⚠️ P2 medium** — `site/next/src/routes/Readme.tsx:19`

The `sanitizeReadmeHtml` function is called twice on the README HTML: once in `renderReadme` and again in the `Readme` component's `dangerouslySetInnerHTML`. The first call in `renderReadme` sanitizes the HTML, and then the `Readme` component sanitizes it again. This double sanitization is redundant but not harmful. However, the second sanitization in `Readme.tsx` uses the same `sanitizeReadmeHtml` function, which is imported from `readme-sanitize.ts`. The issue is that the `Readme` component receives `html` as a prop from `entry-server.tsx`, which already calls `renderReadme()` that returns sanitized HTML. So the second sanitization is unnecessary but not a bug. The real issue is that the `Readme` component's `dangerouslySetInnerHTML` is a potential XSS sink, but the HTML is sanitized before it gets there. The double sanitization is a minor inefficiency, not a defect.

<sub>**How this was verified:** Checked the renderReadme function in readme.ts and the Readme component. Both call sanitizeReadmeHtml.</sub>

</details>

<details><summary><b>Files changed (28)</b> — the diff as I read it</summary>

- `.github/workflows/site-next.yml` — Adds a workflow that builds site/next on push/PR and deploys a Cloudflare Pages preview for same-repo PRs.
- `Makefile` — Adds a site-next target that runs npm ci and npm run build in site/next.
- `site/next/.gitignore` — Ignores node_modules, dist, dist-ssr, and public/media.
- `site/next/README.md` — Documents the prototype site, its stack, build process, and design tokens.
- `site/next/index.html` — Adds the HTML template with markers for head and body content and a JS-enabling script.
- `site/next/package.json` — Defines dependencies and scripts for building, testing, and dev server.
- `site/next/postcss.config.js` — Adds PostCSS config with Tailwind and autoprefixer.
- `site/next/scripts/copy-assets.mjs` — Copies media/logo.png and media/icon.svg from repo root to public/media.
- `site/next/scripts/prerender.mjs` — Prerenders routes to static HTML and checks for hero content.
- `site/next/src/client.ts` — Adds client-side enhancements: mobile menu toggle and copy-to-clipboard buttons.
- `site/next/src/components/Layout.tsx` — Adds layout components: header, docs index, Superuser band, footer, and head tags.
- `site/next/src/components/icons.tsx` — Adds inline SVG icons for star, menu, copy, and check.
- `site/next/src/entry-server.tsx` — Defines routes and server-side render function for the two pages.
- `site/next/src/lib/cn.ts` — Adds class name joiner and shared layout constants.
- `site/next/src/lib/links.ts` — Centralizes outbound links used across the site.
- `site/next/src/lib/readme-sanitize.ts` — Adds HTML sanitizer for README content with allowed tags and URL scheme checks.
- `site/next/src/lib/readme.test.ts` — Adds tests for README rendering, URL rewriting, and sanitization.
- `site/next/src/lib/readme.ts` — Renders README.md to HTML with heading anchors, URL rewriting, and sanitization.
- `site/next/src/lib/stats.ts` — Fetches build-time stats with fallback defaults and caching.
- `site/next/src/routes/Home.tsx` — Adds the landing page with hero, terminal demo, tool grid, and other sections.
- `site/next/src/routes/Readme.tsx` — Adds the README page that renders sanitized HTML.
- `site/next/src/styles/index.css` — Adds global styles, fonts, and base layer.
- `site/next/src/styles/tokens.css` — Defines design tokens as CSS custom properties for light and dark themes.
- `site/next/tailwind.config.ts` — Configures Tailwind with custom breakpoints and js/no-js variants.
- `site/next/tailwind.preset.ts` — Maps design tokens to Tailwind theme colors, fonts, and typography.
- `site/next/tsconfig.json` — Adds TypeScript configuration for the site.
- `site/next/vite.config.ts` — Adds Vite config with a dev SSR middleware.
- `tests/test_util_cli_batch.py` — Adds a helper to patch cmd_batch.time.monotonic and updates two tests to use it.

</details>

| `23c9b4bee7e2` | 4/5 | 2 | llm | 2026-09-14 11:41 UTC |
| `e440ea818b14` | 4/5 | 2 | llm | 2026-09-15 17:20 UTC |
| `b851d35c3df2` | 4/5 | 1 | llm | 2026-09-15 18:44 UTC |
| `b2555b5e1b86` | 4/5 | 1 | llm | 2026-09-15 19:22 UTC |

</details>

<sub>Reviewed `73c6f8901b5c` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 468s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

.github/workflows/site-next.yml
```diff
@@ -0,0 +1,119 @@
...
+        fi
+
+    - name: Deploy preview
+      id: deploy
+      working-directory: .
+      continue-on-error: true
+      run: |
+        set -o pipefail
+        npx --yes "$WRANGLER" pages deploy dist --project-name "$PROJECT" --branch "$HEAD_REF" --commit-dirty=true | tee deploy.log
+        url=$(grep -oE 'https://[a-zA-Z0-9.-]+\.pages\.dev' deploy.log | tail -1)
```
TimeToBuildBob, line 103: ⚠️ **P2** — The workflow's preview job uses `continue-on-error: true` on the deploy step, then the Summary step checks `steps.deploy.outcome`. However, the deploy step's `run` block uses `set -o pipefail` and pipes `npx ... | tee deploy.log`. If the deploy fails, the pipeline fails, but because `continue-on-error: true` is set, the step's outcome is 'failure' and the Summary step correctly falls into the else branch. The issue is that the `url` extraction uses `grep -oE 'https://[a-zA-Z0-9.-]+\.pages\.dev' deploy.log | tail -1`. If the deploy output contains multiple URLs (e.g., a previous deployment URL in the log), `tail -1` may pick the wrong one. More importantly, if the deploy succeeds but the URL is not in the log (e.g., wrangler output format changes), `url` is empty and the Summary step prints an empty deployment link. This is a best-effort preview, so the impact is a broken preview link in the PR summary, not a build failure. The fix is to parse the URL more robustly or fail the step if the URL is missing.

site/next/src/client.ts
```diff
@@ -0,0 +1,40 @@
...
+  btn.addEventListener("click", async () => {
+    try {
+      await navigator.clipboard.writeText(btn.dataset.copy ?? "");
+      btn.dataset.state = "copied";
+      if (status) status.textContent = "Copied to clipboard";
+      setTimeout(() => {
+        delete btn.dataset.state;
+        if (status) status.textContent = "";
+      }, 1800);
+    } catch {
```
TimeToBuildBob, line 36: ⚠️ **P2** — In `src/client.ts`, the copy-to-clipboard handler sets `btn.dataset.state = "copied"` and then uses `setTimeout` to delete it after 1800ms. If the user clicks the button multiple times quickly, multiple timeouts are scheduled, and the state may be cleared prematurely or the status text may be overwritten. More importantly, the `status` element is a single `#copy-status` span shared by all copy buttons. If there are multiple copy buttons (currently only one), clicking one and then another within 1800ms would cause the status text to be cleared by the first timeout while the second is still showing. This is a minor UI bug. Also, the `catch` block sets the status to 'Copy failed' but does not clear it after a timeout, so the error message persists indefinitely. The success path clears after 1800ms, but the failure path does not. This is a real inconsistency.

site/next/src/lib/readme.ts
```diff
@@ -0,0 +1,175 @@
...
+ * Drop the centered HTML header at the top of the README (logo, title,
+ * pronunciation, nav links, badges). The site already has its own header, and
+ * the badge wall is noise here. Everything from the first Markdown content
+ * after the badges block is kept.
+ */
+export function stripHeader(md: string): string {
+  const badges = md.indexOf("<!-- Badges -->");
+  if (badges !== -1) {
+    const end = md.indexOf("</p>", badges);
+    if (end !== -1) return md.slice(end + "</p>".length).trimStart();
```
TimeToBuildBob, line 29: ⚠️ **P2** — The `stripHeader` function searches for the first `</p>` after the `<!-- Badges -->` comment. If the README's badge block contains a `</p>` inside a nested tag (e.g. a `<p>` containing a `<br>` or a `<span>`), the slice will cut at the first `</p>`, which may be inside the badge block, leaving partial HTML. More concretely, the README's header likely has multiple `<p>` blocks; the first `</p>` after the badges comment might close a paragraph that is not the last one, so the returned markdown starts with the remainder of the badge block, including unclosed tags. The fallback only triggers if `</p>` is not found. The test only covers a simple case. This could cause the rendered README to include stray HTML or lose content. However, the actual README structure is not shown, so this is speculative. _(3/3 passes)_

site/next/src/routes/Readme.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <h1 className="m-0 font-mono text-[44px] font-semibold leading-[1.1] tracking-display text-heading max-sm:text-[34px]">
+          gptme
+        </h1>
+        <p className="m-0 text-muted-text">
+          Rendered from the repository README at build time. <a href={`${links.github}#readme`}>View on GitHub</a>
+        </p>
+      </header>
+      <article
+        className="prose prose-lg max-w-[860px] pb-22 [overflow-wrap:anywhere] max-md:pb-16 max-sm:prose-base"
+        dangerouslySetInnerHTML={{ __html: html }}
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 3d39f00. `/readme/` HTML now goes through `sanitize-html` before `dangerouslySetInnerHTML`, and `rewriteUrl` drops `javascript:`/`data:` and other non-http(s)/mailto schemes. Tests in `site/next/src/lib/readme.test.ts`.

.github/workflows/site-next.yml
```diff
@@ -0,0 +1,119 @@
...
+    runs-on: ubuntu-latest
+    env:
+      CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+      CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+      PROJECT: gptme-org-next
+      WRANGLER: wrangler@4.131.2
+      HEAD_REF: ${{ github.head_ref }}
+    steps:
+    - name: Download dist artifact
+      uses: actions/download-artifact@v6
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e440ea818. Preview deploy now uses `cloudflare/wrangler-action@ebbaa158` (same pin as `webui.yml`). Artifact download is pinned to a commit SHA. Cloudflare tokens are action inputs, not job env, and `npx wrangler` is gone.

site/next/src/client.ts
TimeToBuildBob: Fixed in 3d39f00: one shared timer, cleared on re-click, and the failure path also clears after 1800ms.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `b851d35c3` for the remaining Greptile 3/5 gap.

`Readme.tsx` now runs `sanitizeReadmeHtml` at the `dangerouslySetInnerHTML` sink (the line the P1 pointed at). Producer-only sanitization in `renderReadme()` was not visible there. Unsafe `href`/`src`/`srcset` schemes are dropped at that boundary. Preview `--branch` is a Cloudflare-safe token, not raw `github.head_ref`.

.github/workflows/site-next.yml
TimeToBuildBob: Obsolete after e440ea818 / b851d35c3. Preview URL comes from wrangler-action outputs (`deployment-url`, `pages-deployment-alias-url`), not grep of deploy.log.

site/next/src/lib/readme-sanitize.ts
```diff
@@ -0,0 +1,125 @@
...
+function keepSafeSrc(src: string | undefined): string | undefined {
+  if (!src) return undefined;
+  const trimmed = src.trim();
+  return SAFE_SRC.test(trimmed) ? trimmed : undefined;
+}
+
+function keepSafeSrcset(value: string | undefined): string | undefined {
+  if (!value) return undefined;
+  const kept = value
+    .split(",")
```
TimeToBuildBob, line 24: ⚠️ **P2** — In src/lib/readme.ts, the heading renderer generates the id from slugger.slug(text) where text is the raw heading text, but the README's own table of contents links use GitHub's anchor generation. For headings containing inline code or HTML entities, GitHub's slugger (github-slugger) is the same library, so it should match. However, the renderer passes the raw text to slugger, not the parsed inline tokens. For a heading like "Install `gptme`", slugger.slug("Install `gptme`") returns "install-gptme" (it strips backticks), and GitHub also produces "install-gptme". That matches. But for a heading with a link, e.g. "See [docs](...)", the raw text is "See docs" (marked's token.text includes the link text), and slugger gives "see-docs". GitHub also gives "see-docs". So no mismatch. The real issue is that the heading renderer does not handle duplicate headings the same way GitHub does? github-slugger does handle duplicates by appending -1, -2. That is fine. Another issue: the heading renderer uses escapeAttr on the slug, but slugger output is already safe. Not a bug. Let me look at the sanitizer. The sanitizer's allowedAttributes for a includes "aria-hidden" and "tabindex", which are used by the permalink. But the transformTags for a deletes href if unsafe. The permalink href="#id" is safe. Fine. The real bug: in readme-sanitize.ts, keepSafeSrcset splits on commas, but a data: URL in srcset like "data:image/svg+xml,<svg> 1x" would be split on the comma inside the data URL, producing parts "data:image/svg+xml" and "<svg> 1x". The first part would be checked by keepSafeSrc, which returns undefined for data:, so it is dropped. The second part "<svg> 1x" would be split into url "<svg>" and rest "1x", and keepSafeSrc("<svg>") returns undefined, so dropped. So the whole srcset is dropped. That is safe. But what about a srcset with a comma in a URL? URLs can contain commas, e.g. https://example.com/a,b.png 1x. Splitting on comma would break it into "https://example.com/a" and "b.png 1x". The first part is a valid https URL, kept as "https://example.com/a". The second part "b.png 1x" would be treated as a URL "b.png" with rest "1x", and keepSafeSrc("b.png") returns undefined because it doesn't start with https?:, so dropped. The resulting srcset would be "https://example.com/a", which is a broken URL. This is a correctness bug: valid srcset URLs containing commas are mangled. The README could contain such an image. This is a P2 bug. The test only covers javascript and https URLs without commas. I'll report it.

TimeToBuildBob:
`Test without API keys or extras` failed on `b851d35` at `tests/test_util_cli_batch.py::test_run_one_prompt_invokes_child_process` (`RuntimeError: generator raised StopIteration`).

The test patched `time.monotonic` with a two-value iterator on the shared `time` module. pytest-xdist's multiprocessing logger also calls `time.monotonic()`, exhausted the iterator, and the duration measurement blew up. Unrelated to the site prototype.

Fix is in `b2555b5`: isolate `cmd_batch.time` so the batch clock is a local sequence, not the global monotonic. Extras job is rerunning on that commit.

site/next/src/lib/readme.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (trimmed.startsWith("//")) return rewriteUrl(`https:${trimmed}`);
+  if (/^[a-z][a-z0-9+.-]*:/i.test(trimmed)) {
+    return SAFE_ABSOLUTE.test(trimmed) ? trimmed : "";
+  }
+  const clean = trimmed.replace(/^\.\//, "").replace(/^\//, "");
+  if (clean.startsWith("media/")) return MEDIA_BASE + clean.slice("media/".length);
+  return REPO_BLOB_BASE + clean;
+}
+
+function rewriteSrcset(value: string): string {
```
TimeToBuildBob, line 55: ⚠️ **P2** — The README sanitizer's allowedSchemesAppliedToAttributes includes 'cite', but the allowedAttributes list does not include 'cite' for any tag, so this is harmless. However, the transformTags for 'a' and 'img' only run after sanitize-html's own attribute filtering. The keepSafeHref function allows '#' fragments, but sanitize-html's default allowedSchemesAppliedToAttributes with allowProtocolRelative:false will strip hrefs that are not http/https/mailto. The transformTags for 'a' runs after the default attribute filtering, so a href like 'javascript:alert(1)' is already removed by sanitize-html before keepSafeHref sees it. The test confirms this. The real issue: the sanitizer allows 'class' attributes on many tags, and the README content is inserted via dangerouslySetInnerHTML. A class attribute itself is not a security risk, but the CSS is controlled by the site. No XSS here. Another issue: the rewriteHtmlUrls function in readme.ts runs before sanitizeReadmeHtml, and it rewrites srcset URLs. The rewriteSrcset function splits on commas, but URLs in srcset can contain commas (e.g., data URLs, but those are dropped). For normal URLs, commas are rare. The rewriteUrl function treats any string with a scheme as absolute and returns it if it matches SAFE_ABSOLUTE (http/https/mailto). So a srcset entry like 'https://example.com/a,b.png 1x' would be split into 'https://example.com/a' and 'b.png 1x', breaking the URL. This is a correctness bug for srcset with commas in URLs, but such URLs are uncommon. The sanitizer's keepSafeSrcset also splits on commas, so even if rewriteHtmlUrls didn't break it, the sanitizer would. This is a minor bug, but the PR's own test uses a srcset with a comma-separated list of URLs, not a URL containing a comma. The impact: images with srcset URLs containing commas will be broken. This is a P2 bug.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
WebUI `e2e` failed on `b2555b5` at `e2e/conversation.spec.ts` › Connecting › should connect and list conversations: after clicking "Introduction to gptme", `Hello! I'm gptme, your AI programming assistant` was not visible within 5s (3 retries). Same dev pass: hard-requirement connect test passed, 39 passed / 1 failed.

This PR does not touch `webui/`. WebUI e2e was green on this branch at 17:23Z and 18:34Z. Master WebUI e2e is green. The only commit since the last green e2e is the batch-clock isolation in `tests/test_util_cli_batch.py`.

Reran the failed e2e job on this SHA (run 35011451178). If it fails the same way, treat it as a real assertion rather than a one-off — do not keep rerunning.

site/next/src/lib/readme.ts
TimeToBuildBob: Not blocking. Comma-splitting srcset is a known simplification vs the WHATWG parser (4.8.4.3.10: collect the URL until ASCII whitespace, then strip a trailing comma). So `https://example.com/a,b.png 1x` would survive in a browser and get split here.

Checkable reasons this stays:

1. README at `b2555b5e1b86` has zero `srcset` attributes (`rg -i srcset README.md` is empty).
2. `javascript:` / `data:` / `vbscript:` are dropped by `rewriteUrl` / `keepSafeSrc` before they can ride a comma; this is a broken-image case, not XSS.
3. A full descriptor-tokenizer parser is out of scope for a prototype `/readme/` page.

Accepted trade-off. Revisit if the README grows real srcset markup.

TimeToBuildBob:
## Merge recommendation (Greptile cap adjudication)

Hit the Greptile attempt cap; did **not** re-trigger. Freshness: head `b2555b5e1b86` matches the newest `bob-ai-review` marker. Greptile last reviewed `23c9b4bee7e2` (opening commit). All 7 review threads are now resolved.

### Fixed
- Greptile P1 XSS on `/readme/`: sanitizer at the `dangerouslySetInnerHTML` sink (`3d39f00` / `b851d35c3`). Verified on this head: `Readme.tsx` calls `sanitizeReadmeHtml(html)`; `javascript:` / `data:` / event handlers stripped in tests.
- Greptile P2 mutable actions: preview job pins `download-artifact@018cc2cf` and `wrangler-action@ebbaa158`; tokens are action inputs, not job env (`e440ea818`).
- Copy-button timer + failure-path clear (`3d39f00`).
- Preview URL from wrangler-action outputs, not `grep` of deploy.log (`e440ea818` / `b851d35c3`).
- Batch-clock isolation so pytest-xdist cannot exhaust `time.monotonic` (`b2555b5e1b86`). WebUI e2e rerun on that SHA is green.

### Remaining
- Our P2 `cb380657a8a5` (`site/next/src/lib/readme.ts:55`, srcset comma-in-URL): **accepted-tradeoff**. WHATWG srcset parsing collects the URL until whitespace, so `https://example.com/a,b.png 1x` would survive in a browser and get split here. README at this SHA has **zero** `srcset` attributes. Unsafe schemes are dropped before they can ride a comma. Full descriptor-tokenizer parser is out of scope for a prototype `/readme/` page.
- Greptile score is still 3/5 because it never re-reviewed past `23c9b4bee7e2`. Score-driving findings are the two above, both fixed in later commits and now resolved as outdated.

### Verified against
`b2555b5e1b86c238f580815586d7ab93db92f9bd`

### CI
All required checks green on this head, including extras (`test_util_cli_batch`) and the WebUI e2e rerun.

### Domain risk
Maintainer should glance at the Cloudflare preview (`/readme/` especially) before cutting anything over to gptme.org. This PR does not deploy to gptme.org.

### Convergence
`round_convergence.stable_rounds=0` (status `new_blocking` in the detector because Greptile never posted a second round). After resolving the stale threads: `verdict=converged`, unresolved P1/P2 = 0.

Stop at maintainer judgment. Not auto-merging.

tests/test_util_cli_batch.py
```diff
@@ -12,6 +14,19 @@
...
 
+def _patch_monotonic(monkeypatch, *values: float) -> None:
+    """Give cmd_batch an isolated clock without patching the global module.
+
+    Patching ``time.monotonic`` globally exhausts a short iterator and crashes
+    pytest-xdist's multiprocessing logger (``timeout = deadline - time.monotonic()``).
+    """
+    real_monotonic = time.monotonic
+    seq = iter(values)
+    fake_time = SimpleNamespace(monotonic=lambda: next(seq))
```
TimeToBuildBob, line 25: ⚠️ **P2** — The `_patch_monotonic` helper patches `cmd_batch.time` with a `SimpleNamespace` whose `monotonic` lambda calls `next(seq)`. If `_run_one_prompt` (or any code under test) calls `time.monotonic()` more times than the number of values supplied, `next(seq)` raises `StopIteration`, which propagates as a RuntimeError and fails the test. The previous code had the same limitation, but the new helper is now shared by two tests and the docstring claims it gives an isolated clock; it does not guard against exhaustion. Concretely, `test_run_one_prompt_invokes_child_process` supplies exactly two values (10.0, 12.25) and the production `_run_one_prompt` calls `time.monotonic()` exactly twice (start and end), so it passes today. But any future change that adds a third monotonic call (e.g. a timeout check inside the loop) will make the test crash with StopIteration instead of failing on an assertion, which is a confusing false failure. The fix is to make the fake monotonic return the last value when the iterator is exhausted, or to use a real monotonic-based sequence. This is a test-defect because the helper is new in this PR and the crash mode is a defect in the test infrastructure.

TimeToBuildBob: Not blocking. Finite sequence + `StopIteration` is the intended fail-loud clock, not a defect in the helper.

Checkable on `73c6f8901b5c`:

1. `_run_one_prompt` calls `time.monotonic()` twice per path: `start =` then exactly one `duration_s = time.monotonic() - start` in the success, `TimeoutExpired`, and `ValueError` branches. Never three in one invocation.
2. Both tests supply exactly two values: `(10.0, 12.25)` and `(1.0, 4.4567)`. CI extras job is green on this head.
3. Last-value fallback would hide an extra call and can make duration asserts pass incorrectly. The suggested `next(seq, seq[-1] if seq else 0.0)` is also wrong: `seq` is an iterator, so `seq[-1]` raises `TypeError`.
4. The helper's documented job is isolation from global `time.monotonic` (pytest-xdist logger), not an unbounded clock.

Accepted trade-off. If production adds a third call, the test should fail loudly so the sequence is extended on purpose.

TimeToBuildBob:
## Merge recommendation (Greptile cap adjudication, follow-up)

Hit the Greptile attempt cap; did **not** re-trigger. Freshness: head `73c6f8901b5c` matches the newest `bob-ai-review` marker (`sha: 73c6f8901b5c`). Greptile last reviewed `23c9b4bee7e2` (opening commit) and will not see later fixes.

This is a follow-up to the 2026-09-15 rec on `b2555b5e1b86`. Three site commits landed after that (`ecb8837667d7`, `cae9331471cd`, `73c6f8901b5c`). Our reviewer re-ran on the new head and opened one P2; no unresolved threads remain.

### Fixed
- None in this session (no code change). Prior Greptile P1 XSS and P2 mutable-action pins remain in place on this head: `Readme.tsx` still calls `sanitizeReadmeHtml` at the `dangerouslySetInnerHTML` sink; preview job still pins `download-artifact@018cc2cf` and `wrangler-action@ebbaa158`.

### Remaining
- Our P2 `25330dfdc363` (`tests/test_util_cli_batch.py:25`, `_patch_monotonic` `StopIteration` on exhaustion): **accepted-tradeoff**. `_run_one_prompt` calls `time.monotonic()` twice per path; both tests supply two values; extras CI is green. Last-value fallback would hide extra calls; the suggested `seq[-1]` is invalid on an iterator. Finite sequence is fail-loud by design.
- Our P2 `cb380657a8a5` (srcset comma-in-URL): **accepted-tradeoff** (unchanged). README at this SHA still has zero `srcset` attributes.
- Greptile score is still 3/5 because it never re-reviewed past `23c9b4bee7e2`. Score-driving findings (README XSS, mutable actions) were fixed in later commits and those threads are resolved as outdated.

### Verified against
`73c6f8901b5c3d7e111d9469e2ff7fe9ba8bb7fe`

### CI
All required checks green on this head (tests, extras, e2e, lint, typecheck, preview). `deploy` jobs skipped (expected).

### Domain risk
Maintainer should glance at the Cloudflare preview (`/` and `/readme/`) before cutting anything over to gptme.org. This PR still does not deploy to gptme.org. New since last rec: copy/structure of the landing page (agents, ecosystem, self-hosting) plus a preview-branch slash so the alias URL stays stable.

### Convergence
`round_convergence.stable_rounds=0` (status `new_blocking` in the detector because Greptile never posted a second round). After this disposal: our unresolved findings = 0; Greptile unresolved P1/P2 = 0; `verdict=converged`.

Stop at maintainer judgment. Not auto-merging.

ErikBjare:
@greptileai review

.github/workflows/site-next.yml
```diff
@@ -0,0 +1,126 @@
...
+    # two different branches, so the second one got a suffixed alias
+    # (`feat-x-lexy.…pages.dev`) while the clean hostname stayed pinned to the
+    # first deploy — a preview URL that silently stops updating.
+    - name: Sanitize preview branch name
+      id: branch
+      working-directory: .
+      env:
+        RAW: ${{ github.head_ref }}
+      run: |
+        safe=$(printf '%s' "$RAW" | tr -c 'A-Za-z0-9._/-' '-' | sed 's/--*/-/g; s/^[-/]*//; s/[-/]*$//')
```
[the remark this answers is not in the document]
TimeToBuildBob: Not blocking. `s/--*/-/g` is cleanup after `tr` maps illegal chars to `-`, not another `/`→`-` keying bug.

Checkable on `73c6f8901b5c`:

1. This branch is `feat/site-next-prototype` — no consecutive dashes. The stale-preview bug was collapsing `/` before wrangler; that is already fixed (slash kept, Cloudflare still collapses it for the hostname).
2. Repo convention is `feat/x` / `fix/x`. `feat--x` vs `feat-x` would take two concurrent PRs named that way. We do not name branches with `--`.
3. Dropping the collapse would keep `feat@@x` as `feat--x` after `tr`, which is noisier and does not buy anything this PR hits.

Accepted trade-off. Revisit if we start using `--` in branch names.

TimeToBuildBob:
Greptile re-review is 5/5 (safe to merge). Remaining P2 is preview-branch `--` collapse (`feat--x` vs `feat-x`); accepted — we don't name branches that way, and the `/` vs `-` stale-alias bug is already fixed in `73c6f8901`.

Not merging. Still wants a glance at the preview (`/` and `/readme/`).

ErikBjare:
@TimeToBuildBob Merged, file a task to actually get it deployed

TimeToBuildBob:
Filed the deploy task: it wires a GitHub Pages deploy for `site/next/` (vite build + SSR prerender), verifies the live design, then cuts over the old pandoc `docs.yml` path so there's a single deploy. Tracked locally as `gptme-org-site-next-deploy` (todo, high priority) against this PR; will open the deploy PR against this repo.

TimeToBuildBob:
Deploy PR is up: #3849

- `make site` overlays `site/next/` onto Sphinx docs / media / downloads and deploys through the existing `docs.yml` GitHub Pages path (no second publisher).
- CI green on `be1bbefea11f`. Self-merge is blocked on workflow/Makefile paths — needs a maintainer merge.
- Live gptme.org is still the pandoc homepage until that lands.

Local tracking: `gptme-org-site-next-deploy` (waiting on #3849 merge, then live URL verification).

PR #3840, opened 2026-09-15
fix(webui): guard console-proxy injection against cross-origin iframes

TimeToBuildBob:
## Problem

In the chat browser preview, `handleIframeLoad` reads `iframe.contentWindow.document`. For any cross-origin URL that throws `DOMException: SecurityError`, and the error lands in gptme.ai error tracking. Reported by PostHog self-driving as gptme/gptme-cloud#957 (4 occurrences, 1 user).

## Fix

Wrap the injection in `try/catch`. Same-origin previews keep console capture. Cross-origin previews skip it silently, since injecting there was never possible anyway.

The duplicate-capture half of gptme/gptme-cloud#957 is handled in gptme/gptme-cloud#960.


TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

<sub>ℹ️ Consensus was degraded on this run: 2 of 3 passes answered, so findings were filtered at 2-of-2 agreement rather than 2-of-3 — less filtered than usual; 2 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

**⚠️ P2 medium · 🔒 security** — `webui/src/utils/consoleProxy.ts:69`

The `injectConsoleProxy` function catches `SecurityError` only if `err instanceof DOMException`. In some browsers, cross-origin access errors may be thrown as `DOMException` with name `SecurityError`, but in other environments (e.g., certain WebKit or jsdom versions) the error might be a plain `Error` with name `SecurityError` or a `TypeError` with a message about cross-origin access. The test uses `new DOMException(...)` which is the standard, but the real-world behavior can vary. If the error is not an instance of `DOMException`, the catch block rethrows it, causing the `handleIframeLoad` to throw and potentially break the iframe load event or produce an uncaught error in the console. This defeats the purpose of the PR, which is to swallow cross-origin SecurityError. The fix is to check `err instanceof DOMException && err.name === 'SecurityError'` OR `(err as any)?.name === 'SecurityError'` to be more robust. However, this is a defensive improvement; the current code may work in most modern browsers. The severity is P2 because it's an edge case that may not manifest in the target environment, but it's a real robustness issue.

```suggestion
if (err instanceof DOMException && err.name === 'SecurityError' || (err as Error)?.name === 'SecurityError') {
```

<sub>**How this was verified:** Checked the catch block in consoleProxy.ts and the test file. The test only covers DOMException. Cross-origin access in browsers can throw different error types; for example, accessing a cross-origin window's document in Chrome throws a DOMException, but in some cases (e.g., accessing contentWindow.document on a sandboxed iframe without allow-same-origin) it may throw a SecurityError as a DOMException, but other properties might throw TypeError. The code rethrows any non-DOMException, which could break the iframe load handler.</sub>

</details>

| `c432e5c5c363` | 5/5 | 0 | llm | 2026-09-15 16:22 UTC |
| `c432e5c5c363` | 5/5 | 0 | llm | 2026-09-15 16:52 UTC |

</details>

<sub>Reviewed `67e227ae992f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 311s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Merged by Erik (`424bd1df`). Pre-merge CI green. gptme/gptme-cloud#957 is already closed (PostHog report closed). Duplicate-capture half landed in gptme/gptme-cloud#960.

gptme-cloud still pins gptme at `1541e48` (3 commits behind; this guard is not live on gptme.ai until the next submodule bump). Auto-bump is scheduled 06:00 UTC. No further action on this PR.

PR #3843, opened 2026-09-15
fix(shell): wake idle sessions on background completion

TimeToBuildBob:
## Summary

A noninteractive session could start a background shell job, say it was running, and exit before receiving its result. The completion hook ran only at turn end, after auto-reply/stuck detection, and never waited for a still-running job.

- Deliver completed output at `STEP_PRE`, including server/ACP steps.
- Wake an idle noninteractive CLI through a conversation-specific condition before auto-reply hooks run. Waiting is bounded by `GPTME_WATCH_IDLE_MAX` (default 1800 seconds), yields to prompts/control/subagent notifications, and respects subagent cancellation and wall-clock budgets. A timeout reports once; late output remains available.
- Wake waiters on teardown and discard callbacks from removed jobs.
- Let CLI SIGTERM run existing cleanup, preventing detached background processes from surviving termination.

This extends the conversation-owned job registry shipped in gptme/gptme#3802. Server idle auto-wake and a general multi-source watch tool remain separate follow-ups.

## Validation

- Real chat-loop regression: baseline exits after two model replies and loses completion; fixed loop consumes final output exactly once in the third reply, with no assistant polling/wait call. Foreign conversation output remains isolated.
- `make test` over shell, complete, and hook-registry suites: **1,112 passed**.
- Final completion/control/CLI-smoke/signal/prompt-queue regressions: **40 passed** with strict thread-leak checking. The real `gptme -n` test verifies successful exit and the completion-aware answer in the saved transcript, mocking only provider generation. Timeout/cancel checks preserve terminal status without overwriting an earlier result. Idle steering uses the existing locked queue readers; malformed records cannot terminate the wait, and excess regular/steering prompts remain durable beyond the loop’s queue limit.
- CLI suite plus SIGTERM subprocess regressions: **116 passed**, 18 API/slow/eval cases deselected. SIGTERM first reproduced a surviving real child on baseline; all four signal scenarios now pass.
- Scoped mypy and all pre-commit hooks pass. Full `make typecheck` reports the same **69 baseline errors in 12 files**, with no new errors (verified against the unmodified base using the same environment).

## Review status

Production-config review of predecessor `e2b63a14d` returned 4/5 with no P0/P1 findings (two responding passes). Independent adversarial review found a queue-capacity P2; `f7979575b` fixes it, with three red-to-green tests and an actual-loop replay verifying all 101 prompts arrive in order.

The final-head production review received empty responses from Together in all three passes and stalled in retries for over ten minutes. That attempt was stopped as an infrastructure failure. **Final-head production review remains pending**; no stale verdict is published as current evidence. The independent reviewer confirmed the final fix.


TimeToBuildBob:
## 🤖 AI code review

Confidence 5/5

| `f7979575b208` | 3/5 | 3 | llm | 2026-09-15 19:22 UTC |
| `ebb0bcd2d6a1` | 3/5 | 3 | llm | 2026-09-15 20:22 UTC |

</details>

<sub>Reviewed `9cb572a0532b` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 648s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell_background.py
```diff
@@ -434,6 +454,165 @@ def background_job_completion_hook(
...
+    # STEP_PRE to consume steering unless we actually queue that input here.
+    # Invalid/partial records remain on disk and cannot cause a spurious exit.
+    messages = drain_prompt_queue(logdir, max_items=MAX_PROMPT_QUEUE_SIZE)
+    messages += drain_steer_prompts(
+        logdir, max_items=MAX_PROMPT_QUEUE_SIZE - len(messages)
+    )
+    if messages:
+        return messages
+    try:
+        if (logdir / "control.jsonl").stat().st_size:
```
TimeToBuildBob: ❌ **P1** — The background_job_wait_hook calls _background_wait_input, which checks control.jsonl for any non-zero size and returns a system Message ('Background wait yielded to pending session input.') but never consumes or removes the file. Because the hook runs at every loop.continue checkpoint while a background job is pending, a stale non-empty control file causes the same synthetic message to be yielded every turn. The loop feeds that message to the model, the model replies, and the next loop.continue again sees the same non-empty file and yields the same message again — an endless cycle of model calls until the file is manually removed or the timeout fires. The test test_control_file_reenters_step_checkpoint explicitly asserts the file still exists after the hook runs, confirming the hook does not consume it. In a non-interactive session with a pending job and any non-empty control file, the session never progresses to job completion and burns provider calls indefinitely.

gptme/tools/shell_background.py
```diff
@@ -434,6 +454,165 @@ def background_job_completion_hook(
...
+                messages = list(background_job_completion_hook(manager))
+                if messages or any(job._completion_notified for job in pending):
+                    break
+                if any(
+                    _background_jobs.get(conversation_id, {}).get(job.id) is not job
+                    for job in pending
+                ):
+                    break
+                remaining = deadline - time.monotonic()
+                if remaining <= 0:
```
TimeToBuildBob, line 613: ⚠️ **P2** — In background_job_wait_hook, the timeout path sets job._wait_expired = True for all pending jobs and yields a timeout message. But the timeout message is only yielded once because _wait_expired is set. However, the hook also checks `if messages or not pending: break` at the top of the loop, and after the timeout, the pending list is empty (all _wait_expired), so the next invocation returns immediately. The issue is that the timeout message is yielded as a system message, and then StopPropagation is yielded. But the hook is registered at loop.continue priority 1500, and the completion hook at step.pre priority 900. When the timeout fires, the wait hook yields a system message and StopPropagation. This message is a system message that will be added to the conversation. But the timeout message says 'Automatic waiting is disabled for these jobs; late completions will still be reported.' However, the _wait_expired flag is set on the job, and the completion hook at step.pre will still deliver the completion when it arrives. But the wait hook's pending list filters out _wait_expired jobs, so it won't wait again. That is correct. The bug is that the timeout message is yielded even when there are no pending jobs? No, the timeout only fires when pending is non-empty. Actually, the timeout message is yielded when remaining <= 0, and pending is non-empty. That is fine. But consider the case where GPTME_WATCH_IDLE_MAX is set to 0. The deadline is time.monotonic() + 0, so remaining <= 0 immediately. The hook yields the timeout message and StopPropagation. But the test test_timeout_reports_once_but_preserves_late_completion expects that after the timeout, a second call returns [] because _wait_expired is set. That works. However, the timeout message is a system message that is injected into the conversation. This could be considered a behavior change, but it is intended. I don't see a clear bug here. Let me look for other issues.

tests/test_shell_completion_loop.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def reply(messages: list[Message], *args: object, **kwargs: object) -> Message:
+        observed_prompts.append(list(messages))
+        assert len(observed_prompts) <= len(replies), "Unexpected extra model turn"
+        if len(observed_prompts) == 2:
+            job = get_background_job(1)
+            assert job is not None and job.is_running(), (
+                "Job must remain blocked until the second reply releases it"
+            )
+            release_gate.write_text("continue\n")
```
TimeToBuildBob: ⚠️ **P2** — The FIFO release gate used in test_noninteractive_loop_receives_delayed_owned_completion and test_noninteractive_cli_receives_background_completion relies on a blocking write from an unprivileged test thread. The background job command opens the FIFO with `read -r release < FIFO`, and the test calls release_gate.write_text('continue\n') during the second model reply. If the child process has not yet reached its `read` (e.g., under heavy CI load the subprocess may be delayed past the mocked reply), the write blocks for as long as the child takes to open the FIFO. There is no timeout on the write, so a stalled or mis-scheduled child can hang the entire test suite indefinitely rather than failing with a clear assertion. The tests should use a non-blocking open/write (or a timeout plus explicit skip) so a slow child fails loudly instead of wedging CI.

gptme/tools/shell_background.py
TimeToBuildBob: Fixed in `ebb0bcd2d`. The wait still leaves `control.jsonl` for STEP_PRE, but it now fingerprints that yield and will not inject the same dummy system message on later LOOP_CONTINUE turns. A stale parent-session control file times out (or waits for the job) instead of burning model calls.

Regression: `test_control_file_does_not_reenter_every_turn`.

gptme/tools/shell_background.py
TimeToBuildBob: Not a defect. Timeout is one-shot via `_wait_expired`, late completions still arrive at STEP_PRE, and `GPTME_WATCH_IDLE_MAX=0` is the covered immediate-timeout path. No code change.

tests/test_shell_completion_loop.py
TimeToBuildBob: Fixed in `ebb0bcd2d`. FIFO release writes now use a non-blocking open with a 5s timeout, so a late or stalled child fails loudly instead of wedging the suite.

gptme/tools/shell_background.py
```diff
@@ -434,6 +461,172 @@ def background_job_completion_hook(
...
+    try:
+        stat = control.stat()
+    except FileNotFoundError:
+        return None
+    if not stat.st_size:
+        return None
+    # Leave the file for STEP_PRE. Yield once so a subagent cancel can re-enter
+    # that checkpoint; a second yield of the same contents is an infinite loop
+    # because the parent cancel hook no-ops when agent_id is unset.
+    key = (str(control.resolve()), stat.st_mtime_ns, stat.st_size)
```
TimeToBuildBob: ⚠️ **P2** — In _background_wait_input, the control-file fingerprint uses `stat.st_mtime_ns` and `stat.st_size`. If a control file is written with the same size within the same nanosecond (unlikely but possible on filesystems with coarse timestamps, e.g., some network filesystems or when the file is rewritten quickly), the key will collide and the wait hook will not yield to the new control file. More importantly, the key is added to _control_wait_keys and never removed until reset_background_jobs. If a control file is written, the wait hook yields once, then the file is left on disk. If the same control file is later modified (e.g., a new cancel op appended) but the mtime_ns and size happen to be identical (e.g., the file is truncated and rewritten with the same content length), the wait hook will not yield again. This can cause a subagent cancel or other control op to be missed during a long background wait. The consequence is that a cancel request written to control.jsonl after the first yield may not interrupt the wait, leaving the session stuck until timeout. The fix is to also include a content hash or to remove the key when the file is consumed by STEP_PRE.

tests/test_shell_completion_wait.py
```diff
@@ -0,0 +1,271 @@
...
+) -> None:
+    """A stale control file must not inject a dummy model turn every LOOP_CONTINUE."""
+    monkeypatch.setenv("GPTME_WATCH_IDLE_MAX", "0")
+    start_owned("sleep 30")
+    (tmp_path / "control.jsonl").write_text('{"op":"cancel"}\n')
+    manager = SimpleNamespace(chat_id="owner", logdir=tmp_path)
+    first = list(bg.background_job_wait_hook(manager, False, []))
+    assert isinstance(first[0], Message)
+    assert "pending session input" in first[0].content
+    second = list(bg.background_job_wait_hook(manager, False, []))
```
TimeToBuildBob, line 179: ⚠️ **P2** — In test_shell_completion_wait.py, the test `test_control_file_does_not_reenter_every_turn` sets GPTME_WATCH_IDLE_MAX to '0' and expects the second call to the wait hook to return a timeout message. However, the first call yields a dummy system message for the control file and returns StopPropagation. The second call, with the same control file, should not yield again because the key is in _control_wait_keys. But the test also expects the second call to time out immediately (since limit=0). The wait hook computes `deadline = time.monotonic() + limit` with limit=0, so the first iteration of the while loop will check `remaining <= 0` and set _wait_expired and yield the timeout message. However, the first call already yielded the control message and returned, so the second call starts fresh. The test asserts that the second call's first message does not contain 'pending session input' and does contain 'timed out'. This is correct. But the test does not verify that the control file key is actually preventing re-entry; it just checks the timeout. The test is not vacuous, but it could pass even if the key logic were broken, because the timeout would still occur. However, the test name suggests it is checking the key logic. The test would fail if the key logic were broken and the second call yielded the control message again, because then the first message would contain 'pending session input' and the assertion would fail. So the test is valid.

gptme/tools/shell_background.py
TimeToBuildBob: Fixed in `9cb572a05`. Control-file wait fingerprints now hash file contents, so a same-size rewrite (new cancel/pause op) interrupts wait again even on coarse timestamps.

Regression: `test_control_file_same_size_rewrite_reenters`.

gptme/tools/shell_background.py
TimeToBuildBob: Follow-up: the later review reused this fingerprint for a drain-race P1 (a claimed sibling completion aborting wait for remaining jobs). That path is fixed in `9cb572a05` by recomputing pending from live state before `wait()`. Timeout remains one-shot via `_wait_expired` as previously noted.

Regression: `test_wait_keeps_remaining_jobs_after_concurrent_drain`.

tests/test_shell_completion_wait.py
TimeToBuildBob: Not a defect. The second call would yield 'pending session input' again if the fingerprint failed, so the assertion does check the key logic; GPTME_WATCH_IDLE_MAX=0 only makes the no-reentry path fail fast instead of waiting. No code change.

PR #3845, opened 2026-09-16
fix(eval): keep token counts on timeout instead of zeroing them

TimeToBuildBob:
## Summary

Eval timeouts were recorded as 0-token / no-cost runs. That hides the
failure-cost tail: a model that burns tokens until the wall-clock cap
looks identical to one that never started.

Two paths dropped cost:

1. **Child `TimeoutExpired`** (Claude Code CLI timeout) wrote
   `status=timeout` without a cost block.
2. **Parent `Process.join` timeout** SIGTERM'd the child. The SIGTERM
   handler wrote `status=error` with no cost, and `execute()` then
   overwrote the parent timeout with that error.

## Change

- Snapshot `CostTracker` on success, error, timeout, and SIGTERM.
- Map parent SIGTERM to `status=timeout` (not error).
- If the parent already decided timeout, keep it unless the child
  already wrote success (join race).
- Tests: child timeout/error include cost; parent timeout keeps tokens
  even when the child wrote error.

## Test plan

- [x] `pytest tests/test_eval.py::test_act_process_maps_subprocess_timeout_to_timeout_result`
- [x] `pytest tests/test_eval.py::test_act_process_error_includes_cost`
- [x] `pytest tests/test_eval.py::test_execute_parent_timeout_keeps_child_tokens`
- [ ] CI eval/unit jobs

Force-kill after SIGTERM still cannot recover cost (process is gone).
That path stays 0 tokens; this PR covers the cases where the child
gets to write a result.


TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 3 findings disposed (rejected)

| --- | --- | --- |
| `gptme/eval/run.py:684` | P1 | superseded by latest review (not reproduced) |
| `gptme/eval/run.py:102` | P1 | rejected — Verification said: 'The parent execute() also reads sync_dict after join. No lock is used anywhere.' The next sentence t |
| `gptme/eval/run.py:683` | P1 | superseded by latest review (not reproduced) |
| `gptme/eval/run.py:752` | P1 | rejected — Verification checked execute() 410-421 and success-path 737-752, then treated child success after parent join-timeout as |
| `gptme/eval/run.py:706` | P1 | rejected — Verification says it traced error_handler write-then-SIG_IGN overwriting error with timeout. The finding body instead cl |
| `gptme/eval/run.py:686` | P2 | superseded by latest review (not reproduced) |
| `tests/test_eval.py:746` | P2 | superseded by latest review (not reproduced) |

</details>




This PR changes gptme/eval/run.py so that cost snapshots are captured on success, error, timeout, and SIGTERM paths, and so that a parent join-timeout keeps status=timeout (and the child's cost) unless the child already wrote success. It adds a _cost_snapshot() helper, a _write_result_unless_success() guard, and updates the SIGTERM handler to write a timeout result instead of delegating to error_handler. The tests add coverage for cost inclusion in timeout/error results and for parent-timeout status preservation.

Confidence 3/5

**❌ P1 high** — `gptme/eval/run.py:706`

The SIGTERM handler now writes a timeout result and then calls cleanup_process_group(), which sends SIGTERM to the process group. The handler sets SIG_IGN before writing, but the SIGTERM sent by _graceful_killpg to the process group will be ignored by this process (since SIG_IGN is set), so the handler won't re-enter. However, the handler does not reset the SIGTERM handler to SIG_DFL after cleanup. After act_process returns, the process exits normally, so this is not a problem. But if the process is reused (e.g., in a ProcessPoolExecutor worker), the SIGTERM handler remains SIG_IGN, which could cause the worker to ignore a subsequent SIGTERM from the parent's terminate() call, preventing graceful shutdown. The parent's run_evals() calls process.terminate() on active children, which sends SIGTERM; if the worker has SIG_IGN, it won't terminate and the parent will hang on join(). This is a real risk because act_process runs in a ProcessPoolExecutor worker (via execute() submitted to the executor), and workers are reused across tasks. The fix is to reset the handler to SIG_DFL after cleanup, as the previous code did.

```suggestion
Set signal.signal(signal.SIGTERM, signal.SIG_IGN) before writing the result in error_handler and the TimeoutExpired handler, or use _write_result_unless_success in those handlers as well.
```

<sub>**How this was verified:** Traced the error_handler: it writes the result, then sets SIG_IGN. A SIGTERM arriving between those two lines would trigger sigterm_handler, which calls _write_result_unless_success and overwrites the error with timeout. The same pattern exists in the TimeoutExpired handler at lines 727-730.</sub>

</details>

- `gptme/eval/run.py` — Adds cost snapshots to all result paths, changes SIGTERM handler to write timeout, and adds _write_result_unless_success guard.
- `tests/test_eval.py` — Adds tests for cost in timeout/error results, parent-timeout status preservation, and the new helper.

</details>

| `7feb8ffccafa` | 3/5 | 1 | llm | 2026-09-16 01:15 UTC |
| `f7988bb9650e` | 3/5 | 3 | llm | 2026-09-16 01:42 UTC |
| `34a07a8bbce3` | 5/5 | 0 | llm | 2026-09-16 02:23 UTC |
| `34a07a8bbce3` | 3/5 | 1 | llm | 2026-09-16 02:26 UTC |
| `3208844f3ddc` | 3/5 | 2 | llm | 2026-09-16 03:41 UTC |

</details>

<sub>Reviewed `3208844f3ddc` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 67s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/eval/run.py
```diff
@@ -649,19 +659,30 @@ def error_handler(e):
...
+        duration = time.time() - start
+        result_timeout: ProcessError = {
+            "status": "timeout",
+            "message": "SIGTERM received",
+            "stdout": stdout.getvalue(),
+            "stderr": stderr.getvalue(),
+            "duration": duration,
+            "cost": _cost_snapshot(),
+        }
+        sync_dict["result"] = result_timeout
```
TimeToBuildBob: ❌ **P1** — The SIGTERM handler unconditionally overwrites any existing result in `sync_dict`. If the child agent has already written a success result (e.g., the agent finished just before the parent's timeout and then blocked in `cleanup_process_group`), the SIGTERM handler will replace that success with a timeout result before the parent reads it. The parent's `parent_timeout` logic in `execute` (lines 402-405) only preserves a child's success if it is still present when the parent reads the dict, but because the handler clobbers it, the parent sees `child_status="timeout"` and — since `parent_timeout` is True and `timeout != "success"` — sets the final status to `"timeout"`. The eval is then reported as timed out and the check/run phase is skipped, even though the agent produced a successful result. The PR's stated intent to keep success on a join race is therefore not met for this case. _(3/3 passes)_

TimeToBuildBob: Fixed in f7988bb9650e. SIGTERM now goes through `_write_result_unless_success`, so a completed success is not overwritten; the success path also `SIG_IGN`s before publishing. Tests cover the helper and the execute join-race (`test_execute_parent_timeout_keeps_child_success`).

gptme/eval/run.py
```diff
@@ -78,11 +78,30 @@ class ProcessError(TypedDict):
...
+def _write_result_unless_success(sync_dict, result: ProcessResult) -> None:
+    """Write a process result unless a success is already recorded.
+
+    Parent join-timeout keeps a child's success (join raced with a completed
+    write). The SIGTERM handler must not clobber that success with timeout.
+    """
+    existing = sync_dict.get("result")
+    if isinstance(existing, dict) and existing.get("status") == "success":
+        return
+    sync_dict["result"] = result
```
TimeToBuildBob, line 106: ❌ **P1** — The `_write_result_unless_success` function is not atomic. It does a `sync_dict.get("result")` check and then a separate `sync_dict["result"] = result` assignment. In the SIGTERM handler, this runs in the child process while the parent may be concurrently writing to the same `sync_dict` (a `multiprocessing.Manager` dict). If the parent writes a success result between the child's `get` and `set`, the child's timeout result will overwrite the success. The PR's own comment acknowledges this race ('_write_result_unless_success is the remaining race guard') but the guard is not atomic, so the race it claims to guard against still exists. The consequence is that a successful eval can be reported as a timeout, losing the success status and the files/checks that would have been run. This is a real concurrency bug in the new code.

gptme/eval/run.py
```diff
@@ -649,19 +671,30 @@ def error_handler(e):
...
         # kill child processes gracefully
         cleanup_process_group()
 
-    # handle SIGTERM
+    # handle SIGTERM (parent join-timeout calls Process.terminate)
     def sigterm_handler(*_):
         # Reset to default handler first to prevent recursive SIGTERM loop:
         # _graceful_killpg sends SIGTERM to our own process group, which would
         # re-trigger this handler without this reset.
         signal.signal(signal.SIGTERM, signal.SIG_DFL)
```
TimeToBuildBob: ⚠️ **P2** — The `sigterm_handler` resets the SIGTERM handler to `SIG_DFL` before writing the result and cleaning up. This means that if the `_graceful_killpg` call (which sends SIGTERM to the process group) triggers another SIGTERM to this process, the default handler will terminate the process immediately, potentially before the `sync_dict["result"]` write is flushed to the Manager. The comment says the reset is to prevent a recursive SIGTERM loop, but the consequence is that the timeout result may never be written if the process is killed before the write completes. The PR's goal is to preserve cost on timeout, but this path can still lose the result entirely. The `_write_result_unless_success` call is before `cleanup_process_group()`, but the write to a Manager dict is not synchronous; it may be buffered. If the process dies before the Manager flushes, the parent sees no result and falls into the `else` branch, reporting an error with no cost.

tests/test_eval.py
```diff
@@ -651,6 +680,197 @@ def test_act_process_maps_subprocess_timeout_to_timeout_result():
...
+            return None
+
+        def is_alive(self):
+            return self._alive
+
+        def terminate(self):
+            sync_dict = self._args[4]
+            # Simulate the old SIGTERM handler writing error, which used to
+            # both drop cost and overwrite parent timeout status.
+            sync_dict["result"] = {
```
TimeToBuildBob, line 763: ⚠️ **P2** — The `test_execute_parent_timeout_keeps_child_tokens` test uses a `FakeProc` whose `terminate()` method writes a result with `status: "error"` directly to `sync_dict`, simulating the old SIGTERM handler. However, the new SIGTERM handler writes `status: "timeout"`, not `error`. The test is therefore testing the old behavior, not the new one. The test's comment says 'Simulate the old SIGTERM handler writing error', but the PR's change is to make the SIGTERM handler write timeout. The test should simulate the new handler (writing timeout) to verify the parent timeout preservation logic. As written, the test passes because the parent's `parent_timeout` logic overrides the child's error status, but it does not exercise the new code path where the child writes timeout. This is a test defect: it stubs the wrong behavior and gives a false sense of coverage.

TimeToBuildBob:
@greptileai review

gptme/eval/run.py
TimeToBuildBob: False positive. The verification itself shows the parent only *reads* `sync_dict` after join — it never writes a success result.

The claimed race is: "the parent writes a success result between the child's get and set." That writer does not exist. `execute()` waits on `p.join` / `terminate`, then reads. The only writers are in the child.

get-then-set is same-process: the SIGTERM handler interrupts the child. A `Manager.Lock` would not buy a second writer. Success already `SIG_IGN`s before publishing, so the handler cannot run during that write.

No lock added (YAGNI). Docstring now states the parent-never-writes invariant.

gptme/eval/run.py
TimeToBuildBob: Fixed in 34a07a8bb. Handler now `SIG_IGN`s before writing, same as the success / TimeoutExpired paths, so a process-group SIGTERM cannot SIG_DFL-kill us mid-write.

tests/test_eval.py
TimeToBuildBob: Fixed in 34a07a8bb. `test_execute_parent_timeout_keeps_child_tokens` now stubs the SIGTERM handler writing `timeout`. The error-handler path is a separate test (`test_execute_parent_timeout_overrides_child_error`).

gptme/eval/run.py
```diff
@@ -649,19 +675,31 @@ def error_handler(e):
             "stdout": stdout.getvalue(),
             "stderr": stderr.getvalue(),
             "duration": duration,
+            "cost": _cost_snapshot(),
         }
         sync_dict["result"] = result_error
 
         # kill child processes gracefully
         cleanup_process_group()
```
TimeToBuildBob, line 687: ❌ **P1** — error_handler writes a result with status="error" and then calls cleanup_process_group(), which sends SIGTERM to the process group including the current subprocess. Since the SIGTERM handler is still installed and does not check for an existing non-success result (only preserves "success" via _write_result_unless_success), it overwrites the error result with status="timeout", message="SIGTERM received". Consequently any non-timeout agent exception (e.g. RuntimeError or API failure) is reported as a timeout, losing the actual error message and misclassifying the eval. This is not caught by test_act_process_error_includes_cost because it patches _graceful_killpg, so the self-SIGTERM never fires. Fix by setting signal.signal(signal.SIGTERM, signal.SIG_IGN) in error_handler before calling cleanup, mirroring the success and TimeoutExpired paths. _(2/2 passes)_

TimeToBuildBob: Fixed in 3208844f3. `error_handler` now `SIG_IGN`s before `cleanup_process_group()`, matching the success / TimeoutExpired paths. `test_act_process_error_survives_cleanup_sigterm` sends SIGTERM during cleanup and asserts the result stays `error`.

gptme/eval/run.py
```diff
@@ -706,14 +744,13 @@ def sigterm_handler(*_):
         "duration": duration,
         "log_dir": agent.log_dir,
         "workspace_dir": agent.workspace_dir,
-        "cost": cost_dict,
+        "cost": _cost_snapshot(),
     }
+    # Ignore SIGTERM before publishing success so parent join-timeout cannot
+    # clobber it; _write_result_unless_success is the remaining race guard.
+    signal.signal(signal.SIGTERM, signal.SIG_IGN)
     sync_dict["result"] = result_success
```
TimeToBuildBob, line 752: ❌ **P1** — The success path in act_process writes sync_dict["result"] directly on line 752 instead of going through _write_result_unless_success, contradicting the comment on lines 749-750 that claims that helper is the remaining race guard. SIG_IGN is only installed on the preceding line, so a SIGTERM delivered after agent.act returns (line 712) but before line 751 runs the sigterm_handler first. That handler writes a record with status "timeout" to sync_dict via _write_result_unless_success. Execution then resumes, line 751 sets SIG_IGN, and line 752 overwrites the timeout with result_success, so the final record reports "success". In execute(), parent_timeout is True but child_status == "success" fails the first condition (line 418) and the elif (line 420) sets status to child_status, so the run is recorded as successful. This defeats the PR's goal of preserving timeout status — a run the parent timed out is silently reported as success, hiding a slow-burning model.

TimeToBuildBob: False positive — this is the join-race the PR is designed to keep.

`agent.act()` already returned. The child finished. Parent `join` timeout firing while we publish success is exactly `test_execute_parent_timeout_keeps_child_success`: keep success, do not report timeout.

`_write_result_unless_success` is the SIGTERM handler's guard so a *later* timeout write cannot clobber success. The success path must write directly (or through a helper that still overwrites timeout/error). Routing success through that helper would not change the outcome: existing timeout is not success, so success would still overwrite it.

`execute()` then sees `parent_timeout and child_status == "success"` and keeps success. That is the intended invariant, not a bug: a completed run must not be mis-reported as a timeout.

P2 on `error_handler` write-then-SIG_IGN is a nanosecond window, not the deterministic self-SIGTERM from `cleanup_process_group()` already fixed in 3208844f3. Not changing it.

gptme/eval/run.py
```diff
@@ -649,19 +675,35 @@ def error_handler(e):
...
+        result_timeout: ProcessError = {
+            "status": "timeout",
+            "message": "SIGTERM received",
+            "stdout": stdout.getvalue(),
+            "stderr": stderr.getvalue(),
+            "duration": duration,
+            "cost": _cost_snapshot(),
+        }
+        _write_result_unless_success(sync_dict, result_timeout)
+        cleanup_process_group()
```
TimeToBuildBob, line 706: ❌ **P1** — The SIGTERM handler now writes a timeout result and then calls cleanup_process_group(), which sends SIGTERM to the process group. The handler sets SIG_IGN before writing, but the SIGTERM sent by _graceful_killpg to the process group will be ignored by this process (since SIG_IGN is set), so the handler won't re-enter. However, the handler does not reset the SIGTERM handler to SIG_DFL after cleanup. After act_process returns, the process exits normally, so this is not a problem. But if the process is reused (e.g., in a ProcessPoolExecutor worker), the SIGTERM handler remains SIG_IGN, which could cause the worker to ignore a subsequent SIGTERM from the parent's terminate() call, preventing graceful shutdown. The parent's run_evals() calls process.terminate() on active children, which sends SIGTERM; if the worker has SIG_IGN, it won't terminate and the parent will hang on join(). This is a real risk because act_process runs in a ProcessPoolExecutor worker (via execute() submitted to the executor), and workers are reused across tasks. The fix is to reset the handler to SIG_DFL after cleanup, as the previous code did. _(2/2 passes)_

TimeToBuildBob: False positive. `SIG_IGN` is installed in `act_process`, which `execute()` runs as a fresh `multiprocessing.Process` (line 371), not inside the `ProcessPoolExecutor` worker.

The pool worker runs `execute()`. That process never sets SIG_IGN. The child that does set it exits when `act_process` returns, so there is no reused worker carrying SIG_IGN into a later eval.

`run_evals()` `process.terminate()` of `active_children()` targets those short-lived `act_process` children, not pool workers. Resetting to SIG_DFL after cleanup was the previous P2: process-group SIGTERM then SIG_DFL-killed the child mid Manager-dict write. SIG_IGN is the fix, not a regression.

The "How this was verified" paragraph describes a different race (error_handler write-then-SIG_IGN) than the finding body (pool-worker reuse).

TimeToBuildBob:
Squash-merged as `53fd88323682` at 2026-09-16T03:56:24Z. Pre-merge CI green. Eval timeouts now keep CostTracker snapshots instead of reporting 0 tokens. No further action.

PR #3842, opened 2026-09-15
fix(cli): emit clean error on dataset export -o write failure

TimeToBuildBob:
## Summary

`gptme-util dataset export -o <path>` crashed with a raw Python traceback when the output parent path was missing or was a regular file.

This is the same bug class as gptme/gptme#3824 (`gptme-util status -o`), found by dogfooding the new `dataset` CLI rather than re-probing the saturated HTTP input boundary.

**Before:**
```
$ gptme-util dataset export -o /nonexistent/dir/foo.jsonl -n 1
Traceback (most recent call last):
  ...
FileNotFoundError: [Errno 2] No such file or directory: '/nonexistent/dir/.foo.jsonl.XXXX'
```

**After:**
```
$ gptme-util dataset export -o /tmp/new/subdir/foo.jsonl -n 1
Exported 0 environments.   # parent auto-created, exit 0

$ gptme-util dataset export -o /etc/hostname/x.jsonl -n 1
Error: Failed to write dataset export to /etc/hostname/x.jsonl: ...
```

## How to reproduce

```bash
gptme-util dataset export -o /tmp/definitely-missing-725f/envs.jsonl -n 1
# master: FileNotFoundError traceback
# this branch: auto-mkdir parent, export, exit 0

gptme-util dataset export -o /etc/hostname/x.jsonl -n 1
# master: NotADirectoryError traceback
# this branch: Error: Failed to write dataset export to ... (exit 1, no traceback)
```

## Changes

- `mkdir(parents=True)` before creating the atomic tempfile next to the destination
- `OSError` → `click.ClickException` around tempfile creation and `os.replace`
- Regression tests in `tests/test_dataset.py`

## Test plan

- [x] `python -m pytest tests/test_dataset.py` — 38 passed
- [x] Manual repro of missing parent and parent-is-file paths

Note: overlap-guard false-positive against gptme/gptme#3696 (knowledge schema; disjoint files). This PR only touches `gptme/cli/cmd_dataset.py` and `tests/test_dataset.py`.

PR #3849, opened 2026-09-16
feat(site): deploy design-v3 landing page to gptme.org

TimeToBuildBob:
## Summary

Cut gptme.org over from the pandoc README homepage to the design-v3 site in `site/next/` (gptme/gptme#3839), without a second GitHub Pages deploy.

`make site` now:

1. Prerenders `site/next/` (Vite + SSR)
2. Overlays Sphinx docs, `/downloads/`, `badge.svg`, the full `media/` tree, `CNAME`, and `.nojekyll`
3. Runs `make site-check` so CI cannot publish a pandoc homepage or a tree missing `/readme/` / `/docs/` / `/media/`

`.github/workflows/docs.yml` is still the **only** production deploy (gh-pages). Pandoc is no longer installed in that workflow. `.github/workflows/site-next.yml` stays PR-preview-only (Cloudflare Pages) so the two workflows cannot race.

Old inbound hashes such as https://gptme.org/#installation redirect to `/readme/#installation` when the new homepage has no matching id.

## What stays the same

- `/docs/` (Sphinx)
- `/downloads/`
- `/media/*` (README images keep loading from `https://gptme.org/media/`)
- Custom domain via `CNAME`

## What changes on merge

- `/` is the design-v3 landing page (hero, terminal session, stats)
- `/readme/` is the rendered README with working TOC anchors
- The pandoc template (`site/template.html`) is unused; kept for rollback

## Test plan

- [x] `cd site/next && npm test && npm run build` (11 tests, prerender writes `/` and `/readme/`)
- [x] `make site-assemble site-check` against a stub Sphinx tree
- [x] Local smoke: `/`, `/readme/`, `/docs/`, `/downloads/`, `/media/logo.png`, `/media/screenshots/tui.png` all 200
- [ ] CI `Docs` workflow green on this PR (full Sphinx + assemble + site-check)
- [ ] After merge: fetch https://gptme.org/ and confirm the design-v3 hero; `/readme/` anchors and `https://gptme.org/media/` images; `/docs/` still serves Sphinx

## Rollback

Revert this PR. `docs.yml` will rebuild the old pandoc homepage from `site/template.html`.

PR #3850, opened 2026-09-16
fix(cli): emit clean error on attest sign --out write failure

TimeToBuildBob:
## Summary

`gptme-util attest sign --out` dumped a raw Python traceback when the destination parent was unwritable (`PermissionError` from `mkdir`) or a regular file (`FileExistsError`). Same bug class as #3842 (`dataset export -o`) and #3824 (`status -o`).

`write_attestation()` now converts those `OSError`s to `AttestationError`, which the CLI already maps to a clean `ClickException`. Missing parent dirs still auto-mkdir.

## Test plan

- [x] `PYTHONPATH=<worktree> pytest tests/test_util_cli_attest.py` — 14 passed
- [x] Reproduced pre-fix: `--out /tmp/locked/nested/att.json` (chmod 555 parent) → raw `PermissionError`
- [x] Reproduced pre-fix: `--out /etc/hostname/att.json` → raw `FileExistsError`
- [x] Regression tests cover parent-is-file, mkdir `PermissionError`, and auto-mkdir of nested parents


gptme/attestation.py
```diff
@@ -240,8 +240,13 @@ def write_attestation(
     out_path = destination or default_attestation_path(
         workspace_root, attestation["id"]
     )
-    out_path.parent.mkdir(parents=True, exist_ok=True)
-    out_path.write_text(json.dumps(attestation, indent=2, sort_keys=True) + "\n")
+    try:
+        out_path.parent.mkdir(parents=True, exist_ok=True)
+        out_path.write_text(json.dumps(attestation, indent=2, sort_keys=True) + "\n")
```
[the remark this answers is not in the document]
TimeToBuildBob: Added `test_attest_sign_out_is_directory_emits_clean_error` (83720c328): patches `Path.write_text` to raise `IsADirectoryError` so the write-failure half of the clean-error contract is covered. Note the literal "destination is a directory" case can't reach `write_attestation` — Click's `--out` path option rejects directories at parse time (exit 2) — so the patched-write form is the reachable coverage.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


| --- | --- | --- |
| `tests/test_util_cli_attest.py:388` | P1 | superseded by latest review (not reproduced) |
| `gptme/attestation.py:247` | P2 | rejected |

</details>




### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `gptme/attestation.py:247`

`write_attestation` previously let OSError propagate to callers; now it catches all OSError subclasses and re-raises as `AttestationError`, which is a `ValueError` subclass, not an `OSError`. Any existing caller that catches `OSError` to handle disk-full, permission, or path-construction failures will silently miss the converted error because `AttestationError` is not an `OSError`. For example, a caller doing `try: write_attestation(...) except OSError: ...` will no longer enter that block, and the error surfaces as an unhandled `AttestationError`. The new tests only assert CLI output and do not pin the exception type, so the contract change is not covered.

```suggestion
Consider documenting the exception-type change in the docstring, or making AttestationError subclass OSError so existing OSError handlers continue to work.
```

<sub>**How this was verified:** Checked the except clause at line 246 only catches OSError. UnicodeEncodeError is a subclass of ValueError, not OSError. The CLI tests use CliRunner.invoke without catch_exceptions=False, so they would show tracebacks for uncaught exceptions. The PR's stated goal is to avoid raw tracebacks.</sub>

</details>

| `83720c328698` | 3/5 | 1 | llm | 2026-09-16 18:02 UTC |

</details>

<sub>Reviewed `71ecc62818a3` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 610s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_util_cli_attest.py
```diff
@@ -335,3 +335,141 @@ def test_get_agent_id_falls_back_to_uid_when_username_lookup_fails(monkeypatch):
...
+    repo.mkdir()
+    _init_repo(repo)
+
+    monkeypatch.setenv("CC_MODEL", "gpt-5.4")
+    out_path = tmp_path / "nested" / "att.json"
+
+    def boom(*_args, **_kwargs):
+        raise PermissionError(13, "Permission denied")
+
+    monkeypatch.setattr(Path, "mkdir", boom)
```
TimeToBuildBob, line 391: ❌ **P1** — The new tests monkeypatch Path.mkdir and Path.write_text globally, which also affects the _init_repo helper's use of Path.write_text (repo / "README.md").write_text("hello\n") and the git operations. In test_attest_sign_out_unwritable_parent_emits_clean_error, monkeypatch.setattr(Path, "mkdir", boom) is applied before _init_repo(repo) is called, so _init_repo's internal git init/config/commit will call the patched mkdir and raise PermissionError, causing the test to fail before the CLI is even invoked. The same applies to test_attest_sign_out_is_directory_emits_clean_error, where monkeypatch.setattr(Path, "write_text", boom) is applied before _init_repo, so the README.md write in _init_repo raises IsADirectoryError, failing the test setup. The tests as written cannot pass; they need to either patch after _init_repo or patch only the specific Path instance used for out_path.

TimeToBuildBob:
@greptileai review

tests/test_util_cli_attest.py
TimeToBuildBob: Fixed in 71ecc62818. Scoped the Path.mkdir / Path.write_text monkeypatches to the destination parent/file so they no longer intercept unrelated Path calls during CLI invocation.

gptme/attestation.py
```diff
@@ -240,8 +240,13 @@ def write_attestation(
...
+    except OSError as exc:
+        raise AttestationError(
```
TimeToBuildBob, line 247: ⚠️ **P2** — `write_attestation` previously let OSError propagate to callers; now it catches all OSError subclasses and re-raises as `AttestationError`, which is a `ValueError` subclass, not an `OSError`. Any existing caller that catches `OSError` to handle disk-full, permission, or path-construction failures will silently miss the converted error because `AttestationError` is not an `OSError`. For example, a caller doing `try: write_attestation(...) except OSError: ...` will no longer enter that block, and the error surfaces as an unhandled `AttestationError`. The new tests only assert CLI output and do not pin the exception type, so the contract change is not covered. _(3/3 passes)_

TimeToBuildBob: Won't-fix. The conversion to AttestationError is the bug this PR is shipping: the CLI already maps AttestationError → ClickException (cmd_attest.py:92), and write_attestation has no other in-repo caller. AttestationError stays a ValueError — it is also raised for JSON/validation failures, so subclassing OSError would lie for those paths. Callers that want a clean failure already catch AttestationError; there is no remaining OSError handler to preserve.

TimeToBuildBob:
CI red is the same xdist cwd-leak flake as gptme/gptme#3849, not this attest change. Reran the failed Test job (`35133056011 --failed`).

- **Test without API keys**: `tests/test_tools_shell.py::test_shell_cd_chdir` asserted `os.getcwd()` after shell `cd`, but the process cwd was still `/tmp/pytest-of-runner/.../popen-gw15/test_server_cost_windows_isola0`. All 3 retries failed with that same leaked path.
- This PR only touches `write_attestation` / `tests/test_util_cli_attest.py`. First two Test runs on this branch (`1822bcef`, `83720c32`) were green. Master Test is green. gptme/gptme#3849's rerun of the same flake came back green.

Not mixing flake hardening into this attest PR. Next session: if the rerun is green, this is merge-ready on CI; if `test_shell_cd_chdir` fails again, extract the cwd-leak fix to its own PR.

TimeToBuildBob:
The rerun (attempt 2 of `35133056011`) came back red again, but on a **different** unrelated test — which is what makes this a runner flake rather than this change.

**Attempt 1**: `tests/test_tools_shell.py::test_shell_cd_chdir` (xdist cwd leak, same as gptme/gptme#3849).
**Attempt 2**: `tests/test_server_v2.py::test_start_tool_execution_streams_only_real_tool_output` and `..._uses_branch_for_reload_and_continuation`, both `ERROR at setup` with `assert 409 == 200`.

The 409 is `api_conversation_put` returning "Conversation already exists" from `logdir.mkdir()` hitting `FileExistsError`. `create_conversation()` builds the name as `test-server-v2-{random.randint(0, 1000000)}`, and `tests/test_lessons_auto_include.py` seeds the global `random` module (`random.seed(0/42/99)`) inside its own tests. Under `-n 16` those seeds leak into each worker's global RNG stream, so the "random" names are no longer independent across workers sharing one `get_logs_dir()` — a collision stops being a 1-in-a-million event. That is the most plausible mechanism (not reproduced here) for why both tests on the same worker failed at setup in the same run.

This change touches only `gptme/attestation.py` and `tests/test_util_cli_attest.py` — nothing in the server, shell, or test infrastructure. Master's `Test` workflow is green on the same base (`0881c1d980`).

Reran the failed jobs again (attempt 3). No code change here; the flake and its root cause are a separate concern.

PR #3852, opened 2026-09-17
fix(cli): validate non-empty query in context retrieve/search-conversations; catch RuntimeError

TimeToBuildBob:
## Problem

`gptme-util context retrieve ""` and `gptme-util context search-conversations ""` passed the empty string to `gptme-rag search ''`, which would hang for the full 60-second timeout then dump a raw `RuntimeError` traceback:

```
RuntimeError: gptme-rag command timed out after 60s
  File ".../gptme/cli/util.py", line 498, in context_retrieve
    results = rag_search(query, return_full=full)
  ...
```

Two bugs in one:
1. Empty/whitespace query is accepted and passed through (hangs for 60s)
2. `RuntimeError` from `rag_search` propagates as an unhandled exception with full traceback instead of a clean user-facing error

## Fix

- Reject empty/whitespace queries immediately with `Error: query cannot be empty.` (exit 1) — no subprocess spawned
- Wrap `rag_search` in a `try/except RuntimeError` and emit a clean error message (exit 1) for timeout and other failures

Same fix applied to both `context retrieve` and `context search-conversations`.

## Tests

Two regression tests added (`test_context_retrieve_empty_query`, `test_context_search_conversations_empty_query`) — no `gptme-rag` required, run fully offline.


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Adds empty/whitespace query validation to the context retrieve and search-conversations CLI commands, exiting with code 1 and a clean error message before any RAG subprocess is spawned. Also wraps rag_search calls in try/except RuntimeError to surface timeouts and other failures as clean errors instead of tracebacks. Adds regression tests for empty, whitespace-only, and RuntimeError cases.

Confidence 3/5

**❌ P1 high** — `gptme/cli/util.py:488`

The new empty-query guard runs after the lazy import of `_has_gptme_rag, init, rag_search` from `..tools.rag`. If gptme-rag is not installed, importing `gptme.tools.rag` may itself raise an ImportError (the module likely imports the optional `gptme_rag` package at module level). Previously, an empty query with gptme-rag missing would hit the `_has_gptme_rag()` check and print the friendly 'gptme-rag is not installed' message. Now the import happens first, so an empty query on a system without gptme-rag raises an unhandled ImportError traceback instead of the intended clean error. The guard should be moved before the import, or the import should be wrapped, to preserve the existing behavior for the missing-dependency case.

```suggestion
Move the `from ..tools.rag import ...` line after the empty-query guard, or wrap the import in try/except ImportError.
```

<sub>**How this was verified:** Checked the order in the new code: the `from ..tools.rag import ...` line is at line 486, before the guard at line 488. The existing `_has_gptme_rag()` check at line 492 is after the guard. If the import fails, the guard never runs.</sub>

</details>

- `gptme/cli/util.py` — Adds query.strip() validation and RuntimeError handling to context_retrieve and context_search_conversations.
- `tests/test_util_cli.py` — Adds tests for empty/whitespace query rejection and RuntimeError handling in both context commands.

</details>

| `3a5dfc75e8f9` | 4/5 | 2 | llm | 2026-09-17 01:41 UTC |

</details>

<sub>Reviewed `8cebb79c07ce` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 13s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/cli/util.py
```diff
@@ -495,7 +499,11 @@ def context_retrieve(query: str, full: bool):
     init()
 
     # Search for the query
-    results = rag_search(query, return_full=full)
+    try:
+        results = rag_search(query, return_full=full)
+    except RuntimeError as e:
```
TimeToBuildBob, line 504: ⚠️ **P2** — The RuntimeError handler catches only `RuntimeError` from `rag_search`. The PR description mentions the timeout error is a RuntimeError, but `rag_search` may raise other exceptions (e.g., `subprocess.TimeoutExpired`, `FileNotFoundError`, `OSError`) depending on the gptme-rag implementation. If a non-RuntimeError exception occurs, it will still propagate as an unhandled traceback, defeating the purpose of the fix. The tests only cover RuntimeError, so other failure modes remain unhandled. This is a real gap in error handling, though the specific failing input depends on the rag_search implementation. Since the PR's stated goal is to avoid tracebacks for rag failures, catching only RuntimeError is incomplete.

tests/test_util_cli.py
```diff
@@ -1255,6 +1255,53 @@ def test_context_search_conversations_top_k(tmp_path):
...
+
+    runner = CliRunner()
+    with (
+        patch("gptme.tools.rag._has_gptme_rag", return_value=True),
+        patch(
+            "gptme.tools.rag.rag_search",
+            side_effect=RuntimeError("gptme-rag command timed out after 60s"),
+        ),
+    ):
+        result = runner.invoke(main, ["context", command, "pytest"])
```
TimeToBuildBob, line 1302: ⚠️ **P2** — The new tests for empty query and RuntimeError use `CliRunner().invoke(main, ...)` without isolating the filesystem or environment. The `test_context_rag_search_runtime_error` patches `gptme.tools.rag._has_gptme_rag` and `gptme.tools.rag.rag_search`, but the command also calls `init()` from the same module. If `init()` is not patched, it may attempt to initialize the RAG store, which could fail or have side effects in the test environment. However, the test patches `_has_gptme_rag` to return True, so `init()` is called. If `init()` raises a non-RuntimeError exception, the test would fail with a traceback, but the test asserts no traceback, so it would catch that. The test is not hermetic because it depends on the real `init()` behavior, which may require gptme-rag to be installed or a writable index directory. This could cause flaky failures in environments without gptme-rag. The test should also patch `init()` to a no-op to be hermetic.

gptme/cli/util.py
TimeToBuildBob: False positive. `rag_search` routes every subprocess failure through `_run_rag_cmd` (gptme/tools/rag.py:159), which converts `subprocess.TimeoutExpired` and `CalledProcessError` into `RuntimeError`. The remaining escape — a missing `gptme-rag` binary raising `FileNotFoundError` — is guarded by the `_has_gptme_rag()` check earlier in both commands. Broadening to bare `except Exception` would mask genuine bugs, so the catch stays scoped to `RuntimeError` deliberately.

tests/test_util_cli.py
TimeToBuildBob: Fixed in 8cebb79c: the RuntimeError tests now patch `gptme.tools.rag.init` alongside `_has_gptme_rag`, so `context retrieve` cannot touch the real project config or index directory in CI. (Full test selection passes locally: 21 passed.)

TimeToBuildBob:
@greptileai review

gptme/cli/util.py
```diff
@@ -485,6 +485,10 @@ def context_retrieve(query: str, full: bool):
     """Search indexed documents for relevant context."""
     from ..tools.rag import _has_gptme_rag, init, rag_search  # fmt: skip
 
+    if not query.strip():
```
TimeToBuildBob, line 488: ❌ **P1** — The new empty-query guard runs after the lazy import of `_has_gptme_rag, init, rag_search` from `..tools.rag`. If gptme-rag is not installed, importing `gptme.tools.rag` may itself raise an ImportError (the module likely imports the optional `gptme_rag` package at module level). Previously, an empty query with gptme-rag missing would hit the `_has_gptme_rag()` check and print the friendly 'gptme-rag is not installed' message. Now the import happens first, so an empty query on a system without gptme-rag raises an unhandled ImportError traceback instead of the intended clean error. The guard should be moved before the import, or the import should be wrapped, to preserve the existing behavior for the missing-dependency case. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: False positive — verified against the source. `gptme/tools/rag.py` has no module-level import of the optional `gptme_rag` package (lines 44–58 are stdlib + gptme-internal only). The `gptme-rag` binary is invoked as a subprocess and availability is checked via `shutil.which` in `_has_gptme_rag()`, so importing `gptme.tools.rag` cannot raise ImportError when gptme-rag is missing. Guard placement after the lazy import is safe, and behavior is unchanged from before this PR.

TimeToBuildBob:
Addressed the review round at 8cebb79c:

- **Fixed (P2, test hermeticity):** the `RuntimeError` regression tests now patch `gptme.tools.rag.init` so `context retrieve` cannot touch the real project config/index in CI (`8cebb79c`). Selected tests pass locally (21 passed).
- **Rejected (P1, ImportError premise):** verified `gptme/tools/rag.py` has no module-level dependency on the optional `gptme_rag` package — the binary is a subprocess checked via `shutil.which`, so the lazy import cannot raise ImportError. Thread replied + resolved.
- **Rejected (P2, bare `except Exception`):** `rag_search` routes subprocess timeout/failure through `_run_rag_cmd`, which normalizes both to `RuntimeError`; the missing-binary case is guarded by `_has_gptme_rag()`. Keeping the catch scoped deliberately.

Greptile re-triggered on the new head; CI re-running. Remaining: green CI, then this is merge-ready.

tests/test_util_cli.py
```diff
@@ -1255,6 +1255,24 @@ def test_context_search_conversations_top_k(tmp_path):
     mock_search.assert_called_once_with("pytest", return_full=True, top_k=5)
 
 
+def test_context_retrieve_empty_query():
+    """context retrieve rejects an empty query with exit code 1, no traceback."""
+    runner = CliRunner()
+    result = runner.invoke(main, ["context", "retrieve", ""])
+    assert result.exit_code == 1
+    assert "Error: query cannot be empty" in result.output
+    assert "Traceback" not in result.output
```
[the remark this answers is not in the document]
TimeToBuildBob: Both requested cases are already covered in this PR's test additions:

- Whitespace-only queries: `test_context_whitespace_only_query`, parametrized over both `retrieve` and `search-conversations`.
- Mocked `rag_search` failures: `test_context_rag_search_runtime_error`, also parametrized over both commands, patching `rag_search` to raise `RuntimeError` and asserting a clean exit (no traceback).

The only new combination not literally enumerated is a mocked failure per individual whitespace input, which adds no distinct code path — validation short-circuits before `rag_search` is ever invoked.

TimeToBuildBob:
Status after the latest review cycles — requesting a human merge decision, since our merge gate is now blocked on a finding that was verified as a false premise:

- CI fully green on `8cebb79c` (18/18 checks), mergeStateStatus CLEAN, Greptile 5/5.
- Our AI-review consensus scores 3/5, re-raising the P1 "empty-query guard runs after the rag import, so a missing gptme-rag import would traceback" finding. That premise is false: `gptme/tools/rag.py` never imports `gptme_rag` at module level — the optional dependency is only probed inside `_has_gptme_rag()` (line 154), so importing `gptme.tools.rag` cannot raise ImportError when gptme-rag is missing. The same RuntimeError-only-catch point was a deliberate trade-off (`_run_rag_cmd` normalizes subprocess failures) and was disposed with evidence earlier; the fresh consensus pass re-raises both without new information.
- Re-running the merge gate just re-polls the same consensus, so we've stopped rather than loop.

Everything actionable from both reviewers is addressed on-thread. @ErikBjare — needs your review/merge, or tell us if you'd like the guard moved above the import anyway as a defensive change.

TimeToBuildBob:
New evidence on the mock-path concern raised by the latest consensus pass ("patching `gptme.tools.rag.rag_search` targets the wrong namespace because `util.py` uses `from ..tools.rag import rag_search`"):

The imports in `context_retrieve`/`context_search_conversations` are **function-local**, not module-level — `from ..tools.rag import ...` executes at call time, re-reading the (patched) module attributes. So `patch("gptme.tools.rag.rag_search", ...)` is the correct target and does intercept the call.

Verified empirically just now against the PR head (`8cebb79`):

```
$ pytest tests/test_util_cli.py -k "empty_query or whitespace or runtime_error"
....... [100%]  7 passed in 1.25s
```

That includes both `test_context_rag_search_runtime_error` params asserting `"Error: gptme-rag command timed out after 60s" in result.output` — the patched `side_effect` fired. CI's "Test without API keys" job (a gptme-rag-less environment) also ran these on `8cebb79` and passed, which rules out any real-path invocation or 60s hang.

No code change needed; the PR is unchanged at `8cebb79` with CI 18/18 green and all review threads (ours + Greptile's, which withdrew) resolved. Still requesting a human merge decision from @ErikBjare per the previous status.

PR #3851, opened 2026-09-16
test(lessons): stop seeding the global RNG in dropout tests

TimeToBuildBob:
Fixes a recurring, hard-to-diagnose CI flake that shows up as unrelated PRs failing `Test without API keys`.

## Symptom

gptme/gptme#3850's rerun failed with two `ERROR at setup` in `tests/test_server_v2.py`:

```
ERROR tests/test_server_v2.py::test_start_tool_execution_streams_only_real_tool_output - assert 409 == 200
```

`409` is `api_conversation_put` hitting `FileExistsError` on `logdir.mkdir()` → "Conversation already exists". `create_conversation()` names the conversation with `random.randint(0, 1000000)`, so collisions should be ~1-in-a-million — yet two tests on the same xdist worker failed at once.

## Cause

`tests/test_lessons_auto_include.py` called `random.seed(42/0/99/0)` inside four dropout tests. That mutates the **global** RNG of the worker process; the seed outlives the test. With `-n 16`, workers that pass through those seeds generate correlated "random" sequences, and because the server tests share one `get_logs_dir()` on the filesystem, a name generated on one worker can already exist from another. The collision stops being improbable.

This is a test-isolation bug in its own right: any later test on the same worker that depends on `random` gets a deterministic stream it did not ask for.

## Change

Swap `random.seed(N)` for `monkeypatch.setattr(_random, "random", _random.Random(N).random)` — same deterministic stream for the dropout assertions, reverted automatically after each test, no process-wide mutation. This matches the pattern already used two tests below (`monkeypatch.setattr(_random, "random", lambda: 0.9)`).

## Verification

- `pytest tests/test_lessons_auto_include.py -q` → `68 passed, 1 failed`
- The one failure, `test_policy_manifest_cache_reloads_same_size_preserved_mtime`, is pre-existing on unmodified `master` in a local run of the full file (and passes in isolation) — reproduced with the patch stashed, so it is unrelated to this change.
- `ruff check` / `ruff format --check` clean.

PR #3854, opened 2026-09-17
fix(tests): reset leaked shell workspace cwd state between tests

TimeToBuildBob:
## Problem

`Test without API keys` (xdist `-n 16`) intermittently reds unrelated PRs.
Diagnosed in `journal/2026-09-16/monitoring-pr3850-ci-flake-rerun.md`:

`tests/test_tools_shell.py::test_shell_cd_chdir` asserts `os.getcwd()` after a
shell `cd`, but the process cwd was pinned to
`test_server_cost_windows_isolated_across_conversations`'s leftover workspace
path — deterministic per run, not random.

## Root cause

`gptme/server/session_step.py`'s `step()` calls `set_workspace_cwd(workspace)`
on the first user message of a new conversation. That sets a module-level
`ContextVar` (`_workspace_cwd`) with no matching `.reset()`. Per
`ShellSession._set_cwd`:

```python
if changed and get_workspace_cwd() is None:
    os.chdir(cwd)
```

`_set_cwd` deliberately skips the process-wide `os.chdir()` whenever
`get_workspace_cwd()` is set — that's correct for concurrent server sessions.
But pytest runs tests synchronously in one thread/context, so once a
server-mode test sets that ContextVar, it stays set for every later test on
the same worker. A later CLI-style shell test (`test_shell_cd_chdir`) then
sees its own `shell.run("cd ...")` update the tracked cwd internally while
`os.getcwd()` never moves — it stays wherever the previous test's `step()`
last `os.chdir()`'d.

Reproduced locally by running the two tests in the same process:

```
tests/test_server_skill_events.py::test_server_cost_windows_isolated_across_conversations
tests/test_tools_shell.py::test_shell_cd_chdir
```

```
AssertionError: assert '/tmp/pytest-.../test_server_cost_windows_isola0' == '/tmp/tmpn2ytnmjv'
```

(The sibling flake mechanism, a `random.seed()` leak producing deterministic
`409 Conversation already exists` collisions, was already fixed in #3851.)

## Fix

Two autouse fixtures in `tests/conftest.py`:

- `restore_cwd` — unconditionally restores the process cwd after every test.
  Several code paths do a process-wide `os.chdir()` with no matching restore
  (this one, plus a few `os.chdir(workspace)` fallbacks elsewhere); this closes
  the gap generically instead of chasing each call site.
- `cleanup_shell_after` — now also resets the `_workspace_cwd` ContextVar to
  `None`, so a server test's `set_workspace_cwd()` can never leak into a later
  test's `ShellSession._set_cwd()` decision.

No production code changed — this is a test-isolation fix.

## Verification

- Reproduced the exact CI failure locally (same two tests, same process,
  before the fix).
- Confirmed both tests pass together after the fix.
- Full `tests/test_tools_shell.py` + `tests/test_server_skill_events.py`
  (163 tests) green.
- `ruff format`/`ruff check` clean; pre-commit passed.


TimeToBuildBob:
## 🤖 AI code review

Adds two autouse fixtures to tests/conftest.py: restore_cwd saves the process cwd before each test and restores it after, and cleanup_shell_after now also resets the shell module's _workspace_cwd ContextVar to None after each test. The change is test-only and aims to prevent leaked cwd and workspace-cwd state from one test affecting later tests on the same pytest-xdist worker.

Confidence 3/5

**❌ P1 high** — `tests/conftest.py:456`

The cleanup_shell_after fixture now unconditionally sets shell_module._workspace_cwd.set(None) after every test. However, this resets the ContextVar without using a token from a previous set, which is fine for isolation, but it also resets the workspace cwd for tests that legitimately set it during the test and expect it to be cleared. The real issue is that the fixture does not restore the previous value; it just sets to None. If a test sets _workspace_cwd to a non-None value and then the test itself or a later fixture reads it during teardown (e.g., a shell close that checks get_workspace_cwd()), the value is already None, which may change behavior. But more concretely, the PR's own description says the ContextVar is set with no matching .reset(), and the fix uses .set(None) which is a permanent change, not a reset. This is acceptable for test isolation, but it means that if a test's teardown (e.g., a shell session's close) relies on the workspace cwd being set, it will now see None. The PR claims this is correct, but it could break tests that expect the workspace cwd to be available during teardown. However, without a concrete failing test, this is speculative. The more concrete defect is that the fixture's teardown order relative to restore_cwd is wrong: cleanup_shell_after is defined after restore_cwd, so its teardown runs BEFORE restore_cwd's teardown. That means if cleanup_shell_after's teardown (or a shell close it triggers) does an os.chdir(), restore_cwd will not undo it. The PR's stated goal of 'undoing a chdir made during another fixture's teardown too' is not achieved because restore_cwd's teardown runs first, not last.

```suggestion
To ensure restore_cwd runs last, move its definition to the end of the autouse fixtures, or use a pytest_runtest_teardown hook that runs after all fixture teardowns.
```

<sub>**How this was verified:** Confirmed fixture definition order: restore_cwd at line 317, cleanup_shell_after at line 433. Pytest teardown order is reverse definition order, so cleanup_shell_after's teardown runs before restore_cwd's teardown. Thus any chdir in cleanup_shell_after's teardown is not undone by restore_cwd.</sub>

</details>

- `tests/conftest.py` — Adds restore_cwd autouse fixture that restores os.getcwd() after each test, and extends cleanup_shell_after to reset shell_module._workspace_cwd to None.

</details>

| `80c808e08108` | 5/5 | 0 | llm | 2026-09-17 04:26 UTC |
| `80c808e08108` | 3/5 | 1 | llm | 2026-09-17 04:40 UTC |

</details>

<sub>Reviewed `80c808e08108` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 26s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/conftest.py
```diff
@@ -314,6 +314,23 @@ def auth_headers():
...
+    restore (e.g. ``session_step.py``'s legacy fallback for tools that still use
+    ``Path.cwd()``). Under pytest-xdist a leaked cwd survives to the next test on
+    the same worker, making an unrelated test like ``test_shell_cd_chdir`` fail
+    deterministically depending on run order (gptme/gptme#3850). Defined first
+    among the autouse fixtures so its teardown runs last, undoing a chdir made
+    during another fixture's teardown too.
+    """
+    original_cwd = os.getcwd()
+    yield
+    os.chdir(original_cwd)
```
TimeToBuildBob, line 331: ❌ **P1** — The restore_cwd fixture restores the process cwd after each test, but it does not account for tests that change the cwd and then delete the directory they changed into. If a test does os.chdir(tmp_path) and then the tmp_path is cleaned up (pytest removes tmp_path after the test), the teardown os.chdir(original_cwd) will fail with FileNotFoundError because the current directory no longer exists. This would cause the fixture teardown to error, failing the test even if the test itself passed. The fixture should handle the case where the cwd has been removed, e.g., by catching OSError and falling back to a known-good directory like the original cwd's parent or os.getcwd() after the chdir fails. This is a real defect because tests that chdir into a temporary directory and then rely on pytest's tmp_path cleanup will hit this.

TimeToBuildBob: Disposing this as a false positive — the stated failure mechanism does not reproduce.

The claim is that `restore_cwd`'s teardown `os.chdir(original_cwd)` raises
`FileNotFoundError` when the test chdir'd into a directory that pytest then
deleted. Verified empirically on Linux:

```python
os.chdir(tmp); shutil.rmtree(tmp)
os.chdir(original)        # absolute, still-existing path -> SUCCEEDS
os.getcwd()               # this is what raises FileNotFoundError on a deleted cwd
```

`chdir(2)` resolves an absolute target path from the root, so a deleted *current*
working directory does not make the call fail. The `FileNotFoundError` in the
P1 belongs to `os.getcwd()` on a deleted cwd, which `restore_cwd` never calls —
its teardown chdirs before anything else reads the cwd.

The fixture's whole purpose is to move the process cwd back to a known-good,
existing path; by the time it runs, the tmp directory's deletion is irrelevant.
Adding the suggested `except OSError: os.chdir(os.path.dirname(original_cwd))`
fallback would be error handling for a case that cannot occur.

Consensus note: this finding came from a degraded review (1/3 passes answered —
2 passes failed on upstream HTTP 429 rate limiting), which is why it did not
survive manual verification.

tests/conftest.py
```diff
@@ -429,6 +446,14 @@ def cleanup_shell_after():
...
             logger.warning(f"Error closing shell during test cleanup: {e}")
         shell_module._shell_var.set(None)
+    # A leaked `_workspace_cwd` makes `ShellSession._set_cwd` skip its
+    # process-wide `os.chdir()` in the *next* test too (it only chdirs when
+    # `get_workspace_cwd() is None`), so a server test that calls
+    # `set_workspace_cwd()` silently breaks an unrelated CLI-style shell test
+    # run afterward on the same worker (gptme/gptme#3850). pytest runs tests
+    # synchronously in one thread/context, so a bare `ContextVar.set()` with
+    # no matching `.reset()` persists past the test that made it.
+    shell_module._workspace_cwd.set(None)
```
TimeToBuildBob, line 456: ❌ **P1** — The cleanup_shell_after fixture now unconditionally sets shell_module._workspace_cwd.set(None) after every test. However, this resets the ContextVar without using a token from a previous set, which is fine for isolation, but it also resets the workspace cwd for tests that legitimately set it during the test and expect it to be cleared. The real issue is that the fixture does not restore the previous value; it just sets to None. If a test sets _workspace_cwd to a non-None value and then the test itself or a later fixture reads it during teardown (e.g., a shell close that checks get_workspace_cwd()), the value is already None, which may change behavior. But more concretely, the PR's own description says the ContextVar is set with no matching .reset(), and the fix uses .set(None) which is a permanent change, not a reset. This is acceptable for test isolation, but it means that if a test's teardown (e.g., a shell session's close) relies on the workspace cwd being set, it will now see None. The PR claims this is correct, but it could break tests that expect the workspace cwd to be available during teardown. However, without a concrete failing test, this is speculative. The more concrete defect is that the fixture's teardown order relative to restore_cwd is wrong: cleanup_shell_after is defined after restore_cwd, so its teardown runs BEFORE restore_cwd's teardown. That means if cleanup_shell_after's teardown (or a shell close it triggers) does an os.chdir(), restore_cwd will not undo it. The PR's stated goal of 'undoing a chdir made during another fixture's teardown too' is not achieved because restore_cwd's teardown runs first, not last. _(2/3 passes — pass 3 looked and disagreed)_

TimeToBuildBob: Disposing as a false positive — the finding's own premise refutes its conclusion, and the concrete ordering claim is demonstrably wrong.

The finding states: *"cleanup_shell_after is defined after restore_cwd, so its teardown runs BEFORE restore_cwd's teardown"*, then concludes *"restore_cwd will not undo it"* and *"restore_cwd's teardown runs first, not last"*. Those last two contradict the first sentence, and they are the false ones: if `cleanup_shell_after` tears down **before** `restore_cwd`, then `restore_cwd` tears down **last** and *does* undo any `os.chdir()` made by `cleanup_shell_after` or anything it triggers.

Verified empirically against this repo's pytest version, two same-scope autouse fixtures defined first/second:

```txt
ORDER: ['first:setup', 'second:setup', 'test', 'second:teardown', 'first:teardown']
```

The first-defined autouse fixture tears down **last**. In `tests/conftest.py`, `restore_cwd` is defined at line 317 and `cleanup_shell_after` at line 433, so `restore_cwd`'s teardown runs after `cleanup_shell_after`'s — the docstring's stated goal ("undoing a chdir made during another fixture's teardown too") is achieved exactly as written.

The remaining concern is flagged by the finding itself as speculative ("*without a concrete failing test, this is speculative*"): `_workspace_cwd.set(None)` is the intended per-test isolation for a `ContextVar` whose default is already `None` (`gptme/tools/shell.py:1942`), and it is set at the very end of the fixture — after the shell close — so no teardown step reads a value it still needs.

PR #3860, opened 2026-09-17
test(shell): pin no-retry invariant by watching stdin writes after EPIPE

TimeToBuildBob:
## Summary

- `test_broken_pipe_partial_write_is_not_retried` had an assertion flagged twice by AI reviewer as vacuous: `assert not marker.exists() or marker.read_text() == "x"` passes trivially when the marker file doesn't exist, so it doesn't actually pin the no-retry invariant it claims to test
- New approach: after `BrokenPipeError` fires, watch all subsequent writes to `stdin_fd` for the command bytes; if the command is re-sent to the restarted shell, `state["retried"]` is set and the test fails
- Removes the timing race entirely — no `sleep()`, no marker file, no dependency on bash scheduling the command before restart
- Added two self-validating assertions (`state["partial"]` and `state["raised"]`) so a misconfigured test fails loudly rather than vacuously passing

## Test plan

- [x] `pytest tests/test_shell_restart.py::test_broken_pipe_partial_write_is_not_retried` — passes
- [x] `pytest tests/test_shell_restart.py` — all 20 tests pass


TimeToBuildBob:
## 🤖 AI code review

This PR rewrites test_broken_pipe_partial_write_is_not_retried to watch stdin writes after EPIPE instead of using a marker file, and changes the sleep duration in test_timeout_kills_grandchildren_and_term_ignoring_children from 30 to 7979 to avoid matching concurrent test workers. The new test asserts that the command bytes are not re-sent to the restarted shell's stdin after a BrokenPipeError, with self-validating assertions for the partial write and EPIPE trigger.

Confidence 5/5

<sub>ℹ️ Consensus was degraded on this run: 1 of 3 passes answered, so findings were filtered at 1-of-1 agreement rather than 2-of-3 — less filtered than usual; 1 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

<details><summary><b>1 out-of-scope finding (outside this PR's changed files)</b></summary>

_Each item here is an observation about the PR as a whole — informational only, not scored._

**⚠️ P2 medium** — _PR-level_

The new test's retry detection is defeated by the same fd-reuse problem it tries to avoid. `_is_shell_stdin` compares `fd` against `shell.process.stdin.fileno()` at the time of each write, but the restarted shell's stdin pipe is a new file descriptor. The old `stdin_fd` captured at setup is the original shell's stdin fd; after the restart, the new shell's stdin may reuse that same fd number, but the test's `write_fd` first checks `fd == stdin_fd` for the partial and raised branches. After the restart, if the new stdin fd happens to equal the old `stdin_fd`, the first write to the new shell would be misclassified as the partial write (setting `state[

</details>

- `tests/test_shell_restart.py` — Rewrites the partial-write EPIPE test to accumulate post-EPIPE stdin writes and assert no retry, and changes the sleep duration in the grandchildren timeout test to 7979.

</details>

| `212dfe95282c` | 4/5 | 1 | llm | 2026-09-17 18:01 UTC |
| `50fdc2339556` | 4/5 | 1 | llm | 2026-09-17 19:02 UTC |
| `50fdc2339556` | 5/5 | 0 | llm | 2026-09-17 19:47 UTC |
| `9cdc14b3b5a0` | 5/5 | 0 | llm | 2026-09-17 19:57 UTC |
| `07ca165a7d66` | 5/5 | 0 | llm | 2026-09-17 21:34 UTC |
| `07ca165a7d66` | 5/5 | 0 | llm | 2026-09-17 21:54 UTC |

</details>

<sub>Reviewed `07ca165a7d66` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 1208s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_shell_restart.py
```diff
@@ -171,38 +171,51 @@ def write_fd(fd, data):
...
             end = data.index(cmd_bytes) + len(cmd_bytes) + 1
             return orig_write(fd, data[:end])
+        # Second write: simulate mid-payload EPIPE.
         if fd == stdin_fd and not state["raised"]:
             state["raised"] = True
             raise BrokenPipeError
+        # After the restart: flag any write that re-sends the command bytes.
+        # The restarted shell often reuses the same stdin fd, so this check is
+        # live for all subsequent shell writes.
+        if fd == stdin_fd and command.encode() in data:
```
TimeToBuildBob: ⚠️ **P2** — The retry-detection check `if fd == stdin_fd and command.encode() in data` flags any write containing the command bytes, but the shell protocol writes the command as part of a larger payload that includes a start-marker echo and an end-marker echo. After the restart, the shell's stdin fd is reused, and the test's own subsequent `shell.run("echo ok", output=False)` writes a different command, so that doesn't trigger. However, the check also matches the command bytes appearing in the restart notice or in any other write that happens to contain the same substring. More importantly, the check is only live for writes to the same fd number; if the restarted shell gets a different fd number, the retry would go undetected. The test's claim to pin the no-retry invariant is therefore conditional on fd reuse, which the comment acknowledges but does not guard against. If the restarted shell does not reuse the fd, the test passes vacuously even if the command is re-sent, because the retry write goes to a different fd and `state["retried"]` stays False. This is a test-defect: the assertion can pass while the invariant is violated.

tests/test_shell_restart.py
```diff
@@ -171,38 +171,51 @@ def write_fd(fd, data):
...
+            state["retried"] = True
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 50fdc2339. The retry watchdog now compares each write against the shell's *current* stdin descriptor (`_is_shell_stdin`, reading `shell.process.stdin.fileno()` at write time) rather than the fd captured before the restart, so a retry routed to a new descriptor is still caught. Also verified the assertion isn't vacuous: injecting the retry bug into `_run_pipe` (retry on partial EPIPE) makes this test fail with `command was re-sent to the restarted shell (retry bug)`.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI re-run triggered on run 35258279089. The 3 failing tests (`test_shell_death_restarts_and_restores_state[kill -- -$$--15]`, `test_cli_sigterm_cleans_background_job[terminate]`, `test_cli_sigterm_cleans_background_job[default-return]`) are timing-sensitive signal/process-group tests unrelated to this PR's changes (which only touch `test_broken_pipe_partial_write_is_not_retried`). The first PR commit passed CI cleanly; this appears to be an xdist flake.

tests/test_shell_restart.py
TimeToBuildBob: False positive — the test is correct as written.

`_is_shell_stdin(fd)` evaluates `shell.process.stdin.fileno()` **at call time**, not against the `stdin_fd` captured at test setup. After restart, `shell.process.stdin` is the new stdin pipe for the restarted shell. Any retry from gptme's shell code would write to that same `shell.process.stdin`, so `_is_shell_stdin` correctly matches it regardless of which fd number the OS assigned to the new pipe.

The reviewer's concern about "a write to a different fd that is not the current stdin" doesn't map to realistic retry behaviour: gptme's shell runner writes commands through `shell.process.stdin` (that's the only channel it has to the shell). The first two writes in the test were intercepted through exactly this `os.write` + stdin-fd path, confirming it's the only path. A retry on a completely different fd would not be communicating with the shell and would not constitute the "retry bug" this test is guarding against.

The comment in the code already explains the design: "compare against the *current* stdin rather than the fd captured at setup" — this is the correct invariant for detecting retries after fd reassignment.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

Issue #3811, opened 2026-09-11
Agents don't know about /preview/{port}/ — surface capability via prompt layering + cloud config overlay


ErikBjare:
#3767 added the `/preview/{port}/` proxy (HTTP+WS to 127.0.0.1:{port}; noVNC at /preview/6080/vnc.html) and it's verified live in gptme.ai prod pods — but nothing tells the agent it exists. Erik: "an agent can now start Vite on a port but does it know it can?" It doesn't: no prompt fragment, no env signal.

**What the agent must learn** (short fragment, cloud-gated):
- Bind to 127.0.0.1:{port} (1024-65535, not 5700/5900) → user-reachable at `/preview/{port}/` relative to the instance server URL
- Dev servers with absolute asset paths need a base path (`vite --base=/preview/{port}/`)
- WebSockets work (HMR; noVNC desktop at /preview/6080/vnc.html)
- Surface URLs via the webui's actual live_app/preview panel mechanism (read the webui code for the exact convention — don't invent one)

**Design constraints (Erik):**
1. Use the EXISTING prompt-layering mechanism for interface/deployment-specific injection, not a new ad-hoc env gate. Leads: named prompt sections in `gptme/prompts/__init__.py` (`_build_prompt_sections`); server-side conversation prompt layering in `gptme/server/api_v2.py` (~316-362, ChatConfig.system_prompt → profile fallback); check how existing webui/server features got prompt-surfaced and follow that pattern.
2. Cloud-injected config must NOT go in the user's `config.toml` (users legitimately edit it; clobbering both directions; failure modes must be independent). Follow the `config.local.toml` overlay pattern — likely a dedicated cloud-owned overlay file the pod startup script rewrites idempotently each boot, with documented precedence. A general overlay mechanism in core is fine; a cloud-special hack is not.

Cloud-side counterpart (env/startup-script plumbing) tracked in gptme-cloud#910. Related: gptme-cloud#924/#926 (Phase 2 wildcard subdomains — infra, separate concern).

gargsajal9:
I am preparing a core-only draft PR for this issue. The implementation adds an optional operator-owned `config.runtime.toml` defaults layer (runtime < main < local), additive named `[prompt.fragments]` through the existing prompt builder, and read-only configuration diagnostics. Preview guidance uses the existing `Message.metadata.panel_hints` / `live_app` convention, without replacing user preferences or profiles.

[...]

PR #3861, opened 2026-09-17
fix(webui): resolve server-relative panel srcs against the instance API origin

TimeToBuildBob:
Part of gptme/gptme#3811. `config.runtime.toml` + `[prompt.fragments]` (#3821) landed the core mechanism; the fragment still tells the agent to declare a `live_app` panel at `/preview/{port}/`, and on gptme.ai that URL loads from the SPA origin and 404s.

## Problem

`PanelsPanel` passes `selected.url` / `descriptor.src` straight into the iframe. For a server-relative value like `/preview/5173/` that resolves against the *page* origin. On deployments where the SPA and the instance API are different origins (SPA `gptme.ai`, instance `fleet.gptme.ai/...`), the panel points at the wrong host. `urljoin(base, "/preview/5173/")` is the other footgun — an absolute path replaces the instance prefix, so the join has to be naive concatenation on a slash-normalized base.

Separately, the iframe allowlist only accepts localhost and server-relative srcs, so a resolved instance-API URL would be rejected as "Panel blocked".

## Fix

- `resolvePanelSrc(src, baseUrl)` joins server-relative srcs onto the API base, preserving any instance path prefix, and leaves absolute / protocol-relative / malformed values for the allowlist to judge.
- `isAllowedIframeSrc(src, allowedOrigin?)` now accepts exactly one additional origin, supplied by the caller. This is not a public-internet hole: the caller passes the connected server's origin, and only an exact origin match is allowed (scheme included).
- `SandboxedIframePanel` takes an optional `apiBaseUrl` prop and resolves both `src` and the expected `postMessage` origin against it. Omitted → current behavior (relative src, page origin) unchanged for local dev.
- `PanelsPanel` passes `connectionConfig.baseUrl` down, covering both the `live_app` and `iframe` render paths.

## Tests

16 new assertions across the two existing suites (36 pass):

- join semantics: prefix preservation, trailing-slash tolerance, absolute/protocol-relative passthrough, no-op without a base
- allowlist: API origin accepted, other origins still rejected (including `fleet.gptme.ai.evil.com`, scheme mismatch, malformed/empty allowedOrigin)
- component: relative src resolves to the instance URL; renders unresolved without `apiBaseUrl`; `postMessage` accepted from the resolved origin; foreign origin still blocked

`npm run typecheck`, `eslint`, and `prettier --check src/**/*.{ts,tsx}` clean.

## Not in this PR

- **Cloud startup writing `config.runtime.toml`** — the deployment half, tracked in gptme/gptme-cloud#910. It should land after this rewrite, otherwise the fragment teaches a URL the iframe cannot load.
- **VNC path** — `ComputerPreview.buildVncUrl()` already resolves `/preview/6080/vnc.html` against `baseUrl`, so no change was needed there.
- **Wildcard `{port}-{id}.preview...` origins** (gptme/gptme-cloud#924) — deliberately not touched; this change allows exactly the instance origin, not a preview wildcard.


webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -13,19 +13,31 @@ import { useEffect, useRef, useState } from 'react';
...
   const [autoHeight, setAutoHeight] = useState<number | null>(null);
 
-  const allowed = isAllowedIframeSrc(descriptor.src);
-  const expectedOrigin = allowed ? iframeSrcOrigin(descriptor.src) : null;
+  // Server-relative srcs (e.g. "/preview/5173/") belong to the instance server,
+  // which may be a different origin than the page rendering this panel.
+  const src = resolvePanelSrc(descriptor.src, apiBaseUrl);
+  const apiOrigin = urlOrigin(apiBaseUrl);
+  const allowed = isAllowedIframeSrc(src, apiOrigin);
+  const expectedOrigin = allowed ? iframeSrcOrigin(src, apiOrigin ?? undefined) : null;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 44e4fd63f.

You're right, and the codebase had already reached the same conclusion — inverted — on the server side: `panels_api._build_iframe_panel` warns that a server-relative src with `allow-scripts` "has opaque origin; postMessage bootstrap handshake will not work". `resolveSandbox` drops `allow-same-origin` whenever `allow-scripts` is requested, so *every* scripted panel is opaque. The handshake was dead for all of them, not just for split-origin deployments.

Changes in 44e4fd63f:

- `sandboxHasOpaqueOrigin()` added to `iframePanelPolicy.ts`: true when the resolved sandbox is non-empty and omits `allow-same-origin`.
- `SandboxedIframePanel` accepts `event.origin === "null"` for those frames and replies with `targetOrigin: "*"`. A concrete `targetOrigin` can never match an opaque origin, so identity is pinned by the already-present `event.source === contentWindow` check — that check is the authentication for opaque frames, and it runs before the origin gate.
- Frames that keep their origin (no sandbox, or `allow-same-origin` without scripts) still require an exact origin match, and a `null` origin from those is rejected. The fail-closed test is retained, re-anchored to the non-sandboxed case.
- Reworded the server-side warning to describe the real limitation (identity-pinned messages, no cookies/storage) instead of claiming the handshake cannot work.
- Tests now emit `origin: "null"` for the sandboxed cases. As you noted, `https://fleet.gptme.ai` is not an origin that frame can produce, so the previous assertion was unfalsifiable against a real browser.

Verification: 40/40 tests in the two affected suites, `npm run typecheck`, eslint, and prettier all clean.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 2 findings disposed (fixed in 0b179a3f4; fixed in 0936df6ec)

| --- | --- | --- |
| `webui/src/components/SandboxedIframePanel.tsx:40` | P1 | superseded by latest review (not reproduced) |
| `webui/src/components/SandboxedIframePanel.tsx:72` | P1 | fixed in `0b179a3f4` |
| `webui/src/components/SandboxedIframePanel.tsx:56` | P1 | fixed in `ebd4f2efa` |
| `webui/src/components/SandboxedIframePanel.tsx:100` | P1 | fixed in `0936df6ec` |
| `webui/src/components/__tests__/SandboxedIframePanel.test.tsx:216` | P1 | superseded by latest review (not reproduced) |
| `gptme/server/panels_api.py:211` | P2 | fixed in `05a759c13` |
| `webui/src/components/SandboxedIframePanel.tsx:74` | P2 | rejected |
| `webui/src/components/__tests__/SandboxedIframePanel.test.tsx:80` | P2 | superseded by latest review (not reproduced) |
| `webui/src/components/__tests__/SandboxedIframePanel.test.tsx:94` | P2 | rejected |
| `webui/src/components/__tests__/SandboxedIframePanel.test.tsx:63` | P2 | fixed in `0936df6ec` |
| `webui/src/utils/iframePanelPolicy.ts:52` | P2 | superseded by latest review (not reproduced) |
| `webui/src/utils/iframePanelPolicy.ts:52` | P2 | rejected |
| `webui/src/utils/iframePanelPolicy.ts:147` | P2 | fixed in `0936df6ec` |
| `webui/src/utils/iframePanelPolicy.ts:52` | P2 | fixed in `0936df6ec` |

</details>




This PR makes server-relative iframe panel srcs resolve against the instance API base URL instead of the SPA origin, and extends the iframe allowlist to accept the instance API origin. It adds a resolvePanelSrc helper with path-traversal guards, updates SandboxedIframePanel to use the resolved src and to handle opaque-origin postMessage identity via event.source, and broadens the server-side opaque-origin warning to all allow-scripts panels. Tests are added for the new resolution, allowlist, and bootstrap-once-per-document behavior.

Confidence 3/5

<b>2 findings</b> · ❌ 1 P1 · ⚠️ 1 P2

<details><summary><b>1 inline</b></summary>

**⚠️ P2 medium** — `webui/src/components/SandboxedIframePanel.tsx:74`

The bootstrap-once-per-document guard is re-armed on every src change, but the effect also re-runs when conversationId changes. The guard is intentionally NOT reset on conversationId changes, yet the effect's cleanup and re-run re-registers the message listener. If a conversationId change happens after the frame has loaded and posted ready, the new listener captures the same bootstrappedRef (still true), so a duplicate ready is ignored. However, the effect re-run also re-evaluates `src` and `bootstrappedSrcRef`; since src is unchanged, the guard is not reset. This is correct. But consider the case where the src changes to a new value and then back to the original value. The guard is reset on the first change, and the new document loads. If the src changes back to the original before the new document posts ready, the guard is reset again, allowing a second bootstrap for the same original document? Actually the src change triggers a navigation, so the document is new. The guard reset on src change is correct. The potential issue is that `handleLoad` resets the guard on every load event, including the initial load. If the iframe fires a load event after the bootstrap has already been sent (e.g., due to a subresource load? No, load fires for the document). The load event fires once per document. So the guard is reset exactly once per document. But what if the iframe fires a load event after the ready message? The ready message is posted after the document loads, so the load event fires before the ready. The guard is reset on load, then ready arrives and bootstrap is sent. If the document posts ready twice, the second is ignored. If the document reloads, a new load event resets the guard. This seems correct. However, there is a race: if the ready message arrives before the load event (as the code comments mention), the guard is reset by the src-change effect, so it works. But if the src changes and the new document posts ready before the effect runs? The effect runs synchronously after render, so it resets the guard before the iframe loads. So no issue. I need to find a real bug.

<sub>**How this was verified:** Traced the effect and guard logic; the guard is reset on src change and on load, and not on conversationId change. The tests cover these cases.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`webui/src/components/SandboxedIframePanel.tsx:1`](https://github.com/gptme/gptme/blob/2b18f736582f6003b1c7f0336bdd9518700890a9/webui/src/components/SandboxedIframePanel.tsx#L1)

This PR modifies visual/UI files (`webui/src/components/SandboxedIframePanel.tsx`) but the PR body contains no screenshot or before/after image. Erik's feedback: 'I merged it, but it's STILL gray bg!' (gptme#3321) and 'Add a before/after screenshot' (gptme#3334, #3461). Attach a screenshot showing the rendered change.

```suggestion
Add a Markdown image to the PR body: `![Before](url) ![After](url)`
```

- `gptme/server/panels_api.py` — Changes the opaque-origin warning condition from server-relative srcs to any panel with allow-scripts, and updates the warning text.
- `tests/test_server_panels.py` — Renames the localhost warning test and updates assertions to expect the opaque-origin warning for localhost srcs with allow-scripts.
- `webui/src/components/PanelsPanel.tsx` — Adds apiBaseUrl from useApi and passes it to SandboxedIframePanel, and adds key={conversationId} to both iframe render paths.
- `webui/src/components/SandboxedIframePanel.tsx` — Resolves src against apiBaseUrl, adds opaque-origin handling for postMessage, and adds a bootstrap-once-per-document guard with load-based re-arming.
- `webui/src/components/__tests__/SandboxedIframePanel.test.tsx` — Updates existing tests for opaque-origin behavior and adds tests for src resolution, origin gating, and bootstrap-once-per-document semantics.
- `webui/src/utils/__tests__/iframePanelPolicy.test.ts` — Adds tests for resolvePanelSrc, urlOrigin, and sandboxHasOpaqueOrigin, and updates isAllowedIframeSrc tests for the allowedOrigin parameter.
- `webui/src/utils/iframePanelPolicy.ts` — Adds urlOrigin, resolvePanelSrc, and sandboxHasOpaqueOrigin; extends isAllowedIframeSrc to accept an allowedOrigin parameter.

</details>

| `d56b5f25353e` | 2/5 | 4 | llm | 2026-09-17 18:43 UTC |
| `44e4fd63f991` | 2/5 | 5 | llm | 2026-09-17 19:04 UTC |
| `05a759c1369e` | 2/5 | 4 | llm | 2026-09-17 20:20 UTC |
| `cafd300bda1f` | 2/5 | 5 | llm | 2026-09-17 21:34 UTC |

</details>

<sub>Reviewed `2b18f736582f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 66s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/src/components/SandboxedIframePanel.tsx
TimeToBuildBob, line 49: ❌ **P1** — In SandboxedIframePanel, the expectedOrigin for postMessage is computed as iframeSrcOrigin(src, apiOrigin ?? undefined). When apiBaseUrl is provided and src is server-relative, src is resolved to an absolute URL on the API origin, so iframeSrcOrigin returns that origin. But when apiBaseUrl is not provided and src is server-relative, iframeSrcOrigin falls back to window.location.origin (the SPA origin). The iframe's actual content is loaded from the SPA origin in that case, so the origin gate matches. However, when apiBaseUrl is provided but src is an absolute localhost URL (e.g. http://localhost:5173), resolvePanelSrc leaves it unchanged, apiOrigin is the API origin, and iframeSrcOrigin returns http://localhost:5173. The allowlist accepts localhost, and postMessage is gated to localhost. That is correct. The defect: when apiBaseUrl is provided and src is a server-relative path that resolves to the API origin, the iframe loads from the API origin, but the sandbox attribute may include allow-scripts without allow-same-origin, making the iframe's origin opaque ('null'). The postMessage origin gate expects the API origin, but the iframe's contentWindow.postMessage targetOrigin is set to expectedOrigin, which is fine for sending, but incoming messages from the iframe will have event.origin === 'null' (opaque origin) and will be rejected. This means the bootstrap handshake never completes for server-relative panels on a cross-origin deployment, because the sandboxed iframe without allow-same-origin gets an opaque origin. The existing test 'accepts a postMessage from the resolved API origin' emits from 'https://fleet.gptme.ai' directly, which does not reflect real browser behavior for a sandboxed iframe without allow-same-origin. In a real browser, the iframe's origin would be 'null', so the host would reject the ready message and never send bootstrap. This breaks the live_app preview on gptme.ai, which is the exact scenario this PR aims to fix.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
```diff
@@ -174,4 +174,63 @@ describe('SandboxedIframePanel', () => {
...
+        descriptor={{ ...baseDescriptor, src: '/preview/5173/', title: 'Live App' }}
+        conversationId="conv1"
+        apiBaseUrl="https://fleet.gptme.ai/api/v1/instances/abc"
+      />
+    );
+    const frame = screen.getByTitle('Live App') as HTMLIFrameElement;
+    const postMessage = jest.fn();
+    Object.defineProperty(frame, 'contentWindow', { value: { postMessage }, configurable: true });
+
+    emitFromIframe(frame, 'https://fleet.gptme.ai', { type: 'gptme:ready' });
```
TimeToBuildBob, line 268: ❌ **P1** — The new test 'accepts a postMessage from the resolved API origin' stubs frame.contentWindow with a postMessage jest.fn and dispatches a MessageEvent with origin 'https://fleet.gptme.ai'. In a real browser, a sandboxed iframe without allow-same-origin (which resolveSandbox guarantees when allow-scripts is present) has an opaque origin, so the ready message would arrive with event.origin === 'null', not the API origin. The test therefore passes vacuously for the real handshake path: it asserts the component accepts a message from the API origin, but the component would never receive such a message from the actual iframe. This is a test-defect because it stubs the wrong boundary — it simulates a non-sandboxed iframe origin. The test should emit from 'null' to match the opaque origin, and then the component would correctly reject it, exposing the bootstrap failure described in the component finding. This test gives false confidence that the cross-origin deployment works.

webui/src/utils/iframePanelPolicy.ts
```diff
@@ -14,16 +14,59 @@ const SANDBOX_ALLOWLIST: ReadonlySet<string> = new Set<IframeSandboxToken>([
...
+  if (typeof src !== 'string') return '';
+  const value = src.trim();
+  if (!baseUrl) return value;
+  // "//host" is protocol-relative and "/\host" is a backslash variant; both are
+  // absolute for policy purposes, so leave them for the allowlist to judge.
+  if (!value.startsWith('/') || value.startsWith('//') || value.startsWith('/\\')) return value;
+
+  const base = baseUrl.trim().replace(/\/+$/, '');
+  if (!base) return value;
+  return `${base}/${value.replace(/^\/+/, '')}`;
```
TimeToBuildBob, line 101: ⚠️ **P2** — resolvePanelSrc rewrites every src that begins with a single '/', not just /preview/ paths. Because PanelsPanel now always supplies connectionConfig.baseUrl (lines 205 and 219), any existing server-relative panel src that was previously relative to the page/SPA origin is now prefixed with the API base path. For example, a descriptor with src '/assets/dashboard.html' that the SPA serves at the page origin will be loaded as '{apiBase}/assets/dashboard.html' and 404, since the API base typically includes /api/v1/instances/{id}. The previous contract of isAllowedIframeSrc accepted any single-slash path and the iframe used the raw src, so page-relative panel srcs worked; this PR silently changes that contract for all server-relative paths without a path filter or migration.

webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -7,40 +7,72 @@
...
-      // Strict origin gate: only accept messages from the declared src origin.
-      // Fail closed: if expectedOrigin is null (origin unresolvable), reject all.
-      if (!expectedOrigin || event.origin !== expectedOrigin) return;
+      // Identity gate: only this panel's own frame may drive the protocol.
       if (event.source !== iframeRef.current?.contentWindow) return;
+      // Origin gate. An opaque-origin frame can only ever speak as "null", and
+      // the source check above is what authenticates it. Everything else must
+      // match the declared src origin; fail closed when it is unresolvable.
+      if (opaqueOrigin) {
+        if (event.origin !== 'null') return;
```
TimeToBuildBob, line 95: ❌ **P1** — The opaque-origin postMessage gate accepts any message whose event.origin is the string 'null' as long as event.source equals the iframe's contentWindow. A malicious page that can cause the iframe to navigate to a different document (e.g. via a link or script inside the sandboxed frame, or by the parent setting src to a data: URL) would still have event.source === contentWindow, and the new contentWindow's origin would be opaque 'null'. The origin check is therefore not actually authenticating the frame's origin; it only checks the serialized string 'null', which any opaque-origin frame (including a cross-origin one loaded into the same iframe element) will report. The comment claims identity is pinned by event.source, but event.source is the same for any document loaded in that iframe element, so a panel that navigates itself to an attacker-controlled opaque-origin document can still drive the bootstrap handshake and receive the conversation_id bootstrap payload. This weakens the fail-closed property the previous code had, which rejected all 'null' origins.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
```diff
@@ -55,6 +77,24 @@ describe('SandboxedIframePanel', () => {
     );
   });
 
+  it('rejects a "null" origin from a frame that is not sandboxed (fail-closed)', async () => {
```
TimeToBuildBob, line 80: ⚠️ **P2** — The new test 'rejects a "null" origin from a frame that is not sandboxed (fail-closed)' uses a descriptor with sandbox: [] and emits a message with origin 'null'. The component computes opaqueOrigin = sandboxHasOpaqueOrigin([]) which is false, so it goes to the else branch and checks expectedOrigin. expectedOrigin is iframeSrcOrigin('http://localhost:8080', undefined) which in jsdom resolves to 'http://localhost' (window origin) because the src is absolute localhost:8080, so expectedOrigin is 'http://localhost:8080'. The event.origin 'null' does not match, so the message is rejected. This test passes, but it does not actually test the fail-closed behavior for an opaque-origin frame that is not sandboxed; it tests that a non-sandboxed frame with a mismatched origin is rejected. The more important case is a sandboxed frame (opaqueOrigin true) that emits from a non-null origin, which is already covered by the 'ignores messages from a foreign origin' test (which uses baseDescriptor with allow-scripts and emits from 'https://evil.example.com'). That test passes because opaqueOrigin is true and event.origin !== 'null'. So the new test is somewhat redundant and does not add coverage for a distinct failure mode. However, it is not vacuous; it does assert a real behavior. The test name is slightly misleading but the assertion is valid.

gptme/server/panels_api.py
```diff
@@ -210,9 +210,10 @@ def _build_iframe_panel(
     stripped_src = str(src).strip()
     if stripped_src.startswith("/") and "allow-scripts" in sandbox:
```
TimeToBuildBob: ⚠️ **P2** — The server-side warning in panels_api.py now claims that for server-relative srcs with allow-scripts, 'postMessage traffic is authenticated by frame identity rather than origin'. However, the server does not enforce or verify any frame identity; it only emits a warning string. The actual authentication is implemented in the webui's SandboxedIframePanel. The warning text is informational and does not affect behavior, but it is misleading because it states as fact that frame identity is used, which is only true if the webui is updated to this version. If an older webui is used with a newer server, the warning would be inaccurate. This is a documentation/contract issue rather than a runtime bug, but it could mislead users about the security properties. The warning is also only added for server-relative srcs, but the same opaque-origin situation applies to any src that is not localhost? Actually localhost srcs with allow-scripts also get opaque origin because resolveSandbox drops allow-same-origin when allow-scripts is present. The warning condition is `stripped_src.startswith("/") and "allow-scripts" in sandbox`, so it misses localhost srcs that also have allow-scripts and thus also have opaque origin. This is an inconsistency: the warning should apply to any scripted panel, not just server-relative ones. The webui's sandboxHasOpaqueOrigin correctly returns true for any sandbox without allow-same-origin, including localhost srcs. So the server warning under-reports the opaque-origin condition.

webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -7,40 +7,72 @@
...
+      } else if (!expectedOrigin || event.origin !== expectedOrigin) {
+        return;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0b179a3f4.

Added a `bootstrappedRef` that's reset each effect run (i.e. each time `src` or deps change). Once the bootstrap fires, subsequent `gptme:ready` messages are silently ignored — a navigated document in the same iframe shares the same `contentWindow` but can no longer receive `conversation_id`.

Regression test: 'does not re-bootstrap after gptme:ready fires a second time (bootstrap-once guard)' passes alongside the rest of the suite.

TimeToBuildBob:
@greptileai review

webui/src/components/SandboxedIframePanel.tsx
TimeToBuildBob: Fixed in 0b179a3f4. Added bootstrappedRef (reset each effect run) that lets the bootstrap fire only once per iframe lifecycle. A navigated document in the same iframe shares the same contentWindow but subsequent gptme:ready messages are now silently dropped. Regression test added and passing.

gptme/server/panels_api.py
TimeToBuildBob: Fixed in 05a759c13. Removed the `stripped_src.startsWith("/")` guard so the opaque-origin warning fires for any `allow-scripts` panel (localhost absolute srcs included).

TimeToBuildBob:
@greptileai review

gptme/server/panels_api.py
```diff
@@ -208,11 +208,12 @@ def _build_iframe_panel(
 
     warnings: list[str] = []
     stripped_src = str(src).strip()
-    if stripped_src.startswith("/") and "allow-scripts" in sandbox:
+    if "allow-scripts" in sandbox:
```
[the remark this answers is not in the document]
TimeToBuildBob: False positive — `stripped_src` is used in the `return IframePanelOut(src=stripped_src, ...)` on line 229, so it is not unused. Ruff passes cleanly on the current head (`ruff check gptme/server/panels_api.py` → "All checks passed!"). No code change needed.

webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -7,50 +7,92 @@
...
+  const expectedOrigin = allowed ? iframeSrcOrigin(src, apiOrigin ?? undefined) : null;
+  // A sandbox without `allow-same-origin` gives the frame an opaque origin: its
+  // messages arrive as `event.origin === "null"` and no concrete `targetOrigin`
+  // can address it. `resolveSandbox` drops `allow-same-origin` whenever
+  // `allow-scripts` is requested, so every scripted panel is opaque, and
+  // `event.source` is the only identity gate available.
+  const opaqueOrigin = sandboxHasOpaqueOrigin(descriptor.sandbox);
 
   useEffect(() => {
+    bootstrappedRef.current = false;
```
TimeToBuildBob, line 63: ❌ **P1** — The bootstrap-once guard uses a ref that is reset to false on every effect re-run (line 56: bootstrappedRef.current = false). The effect depends on allowed, expectedOrigin, opaqueOrigin, conversationId, descriptor.bootstrap, and descriptor.resize. If any of these change after the iframe has already loaded and sent gptme:ready, the effect re-runs, resets the guard, and re-registers the message listener. A subsequent gptme:ready from the same frame (e.g., after a navigation) will then pass the guard and send the bootstrap payload again, defeating the navigation-bypass protection the guard is meant to provide. The test only checks a second ready within the same effect instance, not across a re-render that changes a dependency. The consequence is that an attacker-controlled page that navigates the iframe after a dependency change can still receive conversation_id via a second bootstrap.

webui/src/utils/iframePanelPolicy.ts
TimeToBuildBob, line 101: ⚠️ **P2** — resolvePanelSrc treats any value starting with a single '/' as server-relative and joins it onto the base URL, but it does not account for the case where the base URL itself contains a path prefix that is not the instance prefix. The PR description says the join must preserve the instance path prefix, but the implementation simply concatenates base + '/' + src. If the base URL is something like 'https://fleet.gptme.ai/api/v1/instances/abc' and the src is '/preview/5173/', the result is 'https://fleet.gptme.ai/api/v1/instances/abc/preview/5173/', which is correct. However, if the src is a server-relative path that is meant to be relative to the instance root (e.g., '/other/app'), the join produces 'https://fleet.gptme.ai/api/v1/instances/abc/other/app', which may not be the intended URL. The allowlist test at line 46 explicitly accepts 'https://fleet.gptme.ai/other/app' as same-origin, but resolvePanelSrc would never produce that from a server-relative src; it would produce the prefixed version. This is a mismatch between the documented behavior (preserving the prefix) and the allowlist's acceptance of any path on the same origin. The consequence is that a server-relative src like '/other/app' resolves to a URL under the instance prefix, which may 404 if the server only serves /preview/ under that prefix. This is a correctness issue for non-preview server-relative paths.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
```diff
@@ -55,6 +77,24 @@ describe('SandboxedIframePanel', () => {
...
+      />
+    );
+    const frame = getIframe();
+    const postMessage = jest.fn();
+    Object.defineProperty(frame, 'contentWindow', { value: { postMessage }, configurable: true });
+
+    // A frame with a real origin cannot legitimately claim to be opaque.
+    emitFromIframe(frame, 'null', { type: 'gptme:ready' });
+
+    await new Promise((r) => setTimeout(r, 10));
```
TimeToBuildBob, line 96: ⚠️ **P2** — The test 'rejects a "null" origin from a frame that is not sandboxed (fail-closed)' at line 80-96 uses a 10ms setTimeout to wait for the message handler to run, then asserts postMessage was not called. This is a non-hermetic timing-based assertion: if the event loop is slow, the handler may not have run yet, and the assertion passes vacuously. The test should use waitFor with a negative assertion or a more deterministic mechanism. However, the test is checking that a message is NOT sent, which is hard to assert with waitFor. The existing pattern in the file uses setTimeout for negative assertions, so this is consistent. But it is still a test defect because it can pass even if the handler would have sent a message after the timeout. The PR adds this test, so it is a defect in the diff.

webui/src/components/SandboxedIframePanel.tsx
TimeToBuildBob: Fixed in ebd4f2efa. Added `bootstrappedSrcRef` to track the src at which the guard was last reset, and changed the effect to only reset `bootstrappedRef.current` when `src` actually changes — not on every effect re-run. Also added `src` to the dep array so the guard correctly resets when the iframe navigates to a new URL. New regression test: "bootstrap guard persists after a conversationId re-render (dep change without src change)".

webui/src/utils/iframePanelPolicy.ts
TimeToBuildBob: The described behavior is intentional: `resolvePanelSrc` joins server-relative srcs onto the instance API base URL (preserving the instance path prefix). Paths like `/preview/5173/` resolve to `https://fleet.gptme.ai/api/.../instances/abc/preview/5173/` — which is exactly the intent since the preview server runs under that instance prefix. A path meant to be "relative to the origin root" is not a supported use-case for this API; callers wanting root-relative resolution should pass an absolute URL instead. The allowlist test at line 46 accepting any path on the API origin is a separate access-control decision, not a mismatch with resolution semantics.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
TimeToBuildBob: This is a negative assertion (checking that postMessage is NOT called). For negative assertions, `waitFor` with a positive inverse assertion is the standard pattern, but when the handler is synchronous the 10ms timeout is functionally hermetic — the event handler runs synchronously in JSDOM. This is consistent with the existing pattern in this test file. Accepted as a minor test-style trade-off.

webui/src/components/SandboxedIframePanel.tsx
TimeToBuildBob: Fixed in ebd4f2efa — `bootstrappedSrcRef` now tracks the src at which the guard was last reset, and the guard only resets when `src` changes (not on every effect re-run). New regression test confirms guard persists across `conversationId` re-renders. All 17 SandboxedIframePanel tests pass.

TimeToBuildBob:
## PM status: ready for human review

All automated work complete. PR requires human review due to the server path gate (`gptme/server/panels_api.py`).

### Changes pushed this session

**`ebd4f2efa` — fix(webui): guard bootstrap-once by src, not by effect lifecycle**

The bootstrap guard was resetting on every `useEffect` re-run. That meant a `conversationId` change (or any other dep change) would reset the guard while the same iframe document was still loaded — so a navigated attacker-controlled page sharing the same `contentWindow` and opaque `"null"` origin would pass both identity checks and receive a second `gptme:bootstrap` payload containing `conversation_id`.

Fix: track `bootstrappedSrcRef` (the src at last reset). Guard only resets when `src` actually changes. Regression test added.

**`cafd300bd` — fix(webui): key SandboxedIframePanel by conversationId to force remount on switch**

The src-only guard creates a functional gap: switching conversations leaves the old iframe mounted (same panel src, different conversation). Adding `key={conversationId}` to both `SandboxedIframePanel` usages in `PanelsPanel` forces a fresh mount on switch, so each conversation gets its own `contentWindow` and clean bootstrap state. This eliminates the security/functionality trade-off entirely.

### Review status

- **Greptile 4/5** — addressed the only real finding (`stripped_src` "unused variable" was a false positive; `ruff` confirms it is used on line 229 of `panels_api.py`). Score held at 4/5.
- **AI review P1s** — all resolved/disposed. Bootstrap guard fix was the P1; PanelsPanel key addresses the functional gap it introduced.
- **CI** — passing (e2e still running, no failures).
- **Server path gate** — `gptme/server/panels_api.py` touches the server/ path cluster; self-merge blocked per `auto-merge-analysis §3 cluster 3`.

TimeToBuildBob:
@greptileai review

webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -7,50 +7,105 @@
...
-            // Spread descriptor.bootstrap first so the prop-supplied
-            // conversationId always wins over any key in the bootstrap blob.
-            payload: { ...(descriptor.bootstrap ?? {}), conversation_id: conversationId },
-          });
+          // Bootstrap-once guard: a navigated document in the same iframe
+          // shares the same contentWindow (WindowProxy) and an opaque "null"
+          // origin, so it would otherwise pass both identity checks. Sending
+          // the bootstrap payload only once prevents a navigated attacker-
+          // controlled page from receiving conversation_id via a second ready.
+          if (!bootstrappedRef.current) {
```
TimeToBuildBob, line 108: ❌ **P1** — The bootstrap-once guard is reset only when `src` changes. A panel that navigates to a new document within the same iframe (e.g., `location.reload()` or an internal link) keeps the same `src` attribute, so the new document's `gptme:ready` passes both the `event.source === contentWindow` and opaque-`'null'`-origin checks but is suppressed by `bootstrappedRef`. The replacement document never receives `conversation_id` or the `descriptor.bootstrap` payload, breaking panels that legitimately reload or multi-document flows. The test 'does not re-bootstrap after gptme:ready fires a second time' enshrines this behavior without distinguishing attacker navigation from a legitimate reload.

webui/src/utils/iframePanelPolicy.ts
```diff
@@ -81,3 +129,20 @@ export function resolveSandbox(tokens: readonly string[] | undefined): string {
...
+ * Browsers serialise an opaque origin as `"null"` on `postMessage`, so the
+ * host can neither match `event.origin` against a concrete origin nor address
+ * the frame with a concrete `targetOrigin`. Identity must be pinned by
+ * `event.source` instead. Since `resolveSandbox` drops `allow-same-origin`
+ * whenever `allow-scripts` is requested (the dangerous pairing), any sandboxed
+ * panel that can script is opaque.
+ */
+export function sandboxHasOpaqueOrigin(tokens: readonly string[] | undefined): boolean {
+  const sandbox = resolveSandbox(tokens);
+  return sandbox !== '' && !sandbox.split(' ').includes('allow-same-origin');
```
TimeToBuildBob: ⚠️ **P2** — sandboxHasOpaqueOrigin returns `sandbox !== '' && ...`, so it classifies an empty resolved sandbox as non-opaque. But `sandbox={resolveSandbox(descriptor.sandbox)}` (SandboxedIframePanel.tsx:158) renders a present-but-empty `sandbox=""` attribute for any descriptor whose tokens resolve to the empty set — descriptor.sandbox = [], a list of only never-allowed tokens, or a missing sandbox the server already defaulted to []. Per the HTML sandbox spec an empty attribute still activates full sandboxing, so the frame has an opaque 'null' origin while `opaqueOrigin` is false. The origin gate then takes the `else if (!expectedOrigin || event.origin !== expectedOrigin)` branch and rejects the frame's ready as a foreign origin, and the reply target is the concrete expectedOrigin, which cannot address an opaque frame. The break is currently latent because such frames are script-less and cannot post ready, but the classification is wrong and is the exact mechanism that lets the new sandbox:[] tests pass on fabricated origins.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
```diff
@@ -46,6 +46,28 @@ describe('SandboxedIframePanel', () => {
...
+    expect(postMessage).toHaveBeenCalledWith(
+      { type: 'gptme:bootstrap', payload: { conversation_id: 'conv-abc' } },
+      '*'
+    );
+  });
+
+  it('replies to the concrete src origin when the sandbox keeps the frame origin', async () => {
+    render(
+      <SandboxedIframePanel
+        descriptor={{ ...baseDescriptor, sandbox: [] }}
```
TimeToBuildBob: ⚠️ **P2** — resolveSandbox([]) returns '', which the component renders as `sandbox=""` — a present-but-empty attribute that fully sandboxes the frame and gives it an opaque 'null' origin. The new test 'replies to the concrete src origin when the sandbox keeps the frame origin' renders `sandbox: []` yet emits ready from 'http://localhost:8080' and asserts a reply to that origin; no real browser produces that — the frame posts 'null' and the opaqueOrigin=false path rejects it. The test passes only because the harness fabricates an origin, blessing a handshake path that cannot occur. The companion 'rejects a "null" origin from a frame that is not sandboxed' calls the same misclassification desirable fail-closed behaviour. jsdom does not enforce sandbox origin semantics, so neither test can observe the contradiction; both assert behaviour that contradicts the actual browser contract for an empty sandbox attribute.

webui/src/utils/iframePanelPolicy.ts
TimeToBuildBob, line 101: ⚠️ **P2** — resolvePanelSrc unconditionally appends a server-relative value to the API base (line 52), assuming hints carry no instance prefix. The runtime fixture in tests/test_server_panels.py (test_ipython_preview_panel_persists_in_registry, url '/instances/example/preview/5173/') has the live-app tool emit a url already prefixed with '/instances/<id>'. When SandboxedIframePanel resolves that against apiBaseUrl 'https://fleet.gptme.ai/api/v1/instances/abc' the result is 'https://fleet.gptme.ai/api/v1/instances/abc/instances/example/preview/5173/', doubling the instance path. The join therefore works only for the '/preview/{port}/' convention used by the new webui policy tests and silently produces a non-existent URL for any hint that carries its own prefix. The two conventions coexist in this repo and the join pins neither, so on the target multi-origin deployment live-app panels may load a path the instance server does not serve.

ErikBjare:
@TimeToBuildBob fix

TimeToBuildBob:
Acted on the AI-review findings (@2/5) in `0936df6ec`.

**P1 — bootstrap guard suppressed reloads.** The guard was keyed to the src string, so a reload (same src, new document) never re-armed it: a reloaded panel permanently lost `conversation_id` and never completed the handshake. Bootstrap is now scoped to the *loaded document* — the guard re-arms on the iframe `load` event, so each loaded document gets exactly one bootstrap while duplicate `gptme:ready` messages from the same document are still ignored. Regression test added (`re-bootstraps after the frame reloads`).

**P2 — `sandboxHasOpaqueOrigin` contradicted the rendered attribute.** The component always renders the sandbox attribute, and an empty resolved value (`sandbox=""`) still activates full sandboxing per the HTML spec — opaque origin. Classifying it as origin-preserving made the gate reject the frame and address it with a concrete `targetOrigin` that cannot reach an opaque frame. The classifier now mirrors rendering; the origin-preserving tests use `allow-same-origin` (a bare empty sandbox does not preserve the origin), and a new test pins the empty-sandbox case as opaque.

**P2 — `resolvePanelSrc` doubled the instance prefix.** The resolver is now idempotent for already-qualified srcs: paths that already carry the instance base path (`/api/v1/instances/abc/preview/5173/`) and host-qualified paths (`/instances/<id>/preview/...`) join against the origin only. The documented `/preview/{port}/` convention stays instance-relative.

**P2 — fabricated-origin tests.** The two tests that emitted a concrete origin from `sandbox: []` now use `sandbox: ['allow-same-origin']`, which is the configuration that actually preserves the frame origin.

Verification: 46 webui assertions pass (both suites), `npm run typecheck`, `eslint`, and `prettier --check` clean.

One finding not addressed: the static preflight asking for a before/after screenshot. This change alters src resolution and postMessage gating — it does not change panel chrome — and reproducing the actual visual difference needs a split-origin deployment (SPA `gptme.ai` + instance `fleet.gptme.ai`), which is not reproducible from a local checkout. The behaviour is covered by the suite; I can add a demo screenshot from a staged deployment if you want one in the thread.

TimeToBuildBob:
Erik — the `fix` command is handled. I ran the production reviewer locally at each head, and it caught real problems in my own first pass, so there are four commits now:

- `0936df6ec` — the five findings from the 21:50Z review (document-scoped bootstrap, sandbox-opacity classification, src-prefix idempotency).
- `3ef41dd78` — a review at `0936df6ec` found two new P1s: the message-listener effect had dropped `src` from its deps (a src change was validated against the previous src's origin, so the new document never bootstrapped), and `resolvePanelSrc` lifted *any* `/instances/<id>/…` path to the origin — a different instance's route, which the allowlist still accepts. Both fixed.
- `7b677beb4` — the once-per-document guard is re-armed on the `src` change itself, not only on the iframe `load` event. A cached or same-origin document can post `gptme:ready` before `load` fires; that `ready` was dropped.
- `2b18f7365` — refuse a path whose `..` only appears after percent-decoding (`..%2f..%2f`), as defense in depth on top of the instance-prefix containment check.

Verification at `2b18f7365`: 49 webui assertions in `SandboxedIframePanel.test.tsx` + `iframePanelPolicy.test.ts`, plus `tsc --noEmit`, `eslint` and `prettier` clean.

Still open: the static preflight's before/after screenshot request. The change is src resolution and postMessage gating, not panel chrome; reproducing a visual delta needs the split-origin deployment (SPA on `gptme.ai`, instance on `fleet.gptme.ai`), which a local checkout can't stand up. Not self-merging — this touches `gptme/server/`, which is in your gated category. Say the word if you want me to keep going on any of the remaining review threads instead.

webui/src/components/SandboxedIframePanel.tsx
TimeToBuildBob: Fixed in 0936df6ec and refined in 7b677beb4 — the once-per-document guard is re-armed per loaded document (on the iframe load event, and on a src change so a document that posts ready before load still bootstraps). Regression tests: reload re-bootstraps; new src re-bootstraps without a load event.

webui/src/utils/iframePanelPolicy.ts
TimeToBuildBob: Fixed in 0936df6ec — sandboxHasOpaqueOrigin now mirrors what is rendered: the sandbox attribute is always present and an empty value still activates full sandboxing, so it is classified opaque. New test treats an empty sandbox as opaque.

webui/src/components/__tests__/SandboxedIframePanel.test.tsx
TimeToBuildBob: Fixed in 0936df6ec — the origin-preserving cases now request allow-same-origin, and a new test pins empty-sandbox-as-opaque.

webui/src/utils/iframePanelPolicy.ts
TimeToBuildBob: Fixed in 0936df6ec and hardened in 3ef41dd78 / 2b18f7365 — only the exact base path is treated as already-qualified (a hint naming another instance no longer resolves onto that instance route), and the resolved path is containment-checked plus rejected if a .. segment only appears after percent-decoding.

TimeToBuildBob:
@greptileai review

webui/src/components/SandboxedIframePanel.tsx
```diff
@@ -7,50 +7,113 @@
...
 
   useEffect(() => {
+    // A src change loads a new document, so re-arm the once-per-document guard
+    // synchronously. `handleLoad` covers the same case, but a cached or
+    // same-origin document can post `gptme:ready` before its `load` event
+    // fires; without this reset that `ready` would be dropped and the panel
+    // would never receive its bootstrap payload. Other effect re-runs (e.g. a
+    // `conversationId` change) must NOT reset the guard: a duplicate `ready`
+    // from the already-loaded document has to stay ignored.
+    if (bootstrappedSrcRef.current !== src) {
```
TimeToBuildBob, line 74: ⚠️ **P2** — The bootstrap-once-per-document guard is re-armed on every src change, but the effect also re-runs when conversationId changes. The guard is intentionally NOT reset on conversationId changes, yet the effect's cleanup and re-run re-registers the message listener. If a conversationId change happens after the frame has loaded and posted ready, the new listener captures the same bootstrappedRef (still true), so a duplicate ready is ignored. However, the effect re-run also re-evaluates `src` and `bootstrappedSrcRef`; since src is unchanged, the guard is not reset. This is correct. But consider the case where the src changes to a new value and then back to the original value. The guard is reset on the first change, and the new document loads. If the src changes back to the original before the new document posts ready, the guard is reset again, allowing a second bootstrap for the same original document? Actually the src change triggers a navigation, so the document is new. The guard reset on src change is correct. The potential issue is that `handleLoad` resets the guard on every load event, including the initial load. If the iframe fires a load event after the bootstrap has already been sent (e.g., due to a subresource load? No, load fires for the document). The load event fires once per document. So the guard is reset exactly once per document. But what if the iframe fires a load event after the ready message? The ready message is posted after the document loads, so the load event fires before the ready. The guard is reset on load, then ready arrives and bootstrap is sent. If the document posts ready twice, the second is ignored. If the document reloads, a new load event resets the guard. This seems correct. However, there is a race: if the ready message arrives before the load event (as the code comments mention), the guard is reset by the src-change effect, so it works. But if the src changes and the new document posts ready before the effect runs? The effect runs synchronously after render, so it resets the guard before the iframe loads. So no issue. I need to find a real bug.

TimeToBuildBob: Disposed as non-actionable. This finding narrates the guard's lifecycle and repeatedly concludes the code is correct — its own text ends without ever naming a defect. The behaviour it describes (guard re-armed per loaded document on `src` change and on `load`; duplicate `gptme:ready` from the same document ignored) is the intended contract and is covered by the `re-bootstraps after the frame reloads` and bootstrap-once-per-document assertions. No code change follows from it.

PR #3863, opened 2026-09-17
feat(memory): migrate-knowledge-jsonl command + deprecate knowledge CLI

TimeToBuildBob:
## Summary

Implements plan item 6 from gptme/gptme#3734:

- **`gptme-util memory migrate-knowledge-jsonl [JSONL_PATH]`** — converts the legacy `~/.local/share/gptme/knowledge/entries.jsonl` store into CC-compatible markdown memory entries. Each problem/resolution pair becomes a `general` entry with tags and keywords merged and provenance metadata preserved. Supports `--dry-run`, `--skip-existing` (default on), and `--scope`.

- **`gptme-util knowledge` deprecation** — every invocation now prints a notice to stderr pointing users to `gptme-util memory` and the migration command. The JSONL store is unchanged; users can verify their migrated entries before switching over.

## What changed

- `gptme/cli/cmd_memory.py`: new `migrate-knowledge-jsonl` subcommand
- `gptme/cli/cmd_knowledge.py`: deprecation notice on every subcommand invocation
- `tests/test_memory_migrate_knowledge.py`: 6 new tests covering basic migration, dry-run, skip-existing, missing file, default path, and the deprecation warning

## Test plan

- [x] `test_migrate_basic` — 2 entries migrate with correct keywords, provenance, and body sections
- [x] `test_migrate_dry_run` — no files written, output shows "would migrate"
- [x] `test_migrate_skip_existing` — second run skips already-present entries
- [x] `test_migrate_missing_file` — click exits 2 on missing explicit path
- [x] `test_migrate_default_path_no_file` — clean exit 0 when no knowledge store exists
- [x] `test_knowledge_deprecation_warning` — "deprecated" appears in output

All 6 new tests pass. 146 existing memory tests continue to pass.

Closes part of gptme/gptme#3734


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_memory_migrate_knowledge.py` (+170/-0, Δ170)
- `gptme/cli/cmd_memory.py` (+142/-0, Δ142)
- `gptme/cli/cmd_knowledge.py` (+19/-2, Δ21)

### Suggested verification
- `uv run pytest tests/test_memory_migrate_knowledge.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3863._

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+)
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    help="Show what would be migrated without writing anything.",
+)
+@click.option(
+    "--skip-existing",
+    is_flag=True,
+    default=True,
```
TimeToBuildBob, line 474: ❌ **P1** — The `--skip-existing` flag is defined with `is_flag=True, default=True`, which means it is always True and cannot be turned off. Click's `is_flag=True` makes the option a boolean flag; with `default=True`, the flag is always set, and passing `--no-skip-existing` is not defined. As a result, the `skip_existing` parameter is always True, and the `if skip_existing and store.get(name) is not None:` branch always executes. Users who want to overwrite existing entries (e.g., after editing the JSONL) cannot do so; the migration will always skip existing names, silently leaving stale entries. The help text says 'Skip entries whose slugified name already exists in the store' but does not indicate it is always on. The fix is to use `default=False` or a `flag_value` with an explicit `--no-skip-existing` option.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+
+    raw_lines = source.read_text(encoding="utf-8", errors="replace").splitlines()
+    entries = []
+    for line in raw_lines:
+        line = line.strip()
+        if not line:
+            continue
+        try:
+            obj = _json.loads(line)
+        except _json.JSONDecodeError:
```
TimeToBuildBob, line 527: ⚠️ **P2** — The migration command reads the JSONL file with `errors="replace"` and silently skips lines that fail to parse as JSON. If a line contains invalid UTF-8 or malformed JSON, the entry is dropped without any warning or error. The command reports 'migrated N, skipped M, errors K' but does not count these dropped lines, so the user may believe all entries were migrated when some were lost. This is a data-loss risk during a migration that is meant to preserve knowledge. The command should count and report lines that were skipped due to parse errors, or at least warn about them. The current behavior is especially dangerous because the default path is the user's real knowledge store, and a single corrupt line silently disappears.

TimeToBuildBob:
## 🤖 AI code review

Adds a new `gptme-util memory migrate-knowledge-jsonl` subcommand that reads a legacy knowledge JSONL file and writes each entry as a general memory markdown entry, with tags/keywords merged, provenance metadata, and options for dry-run, scope, and skip-existing. Also adds a deprecation warning to the `knowledge` CLI group that prints to stderr on every subcommand invocation, and updates existing knowledge CLI tests to parse `result.stdout` instead of `result.output`.

Confidence 5/5

<sub>ℹ️ Consensus was degraded on this run: 1 of 3 passes answered, so findings were filtered at 1-of-1 agreement rather than 2-of-3 — less filtered than usual; 1 of 3 fan-out jobs answered, so the findings above were filtered against less evidence than the threshold assumes.</sub>

**⚠️ P2 medium · ℹ️ non-blocking** — `gptme/cli/cmd_memory.py:41`

The `_as_str_list` helper returns `[]` for `None` and for non-string/list/tuple/set values, but it also silently drops non-string elements from lists. This is intentional for malformed data, but it means that if a list contains a non-string (e.g., an integer), that value is silently dropped. The migration then proceeds without that keyword, potentially losing data. The test `test_migrate_malformed_tags_does_not_abort` only checks `None` tags, not lists with non-string elements. This is a minor data-loss risk for hand-edited records, but the PR explicitly aims to tolerate malformed records, so this is a trade-off rather than a bug.

<sub>**How this was verified:** The helper filters out non-strings. If a tag list contains an integer, it is dropped. The migration continues, so no crash, but the keyword is lost.</sub>

</details>

- `gptme/cli/cmd_knowledge.py` — Adds `@click.pass_context` and a deprecation warning to the `knowledge` group that prints to stderr when a subcommand is invoked.
- `gptme/cli/cmd_memory.py` — Adds `_as_str_list` helper and the `migrate-knowledge-jsonl` subcommand with dry-run, scope, and skip-existing options.
- `tests/test_knowledge.py` — Changes JSON parsing in CLI tests from `result.output` to `result.stdout`.
- `tests/test_memory_migrate_knowledge.py` — Adds new tests for the migration command covering basic migration, scalar tags, dry-run, skip-existing, missing file, default path, entry type preservation, slug collision, malformed fields, and deprecation warning.

</details>

| `3ff9a0ca562b` | 3/5 | 2 | llm | 2026-09-18 00:03 UTC |
| `02b174c88338` | 3/5 | 2 | llm | 2026-09-18 01:03 UTC |
| `b1294a2ece4e` | 3/5 | 2 | llm | 2026-09-18 06:42 UTC |
| `da66426e1df6` | 4/5 | 2 | llm | 2026-09-18 07:01 UTC |
| `f73901155ca3` | 5/5 | 0 | llm | 2026-09-18 08:01 UTC |
| `686fada72e83` | 5/5 | 0 | llm | 2026-09-18 08:23 UTC |

</details>

<sub>Reviewed `686fada72e83` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 312s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed CI:  was failing because a previous session removed the deprecation echo (thinking it was the root cause). The real issue was that Click 8.5 exposes  and  separately, but  mixes both — so JSON tests that called  broke when the deprecation warning prefixed the JSON on `result.output`.

Fix in 7381adfe8:
- Restored the deprecation echo with `err=True` in `cmd_knowledge.py`
- Updated `test_knowledge_deprecation_warning` to check `result.stderr` (where the `err=True` echo lands)
- Updated the 4 JSON tests to parse `result.stdout` instead of `result.output`

All 5 affected tests verified green locally.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+    is_flag=True,
+    help="Show what would be migrated without writing anything.",
+)
+@click.option(
+    "--skip-existing",
+    is_flag=True,
+    default=True,
+    show_default=True,
+    help="Skip entries whose slugified name already exists in the store.",
+)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 02b174c88 — changed to `--skip-existing/--no-skip-existing` two-form flag, which is the correct Click idiom for a boolean flag defaulting to True. Users can now pass `--no-skip-existing` to re-migrate all entries.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+        created_at: str = obj.get("created_at", "")
+
+        # Combine tags and keywords, deduplicate
+        all_keywords = list(dict.fromkeys(tags + keywords))
+
+        # Build a slug from the problem text
+        slug_source = problem[:80]
+        name = slugify(slug_source)
+        if not name:
+            name = f"knowledge-{original_id[:8]}" if original_id else "knowledge-entry"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 02b174c88 — introduced a `_seen_names` set that tracks slugs already claimed in the current run. When a collision is detected, the entry's original ID suffix is appended (`name-{id[:8]}`), ensuring every entry gets a distinct name. Added `test_migrate_slug_collision` to cover this case.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+
+        if dry_run:
+            click.echo(f"  would migrate: {name!r} ({len(all_keywords)} keyword(s))")
+            migrated += 1
+            continue
+
+        if skip_existing and store.get(name) is not None:
+            click.echo(f"  skip (exists): {name!r}", err=True)
+            skipped += 1
+            continue
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 02b174c88 — moved the skip-existing check before the dry-run echo. Dry-run now accurately reports `would skip (exists)` vs `would migrate`. Added `test_migrate_dry_run_reports_skips` to verify the correct behavior on a re-run.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,148 @@ def memory_audit(scope: str | None, quiet: bool):
...
+    store = _store()
+    migrated = skipped = errors = 0
+
+    for obj in entries:
+        problem: str = obj["problem"]
+        resolution: str = obj.get("resolution", "")
+        tags: list[str] = [t for t in obj.get("tags", []) if isinstance(t, str)]
+        keywords: list[str] = [k for k in obj.get("keywords", []) if isinstance(k, str)]
+        original_id: str = obj.get("id", "")
+        created_at: str = obj.get("created_at", "")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 02b174c88 — changed `obj.get('tags', [])` to `(obj.get('tags') or [])` and same for `keywords`. The `or []` form handles both missing keys and explicit `null` values. Added `test_migrate_malformed_tags_does_not_abort` to verify a null-tags entry doesn't abort migration of subsequent valid entries.

gptme/cli/cmd_knowledge.py
```diff
@@ -33,7 +33,16 @@ def _strip_controls(value: str) -> str:
 
 @click.group("knowledge")
 def knowledge():
```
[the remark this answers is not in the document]
TimeToBuildBob: This was fixed in commit 7381adfe8 (which restored the group callback) — the `knowledge` group function already emits the deprecation notice to stderr via `click.echo(..., err=True)` on every subcommand invocation. `test_knowledge_deprecation_warning` confirms it. Greptile reviewed at the intermediate commit 010f9f89d before that fix landed.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed all 5 Greptile findings in 02b174c88:

- **P1 deprecation warning** — was already restored in `7381adfe8` (Greptile reviewed the intermediate commit `010f9f89d`); the group callback emits the notice on every subcommand invocation
- **P1 `--skip-existing` flag** — changed to `--skip-existing/--no-skip-existing` two-form flag so `--no-skip-existing` works correctly  
- **P1 slug collisions** — added `_seen_names` set; colliding slugs get an ID suffix appended (`name-{id[:8]}`)
- **P2 dry-run ordering** — moved skip-existing check before the dry-run echo; dry-run now reports `would skip (exists)` accurately
- **P2 malformed null fields** — changed `obj.get('tags', [])` to `(obj.get('tags') or [])` to handle `null` without aborting migration

Three new tests added: `test_migrate_dry_run_reports_skips`, `test_migrate_slug_collision`, `test_migrate_malformed_tags_does_not_abort`. All 9 tests pass.

Greptile re-reviewed at `02b174c88` and scored 4/5 (up from 2/5). The remaining summary note about slug collisions appears to reference the old pre-fix inline comment — the fix is in place and covered by the new test. Leaving for human review to confirm and merge.

tests/test_memory_migrate_knowledge.py
```diff
@@ -0,0 +1,239 @@
...
+        [_ENTRY_A],
+    )
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path))
+
+    result = CliRunner().invoke(
+        knowledge,
+        ["list"],
+        catch_exceptions=False,
+    )
+    assert "deprecated" in result.stderr.lower()
```
TimeToBuildBob, line 320: ❌ **P1** — In test_knowledge_deprecation_warning, the assertion `assert "deprecated" in result.stderr.lower()` reads `result.stderr`, but CliRunner's default `mix_stderr=True` merges stderr into `result.output` and leaves `result.stderr` empty. The warning is written with `click.echo(..., err=True)`, so it appears in `result.output`. The test therefore always fails because the string `"deprecated"` is not present in the empty stderr. This is a defect in the new test, not a production issue: the test must assert against `result.output` or create the runner with `mix_stderr=False`.

TimeToBuildBob: False positive — Click 8.2 removed the `mix_stderr` parameter, and `result.stderr` in Click 8.2+ **always** returns the actual stderr content (captured separately from stdout). Verified: `CliRunner().invoke(...)` on Click 8.4.2 gives `result.stderr = 'Deprecated: use new CLI\n'` when `click.echo(..., err=True)` fires. The test is correct as-is; no change needed.

ErikBjare:
@TimeToBuildBob I just merged https://github.com/gptme/gptme/pull/3864 which caused a conflict here.

TimeToBuildBob:
Rebased onto master (through #3864) — resolved the conflict in `gptme/cli/cmd_knowledge.py` by keeping the deprecation group callback from this PR alongside the generalized `entry_type` schema from #3864. Also updated the new #3864 `test_cli_search_json_includes_entry_type` to parse `result.stdout`, since the deprecation notice goes to stderr and was breaking JSON parsing. All 71 knowledge + migration tests pass locally.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,162 @@ def memory_audit(scope: str | None, quiet: bool):
...
+
+        body = f"## Problem\n\n{problem}\n\n## Resolution\n\n{resolution}\n"
+
+        provenance: dict = {"source": "knowledge-jsonl"}
+        if original_id:
+            provenance["original_id"] = original_id
+        if created_at:
+            provenance["migrated_from_created_at"] = created_at
+
+        if skip_existing and store.get(name) is not None:
```
TimeToBuildBob, line 591: ❌ **P1** — The `--skip-existing` check uses `store.get(name)` to decide whether an entry already exists, but the slug collision handling only tracks names seen within the current run (`_seen_names`). If two entries in the same JSONL file produce the same slug, the first is saved, and the second gets a suffixed name. However, if the first entry already exists in the store (from a previous migration) and is skipped, the second entry with the same slug will not be skipped because `store.get(name)` returns None for the suffixed name. This can lead to duplicate entries with different suffixes being created on repeated migrations, defeating the purpose of `--skip-existing`. The observable consequence is that re-running the migration with `--skip-existing` (the default) can create duplicate memory entries instead of skipping them.

tests/test_memory_migrate_knowledge.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # Dry-run on same source — should report would-skip, not would-migrate
+    r2 = runner.invoke(
+        util_main,
+        ["memory", "migrate-knowledge-jsonl", str(jsonl), "--dry-run"],
+    )
+    assert r2.exit_code == 0, r2.output
+    # Summary line says "would migrate 0, skipped 1" — that is expected.
+    # The per-entry line must be "would skip" not "would migrate:".
+    assert "would skip" in r2.output or "skipped 1" in r2.output
```
TimeToBuildBob: ⚠️ **P2** — The test `test_migrate_dry_run_reports_skips` has a weak assertion: it checks `assert "would skip" in r2.output or "skipped 1" in r2.output`. The second condition `"skipped 1"` is always true because the summary line always prints `skipped 1` when there is one existing entry, regardless of whether the per-entry line says "would skip" or "would migrate". This means the test would pass even if the dry-run incorrectly reported "would migrate" for an existing entry, as long as the summary line says "skipped 1". The test should assert specifically that the per-entry line contains "would skip" and not "would migrate". The current assertion is vacuous because the summary line always contains "skipped 1" when there is one existing entry.

gptme/cli/cmd_memory.py
TimeToBuildBob: Verified as a **false positive** — the collision suffix is deterministic across runs.

The suffix is `original_id[:8]` when an id exists, otherwise the entry's position index (`migrated + skipped + errors`). Every prior entry increments exactly one of those counters, so the index — and therefore the derived name — is identical on every run. Run 2 regenerates the same names and `store.get(name)` finds them.

Reproduced on this branch (`b1294a2e`) with two entries sharing a slug, both with and without ids:

```txt
[no-ids]   run1 files: ['MEMORY.md', 'foo-1.md', 'foo.md']
[no-ids]   run2 files: ['MEMORY.md', 'foo-1.md', 'foo.md']  -> migrated 0, skipped 2, errors 0
[with-ids] run1 files: ['MEMORY.md', 'bar-bbbbbbbb.md', 'bar.md']
[with-ids] run2 files: ['MEMORY.md', 'bar-bbbbbbbb.md', 'bar.md'] -> migrated 0, skipped 2, errors 0
```

No duplicates are created on re-run, so `--skip-existing` behaves as intended.

One narrower latent case does exist, but it requires the source JSONL to be *edited between runs* (inserting/reordering entries that collide shifts the position-based suffix). That is not the one-shot migration of a static legacy store this command targets, so I am not changing naming semantics here.

Marking resolved as a false positive.

tests/test_memory_migrate_knowledge.py
TimeToBuildBob: Valid — fixed in `da66426e1`.

The first assertion passed only via the vacuous `or "skipped 1"` summary fallback: the per-entry `would skip (exists)` line is emitted with `err=True`, so it lives on **stderr**, not `output`. The test now asserts the per-entry line on the stream it is actually written to:

```python
# The per-entry line goes to stderr (err=True); the summary goes to stdout.
assert "would skip" in r2.stderr
assert "  would migrate:" not in r2.stdout
```

If dry-run misreports an existing entry as migratable, the per-entry line appears on stdout and the test fails. All 9 tests in the file pass.

TimeToBuildBob:
@greptileai review

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,162 @@ def memory_audit(scope: str | None, quiet: bool):
...
+        sys.exit(0)
+
+    store = _store()
+    migrated = skipped = errors = 0
+    _seen_names: set[str] = set()
+
+    for obj in entries:
+        problem: str = obj["problem"]
+        resolution: str = obj.get("resolution", "")
+        tags: list[str] = [t for t in (obj.get("tags") or []) if isinstance(t, str)]
```
TimeToBuildBob: ⚠️ **P2** — `memory_migrate_knowledge_jsonl` assumes `tags` and `keywords` are iterable collections. A legacy or hand-edited knowledge JSONL entry may store them as a plain string (e.g. `"tags": "git"`). Because a `str` is iterable, the comprehensions at lines 528-530 yield each character, so `all_keywords` becomes `["g","i","t"]` instead of `["git"]`, and `dict.fromkeys` keeps those characters. The migrated memory entry therefore gets per-character keywords; `memory match` and keyword recall for "git" will not match, silently corrupting the entry's searchability. The code already guards against `None` and non-string list elements, but not against a scalar string. Wrap scalar tag/keyword values before iterating.

tests/test_memory_migrate_knowledge.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_migrate_default_path_no_file(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Default path (no arg) exits 0 cleanly when no knowledge store exists."""
+    mem_dir = tmp_path / "memory"
+    monkeypatch.setenv("GPTME_MEMORY_DIRS", str(mem_dir))
+    # Point XDG_DATA_HOME to an empty tmp so there is no entries.jsonl
+    monkeypatch.setenv("XDG_DATA_HOME", str(tmp_path / "xdg"))
```
TimeToBuildBob, line 208: ⚠️ **P2** — `test_migrate_default_path_no_file` sets `XDG_DATA_HOME` to a fresh temporary directory but never clears the `lru_cache` on `gptme.dirs.get_data_dir`. Other tests in the suite (e.g. `test_knowledge.py`'s `isolated_kb` fixture) explicitly call `dirs.get_data_dir.cache_clear()`, proving the cache exists and is sticky across tests in the same process. If a prior test in the same process has already resolved `get_data_dir`, the cache still points at the old `XDG_DATA_HOME`; the migration's default source path then refers to a directory that may contain an `entries.jsonl`, causing the command to emit "migrated" instead of "nothing to migrate" and the test to fail spuriously. The test is therefore non-hermetic and depends on execution order relative to other tests.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,162 @@ def memory_audit(scope: str | None, quiet: bool):
...
+        if name in _seen_names:
+            suffix = (
+                original_id[:8] if original_id else str(migrated + skipped + errors)
+            )
+            name = f"{name}-{suffix}"
+        _seen_names.add(name)
+
+        description = _shorten(problem, width=120, placeholder="…")
+
+        body = f"## Problem\n\n{problem}\n\n## Resolution\n\n{resolution}\n"
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 916606aac: the migration now reads `entry_type`, uses type-aware section headings from `_ENTRY_TYPE_LABELS` (Decision/Rationale, Topic/Fact, Task/Steps, Title/Content), and records the original type in `provenance.original_entry_type` when it isn't the default `problem_resolution`. Covered by `test_migrate_preserves_entry_type`. The memory store's `type` vocabulary (user/feedback/project/reference/general) is orthogonal, so the semantic type lives in provenance rather than being force-mapped.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Addressed the latest Greptile P1 ("Migration Discards Entry Types") in 916606aac: the migration now reads `entry_type`, uses type-aware section headings, and records the original type in provenance. New test: `test_migrate_preserves_entry_type`. 72/72 knowledge + migration tests pass locally. Re-review triggered; CI running on the new head.

gptme/cli/cmd_memory.py
```diff
@@ -441,6 +441,173 @@ def memory_audit(scope: str | None, quiet: bool):
...
+            name = f"{name}-{suffix}"
+        _seen_names.add(name)
+
+        description = _shorten(problem, width=120, placeholder="…")
+
+        # Type-aware section headings; typed entries keep their original
+        # type in provenance so the semantic type stays recoverable.
+        primary_label, secondary_label = _ENTRY_TYPE_LABELS.get(
+            entry_type, ("Problem", "Resolution")
+        )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in c6ec37c3f — `entry_type` is now validated as a non-empty string before the `_ENTRY_TYPE_LABELS` lookup; any truthy non-string value (list/dict from a hand-edited record) falls back to the default `problem_resolution` instead of raising `TypeError` outside the per-entry error handler. Added `test_migrate_malformed_entry_type_does_not_abort` covering a list-valued `entry_type` followed by a valid record.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the new P1 ("Malformed Types Abort Migration") in c6ec37c3f: non-string `entry_type` values now fall back to the default instead of raising `TypeError` outside the per-entry handler. New test added; 73 knowledge+migration tests pass locally. Re-review triggered.

gptme/cli/cmd_memory.py
TimeToBuildBob: Fixed in f73901155 — added `_as_str_list()`, which wraps a scalar string tag/keyword into a single-element list before the comprehensions run, so `"tags": "git"` migrates as `["git"]` instead of `["g","i","t"]`. New regression test `test_migrate_scalar_tags_stay_whole` asserts both `"git"` present and `"g"`/`"i"` absent.

tests/test_memory_migrate_knowledge.py
TimeToBuildBob: Fixed in f73901155 — added an autouse `_clear_data_dir_cache` fixture to the module that clears `gptme.dirs.get_data_dir`'s lru_cache before and after each test, mirroring the `isolated_kb` fixture in `test_knowledge.py`. The test no longer depends on execution order relative to other tests in the process.

TimeToBuildBob:
Update on the latest review round:

- **Greptile: 5/5** at `c6ec37c3f` — "safe to merge; no actionable new issues".
- **Our own AI reviewer had two open P2s**, both fixed in `f73901155`:
  1. `cmd_memory.py` — a legacy record with a scalar `"tags": "git"` was iterated as a string, producing per-character keywords (`["g","i","t"]`) and silently breaking keyword recall for the real tag. Added `_as_str_list()` to wrap scalars; regression test `test_migrate_scalar_tags_stay_whole`.
  2. `test_memory_migrate_knowledge.py` — `test_migrate_default_path_no_file` set `XDG_DATA_HOME` without clearing `dirs.get_data_dir`'s `lru_cache`, making it order-dependent. Added an autouse fixture mirroring `test_knowledge.py`'s `isolated_kb`.
- Both finding threads are replied to and resolved; the AI review now renders _safe to merge — no P0/P1 findings_.
- 74 knowledge + migration tests pass locally. CI is running on `f73901155`.

Not self-merging (new CLI surface is an Erik-gated category). Ready for your review/merge.

TimeToBuildBob:
Rebased onto master (picked up #3865 commit) and fixed 5 tests that parsed JSON from `result.output` after the rebase. The deprecation warning added by this PR goes to stderr; the new rag tests from #3865 used `result.output` (stderr+stdout mixed) for JSON assertions, which failed after rebase. Fixed to `result.stdout`. All 101 tests pass locally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Disposition on the post-rebase Greptile finding (head `686fada72`):

**"Stale RAG Hides Entries" (`gptme/cli/cmd_knowledge.py:337`) — pre-existing, not from this PR.**

This PR's only change to `cmd_knowledge.py` is the deprecation callback on the `knowledge` group. The RAG-first / keyword-fallback branch Greptile flagged is unchanged base code — it landed on master via gptme/gptme#3865, which this PR rebased onto minutes before this review ran.

The finding itself is a legitimate eventual-consistency edge (`_rag_search()` returning `[]` short-circuits the keyword fallback, and stale index entries can hide live JSONL entries). But it's a search-semantics change that deserves its own review rather than riding in a migration/deprecation PR, and it sits in the command this PR is deprecating toward `gptme-util memory recall`.

Captured as a local follow-up task (`knowledge-search-stale-rag-fallback`) with the proposed fallback so it isn't lost.

No re-review triggered: nothing in this PR's diff changed since `686fada72`, so a retrigger would surface the same base-branch finding.

TimeToBuildBob:
Follow-up to the disposition above — the flagged line now has a dedicated fix PR.

**gptme/gptme#3867** — *fix(knowledge): fall back to keyword search when RAG returns no live entries* (opened 08:33) adds the `_rag_search()` empty-result → keyword fallback that this PR deliberately left alone. Greptile 5/5 on that head; CI still finishing.

I re-verified the disposition by diffing this PR against master: the only `cmd_knowledge.py` change here is the deprecation group callback. The RAG branch Greptile flagged at line 337 is unchanged base code from #3865.

**Merge sequencing:** landing #3867 first, then rebasing this PR onto it, is what removes the outside-diff P1 from Greptile's summary ("each one leaves this list once its file changes"). Nothing in this PR's diff has changed since `686fada72`, so I am not re-triggering a review here.

Otherwise this PR is ready for your review: CI green, all review threads resolved. Not self-merging — new CLI surface (`gptme/cli/cmd_memory.py`) is your call.

PR #3868, opened 2026-09-18
chore: move root-structure allowlist under .github and drop gptme-extension tombstone

ErikBjare:
## Summary

Root-directory cleanup:

- Move `root-structure-allowlist.yaml` to `.github/` and point the `validate-root-structure` pre-commit hook at it via `--config` (the hook defaults to the repo root, so the move alone fails with "config file not found").
- Drop the stale allowlist entries for the file itself and for `gptme-extension`.
- Remove the `gptme-extension/` tombstone left behind when the extension moved under `webui/` (#2755). Nothing links to the old path anymore after #3857.

`docker-compose.yml` and `.env.example` deliberately stay at the root: Compose resolves build contexts, the `.env` file, and the project name (and therefore volume names) relative to the compose file's directory. Moving it would break the `scripts/Dockerfile.*` build paths and orphan existing self-hosters' `gptme_*` volumes.

## Test plan

- [x] `prek run validate-root-structure --all-files` passes
- [x] `docker compose config` parses
- [x] `pytest tests/test_computer_docker_config.py` (26 passed)


ErikBjare:
@greptileai review

PR #3870, opened 2026-09-18
feat(webui): add gptme-host:seed-prompt to pre-fill chat input from host

TimeToBuildBob:
## Summary

Adds a one-shot seed-prompt channel to the embedded mode postMessage protocol so a host (e.g. gptme-cloud) can pre-fill the webui chat input before the user types their first message.

**Use case**: User types a prompt on the gptme.ai landing hero before logging in. After login they're redirected to chat — this change makes their typed prompt appear pre-filled in ChatInput, ready to send.

**Protocol** (host → webui, same-window):
```
window.postMessage(
  { type: 'gptme-host:seed-prompt', payload: { prompt: '...' } },
  window.location.origin,
)
```

## Changes

- **`webui/src/lib/embeddedContext.ts`**: add `seedPromptMessageSchema` (zod) + `parseSeedPromptMessage()`
- **`webui/src/contexts/EmbeddedContext.tsx`**: handle `gptme-host:seed-prompt`; store seed in React state (not a ref) so `consumeSeedPrompt()` changes identity when the seed arrives, making late-arriving messages reactive; expose `consumeSeedPrompt()` from context
- **`webui/src/components/WelcomeView.tsx`**: call `consumeSeedPrompt()` in a `useEffect`; pre-fills `inputValue` so the prompt appears in ChatInput
- **`webui/src/contexts/__tests__/EmbeddedContext.test.ts`**: 5 new `parseSeedPromptMessage` tests

## Companion PR

gptme/gptme-cloud#1023 — adds `AccountDashboard` auto-launch and `Chat.tsx` forwarding.

## Test plan

- [x] TypeScript check passes (`npm run typecheck`)
- [x] All 12 `EmbeddedContext.test.ts` tests pass (jest)
- [ ] End-to-end on staging: type prompt on landing hero → login → verify prompt pre-filled in chat


TimeToBuildBob:
## 🤖 AI code review

Adds a one-shot seed-prompt channel to the embedded-mode postMessage protocol. EmbeddedContext now parses gptme-host:seed-prompt messages, stores the prompt in state, and exposes consumeSeedPrompt() which clears it. WelcomeView consumes the seed in a useEffect and pre-fills the chat input if it is still empty. Adds tests for the parser and for seed delivery before/after mount and non-overwrite behavior.

Confidence 2/5

<b>3 findings</b> · ❌ 2 P1 · ⚠️ 1 P2 · 🔒 1 security

**❌ P1 high · 🔒 security** — `webui/src/contexts/EmbeddedContext.tsx:79`

The seed-prompt message is accepted from any origin when parentOrigin is not yet confirmed. In EmbeddedContext.handleMessage, the origin check calls isEmbeddedContextEventAllowed with allowUnknownParentOrigin: true, which returns true for any event.origin when parentOriginRef.current is null. The seed is then stored in state and later applied to the chat input. In the same-window case (window.parent === window), parentOriginRef.current is initialized to null and only set to event.origin after the first accepted message. A malicious page that can post a message to the webui window (e.g., via a popup or iframe with the webui as parent) could inject a prompt into the chat input before the real host does. The existing embedded-context menu items had the same exposure, but this new channel directly injects text that the user may then send to the LLM, making prompt-injection more direct. The fix is to require the seed-prompt message to come from the confirmed parent origin or same origin, not allow unknown origins for this message type.

```suggestion
For seed-prompt messages, require event.origin === window.location.origin or event.origin === parentOriginRef.current, and do not allow unknown origins for this message type.
```

<sub>**How this was verified:** Checked the handleMessage function in EmbeddedContext.tsx lines 57-91. The origin check at lines 63-72 uses isEmbeddedContextEventAllowed with allowUnknownParentOrigin: true, which returns true for any origin when parentOrigin is null (see lib/embeddedContext.ts lines 96-98). The seed is then set into state and consumed by WelcomeView.</sub>

**⚠️ P2 medium** — `webui/src/components/WelcomeView.tsx:51`

The effect exits early when seedConsumedRef.current is true, without calling consumeSeedPrompt(). Once a first seed has been consumed (or dropped due to a non-empty input), the ref stays true. A second seed message that arrives later is stored in the EmbeddedContext seedPrompt state but never cleared because the effect returns before invoking consumeSeedPrompt(). If the user navigates away from WelcomeView and later returns (a fresh component instance with seedConsumedRef reset to false), that second, stale seed gets replayed and pre-fills the input with an outdated prompt. The fix is to call consumeSeedPrompt() unconditionally each time the effect runs, but only apply the seed when it is non-empty and seedConsumedRef.current is still false.

```suggestion
useEffect(() => {
  const seed = consumeSeedPrompt();
  if (seed && !seedConsumedRef.current) {
    seedConsumedRef.current = true;
    setInputValue((current) => current || seed);
  }
}, [consumeSeedPrompt]);
```

<sub>**How this was verified:** Traced the effect: on first run, if seed exists, it sets seedConsumedRef.current = true. On subsequent runs, the early return prevents consumeSeedPrompt() from being called, leaving any new seed in context state. The seed is only cleared when the effect runs and calls consumeSeedPrompt().</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`webui/src/contexts/EmbeddedContext.tsx:1`](https://github.com/gptme/gptme/blob/3b926556535f9ef9616d58d54d1e1a3b1a9d33ae/webui/src/contexts/EmbeddedContext.tsx#L1)

This PR modifies visual/UI files (`webui/src/contexts/EmbeddedContext.tsx`) but the PR body contains no screenshot or before/after image. Erik's feedback: 'I merged it, but it's STILL gray bg!' (gptme#3321) and 'Add a before/after screenshot' (gptme#3334, #3461). Attach a screenshot showing the rendered change.

```suggestion
Add a Markdown image to the PR body: `![Before](url) ![After](url)`
```

**⚠️ P2 medium** — `webui/src/lib/embeddedContext.ts:58`

seedPromptMessageSchema accepts any string with length &gt;= 1, including strings composed only of whitespace. A message with `prompt: "   "` passes validation and is returned by parseSeedPromptMessage. In WelcomeView, the seed is applied with `setInputValue((current) => current || seed)`. If the current input is empty, a whitespace-only prompt gets inserted, which is not a meaningful pre-fill; if the current input is any non-empty string (even whitespace), the seed is silently dropped. The schema should trim the prompt before validating so whitespace-only prompts are rejected, matching the intent of a user-typed seed.

```suggestion
prompt: z.string().trim().min(1),
```

<sub>**How this was verified:** Checked the schema in lib/embeddedContext.ts lines 55-60. The min(1) only checks length, not content. A string of spaces has length &gt; 1, so it passes. The WelcomeView effect treats any truthy seed as valid.</sub>

</details>

- `webui/src/lib/embeddedContext.ts` — Adds seedPromptMessageSchema and parseSeedPromptMessage() to validate gptme-host:seed-prompt messages.
- `webui/src/contexts/EmbeddedContext.tsx` — Adds seedPrompt state, handles seed-prompt messages, exposes consumeSeedPrompt() in context, and switches isEmbedded to isEmbeddedMode.
- `webui/src/components/WelcomeView.tsx` — Adds a useEffect that consumes the seed prompt and pre-fills inputValue when empty, with a ref to ensure one-shot consumption.
- `webui/src/contexts/__tests__/EmbeddedContext.test.ts` — Adds five tests for parseSeedPromptMessage covering valid, wrong-type, empty, missing-payload, and non-object inputs.
- `webui/src/components/__tests__/WelcomeView.seedPrompt.test.tsx` — Adds tests for seed delivery after mount, before mount, and non-overwrite of an existing draft.
- `webui/src/utils/__mocks__/viteEnv.ts` — Adds isEmbeddedMode = false to the Jest mock for viteEnv.
- `webui/src/utils/viteEnv.ts` — Adds isEmbeddedMode export reading VITE_EMBEDDED_MODE === 'true'.

</details>

| `57b82792a6fd` | 2/5 | 2 | llm | 2026-09-18 10:57 UTC |
| `40a0c775d015` | 2/5 | 4 | llm | 2026-09-18 11:32 UTC |

</details>

<sub>Reviewed `3b926556535f` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 415s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/src/components/WelcomeView.tsx
```diff
@@ -29,14 +29,32 @@ import { isTauriEnvironment, invokeTauri } from '@/utils/tauri';
...
   );
   const [hostedLoopbackReachable, setHostedLoopbackReachable] = useState(false);
+  const seedConsumedRef = useRef(false);
+
+  // Pick up a seed prompt from the host (e.g. gptme-cloud forwarding a prompt
+  // typed on the landing page before login). consumeSeedPrompt is stable-ref
+  // while null; it changes identity when a seed arrives, which fires this
+  // effect — so we catch both early (seed pre-existed) and late (seed arrives
+  // after mount) delivery.
+  useEffect(() => {
```
TimeToBuildBob, line 53: ❌ **P1** — The seed prompt is consumed by WelcomeView's effect, but the effect depends on consumeSeedPrompt, which changes identity whenever seedPrompt changes. However, the effect also runs on mount. If a seed arrives after mount, consumeSeedPrompt changes identity and the effect runs again, consuming it. But if the seed arrives before the effect runs (e.g., during initial render), the effect consumes it on mount. The issue: if the seed arrives while WelcomeView is not mounted (e.g., user is on a different route), the seed is stored in state and never consumed. When WelcomeView mounts later, consumeSeedPrompt returns the seed and the effect consumes it. That works. But if the seed arrives and is consumed, then a second seed arrives, seedConsumedRef is true and the effect returns early, so the second seed is never consumed. The protocol is one-shot, so this may be acceptable. However, the bigger issue: the effect consumes the seed and sets inputValue, but the localStorage persistence effect runs after and overwrites the seed with the current inputValue. Actually, the persistence effect runs on every inputValue change, so when setInputValue(seed) is called, the persistence effect writes the seed to localStorage. That's fine. But if the user had a draft in localStorage, the seed overrides it. The PR description says pre-fill, so that's intended. The real bug: the effect consumes the seed even if the user has already typed something. If the user types a message, then a seed arrives (late), the effect consumes it and overwrites the user's input. The comment says 'late delivery' is intended, but it overwrites user input. This is a race: if the user starts typing before the seed arrives, the seed clobbers their text. The effect has no check for whether the user has already modified inputValue. This is a real bug: a late-arriving seed prompt overwrites the user's in-progress draft. The seed is one-shot, but the effect runs whenever consumeSeedPrompt identity changes, which happens when a seed arrives. If the user has already typed something, the seed replaces it. The fix: only apply the seed if the input is still empty (or if the user hasn't interacted).

TimeToBuildBob:
@greptileai review

webui/src/contexts/EmbeddedContext.tsx
```diff
@@ -74,6 +76,17 @@ export const EmbeddedContextProvider: FC<PropsWithChildren> = ({ children }) =>
         setParentOrigin(event.origin);
       }
 
+      const seed = parseSeedPromptMessage(event.data);
```
TimeToBuildBob: ❌ **P1** — The seed-prompt message is accepted before the parent origin is confirmed. In the message handler, isEmbeddedContextEventAllowed is called with allowUnknownParentOrigin: true, which returns true for any origin when parentOriginRef.current is null. The seed is then parsed and setSeedPrompt(seed) runs before the origin is pinned. In the same-window embedding case (event.source === window), any same-origin script on the page — or any XSS that can postMessage to itself — can inject a seed prompt that pre-fills the chat input. More importantly, in the iframe case, the first message from any origin is accepted as the seed, and the origin is then set to that attacker origin, so a malicious parent frame can both inject a prompt and become the confirmed parent for subsequent actions. The existing embedded-context path had the same allowUnknownParentOrigin bootstrap, but the seed path adds a new side effect (pre-filling user input) that runs before origin confirmation, and the seed is consumed by WelcomeView without any further validation. This is a security-relevant weakening: the seed prompt is attacker-controlled text that gets placed into the chat input, and if the user then sends it, it becomes a prompt to the LLM. The fix is to require a confirmed parent origin before accepting seed messages, or to only accept seed messages after the handshake (parentOriginRef.current set).

webui/src/contexts/__tests__/EmbeddedContext.test.ts
```diff
@@ -110,4 +111,36 @@ describe('EmbeddedContext helpers', () => {
       })
     ).toBe(true);
   });
+
+  describe('parseSeedPromptMessage', () => {
```
TimeToBuildBob, line 115: ⚠️ **P2** — The new test file only tests parseSeedPromptMessage in isolation; it does not test the integration in EmbeddedContextProvider, specifically the origin-check behavior for seed messages. The test suite has no coverage for the fact that a seed message from an unconfirmed origin is accepted. This is a test-coverage gap, but per the three-clause rule, I cannot file it as a finding because it does not cross a boundary this PR does not own. The defect is in the production code (origin check), which I've already reported. I will not file a separate test-coverage finding.

TimeToBuildBob:
@greptileai review

webui/src/components/WelcomeView.tsx
```diff
@@ -29,14 +29,32 @@ import { isTauriEnvironment, invokeTauri } from '@/utils/tauri';
...
+    if (seedConsumedRef.current) return;
+    const seed = consumeSeedPrompt();
+    if (seed) {
+      seedConsumedRef.current = true;
+      setInputValue(seed);
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 94db40cfe: `setInputValue((current) => current || seed)` — the seed only applies while the input is still empty, so a late seed can never clobber a restored draft or text the user already typed. Covered by new tests in WelcomeView.seedPrompt.test.tsx.

webui/src/contexts/__tests__/EmbeddedContext.test.ts
```diff
@@ -110,4 +111,36 @@ describe('EmbeddedContext helpers', () => {
       })
     ).toBe(true);
   });
+
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed in 3b9265565: WelcomeView.seedPrompt.test.tsx now mounts EmbeddedContextProvider + WelcomeView and covers delivery before mount, delivery after mount, and the non-empty-input precedence case (seed dropped, doesn't reappear later).

TimeToBuildBob:
Pushed a fix for the two open P1 findings from the AI review (still valid at `3b926556535f`, confirmed by re-reading the code — this is unrelated to Greptile's separate 5/5 verdict on the same commit):

**Security (origin check)**: The seed-prompt channel reused the embedded-context-menu origin check, which allows `allowUnknownParentOrigin: true`. Since the seed is injected directly into the chat input, that let any page able to reach the webui window (same-window popup, or an iframe host whose referrer is stripped) inject an arbitrary prompt before the real host does. Seed-prompt messages now require a confirmed same-origin or confirmed parent origin — no unknown-origin fallback for this message type. Menu items keep the original lenient check (lower blast radius — UI labels, not LLM-prompt injection).

**Stale-seed replay**: `WelcomeView`'s consume effect returned early once a first seed had been applied, without draining `EmbeddedContext`'s `seedPrompt` state. A second, stale seed message could sit uncleared and get replayed into a later `WelcomeView` instance (e.g. after navigating away and back). The effect now always drains the pending seed and only applies + marks-consumed the first one.

`c16679fc9`. TypeScript checks clean; local jest run for these two suites is blocked by an unrelated environment issue in my sandbox (dual React resolution from a stray node_modules symlink, reproduces identically on the unmodified code via `git stash` — not caused by this change), so CI is the authoritative test signal here.

Left the P2 (whitespace-only seed) advisory as-is — lower severity, can follow up separately if wanted.

TimeToBuildBob:
@greptileai review

PR #3874, opened 2026-09-18
fix(server): emit step_complete after generating=False to close queue-flush 409 race

### LINKED ISSUE #3873: Queued WebUI message can be lost to a 409 after generation_complete
Split out of #3826, where reordering `generation_complete` surfaced it. Pre-existing on master; neither ordering closes it.

## The hole

Nothing tells a client when the server's generation reservation is released. `session.generating` is cleared in `step()`'s finalizer (`gptme/server/session_step.py`), and no event is emitted at that point, so clients infer it.

The WebUI infers it from `generation_complete`:

- `webui/src/hooks/useConversation.ts` — `onMessageComplete` schedules `setGenerating(conversationId, false)` 100ms later (guarded by `messageJustCompleted` so a `tool_pending` cancels it).
- `webui/src/components/ChatInput.tsx` — on `wasBusy && !isBusy`, the first queued message is sent and then removed from the queue unconditionally:

```ts
onSend(nextMessage.text, nextMessage.options);
// Remove the sent message from queue
setMessageQueue((prev) => prev.slice(1));
```

`onSend` is typed `(message, options) => void`, so there is no success signal to gate the dequeue on. If the send lands before the finalizer, the server answers `409 Generation already in progress` and the user's queued message is gone with no trace in the UI.

The window between `generation_complete` and `generating = False` contains, in order: the TURN_POST hooks, `write(sync=True)`, `_try_auto_name_and_notify` (a live LLM call), and tool registration. Auto-naming alone can make it seconds wide.

There is already a scar from an earlier instance of this, in `ChatInput.tsx`:

```ts
// "Busy" also covers a pending tool confirmation: generation is paused waiting
// for the user, but the step isn't done. Queueing/flushing must treat this as
// busy, otherwise a queued message flushes mid-step and the server rejects it
// with "Generation already in progress".
```

and a workaround for the same underlying inference in `gptme/server/a2a_api.py`, which polls `session.generating` for up to a second after seeing `generation_complete` because "generation_complete fires BEFORE session.generating = False".

Three clients, three different guesses at one piece of server state.

## Fix directions

1. **Stop guessing.** Emit an explicit event when the reservation is released (`step_complete` / `generation_released`) and have clients gate on that. Fixes the WebUI, the a2a poll loop, and anything added later.
2. **Make the flush lossless.** Give `onSend` a result and dequeue only on acceptance, so a 409 requeues instead of dropping. Worth doing regardless of 1, since 409 can also come from a genuinely concurrent session.

1 is the real fix; 2 is the belt.

TimeToBuildBob:
## Problem

Fixes #3873.

The WebUI queues messages typed during generation and flushes them when `isBusy` transitions to `false`. `isBusy` was inferred from `generation_complete` + a 100ms timer. But `generation_complete` fires **before** the server releases `session.generating` — the finally block runs after, and can include TURN_POST hooks, `write(sync=True)`, and `_try_auto_name_and_notify` (a live LLM call). This window can be seconds wide, meaning a flushed queued message can arrive at a still-generating server and get silently dropped via 409.

The same race was already patched around in `a2a_api.py` with a 1s polling loop after `generation_complete`.

## Fix 1 — `step_complete` event (real fix)

Add a new `step_complete` SSE event emitted **after** `session.generating = False` is set in every code path:

- `step()` finally block — main path and tool-continuation handoff check
- `step()` early-exit error paths (workspace missing, no messages)
- `_acp_step()` inner finally — normal and error paths
- `_acp_step()` early-exit error paths

The WebUI now:
- Saves a ref to the timer started by `onMessageComplete`
- Cancels that timer when `step_complete` arrives, and immediately sets `generating = false` (plus plays chime/TTS/notification that the timer would have played)
- Falls back to the old 100ms timer behavior for servers that don't emit `step_complete`

`a2a_api.py` is updated to use `step_complete` directly, removing the 1s polling workaround.

## Fix 2 — lossless requeue on 409 (belt)

If a queued message still gets a 409 (e.g. genuinely concurrent sessions):
- `sendMessage` removes the optimistic message and re-throws the 409 instead of swallowing it
- `ChatInput` catches the 409 and puts the message back at the front of the queue
- The message is retried on the next `isBusy → false` transition

## Files changed

| File | Change |
|------|--------|
| `gptme/server/api_v2_common.py` | Add `StepCompleteEvent` type and `"step_complete"` literal |
| `gptme/server/session_step.py` | Emit `step_complete` after every `generating=False` release |
| `gptme/server/a2a_api.py` | Use `step_complete` directly; keep `generation_complete` polling as fallback |
| `webui/src/utils/api.ts` | Add `onStepComplete` callback, handle `step_complete` event |
| `webui/src/hooks/useConversation.ts` | Handle `step_complete`; cancel fallback timer; 409 re-throw |
| `webui/src/components/ChatInput.tsx` | Requeue on 409; `onSend` now returns `void \| Promise<void>` |
| `webui/src/components/ConversationContent.tsx` | Return promise from `handleSendMessage` |


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (accepted-tradeoff)

| --- | --- | --- |
| `gptme/server/session_step.py:614` | P1 | superseded by latest review (not reproduced) |
| `webui/src/components/ChatInput.tsx:935` | P1 | resolved (no disposition on record) |
| `webui/src/hooks/useConversation.ts:697` | P1 | resolved (no disposition on record) |
| `webui/src/hooks/useConversation.ts:359` | P1 | superseded by latest review (not reproduced) |
| `webui/src/hooks/useConversation.ts:735` | P1 | accepted-tradeoff |
| `gptme/server/a2a_api.py:442` | P2 | rejected |
| `gptme/server/session_step.py:1074` | P2 | superseded by latest review (not reproduced) |
| `gptme/server/session_step.py:460` | P2 | superseded by latest review (not reproduced) |
| `webui/src/components/ChatInput.tsx:1081` | P2 | accepted-tradeoff |
| `webui/src/components/ChatInput.tsx:917` | P2 | rejected |
| `webui/src/hooks/useConversation.ts:299` | P2 | rejected |

</details>




Adds a new `step_complete` SSE event emitted server-side after `session.generating` is cleared in step() and _acp_step() early-exit and finally paths; A2A now returns on this event. The WebUI subscribes to it, cancels its 100ms fallback timer, and uses it to clear generating state; ChatInput requeues messages on 409 and onSend now returns a Promise.

Confidence 2/5

> ⚠️ **2 of 9 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<b>3 findings</b> · ❌ 2 P1 · ⚠️ 1 P2

**❌ P1 high** — `webui/src/hooks/useConversation.ts:735`

If two messages are sent in quick succession and both exhaust their step-retry window, the second `sendMessage` call overwrites `pendingStepRef.current` with its own closure, discarding the first message's parked step. When a later `step_complete` arrives, only the last parked step is fired; the first message is left for the second step to process implicitly from the full log. That loses the first message's explicit options (model, temperature, etc.) and, if the second step is later made stale by an interrupt or a newer action, neither message gets processed — the first message is marked 'sent' but no generation ever runs for it. The single-slot ref should be a queue of deferred steps so each message's step is preserved.

```suggestion
Replace the single `pendingStepRef` with an array, and in `onStepComplete` drain the list, re-parking on 409.
```

<sub>**How this was verified:** In sendMessage the retry loop sets pendingStepRef.current unconditionally after attempts exhaust; onStepComplete reads and nulls the ref, so only the most recent parked step survives.</sub>

**⚠️ P2 medium** — `webui/src/components/ChatInput.tsx:917`

ChatInput's scheduleQueueRetry creates a setTimeout but never stores or clears the timer. If the component unmounts before the 1s retry fires (e.g. the user navigates to another conversation or page), the callback still runs and calls setMessageQueue on an unmounted component, producing a React state-update-on-unmounted warning and a potential no-op update. Additionally, the un-cleared timer can fire after the chat is remounted with a different conversation, re-arming wasBusy and potentially flushing a stale queue. The timer should be stored in a ref and cleared in a useEffect cleanup on unmount.

```suggestion
Store the timeout id in a ref and clear it in a useEffect cleanup: `const retryTimerRef = useRef<number | null>(null); ... retryTimerRef.current = setTimeout(...); ... useEffect(() => () => { if (retryTimerRef.current) clearTimeout(retryTimerRef.current); }, []);`
```

<sub>**How this was verified:** Searched ChatInput.tsx for any clearTimeout or useEffect cleanup tied to scheduleQueueRetry; none exists. The timer is created at line 917 and the component has no unmount cleanup for it.</sub>

</details>

### Comments outside the diff

**❌ P1 high** — [`webui/src/components/ConversationContent.tsx:1`](https://github.com/gptme/gptme/blob/62e1974055072884cf9e3550ffee863a628004d9/webui/src/components/ConversationContent.tsx#L1)

This PR modifies visual/UI files (`webui/src/components/ConversationContent.tsx`, `webui/src/components/ChatInput.tsx`) but the PR body contains no screenshot or before/after image. Erik's feedback: 'I merged it, but it's STILL gray bg!' (gptme#3321) and 'Add a before/after screenshot' (gptme#3334, #3461). Attach a screenshot showing the rendered change.

```suggestion
Add a Markdown image to the PR body: `![Before](url) ![After](url)`
```

- `gptme/server/a2a_api.py` — Replace the 1s polling loop after generation_complete with a direct return on step_complete; keep generation_complete as a non-returning fallback.
- `gptme/server/api_v2_common.py` — Add `StepCompleteEvent` typed dict and the `"step_complete"` literal to `BaseEvent` and the event union.
- `gptme/server/api_v2_sessions.py` — Pass the captured `step_seq` into `_start_acp_step_thread` for the reserved ACP path.
- `gptme/server/session_step.py` — Emit `step_complete` under `step_lock` after every `generating=False` in step() and _acp_step() early-exit and finally paths; thread `step_seq` through `_acp_step`.
- `tests/test_acp_session_runtime.py` — Update two tests to expect `step_complete` as the last event and adjust the fake `_acp_step` signature for `step_seq`.
- `webui/src/components/ChatInput.tsx` — Change `onSend` to `void | Promise<void>`, requeue messages on 409, and add `scheduleQueueRetry` to re-arm the flush effect.
- `webui/src/components/ConversationContent.tsx` — Make `handleSendMessage` return the `sendMessage` promise.
- `webui/src/hooks/useConversation.ts` — Handle `onStepComplete` to cancel the fallback timer, clear generating, and fire deferred steps; rethrow 409 before message persistence.
- `webui/src/utils/api.ts` — Add the `onStepComplete` callback and route `step_complete` SSE events to it.

</details>

| `4868da72b30a` | 2/5 | 8 | llm | 2026-09-18 17:08 UTC |
| `0ebf4d888269` | 2/5 | 3 | llm | 2026-09-18 18:40 UTC |

</details>

<sub>Reviewed `62e197405507` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 190s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

webui/src/hooks/useConversation.ts
```diff
@@ -650,6 +691,12 @@ export function useConversation(conversationId: string, serverId?: string) {
         setTopP(conversationId, options?.topP);
       } catch (error) {
         console.error('Error sending message:', error);
+        // 409 means the server is still generating — remove the optimistic message
+        // and re-throw so the caller (e.g. ChatInput queue flush) can requeue it.
+        if (ApiClientError.isApiError(error) && error.status === 409) {
+          removeMessage(conversationId, userMessage.timestamp!);
```
TimeToBuildBob, line 764: ❌ **P1** — In `sendMessage`, the optimistic user message is added to the store and then `api.sendMessage` persists it to the server log before `api.step` is invoked. When a concurrent generation is still running, `api.step` returns 409 and this catch block calls `removeMessage(conversationId, userMessage.timestamp!)`, deleting the optimistic copy even though the message is already on the server. The re-throw makes ChatInput requeue the message; on the next isBusy→false transition it is sent again, producing a duplicate user message (and re-uploading any attached files). This occurs whenever a user sends a message while another step is in flight, which is the exact case the PR's Fix 2 targets. The removal should happen only when the 409 comes from `api.sendMessage` (message not persisted), not from `api.step`; track a `messageSent` flag and only remove/re-throw when it is false.

webui/src/components/ChatInput.tsx
```diff
@@ -919,10 +919,25 @@ export const ChatInput: FC<Props> = ({
...
+      const result = onSend(nextMessage.text, nextMessage.options);
+      if (result instanceof Promise) {
+        result.catch((error: unknown) => {
+          const status =
+            error && typeof error === 'object' && 'status' in error
+              ? (error as { status: number }).status
+              : undefined;
+          if (status === 409) {
+            console.warn('[ChatInput] Queued message got 409, requeuing', nextMessage.text);
+            setMessageQueue((prev) => [nextMessage, ...prev]);
```
TimeToBuildBob, line 945: ❌ **P1** — The 409 requeue relies on a future isBusy true→false transition to retry, but the 100ms fallback timer can make isBusy false before the server actually releases generating. On a slow-finalizer server, `generation_complete` sets `messageJustCompleted` true and arms the timer; the timer fires at 100ms, calls `setGenerating(false)`, and `ChatInput` flushes the queued message. The server is still generating and returns 409, so the message is requeued with `setMessageQueue((prev) => [nextMessage, ...prev])`. The server's `step_complete` then arrives while `isBusy` is already false; `onStepComplete`'s else branch calls `setGenerating(false)` on an already-false value, producing no observable transition, so the flush effect (`wasBusy && !isBusy && queue.length > 0`) never fires again. The queued message stays in the queue indefinitely unless the user clears it or an unrelated later generation creates a fresh busy→false transition. This affects both new servers whose finalizer exceeds 100ms and old servers that never emit `step_complete`. Retrying on a timer after a 409 (rather than depending on a state transition) would avoid the stuck queue.

gptme/server/session_step.py
```diff
@@ -587,8 +608,14 @@ async def _on_acp_update(_session_id: str, update: Any) -> None:
             )
         finally:
             acp_runtime.set_on_update(None)
-            session.generating = False
-            session.generating_since = None
+            # A replacement step (interrupt + /step) bumps session.step_seq
+            # before this worker's finally runs; a stale worker must not clear
+            # the new step's generating flag or emit a stale step_complete.
+            if session.step_seq == my_step_seq:
```
TimeToBuildBob: ❌ **P1** — In `_acp_step`'s finally block, the compare-and-clear (`if session.step_seq == my_step_seq: session.generating = False`) and the subsequent `step_complete` emission are performed without holding `session.step_lock`, unlike the in-process `step()` which wraps the equivalent block in the lock. A stale ACP worker (e.g., from a step interrupted by a user) can read a matching step_seq, be descheduled, and then a new `/step` increments step_seq and sets `generating=True` before the stale worker resumes and sets `generating=False`. This clears the new step's reservation, causing the server to report idle while a generation is active, which can allow a second `/step` to start concurrently or make the WebUI think the turn has completed. The fix is to hold `session.step_lock` across the generation-flag clear and the `step_complete` add_event, matching the pattern used in `step()`.

gptme/server/session_step.py
```diff
@@ -1020,13 +1059,19 @@ def _flush_sse_batch() -> None:
...
                     "step() finally: skipping generating=False — "
                     "reservation transferred to continuation (seq %d→%d)",
                     my_step_seq,
                     session.step_seq,
                 )
+        # Emit step_complete AFTER generating=False so clients see the released
+        # state. Only emit when this step owns the reservation; a continuation
+        # step will emit its own step_complete when it finishes.
+        if _step_released:
+            SessionManager.add_event(conversation_id, {"type": "step_complete"})
```
TimeToBuildBob: ⚠️ **P2** — `step_complete` is emitted after `step_lock` is released, so a newer reservation can slip in between the compare-and-clear and the `add_event` call. Step A's finally sets `session.generating = False` and `_step_released = True` under the lock, then unlocks; before `SessionManager.add_event` executes, a concurrent `/step` (or a tool continuation electing a new epoch) acquires the lock, increments `session.step_seq`, and sets `generating = True`. Step A then emits `step_complete` unconditionally because `_step_released` was already captured. The WebUI receives the stale `step_complete` and its `onStepComplete` else branch calls `setGenerating(false)`, clearing the UI's busy state while step B is actively generating. A queued message can flush during this window and get a 409, and the false idle state can suppress feedback for the active generation. The emission should re-check `session.step_seq == my_step_seq` (or be performed atomically with the clearing) so a stale worker never announces completion for an epoch a newer step has taken over.

webui/src/hooks/useConversation.ts
```diff
@@ -284,10 +289,17 @@ export function useConversation(conversationId: string, serverId?: string) {
...
-              // Increased from 0ms to give API events more breathing room
-              setTimeout(() => {
+              // Fallback: clear generating state after 100ms to allow onToolPending to
+              // fire first. On updated servers this timer is cancelled and superseded by
+              // onStepComplete (which fires after generating=False on the server), closing
+              // the race window between generation_complete and generating=False entirely.
+              if (generationCompleteTimerRef.current !== null) {
+                clearTimeout(generationCompleteTimerRef.current);
+              }
+              generationCompleteTimerRef.current = setTimeout(() => {
```
TimeToBuildBob, line 305: ⚠️ **P2** — onMessageComplete schedules a 100ms fallback timer stored in generationCompleteTimerRef, and onStepComplete cancels it only when a step_complete arrives within that window. Nothing cancels the timer when the subscribing effect tears down: the effect's cleanup (lines 570-576) closes the event stream only when document.hidden, and the new ref is never cleared on unmount or conversation switch. If onMessageComplete fires and the user navigates to a different conversation (or the stream reconnects) within the next 100ms, the stale closure still executes setGenerating(conversationId, false), playChime(), speakText(), and notifyGenerationComplete(), producing an audible completion chime, TTS, and a browser notification for a conversation the user has already left; the notification name comes from the stale conversation$. This is a spurious-completion path distinct from the message-loss race covered elsewhere, and the PR's own timer-management ref introduced the cleanup obligation without honoring it.

gptme/server/a2a_api.py
```diff
@@ -439,17 +439,18 @@ def _run_task_blocking(task_id: str, session: ConversationSession) -> None:
...
-                    # generation_complete fires BEFORE session.generating = False
-                    # (the finally block runs after the event). Wait briefly for
-                    # generating to settle so _task_from_conversation sees the
-                    # correct state instead of TASK_STATE_WORKING.
-                    for _ in range(10):  # ~1s max
-                        if not session.generating:
-                            break
-                        session.event_flag.clear()
-                        session.event_flag.wait(timeout=0.1)
+                if event_type == "step_complete":
```
TimeToBuildBob, line 442: ⚠️ **P2** — The A2A path starts the in-process step() with auto_confirm=False, so a response that contains tool uses leaves those tools pending on the session. In-process step()'s finally now emits step_complete whenever it clears the generating reservation (session_step.py lines 1051-1074) regardless of whether pending_tools is non-empty, and _run_task_blocking returns immediately on the first step_complete, after which _task_from_conversation sees generating=False and reports TASK_STATE_COMPLETED. The comment block above _run_task_blocking still promises that tasks requiring tools 'return TASK_STATE_WORKING; clients should poll with GetTask', which is now deterministically false: a message that elicits a tool call is returned to the A2A client as COMPLETED with the assistant's partial response and no indication that a tool awaits confirmation. Previously this outcome only occurred when generating happened to clear within the ~1s poll; emitting step_complete unconditionally on release makes it guaranteed even while tools are still pending.

gptme/server/session_step.py
```diff
@@ -445,8 +454,12 @@ async def _acp_step(
                     "error": "Internal error: ACP runtime not initialized",
                 },
             )
-            session.generating = False
-            session.generating_since = None
+            # No generation was ever started, so don't emit step_complete: the
+            # WebUI's onError handler already clears isBusy for "error" events,
+            # and step_complete would trigger the completion chime/TTS/notify.
+            if session.step_seq == my_step_seq:
```
TimeToBuildBob: ⚠️ **P2** — In `_acp_step`, the three early-exit error paths (ACP runtime not initialized, no user message, no new user message) clear `session.generating` when `session.step_seq == my_step_seq`, but do not emit a `step_complete` event. The PR's stated contract (and the `StepCompleteEvent` docstring) is that `step_complete` fires after `session.generating` is released, and the PR description explicitly lists `_acp_step()` early-exit error paths as places where that event is emitted. The in-process `step()` early-exit error paths do emit it. A client that gates queued-message flushes or turn-finalization solely on `step_complete` — the intended contract — would never observe the release on these ACP paths and would remain stuck in a busy state. The WebUI also handles `error` events, so its own UX recovers, but a strict `step_complete`-gated client (e.g., a future non-WebUI consumer) would hang. Emit `step_complete` after clearing `generating` in these branches for consistency with the documented contract and with the `step()` early exits.

webui/src/hooks/useConversation.ts
TimeToBuildBob: Fixed in 683937854 — once `api.sendMessage` has persisted the message, a 409 from `api.step` no longer removes it or rethrows for a requeue; `step()` is retried with backoff (250ms→2s) and the message is never re-sent. Only a 409 before persistence still removes + requeues.

webui/src/components/ChatInput.tsx
TimeToBuildBob: Fixed in 683937854 — `ChatInput` now re-arms the queue flush on a 1s timer after any 409 requeue (queue-flush and direct-send paths), so a retry no longer depends on a future busy→idle transition.

gptme/server/session_step.py
TimeToBuildBob: Fixed in 683937854 — `_acp_step`'s finally block now does the compare-and-clear and the `step_complete` emit under `session.step_lock`; the three early-exit paths got the same treatment.

gptme/server/session_step.py
TimeToBuildBob: Fixed in 683937854 — `step_complete` is now emitted inside `step_lock` together with the release (also for the two early-exit paths in `step()`), so no new reservation can interleave between release and event.

gptme/server/session_step.py
TimeToBuildBob: Fixed in 683937854 — the three ACP early-exit paths now release under `step_lock` and emit `step_complete`, matching `step()` and the `StepCompleteEvent` contract. The old 'no chime' rationale was stale: `onStepComplete`'s else-branch (no preceding `generation_complete`) only clears `generating`. Updated the two ACP tests that asserted `error` as the last event.

gptme/server/a2a_api.py
TimeToBuildBob: Not introduced by this PR, so declining here. `_task_from_conversation` derives `WORKING` solely from `session.generating`; with `auto_confirm=False` a tool-call response already ended the step with `generating=False` before this PR, and the old `generation_complete` handler returned after the ≤1s settle wait, yielding the same `COMPLETED` state. `step_complete` only makes the previously timing-dependent outcome deterministic. Reporting pending tools (e.g. `TASK_STATE_INPUT_REQUIRED`) is a separate A2A-semantics change — the stale "tasks that require tools return TASK_STATE_WORKING" comment predates this branch and belongs with that follow-up.

webui/src/hooks/useConversation.ts
TimeToBuildBob: Not a bug: the timer closure is bound to the conversation that produced the completion, so its `setGenerating(conversationId, false)`, chime and notification are correct for a generation that really finished — the event stream is deliberately left open on conversation switch (cleanup only closes it when `document.hidden`), so the `step_complete` path fires identically for a conversation the user has left. Cancelling the timer in the effect cleanup would be worse: that effect re-runs on `toast`/`api`/`retryNonce` changes, and on servers that never emit `step_complete` it would leave `isGenerating` stuck true.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -589,6 +592,8 @@ async def _on_acp_update(_session_id: str, update: Any) -> None:
             acp_runtime.set_on_update(None)
             session.generating = False
             session.generating_since = None
+            # Emit step_complete AFTER generating=False so clients see consistent state
+            SessionManager.add_event(conversation_id, {"type": "step_complete"})
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4868da72b — `_acp_step`'s finally block now holds `session.step_lock` across both the compare-and-clear (`if session.step_seq == my_step_seq: session.generating = False`) and the `step_complete` emission. Early-exit error paths got the same treatment in 683937854. A stale ACP worker can no longer clear a replacement step's reservation because it must hold the lock when `step_seq` is checked — a newer reservation will already have incremented `step_seq` past `my_step_seq`.

webui/src/hooks/useConversation.ts
```diff
@@ -650,6 +691,12 @@ export function useConversation(conversationId: string, serverId?: string) {
         setTopP(conversationId, options?.topP);
       } catch (error) {
         console.error('Error sending message:', error);
+        // 409 means the server is still generating — remove the optimistic message
+        // and re-throw so the caller (e.g. ChatInput queue flush) can requeue it.
+        if (ApiClientError.isApiError(error) && error.status === 409) {
+          removeMessage(conversationId, userMessage.timestamp!);
+          throw error;
+        }
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 683937854 — `sendMessage` now tracks a `messageSent` flag set after `api.sendMessage` persists the message. A 409 from `api.step` (post-persistence) no longer removes the optimistic copy or rethrows for a requeue; instead `step()` is retried with 250ms→2s backoff. Only a 409 from `api.sendMessage` itself (pre-persistence) still removes + requeues. The direct-send path can no longer lose an already-persisted message.

TimeToBuildBob:
@greptileai review

webui/src/hooks/useConversation.ts
```diff
@@ -627,29 +672,51 @@ export function useConversation(conversationId: string, serverId?: string) {
...
+              options?.maxTokens,
+              options?.temperature,
+              options?.topP
+            );
+            break;
+          } catch (stepError) {
+            const busy = ApiClientError.isApiError(stepError) && stepError.status === 409;
+            if (!busy || attempt >= STEP_409_BACKOFF_MS.length) throw stepError;
+            await new Promise((resolve) => setTimeout(resolve, STEP_409_BACKOFF_MS[attempt]));
+            if (generationIsStale(epoch)) return;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0ebf4d888. Added `pendingStepRef` — when 409 retries exhaust post-persistence, we park the step closure and fire it from `onStepComplete` (the authoritative server-idle signal) instead of rethrowing. The message stays in 'sent' state and generation starts once the server is free. The closure captures the epoch so a stale-generation guard prevents racing with a newer send.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Bob PM status update

All code review findings addressed. Summary of work done in this session:

**Commits**
- `683937854` — Fixed prior-session P1s: `_acp_step` step_lock guard, `messageSent` flag for direct sends, ChatInput 1s re-arm after 409 requeue
- `97de44b4c` — Clarified misleading comment on `onStepComplete` else-branch
- `0ebf4d888` — Fixed Greptile P1 "Retry window drops generation": when post-persistence 409 retries exhaust (3.75s), the message was falsely marked 'failed'. Now parks the step in `pendingStepRef` and fires it from `onStepComplete` (the authoritative server-idle signal), keeping the message in 'sent' state.

**Current state**
- All Greptile review threads resolved
- Greptile re-review in-progress for head `0ebf4d888`
- Local AI review running to update the marker for the new head

**Remaining blocker**: touches `gptme/server/` — requires your manual review/merge (the server-path gate is intentional and cannot be bypassed by Bob).

webui/src/hooks/useConversation.ts
```diff
@@ -299,6 +317,48 @@ export function useConversation(conversationId: string, serverId?: string) {
...
+                // onToolPending already cleared generating, making this a no-op.
+                setGenerating(conversationId, false);
+              }
+              // Fire any step that was deferred because the retry window exhausted
+              // while the server was still busy.  step_complete is the authoritative
+              // idle signal, so the deferred call is safe to make now.
+              const pendingStep = pendingStepRef.current;
+              if (pendingStep) {
+                pendingStepRef.current = null;
+                void pendingStep();
```
TimeToBuildBob: ❌ **P1** — In `onStepComplete`, the deferred `pendingStep` is invoked with `void pendingStep()` and no `.catch()`. When `sendMessage` exhausts the 409 backoff, it stores a closure in `pendingStepRef` and returns without marking the message failed; the message stays in the log as 'sent'. When `step_complete` later arrives, this closure calls `api.step` again. If the server is still busy (e.g., a new step from another client started between the backoff exhaustion and the event), that call rejects with a 409 (or any error) and the rejection is unhandled – there is no toast, no retry, and the user's message remains in the log with no assistant response. This is a silent message-loss path, which the PR's 'lossless requeue' goal is meant to prevent. Attach a catch handler to the pending step and on another 409 either re-defer it or surface a failure, or the message is lost forever. _(2/3 passes — pass 1 looked and disagreed)_

webui/src/components/ChatInput.tsx
```diff
@@ -1038,7 +1065,26 @@ export const ChatInput: FC<Props> = ({
...
+          const status =
+            error && typeof error === 'object' && 'status' in error
+              ? (error as { status: number }).status
+              : undefined;
+          if (status === 409) {
+            // Server was still generating (concurrent client). The optimistic
+            // message was already removed by sendMessage; requeue it so it's
+            // retried on the next isBusy -> false transition instead of lost.
+            console.warn('[ChatInput] Direct send got 409, requeuing', sentMessage);
+            setMessageQueue((prev) => [{ text: sentMessage, options: sentOptions }, ...prev]);
```
TimeToBuildBob, line 1081: ⚠️ **P2** — In the direct-send 409 requeue path, `ChatInput` puts the original `sentOptions` back onto the queue. Those options still contain `pendingFiles` (raw `File` objects). `sendMessage` in `useConversation` uploads those files via `api.uploadFiles` before calling `api.sendMessage`; when the 409 arrives from `sendMessage`, the upload already happened and the files exist on the server, yet the requeued message will upload them again on the next flush, creating duplicate files. The same flaw exists in the queued-flush 409 path (line 945). The retry should reuse the file paths returned by the first upload (e.g., by passing the uploaded paths in the options instead of the raw File objects), or the upload should be deferred until after a successful message send. _(2/3 passes — pass 1 looked and disagreed)_

webui/src/hooks/useConversation.ts
```diff
@@ -299,6 +317,48 @@ export function useConversation(conversationId: string, serverId?: string) {
...
+              }
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 62e197405 — `pendingStep()` now has a `.catch()` handler: a 409 re-parks the closure into `pendingStepRef.current` (so the next `step_complete` retries), and any other error shows a destructive toast via `getApiErrorPresentation`. The deferred call can no longer silently drop a persisted message.

webui/src/hooks/useConversation.ts
TimeToBuildBob: Fixed in 62e197405 — `pendingStep()` now has a `.catch()` handler: a 409 re-parks the closure into `pendingStepRef.current` (so the next `step_complete` retries), and any other error shows a destructive toast via `getApiErrorPresentation`. The deferred call can no longer silently drop a persisted message.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## Session merge recommendation (2026-09-18)

**Convergence cap hit after 3 rounds** — adjudicating remaining findings against head `62e197405`.

### Fixed this session

| Finding | Severity | Fix |
|---------|----------|-----|
| Deferred generation can be dropped (fp `7faa364a75dc` / Greptile thread [#4049674280](https://github.com/gptme/gptme/pull/3874#discussion_r4049674280)) | **P1** | `62e197405` — `pendingStep().catch()` re-parks on 409 (waits for next `step_complete`), shows destructive toast on other errors. Silent message-loss path closed. |

### Remaining open findings

| Finding | Severity | Disposition |
|---------|----------|-------------|
| `665fc188208b` — ChatInput direct-send requeue passes raw `File` objects in `sentOptions` | P2 | Non-blocking: `sendMessage` calls `uploadFiles` before `sendMessage` on each retry, so files are re-uploaded (not double-sent). The field shape is consistent with how the queue path already works. Declining on this head; separate issue if stale-file handles become a real problem. |

### Verification

- **Fixed against**: `62e197405` (Greptile re-reviewed this head; 6 reviews total)
- **CI**: All checks green except `Test without API keys` (pending — known flaky; unrelated to this PR's changes)
- **Greptile score**: 4/5 after fix (was 2/5 at start of session)

### Domain risk

Server-side step coordination (`gptme/server/session_step.py`) with a new `step_complete` event emitted inside `step_lock`. The epoch-guard and lock placement are correct, but this code path runs on every server response — worth a maintainer smoke-test on the multi-client / concurrent-step edge case.

### Convergence note

`round_convergence.stable_rounds = 0` — each of the 3 rounds found a new P1, which were all fixed. No stable (unchanged) blocking round reached. Per policy, not re-triggering; leaving 4/5 for human merge judgment.

webui/src/components/ChatInput.tsx
TimeToBuildBob: Accepted trade-off, not fixing in this PR. The re-upload only happens when the 409 arrives from `api.sendMessage` *before* the message is persisted (rare: another client won the reservation between the busy check and the POST). In that case the first upload's files are orphaned server-side and the retry uploads them again — wasted bytes, not a lost or duplicated message. Reusing the first upload's paths means threading uploaded-path state through the queue options (`QueuedMessage` shape + `sendMessage` contract), which is a separate change to the upload path and out of scope for closing the step_complete race. If duplicate uploads show up in practice, it deserves its own issue.

webui/src/hooks/useConversation.ts
```diff
@@ -627,29 +699,71 @@ export function useConversation(conversationId: string, serverId?: string) {
...
+          } catch (stepError) {
+            const busy = ApiClientError.isApiError(stepError) && stepError.status === 409;
+            if (!busy) throw stepError;
+            if (attempt >= STEP_409_BACKOFF_MS.length) {
+              // Retry window exhausted; message is already persisted, so re-sending
+              // would duplicate it.  Instead, park the step call and fire it from
+              // onStepComplete — the authoritative server-idle signal — so the
+              // message is not falsely marked failed.
+              const capturedEpoch = epoch;
+              pendingStepRef.current = () => {
```
TimeToBuildBob, line 735: ❌ **P1** — If two messages are sent in quick succession and both exhaust their step-retry window, the second `sendMessage` call overwrites `pendingStepRef.current` with its own closure, discarding the first message's parked step. When a later `step_complete` arrives, only the last parked step is fired; the first message is left for the second step to process implicitly from the full log. That loses the first message's explicit options (model, temperature, etc.) and, if the second step is later made stale by an interrupt or a newer action, neither message gets processed — the first message is marked 'sent' but no generation ever runs for it. The single-slot ref should be a queue of deferred steps so each message's step is preserved. _(2/3 passes — pass 1 looked and disagreed)_

webui/src/components/ChatInput.tsx
```diff
@@ -910,6 +910,16 @@ export const ChatInput: FC<Props> = ({
   // tool execution/confirmation) completes.
   const wasBusy = useRef(false);
 
+  // After a 409 requeue the busy→idle transition may already have happened (the
+  // 100ms fallback timer can beat the server's release), so a transition-driven
+  // retry would never fire. Re-arm the flush effect on a timer instead.
+  const scheduleQueueRetry = () => {
+    setTimeout(() => {
```
TimeToBuildBob, line 917: ⚠️ **P2** — ChatInput's scheduleQueueRetry creates a setTimeout but never stores or clears the timer. If the component unmounts before the 1s retry fires (e.g. the user navigates to another conversation or page), the callback still runs and calls setMessageQueue on an unmounted component, producing a React state-update-on-unmounted warning and a potential no-op update. Additionally, the un-cleared timer can fire after the chat is remounted with a different conversation, re-arming wasBusy and potentially flushing a stale queue. The timer should be stored in a ref and cleared in a useEffect cleanup on unmount. _(2/3 passes — pass 3 looked and disagreed)_

webui/src/hooks/useConversation.ts
TimeToBuildBob: Accepted trade-off. The scenario needs two messages to both exhaust the 3.75s retry window and park back to back. Even then no message is lost: the surviving parked step calls `api.step`, which generates from the full persisted conversation log, so the first message still gets a response — it only loses its own per-call options (model/temperature/top_p). And if the surviving step is stale because the user interrupted or sent something newer, `generationIsStale` skipping it is the intended behaviour, not a drop. A queue of parked steps would replay N `step()` calls for messages one generation already covers. Not worth the machinery here.

webui/src/components/ChatInput.tsx
TimeToBuildBob: Not a bug. The retry callback only calls `setMessageQueue`, which is component-local state: React 18 no longer warns on unmounted setState, and a remounted ChatInput starts with a fresh queue, so a stale timer cannot flush a stale queue into another conversation. Clearing it on unmount would add a ref and cleanup effect for a no-op.

PR #3879, opened 2026-09-19
feat(tools): add ordered ToolSpec.sensitivity for approval gates

TimeToBuildBob:
## What

Phase 0 of a per-tool approval gate for non-technical users: give `ToolSpec` an ordered severity level so a confirm hook can auto-approve safe calls and interrupt on risky ones, replacing the current all-or-nothing gear boolean.

```python
ToolSensitivity = Literal["safe", "moderate", "sensitive", "dangerous"]
```

Default is `"safe"`, so this is non-breaking — no existing tool changes behavior and no consumer is added here.

## Labels

| Level | Tools | Rationale |
|---|---|---|
| `dangerous` | `shell`, `computer`, `ipython` | Arbitrary code execution / machine control, not sandboxed |
| `sensitive` | `browser`, `gh`, `morph` | External service interaction: browser drives external sites, `gh` has external side effects (open/comment/merge), `morph` sends the target file's full contents to OpenRouter/Morph before the local write |
| `moderate` | `save`, `append`, `patch`, `patch_many`, `hashline_edit` | Local, reversible mutations |

## Why these levels (4, not 3)

The originating task paraphrased the design as `safe`/`sensitive`/`destructive`. I followed the design note instead, which defines four levels. A three-level scheme has to classify `save`/`patch` as *destructive* — they already carry that free-form `hints` tag — which would force an approval prompt on every local file write and reproduce exactly the prompt fatigue the feature exists to remove. `moderate` ("local mutations") is the level that makes a selective gear useful.

## `sensitivity` vs the existing `hints`

`hints` is a free-form allowlist tag set (`"read-only"`, `"file-ops"`, `"destructive"`), consumed by `tool_matches_allowlist`. `sensitivity` is a separate, *ordered* axis that a gate can compare against a threshold. A tool can legitimately be hinted `"destructive"` (it mutates state) while being only `"moderate"` to approve (the mutation is local and reversible) — that is the case for the file-mutation tools above. Both are kept; nothing was migrated.

## Subtool inheritance

`browser` and `computer` expose their functionality as `<parent>.<fn>` subtoolspecs. Without inheritance, a gate evaluating the subtool spec would read `"safe"` for every helper call, making the parent label ineffective. `as_function_subtoolspecs()` now propagates the parent level.

## MCP

MCP tools do get a `ToolSpec` (built in `mcp_adapter`), so the "unannotated = moderate" rule is implemented there rather than left to consumers:

- `readOnlyHint` → `safe`
- `destructiveHint` → `dangerous`
- unannotated → `moderate`

An MCP tool may have side effects we cannot see, so `"safe"` is not a safe default. Mirrors the existing annotation→hints mapping in the same block.

## Scope / not in this PR

- No core consumer of the field yet: this only adds metadata. The threshold evaluation lives in the managed service (`GEAR_CONFIRM_THRESHOLD`), which is phase 1.
- The remaining tools keep the `"safe"` default; the full labeling pass is phase 1, once a threshold consumer exists to verify against. Consumers must treat *absence of a ToolSpec entirely* (e.g. raw shell) as `"moderate"` — documented on the field.
- `email`/`calendar`/`payment` tools are not in core (they live in gptme-contrib), so they cannot be labeled here.

## Test plan

- `tests/test_toolspec_sensitivity.py`: default is safe; each level accepted; the field survives `dataclasses.replace()`; the built-in labels assert against module-level specs (availability-independent, since `browser` needs a backend and `computer` is `disabled_by_default`); function subtools inherit each level; MCP annotation mapping table.
- `pytest tests/test_toolspec_sensitivity.py tests/test_toolspec_read_only.py tests/test_mcp_adapter.py` → 89 passed.
- `ruff check` + `ruff format --check` clean; mypy clean on changed files; pre-commit green.

## Note on pre-existing master failure

`tests/test_tools.py::test_tool_descriptions_within_openai_limit` fails locally for `shell` (1030 chars > 1024). I verified this reproduces on unmodified `origin/master` (stashed my changes and re-ran) — it is not from this PR. Flagged here so a red CI on that job is not misattributed.


gptme/tools/base.py
```diff
@@ -515,6 +537,7 @@ class ToolSpec:
     is_mcp: bool = False
     hints: frozenset[str] = field(default_factory=frozenset)
     read_only: bool = False
+    sensitivity: ToolSensitivity = "safe"
```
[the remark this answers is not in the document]
TimeToBuildBob: Addressed the named cases in 3479735f8: `morph` → `moderate` (local, reversible file mutation — same class as `patch`) and `gh` → `sensitive` (external side effects: it can open PRs, comment, and merge).

I'm keeping the `"safe"` default deliberately in this PR: the field is new and has no consumer yet, so nothing is auto-approvable today, and flipping the default to `"moderate"` would simultaneously reclassify unlabeled read-only tools (`read`, `rag`, `vision`, `screenshot`) as approval-worthy. That's why the full labeling pass is phase 1: it must land before `GEAR_CONFIRM_THRESHOLD` is wired. The field docstring already requires consumers to treat a missing `ToolSpec` entirely as `"moderate"`, and the phase-1 pass closes the same gap for present-but-unlabeled specs and third-party plugins.

gptme/tools/base.py
```diff
@@ -57,6 +57,20 @@
...
+#   moderate:  local mutations (file writes, patches, git commits)
+#   sensitive: external side effects (email, calendar, API mutations, browsing)
+#   dangerous: irreversible or high-blast-radius (arbitrary exec, payments, deletes)
+#
+# This is a separate axis from `hints`: hints are free-form allowlist tags
+# ("read-only", "file-ops", "destructive"), whereas sensitivity is the ordered
+# severity an approval gate can compare. A tool may be hinted "destructive"
+# (it mutates state) while being only "moderate" to approve (the mutation is
+# local and reversible).
+ToolSensitivity: TypeAlias = Literal["safe", "moderate", "sensitive", "dangerous"]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 3479735f8: added `SENSITIVITY_ORDER` and `sensitivity_rank()` to `gptme/tools/base.py`, with a test asserting the rank follows severity order rather than the literal string order (alphabetically `"dangerous" < "safe"`, which would invert a gate's decision). Consumers should compare through `sensitivity_rank()` instead of duplicating the mapping.

TimeToBuildBob:
@greptileai review

gptme/tools/morph.py
```diff
@@ -259,6 +259,8 @@ def execute_morph_impl(
     execute=execute_morph,
     block_types=["morph"],
     available=is_openrouter_available,
+    # Local, reversible file mutation — same class as patch/save, not a read.
+    sensitivity="moderate",
```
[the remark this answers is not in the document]
TimeToBuildBob: Good catch — you're right. `execute_morph_impl` reads the full target file and sends it (plus the edit instructions) to `openrouter/morph/morph-v3-fast` before the local write, so a gate that auto-approves `moderate` local mutations would also auto-approve external disclosure of workspace data.

Fixed in 6ab6ee0d9: `morph` is now `sensitive`, in the same class as the other external-interaction tools (`browser`, `gh`). The builtin-sensitivity test is updated and passes.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Confidence 2/5

<details><summary><b>2 findings</b> · ❌ 2 P1</summary>

**❌ P1 high** — `gptme/tools/base.py:583`

Defaulting every unannotated ToolSpec to `safe` makes legacy and third-party tools indistinguishable from tools explicitly reviewed as safe. Existing plugins can perform arbitrary external actions or code execution while omitting this newly added argument, so a threshold consumer will auto-approve them. The docstring's suggestion to trust only explicit labels cannot be implemented because the constructor does not retain whether the default was implicit. Use an unknown/moderate default (or otherwise track explicit annotation) and explicitly label known read-only tools.

**❌ P1 high** — `gptme/tools/mcp_adapter.py:218`

The MCP mapping classifies every explicitly non-destructive, non-read-only tool as `moderate`, without considering `openWorldHint`. Under the new level definitions, `moderate` is for local mutations while external side effects are `sensitive`; an MCP tool that creates an external resource or sends a message can correctly advertise `destructiveHint=False` and `openWorldHint=True` and will therefore bypass a gate configured to interrupt at `sensitive`. Map open-world mutating tools to `sensitive` and reserve `moderate` for closed-world mutations.

</details>

| `6ab6ee0d9004` | 4/5 | 1 | llm | 2026-09-19 04:02 UTC |
| `999242247ee4` | 3/5 | 1 | llm | 2026-09-19 04:05 UTC |
| `f327ca4cda17` | 2/5 | 3 | llm | 2026-09-19 06:10 UTC |

</details>

<sub>Reviewed `fcadfbdbc99d` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 320s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tests/test_toolspec_sensitivity.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.mark.parametrize("tool_name", ["read", "rag", "vision", "screenshot"])
+def test_read_only_tools_are_not_gated(tool_name: str):
+    """Read-only tools must stay at the default 'safe' level."""
+    from gptme.tools import get_available_tools
+
+    matches = [t for t in get_available_tools() if t.name == tool_name]
+    assert matches, f"tool {tool_name!r} not found"
+    assert matches[0].sensitivity == "safe"
+
+
```
TimeToBuildBob: ⚠️ **P2** — The test test_read_only_tools_are_not_gated calls get_available_tools() and asserts that tools named "read", "rag", "vision", "screenshot" have sensitivity == "safe". This test depends on the runtime environment: get_available_tools() only returns tools whose availability check passes (e.g. browser requires a backend, computer is disabled_by_default, rag may require a backend). If any of these tools is not available in the test environment, the assertion `assert matches, f"tool {tool_name!r} not found"` fails, making the test non-hermetic. The test should either use the module-level ToolSpec objects directly (as the BUILTIN_SENSITIVITY test does) or skip when the tool is unavailable. This is a test defect: it will fail in environments where, e.g., the rag tool is not available, even though the code is correct. _(2/3 passes — pass 1 looked and disagreed)_

TimeToBuildBob:
@greptileai review

gptme/tools/mcp_adapter.py
```diff
@@ -196,6 +197,24 @@ def _call_mcp_tool_with_retry(
...
+) -> ToolSensitivity:
+    """Map MCP ToolAnnotations onto ToolSpec.sensitivity.
+
+    An unannotated server — or one that declares itself neither read-only nor
+    destructive — is treated as ``"moderate"``: an MCP tool may have side
+    effects we cannot see, so ``"safe"`` is not a safe default.
+    """
+    if annotations is None:
+        return "moderate"
+    if annotations.readOnlyHint:
```
TimeToBuildBob, line 213: ❌ **P1** — sensitivity_from_annotations() treats an MCP tool with readOnlyHint=True as "safe" even when destructiveHint is also True, because the readOnlyHint check returns before the destructiveHint check. The test suite explicitly pins this behavior: (_annotations(readOnlyHint=True, destructiveHint=True), "safe"). Per the MCP spec, readOnlyHint and destructiveHint are independent hints; a tool can be both read-only and destructive (e.g. a tool that deletes a resource but returns nothing). Labeling such a tool "safe" means an approval gate comparing against a threshold would auto-approve a destructive operation. The existing hints mapping in the same block already handles this correctly: it only adds "destructive" when not ann.readOnlyHint. The sensitivity mapping should mirror that: readOnlyHint should only win when destructiveHint is not True. _(3/3 passes)_

TimeToBuildBob: Fixing the premature resolution here: an explicit `destructiveHint=True` now wins over `readOnlyHint=True` — fixed in f327ca4cd (mapping checks `destructiveHint is True` before the `readOnlyHint` short-circuit, mirroring the hints logic; test updated to pin `(readOnlyHint=True, destructiveHint=True) → "dangerous"` and `(True, False) → "safe"`).

TimeToBuildBob:
@greptileai review

tests/test_toolspec_sensitivity.py
TimeToBuildBob: Fixed in 999242247 — the test now asserts against the module-level `ToolSpec` objects (`gptme.tools.read.tool`, `gptme.tools.rag.tool`, `gptme.tools.vision.tool`, `gptme.tools.screenshot.tool`) instead of `get_available_tools()`, same pattern as `test_builtin_tool_sensitivity`, so it is hermetic regardless of backend availability.

gptme/tools/gh.py
```diff
@@ -766,6 +766,8 @@ def render(args: list[str]) -> str:
     examples=examples,
     execute=execute_gh,
     block_types=["gh"],
+    # External side effects: can open PRs, comment, and merge — not read-only.
+    sensitivity="sensitive",
```
TimeToBuildBob, line 770: ❌ **P1** — The gh tool is labeled sensitivity='sensitive', but its execute_gh() falls through to _passthrough_gh() for any unrecognized command, which runs subprocess.run(['gh'] + list(args)) with no allowlist. This permits arbitrary gh CLI invocations, including `gh api -X DELETE /repos/...` which can irreversibly delete a repository, force-push, or modify repo settings — operations that match the PR's own 'dangerous' definition (irreversible or high-blast-radius, arbitrary exec/payments/deletes). A future approval gate comparing sensitivity to a threshold will treat gh as an external-interaction tool rather than a dangerous one, auto-approving destructive API calls that should require explicit human confirmation in non-interactive sessions. The native handlers (pr merge, run view) are only a subset of what this tool can do; the pass-through is the hole.

gptme/tools/base.py
```diff
@@ -538,6 +578,7 @@ def __init__(
         is_mcp: bool = False,
         hints: frozenset[str] | None = None,
         read_only: bool = False,
+        sensitivity: ToolSensitivity = "safe",
```
TimeToBuildBob, line 581: ⚠️ **P2** — Adding the `sensitivity` parameter to `ToolSpec.__init__` between `read_only` and `hooks` changes the positional argument index of every subsequent parameter. Any existing caller that constructs a `ToolSpec` with `hooks` or `commands` passed positionally will have that value bound to `sensitivity` instead, likely causing a runtime error when the field is later used (e.g. `sensitivity_rank` receives a dict or function object). While the in-repo constructions all use keyword arguments, the class is a documented public plugin API, so external plugins relying on the previous positional order are silently broken by this change, contradicting the PR's non-breaking claim.

gptme/tools/gh.py
TimeToBuildBob: Fixed in fcadfbdbc: `gh` is now `dangerous`, because its unrestricted pass-through can invoke destructive API and settings operations.

gptme/tools/base.py
TimeToBuildBob: Fixed in fcadfbdbc: `sensitivity` now follows the existing `hooks` and `commands` parameters, preserving every prior positional index. A regression test constructs `ToolSpec` with both legacy fields positionally.

gptme/tools/mcp_adapter.py
TimeToBuildBob: Fixed in f327ca4cd and retained in fcadfbdbc: explicit destructiveHint=True wins before readOnlyHint, with regression coverage for conflicting and omitted hints.

gptme/tools/gh.py
TimeToBuildBob: Fixed in fcadfbdbc: gh is classified dangerous because unrestricted pass-through can perform destructive API and settings operations.

gptme/tools/base.py
TimeToBuildBob: Fixed in fcadfbdbc: sensitivity now follows hooks and commands, preserving every prior positional parameter index; regression coverage constructs both legacy fields positionally.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI is green, Greptile is 5/5 on `fcadfbdbc`, and every review thread is resolved. I re-ran the self-merge gate; it still refuses this feature PR because it has no linked issue / prior maintainer product approval, so I did not bypass the gate. Ready for maintainer review and merge.

gptme/tools/base.py
```diff
@@ -540,6 +580,7 @@ def __init__(
         read_only: bool = False,
         hooks: dict[str, tuple[str, HookFunc, int]] | None = None,
         commands: dict[str, Callable] | None = None,
+        sensitivity: ToolSensitivity = "safe",
```
TimeToBuildBob, line 583: ❌ **P1** — Defaulting every unannotated ToolSpec to `safe` makes legacy and third-party tools indistinguishable from tools explicitly reviewed as safe. Existing plugins can perform arbitrary external actions or code execution while omitting this newly added argument, so a threshold consumer will auto-approve them. The docstring's suggestion to trust only explicit labels cannot be implemented because the constructor does not retain whether the default was implicit. Use an unknown/moderate default (or otherwise track explicit annotation) and explicitly label known read-only tools.

gptme/tools/mcp_adapter.py
```diff
@@ -196,6 +197,27 @@ def _call_mcp_tool_with_retry(
...
+    """
+    if annotations is None:
+        return "moderate"
+    if annotations.destructiveHint is True:
+        return "dangerous"
+    if annotations.readOnlyHint:
+        return "safe"
+    if annotations.destructiveHint is not False:
+        return "dangerous"
+    return "moderate"
```
TimeToBuildBob, line 218: ❌ **P1** — The MCP mapping classifies every explicitly non-destructive, non-read-only tool as `moderate`, without considering `openWorldHint`. Under the new level definitions, `moderate` is for local mutations while external side effects are `sensitive`; an MCP tool that creates an external resource or sends a message can correctly advertise `destructiveHint=False` and `openWorldHint=True` and will therefore bypass a gate configured to interrupt at `sensitive`. Map open-world mutating tools to `sensitive` and reserve `moderate` for closed-world mutations.

ErikBjare:
@TimeToBuildBob closing this, not sure it provides any value for us. Just seems like yet another gear-like mechanism which we realistically won't use (just like gears, which should arguably be removed too).

PR #3880, opened 2026-09-19
fix(patch_many): accumulate repeated paths instead of silently dropping hunks

TimeToBuildBob:
## Problem

`patch_many` accepted one `{"path": ..., "patch": ...}` object per change. When the
same path appeared twice, every entry patched the pristine on-disk text (all reads
happened before the deferred write loop), so the later write clobbered the earlier —
while the tool reported `Applied N patch(es) atomically to:` and listed the path N
times. **Silent partial application with a success message.**

Reproduced 2026-09-19 while adding `ToolSpec.sensitivity`: five `base.py` entries
landed as one, two `save.py` entries landed as one. Only an import probe caught it
(`NameError`), not any error from the tool.

## Fix

Apply each entry to the working content built by earlier entries for that path,
instead of re-reading the pristine file for every entry. Each subsequent hunk for a
path now sees the accumulated text.

The resolve-all-before-writing atomicity guarantee is preserved: reads/validation are
still completed for all entries before any write, and rollback still uses the first
on-disk read per path.

The success message now reports hunks applied and files written separately, so it can
never claim more patches than landed:

```
Applied 3 patch(es) atomically to 1 file(s):
  - /path/to/file.py
```

## Tests

Four regression tests added:

- `test_duplicate_paths_accumulate_both_hunks` — three entries for one path all land
- `test_duplicate_paths_message_counts_surviving_patches` — count/listing can't overclaim
- `test_duplicate_paths_atomic_when_later_hunk_fails` — a failing later hunk for an
  accumulated path aborts and writes nothing
- `test_duplicate_paths_via_kwargs_round_trip` — the kwargs entrypoint accumulates too

The first three of these fail on unpatched `master` (verified by reverting the source
change), confirming they cover the actual bug.

## Verification

- `pytest tests/test_tools_patch.py tests/test_tools_patch_many.py tests/test_tools_patch_anchored.py` → **85 passed**
- `ruff check` + `ruff format --check` clean
- `prek` pre-commit hooks green

Found while working on gptme/gptme#3879; that PR did not touch this code path.


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `tests/test_tools_patch_many.py` (+77/-0, Δ77)
- `gptme/tools/patch_many.py` (+38/-23, Δ61)

### Suggested verification
- `uv run pytest tests/test_tools_patch_many.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3880._

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Pushed `976a2c29e` — shortened the `patch_many` tool instructions from 1331 → 949 chars.

The repeated-path guidance added in the previous commit pushed the tool description over OpenAI's 1024-char function-description limit, which failed CI:

- `test_tool_descriptions_within_openai_limit` — `patch_many (1331)`, limit 1024
- `test_get_prompt_full` — prompt 8260 > 8250 ceiling (knock-on from the longer instructions)

Trimmed the two format examples (one path each) and condensed the prose, keeping the multi-hunk and repeated-path guidance that addresses the earlier Greptile P2. `tests/test_tools_patch_many.py` 27/27 pass locally; `ruff check` / `ruff format --check` clean.

gptme/tools/patch_many.py
```diff
@@ -203,31 +197,46 @@ def execute_patch_many_impl(
...
         try:
+            # Count hunks: a Patch is one hunk; a codeblock string may hold
+            # multiple ORIGINAL/UPDATED blocks, each applying as its own hunk.
             if isinstance(patch_src, Patch):
-                new_content = patch_src.apply(original)
+                hunks += 1
+                new_content = patch_src.apply(current[path])
             else:
-                new_content = apply(patch_src, original)
+                hunks += patch_src.count(ORIGINAL)
```
TimeToBuildBob, line 238: ⚠️ **P2** — The hunk counter `hunks += patch_src.count(ORIGINAL)` counts every occurrence of the literal `<<<<<<< ORIGINAL` in the patch string, but `apply()` may expand placeholder markers like `# ...` into multiple replacements (see `_parse_patches_from_content` comment: 'A single patch block may legitimately expand to multiple replacements when `apply()` handles placeholder markers'). For a patch string with one ORIGINAL block containing a placeholder that expands to N replacements, the message reports 1 hunk while N replacements were applied. Conversely, if a patch string contains the literal text `<<<<<<< ORIGINAL` inside a code block being inserted (e.g., adding a conflict-marker example to a file), `count(ORIGINAL)` over-counts. The success message is user-facing and the PR's stated goal is that it 'can never claim more patches than landed'; this can both under- and over-claim depending on content.

gptme/tools/patch_many.py
```diff
@@ -203,31 +197,46 @@ def execute_patch_many_impl(
...
-
-        try:
-            original = path.read_text(encoding="utf-8")
-        except (UnicodeDecodeError, PermissionError, OSError) as e:
-            yield Message(
-                "system",
-                f"Atomic patch aborted: could not read `{path}`: {e}. No files were written.",
-            )
-            return
+        if path not in current:
```
TimeToBuildBob, line 210: ⚠️ **P2** — The new `current` dict is keyed by `Path` objects. `_resolve_path` returns `path_display.resolve()`, which normalizes the path. However, `execute_patch_many_impl` can be called directly with `Path` objects that are not resolved (as in the tests, `tmp_path / "file.py"`). If the same file is passed as two different `Path` objects that resolve to the same location (e.g., one relative and one absolute, or with `..`), `path not in current` will treat them as distinct keys, so the second entry will re-read the pristine file and clobber the first write. The PR's stated fix is to accumulate repeated paths, but it only works when the same `Path` object (or equal `Path` string) is used. In the kwargs path, `_resolve_path` normalizes, so duplicates from kwargs are caught. But the markdown simple format also resolves paths. The direct `execute_patch_many_impl` API is internal, but the tool's contract is that repeated paths accumulate; a caller passing `Path("a")` and `Path("./a")` would get silent clobbering again. This is a correctness gap in the accumulation logic.

TimeToBuildBob:
## 🤖 AI code review

### Needs a look — P2 only

Confidence 4/5

**⚠️ P2 medium** — `gptme/tools/patch_many.py:238`

The success message's hunk count is derived from the number of ORIGINAL markers in the raw patch text (`hunks += patch_src.count(ORIGINAL)` for string patches, and `hunks += 1` for `Patch` objects), but `apply()` can expand a single block containing `# ...` placeholders into multiple disjoint replacements. The comment at `_parse_patches_from_content` (lines 85-87) explicitly acknowledges this expansion. A simple-format patch like the one in `test_simple_format_placeholder_patch_round_trip` therefore reports `Applied 1 hunk(s)` while two or more substitutions are actually performed. The file content is correct; only the human-readable/automated success report undercounts the work done, which can mislead a user or any consumer that parses the tool result.

<sub>**How this was verified:** Checked the implementation: `patch_src.count(ORIGINAL)` counts all occurrences of the ORIGINAL marker, including any inside the original or updated content fields. `apply()` treats the first marker as a hunk start, so a patch whose original text contains `<<<<<<< ORIGINAL` yields count &gt; 1 for a single hunk. This is confirmed by the new tests, which only use marker-free content and pass.</sub>

</details>

| `976a2c29eedd` | 4/5 | 2 | llm | 2026-09-19 04:42 UTC |

</details>

<sub>Reviewed `976a2c29eedd` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm (in-band)` engine · 250s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Merged as `f65cb9050` (2026-09-19 04:56Z) with 18/18 checks green. This was an informational merge notification on my own PR — no further action needed.

The two P2 findings on hunk-count reporting (placeholder-expanded hunks, multi-block `Patch` objects) were deliberately left unfixed before merge: they affect only the human-readable success message, not file contents, and the fix itself notes the top-level-block counting semantics at `patch_many.py:85`.

PR #3877, opened 2026-09-18
test(shell): make SIGTERM regression independent of inherited SIG_IGN

TimeToBuildBob:
Fixes the flaky `tests/test_shell_sigterm.py::test_cli_sigterm_cleans_background_job[terminate|default-return]` failures that have turned master `Test` red twice today (runs 35340263104 at a0418fe3, 35382892924 at da6832bd).

**Cause.** `SIG_IGN` is inherited across `exec`. When a pytest xdist worker has SIGTERM ignored, the CLI child inherits it, `main()` sees `previous_sigterm != SIG_DFL` and (correctly) doesn't install its handler:
- `terminate` sends SIGTERM to a process that ignores it → `TimeoutExpired` after 10s
- `default-return` hits `assert callable(signal.getsignal(SIGTERM))` in the fake chat → empty `AssertionError` → swallowed by `main()`'s `Fatal error occurred` path → exit 1

`ignore` and `custom` still pass, which matches the CI pattern exactly (2 failures, after 3 attempts each, only when the worker's disposition is bad).

**Fix.** The child resets SIGTERM to `SIG_DFL` before the per-mode setup, so the test doesn't depend on the runner's signal state.

**Verified locally.** Running the file under a parent with SIGTERM set to `SIG_IGN`: before = `2 failed, 2 passed`; after = `4 passed`. Normal run: `4 passed`.

Not done here: identifying which earlier test leaks `SIG_IGN` into the worker (`gptme/eval/run.py` sets it in its subprocess paths, so that's the first place to look). The test is now hermetic either way.

PR #3885, opened 2026-09-19
test(eval): isolate process-global side effects of in-process act_process calls

TimeToBuildBob:
Follow-up root cause for the `test_shell_sigterm.py` flake that #3877 made hermetic. The fix in #3877 resets `SIGTERM` to `SIG_DFL` in the child; this removes the leak at its source.

## Root cause

`tests/test_eval.py` calls `act_process()` **in-process**:

```python
act_process(agent=..., sync_dict=..., parallel=True, suppress_output=True)
```

But `act_process` is written to run in a `multiprocessing` child (see `run_evals`). In the success, timeout and error paths it ends with:

```python
signal.signal(signal.SIGTERM, signal.SIG_IGN)
```

and at entry it calls `os.setpgrp()`. Both mutate **process-global** state. In a throwaway child that is intended; in a pytest worker it leaks:

- `SIG_IGN` is inherited across `execve`, so any later test that spawns a CLI child hands it a process that refuses `SIGTERM`. `gptme/cli/main.py` then sees `previous_sigterm != SIG_DFL`, correctly declines to install its handler, and `tests/test_shell_sigterm.py` fails in exactly the two modes #3877 documented (`terminate` → `TimeoutExpired`; `default-return` → non-callable disposition).
- `os.setpgrp()` moves the worker out of its own process group.

## Evidence

Bad ordering, before this PR (2 failed / 3 passed — the same signature as the master CI runs):

```text
$ pytest "tests/test_eval.py::test_act_process_error_includes_cost" tests/test_shell_sigterm.py -q
FAILED tests/test_shell_sigterm.py::test_cli_sigterm_cleans_background_job[terminate]
FAILED tests/test_shell_sigterm.py::test_cli_sigterm_cleans_background_job[default-return]
2 failed, 3 passed
```

A sentinel assertion placed after the `act_process` test:

```text
without the fixture: SIGTERM leaked as <Handlers.SIG_IGN: 1>; pgrp == pid (leaked)
with the fixture:    both restored
```

After this change:

```text
$ pytest tests/test_eval.py -q
31 passed, 2 skipped
```

## Fix

A small fixture that snapshots and restores the `SIGTERM` disposition and the process group around the three tests that call `act_process` in-process. `run_evals` is unaffected — it runs `act_process` in a child where the side effects are contained.

#3877 is the symptomatic defence and its test-side reset stays (it also guards against a leak from *any* test, not just these three); this removes the actual source so test order stops mattering.


TimeToBuildBob:
Closed loop: this PR was merged by @ErikBjare at 10:39:29Z (merge commit `f40b6dbf`).

No Bob action was needed or taken. For the record, the self-merge gate had correctly blocked at 10:36Z (`AI review required`) because the AI-review sweep's 10:21Z attempt for this head fast-failed on an upstream OpenRouter 429 — Greptile was fresh 5/5 at the time. The maintainer merged directly, so the review retry was skipped.

PR #3884, opened 2026-09-19
test(ci): smoke-test frozen gptme-server startup for missing modules

TimeToBuildBob:
## What

CI's PyInstaller job only ran `gptme-server --help`. That never reaches hook/plugin registration, and those modules are imported by string (`importlib.import_module`), so PyInstaller's static analysis can miss them and the frozen binary still passes `--help`. v0.34.0 shipped with 18 hook/plugin modules missing from the sidecar (fixed in #3883) and every check stayed green.

This adds `scripts/smoke_frozen_server.py`: start the frozen binary on a free loopback port with a clean `HOME` and no provider keys, wait for `/api/v2` to answer, and fail on any `No module named` in the startup log. Wired into `build.yml` after the `--help` step on Unix runners (Windows isn't in the matrix).

## Verified

Run locally with a negative and a positive control:

- Released v0.34.0 `gptme-server` (linux-x64): **fails**, 11 missing-module errors (`gptme.hooks.injection_screening`, `cache_awareness`, `workspace_agents`, ...), exit 1.
- Local PyInstaller build with the #3883 spec fix: **passes**, exit 0.

The CI run on this PR is the first exercise on a real runner; if an optional dependency legitimately logs `No module named` at startup there, that will show up in the `PyInstaller binaries` job and I'll narrow the pattern.

## Not covered

The Tauri E2E suite (`tauri/e2e`) runs against a mock `index.html` served on the dev port (`data-ready` attribute), not the real webui plus sidecar, so it can't catch first-run regressions like the `tauri://localhost` retarget fixed in #3882. That needs a separate change: build the real webui and sidecar into the debug app and drive Local -> Connect. Out of scope here.

PR #3886, opened 2026-09-19
test(tauri): real first-run E2E with webui + sidecar

TimeToBuildBob:
Replaces the mock-based E2E with an integration test that builds the real webui dist and PyInstaller sidecar, then verifies the SetupWizard Local -> Connect flow.

**What changed:**
- Add first_run.test.js: waits for the sidecar on `GPTME_SERVER_PORT`, drives the wizard, asserts a *connected* signal (not a stored URL)
- Update tauri.yml E2E job: build webui + sidecar (via spec) instead of placeholders
- Switch build-sidecar.sh to use gptme-server.spec (correct hiddenimports)
- Remove mock HTTP server; run smoke + first_run specs under xvfb
- Delete mock-based chat_flow.test.js (see "E2E coverage" below)

**E2E coverage note (replaces the earlier claim that chat_flow.test.js still covered chat):**
chat_flow.test.js is deleted, not retired behind a second job. It asserted button/input behaviour of a hand-written mock `index.html`, not of the shipped app, and had already left CI coverage when this PR switched the job to the real webui build. Restoring it needs a separate mock-frontend job (a second `cargo build`, ~4-5 min). Net effect: the repo currently has **no** Tauri E2E coverage for chat send/receive or conversation-title updates — desktop chat regressions are undetected until a replacement lands. Flagged as follow-up work, not part of this PR.

**Test-context constraint (why the spec looks the way it does):**
tauri-driver's WebDriver script context is isolated from the page — page globals (`window.__TAURI__`) are undefined there and `localStorage` access throws `SecurityError: The operation is insecure.` So the spec uses only WebDriver element commands plus `GPTME_SERVER_PORT`; it cannot read the server registry out of storage. The #3882 guard is therefore behavioural: with `tauri://localhost` retargeted, the connect fails and the connected-signal wait times out.

**Regression guard for:**
- #3882 (tauri://localhost retarget bug)
- #3883/#3884 (frozen sidecar missing hooks/modules)

**Not in scope:**
- Chat message send/receive (no coverage after chat_flow.test.js deletion — follow-up)
- Cloud setup, provider configuration, first reply generation
- macOS/Windows E2E (Linux only, same as current)

**Open question at current head:** the E2E reaches the wizard step but `button=Get started` does not exist — `WelcomeView` hides the wizard CTA by design for desktop Tauri with a managed local server (#3407), and the wizard did not auto-open in that run. See the review thread; next step is a `browser.getPageSource()` dump before changing selectors.

Fixes #3884


TimeToBuildBob:
@greptileai review

.github/workflows/tauri.yml
```diff
@@ -178,6 +172,17 @@ jobs:
           touch tauri/src-tauri/icons/icon.icns
           touch tauri/src-tauri/icons/icon.ico
 
+      - name: Build webui dist
+        run: |
+          cd webui
+          npm ci
+          npm run build
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9e50c926a: the retained chat-mock generation step is removed from the E2E job. The job now builds the real `webui/dist` and lets `cargo build` embed it via `frontendDist`, so the debug binary contains the real webui and `first_run.test.js` exercises the actual first-run flow. (Mock-based `chat_flow.test.js` can no longer run in this job — its header now says so; restoring it needs a separate mock-frontend job.)

tauri/e2e/test/specs/first_run.test.js
```diff
@@ -0,0 +1,143 @@
...
+    while (Date.now() < deadline) {
+      try {
+        const result = await browser.execute(
+          (url) =>
+            fetch(url, { method: "GET" })
+              .then((r) => r.status)
+              .catch(() => 0),
+          `http://127.0.0.1:${port}/api/v2/models`
+        );
+        if (result === 200) return;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 9e50c926a: the readiness probe now polls `/api/v2/server/health`, which is unauthenticated by design for liveness/readiness (gptme#3701), instead of the bearer-protected `/api/v2/models` that returns 401 to this request. Also broadened the post-Connect wait to accept the "Connected to server" indicator or the persisted loopback baseUrl, since the wizard may auto-advance past the "Continue" button.

TimeToBuildBob:
**Decision on `chat_flow.test.js`**: delete it (option 2).

The test asserts button/input behaviour of a hand-written mock `index.html`, not of the shipped app. It was already removed from CI coverage when this PR switched to the real webui build. Restoring it would cost a second `cargo build` (~4-5 min) for assertions that have low product value.

The right replacement is a real send/receive chat E2E against the sidecar, which is out of scope for this PR. I'll file that as a follow-up.

Will add a commit to remove `tauri/e2e/test/specs/chat_flow.test.js`.

tauri/e2e/test/specs/first_run.test.js
```diff
@@ -0,0 +1,164 @@
...
+          if (await (await $("*=Connected to server")).isExisting()) return true;
+          return await browser.execute(() => {
+            try {
+              const raw = localStorage.getItem("gptme_servers");
+              if (!raw) return false;
+              const registry = JSON.parse(raw);
+              const active = registry.servers?.find(
+                (s) => s.id === registry.activeServerId
+              );
+              return /^http:\/\/127\.0\.0\.1:\d+/.test(active?.baseUrl || "");
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 958dee515. You're right that the stored URL can't prove a connection: `ApiContext.connect()` calls `updateServer()` (which persists the active `baseUrl`) *before* `checkConnection()` runs, so a failed connect still leaves a loopback URL in `gptme_servers`.

The wait condition no longer reads the registry at all. It now accepts only isConnected-derived signals:

- the "Connected to server" indicator or the "Continue" button label (both render only while `isConnected` is true), or
- the wizard advancing past the Local step to the provider ("Bring your own API key") or complete ("You're all set!") step, which `checkProviderAndAdvance` reaches only after a successful connect.

A failed connect now leaves the wizard on the Local step showing its error, so the wait times out and the test fails. Step 9's loopback/`tauri://` assertion is kept as the #3882 URL-correctness guard, with a comment noting it is not connection evidence.

TimeToBuildBob:
**E2E job fix (5f514986f)**: the failure was `WebDriverError: The operation is insecure.` on the first unwrapped `browser.execute`, and its root cause is that WebDriver's script context is isolated from the page.

Evidence from run [35439197147](https://github.com/gptme/gptme/actions/runs/35439197147/job/105887113363):
- `[e2e] get_server_status reported no port; falling back to 15700` — the `browser.execute(() => window.__TAURI__.core.invoke(...))` probe returned nothing, because page globals are not visible in that context (`withGlobalTauri` is set, so the page itself does have it).
- The same execute API then threw `SecurityError: The operation is insecure.` on `localStorage.getItem("gptme-settings")`.

Both are the same cause: the automation script world has no page globals and no reachable origin storage. So the fix removes all page-state access from the test:

1. `resolveSidecarPort()` now reads `GPTME_SERVER_PORT` directly — CI sets it and the Rust `server_port()` reads the same variable, so the app and the test cannot disagree. The IPC probe only *looked* like it was asking the app.
2. Dropped the `localStorage` first-run forcing + reload. `SetupWizard` opens itself when `hasCompletedSetup` is false, which is the default on a clean CI profile (and when storage throws), so the "Get started" button is already on screen.
3. Dropped the final `localStorage.gptme_servers` read for the #3882 guard. That regression is guarded behaviourally: with `tauri://localhost` retargeted, the connect fails, `isConnected` stays false, and the connected-signal wait times out.

The sidecar readiness path itself is fine — `waitForSidecarReady` reached HTTP 200 on `127.0.0.1:15700` before the failure, so the port and the sidecar build are working. Remaining risk is UI-level (does the wizard click-through reach a connected state under xvfb); this push is what tests that.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
**E2E result for 5f514986f** — the script-context failure is fixed; the job now fails one step later, on the wizard entry point.

What the run shows ([35439976438](https://github.com/gptme/gptme/actions/runs/35439976438/job/105889205979)): `readyState` returns `complete`, the sidecar is reached, and then `Expect $(button=Get started) to exist` fails — `findElements` returns `[]`. No `SecurityError`, so the `browser.execute` storage/IPC removal worked.

Why there is no button: `WelcomeView` hides the wizard CTA **by design** for desktop Tauri with a managed local server (`showWizardCta = isDefaultLocalServer && !demoMode && (!isTauri || isRemoteOnlyTauri)`, #3407). So in this build the only entry points to `SetupWizard` are (a) its own auto-open — `isOpen = !demoMode && !settings.hasCompletedSetup && !isHostedPageOrigin()` — and (b) the `setupWizard$` observable, which is unreachable from the isolated WebDriver script context.

In this run (a) did not happen, and I could not determine why from the log alone. `isHostedPageOrigin()` only matches `chat.gptme.org`, so on `tauri://localhost` it should fire on a clean profile — which makes the missing wizard interesting on its own: if a fresh install shows neither the wizard nor a CTA, first-run setup is unreachable, which is the #3882 family this PR exists to guard.

I stopped here rather than guess again. The cheap next step is a diagnostic, not a selector change: dump `browser.getPageSource()` when the CTA lookup fails. `getPageSource` is a WebDriver command (not `browser.execute`), so it works in this context and will show exactly what the app renders — welcome view, wizard, spinner, or error.

Suggest deciding the intended entry point before patching the test: either the wizard should auto-open on a clean desktop install (then this is a product bug and the test is right to fail), or the E2E should drive the wizard through an app-level hook/IPC that the automation context can actually reach.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 1 P1 disposed (fixed in 13c7c9348e)

| --- | --- | --- |
| `tauri/e2e/test/specs/first_run.test.js:93` | P1 | fixed in `13c7c9348e` |
| `gptme/cli/doctor.py:1162` | P2 | rejected — The review verified an older stacked branch state containing gptme/cli/doctor.py; the current PR diff does not contain t |
| `gptme/cli/doctor.py:1173` | P2 | rejected — The review verified an older stacked branch state containing gptme/cli/doctor.py; the current PR diff does not contain t |
| `tauri/e2e/test/specs/first_run.test.js:108` | P2 | fixed in `13c7c9348e` |
| `tauri/e2e/wdio.conf.js:84` | P2 | fixed in `13c7c9348e` |
| `webui/src/components/SetupWizard.tsx:310` | P2 | rejected |

</details>




This PR replaces the mock-based Tauri E2E test with a real integration test that builds the webui dist and PyInstaller sidecar, then drives the SetupWizard Local→Connect flow. It adds a `_check_plugins` diagnostic to gptme-doctor, switches build-sidecar.sh to use a PyInstaller spec, deletes chat_flow.test.js, and adds a `beforeSession` cleanup hook in wdio.conf.js. It also changes SetupWizard.tsx to skip auto-advance on the welcome step.

Confidence 3/5

> ⚠️ **1 of 8 changed files were not fully read.** Claims that an identifier is never assigned, used, or called are discarded unless verification names only fully preloaded files. This coverage describes the bounded context preload; the diff was reviewed.

<details><summary><b>6 findings</b> · ❌ 1 P1 · ⚠️ 5 P2</summary>

**❌ P1 high** — `tauri/e2e/test/specs/first_run.test.js:93`

The `first_run.test.js` uses `$("button=Get started")` and `$("//button[contains(., 'Local')]")` selectors. The `button=Get started` selector is a WebdriverIO text-based selector that matches a button whose text is exactly 'Get started'. However, the SetupWizard's 'Get started' button contains an ArrowRight icon and the text 'Get started' with whitespace/newlines from JSX, so the exact text match may fail. The PR description itself notes that 'button=Get started' does not exist in the current head, suggesting the selector is wrong. The `//button[contains(., 'Local')]` XPath is also fragile: it matches any button containing 'Local' anywhere, which could match 'Local setup' or other buttons, and the test clicks the first match, which may not be the intended 'Local' option in the mode step.

```suggestion
Consider capturing the sidecar process state or app logs on timeout to make failures actionable.
```

<sub>**How this was verified:** Checked the test flow: the sidecar is expected to be started by the Tauri app's setup hook, but the test waits for it after clicking 'Local' and before clicking 'Connect'. The comment says 'It is started by the Tauri app's setup hook, not by this test.'</sub>

**⚠️ P2 medium** — `tauri/e2e/wdio.conf.js:84`

The `beforeSession` cleanup kills whatever process owns the sidecar port using `fuser -k -TERM` then `-KILL`. On a shared CI runner, another user's process could be bound to port 15700 (or 5700 locally), and this would kill it. More importantly, the cleanup runs before every session, but the sidecar is started by the Tauri app's setup hook during the session. If the previous session's sidecar is still starting up (or the port is reused by a different process), the kill may race with the new session's sidecar startup, causing the new sidecar to fail to bind or the test to connect to a stale process. The `fuser` command also requires the `psmisc` package, which may not be installed on minimal runners; the `|| true` swallows that, but then the stale sidecar is not killed and the test may see a pre-existing connected state.

<sub>**How this was verified:** Checked the beforeSession hook in tauri/e2e/wdio.conf.js lines 78-110. The fuser command targets the port, not a specific PID, and the waits are 1000ms and 500ms.</sub>

**⚠️ P2 medium** — `tauri/e2e/test/specs/first_run.test.js:108`

The `first_run.test.js` waits for the sidecar readiness endpoint `/api/v2/server/health` from Node, but the sidecar is started by the Tauri app's setup hook. If the sidecar fails to start (e.g., missing hook modules, the #3883/#3884 regression this test is meant to catch), the test will time out after 60s with a generic error. However, the test does not verify that the sidecar is actually the one started by the current app session — a stale sidecar from a previous run (if the `beforeSession` kill failed) would make the probe succeed immediately, and the test would pass even if the current app's sidecar is broken. The `beforeSession` kill is best-effort and may not work if `fuser` is missing or the port is owned by a different process.

<sub>**How this was verified:** Checked first_run.test.js lines 33-55 and 108; the probe only checks port reachability, not process identity.</sub>

**⚠️ P2 medium** — `gptme/cli/doctor.py:1162`

The `_check_plugins` function imports each tool module with `importlib.import_module(module_name)` and then calls `_discover_tools(ok_modules)`. However, `_discover_tools` may itself import the module again (or rely on the module being imported), and the `importlib.import_module` call has a side effect: it executes the module's top-level code. If a plugin's tool module has side effects (e.g., registering tools globally, starting threads, or requiring configuration), running `gptme-doctor` will trigger those side effects, which could be undesirable. More critically, the function catches `Exception` from `importlib.import_module`, but `_discover_tools` is called outside that try/except. If `_discover_tools` raises (e.g., because a module imports successfully but then fails during tool discovery), the exception propagates out of `_check_plugins` and crashes the entire `run_diagnostics` call, violating the function's stated goal of isolating plugin failures.

```suggestion
Add a type check after extending tools: `tools = [t for t in tools if isinstance(t, ToolSpec)]` or wrap the loop in try/except.
```

<sub>**How this was verified:** Checked the code path: after extending tools, the loop calls `tool.init()` without checking that `tool` is a ToolSpec. If `_discover_tools` returns a non-ToolSpec, `tool.init` may not exist.</sub>

**⚠️ P2 medium** — `gptme/cli/doctor.py:1173`

The `_check_plugins` function calls `tool.init()` for every tool that has an `init` attribute, but it does not check whether `init` is a callable. If a plugin defines `init` as a non-callable (e.g., a string or `None`), the `if not tool.init: continue` check passes for truthy non-callables, and then `tool.init()` raises `TypeError: 'str' object is not callable`. This exception is caught by the `except Exception` block and reported as an ERROR, which is acceptable, but the error message would be misleading ('init() raised' when it actually wasn't callable). More importantly, the function does not validate that the `ToolSpec` returned by `init()` is the same tool or that it has the expected fields; it only checks `isinstance(initialized, ToolSpec)`. A plugin could return a different `ToolSpec` and the check would pass, masking a contract violation.

<sub>**How this was verified:** Checked doctor.py lines 1169-1194; no callable check on tool.init before invocation.</sub>

**⚠️ P2 medium** — `webui/src/components/SetupWizard.tsx:310`

The SetupWizard auto-advance effect now skips the 'welcome' step, but the test first_run.test.js clicks 'Get started' and then 'Local'. The test assumes the wizard auto-opens on a clean profile. However, the PR description notes that in a previous run, the wizard did not auto-open and the 'Get started' button did not exist. The change to the effect (adding 'welcome' to the guard) prevents auto-advance from the welcome step, but it does not ensure the wizard opens. If the wizard does not auto-open, the test will fail at the 'Get started' button wait. The test's comment says the wizard auto-opens when hasCompletedSetup is false, but the PR description indicates this is not reliable. This is a test reliability issue that could cause flaky CI.

<sub>**How this was verified:** The effect now returns early for 'welcome', preventing auto-advance from that step. The test clicks 'Get started' to move to 'mode', so the guard is intentional.</sub>

</details>

- `.github/workflows/tauri.yml` — E2E job now builds real webui dist and sidecar via make tauri-build-sidecar, sets GPTME_SERVER_PORT, and runs smoke + first_run specs.
- `gptme/cli/doctor.py` — Adds _check_plugins() to validate plugin tool init() contracts and wires it into run_diagnostics().
- `tauri/e2e/SPEC-real-first-run.md` — New spec document describing the real first-run E2E approach, scope, and CI integration plan.
- `tauri/e2e/test/specs/chat_flow.test.js` — Deleted the mock-based chat interaction test file.
- `tauri/e2e/test/specs/first_run.test.js` — New test that waits for the sidecar, clicks through the SetupWizard Local flow, and asserts a connected signal.
- `tauri/e2e/wdio.conf.js` — Adds beforeSession cleanup that kills the sidecar port owner and clears Tauri profile directories.
- `tauri/scripts/build-sidecar.sh` — Switches PyInstaller invocation to use gptme-server.spec and moves output from dist/ to tauri/bins.
- `tests/test_doctor.py` — Adds TestCheckPlugins covering no-plugins, bad init, raising init, broken tool module, and mixed module cases.
- `webui/src/components/SetupWizard.tsx` — Adds 'welcome' to the steps that skip auto-advance in the isConnected effect.

</details>

<sub>Reviewed `080000874495` · `openrouter/deepseek/deepseek-v4-flash-0731` · `llm` engine · 74s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

tauri/e2e/test/specs/first_run.test.js
```diff
@@ -0,0 +1,153 @@
...
+    //    `hasCompletedSetup` is false (the default on a clean profile).
+    //    We do NOT probe the sidecar here: it is only needed at step 6 (Connect),
+    //    and probing before the UI renders means a lingering sidecar from a prior
+    //    Tauri session would cause the probe to return immediately — before React
+    //    has mounted the wizard. Give React time with an explicit 30s wait.
+    //
+    //    Forcing `hasCompletedSetup` via `browser.execute(() => localStorage...)`
+    //    is not possible: WebDriver's script context cannot access page storage on
+    //    tauri:// origins (SecurityError: "The operation is insecure.").
+    const getStartedBtn = await $("button=Get started");
```
TimeToBuildBob: ❌ **P1** — The `first_run.test.js` uses `$("button=Get started")` and `$("//button[contains(., 'Local')]")` selectors. The `button=Get started` selector is a WebdriverIO text-based selector that matches a button whose text is exactly 'Get started'. However, the SetupWizard's 'Get started' button contains an ArrowRight icon and the text 'Get started' with whitespace/newlines from JSX, so the exact text match may fail. The PR description itself notes that 'button=Get started' does not exist in the current head, suggesting the selector is wrong. The `//button[contains(., 'Local')]` XPath is also fragile: it matches any button containing 'Local' anywhere, which could match 'Local setup' or other buttons, and the test clicks the first match, which may not be the intended 'Local' option in the mode step.

tauri/e2e/wdio.conf.js
```diff
@@ -62,6 +64,51 @@ exports.config = {
...
+  //
+  // 2. Clear the Tauri WebKit user-data directory so each test starts with a
+  //    clean localStorage / hasCompletedSetup=false.
+  beforeSession: async () => {
+    const { execSync } = require("child_process");
+    const sidecarPort = Number(process.env.GPTME_SERVER_PORT || "5700");
+
+    // Kill whatever process owns the sidecar port (graceful then forceful).
+    try {
+      execSync(`fuser -k -TERM ${sidecarPort}/tcp 2>/dev/null || true`, {
```
TimeToBuildBob: ⚠️ **P2** — The `beforeSession` cleanup kills whatever process owns the sidecar port using `fuser -k -TERM` then `-KILL`. On a shared CI runner, another user's process could be bound to port 15700 (or 5700 locally), and this would kill it. More importantly, the cleanup runs before every session, but the sidecar is started by the Tauri app's setup hook during the session. If the previous session's sidecar is still starting up (or the port is reused by a different process), the kill may race with the new session's sidecar startup, causing the new sidecar to fail to bind or the test to connect to a stale process. The `fuser` command also requires the `psmisc` package, which may not be installed on minimal runners; the `|| true` swallows that, but then the stale sidecar is not killed and the test may see a pre-existing connected state.

tauri/e2e/test/specs/first_run.test.js
```diff
@@ -0,0 +1,153 @@
...
+
+    // 5. In "Choose your setup", click "Local"
+    const localBtn = await $("//button[contains(., 'Local')]");
+    await expect(localBtn).toExist();
+    await localBtn.click();
+
+    // 6. Wait for the sidecar to be ready before clicking Connect. Now that we
+    //    are past the wizard's first step the UI is fully mounted, so the probe
+    //    is no longer racing against React rendering.
+    await waitForSidecarReady(sidecarPort);
```
TimeToBuildBob, line 137: ⚠️ **P2** — The `first_run.test.js` waits for the sidecar readiness endpoint `/api/v2/server/health` from Node, but the sidecar is started by the Tauri app's setup hook. If the sidecar fails to start (e.g., missing hook modules, the #3883/#3884 regression this test is meant to catch), the test will time out after 60s with a generic error. However, the test does not verify that the sidecar is actually the one started by the current app session — a stale sidecar from a previous run (if the `beforeSession` kill failed) would make the probe succeed immediately, and the test would pass even if the current app's sidecar is broken. The `beforeSession` kill is best-effort and may not work if `fuser` is missing or the port is owned by a different process.

gptme/cli/doctor.py
```diff
@@ -1076,6 +1076,133 @@ def _check_mcp_stdio_server(
...
+                            message=(
+                                f"Tool module {module_name!r} failed to import: "
+                                f"{type(exc).__name__}: {exc}"
+                            ),
+                        )
+                    )
+            if ok_modules:
+                from ..tools import _discover_tools
+
+                tools.extend(_discover_tools(ok_modules))
```
TimeToBuildBob: ⚠️ **P2** — The `_check_plugins` function imports each tool module with `importlib.import_module(module_name)` and then calls `_discover_tools(ok_modules)`. However, `_discover_tools` may itself import the module again (or rely on the module being imported), and the `importlib.import_module` call has a side effect: it executes the module's top-level code. If a plugin's tool module has side effects (e.g., registering tools globally, starting threads, or requiring configuration), running `gptme-doctor` will trigger those side effects, which could be undesirable. More critically, the function catches `Exception` from `importlib.import_module`, but `_discover_tools` is called outside that try/except. If `_discover_tools` raises (e.g., because a module imports successfully but then fails during tool discovery), the exception propagates out of `_check_plugins` and crashes the entire `run_diagnostics` call, violating the function's stated goal of isolating plugin failures.

gptme/cli/doctor.py
```diff
@@ -1076,6 +1076,133 @@ def _check_mcp_stdio_server(
...
+        if not tools:
+            # Plugin provides no tools (hooks/commands/providers only) — nothing
+            # to validate here.
+            continue
+
+        for tool in tools:
+            if not tool.init:
+                continue
+            try:
+                initialized = tool.init()
```
TimeToBuildBob: ⚠️ **P2** — The `_check_plugins` function calls `tool.init()` for every tool that has an `init` attribute, but it does not check whether `init` is a callable. If a plugin defines `init` as a non-callable (e.g., a string or `None`), the `if not tool.init: continue` check passes for truthy non-callables, and then `tool.init()` raises `TypeError: 'str' object is not callable`. This exception is caught by the `except Exception` block and reported as an ERROR, which is acceptable, but the error message would be misleading ('init() raised' when it actually wasn't callable). More importantly, the function does not validate that the `ToolSpec` returned by `init()` is the same tool or that it has the expected fields; it only checks `isinstance(initialized, ToolSpec)`. A plugin could return a different `ToolSpec` and the check would pass, masking a contract violation.

webui/src/components/SetupWizard.tsx
```diff
@@ -307,7 +307,7 @@ export function SetupWizard() {
       lastAutoAdvanceBaseUrlRef.current = null;
       return;
     }
-    if (!isOpen || step === 'complete' || step === 'provider') return;
+    if (!isOpen || step === 'welcome' || step === 'complete' || step === 'provider') return;
```
TimeToBuildBob, line 314: ⚠️ **P2** — The SetupWizard auto-advance effect now skips the 'welcome' step, but the test first_run.test.js clicks 'Get started' and then 'Local'. The test assumes the wizard auto-opens on a clean profile. However, the PR description notes that in a previous run, the wizard did not auto-open and the 'Get started' button did not exist. The change to the effect (adding 'welcome' to the guard) prevents auto-advance from the welcome step, but it does not ensure the wizard opens. If the wizard does not auto-open, the test will fail at the 'Get started' button wait. The test's comment says the wizard auto-opens when hasCompletedSetup is false, but the PR description indicates this is not reliable. This is a test reliability issue that could cause flaky CI.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

.github/workflows/tauri.yml
```diff
@@ -178,75 +172,29 @@ jobs:
...
+        run: make tauri-build-sidecar
 
       - name: Build Tauri debug binary
-        run: cd tauri/src-tauri && cargo build
+        # `custom-protocol` makes the binary serve the embedded frontendDist
+        # (webui/dist) from tauri://localhost, exactly like a shipped build.
+        # A bare `cargo build` is a dev-mode binary: it loads `devUrl`
+        # (http://localhost:5701), nothing listens there, and the webview
+        # renders an empty error page that still satisfies the smoke spec.
+        run: cd tauri/src-tauri && cargo build --features tauri/custom-protocol
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `bb0ab44cf`: `desktop_cors_origin()` now allows both the Vite development origin and all embedded Tauri origins regardless of Rust debug/release profile. The custom-protocol debug binary can therefore reach its managed sidecar. Added a Rust regression test covering the complete origin set; it passes locally.

tauri/e2e/wdio.conf.js
```diff
@@ -62,6 +64,51 @@ exports.config = {
...
+        shell: true,
+        stdio: "ignore",
+      });
+      await new Promise((r) => setTimeout(r, 1000));
+      execSync(`fuser -k -KILL ${sidecarPort}/tcp 2>/dev/null || true`, {
+        shell: true,
+        stdio: "ignore",
+      });
+      await new Promise((r) => setTimeout(r, 500));
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `bb0ab44cf`: the E2E suite no longer runs `pkill` or `fuser` at all. `wdio.conf.js` allocates an OS-assigned port before launching the app, exports it through `GPTME_SERVER_PORT` to both Tauri and the test, and relies on the sidecar's existing `--watch-pid` ownership cleanup. It can no longer kill an unrelated listener on port 5700.

tauri/e2e/wdio.conf.js
```diff
@@ -62,6 +65,48 @@ exports.config = {
...
+            rejectPort(new Error("Could not allocate an E2E sidecar port"));
+            return;
+          }
+          server.close((error) => {
+            if (error) rejectPort(error);
+            else resolvePort(address.port);
+          });
+        });
+      });
+      process.env.GPTME_SERVER_PORT = String(sidecarPort);
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `13c7c9348e`: `GPTME_SERVER_PORT` is now reserved in WDIO's launcher-side `onPrepare` hook before `tauri-driver` is spawned, and the spawn receives an explicit copy of that environment. The app and worker therefore inherit the same port. The previous CI log demonstrated the mismatch (`39163`/`40635` in workers while the app used its pre-launch environment); the lifecycle test now asserts reservation precedes spawn and that the exact port is propagated.

tauri/e2e/test/specs/first_run.test.js
```diff
@@ -0,0 +1,153 @@
...
+    //
+    //    Forcing `hasCompletedSetup` via `browser.execute(() => localStorage...)`
+    //    is not possible: WebDriver's script context cannot access page storage on
+    //    tauri:// origins (SecurityError: "The operation is insecure.").
+    const getStartedBtn = await $("button=Get started");
+    await getStartedBtn.waitForExist({
+      timeout: 30000,
+      timeoutMsg: "SetupWizard 'Get started' button did not appear within 30s",
+    });
+    await getStartedBtn.click();
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed across `76b68306e` and `13c7c9348e`. `SetupWizard` now explicitly suppresses connection-driven auto-advance while `step === 'welcome'`, so a fast sidecar cannot remove the first-run CTA. The E2E also uses stable `setup-wizard-get-started` / `setup-wizard-local` hooks rather than text selectors. The focused SetupWizard suite passes (33 tests).

tauri/e2e/test/specs/first_run.test.js
TimeToBuildBob: Fixed in commit 13c7c9348e: both interactions now use explicit `data-testid` contracts (`setup-wizard-get-started` and `setup-wizard-local`) defined on the rendered controls. This removes the icon/whitespace ambiguity and the broad `contains(., 'Local')` match. Selector-contract checks and the 33-test SetupWizard suite pass.

tauri/e2e/wdio.conf.js
TimeToBuildBob: Fixed in commit 13c7c9348e, completing the safe-port change begun in `bb0ab44cf`: the configuration no longer invokes `pkill` or `fuser`, reserves an unoccupied ephemeral port in launcher-side `onPrepare`, and passes it explicitly to `tauri-driver`. No unrelated process is signaled, and the app/test share an isolated port.

tauri/e2e/test/specs/first_run.test.js
TimeToBuildBob: Fixed in commit 13c7c9348e, completing the isolation introduced in `bb0ab44cf`: the app and test now inherit one freshly reserved port before any Tauri process launches. A stale sidecar cannot occupy that new port, and the health probe can only observe the current app session's managed sidecar. The timeout already reports the exact URL and last socket/HTTP error.

gptme/cli/doctor.py
TimeToBuildBob: Rejected as stale/out of scope for this PR head. `gptme/cli/doctor.py` is not changed by the current PR diff: the branch was rebuilt from master specifically to drop the unrelated stacked doctor commits before the Tauri E2E work continued. There is therefore no plugin-discovery code here to fix or merge under gptme/gptme#3886.

gptme/cli/doctor.py
TimeToBuildBob: Rejected as stale/out of scope for this PR head. `gptme/cli/doctor.py` is not changed by the current PR diff: the branch was rebuilt from master specifically to drop the unrelated stacked doctor commits before the Tauri E2E work continued. There is therefore no plugin `init()` contract change here to fix or merge under gptme/gptme#3886.

webui/src/components/SetupWizard.tsx
TimeToBuildBob: Rejected as superseded by the current implementation and evidence. The branch now intentionally suppresses connection-driven auto-advance at `step === 'welcome'`; the wizard opens from the existing fresh-profile condition, and CI on the prior head demonstrably found and clicked “Get started” before failing later at the mismatched sidecar port. Commit `13c7c9348e` also replaces the text selectors with stable test hooks. The claimed missing-wizard failure no longer reproduces.

TimeToBuildBob:
@greptileai review

.github/workflows/tauri.yml
```diff
@@ -178,75 +172,29 @@ jobs:
...
-              document.getElementById('app').setAttribute('data-ready', 'true');
-            </script>
-          </body>
-          </html>
-          HTMLEOF
+          cd webui
+          npm ci
+          npm run build
+
+      - uses: astral-sh/setup-uv@v7
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `1cff1e44e`: the newly added E2E step now pins `astral-sh/setup-uv` to commit `37802adc94f370d6bfd71619e3f0bf239e1f3b78` (the commit behind signed tag v7 / release v7.6.0), so this workflow no longer executes a mutable action tag.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

webui/src/contexts/ApiContext.tsx
```diff
@@ -137,20 +150,54 @@ export function ApiProvider({
...
-          updateServer(activeServer.id, {
-            ...(config.baseUrl !== undefined && { baseUrl: config.baseUrl }),
-            ...(config.authToken !== undefined && { authToken: config.authToken }),
-            ...(config.useAuthToken !== undefined && { useAuthToken: config.useAuthToken }),
-          });
+      // Use the render snapshot that backs this provider. Reading Legend State
+      // imperatively here can lag one React commit behind after the Tauri sync
+      // effect, which made the first manual Connect probe reuse an unauthenticated
+      // client even though the UI already rendered the managed sidecar config.
+      const activeServer = activeServerRef.current;
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `bd935a7bc`: `connect()` now accepts an explicit target server ID for same-tick registry transitions. `switchServer()` passes the selected ID, so config persistence, client caching, and the connection probe all use the selected record rather than the previous render snapshot. The regression test keeps server 1 in the render snapshot, switches to server 2, and asserts server 2 is updated/probed while server 1 is untouched. Focused ApiContext tests pass (15/15), along with TypeScript, ESLint, and scoped pre-commit checks.

TimeToBuildBob:
## Convergence adjudication — maintainer decision remains

- **Fixed:** the final blocking finding, “Server Switch Uses Stale State,” in `bd935a7bc`. Same-tick server switches now pass an explicit target ID into `connect()`, so the selected server—not the previous render snapshot—is updated, cached, and probed. The inline thread has the fix evidence and is resolved.
- **Remaining:** no unresolved P1/P2 findings. No non-blocking finding needed dismissal in this adjudication.
- **Verified against:** Greptile classified head `8b9c02508af1572d50eeabf0a432a1f2d1d07fc1`; the fix is current head `bd935a7bc0699f5d3cddf666d3527dcf6e7575b4`.
- **Verification:** focused `ApiContext` suite 15/15; TypeScript typecheck, ESLint, scoped pre-commit, and diff checks passed locally. The previous head had 21/21 completed GitHub checks green; checks for the new head are starting.
- **Domain risk:** manual maintainer testing should switch between two saved servers and delete the active non-preset server, confirming the new active server connects without mutating the previous record. The Linux first-run E2E remains the main release-path check.
- **Convergence:** detector reports `stable_rounds: 0` (required: 2) because the latest reviewed round introduced this new P1. Per the convergence-cap adjudication rule, I did **not** request another Greptile review. Current detector verdict is converged after disposition: zero unresolved P1/P2 threads.

Stopping at maintainer judgment; no auto-merge.

TimeToBuildBob:
Head `bd935a7bc` is now fully green: 21/21 completed checks pass, including the real Linux `tauri-driver` first-run E2E; all 14 review threads remain resolved. The self-merge gate still denies this PR because the current head lacks a fresh Bob AI-review verdict and the diff touches workflow/WebUI-sensitive paths. Per the convergence-cap adjudication already posted, I am not re-triggering review or bypassing the gate. Waiting for maintainer review/merge.

Issue #3812, opened 2026-09-11
Compaction cleanup: one budget, one trigger, model-participating compaction (default-on)


ErikBjare:
## Summary

gptme's context compaction is mechanically careful (tool-pair integrity, lossless master log, view branches) but the policy layer is a mess: four overlapping mechanisms with different thresholds, a trigger keyed to the raw context window so it never fires on 1M-window models, per-call mid-history mutation that breaks prompt caching, a dead summarize branch, and no way for the model to record state before history is dropped. Meanwhile the fleet's actual failure mode is a hard "Prompt is too long" death and cold resume, because autocompact is opt-in and nobody enables it.

This issue proposes cleaning up to **one budget, one trigger, one pipeline**, with the model participating in compaction, before adding any smarter triggers.

## Evidence

**A 25h deepseek-v4.1-flash agent session (1M window) never compacted** despite the autocompact tool being loaded:

| | tokens |
|---|---|
| log at end (gptme's counter) | 472,808 |
| of which assistant `<think>` content | 286,782 (61%) |
| autocompact trigger (50% of window + massive tool result) | 524,288 |
| `reduce_log` trigger (90%) | 943,718 |

The model had visibly degraded (reasoning stutter, then "Let me write. OK. Now." loops inside `<think>`) from turn ~191 of 467. Old reasoning is replayed verbatim every request (DeepSeek requires it on tool-calling turns), so the tic self-reinforced. Nothing in the default pipeline touches old reasoning until the 50% trigger.

**Across 8,751 local conversations**, autocompact fired in ~20 sessions, all on 200k Anthropic models at 100k–165k tokens. Recorded savings were often negligible while still invalidating the cache (e.g. 112,555 → 111,963; 113,179 → 112,918). One session re-compacted 7 times in 140 messages (pre-#946). The LLM resume path worked the few times it ran.

**Bob's fleet** never enables autocompact; sessions die on `205176 > 200000`-style overflow and resume cold (journal entries 2026-01 through 2026-08). The harness has a 429-resume path but none for overflow.

**Code findings** (master `31a385390`):
- Four layers in `prepare_messages`: `proactive_summarize_log` (env-gated, off) → `reduce_log` (90%/75%, truncate-only) → `prune_ephemeral_messages` → `limit_log` (100%, hard drop). Plus the `autocompact` tool's post-turn hook (50%/90%, 3 phases → view branch → LLM resume fallback).
[...]

PR #3890, opened 2026-09-20
feat(session): add durable conversation checkpoints

TimeToBuildBob:
## Summary

- add a versioned conversation-checkpoint artifact under each session log directory
- expose save/list/resume through \`gptme checkpoint-session\` and \`gptme-util checkpoint-session\`
- add \`/session-checkpoint save|list\` for the active conversation
- capture an offline deterministic summary, recent tool-call metadata, working-tree changes, model, and context boundary
- keep the lossless conversation log untouched and render a \`<<RESUMED SESSION>>\` handoff prompt

## Scope

This is the manual durable-checkpoint MVP from idea #5529. It does **not** wire the 80% automatic trigger yet; that belongs with the one-budget/default-on compaction work in gptme/gptme#3812 after gptme/gptme#3816 lands.

It is deliberately distinct from:
- \`gptme-checkpoint\` / \`/checkpoint\`: Git workspace recovery
- \`/backtrack\`: in-conversation rewind
- transparent context compaction

## Verification

- \`17 passed\` in \`tests/test_checkpoint_session.py\`
- \`96 passed, 1 xfailed\` across checkpoint + resume + utility CLI tests
- ruff format/check clean
- targeted mypy clean
- pre-commit hooks clean
- pre-PR quality gate: 100/100
- in-session review: no P0/P1 findings after one fix iteration

Refs gptme/gptme#3812


TimeToBuildBob:
Addressed all five review findings in `af023e113`:

- ordinary `python`/`json`/`bash` fences are preserved; only registered tool block types are compacted
- no-overwrite saves now publish with an atomic hard-link create, so concurrent same-label saves have exactly one winner
- human-facing resume output strips terminal controls while stored checkpoint JSON stays lossless
- nested checkpoint records and numeric fields are validated strictly on load
- Git status is parsed via NUL-delimited porcelain records, including renames and unusual filenames

Added regressions for each case. Local verification: 48 checkpoint/resume tests passed, targeted mypy passed, and all touched-file pre-commit hooks passed. The prior CI failures were unrelated flaky shell/server tests; the new commit has started a fresh CI run.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the remaining P1 in `292378747`: checkpoint JSON now ASCII-escapes surrogateescape code points, so valid POSIX filenames containing non-UTF-8 bytes no longer make save fail. Added an end-to-end regression that creates such a filename, saves, reloads, and verifies the path round-trip. Local verification: 27 checkpoint tests passed; targeted ruff, formatting, mypy, and pre-commit passed. The manual MVP remains intentionally separate from automatic budget-triggered checkpointing, which stays under gptme/gptme#3812 after gptme/gptme#3816.

TimeToBuildBob:
@greptileai review

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,594 @@
...
+            dir=path.parent,
+            prefix=f".{label}.",
+            suffix=".tmp",
+            delete=False,
+        ) as output:
+            temporary = Path(output.name)
+            # Git paths may contain arbitrary POSIX bytes decoded with
+            # surrogateescape. Escaping non-ASCII keeps those surrogate code
+            # points representable in a strict UTF-8 JSON file.
+            json.dump(checkpoint.to_dict(), output, indent=2, ensure_ascii=True)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in dcf06b2a4 — human-readable resume output now backslash-escapes surrogate code points before click.echo reaches a strict UTF-8 stream. The regression creates an actual invalid-0xff POSIX filename, saves and reloads it losslessly, then verifies CLI resume exits 0 and renders the escaped path.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,591 @@
...
+            capture_output=True,
+            timeout=10,
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return None
+    return result.stdout if result.returncode == 0 else None
+
+
+def _decode_git_path(value: bytes) -> str:
+    return value.decode("utf-8", errors="surrogateescape")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 292378747 — checkpoint JSON uses ensure_ascii=True, so surrogateescape code points are persisted as ASCII JSON escapes. The end-to-end invalid-0xff filename regression verifies save and load round-trip; dcf06b2a4 extends it through CLI resume.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Confidence 3/5

<details><summary><b>4 findings</b> · ❌ 1 P1 · ⚠️ 3 P2</summary>

**❌ P1 high** — `gptme/checkpoint_session.py:381`

Standalone `gptme-util checkpoint-session save` does not initialize the tool registry, but markdown extraction delegates to `ToolUse.iter_from_content()`, whose markdown parser only recognizes language tags from currently loaded tools. Consequently ordinary fenced calls such as `shell` and `save` are omitted from `last_tool_calls` when this feature is used through its production CLI entry point. The tests mask this because the global test fixture calls `init()` before each test. Parse markdown against discovered tool specs or initialize the target session's tools before extraction.

**⚠️ P2 medium** — `gptme/checkpoint_session.py:345`

When the generated summary exceeds `max_chars`, truncating the assembled string from the right preserves the original task and discards the recent conversation tail. A long initial request can therefore remove all current progress and next-step information—the most important part of a handoff checkpoint. Reserve space for, or truncate from, the original-task section so the recent tail is retained.

**⚠️ P2 medium** — `gptme/checkpoint_session.py:374`

Only `description` is capped at 500 characters; the full shell payload is still stored in `ToolCallSnapshot.command`. Shell calls commonly contain large heredocs, so one of the last five calls can make a supposedly compact checkpoint as large as the original payload (and preserve data that `_strip_tool_payloads()` intentionally removed from the summary). Bound or omit the persisted `command` value as well.

**⚠️ P2 medium** — `gptme/checkpoint_session.py:271`

The reported context percentage is based on every raw persisted message rather than the context the model actually receives. For example, runtime model/tool changes append replacement startup prompts, while `prepare_messages()` removes obsolete prompt generations; this code counts all generations and can report a substantially inflated or even greater-than-100% `pct_used`. Compute the boundary from the active/reduced context, or label it as trajectory size instead of model-context consumption.

</details>

| `dcf06b2a4911` | 1/5 | 4 | agent | 2026-09-20 05:31 UTC |

</details>

<sub>Reviewed `89835e427017` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 599s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,594 @@
...
+                (
+                    match.start(),
+                    _snapshot_tool_call(match.group("tool"), kwargs=payload),
+                )
+            )
+    return [snapshot for _, snapshot in sorted(calls, key=lambda item: item[0])]
+
+
+def recent_tool_calls(
+    messages: list[Message], limit: int = 5
```
TimeToBuildBob, line 402: 🛑 **P0** — `git diff --numstat` is run without `--no-textconv` or `--no-ext-diff`. A repository can associate paths in `.gitattributes` with a `diff.<driver>.textconv` or external diff command from `.git/config`, causing Git to execute that command while checkpoint creation collects line counts. `git_inspect_cmd()` only clears `diff.external`, so named drivers remain an arbitrary-code-execution path when checkpointing an untrusted workspace. Disable both text conversion and external diff execution for this inspection.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,594 @@
...
+from .tools import get_available_tools, get_tools
+from .util.git_cmd import git_inspect_cmd
+from .util.tokens import len_tokens
+
+SCHEMA_VERSION = 1
+_CHECKPOINT_DIR = "checkpoints"
+_LABEL_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,79}$")
+_FILE_KEYS = ("path", "file", "filename")
+_WRITE_TOOLS = frozenset({"append", "morph", "patch", "save"})
+_TOOL_BLOCK_RE = re.compile(
```
TimeToBuildBob, line 36: ⚠️ **P2** — Tool-call detection uses bespoke regexes that do not cover formats already supported by `ToolUse`: XML tool calls and variable-length Markdown fences are ignored, while native calls must contain strict, single-line JSON. For example, an XML-formatted session records no recent tool calls and leaves the complete tool payload in the default summary; non-streaming Anthropic calls can also be missed because their dictionary rendering is not strict JSON. Reuse the canonical tool parsers or add equivalent handling and tests for every supported tool format.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,594 @@
...
+        workspace = ChatConfig.from_logdir(logdir).workspace
+    except (OSError, ValueError):
+        workspace = logdir
+
+    checkpoint = ConversationCheckpoint(
+        version=SCHEMA_VERSION,
+        label=label,
+        timestamp=datetime.now(timezone.utc).isoformat(),
+        model=model,
+        context_boundary=_context_boundary(messages, model, model_limit),
```
TimeToBuildBob, line 516: ⚠️ **P2** — The checkpoint file is fsynced before publication, but the directory is never fsynced after `os.link()` or `os.replace()`. Consequently the command can report a successfully saved “durable” checkpoint that disappears after a crash or power loss, especially when the `checkpoints` directory was newly created. Sync the parent directory after publishing the file (and the relevant ancestor after creating the directory), using the repository's durability helper.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,594 @@
...
+        or not isinstance(raw_pct_used, (int, float))
+        or not math.isfinite(raw_pct_used)
+        or raw_pct_used < 0
+    ):
+        raise TypeError("context_boundary.pct_used must be a finite number or null")
+    total_tokens = cast(int, raw_total_tokens)
+    model_limit = cast(int | None, raw_model_limit)
+    pct_used = cast(int | float | None, raw_pct_used)
+    return ContextBoundary(
+        total_tokens, model_limit, float(pct_used) if pct_used is not None else None
```
TimeToBuildBob, line 182: ⚠️ **P2** — The generic record validator permits `None` for every field, including required fields such as `ToolCallSnapshot.tool`, `ToolCallSnapshot.description`, `FileChange.path`, and `FileChange.action`. Such malformed checkpoint records are treated as valid and rendered as `None` rather than being rejected/skipped as the schema promises. Required string fields should be checked separately from the nullable `file`, `command`, and `lines` fields, with coverage for null values.

TimeToBuildBob:
@greptileai review

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,635 @@
...
+        description=description[:500],
+        file=file_path,
+        command=command,
+    )
+
+
+def _tool_calls_from_content(content: str) -> list[ToolCallSnapshot]:
+    calls: list[tuple[int, ToolCallSnapshot]] = []
+    for tool_format in cast(tuple[ToolFormat, ...], ("markdown", "xml", "tool")):
+        for tool_use in ToolUse.iter_from_content(
```
TimeToBuildBob, line 381: ❌ **P1** — Standalone `gptme-util checkpoint-session save` does not initialize the tool registry, but markdown extraction delegates to `ToolUse.iter_from_content()`, whose markdown parser only recognizes language tags from currently loaded tools. Consequently ordinary fenced calls such as `shell` and `save` are omitted from `last_tool_calls` when this feature is used through its production CLI entry point. The tests mask this because the global test fixture calls `init()` before each test. Parse markdown against discovered tool specs or initialize the target session's tools before extraction.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,635 @@
...
+        )
+    parts.append("Recent conversation:")
+    parts.extend(
+        f"{message.role}: {_strip_tool_payloads(message.content).strip()}"
+        for message in recent
+    )
+    summary = "\n\n".join(part for part in parts if part)
+    if len(summary) <= max_chars:
+        return summary
+    return summary[: max_chars - 24].rstrip() + "\n\n[summary truncated]"
```
TimeToBuildBob, line 345: ⚠️ **P2** — When the generated summary exceeds `max_chars`, truncating the assembled string from the right preserves the original task and discards the recent conversation tail. A long initial request can therefore remove all current progress and next-step information—the most important part of a handoff checkpoint. Reserve space for, or truncate from, the original-task section so the recent tail is retained.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,635 @@
...
+    command = value("command")
+    if command is None and tool in {"shell", "shell_compact"}:
+        command = content
+    target = file_path or command
+    description = tool if not target else f"{tool}: {target}"
+    return ToolCallSnapshot(
+        tool=tool,
+        description=description[:500],
+        file=file_path,
+        command=command,
```
TimeToBuildBob, line 374: ⚠️ **P2** — Only `description` is capped at 500 characters; the full shell payload is still stored in `ToolCallSnapshot.command`. Shell calls commonly contain large heredocs, so one of the last five calls can make a supposedly compact checkpoint as large as the original payload (and preserve data that `_strip_tool_payloads()` intentionally removed from the summary). Bound or omit the persisted `command` value as well.

gptme/checkpoint_session.py
```diff
@@ -0,0 +1,635 @@
...
+    token_model = model or "gpt-4"
+    total_tokens = len_tokens(messages, token_model)
+    if model_limit is None and model is not None:
+        try:
+            resolved_model = get_model(model)
+        except (AssertionError, ValueError):
+            resolved_model = None
+        if resolved_model is not None:
+            model_limit = resolved_model.context
+    pct_used = total_tokens / model_limit if model_limit else None
```
TimeToBuildBob, line 271: ⚠️ **P2** — The reported context percentage is based on every raw persisted message rather than the context the model actually receives. For example, runtime model/tool changes append replacement startup prompts, while `prepare_messages()` removes obsolete prompt generations; this code counts all generations and can report a substantially inflated or even greater-than-100% `pct_used`. Compute the boundary from the active/reduced context, or label it as trajectory size instead of model-context consumption.

ErikBjare:
I don't fully understand the value/need of this, feels like we have a bunch of stuff like this already which is already barely used.

TimeToBuildBob:
You're right. This duplicates three existing mechanisms—lossless log resume, `/compact resume`/`RESUME.md`, and `/backtrack`—while adding a fourth persistence format and three new command surfaces. I built the artifact before proving a real user gap or adoption path, so the distinction in the PR description is taxonomy, not demonstrated value.

Closing this rather than polishing the remaining review findings. If automatic context-pressure recovery still has a real gap, it should be solved inside the existing compaction/resume path in gptme/gptme#3812 and gptme/gptme#3816, not with another manual checkpoint mechanism.

PR #3894, opened 2026-09-20
build(deps): add uv-native [dependency-groups] with pytest

TimeToBuildBob:
## Summary

- Adds `[dependency-groups]` (PEP 735) with pytest and test plugins — the uv-native format that `uv sync` reads by default
- Fixes `uv run pytest` in fresh gptme worktrees, which was falling through to the global uv-tools pytest (a separate Python with no gptme deps)
- Keeps `[tool.poetry.group.dev.dependencies]` intact for Poetry users

## Problem

When creating a gptme worktree and running `uv run pytest`, uv creates a venv with the 75 main deps but **not pytest** (it's only in `[tool.poetry.group.dev.dependencies]`, which uv ignores). uv then falls back to the PATH's `pytest` — the global uv-tools pytest at `~/.local/share/uv/tools/pytest/bin/python` — which has none of gptme's dependencies. This causes:

```
ImportError while loading conftest 'tests/conftest.py'
    import requests
E   ModuleNotFoundError: No module named 'requests'
```

This is a recurring reliability alert — 3 sessions hit it in a 6-hour window today (sessions 5522, ebe8, 66d2).

## Fix

Adding `[dependency-groups]` makes `uv sync` (the default) install pytest into the worktree venv. After this, `uv run pytest` uses the venv's Python where `requests` and all gptme deps are present.

```bash
# After this change, in any fresh gptme worktree:
uv run pytest tests/test_doctor.py -q  # ✅ works
```

Co-Authored-By: Bob <bob@superuserlabs.org>


pyproject.toml
```diff
@@ -142,6 +142,86 @@ agent-client-protocol = {version = ">=0.11.0", optional=true, python=">=3.10,<3.
 # tui
 textual = {version = ">=8,<9", optional=true}
 
+# uv-native (PEP 735) — makes `uv run pytest` work in fresh worktrees; keep in sync with [tool.poetry.group.dev.dependencies]
+[dependency-groups]
+test = [
```
TimeToBuildBob: ❌ **P1** — The new `test`, `lint`, `build`, `docs`, and `types` groups are independent, non-optional Poetry groups; listing the same packages again in `dev` does not make those groups part of `dev`. Consequently, existing production-oriented commands such as `poetry export --without dev` in `scripts/Dockerfile` exclude only the aggregate `dev` group and still export pytest, mypy, PyInstaller, Sphinx, lxml, and the other tooling from these groups. This breaks the previous `--without dev` contract and can substantially bloat or break production dependency exports. Mark the auxiliary groups optional for Poetry or change production installs/exports to select only `main`.

pyproject.toml
```diff
@@ -142,6 +142,86 @@ agent-client-protocol = {version = ">=0.11.0", optional=true, python=">=3.10,<3.
...
+    "pytest>=9.1",
```
TimeToBuildBob: ⚠️ **P2** — The duplicated uv constraints are not actually kept in sync with the Poetry constraints. For example, `pytest>=9.1` here permits pytest 10 while Poetry's `pytest = "^9.1"` excludes it; the same missing major-version bounds occur for pytest plugins, PyInstaller, Sphinx, and other tools. Because uv does not consume `poetry.lock` and `uv.lock` is ignored, fresh `uv run` environments can resolve tool versions that the Poetry environment explicitly disallows. Mirror the Poetry upper bounds (for example `pytest>=9.1,<10`) or derive/include a single shared group rather than maintaining divergent specifications.

TimeToBuildBob:
## 🤖 AI code review

### Needs a look — P2 only

Confidence 4/5

> ⚠️ **1 generated file excluded and NOT reviewed:** ``poetry.lock`` (+516/-252 lines, 103KB). Machine-generated output is removed from the diff before review — from what the model sees and from the size gate alike — so nothing below says anything about it.

**⚠️ P2 medium** — `.github/workflows/test.yml:289`

The smoke test invokes `pytest` as a bare executable, so `uv run` can still fall through to a runner-provided executable on the inherited PATH—the exact failure mode this change is intended to prevent. If that global environment has enough dependencies, the job can pass even when the project environment does not contain pytest. Invoke it through the synchronized environment's interpreter (for example, `uv run python -m pytest`) or explicitly assert that pytest is imported from `.venv`.

</details>

| `48742a0ad7bf` | 3/5 | 2 | agent | 2026-09-20 10:22 UTC |
| `d68302a8e8e0` | 5/5 | 0 | agent | 2026-09-20 22:23 UTC |
| `d68302a8e8e0` | 4/5 | 1 | agent | 2026-09-20 23:32 UTC |

</details>

<sub>Reviewed `e841e851d140` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 460s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

pyproject.toml
```diff
@@ -142,6 +142,20 @@ agent-client-protocol = {version = ">=0.11.0", optional=true, python=">=3.10,<3.
...
+    "pytest-asyncio",
+    "pytest-mock",
+    "pytest-timeout>=2.4.0",
+    "pytest-retry>=1.7.0",
+    "pytest-dotenv>=0.5.2",
+    "pytest-cov",
+    "pytest-xdist>=3.8.0",
+    "greenlet",  # needed for playwright coverage
+]
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `d68302a8e`: restored one complete `dev` group containing the full Poetry development toolchain, then regenerated `poetry.lock`. Fresh uv environments now get pytest without displacing lint, typecheck, docs, OpenAPI, or build tools.

pyproject.toml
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in `d68302a8e`: mirrored Poetry’s caret upper bounds across the uv `dev` group, including pytest/plugins, PyInstaller, Sphinx tooling, OpenAPI validation, and tomli, then regenerated `poetry.lock`.

pyproject.toml
TimeToBuildBob: Fixed in `d68302a8e`: removed the independent auxiliary groups and restored one aggregate `dev` group, preserving the existing `poetry export --without dev` production contract.

pyproject.toml
TimeToBuildBob: Fixed in `d68302a8e`: uv constraints now carry the same upper bounds as their Poetry counterparts; the regenerated lockfile and current-head CI are green.

pyproject.toml
```diff
@@ -142,6 +142,43 @@ agent-client-protocol = {version = ">=0.11.0", optional=true, python=">=3.10,<3.
 # tui
 textual = {version = ">=8,<9", optional=true}
 
+# uv-native (PEP 735) — makes `uv run pytest` work in fresh worktrees; keep in sync with [tool.poetry.group.dev.dependencies]
+[dependency-groups]
```
TimeToBuildBob, line 146: ⚠️ **P2** — The new uv-specific dependency declaration is not exercised anywhere in CI: the test, lint, docs, and build workflows all install and run through Poetry. Since `uv.lock` is also ignored, Poetry's lock-sync check cannot validate this path. A malformed or drifting `[dependency-groups]` table could therefore pass every check while `uv run pytest`—the command this change is intended to fix—remains broken. Add a fresh-environment uv smoke test that runs at least one pytest test.

TimeToBuildBob: Fixed in `e841e851d`: added a `Test with uv (PEP 735 smoke)` job that `uv sync`s a fresh environment and runs `uv run pytest` on `tests/test_cli.py::test_help` and `test_version`. That loads `tests/conftest.py` (imports `requests`) through the venv pytest, so a missing or drifting `[dependency-groups]` table fails CI instead of falling through to the global uv-tools pytest. Verified locally: 2 passed.

TimeToBuildBob:
@greptileai review

.github/workflows/test.yml
```diff
@@ -256,6 +256,40 @@ jobs:
...
+    env:
+      RELEASE: false
+      OPENAI_API_KEY: ''
+      ANTHROPIC_API_KEY: ''
+      DEEPSEEK_API_KEY: ''
+      OPENROUTER_API_KEY: ''
+    steps:
+    - uses: actions/checkout@v7
+
+    - uses: astral-sh/setup-uv@v7
```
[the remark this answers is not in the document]
TimeToBuildBob: Declining the SHA pin. This job uses `astral-sh/setup-uv@v7` to match the existing `security-advisories.yml` job and the rest of this repo’s version-tag convention (`actions/checkout@v7`, `actions/setup-python@v7`). Pinning only the new smoke job would be a one-off while the identical action stays tag-pinned next door. Non-blocking; `Test with uv (PEP 735 smoke)` already passed on this head.

TimeToBuildBob:
Automated merge handoff for `e841e851d140`:

- CI gate: satisfied
- Review threads: clear
- `.github/workflows/test.yml`
- `pyproject.toml`

TimeToBuildBob:
## 📋 Reviewer's Brief

**What changed:** Adds a PEP 735 `[dependency-groups]` `dev` group (pytest + the full Poetry toolchain, caret-aligned) so `uv sync` / `uv run pytest` in fresh gptme worktrees stop falling through to the global uv-tools pytest, plus a `Test with uv (PEP 735 smoke)` CI job that actually exercises that path.
**Risk surface:** `poetry.lock` regeneration and a new job in `.github/workflows/test.yml`; the production `poetry export --without dev` contract is unchanged because there is still one aggregate `dev` group.
**Recommended decision:** Merge — CI green including the uv smoke job, Greptile 5/5, self-merge blocked on path policy (`.github/workflows/test.yml`, `pyproject.toml`).

<sub>Auto-generated by Bob · HEAD `e841e85` · [Report issue](https://github.com/ErikBjare/bob/issues)</sub>

.github/workflows/test.yml
```diff
@@ -256,6 +256,40 @@ jobs:
...
+
+    - name: Install project + PEP 735 dev group
+      run: uv sync
+
+    - name: Smoke uv run pytest
+      env:
+        TERM: xterm
+        MODEL: local/test
+      run: |
+        uv run pytest -q \
```
TimeToBuildBob: ⚠️ **P2** — The smoke test invokes `pytest` as a bare executable, so `uv run` can still fall through to a runner-provided executable on the inherited PATH—the exact failure mode this change is intended to prevent. If that global environment has enough dependencies, the job can pass even when the project environment does not contain pytest. Invoke it through the synchronized environment's interpreter (for example, `uv run python -m pytest`) or explicitly assert that pytest is imported from `.venv`.

TimeToBuildBob: Fixed in `8ad460379`: the PEP 735 smoke job now runs `uv run python -m pytest` so pytest is loaded from the synchronized venv interpreter. A runner-PATH pytest can no longer make this job pass when the project env is missing pytest. Verified locally: 2 passed.

TimeToBuildBob:
@greptileai review

PR #3895, opened 2026-09-20
feat(compaction): Phase 1 — one budget, always-on trigger, no per-call mutation (#3812)

TimeToBuildBob:
## Summary

Implements Phase 1 of #3812: one budget, one trigger, on by default.

**Phase 0** (overflow recovery, `compaction.jsonl` events) was shipped in #3816.
**Phase 2** (model-participating checkpoint compaction) is future work.

### Changes

**`context_budget` config layer** (`gptme/context/config.py`, `gptme/util/context_budget.py`):
- New `ContextConfig.budget` field: fraction (0<x≤1) or absolute token count (>1); default None = dynamic computation
- `get_context_budget(model_context)` resolves: `GPTME_CONTEXT_BUDGET` env → `[context] budget` config → `min(0.9×window, window−output−headroom)`
- For 1M-window models: budget=900k instead of the 50% trigger (524k) that never fired in the Flash session

**Remove per-call mutation** (`gptme/logmanager/manager.py`):
- Removed `reduce_log` and `proactive_summarize_log` from `prepare_messages` — these ran on every call and mutated mid-history, the exact pattern that caused cache regressions #2724 and #3620
- `limit_log` stays as a loud last-resort guard (logs a warning when it fires — should never happen in normal operation)

**Always-on compaction hook** (`gptme/init.py`):
- Registers the autocompact TURN_POST hook unconditionally in `init()`, not only when `--tool autocompact` is passed
- No user-visible change when the tool is also explicitly loaded (registry replaces same-name hooks)
- Bob's fleet and sessions that never passed `--tool autocompact` now get automatic compaction

**Unified thresholds** (`gptme/tools/autocompact/decision.py`):
- Both `estimate_compaction_savings` and `should_auto_compact` now use `get_context_budget()` instead of the hardcoded `0.9 * model.context` scatter

**Docs** (`docs/context-compression.rst`):
- Updated to describe one budget, one trigger, default-on pipeline

## Test Plan

- [x] 8 new tests in `tests/test_context_budget.py` (default, fraction, absolute, env override, invalid-ignored, zero-ignored, large-window, minimum-clamp)
- [x] `tests/test_reduce.py` (39 tests) — all pass; `limit_log`/`_drop_orphaned_tool_pairs` behavior unchanged
- [x] `tests/test_prompts.py::test_unchanged_history_produces_identical_provider_request` — passes (no per-call mutation)
- [x] `tests/test_generation_pre_snapshot_boundary.py::test_unchanged_history_produces_identical_provider_request` — passes
- [x] `tests/test_autocompact_scoring.py` and `tests/test_context_providers.py` (99 tests) — all pass


TimeToBuildBob:
## Review guide (auto-posted)

### Key files
- `gptme/util/context_budget.py` (+115/-0, Δ115)
- `tests/test_context_budget.py` (+71/-0, Δ71)
- `gptme/logmanager/manager.py` (+17/-20, Δ37)
- `docs/context-compression.rst` (+24/-6, Δ30)
- `gptme/context/config.py` (+22/-0, Δ22)
- `gptme/init.py` (+21/-0, Δ21)
- `gptme/tools/autocompact/decision.py` (+3/-2, Δ5)

### Suggested verification
- `uv run pytest tests/test_context_budget.py -q`
- `make typecheck`

### Suggested review focus

_Generated by `scripts/github/pr-warmup-review-guide.py` for gptme/gptme#3895._

TimeToBuildBob:
## 🤖 AI code review


### Safe to merge — 5 findings disposed (fixed in e86d57651; other; rejected; fixed in 1be63c387)

| --- | --- | --- |
| `gptme/init.py:87` | P0 | other |
| `gptme/executor.py:64` | P1 | other |
| `gptme/init.py:87` | P1 | fixed in `e86d57651` |
| `gptme/logmanager/manager.py:1121` | P1 | fixed in `1be63c387` |
| `gptme/tools/autocompact/decision.py:129` | P1 | fixed in `e86d57651` |
| `gptme/tools/autocompact/decision.py:132` | P1 | fixed in `e86d57651` |
| `gptme/tools/autocompact/decision.py:129` | P1 | fixed in `e86d57651` |
| `gptme/tools/autocompact/hook.py:109` | P1 | rejected — The verification step is 'a stored log just below budget becomes a provider-rejected request after enrich_messages_with_ |
| `gptme/util/context_budget.py:86` | P1 | fixed in `e86d57651` |
| `gptme/context/config.py:75` | P2 | fixed in `e86d57651` |
| `gptme/init.py:141` | P2 | fixed in `e86d57651` |
| `gptme/tools/autocompact/engine.py:162` | P2 | fixed in `1be63c387` |

</details>




### Not safe to merge — 1 P0 open

Confidence 1/5

<b>6 findings</b> · 🛑 1 P0 · ❌ 4 P1 · ⚠️ 1 P2

<details><summary><b>5 inline</b></summary>

**🛑 P0 critical** — `gptme/init.py:87`

Making autocompaction unconditional also makes the LLM-powered `summarize` branch automatic. That branch parses model-generated “Context Files”, and `_load_context_files` accepts absolute and `~/` paths and reads them without workspace containment or user confirmation. Untrusted conversation or tool content can therefore induce the summarizer to name a sensitive local file such as `~/.ssh/id_rsa`; its contents are inserted into the compacted log and sent to the provider on a subsequent request. Restrict generated paths to files inside the workspace or require explicit confirmation before reading anything outside it.

**❌ P1 high** — `gptme/tools/autocompact/hook.py:109`

The budget decision counts only `manager.log.messages`, but the actual provider input is enlarged later by `enrich_messages_with_context`. A stored log just below the budget can therefore become larger than `window - max_output - headroom` without triggering compaction. `prepare_messages` only calls `limit_log`, which caps against the full model window rather than the new safe ceiling, so it may leave a request that the provider rejects. The trigger needs to account for the prepared/enriched input or reserve its dynamic context explicitly.

**❌ P1 high** — `gptme/executor.py:64`

Registering this hook in the server execution environment does not make server compaction durable. `autocompact_hook` creates and switches to an in-memory view, but subsequent API steps and tool workers construct a new `LogManager` whose `current_view` defaults to `None`; resumed CLI processes do the same. They consequently reload the uncompressed master transcript, discarding the active compacted context and potentially sending it on the very next request. Persist the active view selection and restore it when loading a conversation, or replace the server's active projection through a durable mechanism.

**❌ P1 high** — `gptme/logmanager/manager.py:1121`

Removing request-time reduction leaves a gap in the server tool loop. The server fires `TURN_POST` immediately after the assistant message, before executing and appending its tool results; a continuation then calls `prepare_messages` before another `TURN_POST` can run. A large tool result that crosses the configured safe budget is therefore sent directly to the next provider call. Because this fallback only limits against the full context window, requests between the safe ceiling and the full window can overflow once output tokens are reserved. Compaction must run after tool results are appended or immediately before the continuation request.

**⚠️ P2 medium** — `gptme/tools/autocompact/engine.py:162`

At exactly the configured budget this branch is entered, but the candidate loop immediately stops because it tests `current_tokens <= target_tokens`. A massive tool result at the exact boundary is therefore left unchanged even though the decision layer triggered compaction; the hook then creates an effectively identical view and reports success. The new test misses this because its token mock returns the boundary value only for the original list object, not for `compacted_log`. Stop only when strictly below the target (and test all equivalent lists).

</details>

### Comments outside the diff

**❌ P1 high** — [`gptme/tools/autocompact/hook.py:102`](https://github.com/gptme/gptme/blob/396023aa43dc0a670b0d8f6ec0db8e8c6b4d4458/gptme/tools/autocompact/hook.py#L102)

The 60-second reentrancy throttle is process-global, not scoped to a conversation. Once any always-on session performs compaction, every other concurrent server conversation skips its own budget check for the next minute, allowing unrelated over-budget logs to reach provider calls or the lossy fallback. Track the last attempt per `LogManager`/conversation instead of in the module-global `_last_autocompact_time`.

| `697036580917` | 2/5 | 7 | agent | 2026-09-20 11:01 UTC |

</details>

<sub>Reviewed `396023aa43dc` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 849s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/util/context_budget.py
```diff
@@ -0,0 +1,115 @@
...
+            logger.warning(
+                "GPTME_CONTEXT_BUDGET=%r is not a number; using default", env_val
+            )
+
+    # 2. Config-level budget
+    try:
+        from ..config import get_config  # fmt: skip
+
+        cfg = get_config()
+        budget_cfg = cfg.context.budget if hasattr(cfg, "context") else None
```
TimeToBuildBob: ❌ **P1** — Config-based budgets are never resolved. `Config` has only `user`, `project`, and `chat`; it has no `context` attribute, while project context is stored at `cfg.project.context`. Consequently this branch always yields `None`. User config does not define/parse a context section either and will treat `[context]` as an unknown top-level key, potentially stripping it. Thus the documented project/user `[context] budget` setting has no effect.

gptme/tools/autocompact/decision.py
```diff
@@ -125,7 +126,7 @@ def should_auto_compact(log: list[Message], limit: int | None = None) -> Compact
 
     model = get_default_model() or get_model("gpt-4")
     if limit is None:
-        limit = int(0.9 * model.context)
+        limit = get_context_budget(model.context)
```
TimeToBuildBob: ❌ **P1** — The computed budget is used only for deciding whether to compact and is not passed into `CompressionConfig.limit` by the automatic hook. The compressor therefore falls back to its separate `0.8 * model.context` threshold. For example, with a configured 300k budget on a 1M model, a 310k conversation dominated by a large tool result is classified as `rule_based`, but `auto_compact_log()` sees neither its 700k close threshold nor 800k limit and can return the log unchanged while the hook still creates a supposedly compacted view. Propagate the resolved budget into the compression operation.

gptme/tools/autocompact/decision.py
```diff
@@ -125,7 +126,7 @@ def should_auto_compact(log: list[Message], limit: int | None = None) -> Compact
...
 
     total_tokens = len_tokens(log, model.model)
     close_to_limit = total_tokens >= int(
```
TimeToBuildBob: ❌ **P1** — The supposedly unified trigger still uses a hard-coded 50% of the provider window for the massive-tool-result path rather than the configured budget. Worse, `estimate_compaction_savings()` uses an 80% window threshold before counting those tool-result savings. Between 50% and 80%, a massive tool result therefore triggers this function, is excluded from the savings estimate, and causes an unnecessary LLM-powered `summarize` action instead of the intended rule-based trim—now by default in every session.

gptme/init.py
```diff
@@ -84,6 +84,7 @@ def init(
             first_line,
         )
     init_tools(tool_allowlist)
+    _register_core_compaction_hook()
```
TimeToBuildBob, line 87: ❌ **P1** — Registering the core hook only in `init()` does not make it unconditional across runtime contexts. Hook registries are ContextVar-local, while server/evaluation/subagent execution contexts are initialized through `prepare_execution_environment()`, which calls `init_tools()` and `init_hooks()` directly and never calls this helper. In a server request whose tool allowlist excludes `autocompact`, that context therefore has no compaction hook, contrary to the stated allowlist-independent behavior.

gptme/init.py
```diff
@@ -124,6 +125,26 @@ def init(
...
+    The hook is registered here rather than relying on the autocompact *tool*
+    being explicitly loaded, so it fires regardless of the tool allowlist.
+    If the tool is also loaded, its own register_hooks() call is a no-op because
+    the hook is already registered under the same name.
+    """
+    from .hooks import HookType, register_hook  # fmt: skip
+    from .tools.autocompact.hook import autocompact_hook  # fmt: skip
+
+    register_hook(
+        "autocompact",
```
TimeToBuildBob: ⚠️ **P2** — This registration does not share the same name as the tool-owned hook as the comment claims. `ToolSpec.register_hooks()` prefixes hook names with the tool name, so the existing entry is `autocompact.autocompact`, whereas this adds `autocompact`. When the tool is loaded, both hooks run after every turn; the module-global cooldown usually prevents duplicate mutation, but token counting and decision work are duplicated on non-triggering turns, which is particularly expensive for the large histories this hook handles.

gptme/tools/autocompact/decision.py
TimeToBuildBob: ❌ **P1** — The dynamic default never receives the selected model's output allowance: both callers pass only `model.context`, so this function always reserves the hard-coded 8,192 tokens even when `ModelMeta.max_output` is much larger. This also contradicts the documented `window - max_output - headroom` calculation (and the stated approximately 167k result for a 200k window; the current defaults return 180k). Models with large output limits can therefore trigger compaction too late to leave the promised output headroom.

gptme/context/config.py
```diff
@@ -57,8 +67,20 @@ def from_dict(cls, config_dict: dict[str, Any]) -> "ContextConfig":
             else ContextSelectorConfig()  # Use default instead of None
         )
 
+        budget_raw = config_dict.get("budget")
+        budget: float | int | None = None
+        if budget_raw is not None:
+            try:
+                f = float(budget_raw)
+                budget = int(f) if f > 1 else f
```
TimeToBuildBob: ⚠️ **P2** — Parsing does not enforce the documented domain. Zero, negative values, NaN, and non-integral absolute values are accepted; for example `budget = 1.5` becomes integer `1` and is later silently ignored. TOML `inf` reaches `int(f)` and raises an uncaught `OverflowError`. Validate finiteness and require either `0 < value <= 1` or an integral value greater than 1 here, rather than deferring to silent fallback behavior.

gptme/util/context_budget.py
```diff
@@ -0,0 +1,115 @@
...
+                return budget
+            if 0 < parsed <= 1:
+                budget = int(parsed * model_context)
+                logger.debug(
+                    "Context budget from GPTME_CONTEXT_BUDGET (%.2f × %d): %d tokens",
+                    parsed,
+                    model_context,
+                    budget,
+                )
+                return budget
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e86d57651 — explicit fractions and absolute budgets are now validated for finiteness/integrality and clamped to `window - max_output - headroom`; regression tests cover `1.0`, oversized absolute values, and `inf`.

gptme/tools/autocompact/decision.py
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e86d57651 — the hook resolves the budget once, passes it to both `should_auto_compact()` and `CompressionConfig.limit`, and the engine now trims toward that same limit.

gptme/init.py
```diff
@@ -124,6 +125,26 @@ def init(
...
+        HookType.TURN_POST,
+        autocompact_hook,
+        priority=100,
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in e86d57651 — core registration now uses the tool-qualified `autocompact.autocompact` name, so tool loading replaces rather than duplicates the hook.

gptme/util/context_budget.py
TimeToBuildBob: Fixed in e86d57651 — `UserConfig` now parses and retains `[context]`, resolution uses project-over-user precedence, and tests verify user config is not stripped.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — the resolved budget is propagated through the hook into `CompressionConfig.limit` and used as the engine target.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — removed the separate 50%/80% provider-window triggers; decision and savings estimation now use the one resolved budget.

gptme/init.py
TimeToBuildBob: Fixed in e86d57651 — `prepare_execution_environment()` now registers the core compaction hook in each context, independently of its tool allowlist.

gptme/init.py
TimeToBuildBob: Fixed in e86d57651 — the core hook uses the same `autocompact.autocompact` registry name as `ToolSpec.register_hooks()`.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — decision and hook paths pass `ModelMeta.max_output` into budget resolution; docs now describe model-specific examples.

gptme/context/config.py
TimeToBuildBob: Fixed in e86d57651 — `ContextConfig.from_dict()` rejects non-finite, non-positive, and non-integral absolute values; parametrized regressions cover each case.

gptme/util/context_budget.py
TimeToBuildBob: Fixed in e86d57651 — `UserConfig` now parses and retains `[context]`, resolution uses project-over-user precedence, and tests verify user config is not stripped.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — the resolved budget is propagated through the hook into `CompressionConfig.limit` and used as the engine target.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — removed the separate 50%/80% provider-window triggers; decision and savings estimation now use the one resolved budget.

gptme/init.py
TimeToBuildBob: Fixed in e86d57651 — `prepare_execution_environment()` now registers the core compaction hook in each context, independently of its tool allowlist.

gptme/init.py
TimeToBuildBob: Fixed in e86d57651 — the core hook uses the same `autocompact.autocompact` registry name as `ToolSpec.register_hooks()`.

gptme/tools/autocompact/decision.py
TimeToBuildBob: Fixed in e86d57651 — decision and hook paths pass `ModelMeta.max_output` into budget resolution; docs now describe model-specific examples.

gptme/context/config.py
TimeToBuildBob: Fixed in e86d57651 — `ContextConfig.from_dict()` rejects non-finite, non-positive, and non-integral absolute values; parametrized regressions cover each case.

TimeToBuildBob:
@greptileai review

gptme/tools/autocompact/decision.py
```diff
@@ -117,32 +117,19 @@ def should_auto_compact(log: list[Message], limit: int | None = None) -> Compact
...
-            msg_tokens = len_tokens(msg.content, model.model)
-            if msg_tokens > 2000:  # Threshold for "massive"
-                has_massive_tool_result = True
-                break
-
-    # First check: would we trigger based on token count?
-    would_trigger = total_tokens > limit or (close_to_limit and has_massive_tool_result)
+
+    # One budget means one trigger: no separate provider-window fraction.
+    would_trigger = total_tokens >= limit
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4902f57f9 — the engine now trims when `current_tokens >= target_tokens`, matching the decision's inclusive trigger. Added an exact-budget regression test that proves a tool-heavy log is changed rather than creating an unchanged view.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Fixed the failing no-key CI test in `396023aa4`.

The failure was order-dependent: an earlier test could configure Rich logging on the shared stdout console, so `models info --json` let `get_model()`'s fallback warning prefix the JSON document. JSON mode now suppresses model-resolution logging while preserving the explicit unknown-provider warning on stderr. Added a regression that reproduces the prior stdout-logging state.

Local verification: 18 model CLI tests and 91 compaction/executor tests pass; the exact CI-order reproducer also passes. The in-band reviewer was unavailable due the OpenRouter credit limit, so the wrapper pushed after the required fail-open review attempt.

TimeToBuildBob:
@greptileai review

gptme/init.py
TimeToBuildBob, line 87: 🛑 **P0** — Making autocompaction unconditional also makes the LLM-powered `summarize` branch automatic. That branch parses model-generated “Context Files”, and `_load_context_files` accepts absolute and `~/` paths and reads them without workspace containment or user confirmation. Untrusted conversation or tool content can therefore induce the summarizer to name a sensitive local file such as `~/.ssh/id_rsa`; its contents are inserted into the compacted log and sent to the provider on a subsequent request. Restrict generated paths to files inside the workspace or require explicit confirmation before reading anything outside it.

gptme/tools/autocompact/hook.py
```diff
@@ -106,8 +107,14 @@ def autocompact_hook(
         return
 
     messages = manager.log.messages
```
TimeToBuildBob, line 106: ❌ **P1** — The budget decision counts only `manager.log.messages`, but the actual provider input is enlarged later by `enrich_messages_with_context`. A stored log just below the budget can therefore become larger than `window - max_output - headroom` without triggering compaction. `prepare_messages` only calls `limit_log`, which caps against the full model window rather than the new safe ceiling, so it may leave a request that the provider rejects. The trigger needs to account for the prepared/enriched input or reserve its dynamic context explicitly.

gptme/executor.py
```diff
@@ -63,6 +63,7 @@ def prepare_execution_environment(
 
     # Initialize tools and hooks
```
TimeToBuildBob, line 64: ❌ **P1** — Registering this hook in the server execution environment does not make server compaction durable. `autocompact_hook` creates and switches to an in-memory view, but subsequent API steps and tool workers construct a new `LogManager` whose `current_view` defaults to `None`; resumed CLI processes do the same. They consequently reload the uncompressed master transcript, discarding the active compacted context and potentially sending it on the very next request. Persist the active view selection and restore it when loading a conversation, or replace the server's active projection through a durable mechanism.

gptme/logmanager/manager.py
```diff
@@ -1102,35 +1100,34 @@ def prepare_messages(
...
-    if (len_from := len_tokens(msgs, model.model)) != (
-        len_to := len_tokens(msgs_pruned, model.model)
-    ):
-        logger.debug(f"Reduced log from {len_from // 1} to {len_to // 1} tokens")
+
+    # Last-resort hard limit against the provider window.
+    # This should never fire in normal operation — compaction (via the TURN_POST
+    # autocompact hook) keeps the log below the context budget before we get here.
+    # If it fires, something bypassed compaction; log a warning so it's visible.
     msgs_limited = limit_log(msgs_pruned)
```
TimeToBuildBob, line 1171: ❌ **P1** — Removing request-time reduction leaves a gap in the server tool loop. The server fires `TURN_POST` immediately after the assistant message, before executing and appending its tool results; a continuation then calls `prepare_messages` before another `TURN_POST` can run. A large tool result that crosses the configured safe budget is therefore sent directly to the next provider call. Because this fallback only limits against the full context window, requests between the safe ceiling and the full window can overflow once output tokens are reserved. Compaction must run after tool results are appended or immediately before the continuation request.

gptme/tools/autocompact/engine.py
```diff
@@ -158,9 +157,9 @@ def auto_compact_log(
     tool_result_tokens_saved = 0
     compression_tokens_saved = 0
     current_tokens = len_tokens(compacted_log, model.model)
-    target_tokens = int(0.8 * model.context)  # Target 80% of context
+    target_tokens = limit
 
-    if current_tokens > target_tokens:
+    if current_tokens >= target_tokens:
```
TimeToBuildBob, line 162: ⚠️ **P2** — At exactly the configured budget this branch is entered, but the candidate loop immediately stops because it tests `current_tokens <= target_tokens`. A massive tool result at the exact boundary is therefore left unchanged even though the decision layer triggered compaction; the hook then creates an effectively identical view and reports success. The new test misses this because its token mock returns the boundary value only for the original list object, not for `compacted_log`. Stop only when strictly below the target (and test all equivalent lists).

gptme/init.py
TimeToBuildBob: Fixed in 1be63c387 — `_load_context_files` now resolves each path and skips anything outside the workspace, including `~/`, absolute paths, `../` traversal, and symlinks that escape. Regression tests cover those cases.

gptme/executor.py
TimeToBuildBob: Fixed in 1be63c387 — `switch_view` persists `views/.current`, `LogManager` restores it on load, and `LogManager.load()` attaches append-cursor metadata to the branch rather than clobbering a restored view. A reload test covers the round-trip.

gptme/logmanager/manager.py
TimeToBuildBob: Fixed in 1be63c387 — the server tool worker now runs `autocompact_hook` after tool results are appended and before the continuation request. The reentrancy throttle is per-conversation (not process-global) and allows a retry when the log grew, so a large tool result is not skipped just because TURN_POST already ran on the assistant message.

gptme/tools/autocompact/engine.py
TimeToBuildBob: Fixed in 1be63c387 — the engine now stops only when `current_tokens < target_tokens`, so an exact-budget log is still trimmed. The regression mocks token counts for every equivalent list, not just the original `messages` object.

gptme/tools/autocompact/hook.py
TimeToBuildBob: Won't-fix as an accepted Phase 1 trade-off.

The compaction trigger counts the stored log — the thing the engine can actually reduce. `enrich_messages_with_context` rebuilds ephemeral workspace/git context on every request and is not part of the compacted view. Folding that into the trigger would compact (and cache-bust) based on content the engine cannot remove.

`limit_log` still guards the provider window as a last-resort cap. Accounting for dynamic context in the budget is a later slice, not a blocker for one-budget/always-on.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -1327,6 +1353,16 @@ def stream_tool_output(
                     with SessionManager.conversation_lock(conversation_id):
                         session._executing_tools.discard(claimed_tool_id)
 
+            # Compact after tool results, before the continuation provider call.
+            try:
+                manager = LogManager.load(conversation_id, branch=branch, lock=False)
+                _compact_after_tool_results(manager, session, conversation_id)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 8479975df — post-tool compaction now loads and runs under `SessionManager.conversation_lock`, the same lock as tool-result appends and continuation election. A regression test probes the lock from another thread during compaction.

gptme/logmanager/manager.py
```diff
@@ -317,11 +315,18 @@ def __init__(
...
-        # If a view was requested, load it as the active log
-        if self.current_view and self.current_view in self._views:
-            # When on a view, the "current" log is the view
-            # but we track master separately for dual-write
-            pass  # View is already loaded in _views
+        # Restore the last active view unless the caller selected one.
+        # Compaction switches are otherwise in-memory and lost on the next
+        # LogManager.load() (server tool workers, resumed CLI sessions).
+        if view is None:
+            self._restore_current_view()
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 8479975df — `LogManager` only restores `views/.current` when loading the main branch. An explicit non-main branch keeps `current_view` unset so appends write to that branch, not dual-write to main+view. Regression covers load+append on a feature branch.

TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -1327,6 +1353,22 @@ def stream_tool_output(
...
+            # Compact after tool results, before the continuation provider call.
+            # Serialize with the same conversation lock as tool-result appends
+            # and continuation election so concurrent workers cannot race view
+            # creation or the active-view marker.
+            try:
+                with SessionManager.conversation_lock(conversation_id):
+                    manager = LogManager.load(
+                        conversation_id, branch=branch, lock=False
+                    )
+                    _compact_after_tool_results(manager, session, conversation_id)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 4ed10d5e4 — post-tool compaction still holds `conversation_lock` for load, view creation, and status appends, but `_resume_via_llm` now drops that lock around `llm.reply`. A regression test probes the lock from another thread during the provider call.

TimeToBuildBob:
@greptileai review

gptme/tools/autocompact/resume.py
```diff
@@ -211,7 +235,8 @@ def _resume_via_llm(
             hide=use_view_branch,
         )
         return
-    resume_response = llm.reply(llm_msgs, model=m.full, tools=[], workspace=None)
+    with llm_unlocked or nullcontext():
+        resume_response = llm.reply(llm_msgs, model=m.full, tools=[], workspace=None)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ec82e44cb — after `llm.reply` the summarizer compares `current_view`, log length, and last message against the pre-call snapshot. If another worker mutated the conversation, the summary is discarded instead of creating a view from the stale snapshot.

TimeToBuildBob:
@greptileai review

gptme/init.py
TimeToBuildBob: The fingerprint sits on `gptme/init.py` (always-on hook registration), but the exfil path is closed in `_load_context_files`. At head ec82e44 that helper resolves `~/`, absolute, `../`, and symlink paths and skips anything outside the workspace (`is_relative_to`). Regression: `test_load_context_files_rejects_paths_outside_workspace` and `test_load_context_files_rejects_symlink_escape`.

gptme/executor.py
TimeToBuildBob: The fingerprint sits on `gptme/executor.py` (hook registration in the server env). Durability is in `LogManager`: at head ec82e44, `switch_view` writes `views/.current`, and `LogManager.load()` restores that marker only on the main branch. Reloaded managers keep the compacted projection instead of uncompressed master.

gptme/logmanager/manager.py
TimeToBuildBob: Fixed in 1be63c387. At head ec82e44 the server tool worker runs `autocompact_hook` after tool results are appended and before the continuation request (`_compact_after_tool_results`). The reentrancy throttle is per-conversation and retries when the log grew.

gptme/tools/autocompact/engine.py
TimeToBuildBob: Fixed in 1be63c387. At head ec82e44 the engine stops only when `current_tokens < target_tokens`, so an exact-budget log is still trimmed. The regression mocks token counts for every equivalent list, not just the original `messages` object.

TimeToBuildBob:
## Merge recommendation (Greptile attempt-cap adjudication)

Verified against: `ec82e44cb7aa885a616e8c0f91e8316f2ff2c063` (`fix-3812`).
Greptile last reviewed that same commit (summary footer, 5/5: no newly established actionable issue).
CI: green. Mergeable: CLEAN.

### Fixed
Prior sessions already landed the code. This session verified those fixes at head and closed the leftover threads:

- P0 summarizer path containment (`_load_context_files` workspace-only, including `~/`, absolute, `../`, symlink escape) — `1be63c387`
- Durable compacted views (`views/.current`, restore only on main) — `1be63c387` / `8479975df`
- Post-tool compaction before continuation — `1be63c387`
- Exact-budget engine trim (`current_tokens < target`) — `1be63c387`
- One budget end-to-end (clamp explicit budgets; pass budget into `CompressionConfig.limit`; tool-qualified hook name) — `e86d57651` / `4902f57f9`
- Serialize post-tool compaction on `conversation_lock`; drop that lock around `llm.reply`; discard stale summaries — `8479975df` / `4ed10d5e4` / `ec82e44cb`

### Remaining
None blocking. One accepted Phase 1 trade-off already disposed: compaction trigger counts the stored log, not later `enrich_messages_with_context` (ephemeral, cannot be compacted; `limit_log` still caps the provider window).

Our AI-reviewer: no open findings. Greptile: 8 leftover threads resolved after the replies above; detector verdict `converged`.

### Convergence
`round_convergence.stable_rounds`: **1** (required 2). `should_request_review_after_fixes`: **false** (attempt cap). This session did **not** trigger another Greptile pass. Latest Greptile summary on current head is 5/5 with no new actionable issue.

### Domain risk
Always-on compaction is now core conversation lifecycle, including the LLM summarize path and server-side view persistence under concurrent tool workers. Worth a maintainer smoke-test: a long server conversation with large tool results, plus one `/compact`/summarize while another tool is in flight.

### Not doing
- No further `@greptileai review` (cap).
- No self-merge. New `[context] budget` schema key, new `gptme/util/context_budget.py`, and server lock/view behavior are Erik-gated.
- No rebase (already MERGEABLE/CLEAN).

Maintainer judgment from here.

PR #3898, opened 2026-09-20
feat(shell): promote slow foreground commands

TimeToBuildBob:
## Summary

- promote foreground shell commands that outlive `GPTME_SHELL_FOREGROUND_TIMEOUT` (120s by default) into the existing conversation-owned job registry
- preserve the command's partial output, tracked cwd, eventual completion event, controls, cleanup, and the existing `GPTME_SHELL_TIMEOUT` hard limit
- lazily seed a replacement foreground shell so the next tool call can proceed while the promoted command finishes
- propagate the new setting into sandboxed shells

## Why

A 30-day trajectory audit found 703 gptme shell timeouts, mostly ordinary long-running tests/searches/builds. Claude Code promotes the equivalent calls instead of killing them. The existing background-job delivery path now provides the missing ownership and completion-event primitives, so foreground promotion can reuse that contract instead of adding another monitor mechanism.

## Verification

- `poetry run pre-commit run --files gptme/tools/shell.py gptme/tools/shell_background.py gptme/sandbox.py tests/test_shell_foreground_promotion.py`
- `poetry run pytest tests/test_shell_foreground_promotion.py tests/test_shell_background.py tests/test_shell_completion_loop.py tests/test_shell_completion_wait.py tests/test_foreground_output_dedup.py -q --timeout=10`
- targeted shell/sandbox suite launched separately; any environment-only Docker failures are documented in the PR follow-up

## Scope

This deliberately does **not** restore the old textual `bg` overlay or add a separate busy-session abstraction. Structured `background: true` remains the explicit path; automatic promotion makes the existing persistent shell the job and releases a lazily-created replacement shell at the tracked cwd.


TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -2687,25 +2765,112 @@ def execute_shell_impl(
...
+
+        def run_foreground() -> None:
+            try:
+                result["value"] = shell.run(cmd, output=False, timeout=timeout)
+            except BaseException as exc:
+                error.append(exc)
+
+        worker = threading.Thread(target=run_foreground, daemon=True)
+        worker.start()
+        worker.join(foreground_timeout)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6615d7db3. Ctrl-C during the soft-timeout wait now uses the same `_terminate_interrupted_shell` cleanup as the synchronous path, then joins the worker (and SIGTERM descendants if it is still alive) so a later tool call cannot reuse a busy shell.

gptme/tools/shell.py
```diff
@@ -2219,6 +2284,19 @@ def set_shell(shell: ShellSession) -> None:
     _shell_var.set(shell)
 
 
+def _replace_promoted_shell(shell: ShellSession) -> None:
+    """Release a busy shell and seed the next shell with its tracked cwd."""
+    shell.detach()
+    _workspace_cwd.set(str(shell.get_cwd()))
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6615d7db3. Replacement shells are seeded from the live process cwd (`/proc/<pid>/cwd`, `lsof` on macOS) so an in-flight `cd dest; sleep ...` is visible before the PWD marker lands.

gptme/tools/shell_background.py
```diff
@@ -314,6 +332,43 @@ def _get_background_job(
...
+
+
+def complete_background_job(job: BackgroundJob, stdout: str, stderr: str) -> None:
+    """Store final output and publish one completion for a registered job."""
+    with job._buffer_lock:
+        job._output_callback = None
+        if stdout:
+            job._append_to_buffer_locked(job.stdout_buffer, stdout)
+        if stderr:
+            job._append_to_buffer_locked(job.stderr_buffer, stderr)
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6615d7db3. `_append_to_buffer_locked` now truncates oversized single appends to the 1 MiB tail, so a promoted 32 MiB completion cannot pin the background registry.

gptme/tools/shell.py
```diff
@@ -267,10 +275,12 @@ def trim_blank_lines(text: str) -> str:
...
-conversation-owned process. The call returns a job ID immediately and completion
-is reported automatically. Exact `jobs`, `output <id> [--new]`, `wait <id>
-[timeout]`, and `kill <id>` calls manage matching harness jobs; otherwise Bash
-owns those commands. Use this for dev servers and long builds.
+conversation-owned process. Foreground calls still running after
+`GPTME_SHELL_FOREGROUND_TIMEOUT` (default 120 seconds) are promoted automatically.
+The call returns a job ID and completion is reported automatically. Exact `jobs`,
+`output <id> [--new]`, `wait <id> [timeout]`, and `kill <id>` calls manage matching
+harness jobs; otherwise Bash owns those commands. Use explicit background mode for
+dev servers and long builds.
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 6615d7db3. Instructions now tell the agent to use explicit `background: true` for known long work and to keep going after automatic promotion, instead of documenting job IDs and control syntax.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Confidence 2/5

<details><summary><b>5 findings</b> · ❌ 2 P1 · ⚠️ 3 P2</summary>

**❌ P1 high** — `gptme/tools/shell.py:2307`

Promotion silently discards the persistent shell's state. `_replace_promoted_shell` transfers only the cwd and clears `_shell_var`, so the next `get_shell()` creates a fresh process without exports established by prior commands, such as an activated virtualenv or a modified `PATH`. Because promotion is automatic by default, a long command unexpectedly breaks the documented stateful-shell contract. The replacement should restore the existing state snapshot, or at minimum explicitly report the state loss.

**❌ P1 high** — `gptme/tools/shell.py:2305`

A detached CLI shell can still change the process-wide cwd after promotion. CLI shells start with `_skip_os_chdir=False`, and the worker's copied context still has no workspace value; consequently, when a promoted script later reaches its delimiter after running `cd`, `_set_cwd()` calls `os.chdir()` from the background worker. By then the replacement shell may already have moved elsewhere, so the global cwd jumps asynchronously and relative operations in other tools can target the wrong directory. Detaching/promoting the old shell must disable all subsequent process-wide cwd synchronization.

**⚠️ P2 medium** — `gptme/tools/shell_background.py:375`

Finalizing a promoted job ignores the number of characters evicted by `_append_to_buffer_locked`. Unlike `_append_output`, this path does not advance `_stdout_buffer_start` or `_stderr_buffer_start`. After `output <id> --new` has recorded an offset while the job is running, a completion larger than `_MAX_BUFFER_SIZE` is tail-truncated but still appears to begin at offset zero, causing later incremental reads to skip some or all of the retained final output.

**⚠️ P2 medium** — `gptme/tools/shell.py:2843`

The kill path races with the publisher when setting `PromotedJobProcess.returncode`. `kill_promoted()` and `publish()` can concurrently call `finish()`, whose `is_set()`/assignment/`set()` sequence is not protected by a lock. Both callers can pass the check, allowing the return code to change after waiters have been awakened and making completion nondeterministically report success or SIGTERM. `finish()` needs an atomic, locked first-writer-wins update.

**⚠️ P2 medium** — `gptme/tools/shell.py:2786`

Foreground promotion is unconditionally disabled on Windows even though the new configuration documentation presents `GPTME_SHELL_FOREGROUND_TIMEOUT` as platform-independent. On this supported shell platform, long foreground commands therefore continue blocking until the hard timeout and the new setting has no effect. Either implement a Windows termination/promotion path or document and validate the feature as POSIX-only.

</details>

<sub>Reviewed `2ace5f5bf06b` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 533s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/tools/shell.py
```diff
@@ -2219,6 +2300,19 @@ def set_shell(shell: ShellSession) -> None:
     _shell_var.set(shell)
 
 
+def _replace_promoted_shell(shell: ShellSession) -> None:
+    """Release a busy shell and seed the next shell with its live cwd."""
+    shell.detach()
+    _workspace_cwd.set(str(shell.live_cwd()))
+    _shell_var.set(None)
```
TimeToBuildBob, line 2457: ❌ **P1** — Promotion silently discards the persistent shell's state. `_replace_promoted_shell` transfers only the cwd and clears `_shell_var`, so the next `get_shell()` creates a fresh process without exports established by prior commands, such as an activated virtualenv or a modified `PATH`. Because promotion is automatic by default, a long command unexpectedly breaks the documented stateful-shell contract. The replacement should restore the existing state snapshot, or at minimum explicitly report the state loss.

gptme/tools/shell.py
```diff
@@ -2219,6 +2300,19 @@ def set_shell(shell: ShellSession) -> None:
     _shell_var.set(shell)
 
 
+def _replace_promoted_shell(shell: ShellSession) -> None:
+    """Release a busy shell and seed the next shell with its live cwd."""
+    shell.detach()
```
TimeToBuildBob, line 2454: ❌ **P1** — A detached CLI shell can still change the process-wide cwd after promotion. CLI shells start with `_skip_os_chdir=False`, and the worker's copied context still has no workspace value; consequently, when a promoted script later reaches its delimiter after running `cd`, `_set_cwd()` calls `os.chdir()` from the background worker. By then the replacement shell may already have moved elsewhere, so the global cwd jumps asynchronously and relative operations in other tools can target the wrong directory. Detaching/promoting the old shell must disable all subsequent process-wide cwd synchronization.

gptme/tools/shell_background.py
```diff
@@ -314,6 +341,43 @@ def _get_background_job(
...
+        _jobs_for(conversation_id)[job.id] = job
+    return job
+
+
+def complete_background_job(job: BackgroundJob, stdout: str, stderr: str) -> None:
+    """Store final output and publish one completion for a registered job."""
+    with job._buffer_lock:
+        job._output_callback = None
+        if stdout:
+            job._append_to_buffer_locked(job.stdout_buffer, stdout)
```
TimeToBuildBob: ⚠️ **P2** — Finalizing a promoted job ignores the number of characters evicted by `_append_to_buffer_locked`. Unlike `_append_output`, this path does not advance `_stdout_buffer_start` or `_stderr_buffer_start`. After `output <id> --new` has recorded an offset while the job is running, a completion larger than `_MAX_BUFFER_SIZE` is tail-truncated but still appears to begin at offset zero, causing later incremental reads to skip some or all of the retained final output.

gptme/tools/shell.py
```diff
@@ -2687,25 +2781,144 @@ def execute_shell_impl(
...
+        else:
+            if worker.is_alive():
+                # The command is already executing in ``shell``. Register that
+                # shell as the job and let the existing worker publish its
+                # eventual result.
+                process = PromotedJobProcess()
+
+                def kill_promoted() -> None:
+                    _kill_descendants(shell.process.pid)
+                    process.finish(-signal.SIGTERM)
```
TimeToBuildBob, line 2994: ⚠️ **P2** — The kill path races with the publisher when setting `PromotedJobProcess.returncode`. `kill_promoted()` and `publish()` can concurrently call `finish()`, whose `is_set()`/assignment/`set()` sequence is not protected by a lock. Both callers can pass the check, allowing the return code to change after waiters have been awakened and making completion nondeterministically report success or SIGTERM. `finish()` needs an atomic, locked first-writer-wins update.

gptme/tools/shell.py
```diff
@@ -2687,25 +2781,144 @@ def execute_shell_impl(
     allowlisted = is_allowlisted(cmd, cwd=shell.get_cwd())
 
     start_time = time.monotonic()
-    try:
-        returncode, stdout, stderr = shell.run(cmd, timeout=timeout)
+    foreground_timeout = _get_foreground_timeout()
+    promoted = (
+        not _is_windows
```
TimeToBuildBob, line 2937: ⚠️ **P2** — Foreground promotion is unconditionally disabled on Windows even though the new configuration documentation presents `GPTME_SHELL_FOREGROUND_TIMEOUT` as platform-independent. On this supported shell platform, long foreground commands therefore continue blocking until the hard timeout and the new setting has no effect. Either implement a Windows termination/promotion path or document and validate the feature as POSIX-only.

gptme/tools/shell.py
TimeToBuildBob: Fixed in f8e5a3e25. Promotion now copies the last completed-command env snapshot into `_promoted_next_state` and restores it when the replacement shell is created, so exports such as an activated venv/PATH survive automatic promotion.

gptme/tools/shell.py
TimeToBuildBob: Fixed in f8e5a3e25. `detach()` now sets `_skip_os_chdir` so the worker cannot `os.chdir` after a later `cd`. The live cwd is seeded through `_promoted_next_cwd` instead of mutating `_workspace_cwd` (the server skip-chdir flag).

gptme/tools/shell_background.py
TimeToBuildBob: Fixed in f8e5a3e25. `complete_background_job` now advances `_stdout_buffer_start` / `_stderr_buffer_start` by the characters evicted during tail truncation, so later `output <id> --new` reads still see the retained tail.

gptme/tools/shell.py
TimeToBuildBob: Fixed in f8e5a3e25. `PromotedJobProcess.finish()` is now a locked first-writer-wins update, so `kill_promoted()` and `publish()` cannot clobber each other's return code.

gptme/tools/shell.py
TimeToBuildBob: Documented in f8e5a3e25 as POSIX-only. Windows has no process-group promotion path; `GPTME_SHELL_FOREGROUND_TIMEOUT` is ignored there and long commands wait for `GPTME_SHELL_TIMEOUT`.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -2219,6 +2319,65 @@ def set_shell(shell: ShellSession) -> None:
...
+def _replace_promoted_shell(shell: ShellSession) -> None:
+    """Release a busy shell and seed the next shell with its cwd and exports.
+
+    Do not write the live cwd into ``_workspace_cwd``. That ContextVar is the
+    server skip-chdir flag; mutating it would make later CLI shells skip
+    process-wide chdir, and the worker already has a copied context so it
+    would not see the write. Creating the replacement here would also block
+    the promotion return on ``_init``.
+    """
+    shell.detach()
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in ed1a8690b. A DEBUG trap now writes the cwd/env snapshot immediately before a non-builtin command, so `export API_URL=...; sleep 180` is visible to the replacement shell at promotion time rather than waiting for the post-command snapshot that only runs after sleep finishes.

TimeToBuildBob:
@greptileai review

gptme/tools/shell.py
```diff
@@ -2219,6 +2325,144 @@ def set_shell(shell: ShellSession) -> None:
...
+    """
+    skip = "|".join(_FAST_BASH_BUILTINS)
+    body = (
+        f"case ${{BASH_COMMAND%% *}} in {skip}) ;; "
+        f"*) {_shell_state_snapshot_cmd(state_path)} ;; "
+        "esac"
+    )
+    return (
+        "__gptme_old_debug=$(trap -p DEBUG 2>/dev/null || true)\n"
+        f"trap {shlex.quote(body)} DEBUG\n"
```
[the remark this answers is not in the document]
TimeToBuildBob: Intentional: we save the previous DEBUG trap with `trap -p` and restore it after the command, so session-level tracing is not dropped. Replacing it for the duration of the user command is required so the in-flight env snapshot actually runs. Chaining would also fire user DEBUG handlers during `export -p`, which can recurse or spam traces. The original P1 (`export ...; sleep`) is covered by ed1a8690b.

gptme/tools/shell.py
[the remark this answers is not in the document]
TimeToBuildBob: The original example (`export API_URL=...; sleep 180`) is fixed in ed1a8690b and covered by `test_promoted_command_preserves_inflight_exports`. Rechecked on head `ed1a8690b07e`.

The later summary residual about skipped wrappers is not merge-blocking:

- `source env.sh; sleep` preserves the sourced exports — DEBUG fires before the sibling `sleep`. Verified on this head.
- `eval "export ...; sleep"` preserves exports — DEBUG already fires inside `eval`. Verified on this head.
- Remaining hole: `source script.sh` where the script's own body blocks after mutating env. Bash DEBUG does not enter sourced files unless `functrace`/`extdebug` is on. Enabling those for every command would inherit DEBUG into all functions, which is a larger behavior change than this PR should take at the review cap. The common form `source venv/bin/activate; long_cmd` is the sibling form and works.

Dismissing the residual as an accepted POSIX DEBUG limitation. Not starting another fix round.

TimeToBuildBob:
## Merge recommendation (convergence adjudication)

Adjudicated against head `ed1a8690b07e`. Greptile's last reviewed commit is that SHA. In-band AI review last covered `2ace5f5` (two commits behind); those findings were fixed in `f8e5a3e25` and the threads are resolved here rather than re-scored. Did not retrigger Greptile (attempt cap).

**Fixed (prior sessions, verified still present on this head):**
- P1 interrupt-during-soft-timeout — `6615d7db3`
- P1 live cwd at promotion — `6615d7db3`
- P1 promoted output bound — `6615d7db3`
- P1 replacement shell lost prior exports — `f8e5a3e25`
- P1 detached CLI shell `os.chdir` after promotion — `f8e5a3e25`
- P1 in-flight `export ...; sleep` snapshot — `ed1a8690b` (`test_promoted_command_preserves_inflight_exports`)

**Remaining (non-blocking, dismissed with reasons):**
- Greptile P1 residual "skipped wrappers (`source`/`eval`)" — `eval "export ...; sleep"` and `source env.sh; sleep` both preserve exports on this head. The leftover hole is `source script.sh` whose *own body* blocks after mutating env; bash DEBUG does not enter sourced files unless `functrace`/`extdebug` is on, and enabling that for every command inherits DEBUG into all functions. Common form `source venv/bin/activate; long_cmd` is the sibling form and works. Thread 4060119369.
- Greptile P2 DEBUG-trap suppression — intentional scoped replace + restore; Greptile agreed the finding does not apply. Thread 4060378827.
- AI-review P2 Windows promotion — documented POSIX-only in `f8e5a3e25`.
- AI-review P2 buffer-offset / finish-race — fixed in `f8e5a3e25`.

All Greptile and in-band review threads are resolved.

**CI:** green on `ed1a8690b07e` (lint, typecheck, tests without API keys, Windows path slice, Docker, PyInstaller, Greptile Review). Deploy skipped (expected).

**Domain risk:** foreground promotion + env restore is POSIX process-group machinery. Worth a maintainer manual check of `export FOO=bar; sleep 180` then a follow-up command seeing `FOO`. Do not need to exercise `source`-of-a-blocking-script; that residual is accepted.

**Convergence:** `round_convergence.status = no_current_blocking`, `stable_rounds = 1` / `required_stable_rounds = 2`, `should_request_review_after_fixes = false`. Detector verdict after thread resolution: `converged` (0 unresolved P1/P2). Latest Greptile round (`review:5264455039`) already had `p1_count = 0`.

Ready for maintainer judgment — merge-ready does not mean auto-merge. This is a `feat` without an issue reference, so the self-merge gate stays closed anyway.

PR #3901, opened 2026-09-21
fix(cli): import extras lazily so gptme-server and gptme-dspy start without them

### LINKED ISSUE #290: bug: gptme-server fails with `No module named 'flask'`
Perhaps this is an easy fix, but I didn't see anything for it in the docs.

```
$ gptme-server
  File "/home/filip/.local/bin/gptme-server", line 5, in <module>
    from gptme.server.cli import main
  File "/home/filip/.local/share/pipx/venvs/gptme/lib/python3.12/site-packages/gptme/server/__init__.py", line 5, in <module>
    from .api import create_app
  File "/home/filip/.local/share/pipx/venvs/gptme/lib/python3.12/site-packages/gptme/server/api.py", line 17, in <module>
    import flask
ModuleNotFoundError: No module named 'flask'
```

vanshtaneja23:
# fix(cli): import extras lazily so gptme-server and gptme-dspy start without them

Closes #290 (the `gptme-server` half — the check added then is no longer reachable).

## Problem

Two console scripts reach an optional dependency at module scope, so Python fails the entry-point import before any gptme code runs. On a plain `pip install gptme` (0.34.0 and current master, 4bd2560):

```
$ gptme-server
  File ".../bin/gptme-server", line 5, in <module>
    from gptme.server.cli import main
  File ".../gptme/server/cli.py", line 56, in <module>
    from .app import create_app
  File ".../gptme/server/app.py", line 11, in <module>
    import flask
ModuleNotFoundError: No module named 'flask'

$ gptme-dspy
  File ".../gptme/eval/dspy/prompt_optimizer.py", line 12, in <module>
    import dspy
ModuleNotFoundError: No module named 'dspy'
```

Both places already have a check for exactly this, and both are dead code:

- `gptme/server/cli.py` — `main()` tests `__import__("flask")` and prints `Install them with pip install gptme[server]`, but `from .app import create_app` two lines above the group definition raises first. This is what #290 reported; the check added then stopped being reachable as the module-level imports grew.
- `gptme/eval/dspy/__init__.py` — defines `_has_dspy()`, but `from .cli import main` in the same file imports dspy first. `tests/test_dspy_basic.py` and `tests/test_dspy_hybrid.py` both skip on `_has_dspy()`, so they cannot run on an install without the extra either.

`gptme-acp` is the shape that works today — `gptme/acp/client.py` defers its import and prints `Install with: pip install 'gptme[acp]'`. This makes the other two behave the same way.

## Change

Move the flask and dspy imports into the commands that use them (`serve`, `token`, `openapi`, and the `gptme-dspy` entry function). No behaviour change when the extras are installed.

## Verification

Clean venv, `pip install .` with no extras, current master:

```
$ gptme-server
gptme installed without needed extras for server. Install them with `pip install gptme[server]`
$ gptme-server token
gptme installed without needed extras for server. Install them with `pip install gptme[server]`
$ gptme-server --help          # now works without the extra
$ gptme-dspy
ERROR - gptme.eval.dspy - gptme installed without needed extras for dspy. Install them with `pip install gptme[dspy]`
$ python -m gptme.eval.dspy    # same
```

Checks on the changed files: `ruff format --check`, `ruff check`, `mypy` all clean (ruff 0.16.0, mypy 2.3.0, as pinned in `poetry.lock`).

## Test

`tests/test_entrypoints.py` imports every `[project.scripts]` target in a subprocess with every `optional = true` distribution blocked by a `sys.meta_path` finder. Because the blocking is explicit, it reproduces the failure on a machine that has all the extras installed — which is why CI never caught this.

```
# on master
2 failed, 24 passed        -> gptme-dspy, gptme-server
# with this change
26 passed                  in 1.0s
```

Runs in about a second and covers entry points added later, not just these two.


ErikBjare:
@TimeToBuildBob how come CI doesn't catch this? It has a extra-free job for testing these exact things. Worth a follow-up PR (unless the PR test is enough)

TimeToBuildBob:
The extra-free job tests the pytest suite on a minimal install. It never runs the console scripts.

`test-minimal` (`Test without API keys or extras`) does `poetry install` (no extras) then `make test`. Every `tests/test_server*.py` file starts with `pytest.importorskip("flask", ...)`, so those modules drop out at collection — xdist never even lists `tests/test_server.py`. DSPy tests skip the same way (`_has_dspy()` / `ImportError`). A successful extra-free run on master (`4bd2560`) is `11788 passed, 213 skipped` with zero `tests/test_server.py` / `tests/test_dspy*` lines in the log.

The extras-check in `gptme.server.cli.main()` is unreachable for the same reason the user crash is: the generated wrapper does `from gptme.server.cli import main` before any gptme code runs. `gptme-dspy` is the same shape (`gptme.eval.dspy.__init__` imports `.cli` which imports `dspy` at module scope, so `_has_dspy()` is dead).

The `-E all` job has flask/dspy installed, so the import succeeds there too. Nothing else in CI executes `gptme-server` / `gptme-dspy` as entry points — the Windows "CLI entrypoints" slice is `gptme --help` / `--version`, and the PyInstaller `--help` check is a bundled binary with extras baked in.

This PR's `tests/test_entrypoints.py` is enough. It imports every `[project.scripts]` target in a subprocess with optional dists blocked, so it fails on master for exactly these two and still fails on a fully-provisioned runner. A follow-up that adds `gptme-server --help` to `test-minimal` would be narrower than that test, so I'm not opening one.

Greptile's P2 on dist-name → import-name fallback (`agent-client-protocol` vs `acp`) is real but not a hole here: uninstalled extras still fail the import naturally, and the `-E all` job uses `packages_distributions()`.

Issue #3903, opened 2026-09-21
fix(webui): honor Max tokens on a new conversation initial step


TimeToBuildBob:
## Reproduction

Tested the released Linux AppImage `v0.34.1.dev20260921` from a clean HOME against its bundled sidecar:

1. Open a new conversation.
2. Open **Options** and set **Max tokens** to `64` (the field visibly retains `64`).
3. Send `Reply exactly: packaged desktop reply works` with the default OpenRouter-backed model.

## Actual

The provider rejects the request with HTTP 402 and reports that the request asked for up to **16,000 tokens**. Repeating from a fresh conversation and changing the option before the first send produces the same result.

The raw upstream error metadata is also rendered as a very large destructive toast that fills most of the app window.

## Expected

The initial `/step` request should carry `max_tokens: 64`, so the visible per-chat control governs the first generation as advertised.

## Control

The same packaged app, bundled sidecar, clean HOME, and network path returned the exact requested reply after selecting `openrouter/z-ai/glm-5.2:free`. That rules out the desktop transport and provider-key wiring; the failure is specific to the requested output cap not reaching the paid-model call.

This was found while verifying the post-gptme/gptme#3882 AppImage first-run path.

TimeToBuildBob:
Fixed in gptme/gptme#3905 (merged 2026-09-23). All CI checks pass — the server now forwards the per-step `max_tokens` override from the WebUI through the initial step, tool continuations, and ACP paths.

PR #3905, opened 2026-09-22
fix(server): honor per-step max token override

TimeToBuildBob:
## Summary

- accept and validate max_tokens on POST /api/v2/conversations/{id}/step
- forward the request-scoped override through the worker to the provider call
- preserve the conversation ChatConfig fallback when the request omits the override
- document max_tokens in the StepRequest OpenAPI schema

## Verification

- Regression test: failed before the fix with None instead of 64
- 169 server/session tests passed
- 6 focused max_tokens tests passed with strict thread-leak checking
- Ruff check and format passed
- mypy passed on the affected server modules

Addresses gptme/gptme#3903.


TimeToBuildBob:
@greptileai review

gptme/server/session_step.py
```diff
@@ -706,11 +707,16 @@ def step(
...
+        max_tokens: Optional request-scoped response limit. Falls back to the
+            conversation config when omitted.
     """
 
     # Load chat config and prepare execution environment
     logdir = get_logs_dir() / conversation_id
     chat_config = ChatConfig.load_or_create(logdir, ChatConfig())
+    effective_max_tokens = (
+        max_tokens if max_tokens is not None else chat_config.max_tokens
+    )
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0f95c3b8d. Auto-confirm tool continuations now forward the request-scoped max_tokens into the continuation worker.

gptme/server/api_v2_sessions.py
```diff
@@ -334,6 +334,14 @@ def api_conversation_step(conversation_id: str):
     if model is not None and not isinstance(model, str):
         return flask.jsonify({"error": "model must be a string"}), 400
 
+    max_tokens = req_json.get("max_tokens")
```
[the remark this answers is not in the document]
TimeToBuildBob: Fixed in 0f95c3b8d. /step now forwards max_tokens into the ACP worker and through ACP _meta so the gptme-acp agent honors it on the provider call.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
## 🤖 AI code review

Confidence 2/5

<details><summary><b>2 findings</b> · ❌ 2 P1</summary>

**❌ P1 high** — `gptme/acp/client.py:400`

The ACP extension metadata is passed at the wrong abstraction level. `ClientSideConnection.prompt()` collects its extra keyword arguments into the protocol `_meta` object, while the agent expects the resulting extension as `kwargs["gptme"]`. Passing an `_meta` keyword here therefore nests the payload as `_meta._meta.gptme`; the agent's new `kwargs.get("gptme")` lookup never sees it and silently uses no request limit. The test mocks the connection and consequently only verifies this incorrect intermediate call, not an ACP round trip. Pass `gptme={"max_tokens": max_tokens}` to the connection (or explicitly support the nested form in the agent).

**❌ P1 high** — `gptme/server/session_step.py:1122`

The request-scoped limit is preserved only for auto-confirmed tools. If the initial response leaves a pending tool, `ToolExecution` stores no `max_tokens`; the later confirm/edit/skip handlers call `start_tool_execution()` without this new argument, so the continuation reaches `_start_step_thread()` with `None` and falls back to `ChatConfig.max_tokens`. Thus a step requested with (for example) 64 tokens can use the configured/default limit after an ordinary manual tool confirmation. Persist the effective limit with the pending tool and forward it through every confirmation continuation path.

</details>

<sub>Reviewed `0f95c3b8dd1c` · `openai-subscription/gpt-5.6-sol` · `agent` engine · 439s · [about this reviewer](https://github.com/gptme/gptme-contrib/blob/master/docs/ai-reviewer.md)</sub>

</details>

gptme/acp/client.py
```diff
@@ -392,9 +396,13 @@ async def prompt(
         from acp.schema import TextContentBlock
 
         prompt_content = [TextContentBlock(type="text", text=message)]
+        prompt_kwargs: dict[str, Any] = {}
+        if max_tokens is not None:
```
TimeToBuildBob, line 401: ❌ **P1** — The ACP extension metadata is passed at the wrong abstraction level. `ClientSideConnection.prompt()` collects its extra keyword arguments into the protocol `_meta` object, while the agent expects the resulting extension as `kwargs["gptme"]`. Passing an `_meta` keyword here therefore nests the payload as `_meta._meta.gptme`; the agent's new `kwargs.get("gptme")` lookup never sees it and silently uses no request limit. The test mocks the connection and consequently only verifies this incorrect intermediate call, not an ACP round trip. Pass `gptme={"max_tokens": max_tokens}` to the connection (or explicitly support the nested form in the agent).

gptme/server/session_step.py
```diff
@@ -1097,6 +1119,7 @@ def start_tool_execution(
     *,
     reserved: bool = False,
     branch: str = "main",
+    max_tokens: int | None = None,
```
TimeToBuildBob, line 1123: ❌ **P1** — The request-scoped limit is preserved only for auto-confirmed tools. If the initial response leaves a pending tool, `ToolExecution` stores no `max_tokens`; the later confirm/edit/skip handlers call `start_tool_execution()` without this new argument, so the continuation reaches `_start_step_thread()` with `None` and falls back to `ChatConfig.max_tokens`. Thus a step requested with (for example) 64 tokens can use the configured/default limit after an ordinary manual tool confirmation. Persist the effective limit with the pending tool and forward it through every confirmation continuation path.

gptme/acp/client.py
TimeToBuildBob: Fixed in e5fe60900. ACP prompt now passes `gptme={"max_tokens": ...}` so ClientSideConnection packs it as protocol `_meta.gptme`, and the agent sees `kwargs["gptme"]`.

gptme/server/session_step.py
TimeToBuildBob: Fixed in e5fe60900. `ToolExecution` now stores the request-scoped `max_tokens`, and confirm/edit/skip/auto continuations forward it into the next provider call.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
CI failure on "Test without API keys" was an Ubuntu runner apt 404 (`ghostscript` 10.02.1~dfsg1-0ubuntu7.8 missing from a stale index), not a test regression. The extras job already passed on the previous SHA.

Pushed c9c5fe858: `apt-get update` before the runner package install so the job resolves current noble-updates packages instead of 404ing.

TimeToBuildBob:
@greptileai review

TimeToBuildBob:
Rebased onto current master (`be343267b`).

Conflict was only `.github/workflows/test.yml` on the last commit (`ci(test): apt-get update before installing runner packages`). Master already has the same fix from gptme/gptme#3914 (`apt-get update && apt-get install -y`), so I dropped that commit instead of re-applying a weaker version without `-y`.

The five feature commits are unchanged. HEAD is now `d822707c9`.

TimeToBuildBob:
@greptileai review

